PR #29059, opened 2022-03-09
*DO NOT SUBMIT* Testing behavior when adding an empty CRL from a different issuer/CA cert.

krestofur:
*DO NOT SUBMIT*
PR is just for testing behavior and sharing code.

Changes:
* add new CA cert
* create an empty CRL file from that CA cert
* openssl rehash 
* Add CRL file to CRL directory of `crl_ssl_transport_security_test` and see if it still passes.


markdroth:
This looks abandoned. Please let me know if that's not correct.

Issue #29048, opened 2022-03-08
Interop Cloud-to-Prod and Cloud-to-Prod Tests (internal CI) broken


drfloob:
Possibly an upstream bug. Example: https://source.cloud.google.com/results/invocations/4af24346-95f8-48c9-b432-ac9dc7e8b3cd/log. This has been failing continuously today, starting with an unrelated PR.

The failure is is in the `build_docker_node` step.

Suspicious log:
```
> node ./_postinstall.js

sh: 1: node: Permission denied
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.2.7 (node_modules/gulp/node_modules/chokidar/node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.13: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

npm ERR! Maximum call stack size exceeded
...
2022-03-07 17:51:51,396 FAILED: build_docker_node [ret=127, pid=7796, time=186.8sec]
```

donnadionne:
This is now fixed and tests are passing again

PR #29062, opened 2022-03-09
Use a non-root user in Node interop docker images

murgatroid99:
A recent interop build error, demonstrated in [this log](https://source.cloud.google.com/results/invocations/bd2a383a-8194-4a9a-8f31-527949b45197/targets/github%2Fgrpc%2Finterop_docker_build/tests;group=tests;test=build_docker_nodepurejs;row=1), appears to be the result of a change in the `gulp` npm package triggering nvm-sh/nvm#1407. The recommended solution in that bug, and the only one that I could get to work, was to use a non-root user. I tested it locally, and with the change, `npm install -g gulp` succeeds.

While I was modifying these docker files, I took the opportunity to switch to a newer version of nvm and the newest LTS version of Node. The previous installation of multiple versions of Node was there to support unit tests, but the unit test docker images for Node are no longer built using that template so that part is no longer necessary.

### DIFF
--- a/templates/tools/dockerfile/interoptest/grpc_interop_node/Dockerfile.template
+++ b/templates/tools/dockerfile/interoptest/grpc_interop_node/Dockerfile.template
@@ -18,8 +18,8 @@
   
   <%include file="../../apt_get_basic.include"/>
   <%include file="../../run_tests_python_deps.include"/>
-  <%include file="../../node_deps.include"/>
   <%include file="../../run_tests_addons.include"/>
+  <%include file="../../node_deps.include"/>
   # Define the default command.
   CMD ["bash"]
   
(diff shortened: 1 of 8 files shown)


nicolasnoble (approved):

drfloob:
@murgatroid99 can you confirm that this closes https://github.com/grpc/grpc/issues/29048?

nicolasnoble:
I'll have a stab at it too and see where we can go from there.

murgatroid99:
OK, I think I did it. The only failure in the Interop Cloud-to-Cloud Tests is a timeout in the python -> php7 timeout_on_sleeping_server test. That test was not impacted by any changes in this PR and that is the only occurrence of this failure in several test runs so I am confident that it is just a flake.

PR #29065, opened 2022-03-09
Bump version to 1.45.0-pre1 (on v1.45.x branch)

drfloob:


nicolasnoble (approved):

drfloob:
Interop failures are known, and only affect those interop tests themselves.

The bloat diff failure is due to an internal change in test config that happened after the release branch was cut.

PR #29069, opened 2022-03-10
Fixing use after free bug due to invocation gprc_call_get_peer after grpc_call_cancel

Vignesh2208:
Fixes b/222144056

@veblush


Vignesh2208:
Not sure if this fix is the best thing to do here. Would be curious to hear your thoughts

markdroth (commented):
This approach seems okay to me -- not necessarily ideal, but probbaly workable if we can't figure out something better.

Another possible option would be to make the peer string ref-counted and have the call take a ref to it, so that it is still valid even after the transport is destroyed.  That would be a more invasive change but would also be slightly more general.  Not sure whether or not it's worth doing.

Do we know that this problem exists only during cancellation?  Is there also a window in normal call finishing during which the transport is destroyed before the `grpc_call` object?  If so, the same thing could happen in that case.

PR #29049, opened 2022-03-08
Remove shutdown_cq from grpc_end2end_test_fixture

yashykt:
Having a separate cq for shutdown is unnecessary and causes issues when implementing graceful shutdown, so removing it.

PR #29070, opened 2022-03-10
ObjC: End2EndTests fix

yashykt:
Following up from https://github.com/grpc/grpc/pull/29049


dennycd (approved):
lgtm, thanks for the fix

Issue #25682, opened 2021-03-11
[C#] server sending responseheader in unary call not received by the client until the whole response is returned


alexhcheng:
### What version of gRPC and what language are you using?
2.26

### What operating system (Linux, Windows,...) and version?
Linux

### What runtime / compiler are you using (e.g. python version or version of gcc)
dotnet 3.1

### What did you do?
Client 
1. makes a unary request to the server.
2. call await response.ResponseHeaderAsync
3. call await response.ResponseAsync

Server
1. call await WriteResponseHeadersAsync()
2. do some heavy processing, then send the response

### What did you expect to see?
Client to get the header first
Then after a while get the response.

### What did you see instead?
ResponseHeaderAsync doesn't return until the server send the actual response.

drfloob:
It might be worth seeing if you could reproduce this with a more recent gRPC version. v2.26 was released in 2019. Some of the [ResponseHeaderAsync tests](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Tests/ResponseHeadersTest.cs#L65-L81) might be informative for you here as well, they exercise this behavior. I'm not sure we're looking at a bug here, this might be a better question for another forum (stackoverflow, or the grpc mailing list). @jtattermusch, please take a look, maybe my assessment is off.

alexhcheng:
hi @drfloob , i did try the latest build v2.38 and it has the same behavior.  Perhaps this is by design?  there is no way to send the responseheader back earlier in a unary call?

jtattermusch:
> hi @drfloob , i did try the latest build v2.38 and it has the same behavior. Perhaps this is by design? there is no way to send the responseheader back earlier in a unary call?

It should definitely be possible for the server to send the response headers independently from sending the first response (That is send response headers form the server side handler and receive them on the client regardless of how late you return the response from the server-side handler).
Can you please share a ready-to-run reproduction of the problem (e.g. as a github repo)? As @drfloob  has pointed out, we basically have a test that makes sure that the above works, so it would be interesting to see what is it that you're doing differently.

rburns1293:
The test below reproduces the issue. It'll only pass if the Computation delay is less than the 10s wait time of the ResponseHeadersAsync. 
```C#
public async Task ResponseHeadersAsync_UnaryCall()
        {
            helper.UnaryHandler = new UnaryServerMethod<string, string>(async (request, context) =>
            {
                await context.WriteResponseHeadersAsync(headers);
[...]

PR #29072, opened 2022-03-10
add tests for #25682, at least for server streaming and bidi streaming calls.

jtattermusch:
#25682 was confirmed for unary and client streaming calls, so at least adding tests to make sure for server streaming and duplex streaming calls things work as expected.


jtattermusch:
CC @rburns1293

jtattermusch:
Known failures:
- node interop tests are failing on master as well.

apolcyn (approved):

PR #29082, opened 2022-03-12
Add args to python-helloworld example

aaronlelevier:
Hi,

I would like to submit a PR for the Python example to have the same input/output as the [Golang QuickStart example](https://grpc.io/docs/languages/go/quickstart/).

- The Golang example accepts an optional `--name` argument

  - https://github.com/grpc/grpc-go/blob/master/examples/helloworld/greeter_client/main.go
- The Golang example outputs the received `name`
  
  - https://github.com/grpc/grpc-go/blob/master/examples/helloworld/greeter_server/main.go

@veblush

### DIFF
--- a/examples/python/helloworld/greeter_client.py
+++ b/examples/python/helloworld/greeter_client.py
@@ -15,6 +15,7 @@
 
 from __future__ import print_function
 
+import argparse
 import logging
 
 import grpc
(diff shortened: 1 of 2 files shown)


examples/python/helloworld/greeter_client.py
```diff
@@ -15,23 +15,30 @@
...
+import argparse
 import logging
 
 import grpc
 import helloworld_pb2
 import helloworld_pb2_grpc
 
 
-def run():
+def run(args):
```
lidizheng, line 26: If you are comfortable with type annotations, can you help to annotate the input argument and return?

examples/python/helloworld/greeter_server.py
```diff
@@ -24,6 +24,7 @@
 class Greeter(helloworld_pb2_grpc.GreeterServicer):
 
     def SayHello(self, request, context):
+        print("Received:", request.name)
```
lidizheng, line 27: Can we update the print message to make sure client and server print messages are similar? For examle "Greeter server received: ....".

PR #17511, opened 2018-12-14
Do not take ownership for grpc server global callbacks.

wjywbs:
The function comment of grpc::Server::SetGlobalCallbacks() states that "Can only be called once per application. Does not take ownership of callbacks". However, the implementation uses shared_ptr which takes ownership. Make it the same as ClientContext's GlobalCallbacks.

PR #29084, opened 2022-03-14
Do not take ownership for grpc server global callbacks.

wjywbs:
The function comment grpc::Server::SetGlobalCallbacks() states that "Can only be called once per application. Does not take ownership of callbacks". However, the implementation uses shared_ptr which takes ownership. Make it the same as ClientContext's GlobalCallbacks.

Previous PR #17511 was closed without merged.

@veblush


wjywbs:
Pinging @markdroth @vjpai. Thanks.

markdroth:
Craig, since this API is used mainly internally, can you please review?  Thanks!

PR #29023, opened 2022-03-04
Revert "Revert "Revert "Add Python Reflection Client"""

drfloob:
Reverts grpc/grpc#28879

This breaks portability and ARM tests.

PR #29085, opened 2022-03-14
Add Python Reflection Client

tomerv:
Implement ProtoReflectionDescriptorDatabase in Python to support
client-side reflection sevices.

See gRFC L95.

See also #29023.

@lidizheng

### DIFF
--- a/src/python/grpcio_tests/tests/reflection/_reflection_client_test.py
+++ b/src/python/grpcio_tests/tests/reflection/_reflection_client_test.py
@@ -1,0 +1,151 @@
+# Copyright 2022 gRPC authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests of grpc_reflection.v1alpha.reflection."""
+
+import unittest
+
+from google.protobuf.descriptor_pool import DescriptorPool
+import grpc
+from grpc_reflection.v1alpha import reflection
+from grpc_reflection.v1alpha.proto_reflection_descriptor_database import \
+    ProtoReflectionDescriptorDatabase
+
+from src.proto.grpc.testing import test_pb2
+# Needed to load the EmptyWithExtensions message
+from src.proto.grpc.testing.proto2 import empty2_extensions_pb2
+from tests.unit import test_common
+
+_PROTO_PACKAGE_NAME = "grpc.testing"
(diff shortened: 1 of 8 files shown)


src/python/grpcio_tests/tests/reflection/_reflection_client_test.py
```diff
@@ -0,0 +1,151 @@
...
+    def testFindMethodError(self):
+        service_name = self._SERVICE_NAMES[0]
+        service_desc = self.desc_pool.FindServiceByName(service_name)
+
+        # FindMethodByName sometimes raises a KeyError, and sometimes returns None.
+        # See https://github.com/protocolbuffers/protobuf/issues/9592
+        with self.assertRaises(KeyError):
+            res = service_desc.FindMethodByName(_INVALID_SYMBOL_NAME)
+            if res is None:
+                raise KeyError()
```
lidizheng, line 132: Can you share more detail here? Are we going to remove this test, or what improvement did we add to make it compatible?

https://github.com/grpc/grpc/pull/29023#issuecomment-1060399713

tomerv: The original test was
```
        with self.assertRaises(KeyError):
            service_desc.FindMethodByName(_INVALID_SYMBOL_NAME)
```
I changed it to translate a `None` return value to a `raise KeyError`.

lidizheng: Thanks for the clear write-up in https://github.com/protocolbuffers/protobuf/issues/9592.

Checking other find* methods ([Python code](https://github.com/protocolbuffers/protobuf/blob/349d74d92e8268c41fbb726b0a8292d588c63e20/python/google/protobuf/descriptor_database.py)), they favor the raise behavior. Do you know how the C-Extension implementation will behave? Based on [code](https://github.com/protocolbuffers/protobuf/blob/2d69d44cc18beaa0a3a62d353f80aead669abb5c/python/google/protobuf/pyext/descriptor_database.cc#L113), their return type is `bool`.

tomerv: In the code you linked, the return type is `bool` but internally it calls a Python function (`PyObject_CallMethod(...)`). In this context we have a call chain that look like `Python -> C++ -> Python`, where the outer Python code is the one doing the call `desc_database.FindFileByName(...)` and the inner is the `FindFileByName` implementation. I'm not an expert on this, but I believe that if the inner Python code raises an exception then it is immediately caught by the outer Python code - it doesn't even do exception folding of that C++ function, just sort of steals the context.

If the exception is not caught by the outer Python code, it looks like this:
```
  File "test.py", line 12, in <module>
    desc_pool.FindFileByName("foo")
KeyError: "Couldn't find file foo"
```
Note how all the traceback of the inner `C++ -> Python` is missing as result of this unusual flow.

I hope this answers your question.

lidizheng (approved):
LGTM.

tomerv:
> LGTM.

Thanks! When can this be merged?

lidizheng:
@gnossen PTAL.

@tomerv The policy was each PR requires 2 maintainers review.

gnossen (approved):

PR #29090, opened 2022-03-14
xds: accept resources wrapped in a Resource message

markdroth:
The client sends the `xds.config.resource-in-sotw` capability to let the server know it can do this.

### DIFF
--- a/src/core/ext/xds/xds_api.cc
+++ b/src/core/ext/xds/xds_api.cc
@@ -403,7 +403,7 @@ absl::Status XdsApi::ParseAdsResponse(const XdsBootstrap::XdsServer& server,
     absl::string_view serialized_resource =
         UpbStringToAbsl(google_protobuf_Any_value(resources[i]));
     // Unwrap Resource messages, if so wrapped.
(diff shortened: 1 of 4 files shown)


src/core/ext/xds/xds_api.cc
```diff
@@ -399,6 +402,23 @@ absl::Status XdsApi::ParseAdsResponse(const XdsBootstrap::XdsServer& server,
         "type.googleapis.com/");
     absl::string_view serialized_resource =
         UpbStringToAbsl(google_protobuf_Any_value(resources[i]));
+    // Unwrap Resource messages, if so wrapped.
+    if (type_url == "envoy.service.discovery.v2.Resource" ||
```
menghanl, line 406: v2 should be `envoy.api.v2.Resource`, right?

markdroth: Good catch!  Fixed.

(I don't expect that we'll actually ever use this in v2; I added it here just for completeness.)

apolcyn (approved):

PR #29094, opened 2022-03-14
xds: reset xDS stream backoff even if all resources are invalid

markdroth:


src/core/ext/xds/xds_client.cc
```diff
@@ -1204,6 +1204,7 @@ bool XdsClient::ChannelState::AdsCallState::OnResponseReceivedLocked() {
             xds_client(), chand()->server_.server_uri.c_str(),
             status.ToString().c_str());
   } else {
+    seen_response_ = true;
```
apolcyn, line 1207: At first glance, this doesn't look quite the same as [Go](https://github.com/grpc/grpc-go/pull/5241), [Java](https://github.com/grpc/grpc-java/blob/87e13daf1a0a01afafb3e0f2818e02fe5c624ad1/xds/src/main/java/io/grpc/xds/AbstractXdsClient.java#L420), or [envoy](https://github.com/envoyproxy/envoy/blob/82b1f66df20d9577a24ea632f149481ba2f99edd/source/common/config/grpc_stream.h#L92), because we're resetting backoff only if the response message parsed.

But I take it we are doing that here because those other implementations are implicitly doing that?

apolcyn (approved):

src/core/ext/xds/xds_client.cc
markdroth: Yes, I believe that in all of those other implementations, the protobuf deserialization will happen before they consider the message to be received; the read will basically fail if the protobuf deserialization fails.

In any case, I think that's the right thing to do, because it's a completely reasonable sanity check that we've gotten a valid response from the server.  We don't have to care whether any of the resources in the response are valid, but we should care that the basic transport protocol semantics are being followed, or else we're not really in contact with a healthy server.

Issue #29098, opened 2022-03-15
PHP MemoryLeakTest consistently failing on ARM64 linux


jtattermusch:
When running PHP MemoryLeakTest on a ARM64 linux machine, valgrind reports `Syscall param epoll_ctl(event) points to uninitialised byte(s)`:

```
+ /usr/bin/valgrind --error-exitcode=10 --leak-check=yes /usr/local/bin/php -d extension_dir=/var/local/git/grpc/src/php/bin/../ext/grpc/modules -d extension=grpc.so -d max_execution_time=300 ../tests/MemoryLeakTest/MemoryLeakTest.php
==12359== Memcheck, a memory error detector
==12359== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==12359== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==12359== Command: /usr/local/bin/php -d extension_dir=/var/local/git/grpc/src/php/bin/../ext/grpc/modules -d extension=grpc.so -d max_execution_time=300 ../tests/MemoryLeakTest/MemoryLeakTest.php
==12359==
==12359== Syscall param epoll_ctl(event) points to uninitialised byte(s)
==12359==    at 0x4FCFAC8: epoll_ctl (syscall-template.S:120)
==12359==    by 0x776489F: grpc_is_epollexclusive_available (is_epollexclusive_available.cc:70)
==12359==    by 0x775EDA3: grpc_init_epollex_linux(bool) (ev_epollex_linux.cc:1630)
==12359==    by 0x776174F: try_engine (ev_posix.cc:169)
==12359==    by 0x776174F: grpc_event_engine_init() (ev_posix.cc:219)
==12359==    by 0x77647BF: iomgr_platform_init() (iomgr_posix.cc:43)
==12359==    by 0x7764113: grpc_iomgr_init() (iomgr.cc:65)
==12359==    by 0x77ED7AB: grpc_init (init.cc:185)
==12359==    by 0x745E3AF: zm_activate_grpc (in /var/local/git/grpc/src/php/ext/grpc/modules/grpc.so)
==12359==    by 0x518E4F: zend_activate_modules (zend_API.c:2600)
==12359==    by 0x4B3493: php_request_startup (main.c:1694)
==12359==    by 0x5C542B: do_cli (php_cli.c:982)
==12359==    by 0x1F982F: main (php_cli.c:1403)
==12359==  Address 0x1ffeffecd4 is on thread 1's stack
[...]

PR #29099, opened 2022-03-15
Disable PHP valgrind memoryleak test on arm64

jtattermusch:
Temporarily disable the test on ARM64 to avoid continuous failure.
See https://github.com/grpc/grpc/issues/29098


jtattermusch:
Known failures:
- node interop build is failing on master as well
- the C# errors are unrelated.

PR #29101, opened 2022-03-15
Grpc.Tools: Add macOS ARM64 support for c#

xsoheilalizadeh:
This is going to add macOS ARM64 support to the c# package, I tried to build the project with current version of bazel but it was not successful, with the 5.0.0 version I could build it successfully. 

@jtattermusch @donnadionne


xsoheilalizadeh:
Hi @jtattermusch, do you have any comments on this that could help?

Currently, [dotnet/aspnetcore](https://github.com/dotnet/aspnetcore) source code build is failing on macOS ARM64 because the gRPC package doesn't support it.

tools/bazel
```diff
@@ -42,7 +42,7 @@ fi
 
 # IMPORTANT: if you update the version here, other parts of infrastructure might needs updating as well
 # (e.g. sanity checks, bazel toolchains etc.)
-VERSION=${OVERRIDE_BAZEL_VERSION:-4.2.1}
+VERSION=${OVERRIDE_BAZEL_VERSION:-5.0.0}
```
jtattermusch, line 45: note that the protoc and grpc_charp_plugin binaries aren't being built by bazel. See https://github.com/grpc/grpc/blob/3c558dd7962f2f2aea6b6a5759f65fa6f11db847/tools/internal_ci/macos/grpc_distribtests_csharp.sh for more info.

jtattermusch (changes requested):
I think overall the intent makes sense but the problem is the actual build of the protoc and grpc_csharp_build binaries. They'd need to be built on our CI and in our CI, we currently don't have access to ARM-based macs.

Potentially, we could make use of the emulator and simply use the x64 binaries instead of the ARM64 binaries, but this is still something that would need to be tested first.

jtattermusch:
> Hi @jtattermusch, do you have any comments on this that could help?
> 
> Currently, [dotnet/aspnetcore](https://github.com/dotnet/aspnetcore) source code build is failing on macOS ARM64 because the gRPC package doesn't support it.

Please provide details.

xsoheilalizadeh:
> > Hi @jtattermusch, do you have any comments on this that could help?
> > Currently, [dotnet/aspnetcore](https://github.com/dotnet/aspnetcore) source code build is failing on macOS ARM64 because the gRPC package doesn't support it.
> 
> Please provide details.

Thanks, you can see the build errors section in the following comment, gRPC.Tools is using `macosx_x64/protoc` and It returns  _Bad CPU type in executable_ as an error message.


Probably it could be possible by rosetta, but I had the native binary in mind. Is the CI dependent on https://github.com/actions/virtual-environments/issues/2187 or could we find another solution, I'm curious to know what other languages do in this case?

https://github.com/dotnet/aspnetcore/issues/40109#issuecomment-1067142076

jtattermusch:
> dotnet/aspnetcore#40109 (comment)



> > > Hi @jtattermusch, do you have any comments on this that could help?
> > > Currently, [dotnet/aspnetcore](https://github.com/dotnet/aspnetcore) source code build is failing on macOS ARM64 because the gRPC package doesn't support it.
> > 
> > 
> > Please provide details.
> 
> Thanks, you can see the build errors section in the following comment, gRPC.Tools is using `macosx_x64/protoc` and It returns _Bad CPU type in executable_ as an error message.

Can you try installing rosetta emulator? I think the x64 version of protoc should be able to run on M1 Mac (can you try that manually)?

> 
> Probably it could be possible by rosetta, but I had the native binary in mind. Is the CI dependent on [actions/virtual-environments#2187](https://github.com/actions/virtual-environments/issues/2187) or could we find another solution, I'm curious to know what other languages do in this case?

No, we are not using github actions, we have use internal CI called Kokoro (and we currently don't have infrastructure to be able to run test jobs on M1).
For other gRPC languages this is not different. We currently don't provide M1 support (precisely because we don't a have good way of testing / building artifacts).

> 
> [dotnet/aspnetcore#40109 (comment)](https://github.com/dotnet/aspnetcore/issues/40109#issuecomment-1067142076)

From that log it's not obvious where exactly you're using Grpc.Tools in dotnet/aspnetcore  and for what reason. 
- is that use really necessary or it is just a convenience (e.g to make tests simpler).
- do you have other workarounds (e.g. if this is a test, you can skip it on mac M1)? Also, Grpc.Tools simply generates .cs file in the pre-build step, so on M1 you could skip that step and use pre-generated .cs files instead.

Issue #29044, opened 2022-03-07
Fork support is only compatible with the epoll1 and poll polling strategies


jaxxstorm:
### What version of gRPC and what language are you using?

v1.44.0
Python

### What operating system (Linux, Windows,...) and version?

OS X

### What runtime / compiler are you using (e.g. python version or version of gcc)

3.9.2

### What did you do?

Since the release of `v1.44.0` we're getting errors thrown that says:

```
Fork support is only compatible with the epoll1 and poll polling strategies
```

We haven't made any other changes to our code, and this didn't happen beforehand. I looked through the changelog, and it's difficult to understand why these errors are now surfacing.

### What did you expect to see?

Not an error

### What did you see instead?

```
Fork support is only compatible with the epoll1 and poll polling strategies
```

maxfirman:
Here is a minimal reproducible example of the hello world `greeter_client.py`:

```python
from __future__ import print_function

import logging
from concurrent.futures import ProcessPoolExecutor

import grpc

import helloworld_pb2
import helloworld_pb2_grpc

NUM_REQUESTS = 10
POOL_SIZE = 2


def worker(message):
    with grpc.insecure_channel('localhost:50051') as channel:
        stub = helloworld_pb2_grpc.GreeterStub(channel)
        response = stub.SayHello(helloworld_pb2.HelloRequest(name=str(message)))
        print("Greeter client received: " + response.message)


def run():
    with grpc.insecure_channel('localhost:50051') as channel:
        with ProcessPoolExecutor(POOL_SIZE) as p:
            p.map(worker, range(NUM_REQUESTS))


if __name__ == '__main__':
    logging.basicConfig()
    run()

```

drfloob:
I can't reproduce this on my mac with the latest OS X, and grpcio v1.44.0. On Mac, gRPC defaults to use the `poll` polling engine. I'd be curious if your code worked if you ran the same process with the environment variables `GRPC_POLL_STRATEGY=poll`, which should be redundant on Mac, and `GRPC_VERBOSITY=debug`. The first few lines of output will show you which polling engine you're using.

```
(pr29044) my-machine::tmp/pr29044 » GRPC_VERBOSITY=debug python main.py
D0308 09:04:42.287589000 4498904576 ev_posix.cc:172]                   Using polling engine: poll
D0308 09:04:42.287663000 4498904576 lb_policy_registry.cc:42]          registering LB policy factory for "grpclb"
[...]

PR #28566, opened 2022-01-14
Log errors for unsupported fork scenarios in Python

drfloob:
Previously it logged info-level warnings. This also now returns early from prefork if the polling strategy is not set, or otherwise not supported.

See #28557

PR #29103, opened 2022-03-15
Fix prefork handler register's default behavior

lidizheng:
The fork handlers in Core was on by default (if compile-time flag is given), and can be explicitly controlled by `GRPC_ENABLE_FORK_SUPPORT` env var. However, in Python, the parsing of `GRPC_ENABLE_FORK_SUPPORT` is off by default.

The fork-support's use case limitation: 1. post-fork pattern; 2. Python (sync stack) or C++ (status unknown); 3. polling strategy needs to be `poll`.

Most Linux distros these days have more performant default polling strategy than `poll`.

This creates an unexpected behavior. Users who unaware of fork support are forced to register fork handlers, then the fork handlers complain about the polling strategy. This was silently broken, until we decided to bump up the error level in the prefork handler: https://github.com/grpc/grpc/pull/28566.

---

This PR changes the default from true (if fork-support compiled) to false. This should not affect Python users, because for Python's fork behavior to work, users need to explicitly specify `GRPC_ENABLE_FORK_SUPPORT` to true.

~~This might be a backward incompatible change to C++ users, if they rely on the fork-on-by-default behavior. However, I'm not sure about the adoption status of C++ forking. I can create another compile-time flag just to change the default.~~

Fixes https://github.com/grpc/grpc/issues/29044

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -391,10 +391,7 @@ if "win32" in sys.platform:
         # on msvc, but only for 32 bits
         DEFINE_MACROS += (('NTDDI_VERSION', 0x06000000),)
 else:
-    DEFINE_MACROS += (
-        ('HAVE_CONFIG_H', 1),
-        ('GRPC_ENABLE_FORK_SUPPORT', 1),
-    )
+    DEFINE_MACROS += (('HAVE_CONFIG_H', 1),)
 
 LDFLAGS = tuple(EXTRA_LINK_ARGS)
 CFLAGS = tuple(EXTRA_COMPILE_ARGS)


mwtian:
Thanks for the fix! This would allow us to upgrade to newer `grpcio` packages.
Btw, since the usage of `GRPC_ENABLE_FORK_SUPPORT` macro is removed, should the line below where the macro is populated also be removed? https://github.com/grpc/grpc/blob/6ba998854ff8aa645be8e589ebb9319590278dfc/setup.py#L396
And if we don't want to break C++ fork users (if there is any), maybe just removing the line above for Python extensions would be enough?

lidizheng:
@mwtian There are users adopted fork support. We still hope to offer post-fork support, if the env var is given. So, gRPC Python should still include fork support in its binary.

Removing the macro completely might require more evidence.

gnossen (approved):

PR #29108, opened 2022-03-15
Python: Unify API between RPCState and ServicerContext(s)

TimPansino:
@donnadionne

# Overview

New Relic is looking to provide instrumentation to customers on the Python async implementation of gRPC. The information required to report properly attributed telemetry to customers is not completely available, the following changes are designed to fix that.

* Information available on SyncServicerContext differs from ServicerContext making available information different between sync and async handlers.
  * Fixed by supplying missing attributes on the SyncServicerContext
* RPCState contains the host but is irretrievable from Python level, only available in binary extension.
  * Add method to allow extraction of host that matches the extraction for method, as well as providing this API on the ServicerContext and SyncServicerContext to make available from Python.
* Target attribute is not available on channel objects from the Python level which differs from sync gRPC's implementation.
  * Add method to extract it.

### DIFF
--- a/src/python/grpcio/grpc/_cython/_cygrpc/aio/channel.pyx.pxi
+++ b/src/python/grpcio/grpc/_cython/_cygrpc/aio/channel.pyx.pxi
@@ -66,6 +66,9 @@ cdef class AioChannel:
                 try_to_connect,
             )
 
+    def target(self):
+        return self._target
+
     async def watch_connectivity_state(self,
                                        grpc_connectivity_state last_observed_state,
                                        object deadline):
--- a/src/python/grpcio/grpc/aio/_base_server.py
+++ b/src/python/grpcio/grpc/aio/_base_server.py
@@ -371,3 +371,23 @@ class ServicerContext(Generic[RequestType, ResponseType], abc.ABC):
         Returns:
           A bool indicates if the RPC is done.
         """
+
+    def host(self):
+        """Accesses the value of the host used to connect to the RPC.
+
+        This is an EXPERIMENTAL API.
+
+        Returns:
+          The Host value for the RPC.
+        """
+        raise NotImplementedError()
+
+    def method(self):
+        """Accesses the value of the method used to connect to the RPC.
+
+        This is an EXPERIMENTAL API.
+
+        Returns:
+          The Method value for the RPC.
+        """
+        raise NotImplementedError()
--- a/src/python/grpcio_tests/tests_aio/unit/context_peer_test.py
+++ b/src/python/grpcio_tests/tests_aio/unit/context_peer_test.py
@@ -16,7 +16,7 @@
 import asyncio
 import logging
 import os
-from typing import Callable, Iterable, Sequence, Tuple
+from typing import Callable, Iterable, Sequence, Tuple, Union
 import unittest
 
 import grpc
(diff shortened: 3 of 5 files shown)


TimPansino:
I acknowledge the tiny change to channels is sort of tacked onto this PR, but seeing as the whole thing is quite small I felt 2 separate PRs would be more work for the maintainers. Let me know if you want that split out for any reason.

lidizheng:
@TimPansino Thanks for the PR. To make this change complete, we will need to:

1. Update the "interface" which also generates our document, for example https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/aio/_base_call.py;
2. Add unit tests similar to https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests_aio/unit/context_peer_test.py.

TimPansino:
I've added sync and async testing to check these attributes exist on the context within a handler. Let me know if you'd like any other changes made.

src/python/grpcio_tests/tests_aio/unit/context_peer_test.py
```diff
@@ -31,33 +31,58 @@
...
         await server.start()
 
         # Creates a channel
         async with aio.insecure_channel('localhost:%d' % port) as channel:
-            response = await channel.unary_unary(_TEST_METHOD)(_REQUEST)
-            self.assertEqual(_REQUEST, response)
+            for method in ("test_async", "test_sync"):
+                _test_method = '/%s/UnaryUnary' % method
+                response = await channel.unary_unary(_test_method)(_REQUEST)
+                self.assertEqual(_REQUEST, response)
```
lidizheng, line 85: The content of the test looks good, can you split the unit test to another file under `src/python/grpcio_tests/tests_aio/unit`? Or update the name of this test file.

You might also need to update the `tests.json` file and `BUILD.bazel` file.

src/python/grpcio/grpc/aio/_base_server.py
```diff
@@ -371,3 +371,23 @@ def done(self) -> bool:
         Returns:
           A bool indicates if the RPC is done.
         """
+
+    def host(self):
```
lidizheng, line 375: `-> string`, same for other return type annotations.

src/python/grpcio/grpc/_cython/_cygrpc/aio/channel.pyx.pxi
```diff
@@ -66,6 +66,9 @@ cdef class AioChannel:
                 try_to_connect,
             )
 
+    def target(self):
+        return self._target
```
lidizheng, line 70: This is not documented or tested.

src/python/grpcio_tests/tests_aio/unit/context_peer_test.py
```diff
@@ -31,33 +31,58 @@
...
+
             self.assertEqual(_REQUEST, request)
             # The peer address could be ipv4 or ipv6
             self.assertIn('ip', context.peer())
+
+            self.assertIn('localhost', context.host().decode())
+            self.assertEqual('/%s/UnaryUnary' % method_name, context.method().decode())
+            self.assertEqual(0, context.code())
+            self.assertSequenceEqual((), context.trailing_metadata())
+            self.assertIn(servicer_class, str(type(context)))
```
lidizheng, line 56: L56: This might be too specific, let's not test the internals.

src/python/grpcio_tests/tests_aio/unit/context_peer_test.py
```diff
@@ -31,33 +31,58 @@
...
+        server.add_generic_rpc_handlers((handler_async, handler_sync))
         port = server.add_insecure_port('[::]:0')
         await server.start()
 
         # Creates a channel
         async with aio.insecure_channel('localhost:%d' % port) as channel:
-            response = await channel.unary_unary(_TEST_METHOD)(_REQUEST)
-            self.assertEqual(_REQUEST, response)
+            for method in ("test_async", "test_sync"):
+                _test_method = '/%s/UnaryUnary' % method
```
lidizheng, line 83: Can we use f-string here?

lidizheng (commented):
Thanks for the update. Looks good in general.

We are updating the surface of gRPC Python. For any API change, there is a process of gRFC. In short, we should write a short description of what we are trying to do, and the changes. You can find several Python examples there: https://github.com/grpc/proposal. This gives the community an opportunity to chime in.

PR #29119, opened 2022-03-16
Backport "decrease verbosity of alts handshaker logs which can happen during cancellation (#29058)" to 1.45.x

apolcyn:
Backports #29058


apolcyn:
bazel macos is a tool failure

bloat diff failed, but log is not accessible (should not be related to this change)

memory diff failed, but log is not accessible (should not be related to this change)

interop tests are a known node build failure

PR #29123, opened 2022-03-16
Include darwin-stub.h for latest libuv v1.44.1 build break for iOS

dennycd:
Include darwin-stub.h for libuv build for MacOS/iOS platform.  This file is needed in order to build latest libuv release (v1.44.1) on onwards 

NOTES: 
 - not merging until PR 27531 timer impl is landed with libuv version bump 

@HannahShiSFB 

---

Upstream libuv patch for iOS at  https://github.com/libuv/libuv/pull/3563


drfloob (approved):
Thanks for the fix, Denny!

drfloob (changes requested):
The file should be added in `DARWIN_LIBUV_HEADERS`, not sources.

PR #29124, opened 2022-03-16
Softfail when receiving a X509_V_ERR_UNABLE_TO_GET_CRL error

krestofur:
Although this isn't standard OpenSSL behavior, the gRPC auth team currently has a need for soft-failing on X509_V_ERR_UNABLE_TO_GET_CRL errors.

### DIFF
--- a/test/core/tsi/crl_ssl_transport_security_test.cc
+++ b/test/core/tsi/crl_ssl_transport_security_test.cc
@@ -43,6 +43,7 @@ const int kSslTsiTestRevokedKeyCertPairsNum = 1;
 const int kSslTsiTestValidKeyCertPairsNum = 1;
 const char* kSslTsiTestCrlSupportedCredentialsDir =
     "test/core/tsi/test_creds/crl_data/";
+const char* kSslTsiTestFaultyCrlsDir = "bad_path/";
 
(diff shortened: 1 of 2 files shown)


test/core/tsi/crl_ssl_transport_security_test.cc
```diff
@@ -43,17 +43,22 @@ const int kSslTsiTestRevokedKeyCertPairsNum = 1;
...
     "test/core/tsi/test_creds/crl_data/";
+const char* kSslTsiTestFaultyCrlsDir = "bad_path/";
 
 class CrlSslTransportSecurityTest
     : public testing::TestWithParam<tsi_tls_version> {
  protected:
   // A tsi_test_fixture implementation.
   class SslTsiTestFixture {
    public:
+    // When use_faulty_crl_directory is set, the crl_directory of the
```
krestofur, line 54: Currently only adds the faulty directory to the client. I think this may be sufficient for testing the softfail but I can add another unit test for the server if need be.

ZhenLian (approved):
Thanks for making the changes Chris!

ZhenLian:
Bazel C/C++ Dbg MacOS fails because of the tool failure. I've seen it passed in the previous runs. All other tests are good.

PR #29130, opened 2022-03-17
DO NOT MERGE: remove legacy .csproj distribtest for C#

jtattermusch:


jtattermusch:
could be done as possible simplification againts the v1.46.x branch.

jtattermusch:
not worth doing now after https://github.com/grpc/grpc/pull/29225 is in.

PR #28667, opened 2022-01-24
Add EventEngine end2end test logic to the gRPC build/test system

drfloob:
All tests that are configured with `uses_polling=True` are now exercised
against every known EventEngine implementation in gRPC.

For each iomgr poller, the default EventEngine factory is used and the
tests are executed as they were before.

For every additional EventEngine implementation, all polling tests are
additionally run once with that EventEngine and one of the pollers
(arbitrarily the first in the list).

This adds ~350 polling tests to //test/core/... when accounting for two 
EventEngines. When iomgr is removed, this instead will be eliminating ~700 
polling tests.

@tamird

PR #29140, opened 2022-03-17
Remove deprecated use_cfstream arg from gRPC build system

drfloob:
It is unused. This removal helps with #28667, and requires a
cherry-pick.

CC @dennycd

### DIFF
--- a/bazel/grpc_build_system.bzl
+++ b/bazel/grpc_build_system.bzl
@@ -50,12 +50,6 @@ def if_windows(a):
         "//conditions:default": [],
     })
 
-def if_mac(a):
-    return select({
-        "//:mac_x86_64": a,
-        "//conditions:default": [],
-    })
-
 def _get_external_deps(external_deps):
     ret = []
     for dep in external_deps:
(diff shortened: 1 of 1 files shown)


nicolasnoble (approved):

jtattermusch:
Are you 100% sure it's unused? I think we have a cfstream bazel job somehere (maybe https://github.com/grpc/grpc/blob/208619cdce3b7cdc681cb50c29db619e46dd5af6/tools/internal_ci/macos/grpc_run_bazel_isolated_tests.sh#L28, https://github.com/grpc/grpc/blob/master/tools/internal_ci/macos/grpc_cfstream.cfg)

Since cfstream is mostly for mobile, @dennycd  should also review.

drfloob:
@jtattermusch this does not change the effect of the `GRPC_CFSTREAM` define, it only removes an unused argument from the bazel build system. I chatted with @dennycd about this earlier, and I confirmed the argument is not used internally. https://github.com/grpc/grpc/blob/208619cdce3b7cdc681cb50c29db619e46dd5af6/src/core/lib/iomgr/iomgr_posix_cfstream.cc#L19-L22

PR #29157, opened 2022-03-18
Don't crash in epollex if we can't find a pollset when deleting it

apolcyn:
epollex can do an [early exit](https://github.com/grpc/grpc/blob/7b4a5fcba522a8449ef332cd2f092c28c22035d1/src/core/lib/iomgr/ev_epollex_linux.cc#L1461) while adding a pollset to a pollset set, for example if it fails to create an epoll fd because we hit the process fd limit.

Since the API doesn't signal an error, the caller will later delete that pollset from the same pollset set.

We shouldn't crash when that happens, though.


apolcyn:
obsoleted by https://github.com/grpc/grpc/pull/29160

PR #29174, opened 2022-03-21
xds_cluster_resolver: fix bug that caused us to incorrectly reuse priority child names

markdroth:
We were seeing a crash due to [this assertion failure](https://github.com/grpc/grpc/blob/6b89acc7ba5a3d9cd6623841bb5f8bc1609df613/src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_impl.cc#L485) in the xds_cluster_impl policy.  It was being triggered by the following scenario:

1. We had an aggregate cluster pointing to an EDS cluster and then a LOGICAL_DNS cluster.  The EDS cluster returned two priorities, priority 0 with locality A and priority 1 with localities B and C, and all of the endpoints in both priorities were unreachable.  The LOGICAL_DNS cluster was failing to return endpoints, which we [treat as an empty priority with no localities](https://github.com/grpc/grpc/blob/6b89acc7ba5a3d9cd6623841bb5f8bc1609df613/src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc#L671).  The xds_cluster_resolver policy assigned child policy names "child0" to priority 0 from the EDS cluster, "child1" to priority 1 from the EDS cluster, and "child2" to the LOGICAL_DNS cluster.
2. We got an update for the EDS cluster with priority 0 now containing localities A, B, and C and priority 1 containing locality D.  This caused the code in xds_cluster_resolver to reuse the child policy name "child0" for priority 0, since locality A was in that child in the old update and was still present in the new update.  However, it then incorrectly reused "child2" for priority 1 from the EDS cluster (because it thought that child name was not used due to the LOGICAL_DNS result not containing any localities), which resulted in sending an update to that child that changed its cluster name, this triggering the assertion above.

This PR fixes the xds_cluster_resolver policy to assign child policy names separetely within each discovery mechanism instead of doing it once across all discovery mechanisms.  We avoid conflicts between discovery mechanisms by including the cluster name in the child policy name.  This approach also solves potential problems (not actually seen in the wild) where multiple discovery services might report the same locality name.

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
@@ -495,8 +495,8 @@ void XdsClusterResolverLb::LogicalDNSDiscoveryMechanism::ResolverResultHandler::
 
 const XdsClusterResolverLbConfig::DiscoveryMechanism&
 XdsClusterResolverLb::DiscoveryMechanismEntry::config() const {
-  return discovery_mechanism->parent()->config_->discovery_mechanisms()[
-      discovery_mechanism->index()];
+  return discovery_mechanism->parent()
+      ->config_->discovery_mechanisms()[discovery_mechanism->index()];
 }
 
 std::string XdsClusterResolverLb::DiscoveryMechanismEntry::GetChildPolicyName(
@@ -504,8 +504,7 @@ std::string XdsClusterResolverLb::DiscoveryMechanismEntry::GetChildPolicyName(
   const size_t child_number = priority_child_numbers[priority];
   const auto& discovery_config = config();
   absl::InlinedVector<absl::string_view, 9> parts = {
-    "{cluster=", discovery_config.cluster_name, ","
-  };
+      "{cluster=", discovery_config.cluster_name, ","};
   if (!discovery_config.eds_service_name.empty()) {
     parts.push_back("eds_service_name=");
     parts.push_back(discovery_config.eds_service_name);
@@ -761,9 +760,9 @@ ServerAddressList XdsClusterResolverLb::CreateChildPolicyAddressesLocked() {
           }
           addresses.emplace_back(
               endpoint
-                  .WithAttribute(kHierarchicalPathAttributeKey,
-                                 MakeHierarchicalPathAttribute(
-                                     hierarchical_path))
+                  .WithAttribute(
+                      kHierarchicalPathAttributeKey,
+                      MakeHierarchicalPathAttribute(hierarchical_path))
                   .WithAttribute(kXdsLocalityNameAttributeKey,
                                  absl::make_unique<XdsLocalityAttribute>(
                                      locality_name->Ref()))
@@ -791,7 +790,7 @@ XdsClusterResolverLb::CreateChildPolicyConfigLocked() {
           discovery_entry.latest_update->priorities[priority];
       Json child_policy;
       if (!discovery_entry.discovery_mechanism->override_child_policy()
-              .empty()) {
+               .empty()) {
         child_policy =
             discovery_entry.discovery_mechanism->override_child_policy();
       } else {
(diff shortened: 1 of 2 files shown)


src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
```diff
@@ -510,6 +489,35 @@ void XdsClusterResolverLb::LogicalDNSDiscoveryMechanism::ResolverResultHandler::
...
+XdsClusterResolverLb::DiscoveryMechanismEntry::config() const {
+  return discovery_mechanism->parent()
+      ->config_->discovery_mechanisms()[discovery_mechanism->index()];
+}
+
+std::string XdsClusterResolverLb::DiscoveryMechanismEntry::GetChildPolicyName(
+    size_t priority) const {
+  const size_t child_number = priority_child_numbers[priority];
+  const auto& discovery_config = config();
+  absl::InlinedVector<absl::string_view, 9> parts = {
```
apolcyn, line 506: should this be 10 instead of 9?

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
```diff
@@ -510,6 +489,35 @@ void XdsClusterResolverLb::LogicalDNSDiscoveryMechanism::ResolverResultHandler::
...
+      "{cluster=", discovery_config.cluster_name, ","};
```
apolcyn, line 507: shouldn't the cluster name by itself be enough to differentiate amongst these priorities?

markdroth: Good point.  Done.

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
markdroth: Actually, it should have been 7.  But this has gone away based on the change I made in response to your other comment.

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
apolcyn: Followup question, what happens if we get an update for an EDS cluster which changes the EDS service name or the load reporting service name?

Not sure if that's possible in practice. But if it was, could that re-use a priority and so crash in the `XdsClusterImpl::Update`?

markdroth: The EDS service name is set in the CDS resource, not the EDS resource -- it's basically the way that the CDS resource tells the client which EDS resource to use.  And it is certainly possible to get a CDS update that changes the EDS service name.

The config of the xds_cluster_resolver policy includes a list of discovery mechanisms, and each discovery mechanism includes the [EDS service name](https://github.com/grpc/grpc-proto/blob/ab96cf12ec7ce135e03d6ea91d96213fa4cb02af/grpc/service_config/service_config.proto#L399) and [LRS server name](https://github.com/grpc/grpc-proto/blob/ab96cf12ec7ce135e03d6ea91d96213fa4cb02af/grpc/service_config/service_config.proto#L383).  When the xds_cluster_resolver policy gets a config update that changes the list of discovery mechanisms, it will [gracefully swap itself out](https://github.com/grpc/grpc/blob/8f3cd544cdd95092d514ae30bb67a037132d2e35/src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc#L1240) -- i.e., instead of updating the existing instance, it will create a whole new instance of the policy, and it will do a graceful swap over to that new instance.

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
```diff
@@ -790,139 +768,117 @@ RefCountedPtr<LoadBalancingPolicy::Config>
...
+          };
+          Json::Object& config =
+              *(*child_policy.mutable_array())[0].mutable_object();
+          auto it = config.begin();
+          GPR_ASSERT(it != config.end());
+          (*it->second.mutable_object())["targets"] =
+              std::move(weighted_targets);
+        } else {
+          auto it = xds_lb_policy.find("RING_HASH");
+          GPR_ASSERT(it != xds_lb_policy.end());
```
apolcyn, line 832: I know it's pre-existing, but does this mean we'll crash if we get a cluster with a new LB policy configured? Or is the client doing validation before we reach here?

apolcyn (approved):

PR #29175, opened 2022-03-21
Add bazel cpp distribtest for grpc_cc_library

gnossen:


.bazelignore
```diff
@@ -21,6 +21,7 @@ third_party/upb
 third_party/xds
 
 bazel/test/python_test_repo
```
jtattermusch, line 23: the python_test_repo here is basically a bazel distribtest for grpc python, right?
We should make the path to python and c++ bazel distribtests consistent. `bazel/test/python_test_repo` and `test/distrib/bazel/cpp` seems quite random.
Can you at least add a TODO to make this consistent?

test/distrib/bazel/cpp/greeter_client.cc
```diff
@@ -0,0 +1,104 @@
+/*
```
jtattermusch, line 1: is this a verbatim copy of greeter_client.cc from examples? if so, we should at least add a TODO to deduplicate.
I don't think we want to have multiple copies of greeter_client.cc in the repo.

Same for greeter_server.cc and the .proto file.

test/distrib/bazel/cpp/tools/bazel
```diff
@@ -0,0 +1 @@
+../../../../../tools/bazel
```
jtattermusch, line 1: add a comment that explains the idea behind this redirection (I understand, but not everyone will).

test/distrib/bazel/test_single_bazel_version.sh
```diff
@@ -45,3 +45,7 @@ export OVERRIDE_BAZEL_VERSION="$VERSION"
 # when running under bazel docker image, the workspace is read only.
 export OVERRIDE_BAZEL_WRAPPER_DOWNLOAD_DIR=/tmp
 bazel build -- //... "${EXCLUDED_TARGETS[@]}"
+
+cd test/distrib/bazel/cpp/
+
+bazel test //:all
```
jtattermusch, line 51: nit: for a test that has 2 actions (`bazel build -- //... "${EXCLUDED_TARGETS[@]}"` and `bazel test //:all`, running the in a sequence is generally speaking an antipattern  (if the first action fails, the other one will never run).
but in this case, since the first step tries to build grpc, it is perhaps reasonable to assume that if grpc build fails, there is no point of trying to build greeter on top of grpc.

test/distrib/bazel/cpp/greeter_client.cc
gnossen: greeter_client.cc is the same, but greeter_server.cc had to be modified.

> I don't think we want to have multiple copies of greeter_client.cc in the repo.

I don't know if I agree with this. greeter_client and greeter_server are the two simplest usages of gRPC possible, by definition. I could turn greeter_client.cc into a symlink (I think) but I'm not sure how much value we would derive from that.

test/distrib/bazel/test_single_bazel_version.sh
gnossen: That's a good point. I can see there being cases where some (say) Android target fails to build but the distribtest would not itself fail. I'll run them all in sequence regardless of the results and report errors at the end.

test/distrib/bazel/cpp/tools/bazel
jtattermusch: Ah ok, thats fair then.

test/distrib/bazel/cpp/greeter_client.cc
jtattermusch: I still think there's value in de-duplicating copies of greeter_*.cc, even though this is the "simplest possible" usage, it's still 100 LOC of not entirely trivial code.
But we can leave as is for now (I'd appreciate ideas for how to deduplicate in the future though, having a symlink doesn't seem that bad actually, but it would have to be for all files - client, server and the .proto file)

jtattermusch (approved):
LGTM once the tests are green.

test/distrib/bazel/cpp/greeter_client.cc
gnossen: Ah, I have to correct myself. All three of these files actually had to be modified. The proto lives at a different path from in the parent repo, so the `#includes` and proto package differ from what exists in the examples directory (as they should, since the package should mirror the directory structure).

The server also had to be changed to not use a fixed port. This would lead to port clashes and test failures when running concurrently. But you don't want to make the quality of the example code _worse_ by making changes for test considerations.

Issue #17340, opened 2018-11-29
SSL_ERROR_WANT_WRITE error when using large certificate


jeffdunn:
### What version of gRPC and what language are you using?
1.14.1, C++
 
 
### What operating system (Linux, Windows, …) and version?
CentOS 7
 

### What runtime / compiler are you using (e.g. python version or version of gcc)
gcc 5.5.0
 
### What did you do?
I created a sample client/server which uses secure channels. I attempted to use existing certificates from our production stack, but found that depending on the size of the certificate, the server may fail with a protocol error.

This can be reproduced with a simple server and a generated certificate with many Subject Alternative Name entries.

**Repro steps:**
- Generate large certificate using [certstrap](https://github.com/square/certstrap)
    - Create [generate_domains.py](https://github.com/grpc/grpc/files/2630392/generate_domains.py.txt)
    - `./bin/certstrap init --common-name GrpcTestingCA`
    - `./bin/certstrap request-cert --domain full_hostname_goes_here.com,(python ./generate_domains.py 128) --uri (python ./generate_domains.py 58) --common-name GrpcTesting`
    - `./bin/certstrap sign GrpcTesting --CA GrpcTestingCA`
    - See [certs.tar.gz](https://github.com/grpc/grpc/files/2630516/certs.tar.gz) for sample certs from above steps 
- Run the following:
    - [GrpcClientMain.cpp](https://github.com/grpc/grpc/files/2630426/GrpcClientMain.cpp.txt)
    - [GrpcServerMain.cpp](https://github.com/grpc/grpc/files/2630431/GrpcServerMain.cpp.txt)
    - [HelloGrpcService.proto](https://github.com/grpc/grpc/files/2630434/HelloGrpcService.proto.txt)

### What did you expect to see?
A successful connection established with a request/response.
 
### What did you see instead?

When `GRPC_TRACE=all` is set:
```
GrpcServer grpc/protobuf running on port: 20133
E1129 10:43:02.077461452 2634068 lockfree_event.cc:94]       LockfreeEvent::NotifyOn: 0x610000004eb0 curr=(nil) closure=0x612000002de8
E1129 10:43:05.549992245 2634071 lockfree_event.cc:209]      LockfreeEvent::SetReady: 0x610000004eb0 curr=0x612000002de8
E1129 10:43:05.550497959 2634071 lockfree_event.cc:94]       LockfreeEvent::NotifyOn: 0x6100000080b0 curr=(nil) closure=0x616000120220
[...]

PR #29176, opened 2022-03-21
Handle SSL_ERROR_WANT_WRITE error

yihuazhang:
The existing SSL implementation will fail a handshake if SSL returns SSL_ERROR_WANT_WRITE error. It is not a correct behavior because SSL_ERROR_WANT_WRITE is a transient error which just tells us to drain the buffer holding handshake data received from SSL. So all we need is to repeatedly drain the buffer until all handshake data is received from SSL. This PR addresses the issue.

Another concern is that a peer may receive handshake data from a peer that is larger than its bio buffer size. However in gRPC, the caller will retrieve [8KB](https://github.com/grpc/grpc/blob/master/include/grpc/impl/codegen/grpc_types.h#L333) data from the wire each time, and deliver it to SSL TSI to consume. Since the default size of the bio pair buffer is 17KB, we will not have the problem that the data received from a peer is larger than bio buffer size.

@markdroth


yihuazhang:
The behavior leading to SSL_ERROR_WANT_WRITE  can be triggered in multiple ways. For example, 1) use a large CA trust bundle at the server side or 2) configure a small SSL bio buffer. In the test, I used 2) because it is easier to configure the BIO buffer than creating a large trust bundle and shipping it with the gRPC package. 

Also notice that, in the unit test we selected the minimum size for network BIO buffer as 4KB and SSL bio buffer size as 256 bytes. It is because for network BIO buffer, the default handshake buffer size used in the `ssl_transport_security_test` is 4KB, and we need network BIO buffer to be at least as large as the handshake buffer since in our code we do not handle the situation when the bytes received from peer is larger than the size of local buffer. Regarding the SSL bio buffer size, we cannot go lower than 256 bytes because otherwise, it fails to hold ClientHello message at the client side.

yihuazhang:
Friendly ping :)

src/core/tsi/ssl_transport_security.h
```diff
@@ -217,6 +217,15 @@ tsi_result tsi_ssl_client_handshaker_factory_create_handshaker(
     tsi_ssl_client_handshaker_factory* factory,
     const char* server_name_indication, tsi_handshaker** handshaker);
 
+/* Creates a client handshaker with a custom BIO pair.
+ * The default size of each BIO buffer is 0, which translates to 17KB in
+ * boringSSL. */
+tsi_result
+tsi_ssl_client_handshaker_factory_create_handshaker_with_custom_bio_pair(
```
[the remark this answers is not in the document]
yihuazhang: Done.

src/core/tsi/ssl_transport_security.h
```diff
@@ -354,6 +363,14 @@ tsi_result tsi_create_ssl_server_handshaker_factory_with_options(
 tsi_result tsi_ssl_server_handshaker_factory_create_handshaker(
     tsi_ssl_server_handshaker_factory* factory, tsi_handshaker** handshaker);
 
+/* Creates a server handshaker with a custom BIO pair.
+ * The default size of each BIO buffer is 0, which translates to 17KB in
+ * boringSSL. */
+tsi_result
+tsi_ssl_server_handshaker_factory_create_handshaker_with_custom_bio_pair(
```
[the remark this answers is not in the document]
yihuazhang: Done.

src/core/tsi/ssl_transport_security.h
```diff
@@ -209,19 +209,13 @@ tsi_result tsi_create_ssl_client_handshaker_factory_with_options(
   - server_name_indication indicates the name of the server the client is
     trying to connect to which will be relayed to the server using the SNI
     extension.
+  - network_bio_buf_size and ssl_bio_buf_size represent BIO pair buffers used in
```
markdroth, line 212: Wait... are we not actually passing in a non-zero value here from any of the callers?  If not, why do we need to add these parameters in the first place?

yihuazhang: We passed non-zero values to the callers in `ssl_transport_security_test.cc`. For non-test callers, the passed in values are always zero.

markdroth: Ah, okay.

In the future, when we convert this code to a C++ API, we should consider making an args struct, so that callers need to specify only the parameters that are getting non-default values.  But this is fine for now.

markdroth (approved):

yihuazhang:
Thanks a lot for the review, Mark. Will merge it once confirming it does not break internal targets.

sfc-gh-yderevenets:
I would expect that this PR fixes https://github.com/grpc/grpc/issues/17340.
Probably the issue can be closed then.

Issue #29151, opened 2022-03-18
Installation of python in pre-build should install latest patch for each Python version (not 3.x.0)


jtattermusch:
In https://github.com/grpc/grpc/pull/28963/commits/35efba19fb271a6c4e500961d3f1d1518a3c3ab9 updated the patches for pythons installed in prepare_build_macos_rc to fix https://github.com/grpc/grpc/pull/28963 (see discussion in thread).

I discovered that there are more place in which we should update the pre-install scripts to install a newer patch of python:
- https://github.com/grpc/grpc/blob/83ccb0deed429d77c0a3703784938c45dc989b74/tools/internal_ci/helper_scripts/install_python_interpreters.ps1#L64-L113 (currently 3.8.0 and 3.9.0 etc are being installed).
- https://github.com/grpc/grpc/blob/master/templates/tools/dockerfile/compile_python_38.include, 

(and possibly other places).

Currently I'm not aware of this breaking anything, but we should fix at some point.

lidizheng:
In the Windows Python installation, we used officially provided installer. However, not all releases provide that (e.g., [3.8.13](https://www.python.org/downloads/release/python-3813/)). That puts this effort in a strange place, where we upgraded, but not with all CVE fixes.

Filed https://github.com/grpc/grpc/pull/29184 and https://github.com/grpc/grpc/pull/29186.

lidizheng:
Both PRs are merged now, in future, we could have an automated script for this task, like https://github.com/grpc/grpc/blob/master/tools/release/update_supported_bazel_versions.sh. Closing this issue for now.

jtattermusch:
> Both PRs are merged now, in future, we could have an automated script for this task, like https://github.com/grpc/grpc/blob/master/tools/release/update_supported_bazel_versions.sh. Closing this issue for now.

Thanks for the fix!

PR #29186, opened 2022-03-22
Upgrade Python versions for CI Dockerfiles

lidizheng:
As title.

Related https://github.com/grpc/grpc/issues/29151


lidizheng:
@gnossen @jtattermusch PTAL.

gnossen:
Definitely not a blocker for this PR, but it might be nice to have a script to automatically do this update for you. [Something recent and similar](https://github.com/grpc/grpc/blob/master/tools/release/update_supported_bazel_versions.sh).

gnossen (approved):

lidizheng:
TIL this script exists, we need to rebuild Docker images for Python, otherwise it should be similar to that script.

PR #29188, opened 2022-03-22
priority LB: avoid possibility of rescheduling a timer before it fires

markdroth:


src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -628,122 +765,25 @@ void PriorityLb::ChildPriority::OnConnectivityStateUpdateLocked(
...
-              "[priority_lb %p] child %s (%p): deactivation timer fired, "
-              "deleting child",
-              priority_policy_.get(), name_.c_str(), this);
-    }
-    deactivation_timer_callback_pending_ = false;
-    priority_policy_->DeleteChild(this);
-  }
-  Unref(DEBUG_LOCATION, "ChildPriority+timer");
-  GRPC_ERROR_UNREF(error);
+  deactivation_timer_.reset();
```
apolcyn, line 786: nit: also inline this method, for readability?

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -170,6 +167,38 @@ class PriorityLb : public LoadBalancingPolicy {
...
+      void Orphan() override;
+
+     private:
+      static void OnTimer(void* arg, grpc_error_handle error);
+      void OnTimerLocked(grpc_error_handle);
+
+      RefCountedPtr<ChildPriority> child_priority_;
+      grpc_timer timer_;
+      grpc_closure on_timer_;
+      bool timer_pending_ = false;
```
apolcyn, line 183: nit: can we just set `timer_pending_ = true` here, and below in the failover timer, to save a line in the ctor bodies of each? (I saw we're doing this in the weighted targets fix - https://github.com/grpc/grpc/pull/29203/files)

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -628,122 +765,25 @@ void PriorityLb::ChildPriority::OnConnectivityStateUpdateLocked(
...
-    }
-    failover_timer_callback_pending_ = false;
-    OnConnectivityStateUpdateLocked(
-        GRPC_CHANNEL_TRANSIENT_FAILURE,
-        absl::Status(absl::StatusCode::kUnavailable, "failover timer fired"),
-        nullptr);
-  }
-  Unref(DEBUG_LOCATION, "ChildPriority+OnFailoverTimerLocked");
-  GRPC_ERROR_UNREF(error);
+  failover_timer_ = MakeOrphanable<FailoverTimer>(Ref());
```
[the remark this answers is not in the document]
markdroth: Done.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
markdroth: Can't do that with this one, because it gets called from outside of this object, so the caller doesn't have direct access to the `deactivation_timer_` data member.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
markdroth: Good catch; I meant to come back and make the same change here but forgot.  Done.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -501,6 +518,136 @@ void PriorityLb::SelectPriorityLocked(uint32_t priority) {
...
+  if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_priority_trace)) {
+    gpr_log(GPR_INFO,
+            "[priority_lb %p] child %s (%p): deactivating -- will remove in "
+            "%" PRId64 "ms",
+            child_priority_->priority_policy_.get(),
+            child_priority_->name_.c_str(), child_priority_.get(),
+            kChildRetentionInterval.millis());
+  }
+  GRPC_CLOSURE_INIT(&on_timer_, OnTimer, this, nullptr);
+  Ref(DEBUG_LOCATION, "Timer").release();
```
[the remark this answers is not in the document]
apolcyn: Seeing that the ExecCtx::Now API has the time cache optimization on by default, I'm ok with this reasoning.

That said, the time cache has turned out to cause subtle and hard-to-debug issues more than once before. So IMO our policy should be to invalidate the time before every timer, and not invalidate it only if it's determined to be needed for performance (until that PR is merged).

But, granted there's no indication that this particular usage would cause issues.

apolcyn (approved):
LGTM. Please see my comment on weighted target PR though: https://github.com/grpc/grpc/pull/29203#pullrequestreview-920806988

I think that minor change also can apply here

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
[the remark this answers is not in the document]
markdroth: I think we have consensus that we want to get to a future where we do time caching on the data plane but not anywhere on the control plane.  However, we need to find a reasonable API for doing that; it is not sustainable to expect all timer callers to explicitly call `InvalidateNow()`, and I would be opposed to that as a general approach to addressing this problem.  I think we are likely to find a better approach as part of the promise conversion work that Craig is doing.

In any case, this discussion is far outside of the scope of this PR.

markdroth (commented):
> LGTM. Please see my comment on weighted target PR though: [#29203 (review)](https://github.com/grpc/grpc/pull/29203#pullrequestreview-920806988)
> 
> I think that minor change also can apply here

As per my reply there, it looks like there was actually no need for the timer callback to check `shutting_down_` in the first place, because whenever that's set, `timer_pending_` will already be false, so I've removed the check on `shutting_down_`.

PR #29194, opened 2022-03-23
client channel: improve trace logging

markdroth:
Improve client channel trace logging:
- Move all trace messages in `ClientChannel::CallData` to the `client_channel_call` tracer.
- Move all trace messages in `ClientChannel::LoadBalancedCall` to a new `client_channel_lb_call` tracer.
- Rename `client_channel_routing` tracer to just `client_channel`, since it is now only the control-plane piece of routing.
- Add some additional trace logging messages to show how ops flow through name resolution and load balancing.

### DIFF
--- a/src/core/ext/filters/client_channel/client_channel.cc
+++ b/src/core/ext/filters/client_channel/client_channel.cc
@@ -2591,7 +2591,11 @@ ClientChannel::LoadBalancedCall::LoadBalancedCall(
       on_call_destruction_complete_(on_call_destruction_complete),
       call_dispatch_controller_(call_dispatch_controller),
       call_attempt_tracer_(
-          GetCallAttemptTracer(args.context, is_transparent_retry)) {}
+          GetCallAttemptTracer(args.context, is_transparent_retry)) {
+  if (GRPC_TRACE_FLAG_ENABLED(grpc_client_channel_lb_call_trace)) {
+    gpr_log(GPR_INFO, "chand=%p lb_call=%p: created", chand_, this);
+  }
+}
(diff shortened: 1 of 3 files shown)


src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -2261,30 +2270,27 @@ grpc_error_handle ClientChannel::CallData::ApplyServiceConfigToCallLocked(
...
   if (service_config_call_data != nullptr) {
     service_config_call_data->call_dispatch_controller()->Commit();
   }
   // Chain to original callback.
-  Closure::Run(DEBUG_LOCATION, self->original_recv_trailing_metadata_ready_,
+  Closure::Run(DEBUG_LOCATION, calld->original_recv_trailing_metadata_ready_,
                GRPC_ERROR_REF(error));
 }
 
-void ClientChannel::CallData::
```
apolcyn, line 2276: nit: does inlining this method need to be done as a part of this change?

Can we undo if if not?

apolcyn (approved):

src/core/ext/filters/client_channel/client_channel.cc
markdroth: It doesn't strictly need to be here, but it makes sense to be.  If I'd left this method in place, I would have had to add the `grpc_call_element*` parameter to it, since I needed that to be the parameter for the intercepted callback, so that the callback can log `chand`.  Since I already had all of the state in the caller, I just moved the code there to avoid some boilerplate.

PR #29198, opened 2022-03-23
client channel: record call completion even if recv_trailing_metadata was not started

markdroth:
We are definitely seeing cases in the wild where a call is cancelled and unreffed without starting the recv_trailing_metadata op.  See internal b/226204301 for details.

Note that the place where we actually detected this was an assertion in the LB policy `SubchannelCallTracker` API, not the `CallTracer` API, but it's the same symptom in both cases.

Can you please add a test to test/cpp/ext/filters/census/stats_plugin_end2end_test.cc to make sure the OpenCensus code is correctly handling this case?  It looks like it should work, but let's be sure.

I am going to work on adding a test for the LB policy API.

### DIFF
--- a/src/core/ext/filters/client_channel/client_channel.cc
+++ b/src/core/ext/filters/client_channel/client_channel.cc
@@ -2527,7 +2527,8 @@ class ClientChannel::LoadBalancedCall::BackendMetricAccessor
       : lb_call_(lb_call) {}
 
   const BackendMetricData* GetBackendMetricData() override {
-    if (lb_call_->backend_metric_data_ == nullptr) {
+    if (lb_call_->backend_metric_data_ == nullptr &&
+        lb_call_->recv_trailing_metadata_ != nullptr) {
       if (const auto* md = lb_call_->recv_trailing_metadata_->get_pointer(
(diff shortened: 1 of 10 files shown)


yashykt (approved):
Can you create an issue to track adding such a test so I don't forget? Also, how would one test this? Didn't we change our C++ API to always start the recv trailing metadata op?

markdroth:
See the test I just added for the LB policy API.  I think it's pretty trivial to trigger this problem using the bidi sync API.

Let me see if I can add the one for opencensus myself.  I'd prefer to get that test in with this PR if possible, so that we know we're not introducing additional bugs here.

markdroth:
I've added the opencensus test.  I'm glad that I wrote it, beacuse it did find a bug in my code, but the test itself isn't quite as valuable as I'd like, since it doesn't actually fail without the code change in this PR.  The problem is that the `RecordReceivedTrailingMetadata()` method, which is what we're calling here, actually can't record anything in this case, so all it does is set `status_code_`.  However, the only case where we can trigger this code-path is by cancelling the call, and in that case `status_code_` was already set via `RecordCancel()`.

Anyway, PTAL.

yashykt (approved):

PR #29202, opened 2022-03-23
Refactor end2end tests to exercise each EventEngine

drfloob:
This third version has a few differences:
 * maintains the status quo using `sh_test` on a single binary for each `<end2end test> x <fixture>`, but now using different `--poller` and `--engine` args
 * works out of the box on Windows (avoids more days debugging complex linker problems)
 * avoids compile/link time growth of previous attempts
 * centralizes the  `<poller> x <engine>` cartesian product logic to allow reuse in both cc_test and sh_test

There is broader goal to rewrite end2end tests using googletest eventually, so this work may be temporary. This solution enables us to test different EventEngine implementations today.

@markdroth


drfloob:
@jtattermusch that's great to hear, that work should be restarted soon, enough people are interested. Also, if my PR here is ready to go, which I think it is, we should land it so we can exercise some EventEngines and work on the gtestification separately.

CMakeLists.txt
```diff
@@ -12200,44 +12199,6 @@ target_link_libraries(interop_server
...
-if(gRPC_BUILD_TESTS)
-if(_gRPC_PLATFORM_LINUX OR _gRPC_PLATFORM_MAC OR _gRPC_PLATFORM_POSIX)
-
-  add_executable(interop_test
-    test/cpp/interop/interop_test.cc
-    third_party/googletest/googletest/src/gtest-all.cc
-    third_party/googletest/googlemock/src/gmock-all.cc
-  )
-
-  target_include_directories(interop_test
```
jtattermusch, line 12213: any reason why interop_test be no longer buildable with cmake? Or is that an oversight?

test/core/address_utils/parse_address_test.cc
```diff
@@ -129,7 +129,7 @@ static void test_grpc_parse_ipv6_invalid(const char* uri_text) {
 }
 
 int main(int argc, char** argv) {
-  grpc::testing::TestEnvironment env(argc, argv);
+  grpc::testing::TestEnvironment env(&argc, argv);
```
jtattermusch, line 132: why is changing the signature of TestEnvironment important?  Vast majority of the changes in this PR is the change of this line, and yet this mechanic changes isn't contained in a separate commit, which makes it hard to review the rest of the change (needle in haystack).
Can you include all the refactorings in a separate commit (separate from the actual logic of the change)?

test/core/end2end/run.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
```
jtattermusch, line 1: the name `run.sh` doesn't seem very descriptive, `run_configured_test.sh`?

test/core/end2end/run.sh
```diff
@@ -12,8 +13,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 set -ex
-export GRPC_POLL_STRATEGY=$1
```
jtattermusch, line 17: It seems that if you introduced an env variable (e.g. `GRPC_TEST_EVENT_ENGINE` since it seems to be only used in tests) that will be used by TestEnvironment, you could have saved yourself (and the reviewer) 300 occurrences of `grpc::testing::TestEnvironment env(&argc, argv);` (and other related edits).

bazel/grpc_build_system.bzl
```diff
@@ -389,14 +450,55 @@ def grpc_generate_one_off_targets():
...
 
-def grpc_sh_test(name, srcs, args = [], data = [], tags = []):
-    native.sh_test(
-        name = name,
-        srcs = srcs,
-        args = args,
-        data = data,
-        tags = tags,
-    )
+def grpc_sh_test(name, srcs = [], args = [], data = [], uses_polling = True, size = "medium", timeout = None, tags = [], exec_compatible_with = [], exec_properties = {}, shard_count = None, flaky = None, exclude_pollers = [], test_each_engine = True):
```
jtattermusch, line 453: why do you need to change grpc_sh_test definition? It seems that all the test that use pollers/event engine are grpc_cc_tests?

test/core/end2end/run.sh
drfloob: I used that approach in an earlier version. Because we support bazel 3.x, we need to use something like a generic `run.sh` (or a genrule, if the script itself were offensive to some sensibilities), so packaging the test execution as a `"<command> [args]"`string  made the most sense. Also, I didn't think ~200 copies of a one-character change would be very onerous to review, apologies if that was annoying to you. Aside from the burden of changing a function signature (which is already done), what arguments do you have that environment variables are a superior approach?

bazel/grpc_build_system.bzl
drfloob: You missed its usage in `test/core/end2end/generate_tests.bzl`, PTAL. The sh_tests need to be parameterized for the product of pollers and engines, same as cc_test.

test/core/address_utils/parse_address_test.cc
drfloob: > why is changing the signature of TestEnvironment important?

There are situations where command line arguments need to be consumed, not just parsed. In the microbenchmarks, for example, the benchmark framework will complain if an unknown argument is present. Making `argc` a reference argument is a fairly lightweight way to accomplish that.

> Can you include all the refactorings in a separate commit 

It would be harder to justify the refactoring outside this PR. This PR gives it a new concrete usage. I imagine your pushback would be stronger if that PR landed in your review queue as a separate change without its context. If you agree the change would be useful, I can go through the exercise of extracting it to make its review easier.

test/core/end2end/run.sh
drfloob: Fair point. This script could do more than execute configured tests, it's a generic `"$@"`. This general script replaces two near-identical scripts that had overly-specific names, which I think may have hindered their reuse. It could more appropriately be `run_arguments.sh` or `run_anything.sh`. I could probably replace it with a genrule that does the same, but that would be over-engineering.

For what it's worth, this is not unprecedented, see `run.sh` in xds_k8s_test_driver for example. I think its location is sufficient enough context to make the name not terribly evil.

CMakeLists.txt
drfloob: I'll add that I think this would be a special case, the only binary in a test folder that would need special handling because no exported tests depend on it. I didn't do anything to explicitly exclude it, per se.

test/core/bad_ssl/generate_tests.bzl
```diff
@@ -42,7 +42,10 @@ def grpc_bad_ssl_tests():
         grpc_cc_binary(
             name = "bad_ssl_%s_server" % t,
             srcs = ["servers/%s.cc" % t],
-            deps = [":bad_ssl_test_server"],
+            deps = [
+                ":bad_ssl_test_server",
+                "//test/core/util:grpc_test_util",
```
markdroth, line 47: Why is this dependency needed?

test/core/address_utils/parse_address_test.cc
markdroth: I tend to agree that this seems a little invasive.  Is there an easy way to work around this in the benchmark tests?  If not, this approach is workable, but it would be nice to touch fewer files here.

bazel/grpc_build_system.bzl
```diff
@@ -282,17 +352,18 @@ def grpc_cc_test(name, srcs = [], deps = [], external_deps = [], args = [], data
         flaky: Whether this test is flaky.
         copts: Add these to the compiler invocation.
         linkstatic: link the binary in static mode
+        exclude_pollers: list of poller names to exclude for this set of tests.
+        test_each_engine: set to False if the test is not sensitive to
```
markdroth, line 356: How about calling this parameter `uses_event_engine`?  That seems like a much more intuitive name from the perspective of someone reading a test definition.

Also, I suspect that there are a lot of tests that are not setting this to False that should be.  Just from a quick scan, I suspect that almost everything under test/core *aside* from the end2end tests should be doing this.

markdroth (commented):
Thanks for doing this!  The approach is fine, as per our previous discussion.  Comments are mostly minor.

test/core/address_utils/parse_address_test.cc
ctiller: Could we just divorce the TestEnvironment bits from the microbenchmarks?
A quick survey says that they're probably not overly benefiting from this abstraction.

drfloob: > Could we just divorce the TestEnvironment bits from the microbenchmarks?

I think so. However the microbenchmarks were not the only tests sensitive to unknown CLI args. It was a small number of tests, single digits I think. `test/core/fling/server.cc` is another example that would need some cleanup. This `argc` mutation approach seemed simple, I'm not sure why it's controversial. I don't buy the argument that broad changes should be inherently controversial, I see them in this codebase all the time. This one looks harmless.

If we did a broader cleanup to remove unknown argument sensitivities, we might be able to use absl::Flags instead, which would be another improvement, and another fairly trivial change to this TestEnvironment constructor definition (probably no need for argc and argv). I'd be happy to put that cleanup on a "things to do while compiling" queue. But I don't think that cleanup should block this work.

test/core/bad_ssl/generate_tests.bzl
drfloob: it's not needed, good catch. This carried over from a previous incarnation.

bazel/grpc_build_system.bzl
drfloob: Done. I used that name in a previous PR, but decided that it was misleading: most tests will certainly _use_ an EventEngine, but many of those tests may not care if they're exercised against every EventEngine implementation. It's not a big problem, though.

> Also, I suspect that there are a lot of tests that are not setting this to False that should be.

Agreed. I caught a handful already in the promise and avl code. We can catch more as we go along. It might make sense to have this set to False by default, but that would conflict with how the poller expansion works today.

drfloob: It's probably a sufficient heuristic at the moment to add `uses_event_engine=False` wherever `uses_polling=False` is set. There will be some exceptions to that, such as the EventEngine conformance test suite, but that may be a good first pass. WDYT?

drfloob: I did a first pass on the ~150 tests where `uses_polling=False`. Most are isolated tests that also do not need to exercise all EventEngines, with only a few exceptions. I also ensured that the `message_compress` tests are not parameterized on `poller x engines`, I don't think they need to be.

CMakeLists.txt
jtattermusch: well something in this PR clearly excludes this target from the cmake build and I think it's important to understand what caused it. We can't just make changes that have side effects without understanding what is causing those side effects and having a discussion whether these side effects make sense or not.
I pointed out this one since it seemed that removing interop_test target from cmake has nothing to do with the purpose of this change, so to me it's a red flag that the PR does something we don't quite understand.

So I don't feel strongly about keeping interop_test as part of cmake build (probably not a disaster if it is removed), but I'd like to see that effort has been put in keeping it there and that we have a good understanding why it disappeared.

bazel/grpc_build_system.bzl
```diff
@@ -259,7 +262,74 @@ def ios_cc_test(
...
+        # handles other platforms.
+        poller_config.append({
+            "name": name,
+            "srcs": srcs,
+            "deps": deps,
+            "tags": tags + ["no_linux"],
+            "args": args,
+        })
+    else:
+        for engine_name, engine in EVENT_ENGINES.items():
```
jtattermusch, line 313: I looked at the list of bazel tests that ran on this PR and I never saw the "@engine=" in the name of a test, which seems suspicious?

bazel/grpc_build_system.bzl
```diff
@@ -259,7 +262,74 @@ def ios_cc_test(
             deps = ios_test_deps,
         )
 
-def grpc_cc_test(name, srcs = [], deps = [], external_deps = [], args = [], data = [], uses_polling = True, language = "C++", size = "medium", timeout = None, tags = [], exec_compatible_with = [], exec_properties = {}, shard_count = None, flaky = None, copts = [], linkstatic = None):
+def expand_tests_for_each_poller_and_engine(name, srcs, deps, tags, args, exclude_pollers, uses_event_engine):
```
jtattermusch, line 265: suggestion: to make sure this logic here is correct and give the expected behavior, it might be useful to use `bazel query` to generate a list of tests we'd run on linux and  on non-linux  (use the tags to filter them) and then post the diff before and after.
That would make it clear which tests were added (if any) and which tests were removed (which could be intentional or by mistake). This would also give us a good idea whats the diff on number of tests targets.

jtattermusch (commented):
I reviewed the changes from the build and infrastructure perspective (mostly  bazel/grpc_build_system.bzl changes) and left some comments.

After the comments are addressed, I'm happy with the PR. I'll leave the C/C++ review to @markdroth  and @ctiller (I'll defer the decision on `grpc::testing::TestEnvironment env(&argc, argv);` changes to them as well).

bazel/grpc_build_system.bzl
```diff
@@ -36,6 +36,11 @@ load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl
 # The set of pollers to test against if a test exercises polling
 POLLERS = ["epoll1", "poll"]
 
+# The set of known EventEngines to test
+EVENT_ENGINES = {
+    "default": {"tags": []},  # must not include tag "no_linux"
```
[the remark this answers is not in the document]
drfloob: Removed, it's a misleading comment, thanks for pointing it out. Similar to iomgr, I've made the assumption that the default engine is usable on all platforms (whatever the default happens to be on that platform).

test/core/address_utils/BUILD
```diff
@@ -51,6 +51,7 @@ grpc_cc_test(
     srcs = ["parse_address_with_named_scope_id_test.cc"],
     language = "C++",
     tags = ["no_windows"],
+    uses_event_engine = True,  # uses getaddrinfo
```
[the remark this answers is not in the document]
drfloob: Fixed.

bazel/grpc_build_system.bzl
jtattermusch: sg, thanks for sharing (and +1 for keeping the set of tests identical, which makes checking correctness much easier).

I'd recommend to make the diff also based on platform tags  (no_linux, no_mac, no_windows), since having these tags assigned correctly is pretty important for the right tests to be run on all platforms.

CMakeLists.txt
drfloob: > We can't just make changes that have side effects without understanding what is causing those

You're right to point it out, absolutely, and having a fruitful discussion is important. 

> to me it's a red flag that the PR does something we don't quite understand.

Not the case, the mechanics are clear. This target was a unique case beforehand, that's not new. It's plausible that this target was added into cmake accidentally in the first place, though I'm not sure our CI would be sensitive to that (our master builds would, at least). `interop_client` and `interop_server` still exist in the cmake build, which is important because I _did_ find uses of them via cmake in our master builds. If this target is relied upon via cmake, it's not obvious where or how. Putting in extra effort to keep what _may be_ a mistake is not the right move in my opinion. If some tests break due to non-obvious usages, that's a problem that this PR would help identify. 

All in all, a clear improvement is presented here, a best effort was made, our builds and various test environments are complex, and in any case rollbacks happen frequently on this repo in the normal course of operation due to issues that can be difficult to predict. I'd be happy to be shown it's needed, and put in the work required to document that somewhere obvious, and handle this edge case in the cmake export.

test/core/address_utils/parse_address_test.cc
drfloob: > I think it would be cleaner to fix those tests instead of making this broader change.

I wouldn't label them as things that need fixing. We could definitely change their approaches to arg parsing, but I think that can be done in the other cleanup as you suggested. 

> the change seems to make the API a bit less intuitive

+1 to Craig's comment below, this is a standard approach we can find in our own codebase and dependencies, I did not invent it. I disagree that it's less intuitive.

bazel/grpc_build_system.bzl
drfloob: Done.

bazel/grpc_build_system.bzl
drfloob: Good idea, and there are no differences.

On both branches:

```
bazel query 'attr(tags, "no_linux", test/core/end2end:all)' | wc -l
1907
bazel query 'attr(tags, "no_windows", test/core/end2end:all)' | wc -l
4489
bazel query 'attr(tags, "no_mac", test/core/end2end:all)' | wc -l
3978
```

Similarly, I ran `bazel query 'attr(tags, "no_mac", test/core/end2end:all)' > /tmp/<branch>.no_mac.diff` for all three tags, diffed the results, and they are all empty.

markdroth (approved):
I'm still not crazy about the `TestEnvironment()` API change, but I'm not going to hold up this PR over it.

Thanks for doing this.

PR #29203, opened 2022-03-23
weighted_target_lb: avoid possibility of rescheduling a timer before it fires

markdroth:


src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
```diff
@@ -401,6 +413,53 @@ void WeightedTargetLb::UpdateStateLocked() {
...
+  (void)GRPC_ERROR_REF(error);  // ref owned by lambda
+  self->weighted_child_->weighted_target_policy_->work_serializer()->Run(
+      [self, error]() { self->OnTimerLocked(error); }, DEBUG_LOCATION);
+}
+
+void WeightedTargetLb::WeightedChild::DelayedRemovalTimer::OnTimerLocked(
+    grpc_error_handle error) {
+  if (error == GRPC_ERROR_NONE && timer_pending_ &&
+      !weighted_child_->shutdown_ && weighted_child_->weight_ == 0) {
+    timer_pending_ = false;
```
apolcyn, line 455: nit: any reason not to set `timer_pending_ = false`, unconditionally from this method?

Otherwise, I think there would be no bug, but `Orphan` would still log `cancelling delayed removal timer`, which might be confusing.

Same suggestion for timers in priority lb PR too, actually: https://github.com/grpc/grpc/pull/29188/files

apolcyn (approved):

src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
markdroth: I don't think that case is actually possible, because in basically any of those other cases, `timer_pending_` will already be false.

I think this means that we actually don't need to check `shutdown_` or `weight_` at all, so I've changed the code accordingly.

PR #29207, opened 2022-03-24
Make scenario results to include start and end time for a test scenario.

wanlin31:
This commit update the scenario results to include the start_timestamp
and end_timestamp. These two fields are used in prometheus range queries
to indicate the period we are pulling data for.

### DIFF
--- a/test/cpp/qps/driver.cc
+++ b/test/cpp/qps/driver.cc
@@ -32,6 +32,7 @@
 #include <grpcpp/client_context.h>
 #include <grpcpp/create_channel.h>
 
+#include "google/protobuf/timestamp.pb.h"
 #include "src/core/lib/gpr/env.h"
 #include "src/core/lib/gprpp/host_port.h"
 #include "src/core/lib/profiling/timers.h"
@@ -573,6 +574,11 @@ std::unique_ptr<ScenarioResult> RunScenario(
 
   // Start a run
   gpr_log(GPR_INFO, "Starting");
+
+  google::protobuf::Timestamp start_timestamp;
+  start_timestamp.set_seconds(time(NULL));
+  start_timestamp.set_nanos(0);
+
   for (size_t i = 0; i < num_servers; i++) {
     auto server = &servers[i];
     if (!server->stream->Write(server_mark)) {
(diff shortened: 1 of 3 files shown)


test/cpp/qps/driver.cc
```diff
@@ -32,6 +32,7 @@
 #include <grpcpp/client_context.h>
 #include <grpcpp/create_channel.h>
 
+#include "google/protobuf/timestamp.pb.h"
```
paulosjca, line 35: Try using "proto.h" instead of "pb.h" (it is a smaller file).

test/cpp/qps/driver.cc
```diff
@@ -573,6 +574,11 @@ std::unique_ptr<ScenarioResult> RunScenario(
 
   // Start a run
   gpr_log(GPR_INFO, "Starting");
+
+  google::protobuf::Timestamp start_timestamp;
```
paulosjca, line 578: There is no need to create a timetamp object for start and end. Instead, you can do auto start_time = time(Null) and then update the field in the proto directly at the end.

test/cpp/qps/driver.cc
```diff
@@ -650,6 +660,11 @@ std::unique_ptr<ScenarioResult> RunScenario(
     rrc->set_status_code(it->first);
     rrc->set_count(it->second);
   }
+
+  // Fill in start and end time for the test scenario
+  result->mutable_summary()->mutable_start_time()->CopyFrom(start_timestamp);
```
paulosjca, line 665: Here, if you use start_time and end_time of type time_t, you can just call set_seconds(start_time) on the mutable_start_time() to set the time. Since we are using proto3, it is not necessary to call set_nanos(0).

paulosjca (commented):
Check the sanity checks for code formatting errors. You will need to run a script to format the C++ code.

test/cpp/qps/driver.cc
wanlin31: I have tried to compile timestamp.proto, seems like outside google it only generate the .pb.h file, could not find a .proto.h file.

paulosjca (approved):

paulosjca (approved):

gnossen (approved):

paulosjca (approved):

PR #29215, opened 2022-03-24
ORCA: implement ORCA RPC service for OOB backend metric reporting

markdroth:


markdroth:
This is once again ready for review.

yashykt:
Since this is C++ and not Core, we should be able to use proto libraries instead of upb correct? (similar to channelz, csds, health, load_reporter)

markdroth:
We could use protobuf, but there's no real benefit to doing so, especially since we're depending on the generated code from upb anyway for the client side, where we can't depend on proto.

The health check service does actually use upb on the server side as well, for the same reason.  I'm not sure why we're depending on protobuf instead of using upb for channelz, csds, and load reporter; I think those should probably all be changed to use upb instead.  The only service that really needs to actually use protobuf is the reflection service, for obvious reasons.

src/cpp/server/orca/orca_service.cc
```diff
@@ -0,0 +1,224 @@
...
+    Slice slice;
+    request_buffer->DumpToSingleSlice(&slice);
+    // Parse request proto.
+    upb::Arena arena;
+    xds_service_orca_v3_OrcaLoadReportRequest* request =
+        xds_service_orca_v3_OrcaLoadReportRequest_parse(
+            reinterpret_cast<const char*>(slice.begin()), slice.size(),
+            arena.ptr());
+    if (request == nullptr) {
+      Finish(Status(StatusCode::UNKNOWN, "could not parse request proto"));
```
yashykt, line 51: wouldn't internal be a better error code?

include/grpcpp/ext/orca_service.h
```diff
@@ -0,0 +1,82 @@
...
+// utilization metrics to clients.
+class OrcaService : public Service {
+ public:
+  struct Options {
+    // Minimum report interval.  If a client requests an interval lower
+    // than this value, this value will be used instead.
+    int min_report_duration_ms = 30 * 1000;
+
+    Options() = default;
+    Options& set_min_report_duration_ms(int ms) {
```
yashykt, line 44: Quick check on whether we can use absl::Duration

yashykt (approved):

include/grpcpp/ext/orca_service.h
markdroth: I had assumed that we didn't want to use this yet based on past discussions, but maybe this is an edge case.  I've started an internal discussion about this, and we'll see what the consensus is.

I don't think this needs to block this PR, because the API is still marked as experimental, so I can change this if we decide to use `absl::Duration`.

src/cpp/server/orca/orca_service.cc
markdroth: Good catch.  Looks like we use INTERNAL for the public C++ API, so we should do the same thing here.

Done.

include/grpcpp/ext/orca_service.h
markdroth: I've changed it to use `absl::Duration`, as per the result of the internal discussion.

yashykt (approved):

PR #29217, opened 2022-03-24
HTTP Proxy: Ignore empty entry in no_proxy list

### LINKED ISSUE #26858: `no_proxy` with an empty string causes grpcio to wildcard everything
`Version` : 1.39.0
`Language`: Python

`OS`: Linux

`Runtime`: Python 3.8.11

### What did you do?
`Steps to reproduce`:

1. We were running google cloud monitioring apis which were using `grpcio` library underneath.
2. To get past our proxy we had `https_proxy` and `no_proxy` variables set as following
```shell
export https_proxy=$proxy_server
export no_proxy=",127.0.0.1,localhost"
```
Notice the leading `,` for `no_proxy`. This was because we set a few values for `no_proxy` commonly leaving individual applications to add to that list using their own local variables which when unset resulted in the empty string at the beginning. 
3. This made `grpcio` wildcard everything and the requests were not hitting the proxy servers. 

Requests forwarded to the proxy server.

### What did you see instead?
We were seeing the following error - 
```
Unhandled exception while getting metrics, project=‘*****',
  File “/*/site-packages/google/api_core/grpc_helpers.py", line 67, in error_remapped_callable
    return callable_(*args, **kwargs)
  File “/*/site-packages/grpc/_channel.py", line 946, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File “/*/site-packages/grpc/_channel.py", line 849, in _end_unary_response_blocking
    raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.UNAVAILABLE
	details = "failed to connect to all addresses"
	debug_error_string = "{"created":"@1626360352.381792140","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3009,"referenced_errors":[{"created":"@1626360352.381790821","description":"failed to connect to all addresses","file":"src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc","file_line":398,"grpc_status":14}]}"
>
```

This happens even when there is a `trailing-comma`

yashykt:
Fix #26858


markdroth (commented):
How about instead changing this code to use `absl::StrSplit()` with `absl::SkipEmpty()`, so that it doesn't return the empty entries in the first place?

yashykt:
Done! Added some testing while I was at it.

src/core/ext/filters/client_channel/http_proxy.cc
```diff
@@ -119,111 +120,95 @@ std::string MaybeAddDefaultPort(absl::string_view target) {
...
+  if (uri->scheme() == "unix") {
+    gpr_log(GPR_INFO, "not using proxy for Unix domain socket '%s'",
+            server_uri);
+    goto no_use_proxy;
+  }
+  /* Prefer using 'no_grpc_proxy'. Fallback on 'no_proxy' if it is not set. */
+  no_proxy_str = gpr_getenv("no_grpc_proxy");
+  if (no_proxy_str == nullptr) no_proxy_str = gpr_getenv("no_proxy");
+  if (no_proxy_str != nullptr) {
+    static const char* NO_PROXY_SEPARATOR = ",";
```
markdroth, line 154: I think this can just be inlined in the call to `absl::StrSplit()`.

src/core/ext/filters/client_channel/http_proxy.cc
```diff
@@ -119,111 +120,95 @@ std::string MaybeAddDefaultPort(absl::string_view target) {
...
-            use_proxy = false;
-            break;
-          }
-        }
-        for (size_t i = 0; i < num_no_proxy_hosts; i++) {
-          gpr_free(no_proxy_hosts[i]);
+      gpr_free(no_proxy_str);
+    } else {
+      size_t uri_len = server_host.size();
+      std::vector<std::string> no_proxy_hosts =
```
markdroth, line 167: Suggest using `absl::string_view` instead of `std::string` here to avoid unnecessary copies.

src/core/ext/filters/client_channel/http_proxy.cc
```diff
@@ -119,111 +120,95 @@ std::string MaybeAddDefaultPort(absl::string_view target) {
...
+          absl::StrSplit(no_proxy_str, NO_PROXY_SEPARATOR, absl::SkipEmpty());
+      for (const auto& no_proxy_entry : no_proxy_hosts) {
+        size_t no_proxy_len = no_proxy_entry.size();
```
markdroth, line 170: This can just check `absl::EndsWithIgnoreCase(no_proxy_entry, server_host)`.

src/core/ext/filters/client_channel/http_proxy.cc
```diff
@@ -119,111 +120,95 @@ std::string MaybeAddDefaultPort(absl::string_view target) {
...
+  if (*name_to_resolve == nullptr) return false;
+  char* no_proxy_str = nullptr;
+  std::string server_target;
+  absl::StatusOr<URI> uri = URI::Parse(server_uri);
+  if (!uri.ok() || uri->path().empty()) {
+    gpr_log(GPR_ERROR,
+            "'http_proxy' environment variable set, but cannot "
+            "parse server URI '%s' -- not using proxy. Error: %s",
+            server_uri, uri.status().ToString().c_str());
+    goto no_use_proxy;
```
markdroth, line 143: Suggest using a lambda to eliminate these `goto`s:

```
auto no_use_proxy = [&]() {
  gpr_free(*name_to_resolve);
  *name_to_resolve = nullptr;
  gpr_free(user_cred);
  return false;
};
```

Then you can replace all instances of `goto no_use_proxy` with `return no_use_proxy()`.

src/core/ext/filters/client_channel/http_proxy.cc
```diff
@@ -119,111 +120,95 @@ std::string MaybeAddDefaultPort(absl::string_view target) {
...
+  }
+  if (uri->scheme() == "unix") {
+    gpr_log(GPR_INFO, "not using proxy for Unix domain socket '%s'",
+            server_uri);
+    goto no_use_proxy;
+  }
+  /* Prefer using 'no_grpc_proxy'. Fallback on 'no_proxy' if it is not set. */
+  no_proxy_str = gpr_getenv("no_grpc_proxy");
```
markdroth, line 151: Suggest using `UniquePtr<char>` here to avoid needing to manually free the string in various places later.

src/core/ext/filters/client_channel/http_proxy.cc
```diff
@@ -119,111 +120,95 @@ std::string MaybeAddDefaultPort(absl::string_view target) {
...
   }
-
-  bool MapAddress(const grpc_resolved_address& /*address*/,
-                  const grpc_channel_args* /*args*/,
-                  grpc_resolved_address** /*new_address*/,
-                  grpc_channel_args** /*new_args*/) override {
-    return false;
+  server_target =
+      MaybeAddDefaultPort(absl::StripPrefix(uri->path(), "/")).c_str();
+  grpc_arg args_to_add[2];
```
markdroth, line 186: Suggest using `absl::InlinedVector<grpc_arg, 2>` to make the logic below a little simpler.

test/core/client_channel/http_proxy_mapper_test.cc
```diff
@@ -0,0 +1,87 @@
...
+  gpr_setenv("no_proxy", "");
+  grpc_arg proxy_arg = grpc_channel_arg_string_create(
+      const_cast<char*>(GRPC_ARG_HTTP_PROXY),
+      const_cast<char*>("http://proxy.google.com"));
+  grpc_channel_args args = {1, &proxy_arg};
+  grpc_channel_args* new_args = nullptr;
+  char* name_to_resolve = nullptr;
+  EXPECT_TRUE(HttpProxyMapper().MapName("dns:///test.google.com:443", &args,
+                                        &name_to_resolve, &new_args));
+  gpr_free(name_to_resolve);
```
markdroth, line 41: Please check the value of `name_to_resolve` and `new_args` here.

test/core/client_channel/BUILD
```diff
@@ -30,6 +30,18 @@ grpc_cc_test(
     ],
 )
 
+grpc_cc_test(
+    name = "http_proxy_mapper_test",
```
markdroth, line 34: This can have `uses_polling = False`.

markdroth (commented):
Thanks for doing this!

src/core/ext/filters/client_channel/http_proxy.cc
yashykt: Nice!

src/core/ext/filters/client_channel/http_proxy.cc
```diff
@@ -161,53 +165,40 @@ bool HttpProxyMapper::MapName(const char* server_uri,
...
       const_cast<char*>(GRPC_ARG_HTTP_CONNECT_SERVER),
-      const_cast<char*>(server_target.c_str()));
+      const_cast<char*>(server_target.c_str())));
   if (user_cred != nullptr) {
     /* Use base64 encoding for user credentials as stated in RFC 7617 */
-    char* encoded_user_cred =
-        grpc_base64_encode(user_cred, strlen(user_cred), 0, 0);
+    auto encoded_user_cred =
+        UniquePtr<char>(grpc_base64_encode(user_cred, strlen(user_cred), 0, 0));
     std::string header =
```
markdroth, line 192: This needs to be declared outside of this scope, or else the entry in `args_to_add` will be pointing to already-freed memory when we call `grpc_channel_args_copy_and_add()`.

markdroth (commented):
Looks great, modulo one minor bug and some test failures.

src/core/ext/filters/client_channel/http_proxy.cc
yashykt: That's exactly why the tests were failing. Done!

markdroth (approved):
Looks great!

yashykt:
Thanks for reviewing!

PR #25272, opened 2021-01-26
Implement 3rd-party (envoy) proto generation mechanism

lidizheng:
This PR improves our infrastructure code to support 3rd party proto dependencies, and allow them to generated C++ stubs via protoc compilation. It works for both Bazel and CMake.

The 3rd party proto libraries supported by this PR:
* envoyproxy/data-plane-api
* googleapis/googleapis
* cncf/udpa
* envoyproxy/protoc-gen-validate
* census-instrumentation/opencensus-proto
* ProtoBuf's well known protos (switched them to this path)

# Bazel

For Bazel compilation, the general idea is letting Bazel treat `envoy-api` as a sub-repository. It's done by injecting a WORKSPACE file for it. But to avoid circular dependencies, we need to explicitly depend on transitive libraries. Most of the transitive libraries are already included in gRPC as Git submodules, except `opencensus-proto` which was needed by Envoy's [`opencensus.proto`](https://github.com/envoyproxy/data-plane-api/blob/main/envoy/config/trace/v3/opencensus.proto). The added `grpc_deps` are:

* googleapis
* re2
* gazelle
* opencensus-proto
* protoc-gen-validate
* io_bazel_rules_go (update required by protoc-gen-validate)

One extra setup is added for googleapis to WORKSPACE, so it exposes the APIs in a subset of languages instead of all of them.

~~Current version of `re2` failed ASAN (see [log](https://source.cloud.google.com/results/invocations/c4f25dfd-70e7-472e-b479-86706214dae1/targets/%2F%2Ftest%2Fcpp%2Fend2end:channelz_service_test@poller%3Depoll1/log)). So, this PR disabled ODR check in ASAN (feel free to suggest other ways).~~

One caveat is the proto generation will fail on Windows RBE, since it doesn't have good Golang support (see [log](https://source.cloud.google.com/results/invocations/9ccc7d69-affa-483f-886a-d765e0021536/log)). The Golang tools is required by protoc-gen-validate which is a ProtoBuf plugin wrote by Envoy team, and it needs to be executed for every single Envoy proto files. CC @jtattermusch do you have more context about Golang on Windows RBE?

Overall, Bazel has generally good external library support. Follow the error log and fix the complains one-by-one, and it builds magically in the end.

# ProtoBuf Inconsistent Path Problem

[...]

PR #29219, opened 2022-03-25
[xDS Proto] Update Bazel build dependencies for Envoy API

lidizheng:
First step of https://github.com/grpc/grpc/pull/25272.

---

`third_party/protoc-gen-validate.patch`: The HEAD and latest releases of `protoc-gen-validate` have an incorrect hash value for Golang's ProtoBuf library. It's unclear if this is specific to my local environment (hard to believe another well-known library would have obvious build error), but applying this explicit patch did fix the dependency issue.

Envoy protos depends on the `protoc-gen-validate` to generate C++ proto libraries. It's written in Go, so this PR updates several Go dependencies to make it buildable.

Changes:
1. Upgraded `io_bazel_rules_go`;
2. Upgraded `bazel_gazelle`;
3. Upgraded `com_envoyproxy_protoc_gen_validate` (with patch);
4. Removed submodule `protoc_gen_validate` (this was required when we were trying another CMake solution, no longer required).

CC @veblush


markdroth:
Instead of patching here, can you submit a PR to the proto-gen-validate repo to fix the problem upstream?

lidizheng:
@markdroth Posted https://github.com/envoyproxy/protoc-gen-validate/pull/582 to the upstream repo.

lidizheng:
@jtattermusch PTAL.

For the Go dependency issue, if Envoy folks are acceptive to the change, I will create a follow-up PR to remove the patch.

gnossen (approved):

depthwise:
This PR makes it impossible to cleanly include GRPC as an external dependency into workspaces that already include rules_go. Or at least it's unclear how to do so.

gnossen:
@depthwise Can you please add some more detail? Is the issue the call to `go_register_toolchains`?

gowtham-sundara:
@gnossen Yes, you cannot have 2 go_register_toolchains and let's say I want go1.19 in my repo, it's impossible. The older version allowed for dual registration, the new one from io_rules_go does not and makes using this impossible.

depthwise:
@gnossen yep. Once I patch it out, things work OK. There's got to be a way to check if it's declared already and not re-declare it if it is.

dbernadett:
@gnossen I am also seeing this issue. I'll try some patches, but it would be nice if this didn't have to be done.

PR #29224, opened 2022-03-25
Fix double-space in xmldoc generation.

xclud:
Fix `<summary>` tag in C# xmldoc generation by removing a double-space.

release notes: yes
lang/csharp

@markdroth


jtattermusch (approved):
Thanks for the fix. LGTM.

PR #29225, opened 2022-03-25
Remove C# implementation (individual packages will continue to be maintained through v2.46.x patches or moved to grpc-dotnet)

jtattermusch:
Remove C# from the master branch.
Motivation:
- unblock/simplify planned major changes to the C-core stack by removing Grpc C# (the Grpc.Core implementation) from the build tree.
- reorganize the landscape of gRPC C#/.NET to better match the current support status of gRPC C# (Grpc.Core is currently in maintenance mode and grpc-dotnet is the recommended implementation).

The plan:
- The packages `Grpc`, `Grpc.Core`, `Grpc.Core.Testing`, `Grpc.Core.NativeDebug` and `Grpc.Core.Xamarin` will continue to live in maintenance mode on the https://github.com/grpc/grpc/tree/v1.46.x branch (and there will be occasional patch releases of thereof - e.g. when there's a need for critical/security fix).

- The packages `Grpc.Core.Api`, `Grpc.Auth`, `Grpc.HealthCheck` and `Grpc.Reflection` will be moved to https://github.com/grpc/grpc-dotnet where their development will continue (note that all these packages are implementation-agnostic, and they are also used by grpc-dotnet, so it makes sense to move them there).

- The package `Grpc.Tools` (which provides the protobuf/grpc codegen build integration) will for now stay on the master branch of grpc/grpc (from where it will continue to be released along with other gRPC languages). There are more technical details that need to be figured out and this packages is quite tightly bound to the build infrastructure in the grpc/grpc repository (the artifact/packages build flow, the protoc artifact build etc.) and it also relies on having an up-to-date version of protoc (which comes from third_party/protobuf and would be difficult to upgrade on the v1.46.x branch). The eventual goal is to also move `Grpc.Tools` to elsewhere (probably the grpc-dotnet repository), but more work is needed there.
For now, this is a good solution since this package only depends on protoc and grpc protoc plugins, and is completely independent from C-core.


jtattermusch:
CC @JamesNK

jtattermusch:
Ideally, we'd only merge once the PRs that move some of the packages to grpc-dotnet are all fully ready and approved:
- https://github.com/grpc/grpc-dotnet/pull/1651
- https://github.com/grpc/grpc-dotnet/pull/1652
- https://github.com/grpc/grpc-dotnet/pull/1708

jtattermusch:
Known failures:
b/230080881

apolcyn (approved):

jtattermusch:
- https://github.com/grpc/grpc-dotnet/pull/1651 has already been merged 
- https://github.com/grpc/grpc-dotnet/pull/1652 and https://github.com/grpc/grpc-dotnet/pull/1708 are fully ready and are waiting for final approval (they will be merged very soon).

So it seems nothing prevents this PR from being merged.

PR #29229, opened 2022-03-25
Revert "Fix prefork handler register's default behavior"

lidizheng:
Reverts grpc/grpc#29103

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -391,7 +391,10 @@ if "win32" in sys.platform:
         # on msvc, but only for 32 bits
         DEFINE_MACROS += (('NTDDI_VERSION', 0x06000000),)
 else:
-    DEFINE_MACROS += (('HAVE_CONFIG_H', 1),)
+    DEFINE_MACROS += (
+        ('HAVE_CONFIG_H', 1),
+        ('GRPC_ENABLE_FORK_SUPPORT', 1),
+    )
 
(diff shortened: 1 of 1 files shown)


lidizheng:
@gnossen PTAL.

gnossen (approved):

Stephan-Walter:
We experience right now excessive log spam with our python applications. With respect to my understanding, the root cause is, that the fork is enabled with grpc 1.47.0 per default. In the past this was not the case, and with respect to my understanding right now, it is consent that this should be reverted. Is this correct?

Right now 1.47.0 is still the newest version on PyPi. When will we see a newer version there, and will it already contain the old default setting?

gnossen:
> With respect to my understanding, the root cause is, that the fork is enabled with grpc 1.47.0 per default

Yes. And it is also enabled by default in all versions for the past several years.

> In the past this was not the case
No, fork support has been enabled by default in all versions for the past several years (basically since the feature was introduced).

> it is consent that this should be reverted. Is this correct?
This revert PR re-enables fork support by default. https://github.com/grpc/grpc/pull/29103 was an attempt to turn this off by default, but when we pushed a release with this setting off by default, we saw massive breakage in our users who were using this feature without realizing.

I think future efforts to address this problem should try a different strategy.

Issue #28557, opened 2022-01-14
Python TSI_DATA_CORRUPTED after Popen()


spikecurtis:
### What version of gRPC and what language are you using?
Python 3.8.10
grpcio 1.43.0

### What operating system (Linux, Windows,...) and version?
Ubuntu 20.04

### What runtime / compiler are you using (e.g. python version or version of gcc)
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux

### What did you do?
When I use a gRPC client to make rpc calls immediately after a call to subprocess.Popen(), intermittently, the next rpc call fails. 
### What did you expect to see?

My rpc should return without error.

### What did you see instead?

```
"{"created":"@1642115822.017133164","description":"Error received from peer ipv4:127.0.0.1:50686","file":"src/core/lib/surface/call.cc","file_line":1074,"grpc_message":"Stream removed","grpc_status":2}"
```

Turning on additional gRPC logging with `GRPC_VERBOSITY=debug GRPC_TRACE=tsi,subchannel,subchannel_pool,transport_security`,

I see

```
I0113 15:17:01.721348644  906598 ssl_transport_security.cc:223]      HANDSHAKE START -       TLS client start_connect  - !!!!!!
I0113 15:17:01.721527927  906598 ssl_transport_security.cc:223]                 LOOP -    TLS client enter_early_data  - !!!!!!
I0113 15:17:01.721586410  906598 ssl_transport_security.cc:223]                 LOOP -   TLS client read_server_hello  - !!!!!!
I0113 15:17:01.727443284  899282 ssl_transport_security.cc:223]                 LOOP - TLS 1.3 client read_hello_retr  - !!!!!!
I0113 15:17:01.727508497  899282 ssl_transport_security.cc:223]                 LOOP - TLS 1.3 client read_server_hel  - !!!!!!
I0113 15:17:01.727732797  899282 ssl_transport_security.cc:223]                 LOOP - TLS 1.3 client read_encrypted_  - !!!!!!
I0113 15:17:01.727762972  899282 ssl_transport_security.cc:223]                 LOOP - TLS 1.3 client read_certificat  - !!!!!!
I0113 15:17:01.727831564  899282 ssl_transport_security.cc:223]                 LOOP - TLS 1.3 client read_server_cer  - !!!!!!
[...]

PR #29232, opened 2022-03-25
Change the unsupported polling strategy log for forking to GPR_INFO

lidizheng:
As title.

Related https://github.com/grpc/grpc/issues/29044

### DIFF
--- a/src/core/lib/iomgr/fork_posix.cc
+++ b/src/core/lib/iomgr/fork_posix.cc
@@ -67,7 +67,7 @@ void grpc_prefork() {
   if (poll_strategy_name == nullptr ||
       (strcmp(poll_strategy_name, "epoll1") != 0 &&
        strcmp(poll_strategy_name, "poll") != 0)) {
-    gpr_log(GPR_ERROR,
+    gpr_log(GPR_INFO,
             "Fork support is only compatible with the epoll1 and poll polling "
             "strategies");
     return;


gnossen (approved):

drfloob:
Does changing the log level solve a problem? It was set to error-level logging due to mysterious failures seen in https://github.com/grpc/grpc/issues/28557

lidizheng:
@drfloob  Nope. It doesn't. The mysterious failure should be a GCF specific bug.

We are in a dilemma here. If we log it as ERROR, people could fix https://github.com/grpc/grpc/issues/28557 more easily, but people also got unnecessarily noticed https://github.com/grpc/grpc/issues/29044.

Overall, for https://github.com/grpc/grpc/issues/28557, we updated our document. Reverting this line to previous state is the lesser evil.

drfloob:
Makes sense, thanks for explaining @lidizheng

drfloob (approved):

lidizheng:
(Plan to hold this for another day, I hope to see if we can dig more into the mutex issue.)

lidizheng:
Merging now, as we are unable to produce the error any more, until we found the concrete root cause, I think we will stuck with the inconsistent fork handlers (injected in Core, not in Python) at this moment to maintain the backward compatibility.

JMMarchant:
Similarly seeing the `skipping fork() handlers` message

PR #29235, opened 2022-03-28
do not publish vanilla linux arm32 wheels that break musl OS installs

### LINKED ISSUE #27512: Unable to run grpcio wheel on alpine
`1.41.0` in python

`arm32v7/alpine:3.12`

`Python 3.8.10`

### What did you do?

Given an example `Dockerfile` like this:

```
FROM arm32v7/alpine:3.12

RUN apk add --no-cache python3 py3-pip
RUN apk add libstdc++ libc6-compat
RUN python3 --version

WORKDIR /code
COPY requirements.txt .
COPY example.py .

RUN pip3 install --no-cache-dir --only-binary=grpcio -r requirements.txt

CMD python3 ./example.py
```

Note that this also has problems finding the shared library without `libstdc++` and `libc6-compat` but the errors has changed across grpc versions.

`requirements.txt`:
```
grpcio==1.41.0
google-cloud-pubsub
```

`example.py`
```
#!/usr/bin/python

from google.cloud import pubsub_v1

def main():
  print("Hello world")

if __name__ == "__main__":
  main()
```

Then run a build like this:

```
$ docker build --tag grpcio-armv7 .
```

```
$ docker run grpcio-armv7
Hello world
```

Note, that this is what I see for alpine images on other architectures or different armv7 images like this:
```
FROM arm32v7/python:3.8-buster
```

### What did you see instead?

```
$ docker run grpcio-armv7
  File "./example.py", line 3, in <module>                                      
    from google.cloud import pubsub_v1                                          
  File "/usr/lib/python3.8/site-packages/google/cloud/pubsub_v1/__init__.py", li
ne 17, in <module>                                                              
    from google.cloud.pubsub_v1 import types                                    
  File "/usr/lib/python3.8/site-packages/google/cloud/pubsub_v1/types.py", line 
25, in <module>                                                                 
    from google.api_core import gapic_v1                                        
  File "/usr/lib/python3.8/site-packages/google/api_core/gapic_v1/__init__.py", 
line 16, in <module>                                                            
    from google.api_core.gapic_v1 import config                                 
  File "/usr/lib/python3.8/site-packages/google/api_core/gapic_v1/config.py", line 23, in <module>
    import grpc
  File "/usr/lib/python3.8/site-packages/grpc/__init__.py", line 22, in <module>
    from grpc import _compression
  File "/usr/lib/python3.8/site-packages/grpc/_compression.py", line 15, in <module>
ImportError: Error relocating /usr/lib/python3.8/site-packages/grpc/_cython/cygrpc.cpython-38-arm-linux-gnueabihf.so: _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE7reserveEv: symbol not found

When using an older version of `grpcio` such as 1.38.X, I see an error message like:
```
  File "./example.py", line 3, in <module>
    from google.cloud import pubsub_v1
  File "/usr/lib/python3.8/site-packages/google/cloud/pubsub_v1/__init__.py", line 17, in <module>
    from google.cloud.pubsub_v1 import types
  File "/usr/lib/python3.8/site-packages/google/cloud/pubsub_v1/types.py", line 25, in <module>
    from google.api_core import gapic_v1
  File "/usr/lib/python3.8/site-packages/google/api_core/gapic_v1/__init__.py", line 16, in <module>
    from google.api_core.gapic_v1 import config
  File "/usr/lib/python3.8/site-packages/google/api_core/gapic_v1/config.py", line 23, in <module>
    import grpc
  File "/usr/lib/python3.8/site-packages/grpc/__init__.py", line 23, in <module>
    from grpc._cython import cygrpc as _cygrpc
ImportError: Error relocating /usr/lib/python3.8/site-packages/grpc/_cython/cygrpc.cpython-38-arm-linux-gnueabihf.so: __strftime_l: symbol not found
```

This is similar to the first post of https://github.com/grpc/grpc/issues/24808 where the armv7 binary wheel has problems with alpine.


This seems to have been an `armv7` plus `alpine` issue that first happened last year. I have some history in https://github.com/home-assistant/core/issues/56669 though this was first seen with https://github.com/home-assistant/core/issues/40266 and https://github.com/home-assistant/core/issues/40148

The discussion in https://github.com/grpc/grpc/issues/26620 seems quite relevant.  Is building musllinux wheels for armv7 the right solution?

aptalca:
Currently, vanilla wheels for linux arm32/armv7l are published on pypi. Unfortunately, these wheels are glibc based, but pip has no understanding of that fact. Therefore, when one tries to `pip install grpcio` on a musl based OS like alpine on armv7l, pip happily installs the vanilla glibc wheel from pypi and results in a broken package install.

For x86_64 and aarch64, the wheels published are not vanilla and are container based (manylinux). These container based wheels are properly recognized by pip as glibc based and not compatible with Alpine, thus `pip install grpcio` on alpine x86_64 and aarch64 result in compilation from source (as expected), resulting in working packages.

Since pypi is a universal repo, I don't believe the vanilla cpython wheels should be published there as they break installs on non-glibc platforms. In this case, the vanilla arm32/armv7l wheels are probably better suited for non-universal repos like piwheels.org.

For x86_64 and aarch64, container based wheels like manylinux and musllinux are perfectly appropriate as pip recognizes their platform compatibility. Unfortunately there is no container support for arm32, therefore uploading official arm32 wheels to pypi is not recommended (relevant only for cpython wheels, and not pure python as pure python wheels do not have a dependency on glibc or musl and are universal).

This PR attempts to remove the publishing of vanilla arm32/armv7l linux wheels to pypi. I'm not super familiar with your ci so I attempted this as a best effort. There may be other modifications necessary that I missed.

You can see a more in depth comment on the issue here: https://github.com/grpc/grpc/issues/27512#issuecomment-1030884361

Thanks

closes #27512 

EDIT: I'd be happy to post logs of failures on alpine armhf with easily reproducible docker build lines using qemu if needed.


lidizheng:
To me, it's a bit surprising that `pip` won't automatically figure out the platform and cpu architecture.

cc @jtattermusch might have more context here.

aptalca:
Right. With vanilla wheels, pip goes by the cpython version and the arch, but it has no idea about the shared library dependencies. In this case it's glibc vs musl.

A more pronounced issue for example is with cffi where even though alpine 3.14 and 3.15 have the same cpython version, a vanilla cffi wheel built on 3.14 does not work on 3.15 (on the same arch) due to differences in system libffi version in the two OS versions.

The ultimate solution is to provide container based wheels like manylinux and musllinux (where these deps can be included), which you guys already do for x86_64 and aarch64. But as I mentioned before those containers don't support arm32 so it's best not to publish a wheel which is not universal.

We publish vanilla wheels in our repo for use in our docker images, but we have to separate them not just by glibc vs musl, but also by alpine version:
https://wheel-index.linuxserver.io/alpine/ (for alpine 3.14)
https://wheel-index.linuxserver.io/alpine-3.15
https://wheel-index.linuxserver.io/ubuntu/
and during build, we specify the repo for each image (ie. `pip install --find-links https://wheel-index.linuxserver.io/alpine-3.15/`)

gnossen:
This is indeed surprising behavior out of pip. But based on your recommendation, vanilla linux wheels should _never_ be published because we can never be certain of the deployment libc (or a host of other factors). Nowadays, with container-based wheel specifications becoming more common, this seems much more reasonable. But back when these artifacts were first introduced, there really didn't seem to be any alternative.

We do often direct people to [piwheels](https://www.piwheels.org/project/grpcio/). However, these ARM artifacts have been here for _years_. It might come as quite a surprise to our users if they were removed. We would at least need to make a proposal on the Google group and wait a week or so for comments to come in.

@aptalca A question to you. We don't currently publish any musl artifacts. What behavior do you expect on an arm32 alpine machine? A from-source build?

aptalca:
@gnossen is there any update on the discussion within the community? Should I bother rebasing this since there have been several updates to the repo that created conflicts?

Thanks

gnossen:
@aptalca Sorry I missed your last ping. Yes, we are now publishing x86 musl wheels, but that doesn't really solve your issue.

I still think that removing the distroless armv7 artifacts could be quite jarring to users. Bigquery indicates that the armv7 artifacts get ~4K downloads per day:

AitxiaHKxC3nLWT

I don't think the difficult part here is the code change -- it's gaining the community consensus. Can you please write up your arguments to [the grpc-io mailing list](https://groups.google.com/g/grpc-io)? If there are no objections after two weeks, we should be able to go forward with this.

aptalca:
@gnossen thanks so much for the heads up. I'll go ahead and post there and will propose moving the armv7 wheel to a platform specific repo like https://www.piwheels.org/ which is a lot more suitable.

aptalca:
Added: https://groups.google.com/u/8/g/grpc-io/c/KOsb0tqRcDM

aptalca:
@gnossen thanks for the feedback on this, but it doesn't seem like anyone else had any interest in this issue.

It still remains an issue where a basic `pip install grpcio` on a musl based armhf device will install a broken package.

However, we, as an org, decided to deprecate armhf for all of our images in a couple of months and we already deprecated our Home Assistant armhf build, which is where grpcio is used.

So I also no longer have an interest in this issue. Closing.

PR #29238, opened 2022-03-28
Adding http/1.1 support in httpcli.

### LINKED ISSUE #19173: httpcli should support HTTP 1.1
httpcli [uses](https://github.com/grpc/grpc/blob/master/src/core/lib/http/format_request.cc#L36) HTTP/1.0. 
Istio Envoy requires HTTP/1.1 or above. HTTP 1.0 support is available in Istio 1.1 ([PR](https://github.com/istio/istio/issues/11235)), which is still in [alpha](https://cloud.google.com/istio/docs/istio-on-gke/installing) on GKE. On Istio 1.1, users still need to set an environment HTTP10 to enable HTTP/1.0 support.

When using grpc httpcli to contact the metadata server, requests are intercepted by Envoy proxy which returns HTTP 426 Upgrade required for any HTTP/1.0 requests.
 
 1.21
 
 Linux, Ubuntu 14.04
 
### What did you do?
httpcli request to "http://metadata.google.internal.//computeMetadata/v1/instance/service-accounts/default/token" with required "Metadata-Flavor: Google" header
 
 HTTP status 200 and a token
 
### What did you see instead?
HTTP Status 426 upgrade required
 
Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).
Patched [format_request.cc](https://github.com/grpc/grpc/blob/master/src/core/lib/http/format_request.cc#L36) to 1.1 and the request succeeds.

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.
 
https://github.com/GoogleCloudPlatform/cloud-profiler-java
and other projects where grpc httpcli is used to contact the GCP metadata server.

nicolasnoble:
Fixes #19173.

Supersedes #29226.


ctiller:
Hey I think this looks good. Can you try running the two fuzzers in test/core/http for like... 20-30 minutes each before submission just to squeeze out anything problematic?

something like:
```
bazel run test/core/http/response_fuzzer -c dbg --config fuzzer_asan
```

(or better if you have an idle 100 cores somewhere:
```
bazel run test/core/http/response_fuzzer -c dbg --config fuzzer_asan -- -jobs=100 -workers=100
```
)

ctiller (approved):

nicolasnoble:
Good point. I have roughly 50 idle cores at home so I'll try that for some time :)

nicolasnoble:
Ran this over 50 cores for over an hour without issue.

PR #29253, opened 2022-03-29
xds: add node ID to status messages generated by XdsClient

markdroth:
We don't currently have any tests that check the status messages.  That's something we should definitely fix, but it will be part of a broader effort related to reliability.  In the interim, I have manually verified that this works.

### DIFF
--- a/src/core/ext/xds/xds_client.cc
+++ b/src/core/ext/xds/xds_client.cc
@@ -2163,8 +2163,9 @@ void XdsClient::ResetBackoff() {
 void XdsClient::NotifyOnErrorLocked(absl::Status status) {
   const auto* node = bootstrap_->node();
   if (node != nullptr) {
-    status = absl::Status(status.code(), absl::StrCat(
-        status.message(), " (node ID:", bootstrap_->node()->id, ")"));
+    status = absl::Status(
+        status.code(), absl::StrCat(status.message(),
+                                    " (node ID:", bootstrap_->node()->id, ")"));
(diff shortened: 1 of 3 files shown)


apolcyn:
just an idea, could this test be leveraged: https://github.com/grpc/grpc/blob/93d47104daa789046b1d79db23faf52747450f58/test/cpp/end2end/xds/xds_end2end_test.cc#L2501 to also check for node ID?

yashykt (approved):

src/core/ext/xds/xds_client.cc
```diff
@@ -2200,6 +2161,11 @@ void XdsClient::ResetBackoff() {
 }
 
 void XdsClient::NotifyOnErrorLocked(absl::Status status) {
+  const auto* node = bootstrap_->node();
+  if (node != nullptr) {
+    status = absl::Status(status.code(), absl::StrCat(
```
yashykt, line 2166: would this capture everything we need from `status`?

yashykt: For example, if a payload was set on the status, this would miss that. cc @veblush

markdroth: We never add any payload for any of the statuses that we generate internally.

markdroth (commented):
I've added the test suggested by @apolcyn.  It's not a comprehensive solution, but it does at least check that the output is right in one case.

PR #29254, opened 2022-03-29
[xDS Proto] Enhence gRPC buildgen for 3rd party proto compilation

lidizheng:
Second step of https://github.com/grpc/grpc/pull/25272.

---

This PR includes:

* A CMake feature allowing gRPC build dependencies to be downloaded during the build time;
  * See `download_archive.cmake`, the dependency information (name, prefix, download url) is extracted from `grpc_deps.bzl`
* A CMake feature to generate C++ files for protos in third party folders.
  * This PR copies the proto files into a centralized location, see below section for why.

---

# ProtoBuf Inconsistent Path Problem

Without Bazel, we have to directly interact with `protoc`, which has some quirks that it is very picky about paths. One needs to find the right combination of working directory (`pwd`), import directory (`-I`), and import path (how one proto is meant to be imported). For example, let's say we have 2 proto files each in a different 3rd_party library:

```protobuf
# third_party/lib_a/a.proto
...
message NotImportant {}

# third_party/lib_b/b.proto

import "a.proto";
...
```

| Proto file location                        | Current working directory | Expected Descriptor Name for "a.proto" | Generated Descriptor Path                          | How "b.pb.h" Imports "a.pb.h" | Generated "a.pb.h" Header Path           |
| ------------------------------------------ | ------------------------- | -------------------------------------- | -------------------------------------------------- | ----------------------------- | ---------------------------------------- |
| Untouched                                  | .                         | descriptor_table_a_proto               | descriptor_table_third_party_lib_a_a_proto &cross; | #include "a.pb.h"             | $OUTPUT/third_party/lib_a/a.pb.h &cross; |
| Untouched                                  | third_party               | descriptor_table_a_proto               | descriptor_table_lib_a_a_proto &cross;             | #include "a.pb.h"             | $OUTPUT/lib_a/a.pb.h &cross;             |
| Both copied to project root                | .                         | descriptor_table_a_proto               | descriptor_table_a_proto &check;                   | #include "a.pb.h"             | $OUTPUT/a.pb.h  &check;                  |
| Only **b.proto** is copied to project root | .                         | descriptor_table_a_proto               | descriptor_table_third_party_lib_a_proto &cross;   | #include "a.pb.h"             | $OUTPUT/third_party/lib_a/a.pb.h &cross; |

As above, the ProtoBuf C++ generated code brutal-forcely translate the import line `import "third_party/lib_a/a.proto";` to how its dependency headers should be imported, and what its dependency descriptor should be named. But it somehow generated descriptor name based on relative path. The inconsistent header file path is annoying but not blocking the build. However, the difference in descriptor name will cause the build to fail deterministically. This issue troubles many ProtoBuf adopters that it blocks people from using multiple repos to store proto files.

To make `protoc` work, we can simulate how it is intended to be used in Google, which is gathering the build metadata, find the expected import path, and rebuild the directories, like a monorepo, or like Bazel.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -429,47 +429,18 @@ function(protobuf_generate_grpc_cpp)
     return()
   endif()
 
-  set(_protobuf_include_path -I . -I ${_gRPC_PROTOBUF_WELLKNOWN_INCLUDE_DIR})
   foreach(FIL ${ARGN})
-    get_filename_component(ABS_FIL ${FIL} ABSOLUTE)
-    get_filename_component(FIL_WE ${FIL} NAME_WE)
-    file(RELATIVE_PATH REL_FIL ${CMAKE_CURRENT_SOURCE_DIR} ${ABS_FIL})
-    get_filename_component(REL_DIR ${REL_FIL} DIRECTORY)
-    set(RELFIL_WE "${REL_DIR}/${FIL_WE}")
-
-    #if cross-compiling, find host plugin
-    if(CMAKE_CROSSCOMPILING)
-      find_program(_gRPC_CPP_PLUGIN grpc_cpp_plugin)
-    else()
-      set(_gRPC_CPP_PLUGIN $<TARGET_FILE:grpc_cpp_plugin>)
-    endif()
-
-    add_custom_command(
-      OUTPUT "${_gRPC_PROTO_GENS_DIR}/${RELFIL_WE}.grpc.pb.cc"
-             "${_gRPC_PROTO_GENS_DIR}/${RELFIL_WE}.grpc.pb.h"
-             "${_gRPC_PROTO_GENS_DIR}/${RELFIL_WE}_mock.grpc.pb.h"
-             "${_gRPC_PROTO_GENS_DIR}/${RELFIL_WE}.pb.cc"
-             "${_gRPC_PROTO_GENS_DIR}/${RELFIL_WE}.pb.h"
-      COMMAND ${_gRPC_PROTOBUF_PROTOC_EXECUTABLE}
-      ARGS --grpc_out=generate_mock_code=true:${_gRPC_PROTO_GENS_DIR}
-           --cpp_out=${_gRPC_PROTO_GENS_DIR}
-           --plugin=protoc-gen-grpc=${_gRPC_CPP_PLUGIN}
-           ${_protobuf_include_path}
-           ${REL_FIL}
-      DEPENDS ${ABS_FIL} ${_gRPC_PROTOBUF_PROTOC} ${_gRPC_CPP_PLUGIN}
-      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-      COMMENT "Running gRPC C++ protocol buffer compiler on ${FIL}"
-      VERBATIM)
+    protobuf_generate_grpc_cpp_with_import_path_correction(${FIL} ${FIL})
   endforeach()
 endfunction()
 
-#  protobuf_generate_grpc_cpp_with_import_path
+#  protobuf_generate_grpc_cpp_with_import_path_correction
 #  --------------------------
 #
 #   Add custom commands to process ``.proto`` files to C++ using protoc and
 #   GRPC plugin::
 #
-#     protobuf_generate_grpc_cpp_with_import_path <FILE_LOCATION> <IMPORT_PATH>
+#     protobuf_generate_grpc_cpp_with_import_path_correction <FILE_LOCATION> <IMPORT_PATH>
 #
 #   ``FILE_LOCATION``
 #     The relative path of the ``.proto`` file to the project root
@@ -482,7 +453,7 @@ endfunction()
 #     correct symbol being generated and C++ include path being correct.
 #     More info can be found at https://github.com/grpc/grpc/pull/25272.
 #
-function(protobuf_generate_grpc_cpp_with_import_path FILE_LOCATION IMPORT_PATH)
+function(protobuf_generate_grpc_cpp_with_import_path_correction FILE_LOCATION IMPORT_PATH)
   if(NOT FILE_LOCATION)
     message(SEND_ERROR "Error: PROTOBUF_GENERATE_GRPC_CPP() called without any proto files")
     return()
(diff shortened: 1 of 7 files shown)


lidizheng:
Protoc compilation in Artifact Build Windows is failed due to `No error`:

```
CMake Error at cmake/download_archive.cmake:35 (file):
  file RENAME failed to rename
    T:/altsrc/github/grpc/workspace_protoc_windows_x64/cmake/build/http_archives/xds-cb28da3451f158a947dfc45090fe92b07b243bc1
  to
    T:/altsrc/github/grpc/third_party/udpa
  because: No error
```

lidizheng:
@jtattermusch @gnossen PTAL.

gnossen (approved):
I'm not a cmake expert, but this looks quite impressive, especially considering it's generated from the Bazel build.

tools/buildgen/extract_metadata_from_bazel_xml.py
```diff
@@ -45,6 +45,48 @@
...
+    'com_github_cncf_udpa':
+        ExternalProtoLibrary(destination='third_party/xds',
+                             proto_prefix='third_party/xds/'),
+    'opencensus_proto':
+        ExternalProtoLibrary(destination='third_party/opencensus-proto/src',
+                             proto_prefix='third_party/opencensus-proto/src/'),
+}
+
+
+def _maybe_is_supported_external_lib(name: str) -> Optional[str]:
```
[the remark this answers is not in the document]
lidizheng: Updated. That's much better.

CMakeLists.txt
```diff
@@ -333,49 +414,59 @@ file(MAKE_DIRECTORY ${_gRPC_PROTO_GENS_DIR})
 #   Add custom commands to process ``.proto`` files to C++ using protoc and
 #   GRPC plugin::
 #
-#     protobuf_generate_grpc_cpp [<ARGN>...]
+#     protobuf_generate_grpc_cpp <FILE_LOCATION> <IMPORT_PATH>
```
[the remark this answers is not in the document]
lidizheng: My take is that this is specific to gRPC. Searching "how to compile protobuf with cmake", it seems people are using their own functions.

@jtattermusch correct me if I'm wrong.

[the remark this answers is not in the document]
lidizheng: Searching this function in GitHub, it seems people do depend on this method for their CMake builds: https://github.com/search?l=CMake&q=protobuf_generate_grpc_cpp&type=Code

I'm not sure if we have contracts or not, but I think I can keep the original version at side.

tools/buildgen/extract_metadata_from_bazel_xml.py
```diff
@@ -149,17 +194,39 @@ def _extract_nonpublic_headers(bazel_rule: BuildMetadata) -> List[str]:
 def _extract_sources(bazel_rule: BuildMetadata) -> List[str]:
     """Gets list of source files from a bazel rule"""
     result = []
-    for dep in bazel_rule['srcs']:
-        if dep.startswith('//') and (dep.endswith('.cc') or dep.endswith('.c')
-                                     or dep.endswith('.proto')):
-            result.append(_extract_source_file_path(dep))
+    for src in bazel_rule['srcs']:
+        if src.endswith('.cc') or src.endswith('.c') or src.endswith('.proto'):
+            if src.startswith('//'):
```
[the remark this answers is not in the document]
gnossen: It's probably a mostly theoretical concern. If it does cause a problem, we should be able to tell from broken tests.

test/distrib/cpp/run_distrib_test_cmake.bat
```diff
@@ -68,7 +68,11 @@ popd
 
 @rem Just before installing gRPC, wipe out contents of all the submodules to simulate
 @rem a standalone build from an archive
-git submodule deinit --all --force
+@rem NOTE(lidiz) We used to use "git submodule deinit", but it leaves an empty
+@rem folder for deinit-ed submodules, blocking the CMake download. For users
+@rem downloaded gRPC code as an archieve, they won't have submodule residual
```
jtattermusch, line 73: type: archieve

cmake/download_archive.cmake
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set(_download_archive_TEMPORARY_DIR ${CMAKE_BINARY_DIR}/http_archives)
+file(MAKE_DIRECTORY ${_download_archive_TEMPORARY_DIR})
+
+# This is basically Bazel's http_archive.
+# Note that strip_prefix strips the directory path prefix of the extracted
+# archive content, and it may strip multiple directories.
+function(download_archive destination url hash strip_prefix)
```
jtattermusch, line 21: note that unlike bazel, you only support a single URL here (and when I look at the generated CmakeLists.txt file, sometimes the "primary" url is the one from mirror and sometimes the one from primary source.
Looks like we should support multiple URLs here, to reach parity with what we have in the bazel_deps.bzl

jtattermusch: Idea: perhaps you e.g. pass the URLs as a single "url" param separated e.g. by semicolon?

"https://storage.googleapis.com/grpc-bazel-mirror/github.com/envoyproxy/data-plane-api/archive/9c42588c956220b48eb3099d186487c2f04d32ec.tar.gz;https://github.com/envoyproxy/data-plane-api/archive/9c42588c956220b48eb3099d186487c2f04d32ec.tar.gz"

jtattermusch, line 21: qq: have you investigated whether cmake already has some useful features we could reuse that are sophisticated than `file(DOWNLOAD)`? I want to avoid us reinventing the wheel.

CMakeLists.txt
```diff
@@ -318,11 +318,92 @@ include(cmake/ssl.cmake)
 include(cmake/upb.cmake)
 include(cmake/xxhash.cmake)
 include(cmake/zlib.cmake)
+include(cmake/download_archive.cmake)
+
+# Setup external proto library at third_party/envoy-api with 2 download URLs
+if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api)
+  # Download the archive via HTTP, validate the checksum, and extract to third_party/envoy-api.
+  download_archive(
```
jtattermusch, line 326: note that this basically kills the idea of having offline cmake builds (which might or might not be a good idea, I'll have to give that a thought).

tools/buildgen/extract_metadata_from_bazel_xml.py
```diff
@@ -1103,6 +1212,14 @@ def _detect_and_print_issues(build_yaml_like: BuildYaml) -> None:
 #   'tests': { TARGET_DICT_FOR_TEST_XYZ, ...} }
 build_yaml_like = _convert_to_build_yaml_like(all_targets_dict)
 
+# Step 7: generates build metadata for external ProtoBuf libraries.
+# We only want the ProtoBuf sources from these ProtoBuf dependencies, which may
+# not present in our release source tar balls. These rules will be used in CMake
```
jtattermusch, line 1217: nit: typo: "which may not [be] present"

tools/buildgen/plugins/check_attrs.py
```diff
@@ -89,6 +89,13 @@ def subset_of(values):
         'vs_proj_dir': anything(),
         'zlib': one_of((True,)),
     },
+    'external_library': {
```
jtattermusch, line 92: you're calling this "external_library", but the only instances we currently plan to have are all "external proto libraries", so let's give some thought to how this should be named. Also the "proto_prefix" attributes signifies that this has something to do with protos.

From some other perspective, these are "external archives" to download rather than libraries.

Btw, we currently import the entire third_party/upb subtree into our git repo (because ubp isn't installable on its own so we embed it in our build), so perhaps we could consider making upb and "external library" as well.
CC @veblush

CMakeLists.txt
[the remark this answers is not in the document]
lidizheng: The original version kept as `protobuf_generate_grpc_cpp` (no change), new generation function renamed to `protobuf_generate_grpc_cpp_with_import_path`.

test/distrib/cpp/run_distrib_test_cmake.bat
lidizheng: Good catch! Fixed.

cmake/download_archive.cmake
lidizheng: I'm taking a slightly more verbose approach here. In the CMake template, we will iterate through all download URLs and generate a `download_archive` call for each URL:

```cmake
# Setup external proto library at third_party/opencensus-proto/src with 2 download URLs
if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src)
  # Download the archive via HTTP, validate the checksum, and extract to third_party/opencensus-proto/src.
  download_archive(
    ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src
    https://storage.googleapis.com/grpc-bazel-mirror/github.com/census-instrumentation/opencensus-proto/archive/v0.3.0.tar.gz
    b7e13f0b4259e80c3070b583c2f39e53153085a6918718b1c710caf7037572b0
    opencensus-proto-0.3.0/src
  )
endif()
if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src)
  # Download the archive via HTTP, validate the checksum, and extract to third_party/opencensus-proto/src.
  download_archive(
    ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src
    https://github.com/census-instrumentation/opencensus-proto/archive/v0.3.0.tar.gz
    b7e13f0b4259e80c3070b583c2f39e53153085a6918718b1c710caf7037572b0
    opencensus-proto-0.3.0/src
  )
endif()
```

lidizheng: There is a `ExternalProject` suite of commands: https://cmake.org/cmake/help/latest/module/ExternalProject.html

To use it, we might need to bump up our minimum CMake version to at least 3.7 (we are on 3.5.1 now). WDYT?

tools/buildgen/extract_metadata_from_bazel_xml.py
lidizheng: Good catch! Fixed.

CMakeLists.txt
lidizheng: I see 2 alternatives here:

1. Make all dependencies here sub-tree (like what we did for upb, xxhash: https://github.com/grpc/grpc/tree/master/third_party);
2. Stop supporting C++ xDS related targets in CMake (CSDS, Admin, and several tests).

WDYT?

lidizheng (commented):
Thanks for the review. A few comments might demand more discussion.

tools/buildgen/plugins/check_attrs.py
lidizheng: Renamed to `external_proto_library` (feel free to suggest other names).

Yep, I do plan to look into the possibility of reducing our subtrees after the copied xds protos are gone. I recall one of the blocker: supporting `Make` without subtree is hard.

cmake/download_archive.cmake
jtattermusch: Alright, I didn't notice that. I think your current approach has a reasonable balance between simplicity and robustness. Thanks for the explanation.

jtattermusch: I think ExternalProject has been available much earlier than cmake 3.7 (unless you need some specific features that were introduced in 3.7. do you?), but there might be other issues with using ExternalProject.

Generally, the problem with ExternalProject is that it only download the dependencies in the "build" phase, which might be too late in some cases (e.g. when you need the external project to do something in the "configure" phase). This might be ok for the external proto libraries, since they don't have a CMakeLists.txt themselves, but it would not work in some more complex cases (e.g. download UPB dependency and than include it in a build).
IIRC the `ExternalProject` isn't very popular due to its significant limitations.

In cmake 3.11 version, the `FetchContent` was introduced, which is much more flexible and usable as AFAICT, it's now the "recommended" one to use over the `ExternalProject`. 
https://cmake.org/cmake/help/latest/module/FetchContent.html


Quoting from the docs:
> This module enables populating content at configure time via any method supported by the [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html#module:ExternalProject) module. Whereas [ExternalProject_Add()](https://cmake.org/cmake/help/latest/module/ExternalProject.html#command:externalproject_add) downloads at build time, the FetchContent module makes content available immediately, allowing the configure step to use the content in commands like [add_subdirectory()](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory), [include()](https://cmake.org/cmake/help/latest/command/include.html#command:include) or [file()](https://cmake.org/cmake/help/latest/command/file.html#command:file) operations.

See e.g. https://bewagner.net/programming/2020/05/02/cmake-fetchcontent/ for more info.


TL;DR from this is that for now we can keep the current `file(DOWNLOAD ` and maybe in the future we can look into converting our build to use FetchContent (some time in the future, when requiring e.g. cmake 3.14 isn't an issue for us), that has some potential for simplifying our build.

CMakeLists.txt
jtattermusch: Let's use the approach suggested in this PR.
Btw, it seems that if you pre-populate the right directories (e.g. `third_party/envoy-api`) with what it should contains, there will be no need to perform a download.

templates/CMakeLists.txt.template
```diff
@@ -416,6 +452,67 @@
...
+  #   ``IMPORT_PATH``
+  #     The proto import path that itself expected to be placed in. For
+  #     example, a "bar.proto" file wants to be imported as
+  #     `import "foo/bar.proto"`. Then we should place it under
+  #     "<ProtoBuf_Include_Path>/foo/bar.proto" instead of
+  #     "<ProtoBuf_Include_Path>/third_party/foo/bar.proto". This ensures
+  #     correct symbol being generated and C++ include path being correct.
+  #     More info can be found at https://github.com/grpc/grpc/pull/25272.
+  #
+  function(protobuf_generate_grpc_cpp_with_import_path FILE_LOCATION IMPORT_PATH)
```
jtattermusch, line 474: Looks like this functions is 95% the same as the original `protobuf_generate_grpc_cpp` you copied.
Instead of duplicating the code (bad for maintainability), can you adjust the original function to be able to handle the case that requires putting protos into the centralized location?

Btw, I think many of our .proto in the grpc/grpc repo had their import statements modified to workaround the same issue as you're trying to workaround here.
Perhaps there should be some cleanup afterwards to make sure the import statements in all .proto files (inside and outside of grpc/grpc) are used consistently?

jtattermusch: Also, in the current form, it's quite non-obvious what's actually different about the `protobuf_generate_grpc_cpp` vs `protobuf_generate_grpc_cpp_with_import_path`. It would be better to use a single function everywhere and have only one copy for "protobuf_generate" function (but set different args depending on what our needs are).

jtattermusch: To Illustrate the hacks we've done to the .proto files in our repository, see the same proto in grpc/grpc and grpc-java:

https://github.com/grpc/grpc/blob/61b34dfaee80d36b505fbb488cc73b2afd803c2e/src/proto/grpc/testing/control.proto#L17

https://github.com/grpc/grpc-java/blob/78ccc81fd5c3a9c94c8a202f141e477b946a263b/benchmarks/src/main/proto/grpc/testing/control.proto#L16

as you can see, grpc-java uses the "canonical import path" (`grpc/testing/payloads.proto` "as in google3"), but C-core uses the tweaked one: `src/proto/grpc/testing/payloads.proto`.

jtattermusch (commented):
We're almost there but added a few more comments.

CMakeLists.txt
```diff
@@ -318,18 +318,103 @@ include(cmake/ssl.cmake)
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api
+    https://storage.googleapis.com/grpc-bazel-mirror/github.com/envoyproxy/data-plane-api/archive/9c42588c956220b48eb3099d186487c2f04d32ec.tar.gz
+    c5807010b67033330915ca5a20483e30538ae5e689aa14b3631d6284beca4630
+    data-plane-api-9c42588c956220b48eb3099d186487c2f04d32ec
+  )
+endif()
+if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api)
```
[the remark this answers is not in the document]
lidizheng: With `EXPECTED_HASH` field set in the `file(DOWNLOAD ...)` command, the CMake script will throw an error if the download failed in certain way. The behavior I observed for CMake 3.19.4 is that if the first download attempt failed, no folder will be created, the second download attempt (with different url) will start. However, the configuration phase will be stopped with message "-- Configuring incomplete, errors occurred!".

As a workaround, I added following log line to ask users to re-run the configuration stage:

```
  message(STATUS "Downloading from ${url}, if failed, please try configuring again")
```

The actual console output looks like:

```
$ cmake build
CMake Warning at third_party/abseil-cpp/CMakeLists.txt:74 (message):
  A future Abseil release will default ABSL_PROPAGATE_CXX_STD to ON for CMake
  3.8 and up.  We recommend enabling this option to ensure your project still
  builds correctly.


-- Failed to find LLVM FileCheck
-- git version: v1.6.0-6-g0baacde3 normalized to 1.6.0.6
-- Version: 1.6.0.6
-- Performing Test HAVE_THREAD_SAFETY_ATTRIBUTES -- failed to compile
-- Performing Test HAVE_STD_REGEX -- success
-- Performing Test HAVE_GNU_POSIX_REGEX -- failed to compile
-- Performing Test HAVE_POSIX_REGEX -- success
-- Performing Test HAVE_STEADY_CLOCK -- success
--
-- 3.19.4.0
CMake Deprecation Warning at third_party/zlib/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Downloading from https://storage.333googleapis.com/grpc-bazel-mirror/github.com/envoyproxy/data-plane-api/archive/9c42588c956220b48eb3099d186487c2f04d32ec.tar.gz, if failed, please try configuring again
CMake Error at cmake/download_archive.cmake:25 (file):
  file DOWNLOAD HASH mismatch

    for file: [/usr/local/google/home/lidiz/src/grpc/cmake/build/http_archives/data-plane-api-9c42588c956220b48eb3099d186487c2f04d32ec.tar.gz]
      expected hash: [c5807010b67033330915ca5a20483e30538ae5e689aa14b3631d6284beca4630]
        actual hash: [e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855]
             status: [6;"Couldn't resolve host name"]

Call Stack (most recent call first):
  CMakeLists.txt:326 (download_archive)


-- Downloading from https://github.com/envoyproxy/data-plane-api/archive/9c42588c956220b48eb3099d186487c2f04d32ec.tar.gz, if failed, please try configuring again
-- Configuring incomplete, errors occurred!
```

templates/CMakeLists.txt.template
lidizheng: New method renamed to `protobuf_generate_grpc_cpp_with_import_path_correction`, the old method will be a thin wrapper around the new method as suggested. Manually tested that the old method works (able to compile existing protos under src/proto).

For user's usage (https://github.com/search?q=protobuf_generate_grpc_cpp&type=code):

* gRPC as Submodule: https://github.com/IvanSafonov/grpc-cmake-example
* ProtoBuf provides a more generic command: https://www.f-ax.de/dev/2020/11/08/grpc-plugin-cmake-support.html
* Copied the implementation ([example](https://github.com/sandvikcode/pubsuber/blob/1083c20e71fc057accf1bc0e7fbb31008efaf885/cmake/grpc.cmake))

As discussed, we might want to continue our support for the submodule use case, so keeping the original signature might be worthwhile. I didn't find any usage of using this method from the CMake gRPC target (probably because we don't support this approach yet).

Given that the ProtoBuf repo did provide a plugin command, we can recommend users to use that instead of our CMake "internal" command.

---

For the import path tweak, I will work on that in follow-up PRs.

tools/buildgen/extract_metadata_from_bazel_xml.py
```diff
@@ -45,6 +45,48 @@
...
+        ExternalProtoLibrary(destination='third_party/envoy-api',
+                             proto_prefix='third_party/envoy-api/'),
+    'com_google_googleapis':
+        ExternalProtoLibrary(destination='third_party/googleapis',
+                             proto_prefix='third_party/googleapis/'),
+    'com_github_cncf_udpa':
+        ExternalProtoLibrary(destination='third_party/xds',
+                             proto_prefix='third_party/xds/'),
+    'opencensus_proto':
+        ExternalProtoLibrary(destination='third_party/opencensus-proto/src',
```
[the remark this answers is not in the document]
lidizheng: I see the potential strip path mismatch and version mismatch. As for current state of this PR's latest commit, they are in sync, both of them are using the release tag `v0.3.0` of `opencensus-proto`.

```
$ git submodule
 4aa53e15cbf1a47bc9087e6cfdca214c1eea4e89 third_party/opencensus-proto (v0.3.0)

        http_archive(
            name = "opencensus_proto",
            sha256 = "b7e13f0b4259e80c3070b583c2f39e53153085a6918718b1c710caf7037572b0",
            strip_prefix = "opencensus-proto-0.3.0/src",
            urls = [
                "https://storage.googleapis.com/grpc-bazel-mirror/github.com/census-instrumentation/opencensus-proto/archive/v0.3.0.tar.gz",
                "https://github.com/census-instrumentation/opencensus-proto/archive/v0.3.0.tar.gz",
            ],
        )
```

cmake/download_archive.cmake
lidizheng: As we discussed offline, there are rooms for improvement. We can migrate to the newer `FetchContent` command, but we will need to investigate it's behavior (whether it's in BUILD stage or CONFIGURE stage).

templates/CMakeLists.txt.template
```diff
@@ -382,40 +418,72 @@
...
+    set(_protobuf_include_path -I . -I <%text>${_gRPC_PROTOBUF_WELLKNOWN_INCLUDE_DIR}</%text>)
+    # compute the relative path
+    get_filename_component(ABS_FIL <%text>${_gRPC_PROTO_SRCS_DIR}/${IMPORT_PATH}</%text> ABSOLUTE)
+    get_filename_component(FIL_WE <%text>${_gRPC_PROTO_SRCS_DIR}/${IMPORT_PATH}</%text> NAME_WE)
+    file(RELATIVE_PATH REL_FIL <%text>${_gRPC_PROTO_SRCS_DIR}</%text> <%text>${ABS_FIL}</%text>)
+    get_filename_component(REL_DIR <%text>${REL_FIL}</%text> DIRECTORY)
+    set(RELFIL_WE "<%text>${REL_DIR}/${FIL_WE}</%text>")
+    # copy the proto file to a centralized location
+    file(MAKE_DIRECTORY <%text>${_gRPC_PROTO_SRCS_DIR}/${REL_DIR}</%text>)
+    file(COPY <%text>${CMAKE_CURRENT_SOURCE_DIR}/${FILE_LOCATION}</%text> DESTINATION <%text>${_gRPC_PROTO_SRCS_DIR}/${REL_DIR}</%text>)
```
jtattermusch, line 460: one more concern: will copying the .proto file from the original location to `${_gRPC_PROTO_SRCS_DIR}` break cmake's ability to perform incremental builds? (since if the copy happens always, the .proto file will always have timestamp updated and will look like a modified source file (which will then trigger unnecessary rebuild?).

It would be good to clarify what happens here, make sure that resulting behavior is what we want (incremental builds should still work) and add a comment that explains why that's the case.

jtattermusch: Ok, the documentation says this:

> The COPY signature copies files, directories, and symlinks to a destination folder. Relative input paths are evaluated with respect to the current source directory, and a relative destination is evaluated with respect to the current build directory. Copying preserves input file timestamps, and optimizes out a file if it exists at the destination with the same timestamp. Copying preserves input permissions unless explicit permissions or NO_SOURCE_PERMISSIONS are given (default is USE_SOURCE_PERMISSIONS).

so it looks like we should be fine - can you please check that the incremental builds still work, just to be sure.

jtattermusch (approved):
LGTM upon improving the readability  in the protobuf_generate_grpc_cpp_with_import_path_correction (see comments).

templates/CMakeLists.txt.template
```diff
@@ -382,40 +418,72 @@
...
+  #     More info can be found at https://github.com/grpc/grpc/pull/25272.
+  #
+  function(protobuf_generate_grpc_cpp_with_import_path_correction FILE_LOCATION IMPORT_PATH)
+    if(NOT FILE_LOCATION)
+      message(SEND_ERROR "Error: PROTOBUF_GENERATE_GRPC_CPP() called without any proto files")
+      return()
+    endif()
+
+    set(_protobuf_include_path -I . -I <%text>${_gRPC_PROTOBUF_WELLKNOWN_INCLUDE_DIR}</%text>)
+    # compute the relative path
```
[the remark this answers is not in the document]
lidizheng: Added comments for every path computation (acronyms are terrible, I needed to re-learn CMake). Each comment sections is attached with an example `health.proto`.

It's one of the tweaked proto we have, it's currently located at `src/proto/grpc/health/v1/health.proto`, but the expected location is `grpc/health/v1/health.proto`.

templates/CMakeLists.txt.template
```diff
@@ -382,40 +418,72 @@
...
+    get_filename_component(ABS_FIL <%text>${_gRPC_PROTO_SRCS_DIR}/${IMPORT_PATH}</%text> ABSOLUTE)
+    get_filename_component(FIL_WE <%text>${_gRPC_PROTO_SRCS_DIR}/${IMPORT_PATH}</%text> NAME_WE)
```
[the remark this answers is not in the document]
lidizheng: Comments added.

templates/CMakeLists.txt.template
lidizheng: Checked with target `grpc`. If I modified one of the source files, the incremental works as expected.

lidizheng:
None of the C#, PHP and macOS failures are related to this PR. Merging.

PR #29271, opened 2022-03-31
xds_end2end_test: make each individual test start the number of backends it needs

markdroth:
I also cleaned up the interface to avoid passing parameters into the base class.

This is a first step toward splitting the file up into multiple pieces.


yashykt (approved):
Thanks for doing this!

PR #29277, opened 2022-03-31
Exclude venvs from formatting

gnossen:
I often find that `yapf` and `isort` take a *long* time to complete. It seems this is due to them formatting all of the virtual environments I have scattered throughout my source tree.


lidizheng (approved):
Thanks for fixing this, nice QoL improvment!

PR #29285, opened 2022-04-01
xds_end2end_test: Move CSDS tests to their own file

markdroth:


test/cpp/end2end/xds/BUILD
```diff
@@ -140,6 +139,32 @@ grpc_cc_test(
     ],
 )
 
+grpc_cc_test(
+    name = "xds_csds_end2end_test",
```
lidizheng, line 143: optional: we might be able to drop the `xds_` prefix for tests, since the directory already include this information.

lidizheng (approved):
Thanks for working on this

test/cpp/end2end/xds/BUILD
markdroth: I think it's still useful to have the prefix, since there may be cases where the filename is used without the complete path.

PR #29292, opened 2022-04-01
Add resource quota reclamation to tcp endpoint

ananda1066:


src/core/lib/iomgr/tcp_posix.cc
```diff
@@ -661,7 +664,30 @@ static void tcp_destroy(grpc_endpoint* ep) {
...
 
-static void tcp_trace_read(grpc_tcp* tcp, grpc_error_handle error) {
+static void maybe_post_reclaimer(grpc_tcp* tcp)
+    ABSL_EXCLUSIVE_LOCKS_REQUIRED(tcp->read_mu) {
+  if (!tcp->has_posted_reclaimer) {
+    tcp->has_posted_reclaimer = true;
+    tcp->memory_owner.PostReclaimer(
+        grpc_core::ReclamationPass::kBenign,
+        [tcp](absl::optional<grpc_core::ReclamationSweep> sweep) {
+          if (sweep.has_value()) {
```
ctiller, line 674: It might be worth reducing the hadoken factor here... some suggestions:
1. `if (!sweep.has_value()) return;`
2. moving the body of this into a perform_reclaimation() function that we call from the lambda

ctiller (approved):
This looks good... one style nit to consider.

I wondered about using `MutexLock`, `ReleasableMutexLock` to avoid leaked `Unlock` problems on the mutexes, but I think it'd end up overall less clear.

src/core/lib/iomgr/tcp_posix.cc
ananda1066: Done

PR #28666, opened 2022-01-24
[Ruby]fix a rake compile error since rake-compiler ~> 1.1

HannahShiSFB:
Grpc Rakefile has a cross_compiling callback (https://github.com/grpc/grpc/blob/master/Rakefile#L36) which replace the Gem `spec.files` with some files under src/ruby/xxx.
After https://github.com/rake-compiler/rake-compiler/pull/191/files in rake-compiler 1.1.2, grpc_c.so files stopped being installed to the grpc source directory (src/ruby/lib/grpc/<x.y>/)
As a result, `spec.files` will no longer include the extension binary files, therefore not the built gem file.

Instead of replacing the whole `spec.files` (which build in the gemspec file and updated during rake compile), this PR filter from it for the wanted files to be packaged (i.e. exclude grpc core files, etc.), so no such files will be left behind.

grpc.gemspec is also updated so that pb test files are classified correctly.

grpc.rb is changed because although rake-compiler stopped installing grpc_c.so files, the versioned directories are still being created, which results in the directory test success but no so file to load.

@stanley-cheung 
@apolcyn

PR #28590, opened 2022-01-17
Macos build artifacts speedup (by removing ruby artifact build)

jtattermusch:
As per offline discussion - building ruby gem on macos seems redundant and bring unnecessary complexity and slows down the build. The gems are already built using rake-compiler-dock on linux.

PR #14634, opened 2018-03-08
Remove source code and native-to-build-host grpc_c.so from ruby linux binary packages

apolcyn:
Fixes https://github.com/google/protobuf/issues/4210 for the gRPC ruby package (the google-protobuf package should need a similar fix). Prompted by the suggestion in https://github.com/rake-compiler/rake-compiler/issues/146

Running an artifact build in https://grpc-testing.appspot.com/job/gRPC_build_artifacts/2059/

----------------------

PTAL I think this is ready for a look.

The artifact build on jenkins has been timing out and I haven't been able to get a build on jenkins, but I built packages from this PR locally and did local testing. I also ran the ruby distrib tests locally (except for the ones based off of ubuntu images).

For comparison, here is the content of the latest (1.10) `x86_64-linux` gem:
* https://gist.github.com/apolcyn/71024dcc99b52dd3b7fdee763e69dd96

and here is the content of the `x86_64-linux` 
 gem built from this PR:
* https://gist.github.com/apolcyn/806e64e7d6c9d458fb8cc2c086e3584d

... with the source code and `src/ruby/lib/grpc/grpc_c.so` now removed.

PR #29304, opened 2022-04-04
Upgrade ruby rake-compiler-dock images

apolcyn:
In particular, this picks up the fix in https://github.com/grpc/grpc/issues/28271#issuecomment-986300461


apolcyn:
Ruby distrib tests are currently failing with:

```
A grpc_c.so file in this binary gRPC package is dynamically linked to libruby.
+ [[ 1 != 0 ]]
+ echo 'A grpc_c.so file in this binary gRPC package is dynamically linked to libruby.'
++ find /usr/local/rvm/gems/ruby-2.7.0 -name grpc_c.so -exec ldd '{}' ';'
++ grep -c 'not found'
+ DEPENDENCY_NOT_FOUND=1
+ [[ 1 != 0 ]]
+ echo 'A grpc_c.so file in this binary gRPC package has an non-portable dependency.'
A grpc_c.so file in this binary gRPC package has an non-portable dependency.
```

johnnyshields:
May we get this into the 1.46.0 release?

apolcyn:
Sorry for delay, I would like to get this change in but have not had time to solve the issue found in the test.

The problem that the test found is:

```
$ ldd grpc-1.46.0.dev-x86_64-linux/src/ruby/lib/grpc/grpc_c.so
	linux-vdso.so.1 (0x00007fffed1e6000)
	libruby.so.3.1 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fc712979000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fc71279f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fc7131c8000)
```

So the new packages have an accidentally-built binary that's native to the build host, rather than cross-compiled. This undoes the change in https://github.com/grpc/grpc/pull/14634, which I'd prefer to avoid.

Need to figure out how to fix that, first.

apolcyn:
Very likely we need to get https://github.com/grpc/grpc/pull/28666 in first

apolcyn:
Temporarily cherry-picked https://github.com/grpc/grpc/pull/28666 into this branch to see if it fixes things

apolcyn:
The latest packages still have the non-version-specific library, but it no longer has the `libruby` dependency:

```
$ find grpc-1.47.0.dev-x86_64-linux/ -name grpc_c.so
grpc-1.47.0.dev-x86_64-linux/src/ruby/lib/grpc/2.6/grpc_c.so
grpc-1.47.0.dev-x86_64-linux/src/ruby/lib/grpc/2.7/grpc_c.so
grpc-1.47.0.dev-x86_64-linux/src/ruby/lib/grpc/grpc_c.so
grpc-1.47.0.dev-x86_64-linux/src/ruby/lib/grpc/3.1/grpc_c.so
grpc-1.47.0.dev-x86_64-linux/src/ruby/lib/grpc/2.5/grpc_c.so
grpc-1.47.0.dev-x86_64-linux/src/ruby/lib/grpc/3.0/grpc_c.so
$ ldd grpc-1.47.0.dev-x86_64-linux/src/ruby/lib/grpc/grpc_c.so
	linux-vdso.so.1 (0x00007ffc59776000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f19e6e68000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f19e6c8e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f19e76c4000)
```

jtattermusch:
+1 to building the ruby mac artifacts on linux instead of on macos (where it's slow and requires hacks). If that's what you end up doing, we should consider doing more cleanup (see my older PR https://github.com/grpc/grpc/pull/28590/files, e.g. removing the tools/distrib/build_ruby_environment_macos.sh prepare script.)

tools/internal_ci/macos/grpc_distribtests_ruby.sh
```diff
@@ -15,42 +15,4 @@
...
-# TODO(jtattermusch): Here we would normally run ruby macos distribtests, but currently no such tests are defined
-# in distribtest_targets.py
-
-tools/internal_ci/helper_scripts/store_artifacts_from_moved_src_tree.sh
-
-if [ "$FAILED" != "" ]
-then
-  exit 1
-fi
+echo "TODO(apolcyn): remove ruby macos distrib jobs, they don't test anything"
```
jtattermusch, line 18: +1 for keeping the job for a while as a no-op. Might come handy if we need to revert and it avoids making this PR too complex.

jtattermusch (approved):
LGTM as long as you've made sure that the ruby gems work well after this change is applied.
Can you summarize the test you've run and the results before merging?
This is especially important for the darwin gems, since this PR removes the possibility to build them on mac (and the rake-compiler-dock will be used from now on).

apolcyn:
> Can you summarize the test you've run and the results before merging?
This is especially important for the darwin gems, since this PR removes the possibility to build them on mac (and the rake-compiler-dock will be used from now on).

This change shouldn't affect most packages.

I manually tested as follows:

- Running the [greeter example](https://github.com/grpc/grpc/tree/master/examples/ruby) with the linux-x86_64 packages on ruby 2.5, 3.0, and 3.1
- Ran `ldd` on all `grpc_c.so` files in the linux x86_64 package and saw there are no new or missing dependencies
- I spot checked that the ruby source-only packages are installable (on ruby 3.1).
- Running the [greeter example](https://github.com/grpc/grpc/tree/master/examples/ruby) with the `darwin-x86_64` packages on ruby 2.5, 3.0, 3.1
- Compared the output of `otool -L grpc-1.45.0-universal-darwin/src/ruby/lib/grpc/3.0/grpc_c.bundle` (for a previous macos package) with `otool -L grpc-1.47.0.dev-x86_64-darwin/src/ruby/lib/grpc/3.0/grpc_c.bundle` (a new macos package), and saw that there are no new dependencies

apolcyn:
linux build only looks like a testing infra error:

```
kokoro_log_reader: Exhausted retries waiting for bash to return an integer exit code, returning exit code 137 instead.
```

Macos failures are due to failures of `HttpRequestTest.CancelGetDuringDNSResolution`, which is a known issue fixed by https://github.com/grpc/grpc/pull/29634

PR #29311, opened 2022-04-05
Wrap grpc-generated code with port_def/port_undef protobuf headers

smertnik3sh:
Some windows-specific macros looks like common method names. For example macro GetObject which defined in wingdi.h (transitively included in windows.h):

```#ifdef UNICODE
#define GetObject  GetObjectW
#else
#define GetObject  GetObjectA
#endif
```

This can cause unexpected problems when compiling "GetObject"-containg services for windows. In proto messages generated code port_def.inc/port_undef.inc include solves this problem. I suppose in grpc-generated code this wrap can be used too.

@veblush


georgthegreat:
@veblush, could you, please, take a look?

veblush:
I don't understand what problem you want to solve with this PR? Would you elaborate it more?

veblush:
Closing this since I didn't get information on this change but please feel free to reopen it if this is required.

PR #29313, opened 2022-04-05
weighted_target and xds_cluster_manager: don't update picker while update is in flight

markdroth:
This is the same change that was made for the priority policy in #29252.  It's not required in these other policies for correctness reasons, but it is a useful optimization that avoids unnecessary picker churn and makes it easier to understand what's happening when reading trace logs.  In general, we want to do this for any LB policy that has multiple active children.


markdroth:
Looks like this is actually needed to fix a bug.  See internal b/223861560.

markdroth:
I still need to add a test here...

markdroth:
The test has been added.

veblush (approved):

PR #29315, opened 2022-04-05
Correctly handle shebangs when fixing copyright

gnossen:
Previously, when running `tools/distrib/check_copyright.py --fix`, shebangs (e.g. `#!/bin/bash`) would end up placed _after_ copyright notice comment blocks. This PR ensures that shebangs remain at the beginning of files when running this script.


jtattermusch (approved):
LGTM assuming you've tested this.

PR #29319, opened 2022-04-05
Use "main" as the protobuf head branch name

lidizheng:
As title.

https://fusion2.corp.google.com/invocations/6f897875-3355-4b62-a4af-43e3b2a6ab36/targets

### DIFF
--- a/tools/internal_ci/linux/grpc_build_protobuf_at_head.cfg
+++ b/tools/internal_ci/linux/grpc_build_protobuf_at_head.cfg
@@ -29,3 +29,8 @@ env_vars {
   key: "RUN_TESTS_FLAGS"
   value: "protobuf"
 }
+
+env_vars {
+  key: "BAZEL_FLAGS"
+  value: "main"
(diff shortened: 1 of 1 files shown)


jtattermusch:
The error link if wrong (point to an unrelated build?)
I think you meant this link:
https://source.cloud.google.com/results/invocations/630da787-1f66-479d-9a6d-d0bca653601b/log (which seem to be the first failure of this kind for the job).

```
+ git fetch origin
+ git checkout origin/master
error: pathspec 'origin/master' did 
```

context: Looks like protobuf has recently renamed their `master` branch to `main`

jtattermusch (approved):
LGTM, please verify that this unbreaks the grpc_build_protobuf_at_head job (with an adhoc build) before merging.

gnossen (approved):

lidizheng:
Merging. There are other build problems toward ProtoBuf at head with Python. I will try to dig into that.

PR #13494, opened 2017-11-23
C++-ize backoff

dgquintas:
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/13494)

PR #29320, opened 2022-04-06
client_lb_end2end_test: refactor connection delay injection into its own library

markdroth:
Also change the delay to work asynchronously, via a timer, rather than blocking a thread.

This is in preparation for using this code in other tests.

### DIFF
--- a/test/cpp/end2end/connection_delay_injector.cc
+++ b/test/cpp/end2end/connection_delay_injector.cc
@@ -22,8 +22,8 @@
 
 #include "src/core/lib/channel/channel_args.h"
 #include "src/core/lib/gprpp/time.h"
-#include "src/core/lib/iomgr/timer.h"
 #include "src/core/lib/iomgr/tcp_client.h"
+#include "src/core/lib/iomgr/timer.h"
 
 // defined in tcp_client.cc
 extern grpc_tcp_client_vtable* grpc_tcp_client_impl;
--- a/test/cpp/end2end/connection_delay_injector.cc
+++ b/test/cpp/end2end/connection_delay_injector.cc
@@ -1,0 +1,116 @@
+// Copyright 2016 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


test/cpp/end2end/connection_delay_injector.cc
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  ~InjectedDelay() { grpc_channel_args_destroy(channel_args_); }
+
+ private:
+  static void TimerCallback(void* arg, grpc_error_handle /*error*/) {
+    auto* self = static_cast<InjectedDelay*>(arg);
+    g_original_vtable->connect(self->closure_, self->endpoint_,
+                               self->interested_parties_, self->channel_args_,
+                               &self->address_, self->deadline_);
```
apolcyn, line 67: Can we subtract the initial delay from the delay used here? Thinking without that, this may cause surprising behavior

apolcyn (approved):

test/cpp/end2end/connection_delay_injector.cc
markdroth: We account for that on line 55 above.  This is basically what the original code in client_lb_end2end_test does.

apolcyn: For least surprising behavior, though, shouldn't line 55 really be removed?

Also shouldn't `duration` be set to something like `max(duration, deadline - Now())` ?

Otherwise, if subchannel connection timeout is e.g. 20 seconds, and we inject a 30 second delay, the subchannel won't actually receive it's timeout until 30 seconds later?

markdroth: Good point.  I had thought there must have been some reason the existing code worked this way, but I traced it back to #13494, and I don't see anything in that (long) review that explains why it's done this way.  And the tests do seem to pass without line 55.  I'll remove that line in a subsequent PR.

I don't think we need to subtract from deadline, because the deadline is a timestamp, not a duration.  If the delay is longer than `deadline - Now()`, then deadline will be in the past when the connection attempt starts, in which case it will fail immediately.

markdroth: Oh wait, sorry, you were saying the opposite: you were saying we should cap the delay by the deadline.  Yeah, that's probably a good idea.  I'll make that change as well.

markdroth: Sent you #29349 with this change.

Issue #27932, opened 2021-11-03
Artifact Build Mac OS Failure


gnossen:
```
kokoro_log_reader: Exhausted retries waiting for bash to return an integer exit code, returning exit code 137 instead. This is not a Kokoro error! This usually means: 1) Your build was killed by something in the VM 2) Exited normally, but that the parent shell of the nohup'd build process is IO blocked and could not write out the exit_code_file in 60 seconds. In this case be sure that all processes launched by your build are properly cleaned up before exiting to free up resources and reduce IO contention, or 3) There was a BSOD, the VM rebooted, the builder has reconnected but the process is of course gone.


[ID: 2961102] Build finished after 1739 secs, exit value: 137
```

https://source.cloud.google.com/results/invocations/094c5a19-0b34-4b73-bd5c-be08d23cba67/log

PR #29322, opened 2022-04-06
Expose NoOpCertificateVerifier to C++

ZhenLian:
This is originally raised in an internal bug.
The original idea for not having NoOpCertificateVerifier is to raise security bars. NoOpCertificateVerifier can expose security issues if not properly used. So we tried to "hide" it and hope that will help us limit its usage.
As Advanced TLS becomes popular, there are more usages of  NoOpCertificateVerifier internally, and their reasons for using it are all valid. In that case, we can expose this in C++, with better documentations, so that users would get a sense of what are implied when using this class.
I will clean-up all the internal sites as long as this goes in.

### DIFF
--- a/include/grpcpp/security/tls_certificate_verifier.h
+++ b/include/grpcpp/security/tls_certificate_verifier.h
@@ -214,9 +214,30 @@ class ExternalCertificateVerifier {
       request_map_ ABSL_GUARDED_BY(mu_);
 };
 
+// A CertificateVerifier that will perform hostname verification, to see if the
+// target name set from the client side matches the identity information
+// specified on the server's certificate.
 class HostNameCertificateVerifier : public CertificateVerifier {
  public:
   HostNameCertificateVerifier();
+};
+
+// A sample ExternalCertificateVerifier(which is generally implemented by users)
+// that doesn't perform any additional checks other than certificate
+// verification, if specified.
+// Note: using this solely without any other authentication mechanisms on the
+// peer identity will leave your applications to the MITM(Man-In-The-Middle)
+// attacks. Users should avoid doing so in production environments.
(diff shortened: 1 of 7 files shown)


include/grpcpp/security/tls_certificate_verifier.h
```diff
@@ -214,11 +214,32 @@ class ExternalCertificateVerifier {
...
   HostNameCertificateVerifier();
 };
 
+// A sample ExternalCertificateVerifier(which is generally implemented by users)
+// that doesn't perform any additional checks other than certificate
+// verification, if specified.
+// Note: using this solely without any other authentication mechanisms on the
+// peer identity will leave your applications to the MITM(Man-In-The-Middle)
+// attacks. Users should avoid doing so in production environments.
+class NoOpCertificateVerifier : public ExternalCertificateVerifier {
```
markdroth, line 231: Why make this an external verifier?  If we're providing this out of the box, let's just implement it inside core, the same way that we do for the hostname verifier, so that we don't have an extra and unnecessary level of indirection.

markdroth (approved):
Looks great!

ZhenLian:
All failures seem irrelevant(seems something wrong with the MacOS test infra):
Artifact Build MacOS - https://github.com/grpc/grpc/issues/27932 
Bazel C/C++ Dbg MacOS - an irrelevant `h2_sockpair_test` failed
Distribution Tests Ruby MacOS - https://github.com/grpc/grpc/issues/29487

I am going to merge now. 

---------------------------------------------------------------------------------

A note for the importer:
This PR is unlikely to get rolled back. But just in case it breaks the import: feel free to roll it back, and I will work on it once I am back(I am out until next Thursday).

PR #29169, opened 2022-03-21
Support unix socket in grpc_sockaddr_to_string

yousukseung:
Also adds an optional flag to indicate an explicit error.

@donnadionne

PR #29323, opened 2022-04-06
Fix grpc_sockaddr_to_uri to handle embedded nulls for unix_abstract resolved addresses

anicr7:
Currently when an embedded null is present, it is left as is. This causes an issue when grpc_sockaddr_to_uri is followed by any c style operations like copy, as the string is truncated at the non-encoded null character. For example this is triggered when channel args containing a string channel arg is copied.

To prevent this properly encode the URI with %00.

@markdroth


src/core/lib/address_utils/sockaddr_utils.cc
```diff
@@ -51,11 +52,13 @@ static std::string grpc_sockaddr_to_uri_unix_if_possible(
   }
   const auto* unix_addr = reinterpret_cast<const struct sockaddr_un*>(addr);
   if (unix_addr->sun_path[0] == '\0' && unix_addr->sun_path[1] != '\0') {
```
markdroth, line 54: A better option here would be to use our [URI library](https://github.com/grpc/grpc/blob/master/src/core/lib/uri/uri_parser.h), which will also handle percent-encoding any other special characters that may need it:

```
std::string scheme;
std::string path;
if (unix_addr->sun_path[0] == '\0' && unix_addr->sun_path[1] != '\0') {
  scheme = "unix-abstract";
  path = std::string(unix_addr->sun_path + 1, resolved_addr->len - sizeof(unix_addr->sun_family) - 1);
} else {
  scheme = "unix";
  path = unix_addr->sun_path;
}
absl::StatusOr<URI> uri = URI::Create(
    std::move(scheme), /*authority=*/"", std::move(path),
    /*query_parameter_pairs=*/{}, /*fragment=*/"");
if (!uri.ok()) return "";
return uri->ToString();
```

We should probably do the same thing in `grpc_sockaddr_to_uri()`.

Feel free to make this a TODO assigned to me if you don't want to tackle this now.

test/core/address_utils/sockaddr_utils_test.cc
```diff
@@ -206,6 +209,33 @@ TEST(SockAddrUtilsTest, SockAddrToString) {
   EXPECT_TRUE(grpc_sockaddr_to_uri(&phony).empty());
 }
 
+#ifdef GRPC_HAVE_UNIX_SOCKET
+
+grpc_resolved_address MakeAddrUnix(std::string path) {
```
markdroth, line 214: Instead of writing your own function to do this, please use [`UnixSockaddrPopulate()`](https://github.com/grpc/grpc/blob/ab2193d8c178c41d2b097dc8b35a54bfdd9e928b/src/core/lib/address_utils/parse_address.h#L72) and [`UnixAbstractSockaddrPopulate()`](https://github.com/grpc/grpc/blob/ab2193d8c178c41d2b097dc8b35a54bfdd9e928b/src/core/lib/address_utils/parse_address.h#L77).

markdroth (commented):
Thanks for the fix!

Note that #29169 is touching some of this same code, so either you or @yousukseung will need to merge the changes.

test/core/address_utils/sockaddr_utils_test.cc
anicr7: Done

src/core/lib/address_utils/sockaddr_utils.cc
anicr7: Moved the logic in grpc_sockaddr_to_uri_unix_if_possible to use the URI library.

For grpc_sockaddr_to_uri, migrating to URI library causes some issues because of the square brackets around the IPV6 addresses. 

The URI library encodes the "[" with %5B  and "]" with %5D. This seems like a bug in URI parser library as brackets around host should be fine. WDYT?

markdroth: No, the URI parser implementation is correct.  `[` and `]` are legal characters in the authority part of a URI (see [RFC-3986 section 3.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2), but not in the path part ([RFC-3986 section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3), and here we are encoding the address in the path.

I think we should fix callers to deal with this properly.  They can parse the URI string using `URI::Parse()` and then access the percent-decoded address via `uri->path()`.

But as I said before, feel free to leave this as a TODO for now if you don't want to fix all the callers.

src/core/lib/address_utils/sockaddr_utils.h
```diff
@@ -24,6 +24,7 @@
 #include <string>
 
 #include "src/core/lib/iomgr/resolved_address.h"
+#include "src/core/lib/uri/uri_parser.h"
```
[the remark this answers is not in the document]
anicr7: Moved, done.

src/core/lib/address_utils/sockaddr_utils.cc
anicr7: Created a new function grpc_sockaddr_to_uri_decoded(maybe there is a better name?), and added the logic to perform URI::Parse() followed by constructing the path using uri->scheme and uri->path.

Let me know if this looks better.

markdroth: I'd really rather not have two different methods to maintain here.  The API really makes no sense, and this just prolongs the agony of having this broken.

How many callers would have to be updated if we just fixed the behavior of the existing `grpc_sockaddr_to_uri()` function, and how hard would that be to do?  Remember that there may be some internal callers, so you'll have to check more than just the OSS code.

If this isn't hard, let's just fix it.  If it is too hard to do right now, then let's just leave `grpc_sockaddr_to_uri()` the way it is now, but with a TODO indicating that it needs to be fixed, since I don't think fixing this is a blocker for your other PR.

anicr7: The latest changes does fix grpc_sockaddr_to_uri. I created grpc_sockaddr_to_uri_decoded to minimize the repetition across the callers which do want to use grpc_sockaddr_to_uri but want the previous behavior. Internally grpc_sockaddr_to_uri_decoded parses the URI string using URI:Parse and gets the path. 

This function was created because there are ~20 callers on the OSS code which used grpc_sockaddr_to_uri incorrectly and I didn't want to repeat the URI::Parse + extracting path everywhere. Do you recommend I remove `grpc_sockaddr_to_uri_decoded` function and use grpc_sockaddr_to_uri directly? I can repeat the logic( URI:Parse/get path) in the calling site instead.

markdroth: When you say "used grpc_sockaddr_to_uri incorrectly", what do you mean?  If we just return the properly encoded URI string to those callers, what would break?  Most of the callers look like they're just using the URI string for cosmetic purposes, and I'm fine with that changing to use proper encoding.

My argument here is that the thing we're returning is a URI string, so it needs to be a valid URI string.  There is no such thing as a "decoded URI string" -- that's not how URI processing works.  A URI exists either in string form with proper encoding, or in parsed form where the strings are decoded.  If you convert to string form without proper encoding, then what you have is an invalid URI string.  I don't think that's something we should be providing an API for.  And I don't think it makes sense for the callers to do this decoding themselves either, because they should not be using this invalid representation in the first place.

anicr7: Mark, if we are ok with the logging being the actual URI string we actually don't need any changes on the callers.

The only changes needed are in couple of unit tests where the non-encoded URI was expected. I changed the expectation to URI format in these tests, rather than extract the non-URI string. Let me know if this looks better.

markdroth (approved):
Looks great!

markdroth:
Before merging, please test to see if there will be any internal changes needed for this PR.  If there are, we should do this as a cherry-pick import.

anicr7:
> Before merging, please test to see if there will be any internal changes needed for this PR. If there are, we should do this as a cherry-pick import.

Looks like there are multiple internal changes required for this PR. I am going to work on the internal changes, but to unblock the TCP connect handshaker I am reverting the changes to grpc_sockaddr_uri and keeping only the changes needed for UDS.

I'll create a separate PR once I have the internal changes ready.

ctiller:
Going to just suggest adding a fuzzer that checks that this function either
returns an error or a string that our uri parser accepts as valid.

src/core/lib/address_utils/sockaddr_utils.cc
```diff
@@ -236,6 +244,7 @@ std::string grpc_sockaddr_to_uri(const grpc_resolved_address* resolved_addr) {
   if (scheme == nullptr || strcmp("unix", scheme) == 0) {
     return grpc_sockaddr_to_uri_unix_if_possible(resolved_addr);
   }
+  // TODO(anramach): Encode the string using URI::Parse before returning.
```
markdroth, line 247: I think you mean `URI::Create()` and `URI::ToString()`, not `URI::Parse()`.

markdroth (approved):
This looks good!

I like Craig's suggestion about having a fuzzer.  But since you're going to be doing follow-up work for `grpc_sockaddr_to_uri()`, you can write the fuzzer as part of that PR if you'd like.

src/core/lib/address_utils/sockaddr_utils.cc
anicr7: Yes, thanks for catching that. Fixed!

PR #29326, opened 2022-04-06
fix security connector segfault

dentiny:
Passing `nullptr` to `absl::string_view` (https://github.com/grpc/grpc/blob/c0e2fe9ea040b6e1e901b540eb29f318d9c1c541/src/core/lib/security/security_connector/security_connector.cc#L48) is not allowed and would lead to segfault.

@veblush


dentiny:
I don't think I have the permission to add a label :(

Mrmaxmeier:
As far as I can tell, this change was merged via / included in https://github.com/grpc/grpc/pull/29346

dentiny:
The fix has been merged in another PR.

Issue #26857, opened 2021-08-02
Cross-compiled gRPC is missing binaries for the target


xhochy:
This is a follow-up issue to the PR https://github.com/grpc/grpc/pull/26148 that introduced the problem. I could just send a PR that reverts it but that reintroduces the problem again that led to that PR. I sadly don't know the intention of their PR as we have been cross-compiling gRPC successfully since at least 1.32.

### What version of gRPC and what language are you using?

1.39

### What operating system (Linux, Windows,...) and version?

Linux and macOS

### What runtime / compiler are you using (e.g. python version or version of gcc)

Various; this issue is independent of that.

### What did you do?

We are building gRPC through cross-compilation for e.g. Apple M1 or Linux aarch64 and providing the output as conda packages that are intended to be used on these architectures. There you will need a native version of `grpc_cpp_plugin` to natively build packages that depend on gRPC.

### What did you expect to see?

`$PREFIX/bin/grpc_cpp_plugin` should be built and installed into the target prefix.

### What did you see instead?

The above-mentioned plugin was neither built nor installed. This makes sense if the build of gRPC would only be used in a cross-compiling environment but the output of the mentioned build here will be used for cross-compiling as well as compiling natively on the target system later.

### Anything else we should know about your project / environment?

We will revert the PR https://github.com/grpc/grpc/pull/26148 in conda-forge for now for the 1.39 builds to get the users of the cross-compiled package back to a working state.

veblush:
@xhochy Our of curiosity, why do you use `install` for building packages? I guess it might be for getting built artifacts. Is there an alternative way for getting those?

@UVV-gh Could you explain how installing cross-compiled protoc was a problem for your case?

UVV-gh:
It seems like @xhochy has a different use case, he actually uses cross-compiled gRPC plugins on the target. Yocto-based generated SDK is used on the _host_ for cross-compilation for the target. SDK consists of two sysroots, one for the target and one for the host, the latter would contain gRPC plugins, which would have to be searched manually and substituted to CMake.

clementperon:
There is a pull request to have a proper cross compilation support between gRPC libs and gRPC binaries: 
https://github.com/grpc/grpc/pull/22498

I add it to my Yocto and I have now a proper SDK and a proper cross compilation.

xhochy:
> @xhochy Our of curiosity, why do you use `install` for building packages? I guess it might be for getting built artifacts. Is there an alternative way for getting those?

[...]

PR #26148, opened 2021-05-02
cmake: don't export grpc_cpp_plugin when crosscompiling

UVV-gh:
Plugin should not be exported for a cross-compilation environment,
because it has to run on a host/build system. If it's exported then
configure stage fails with an error, that imported target references
non-existing file. Even if the file is found, it can't be used on a host.

Signed-off-by: Vyacheslav Yurkov <uvv.mail@gmail.com>

@drfloob

PR #29328, opened 2022-04-06
cmake: add separate export for plugin targets

herbrechtsmeier:
Add a separate CMake export for the plugin targets to separate binaries
and libraries into their own CMake target export files. Skip the cross
compiled binary plugin targets during cross compile because they are not
usable and not always available.

The Yocto build system doesn't install cross compiled binaries into the
target sysroot. This makes the CMake gRPC config useless as it checks
the existent of binaries and fails without the binaries.

This is a strip down version of #22498 and related to #26148 and #26857.

@veblush @jtattermusch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10936,7 +10936,7 @@ target_link_libraries(grpc_cpp_plugin
 
 
 if(gRPC_INSTALL)
-  install(TARGETS grpc_cpp_plugin EXPORT gRPCTargets
+  install(TARGETS grpc_cpp_plugin EXPORT gRPCPluginTargets
     RUNTIME DESTINATION ${gRPC_INSTALL_BINDIR}
     BUNDLE DESTINATION  ${gRPC_INSTALL_BINDIR}
     LIBRARY DESTINATION ${gRPC_INSTALL_LIBDIR}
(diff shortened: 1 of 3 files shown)


clementperon:
Hi, 

I tested this patch with host compilation (compiler and libs for x86_64) and cross compilation (compiler for x86_64 and libs for ARM64) on top of gRPC tag "1.45.1".

All of them works properly.

CMake used:
```
# Find Protobuf installation
set(protobuf_MODULE_COMPATIBLE TRUE)
find_package(Protobuf REQUIRED)
message(STATUS "Using protobuf ${Protobuf_VERSION}")
set(_PROTOBUF_LIBPROTOBUF protobuf::libprotobuf)

# Find gRPC installation Looks for gRPCConfig.cmake file installed by gRPC's
# cmake installation.
find_package(gRPC CONFIG REQUIRED)
message(STATUS "Using gRPC ${gRPC_VERSION}")
set(_REFLECTION gRPC::grpc++_reflection)
set(_GRPC_GRPCPP gRPC::grpc++)

# Get full path of protobuf/gRPC compiler
set(_PROTOBUF_PROTOC $<TARGET_FILE:protobuf::protoc>)
if(CMAKE_CROSSCOMPILING)
        find_program(_GRPC_CPP_PLUGIN_EXECUTABLE grpc_cpp_plugin)
else()
        set(_GRPC_CPP_PLUGIN_EXECUTABLE $<TARGET_FILE:gRPC::grpc_cpp_plugin>)
endif()
```

Tested-by "Clément Péron" <peron.clem@gmail.com>

PR #29330, opened 2022-04-06
unbreak proto regenerate scripts for C#, ruby, php by allowing protobuf well known types

jtattermusch:
Running the regenerate scripts currently gives an error:

```
src/csharp/generate_proto_csharp.sh 
google/protobuf/timestamp.proto: File not found.
src/proto/grpc/testing/control.proto:19:1: Import "google/protobuf/timestamp.proto" was not found or had errors.
src/proto/grpc/testing/control.proto:272:3: "google.protobuf.Timestamp" is not defined.
src/proto/grpc/testing/control.proto:273:3: "google.protobuf.Timestamp" is not defined
```

which consequently breaks the process for updating third_party/protobuf submodule.

The problem was introduced by this change:
https://github.com/grpc/grpc/commit/d2aafef157c3d2eb62b26c17af049762e9e596d1#diff-2202fdf8e6917d5dd7e99d5b254a419115fd79780dc18809a5b0a0d144cfc020R19

### DIFF
--- a/src/csharp/Grpc.IntegrationTesting/Control.cs
+++ b/src/csharp/Grpc.IntegrationTesting/Control.cs
@@ -26,86 +26,89 @@ namespace Grpc.Testing {
           string.Concat(
             "CiRzcmMvcHJvdG8vZ3JwYy90ZXN0aW5nL2NvbnRyb2wucHJvdG8SDGdycGMu",
             "dGVzdGluZxolc3JjL3Byb3RvL2dycGMvdGVzdGluZy9wYXlsb2Fkcy5wcm90",
-            "bxoic3JjL3Byb3RvL2dycGMvdGVzdGluZy9zdGF0cy5wcm90byIlCg1Qb2lz",
-            "c29uUGFyYW1zEhQKDG9mZmVyZWRfbG9hZBgBIAEoASISChBDbG9zZWRMb29w",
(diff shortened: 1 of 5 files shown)


wanlin31:
Sorry that PR break the process for updating third_party/protobuf submodule. Can I ask what is the best practice to avoid this from happening again? Any test I should specifically look at? Thank you.

wanlin31 (approved):

paulosjca (approved):

gnossen (approved):

jtattermusch:
> Sorry that PR break the process for updating third_party/protobuf submodule. Can I ask what is the best practice to avoid this from happening again? Any test I should specifically look at? Thank you.

This one would have been hard to predict, so don't worry about this too much:
- Ideally we'd have a sanity test on presubmit to check whether the .proto regeneration scripts still work (which would have caught this), but we currently don't have such test
- the only alternative remedy would be to get more folks involved in the .proto file changes - since the .proto files represent our data model / API contract, it's generally good to make such changes visible - with more people involved, someone might then notice the change could cause this unexpected issue.

jtattermusch:
Known failures:
ruby linux - I've seen the same issue on master as well (https://source.cloud.google.com/results/invocations/3546bace-3614-4fc1-b854-c43c1b7b9d8f/targets/github%2Fgrpc%2Frun_tests%2Fruby_linux_opt_native%2Ftools%2Frun_tests%2Fhelper_scripts%2Frun_ruby.sh/tests)

PR #29332, opened 2022-04-06
priority and ring_hash LBs: fix interactions when using ring_hash under priority

markdroth:
Fixes interactions between the `priority` and `ring_hash` LB policies, as per the changes in grpc/proposal#296.  Also fixes some unrelated bugs found while looking at the code.

Specific changes in `priority`:
- Use the same idempotent logic to choose a priority in all cases (e.g., when processing a child connectivity state update and when processing an update from the parent).  This avoids problems such as failing to stop using the current priority when it transitions from `READY` to `CONNECTING`, but then seeing a duplicate `TRANSIENT_FAILURE` update from a higher priority that causes us to deselect the current priority because it is in state `CONNECTING` but does not have a failover timer pending.
- Keep track of whether a child has seen `TRANSIENT_FAILURE` more recently than `IDLE` or `READY`, and use this to decide whether to restart the failover timer when a child reports `CONNECTING`.  This ensures that we properly start the failover timer when the `ring_hash` child policy transitions from `IDLE` to `CONNECTING` at startup.
- Do not cancel failover timer when deactivating a child.  This avoids problems where a priority is deactivated while connecting and then reactivated before it finishes connecting.

Specific changes in `ring_hash`:
- Apply new aggregation rule: If there is at least one subchannel in state `TRANSIENT_FAILURE` and there are more than one subchannel, report state `CONNECTING`.  If we hit this rule, proactively start a subchannel connection attempt.
- Properly start a proactive subchannel connection attempt at subchannel list creation if the subchannels are already in a state that makes that appropriate.  Also update the connectivity state visible to the picker at subchannel list creation.
- Use a `ring_hash` picker regardless of what connectivity state we report.
- Treat subchannels that go from `READY` to `TRANSIENT_FAILURE` as being in state `IDLE`, as per the gRFC.
- Use logical connectivity state (after applying special cases for sticky-TF and `IDLE`) for both aggregation rules and for the state visible to the picker.


test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -6429,6 +6430,75 @@ TEST_P(CdsTest, AggregateClusterFallBackFromRingHashAtStartup) {
...
+       0},
+      {"locality1",
+       {MakeNonExistantEndpoint(), MakeNonExistantEndpoint()},
+       kDefaultLocalityWeight,
+       1},
+  });
+  balancer_->ads_service()->SetEdsResource(BuildEdsResource(args));
+  // Populate new CDS resources.
+  Cluster eds_cluster = default_cluster_;
+  eds_cluster.set_name(kEdsClusterName);
```
apolcyn, line 6453: nit: do we not need to set `eds_cluster.set_type(CLUSTER::EDS)` ?

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -6429,6 +6430,75 @@ TEST_P(CdsTest, AggregateClusterFallBackFromRingHashAtStartup) {
...
+                                   new_route_config);
+  // Set Logical DNS result
+  {
+    grpc_core::ExecCtx exec_ctx;
+    grpc_core::Resolver::Result result;
+    result.addresses = CreateAddressListFromPortList(GetBackendPorts());
+    logical_dns_cluster_resolver_response_generator_->SetResponse(
+        std::move(result));
+  }
+  // Inject connection delay to make this act more realistically.
```
apolcyn, line 6493: why do we need this delay? Do we actually need 500ms to trigger a certain previously bad behavior? Or, is the goal to make sure that TCP connect fails asynchronously?

Note that in the integration test for the bug I originally filed, connections for ring hash connections actually fail immediately since the addresses are unroutable. The address in the logical_dns_cluster of the integration test may take a bit longer to connect, though.

apolcyn (commented):
https://github.com/grpc/grpc/pull/29332/commits/f80fcac7a2081cc752b1ded13189e6fc2e662b1a looks mostly good

A couple of comments on the test

test/cpp/end2end/xds/xds_end2end_test.cc
markdroth: That's already set in `default_cluster_`, which we're copying, so we don't need to set it here.

test/cpp/end2end/xds/xds_end2end_test.cc
markdroth: Without the fix, the test does not fail without this delay, so it's not a useful test.  The problem is that we need the subchannels to report CONNECTING for a sufficiently long period of time to allow the right sequence of events to be seen by the priority policy.  That's what I saw happening in the log you posted on the internal bug.

markdroth (commented):
Note that I need to merge #29320 before this, so please review that as well.  Thanks!

markdroth:
Looks like the test needs a little more work here.  Since merging #29321, the test is no longer failing without the code change here.  I'll work on changing it so that it does so.

markdroth:
I significantly rewrote the test to cover a simpler case but to do it in a way that exerts more control over the timing of the individual subchannel connection attempts.  I also simplified the logic in the priority policy so that we always use the same code to choose a priority whenever any child reports a connectivity state change, which should systemically avoid the class of bugs where the logic for choosing a child does not match the logic for reacting to the current child's state changes.

PTAL.

markdroth:
Eric pointed out that the approach of always starting the failover timer in CONNECTING is not quite right, since that will cause us to select a higher-than-current priority that transitions back to CONNECTING when we have already been using a lower priority that is in state READY.  We're still discussing the right way to fix this.

markdroth:
I decided to restore the original test here, even though that test passed without the rest of this PR once I merged #29321, because the latter was actually kind of only a partial fix that is completed by this PR.

I have one more test that I want to add, which I'll work on tomorrow, which is to ensure that ring_hash will proactively start subchannel connection attempts in TF and CONNECTING even without any picks.

markdroth:
I've added the test that ring_hash continues connecting even without picks.

I believe this PR is now ready for review.

src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
```diff
@@ -617,41 +652,41 @@ void RingHash::RingHashSubchannelData::UpdateConnectivityStateLocked(
...
-    if (connectivity_state == GRPC_CHANNEL_READY) {
-      seen_failure_since_ready_ = false;
-      subchannel_list()->UpdateStateCountersLocked(
-          GRPC_CHANNEL_TRANSIENT_FAILURE, connectivity_state);
+  if (last_connectivity_state_ == GRPC_CHANNEL_TRANSIENT_FAILURE) {
+    // If not transitioning to READY, ignore the update, since we want
+    // to continue to consider ourselves in TRANSIENT_FAILURE.
+    if (connectivity_state != GRPC_CHANNEL_READY) return;
+  } else if (connectivity_state == GRPC_CHANNEL_TRANSIENT_FAILURE) {
+    // If we go from READY to TF, treat it as IDLE.
```
apolcyn, line 666: nit: as per our offline chat, can we add more to this comment about why we're transitioning to IDLE (e.g. consistency with other langs and better affinity in case the connection just periodically breaks) ?

src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
```diff
@@ -552,47 +562,72 @@ void RingHash::RingHashSubchannelList::UpdateStateCountersLocked(
...
-  if (num_idle_ > 0 && num_transient_failure_ < 2) {
-    p->channel_control_helper()->UpdateState(
-        GRPC_CHANNEL_IDLE, absl::Status(),
-        absl::make_unique<Picker>(p->Ref(DEBUG_LOCATION, "RingHashPicker"),
-                                  p->ring_));
-    return false;
+    state = GRPC_CHANNEL_READY;
+  } else if (num_transient_failure_ >= 2) {
+    state = GRPC_CHANNEL_TRANSIENT_FAILURE;
+    status = absl::UnavailableError("connections to backends failing");
```
apolcyn, line 589: nit: we could give a little more insight in this error message with: "connections to two or more backends failing" (maybe we'd need to include something to make it clear that the error was generated by ring hash)

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -4861,6 +4863,209 @@ TEST_P(CdsTest, AggregateClusterFallBackFromRingHashAtStartup) {
...
+  ConnectionInjector connection_attempt_injector(backends_[0]->port(),
+                                                 backends_[1]->port());
+  // Wait for P0 backend.
+  // Increase timeout to account for subchannel connection delays.
+  WaitForBackend(0, WaitForBackendOptions(), RpcOptions().set_timeout_ms(2000));
+  // Bring down the P0 backend.
+  ShutdownBackend(0);
+  // Allow the connection attempt to the P1 backend to resume.
+  connection_attempt_injector.CompletePriority1Connection();
+  // Wait for P1 backend to start getting traffic.
```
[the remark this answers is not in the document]
markdroth: We shut down backend 0 before we allow the attempt to connect to backend 1 even begin.  In practice, that's enough to ensure that we stop using P0 and check P1 before P1 becomes READY.

If I change `PriorityLb::ChildPriority::MaybeDeactivateLocked()` to cancel the failover timer, then this test fails consistently (the RPCs fail with "no ready priority").

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -5769,6 +5974,79 @@ TEST_P(CdsTest, RingHashIdleToReady) {
...
+        grpc_core::MutexLock lock(&mu_);
+        const int port = grpc_sockaddr_get_port(addr);
+        gpr_log(GPR_INFO, "==> HandleConnection(): seen_port_=%d, port=%d",
+                seen_port_, port);
+        // Initial attempt should be for port0_, which should fail.
+        // Cancel the RPC at this point, so that it's no longer
+        // queued when the LB policy updates the picker.
+        if (!seen_port_ && port == port_) {
+          gpr_log(GPR_INFO, "*** CANCELLING RPC");
+          rpc_->CancelWithoutJoining();
```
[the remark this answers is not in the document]
markdroth: Good idea, that does make it a bit simpler.  Done.

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -5751,6 +5956,88 @@ TEST_P(CdsTest, RingHashIdleToReady) {
...
+       CreateMetadataValueThatHashesToBackendPort(non_existant_endpoint.port)}};
+  rpc.StartRpc(stub_.get(), RpcOptions().set_timeout_ms(0).set_metadata(
+                                std::move(metadata)));
+  // Wait for the RPC to trigger the P0 connection attempt, then cancel it.
+  connection_injector.WaitForP0ConnectionAttempt();
+  rpc.CancelRpc();
+  // Wait for channel to become connected without any pending RPC.
+  EXPECT_TRUE(channel_->WaitForConnected(grpc_timeout_seconds_to_deadline(5)));
+  // RPC should have been cancelled.
+  EXPECT_EQ(StatusCode::CANCELLED, rpc.GetStatus().error_code());
```
apolcyn, line 6036: nit: can we move this status check up to before `WaitForConnected`, i.e. just after `rpc.CancelRpc()`?

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -5751,6 +5956,88 @@ TEST_P(CdsTest, RingHashIdleToReady) {
...
+        const int port = grpc_sockaddr_get_port(addr);
+        gpr_log(GPR_INFO, "==> HandleConnection(): seen_port_=%d, port=%d",
+                seen_port_, port);
+        // Initial attempt should be for port0_, which should fail.
+        // Cancel the RPC at this point, so that it's no longer
+        // queued when the LB policy updates the picker.
+        if (!seen_port_ && port == port_) {
+          gpr_log(GPR_INFO, "*** SEEN P0 CONNECTION ATTEMPT");
+          seen_port_ = true;
+          cond_.Signal();
```
apolcyn, line 6001: since we still attempt the connection right after signaling here, I think we don't guarantee that the current RPC doesn't attempt the next address

I think we need to pause the attempt here and return early. And then resume the connection attempt from the other thread, after cancelling the RPC ?

test/cpp/end2end/xds/xds_end2end_test.cc
markdroth: The status check really doesn't do anything except read the status.  It doesn't affect the timing of the test at all.  In particular, note that `CancelRpc()` will actually join the thread that was sending the RPC, so we already know it's complete when that returns.

test/cpp/end2end/xds/xds_end2end_test.cc
markdroth: I think the test is reliable as it is.  When I comment out [ring_hash.cc line 595](https://github.com/grpc/grpc/pull/29332/files#diff-89b78f1c2c4146ed13bf937b17468e0ca1ea1630d03a685bdf3928bc12d01b4fR595), the test fails consistently.

apolcyn: OK, this this improvement could make the test more deterministic. But this is fine if test is failing consistently, though.

apolcyn (approved):

PR #29334, opened 2022-04-06
Adjusting tcp min read chunk size growth to reduce wastage of incoming buffer size

Vignesh2208:
Double the tcp min chunk size when a recvmsg indicates more data is available to read immediately. This reduces wasted space in the incoming read buffer.

@markdroth


src/core/lib/iomgr/tcp_posix.cc
```diff
@@ -836,11 +844,21 @@ static void maybe_make_read_slices(grpc_tcp* tcp) {
     int target_length = static_cast<int>(tcp->target_length);
     int extra_wanted =
         target_length - static_cast<int>(tcp->incoming_buffer->length);
+    if (tcp->last_read_completed) {
+      // Set it to false again to start the next block of reads
+      tcp->last_read_completed = false;
+      // Reset last_min_read_chunk_size for the next block of reads
+      tcp->last_min_read_chunk_size = tcp->min_read_chunk_size;
```
ananda1066, line 851: nit: maybe rename `curr_min_read_chunk_size` and `starting_min_read_chunk_size` or something similar?

Vignesh2208: Changed it now to curr_min_read_chunk_size and curr_read_completed

ananda1066 (approved):

src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -61,6 +61,8 @@
 #include "src/core/lib/iomgr/unix_sockets_posix.h"
 #include "src/core/lib/resource_quota/api.h"
 
+static gpr_atm num_dropped_connections = 0;
```
ctiller, line 64: let's use `std::atomic`

Vignesh2208: Done.

PR #29336, opened 2022-04-06
Add scenario generation for PSM tests.

wanlin31:
This commit adds the PSM base scenario to each language in
scenario_config.py allowing per language generation of PSM
base scenarios. The PSM base scenarios will be further updated
in loadtest_config.py for number of client channels and the
number of async_server_thread. After update of the two
aforementioned fields, scenarios varying offered_load will
be generated from the base scenarios.

loadtest_config.py is updated to take number of client channels,
the number of async_servers and a list of targeted offered_load.

### DIFF
--- a/tools/run_tests/performance/loadtest_config.py
+++ b/tools/run_tests/performance/loadtest_config.py
@@ -30,7 +30,7 @@ import json
 import os
 import string
 import sys
-from typing import Any, Dict, Iterable, Mapping, Optional, Type
+from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Type
 
 import yaml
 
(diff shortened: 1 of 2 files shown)


tools/run_tests/performance/loadtest_config.py
```diff
@@ -378,6 +417,19 @@ def main() -> None:
...
+    argp.add_argument(
+        '-offered_loads_list',
+        nargs="*",
+        type=int,
+        default=[],
+        help='A list of offered loads that the tests are running with.')
+    argp.add_argument('-client_channel',
+                      type=int,
+                      help='Number of client channel opened for the test')
+    argp.add_argument(
```
paulosjca, line 429: I would call the argument '--server_threads' and in the description write 'number of async server threads'.

I would also list the arguments in the logical order: client_channels, server_threads, offered_loads.

tools/run_tests/performance/loadtest_config.py
```diff
@@ -118,17 +118,54 @@ def gen_run_indices(runs_per_test: int) -> Iterable[str]:
...
+            scenarios: Iterable[Mapping[str,
+                                        Any]]) -> Iterable[Mapping[str, Any]]:
+        """Transforms scenarios by inserting num of client channels, number of async_server_threads and offered_load."""
+
+        for base_scenario in scenarios:
+            base_scenario['client_config']['client_channels'] = client_channels
+            base_scenario['server_config'][
+                'async_server_threads'] = async_server_threads
+            base_name = base_scenario['name']
+            for offered_load in offered_loads:
```
paulosjca, line 141: There may be a wrinkle here if we assume that len(offered_loads) may be zero (and in that case we just return the scenario without changing the offered load). I would handle that by having a function scenario_name for the name:

```python
def scenario_name(base_name, client_channels, server_threads, offered_load):
  elements = [ base_name ]
  if client_channels:
    elements.append(f'{%d}channels', client_channels)
  if server_threads:
    elements.append(f'{%d}threads', server_threads)
  if offered_load:
    elements.append(f{%d}load', offered_load)
  return '_´.join(elements)
```

Given this function, I would then write the loop as follows:

```python
base_name = base_scenario['name']

if not offered_loads:
  base_scenario['name'] = scenario_name(base_name, client_channels, server_threads, 0)
  yield base_scenario
  return

for offered_load in offered_loads:
  scenario = copy.deepcopy(base_scenario)
  scenario['name'] = scenario_name(base_name, client_channels, server_threads, offered_load)
  scenario['client_config´]['load_params'] = { 'poisson': { 'offered_load': offered_load } }
```

tools/run_tests/performance/loadtest_config.py
```diff
@@ -118,17 +118,72 @@ def gen_run_indices(runs_per_test: int) -> Iterable[str]:
...
+        elements.append(f'{offered_load}load')
+    return '_'.join(elements)
+
+
+def scenario_transform_function(
+    client_channels: int, server_threads: int, offered_loads: Iterable[int]
+) -> Optional[Callable[[Iterable[Mapping[str, Any]]], Iterable[Mapping[str,
+                                                                       Any]]]]:
+    """Returns a transform to be applied to a list of scenarios."""
+    if not any([client_channels, server_threads, len(offered_loads)]):
```
paulosjca, line 137: Nit: I would use () instead of [] since it is immutable, but it is not important.

tools/run_tests/performance/loadtest_config.py
```diff
@@ -118,17 +118,72 @@ def gen_run_indices(runs_per_test: int) -> Iterable[str]:
...
+
+        for base_scenario in scenarios:
+            base_name = base_scenario['name']
+            if client_channels:
+                base_scenario['client_config'][
+                    'client_channels'] = client_channels
+            if server_threads:
+                base_scenario['server_config'][
+                    'async_server_threads'] = server_threads
+            if not offered_loads:
```
paulosjca, line 153: Nit: I would add a blank line before this line and after "return", to make it clear that it breaks there. Probably do the same for the two if block above it as well.

tools/run_tests/performance/loadtest_config.py
```diff
@@ -118,17 +118,72 @@ def gen_run_indices(runs_per_test: int) -> Iterable[str]:
         yield index_fmt.format(i)
 
 
+def scenario_name(base_name, client_channels, server_threads, offered_load):
```
paulosjca, line 121: Nit: I would add a docstring, something like

```python
    """Construct scenario name from base name and modifiers."""
```

paulosjca (approved):

tools/run_tests/performance/loadtest_config.py
wanlin31: I appreciate you point this out :)

tools/run_tests/performance/loadtest_config.py
```diff
@@ -378,6 +440,18 @@ def main() -> None:
...
+                      help='Number of client channels.')
+    argp.add_argument('--server_threads',
+                      type=int,
+                      help='Number of async server threads.')
+    argp.add_argument(
+        '--offered_loads',
+        nargs="*",
+        type=int,
+        default=[],
+        help='A list of offered loads that the tests are running with.')
```
gnossen, line 454: This is a bit vague. How about "A list of QPS values at which each load test scenario will be run."?

tools/run_tests/performance/loadtest_config.py
```diff
@@ -118,17 +118,77 @@ def gen_run_indices(runs_per_test: int) -> Iterable[str]:
...
+        elements.append('{:d}channels'.format(client_channels))
+    if server_threads:
+        elements.append('{:d}threads'.format(server_threads))
+    if offered_load:
+        elements.append('{:d}offered_load'.format(offered_load))
+    return '_'.join(elements)
+
+
+def scenario_transform_function(
+    client_channels: int, server_threads: int, offered_loads: Iterable[int]
```
gnossen, line 135: It looks like `None` is a permissible value for `client_channels` and `server_channels`. Their type should be `Optional[int]`.

gnossen (approved):

Issue #29338, opened 2022-04-06
ProtoBuf at HEAD test failed to built with gRPC Python on Alpine


lidizheng:
```
grpc_tools/main.cc:27:10: fatal error: google/protobuf/compiler/python/python_generator.h: No such file or directory
       27 | #include <google/protobuf/compiler/python/python_generator.h>
          |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...

2022-04-06 21:17:03,265 FAILED: /var/local/git/grpc/tools/run_tests/helper_scripts/build_python.sh [ret=1, pid=100, time=478.7sec]
command: "tools/run_tests/run_tests.py --use_docker -t -j 8 -x run_tests/python_linux_dbg_native_default_python_alpine_buildonly/sponge_log.xml --report_suite_name python_linux_dbg_native_default_python_alpine_buildonly -l python -c dbg --iomgr_platform native --arch default --compiler python_alpine --build_only --report_multi_target"
dockerfile: tools/dockerfile/test/python_alpine_x64
kokoro job name: grpc/core/master/linux/grpc_build_protobuf_at_head
```

Issue #29297, opened 2022-04-02
gRPC-1.45.1 does not compile with protobuf-3.20.0


coryan:
### What version of gRPC and what language are you using?

C++, gRPC-1.45.1

### What operating system (Linux, Windows,...) and version?

The problem manifests in Linux, Windows, and macOS.

### What runtime / compiler are you using (e.g. python version or version of gcc)

This is not compiler or runtime specific.

### What did you do?

Try to compile gRPC-1.45.1 with Protobuf-3.20.0.  Using this [small program](https://github.com/grpc/grpc/files/8402716/protobuf-vs-grpc.tar.gz)

### What did you expect to see?

A successful build.

### What did you see instead?

Bazel errors:

```
ERROR: /usr/local/google/home/coryan/.cache/bazel/_bazel_coryan/1597fbe7e0d6911d6fc0820b75877b84/external/com_envoyproxy_protoc_gen_validate/validate/BUILD:28:17: Label '@com_google_protobuf//:protobuf_python' is duplicated in the 'deps' attribute of rule 'validate_py'
ERROR: /usr/local/google/home/coryan/.cache/bazel/_bazel_coryan/1597fbe7e0d6911d6fc0820b75877b84/external/envoy_api/envoy/config/core/v3/BUILD:7:18: Target '@com_envoyproxy_protoc_gen_validate//validate:validate_proto' contains an error and its package is in error and referenced by '@envoy_api//envoy/config/core/v3:pkg'
ERROR: Analysis of target '//:test' failed; build aborted: 
INFO: Elapsed time: 1.426s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (35 packages loaded, 9 targets configured)
```

gnossen:
This looks like an issue between Protobuf and the Envoy project. Envoy is using the protobuf repo's (private, internal) `py_proto_library` rule [here](https://github.com/envoyproxy/protoc-gen-validate/blob/dfcdc5ea103dda467963fb7079e4df28debcfd28/validate/BUILD#L28). It looks like this might have been caused by [this PR](https://github.com/protocolbuffers/protobuf/pull/9187) on protobuf.

In fact, it looks like [they are aware of the problem](https://github.com/protocolbuffers/protobuf/issues/9688) and [have already rolled the PR back](https://github.com/protocolbuffers/protobuf/pull/9187#issuecomment-1082333538). Going ahead and closing as this isn't really a gRPC issue.

PR #29341, opened 2022-04-07
Fix header file import for latest ProtoBuf in grpcio-tools

lidizheng:
Fixes https://github.com/grpc/grpc/issues/29338.

### DIFF
--- a/tools/distrib/python/grpcio_tools/grpc_tools/main.cc
+++ b/tools/distrib/python/grpcio_tools/grpc_tools/main.cc
@@ -24,7 +24,7 @@
 #include <google/protobuf/compiler/code_generator.h>
 #include <google/protobuf/compiler/command_line_interface.h>
 #include <google/protobuf/compiler/importer.h>
-#include <google/protobuf/compiler/python/python_generator.h>
+#include <google/protobuf/compiler/python/generator.h>
 #include <google/protobuf/descriptor.h>
 #include <google/protobuf/io/zero_copy_stream_impl_lite.h>
 


lidizheng:
Triggered prod:grpc/core/master/linux/grpc_build_protobuf_at_head
* http://sponge/64ab4e1f-68c9-4acf-8ecc-099b0440e224

jtattermusch:
Well, as is, this would break build against the current version of third_party/protobuf submodule.
```
grpc_tools/main.cc:27:10: fatal error: google/protobuf/compiler/python/generator.h: No such file or directory
```

Actually, the situation is quite bad, since it seems that protobuf has recently renamed lots of their public headers.
E.g. see https://github.com/protocolbuffers/protobuf/pull/9669/files#diff-0dd62c8d5a58dc167544764d463de9e6129568f09bbe20145f6ab7d4632f6608R111

And it's not just about python, other languages would likely suffer from a similar problem:
https://github.com/protocolbuffers/protobuf/pull/9669/files#diff-deef97e4c58bafd49f04825aa86dcc91ee390d5d23d041c3d7ccb9d4529ee72b

lidizheng:
I'm planning to update the ProtoBuf submodule again, and try to fix breakages. I see ProtoBuf pushed packages to various package management plaforms (like PyPI), it might be harder to yank all that.

The ProtoBuf build at head job is failing constantly for at least 2 weeks (around the mentioned commit). We might want to dig into why this didn't raise an alarm for both teams. https://screenshot.googleplex.com/Ahybmk6DF5WkTtG

lidizheng:
Updating ProtoBuf is blocked by another issue: https://github.com/grpc/grpc/issues/29297.

lidizheng:
We probably need to figure out who should fix the issue first.

PR #29345, opened 2022-04-07
CFStream Test Suite Update Part1

HannahShiSFB:
run CFStream test by bazel

https://github.com/grpc/grpc-ios/issues/48

@veblush
@dennycd

### DIFF
--- a/test/core/iomgr/BUILD
+++ b/test/core/iomgr/BUILD
@@ -14,6 +14,7 @@
 
 load("//bazel:grpc_build_system.bzl", "grpc_cc_library", "grpc_cc_test", "grpc_package")
 load("//bazel:custom_exec_properties.bzl", "LARGE_MACHINE")
+load("@build_bazel_rules_apple//apple:ios.bzl", "ios_unit_test")
 
 licenses(["notice"])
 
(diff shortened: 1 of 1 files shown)


test/core/iomgr/BUILD
```diff
@@ -363,3 +364,20 @@ grpc_cc_test(
         "//test/core/util:grpc_test_util",
     ],
 )
+
+objc_library(
```
dennycd, line 368: recommend we move this into src/objective-c/tests since this is ios platform specific test using objc , thanks

dennycd (commented):
looks like there is some bazel related build failures. Can you rebase and try again, thanks 1

test/core/iomgr/BUILD
HannahShiSFB: This will require copy CFStream test files (test/core/iomgr/ios/CFStreamTests/*) over to src/objective-c/tests (maybe put together with CoreTests?).

Before that, I have some questions in running them with bazel:
1. Currently CFStreamTests are run under TSAN, ASAN and MSAN as well, I’m not sure how to configure CI to run them with sanitizers.
2. It appears that bazel doesn’t support run MSAN with iOS, is this something can be configured or not supported in bazel? Can we drop the MSAN support or should we keep the Xcode tests?

Error running `bazel run --config=msan //test/core/iomgr:CFStreamTest`
clang: error: unsupported option '-fsanitize=memory' for target 'x86_64-apple-ios-simulator’

dennycd: Thanks Hannah, looks like the existing CFStream test is applying TSAN/ASAN/MSAN via xcode build schema options.  There is no exact mapping directly from bazel configurations (-fsanitize or -config=msan are low-level clang compiler flags and may not work with the ios simulator runner, thus the error). 

We likely need to apply this via [xcconfig](https://pewpewthespells.com/blog/xcconfig_guide.html) by setting the corresponding [xcode build configuration setting ](https://xcodebuildsettings.com/) for TSAN/ASAN/MSAN (e.g. ENABLE_THREAD_SANITIZER). Let's chat offline for more details on this one, thanks !

PR #29347, opened 2022-04-07
Update third_party/protobuf to 3.20.1-rc1

mkruskal-google:


jtattermusch (changes requested):
Marking as "request changes" since this change is for test only and should not be merged.
(we will only merge once 3.20.1 stable is released).

yashykt:
Closing since it seems outdated.

PR #29353, opened 2022-04-08
Simplify running bazel with structured test results on CI

jtattermusch:
To run bazel in a way that produces structured test results in kokoro (the "Bazel" link on test result, the resultstore link for the bazel invocation, the results in bigquery) is nontrivial and contains a lot of boilerplate that ends up being duplicated in many of our test scripts.

This PR introduces a `bazel_report_helper.py` that generates a "bazel wrapper" script that adds the right bazel arguments and followup actions to accomplish all of the above.

- add bazel_report_helper.py (to encapsulate 1.the "Bazel" link on test result 2. the resultstore link for the bazel invocation 3. the results in bigquery)
- add experimental feature to generate top level XML report for each bazel invocation (currently mostly unused but will come handy in follow-up PRs when I want to use bazel report helper for our other builds).
- convert some of our test scripts to use the generated bazel wrapper (and thus remove the boilerplate).

More bazel jobs will be updated to use the wrapper (and thus produce easier to digest test results and persist the test history).

### DIFF
--- a/tools/internal_ci/linux/grpc_bazel_on_foundry_base.sh
+++ b/tools/internal_ci/linux/grpc_bazel_on_foundry_base.sh
@@ -23,28 +23,10 @@ source tools/internal_ci/helper_scripts/prepare_build_linux_rc
 # make sure bazel is available
 tools/bazel version
 
-# to get "bazel" link for kokoro build, we need to generate
-# invocation UUID, set a flag for bazel to use it
-# and upload "bazel_invocation_ids" file as artifact.
-BAZEL_INVOCATION_ID="$(uuidgen)"
-echo "${BAZEL_INVOCATION_ID}" >"${KOKORO_ARTIFACTS_DIR}/bazel_invocation_ids"
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path bazel_rbe
 
-tools/bazel \
+bazel_rbe/bazel_wrapper \
   --bazelrc=tools/remote_build/linux_kokoro.bazelrc \
   test \
-  --invocation_id="${BAZEL_INVOCATION_ID}" \
-  --workspace_status_command=tools/remote_build/workspace_status_kokoro.sh \
-  $@ \
-  -- //test/... || FAILED="true"
-
-if [ "$UPLOAD_TEST_RESULTS" != "" ]
-then
-  # Sleep to let ResultStore finish writing results before querying
-  sleep 60
-  python3 ./tools/run_tests/python_utils/upload_rbe_results.py
-fi
-
-if [ "$FAILED" != "" ]
-then
-  exit 1
-fi
+  "$@" \
+  -- //test/...
--- a/tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
+++ b/tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
@@ -14,18 +14,23 @@
 # limitations under the License.
 #
 # Test full Bazel
-#
-# NOTE: No empty lines should appear in this file before igncr is set!
-set -ex -o igncr || set -ex
+
+set -ex
 
 mkdir -p /var/local/git
 git clone /var/local/jenkins/grpc /var/local/git/grpc
 (cd /var/local/jenkins/grpc/ && git submodule foreach 'cd /var/local/git/grpc \
 && git submodule update --init --reference /var/local/jenkins/grpc/${name} \
 ${name}')
-cd /var/local/git/grpc/test
+cd /var/local/git/grpc
 TEST_TARGETS="//src/python/... //tools/distrib/python/grpcio_tools/... //examples/python/..."
 BAZEL_FLAGS="--test_output=errors"
-bazel test ${BAZEL_FLAGS} ${TEST_TARGETS}
-bazel test --config=python_single_threaded_unary_stream ${BAZEL_FLAGS} ${TEST_TARGETS}
-bazel test --config=python_poller_engine ${BAZEL_FLAGS} ${TEST_TARGETS}
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests
+python_bazel_tests/bazel_wrapper test ${BAZEL_FLAGS} ${TEST_TARGETS}
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests_single_threaded_unary_streams
+python_bazel_tests_single_threaded_unary_streams/bazel_wrapper test --config=python_single_threaded_unary_stream ${BAZEL_FLAGS} ${TEST_TARGETS}
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests_poller_engine
+python_bazel_tests_poller_engine/bazel_wrapper test --config=python_poller_engine ${BAZEL_FLAGS} ${TEST_TARGETS}
--- a/tools/run_tests/python_utils/bazel_report_helper.py
+++ b/tools/run_tests/python_utils/bazel_report_helper.py
@@ -43,7 +43,7 @@ def _platform_string():
         return 'posix'
 
 
-def _append_to_kokoro_bazel_invocations(invocation_id):
+def _append_to_kokoro_bazel_invocations(invocation_id: str) -> None:
     """Kokoro can display "Bazel" result link on kokoro jobs if told so."""
     # to get "bazel" link for kokoro build, we need to upload
     # the "bazel_invocation_ids" file with bazel invocation ID as artifact.
@@ -65,7 +65,8 @@ def _append_to_kokoro_bazel_invocations(invocation_id):
         pass
 
 
-def _generate_junit_report_string(report_suite_name, invocation_id, success):
+def _generate_junit_report_string(report_suite_name: str, invocation_id: str,
+                                  success: bool) -> None:
     """Generate sponge_log.xml formatted report, that will make the bazel invocation reachable as a target in resultstore UI / sponge."""
     bazel_invocation_url = 'https://source.cloud.google.com/results/invocations/%s' % invocation_id
     package_name = report_suite_name
@@ -93,8 +94,8 @@ def _generate_junit_report_string(report_suite_name, invocation_id, success):
     return '\n'.join(lines)
 
 
-def _create_bazel_wrapper(report_path, report_suite_name, invocation_id,
-                          upload_results):
+def _create_bazel_wrapper(report_path: str, report_suite_name: str,
+                          invocation_id: str, upload_results: bool) -> None:
     """Create a "bazel wrapper" script that will execute bazel with extra settings and postprocessing."""
 
     os.makedirs(report_path, exist_ok=True)
--- a/tools/run_tests/python_utils/bazel_report_helper.py
+++ b/tools/run_tests/python_utils/bazel_report_helper.py
@@ -54,13 +54,11 @@ def _append_to_kokoro_bazel_invocations(invocation_id: str) -> None:
                   'a') as f:
             f.write(invocation_id + '\n')
         print(
-            'Added invocation ID %s to kokoro "bazel_invocation_ids" artifact' %
-            invocation_id,
+            f'Added invocation ID {invocation_id} to kokoro "bazel_invocation_ids" artifact',
             file=sys.stderr)
     else:
         print(
-            'Skipped adding invocation ID %s to kokoro "bazel_invocation_ids" artifact'
-            % invocation_id,
+            f'Skipped adding invocation ID {invocation_id} to kokoro "bazel_invocation_ids" artifact',
             file=sys.stderr)
         pass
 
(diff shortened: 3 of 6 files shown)


tools/internal_ci/linux/grpc_bazel_on_foundry_base.sh
```diff
@@ -23,28 +23,10 @@ source tools/internal_ci/helper_scripts/prepare_build_linux_rc
...
-  # Sleep to let ResultStore finish writing results before querying
-  sleep 60
-  python3 ./tools/run_tests/python_utils/upload_rbe_results.py
-fi
-
-if [ "$FAILED" != "" ]
-then
-  exit 1
-fi
+  "$@" \
```
lidizheng, line 31: Quoting `$@` is a behavior change. Are we certain we want this?

tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
```diff
@@ -14,18 +14,23 @@
...
-bazel test --config=python_poller_engine ${BAZEL_FLAGS} ${TEST_TARGETS}
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests
+python_bazel_tests/bazel_wrapper test ${BAZEL_FLAGS} ${TEST_TARGETS}
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests_single_threaded_unary_streams
+python_bazel_tests_single_threaded_unary_streams/bazel_wrapper test --config=python_single_threaded_unary_stream ${BAZEL_FLAGS} ${TEST_TARGETS}
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests_poller_engine
+python_bazel_tests_poller_engine/bazel_wrapper test --config=python_poller_engine ${BAZEL_FLAGS} ${TEST_TARGETS}
```
lidizheng, line 36: Since the Python script will geneate the wrapper command, have we considered the alternative of sending all arguments to the Python script? For example:

```bash
python3 tools/run_tests/python_utils/bazel_report_helper.py \
   --report_path python_bazel_tests_poller_engine \
   --bazel_args="test --config=python_poller_engine ${BAZEL_FLAGS} ${TEST_TARGETS}"
```

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+    required=True,
+    type=str,
+    help=
+    'Path under which the bazel wrapper and other files are going to be generated'
+)
+argp.add_argument('--report_suite_name',
+                  default='bazel_invocations',
+                  type=str,
+                  help='Test suite name to use in generated XML report')
+args = argp.parse_args()
```
lidizheng, line 241: optional: I would recommend to move the ArgumentParser portion to the top (for better visibility)

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+# generate new bazel invocation ID
+invocation_id = str(uuid.uuid4())
+
+report_path = args.report_path
+report_suite_name = args.report_suite_name
+upload_results = True if os.getenv('UPLOAD_TEST_RESULTS') else False
+
+_append_to_kokoro_bazel_invocations(invocation_id)
+_create_bazel_wrapper(report_path, report_suite_name, invocation_id,
+                      upload_results)
```
lidizheng, line 252: Let's create a main function, and protect this file with:

```py
if __name__ == '__main__':
    # Execute when the module is not initialized from an import statement.
    ...
```

https://docs.python.org/3/library/__main__.html#idiomatic-usage

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+        return 'windows'
+    elif platform.system() == 'Darwin':
+        return 'mac'
+    elif platform.system() == 'Linux':
+        return 'linux'
+    else:
+        return 'posix'
+
+
+def _append_to_kokoro_bazel_invocations(invocation_id):
```
lidizheng, line 41: Recommend type annotations.

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+                                          invocation_id,
+                                          success=False))
+    with open(success_report_filename, 'w') as f:
+        f.write(
+            _generate_junit_report_string(report_suite_name,
+                                          invocation_id,
+                                          success=True))
+
+    # generate the bazel wrapper for linux/macos
+    with open(bazel_wrapper_filename, 'w') as f:
```
lidizheng, line 142: Have we considered the alternative of creating one or more Bash scripts accepting different configs (than trying to generate a bash script)? Current version works, but we might need to maintain the generation script for years. Any effort making it easier to maintain might be quite rewarding down the line.

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+        intro_lines = [
+            '#!/bin/bash',
+            'set -ex',
+            '',
+            'tools/bazel --bazelrc="%s" "$@" || FAILED=true' %
+            bazel_rc_filename,
```
lidizheng, line 148: nit: Let's use f-string if possible?

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+            '',
+        ]
+
+        if upload_results:
+            upload_results_lines = [
+                '# Sleep to let ResultStore finish writing results before querying',
+                'sleep 60',
```
lidizheng, line 155: We might want to make this configurable. E.g., a script-wide constant at the top. Also... if the script is going to be disposable, it probably isn't that useful to add comments in it?

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+            _generate_junit_report_string(report_suite_name,
+                                          invocation_id,
+                                          success=False))
+    with open(success_report_filename, 'w') as f:
+        f.write(
+            _generate_junit_report_string(report_suite_name,
+                                          invocation_id,
+                                          success=True))
+
+    # generate the bazel wrapper for linux/macos
```
lidizheng, line 141: Should we add bazel_wrapper to gitignore?

tools/run_tests/python_utils/bazel_report_helper.py
```diff
@@ -0,0 +1,252 @@
...
+
+        lines = [
+            line + '\n'
+            for line in intro_lines + upload_results_lines + outro_lines
+        ]
+        f.writelines(lines)
+    os.chmod(bazel_wrapper_filename, 0o775)  # make the unix wrapper executable
+
+    # generate bazel wrapper for windows
+    with open(bazel_wrapper_bat_filename, 'w') as f:
```
lidizheng, line 182: I would recommend to split the two generation process into two functions.

lidizheng (commented):
Thanks for working on this. This will make Bazel test results much better presented with ResultStore and Kokoro.

lidizheng:
@gnossen Can you also help to take a look?

gnossen:
Could you integrate this into the `tools/bazel` script and condition activation of the test reports on an environment variable so that we don't have to update all of the call sites?

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: Yup, good point, I'll do that.

tools/internal_ci/linux/grpc_bazel_on_foundry_base.sh
jtattermusch: Yes, that was an intentional change. "$@"  is more correct and afaik introducing it didn't break any builds.

tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
jtattermusch: > Could you integrate this into the `tools/bazel` script and condition activation of the test reports on an environment variable so that we don't have to update all of the call sites?

I'd rather not, at least not initially.  I don't want to create one complex wrapper that "does everything". Currently `tools/bazel` takes care of selecting the right bazel version but doesn't touch the bazel flags at all and it doesn't know anything about CI, kokoro or resultstore, reports. Purpose of the new wrapper is completely different so I think layering purpose specific wrappers on top of each other is actually better solution that having everything together in a single wrapper.

I think having to update the call sites is actually what we want initially - it makes the introduction of the reports/uploading logic more apparent and clearly differentiates between places where we do upload structured results and where we don't (yet).

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: I considered having wrapper scripts at first, but I wasn't happy with the result. What I ran into:
- we need the logic for both a .sh wrapper and .bat wrapper. if there's a common python script, most of the "work" (e.g. generate the UUID, generate the rc file, generate the report etc.) can be done by the python script (with good comments and readability), while we just bake the hardcoded results into the generated wrapper. In bash/bat, even "trivial" task quickly become a pain  and the resulting "code" has bad readability and also is duplicated.
- in python, coming up with future tweaks is much simpler than in .sh/.bat
- with pre-existing bash scripts, passing the configuration to the wrapper script isn't easy. Since we need the wrapper to forward all the command line arguments to bazel, we are left with setting environment variables to pass values like the invocation ID to the script, at which point simply baking the arguments into generated code just comes off as easier, especially since we're already generating other files (e.g. the bazelrc and the xml report).

I've given this quite a bit of thought and the current version is what I believe makes the most sense. I'm definitely open to improvements in the future, but I think it's going to be hard to come up with major simplifications. But yeah, often times simplifying stuff improves our ability to see even more possible improvements, so I'll be looking whether we can improve this even further (as followup work).

jtattermusch:
> Could you integrate this into the `tools/bazel` script and condition activation of the test reports on an environment variable so that we don't have to update all of the call sites?

I'd rather not, at least not initially.  I don't want to create one complex wrapper that "does everything". Currently `tools/bazel` takes care of selecting the right bazel version but doesn't touch the bazel flags at all and it doesn't know anything about CI, kokoro or resultstore, reports. Purpose of the new wrapper is completely different so I think layering purpose specific wrappers on top of each other is actually better solution that having everything together in a single wrapper.

I think having to update the call sites is actually what we want initially - it makes the introduction of the reports/uploading logic more apparent and clearly differentiates between places where we do upload structured results and where we don't (yet).

tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
jtattermusch: - I think this would only make sense if python would invoke the bazel (otherwise I don't see how that simplifies stuff, we would just need to persist all the arguments into the wrapper and then use them). But invoking bazel from python is something we'd have to test (especially on windows, it could mess with things in unexpected ways).
- I think having the arbitrary args pass through python is potentially risky (we would need to pay close attention to quoting, splitting the args etc.) - I think this is extra trouble we don't need. Just forwarding "$@" in the wrapper seems safer, simpler and less error-prone.

tools/run_tests/python_utils/bazel_report_helper.py
lidizheng: Good point about some logic is much easier to implement in Python than bash or batch script. If the options are using Python to wrap shell script vs. using shell script to wrap Python commands, the former is more manageable and clear (with the comments and abstractions).

tools/run_tests/python_utils/bazel_report_helper.py
lidizheng: Bump. This impacts whether we will keep the generated files permanently.

lidizheng (approved):
PTAL at some smaller comments, LGTM.

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: done.

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: done.

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: done.

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: done (introducing `__main__` has solved this).

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: done.

tools/run_tests/python_utils/bazel_report_helper.py
jtattermusch: I'll leave as is as the function already generates multiple different files and not just the two wrappers.

jtattermusch:
Turns out the ubuntu1604 workers (used by bazel RBE builds that don't run under a docker container) are still using python3.5, and python3.6 is required for "f-strings". So I'm going to revert the "f-strings" refactoring for now, and reintroduce it as a followup (to avoid making this PR complicated).

PR #29354, opened 2022-04-08
Better bazel C++ iOS tests (with structured test results)

jtattermusch:
~~Based on https://github.com/grpc/grpc/pull/29353 (ignore the first few commits).~~ (rebased)

- cleanup and modernize the bazel_cpp_ios tests (use a single bazel invocation, produce test results in resultstore)
- use bazel_report_helper.py for the cpp ios tests.
- remove the unused bazel_core_ios tests (the test configs and script currently exist but they are not run at all by kokoro).

PR #29355, opened 2022-04-08
Simplify ARM64 bazel invocation (and produce resultstore bazel results).

jtattermusch:
Followup for https://github.com/grpc/grpc/pull/29354.

- produce the bazel invocation link for linux ARM64 bazel C/C++ tests
- make use of the bazer_report_helper.py


tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
```diff
@@ -25,5 +25,13 @@ cd /var/local/git/grpc
 # tests require port server to be running
 python3 tools/run_tests/start_port_server.py
 
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path bazel_test_c_cpp
+
 # test gRPC C/C++ with bazel
```
lidizheng, line 30: Since the bazel wrapper is generated for a single use, I would suggest to put line 28 and line 31 together and move the comment up. This sends a better signal to readers that these two commands are coupled and are expected to be called together.

tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
```diff
@@ -25,5 +25,13 @@ cd /var/local/git/grpc
...
-tools/bazel test --config=opt --test_output=errors --test_tag_filters=-no_linux,-no_arm64 --build_tag_filters=-no_linux,-no_arm64 --flaky_test_attempts=1 --runs_per_test=1 //test/...
+bazel_test_c_cpp/bazel_wrapper \
+  --bazelrc=tools/remote_build/include/test_locally_with_resultstore_results.bazelrc \
```
lidizheng, line 32: Should this bazelrc be added via the Bazel wrapper?

tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
```diff
@@ -25,5 +25,13 @@ cd /var/local/git/grpc
...
+  test --config=opt \
```
lidizheng, line 33: Are we sure we want to remove --flaky_test_attempts=1 and --runs_per_test=1?

tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
jtattermusch: done.

tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
jtattermusch: yes, those flags were initially used to assess the level of test flakiness on ARM64, but we're past that point and we no longer need them. I'm trying to make the arm64 config as close to the x64 config as possible.

tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
jtattermusch: not really (or at least not yet). In all the other builds we have, the options from  test_locally_with_resultstore_results.bazelrc are set separately from the bazel wrapper (the wrapper only sets the "kokoro specific" bits, like the kokoro "Bazel" link, kokoro job invocation details etc.

We can potentially reconsider as a followup.

jtattermusch:
Adhoc run: http://sponge/3d5b120a-2941-43ab-87b5-3e40b4649b06

tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
lidizheng: Sounds good. It's not a blocker for this PR.

With the bazel wrapper, the script can do many magical things. The bazelrc can be picked by the input of the bazel report helper or by the platform.

lidizheng (approved):

tools/internal_ci/linux/arm64/grpc_bazel_test_c_cpp_in_docker.sh
jtattermusch: Agreed. We need to find good balance between "too much magic" (wrapper has too much automatic stuff in and stops being transparent and predictable) and "not enough magic" (does too little so it doesn't help in practice) though.

PR #29363, opened 2022-04-08
Added GRPC_MUST_USE_RESULT_WHEN_USE_STRICT_WARNING to grpc::Status

veblush:
To prevent gRPC from having https://github.com/grpc/grpc/pull/29358, `GRPC_MUST_USE_RESULT_WHEN_USE_STRICT_WARNING` is added and applied to `grpc::Status`. This is working as `GRPC_MUST_USE_RESULT` but only when it's built with `USE_STRICT_WARNING`. Since `grpc::Status` is public API, we should be careful when changing its prototype although changing this attribute isn't necessarily a breaking change but it may introduce surprising errors.

### DIFF
--- a/include/grpc/impl/codegen/port_platform.h
+++ b/include/grpc/impl/codegen/port_platform.h
@@ -528,6 +528,11 @@ typedef unsigned __int64 uint64_t;
 #define GRPC_MUST_USE_RESULT
 #define GPR_ALIGN_STRUCT(n)
 #endif
+#ifdef USE_STRICT_WARNING
+#define GRPC_MUST_USE_RESULT_WHEN_USE_STRICT_WARNING GRPC_MUST_USE_RESULT
+#else
+#define GRPC_MUST_USE_RESULT_WHEN_USE_STRICT_WARNING
+#endif
 #endif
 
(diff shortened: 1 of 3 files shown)


markdroth (approved):
This could conceivably still be a breaking change for users that build with bazel and `--define=use_strict_warning=true`, but that's probably a narrow enough set of users that it's okay.

Thanks for doing this!

include/grpc/impl/codegen/port_platform.h
```diff
@@ -528,6 +528,11 @@ typedef unsigned __int64 uint64_t;
 #define GRPC_MUST_USE_RESULT
 #define GPR_ALIGN_STRUCT(n)
 #endif
+#ifdef USE_STRICT_WARNING
+#define GRPC_MUST_USE_RESULT_WHEN_USE_STRICT_WARNING GRPC_MUST_USE_RESULT
```
ctiller, line 532: Can we get a comment about what this is and when it should be used?

veblush: Added. PTAL.

ctiller (approved):

PR #29372, opened 2022-04-11
Adding is_optional case to RLS

donnadionne:


markdroth:
Can you please recreate this as a revert of #29299, with a separate commit for whatever additional changes are here?  That will make it much easier to review.

donnadionne:
> Can you please recreate this as a revert of #29299, with a separate commit for whatever additional changes are here? That will make it much easier to review.

I was not able to create a revert successfully as i think the test file has changed for the revert to be clean.  So I did it manually and discovered indeed the test file had 2 major changes:
1. the scoped env var
2. Needing to call Create on endpoints.

After adjusting these 2 items, all tests passed.

PR #29373, opened 2022-04-12
Adding is_optional case to RLS revert the revert of 29259

donnadionne:


markdroth:
Even if there are merge conflicts, the change can still be done as a revert with manual conflict resolution.  I've sent you #29378 to do this.

PR #29375, opened 2022-04-12
Disable layering check for Objective-C

googlewalt:
These targets are currently not layering check clean.  Disabling it
ensures that they will continue to build even when we enable layering
check for Objective-C internally.

@veblush

### DIFF
--- a/src/objective-c/BUILD
+++ b/src/objective-c/BUILD
@@ -19,7 +19,10 @@ load("@build_bazel_rules_apple//apple:resources.bzl", "apple_resource_bundle")
 
 licenses(["notice"])
 
-package(default_visibility = ["//visibility:public"])
+package(
+    default_visibility = ["//visibility:public"],
+    features = ["-layering_check"],
+)
 
 exports_files(["LICENSE"])
 


markdroth:
I suspect that this is not what we actually want, since this would disable the layering check for all builds, not just ObjC.  But I'll let Nico and Denny decide.

googlewalt:
I think there are only grpc_objc_libraries (that uses layering check) in this BUILD file?  But if you prefer, instead of disabling at the package level, I can disable for the individual rules instead.

markdroth:
Oh, sorry, you're right -- for some reason I thought this was being done in the top-level BUILD file.

Anyway, I'll let Nico and Denny review.

nicolasnoble:
Yeah if it's only for objective-c, I don't really have any objection myself. @dennycd ?

dennycd (approved):
Thanks @googlewalt .  changes LGTM.  Are there any existing build logs that we can check to 
 see if gRPC/ObjC targets will pass layering check once they are turned on ?

googlewalt:
Thanks for the review.  I am sending you a link to the internal logs via email.

PR #29377, opened 2022-04-12
iOS binary size script cleanup

jtattermusch:
The original version of the script has terrible readability. Tried to improve that at least a bit.
- also made the script always print the sizes, even when diff is not significant.

The next step should be looking into how to speed up the jobs (currently `grpc/core/pull_request/macos/grpc_ios_binary_size` takes about 40mins which is way too much for what it's doing).
I wanted to first get a sense of what exactly the job is doing, before we make modifications (hence the cleanup).

### DIFF
--- a/tools/internal_ci/macos/grpc_ios_binary_size.cfg
+++ b/tools/internal_ci/macos/grpc_ios_binary_size.cfg
@@ -1,0 +1,33 @@
+# Copyright 2022 The gRPC Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Config file for the internal CI (in protobuf text format)
+
+# Location of the continuous shell script in repository.
+build_file: "grpc/tools/internal_ci/macos/grpc_ios_binary_size.sh"
+timeout_mins: 90
+before_action {
+  fetch_keystore {
+    keystore_resource {
+      keystore_config_id: 73836
+      keyname: "grpc_checks_private_key"
+    }
+  }
+}
+action {
+  define_artifacts {
+    regex: "**/*sponge_log.*"
+    regex: "github/grpc/reports/**"
+  }
+}
--- a/tools/internal_ci/macos/grpc_ios_binary_size.sh
+++ b/tools/internal_ci/macos/grpc_ios_binary_size.sh
@@ -27,5 +27,14 @@ cd $(dirname $0)/../../..
 export PREPARE_BUILD_INSTALL_DEPS_OBJC=true
 source tools/internal_ci/helper_scripts/prepare_build_macos_rc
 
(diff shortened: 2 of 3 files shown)


jtattermusch:
Example of the new output (see "Checks" tab).
```
Objective-C binary sizes
**********************STATIC******************
       New size                      Old size
      2,606,870       Core (=)      2,606,870
        154,631       ObjC (=)        154,631
        789,720  BoringSSL (=)        789,720
        385,369   Protobuf (=)        385,369

      4,281,386      Total (=)      4,281,386

 No significant differences in binary sizes

********************FRAMEWORKS****************
       New size                      Old size
     17,080,256       Core (=)     17,080,256
      1,308,471       ObjC (=)      1,308,471
              0  BoringSSL (=)              0
      1,682,420   Protobuf (=)      1,682,420

     27,622,048      Total (>)     27,622,047

 No significant differences in binary sizes
```

dennycd:
Thanks Jan.  Looks like this is using cocoapod & xcodebuild, and we should be able to switch to bazel-based build with src/objective-c/tests/build_one_example_bazel.sh script.  I will follow up with an update on this. 

Currently the iOS binary diff is non-blocking. It may not be much value to run per PR since most PRs are C-core changes and it is essentially just measuring c-core binary size diffs (I think we already have a separate binary size tracking for that ?).   Ideally if we could 

 1. if viable, only trigger iOS binary size diff if PR touches src/objective-c/
 2. always run on master like u mentioned

I think # 1 would still be useful for us to identify binary size regression coming from ObjC wrapper layer changes.

jtattermusch:
> Thanks Jan. Looks like this is using cocoapod & xcodebuild, and we should be able to switch to bazel-based build with src/objective-c/tests/build_one_example_bazel.sh script. I will follow up with an update on this.

If it is possible to build with bazel and still get useful binary size stats (this is something you'd have to look into), I think that would be our strongly preferred approach.
bazel is easier to deal with (especially if you need to build multiple revisions in a single build since bazel is much less prone to pollution from previous builds). And we could use the remote cache to make the builds much faster.

> 
> Currently the iOS binary diff is non-blocking. It may not be much value to run per PR since most PRs are C-core changes and it is essentially just measuring c-core binary size diffs (I think we already have a separate binary size tracking for that ?). Ideally if we could
> 
> 1. if viable, only trigger iOS binary size diff if PR touches src/objective-c/

In theory yes, it's possible, but experience has shown that these tricks are fragile and hard to maintain, so I'd rather not do it unless absolutely necessary (and it looks like we have other options here).

> 2. always run on master like u mentioned

we could definitely run the job on master, and it should be quite easy to setup (but of course it would only show the current numbers, not the diff).

> 
> I think # 1 would still be useful for us to identify binary size regression coming from ObjC wrapper layer changes.

jtattermusch:
adhoc master run: https://fusion2.corp.google.com/invocations/2b7fd9ad-af01-46fb-ac99-6adcb11d360c/targets

jtattermusch:
@dennycd  added a job config to enable running on master. PTAL.

dennycd (approved):

ctiller:
Note that if we did restrict the paths that trigger this we'd want to include src/core, include too, since they contribute a bulk of the bytes to an ios build.

jtattermusch:
> Note that if we did restrict the paths that trigger this we'd want to include src/core, include too, since they contribute a bulk of the bytes to an ios build.

Not sure I understand that correctly. We haven't changed what gets included in the "iOS binary size" at all, this PR just refactors the script and the format used to print the results.
Also, AFAIK "Core" is already included in the calculation (see https://github.com/grpc/grpc/pull/29377#issuecomment-1097030349, isn't that what you meant?)

PR #29380, opened 2022-04-12
Workaround for the unaligned malloc

veblush:
Upb arena relies on the assumption where malloc returns aligned memory but it doesn't hold on Windows with MSVC 2015. As a quick workaround, I removed the assertion mainly because Windows on x86 architecture should be fine with unaligned memory. Also this wasn't a regression caused by the latest upb but having more conservative assertions. In addition to this SSE4 code is disabled as a safety measure because it may expect aligned pointers.

Actual fix will land later once upb fixes this issue and this PR should be enough to unblock gRPC release.

b/228862443

### DIFF
--- a/third_party/upb/third_party/utf8_range/utf8_range.h
+++ b/third_party/upb/third_party/utf8_range/utf8_range.h
@@ -1,5 +1,5 @@
 
-#if (defined(__ARM_NEON) && defined(__aarch64__)) || defined(__SSE4_1__)
+#if 0
 int utf8_range2(const unsigned char* data, int len);
 #else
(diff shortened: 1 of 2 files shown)


haberman (approved):

lidizheng:
The re-triggered run seems to fix the Kokoro issue, the portability test didn't end early, it's a good sign.

lidizheng:
@veblush The portability tests passed. The failed C++ macOS looks unrelated to this PR. Can you help to double check and merge if it's ready?

veblush:
@lidizheng Thanks for running the portability test. And `Basic Tests C/C++ MacOS` doesn't seem to relate to this PR

PR #29387, opened 2022-04-12
Revert "Simplify running bazel with structured test results on CI"

lidizheng:
Reverts grpc/grpc#29353

PR #29301, opened 2022-04-04
Cleanup configuration of bazel RBE jobs

jtattermusch:
Followup for https://github.com/grpc/grpc/pull/29353

- cleanup and simplify .cfg and .sh files for linux bazel RBE jobs (make use of env variables instead of helper .sh scripts)
- enable uploading of XML report artifacts for selected bazel jobs (the idea is to enable the bazel_report_helper.py reports to show up in resultstore/sponge).
- use "respawn in moved src tree" for linux bazel RBE jobs (to speed up job finalization).

PR #29389, opened 2022-04-13
Fix bazel RBE builds broken on master

jtattermusch:
Fixup for #29353.
Supersedes https://github.com/grpc/grpc/pull/29387.

Sorry for the breakage.

The concern in https://github.com/grpc/grpc/pull/29353/files#diff-fb339b3529fae792999cdf0604ce818c3a873177ee85aca0c52ac08435cb2353R31 turns out to be fair, I missed that some of the job configs are explicitly passing e.g. `"--config=dbg --cache_test_results=no"` as a single arg (which makes no sense).

### DIFF
--- a/tools/internal_ci/linux/grpc_bazel_on_foundry_dbg.sh
+++ b/tools/internal_ci/linux/grpc_bazel_on_foundry_dbg.sh
@@ -16,5 +16,4 @@
 set -ex
 
 export UPLOAD_TEST_RESULTS=true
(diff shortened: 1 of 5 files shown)


jtattermusch:
The already approved change https://github.com/grpc/grpc/pull/29301 also seems to be fixing the problem (side effect of refactoring done there), so I think I'm going to close this PR and merge https://github.com/grpc/grpc/pull/29301 instead.

jtattermusch:
Superseded by https://github.com/grpc/grpc/pull/29301.

PR #29392, opened 2022-04-13
secure endpoint: Don't allocate slice buffers when using the zero copy protector, as they are entirely unused.

dpcollins-google:
@markdroth


ZhenLian (approved):
The changes LGTM. Some tests failed to run. Let me re-run it to see if they still fail.

markdroth (approved):
Yes, the test failures look unrelated.

PR #29394, opened 2022-04-13
Restructure HPackParserTable to not use an inline table, and use a std::vector for its ring buffer instead.

dpcollins-google:
Before this change, it uses 6.2KiB by default, 2/3 of the size of grpc_chttp2_transport. https://screenshot.googleplex.com/6qdcybty2oCsGjG

After this change, it uses 120 bytes https://screenshot.googleplex.com/4xGWKmZZz68VE4F

This class uses up substantial amounts of memory, as it is allocated on a per-stream basis. This should reduce the size of grpc_chttp2_stream substantially

It should cause a significant memory reduction, as the HPack extension table should almost never be fully populated in terms of number of entries, especially since common headers already exist in a static table and entries are highly likely to take more memory than the absolute minimum.

See cl/441088558 for more internal context
@ctiller

### DIFF
--- a/src/core/ext/transport/chttp2/transport/hpack_parser_table.h
+++ b/src/core/ext/transport/chttp2/transport/hpack_parser_table.h
@@ -97,8 +97,8 @@ class HPackTable {
     uint32_t first_entry_ = 0;
     // How many entries are in the table.
     uint32_t num_entries_ = 0;
-    // Maximum number of entries we could possibly fit in the table, given defined
-    // overheads.
+    // Maximum number of entries we could possibly fit in the table, given
+    // defined overheads.
     uint32_t max_entries_ = hpack_constants::kInitialTableEntries;
 
     std::vector<Memento> entries_;
--- a/src/core/ext/transport/chttp2/transport/hpack_parser_table.h
+++ b/src/core/ext/transport/chttp2/transport/hpack_parser_table.h
@@ -30,40 +30,6 @@
 
 namespace grpc_core {
 
-using Memento = ParsedMetadata<grpc_metadata_batch>;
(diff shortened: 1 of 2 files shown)


ctiller (approved):

src/core/ext/transport/chttp2/transport/hpack_parser_table.h
```diff
@@ -30,6 +30,40 @@
 
 namespace grpc_core {
 
+using Memento = ParsedMetadata<grpc_metadata_batch>;
```
ctiller, line 33: This is probably too generic of a name to leave at the top level of the namespace.

Since we're having some build problems anyway, let's remove it and make `MementoRingBuffer` a child class of `HPackTable`.

Build problem:
```
var/local/git/grpc/src/core/ext/transport/chttp2/transport/hpack_parser_table.h:79:9: error: declaration of ‘using Memento = using Memento = class grpc_core::ParsedMetadata<grpc_metadata_batch>’ changes meaning of ‘Memento’ [-fpermissive]
   79 |   using Memento = Memento;
      |         ^~~~~~~
/var/local/git/grpc/src/core/ext/transport/chttp2/transport/hpack_parser_table.h:33:7: note: ‘Memento’ declared here as ‘using Memento = class grpc_core::ParsedMetadata<grpc_metadata_batch>’
   33 | using Memento = ParsedMetadata<grpc_metadata_batch>;
      |       ^~~~~~~
```

dpcollins-google: Fixed

ctiller (approved):

PR #29396, opened 2022-04-13
[PSM interop] Add a magical wait before configuring and querying

lidizheng:
As title.

Fixes b/228743575.


ejona86:
I don't think we need any sleep when changing configuration. We always change away from the previous method, so any results from the old configuration are ignored. The only possibility of race here (other than completely busted client) is if RPCs from two subtests ago (that started slightly before _SAMPLE_DURATION ago) are still running. In that case, I think I'd suggest increasing _SAMPLE_DURATION instead.

lidizheng:
Discussed offline, with more data points, we found in the past 10 days, the flake happened 4 times (40%). All of them are off-by-one. This might be: An issue with interop client related to configuration tear or status code counting.

If the interop client can not be fixed, we will increase the sample duration (which will be a more acceptable form of this PR).

lidizheng:
Superseded by https://github.com/grpc/grpc/pull/29423

PR #29397, opened 2022-04-13
HTTP: Conditionally allow PUT requests

yashykt:
Builds on #29381 (which allows gRPC Core servers to handle PUT requests). We don't want to expose this ability unconditionally so guarding it with a global.

Also added RBAC tests to make sure it does the right thing for PUT requests. (Needed to add a channel arg for testing purposes to indicate that PUT requests need to be sent.)


src/core/ext/filters/http/server/http_server_filter.h
```diff
@@ -26,4 +26,9 @@
 /* Processes metadata on the server side for HTTP2 transports */
 extern const grpc_channel_filter grpc_http_server_filter;
 
+namespace grpc_core {
+// Temporary code that allows servers to accept PUT requests.
```
veblush, line 30: Can you make this more discouraging to use? We can put warnings in the comment (something like this will be removed)?

veblush (approved):
Thanks for the PR!

src/core/ext/filters/http/server/http_server_filter.h
yashykt: Done!

src/core/ext/filters/http/client/http_client_filter.cc
```diff
@@ -125,17 +129,23 @@ ArenaPromise<ServerMetadataHandle> HttpClientFilter::MakeCallPromise(
...
 
 absl::StatusOr<HttpClientFilter> HttpClientFilter::Create(ChannelArgs args,
                                                           ChannelFilter::Args) {
   auto* transport = args.GetObject<grpc_transport>();
   if (transport == nullptr) {
     return absl::InvalidArgumentError("HttpClientFilter needs a transport");
   }
-  return HttpClientFilter(SchemeFromArgs(args),
-                          UserAgentFromArgs(args, transport->vtable->name));
+  absl::optional<int> use_put_requests =
```
markdroth, line 144: This can be:

```
bool use_put_requests =
    args.GetInt(GRPC_ARG_TEST_ONLY_USE_PUT_REQUESTS).value_or(false);
```

You can probably even just avoid the local variable and inline this expression in the `HttpClientFilter` instantiation below.

src/core/ext/filters/http/server/http_server_filter.h
```diff
@@ -26,4 +26,10 @@
 /* Processes metadata on the server side for HTTP2 transports */
 extern const grpc_channel_filter grpc_http_server_filter;
 
+namespace grpc_core {
+// Temporary code that allows servers to accept PUT requests. DO NOT USE WITHOUT
+// PERMISSION.
+void InternalOnlyAllowPutRequests();
```
markdroth, line 32: Just to be safe, let's call this something like `InternalOnlyDoNotUseUnlessYouHavePermissionFromGrpcTeamAllowPutRequests()`. :)

markdroth (approved):
Just a couple of minor comments, otherwise this looks great!

src/core/ext/filters/http/server/http_server_filter.h
ctiller: ...AllowBrokenPutRequests...

Broken being load bearing

src/core/ext/filters/http/client/http_client_filter.cc
yashykt: Done!

src/core/ext/filters/http/server/http_server_filter.h
yashykt: Done!

PR #29404, opened 2022-04-14
Fixing RouteGuide's C++ Reactor example.

nicolasnoble:
The current method involves locking and unlocking a mutex from different threads, which isn't allowed. Changing the strategy a bit to address this.


nicolasnoble:
Apparently @vjpai is still a member, so, lol.

examples/cpp/route_guide/route_guide_callback_server.cc
```diff
@@ -201,25 +201,27 @@ class RouteGuideImpl final : public RouteGuide::CallbackService {
...
-            NextWrite();
-          });
+          // Unlike the other example in this directory that's not using
+          // the reactor pattern, we can't grab a local lock to secure the
+          // access to the notes vector, because the reactor will most likely
+          // make us jump threads, so we'll have to use a different locking
+          // strategy. We'll grab the lock locally to build a copy of the
+          // list of nodes we're going to send, then we'll grab the lock
+          // again to append the received note to the existing vector.
+          mu_->lock();
```
vjpai, line 213: Since it's now structured this way, why not a lock guard?

examples/cpp/route_guide/route_guide_callback_server.cc
```diff
@@ -228,25 +230,21 @@ class RouteGuideImpl final : public RouteGuide::CallbackService {
...
+        if (notes_iterator_ != to_send_notes_.end()) {
+          StartWrite(&*notes_iterator_);
           notes_iterator_++;
-          if (n.location().latitude() == note_.location().latitude() &&
-              n.location().longitude() == note_.location().longitude()) {
-            StartWrite(&n);
-            return;
-          }
+        } else {
+          mu_->lock();
```
vjpai, line 237: Same about lock guard...

examples/cpp/route_guide/route_guide_callback_server.cc
vjpai: BTW one of the side reasons for this example was to use a thread inside a callback since we should have some example of what to do with long blocking ops. So even if this code wasn't the right one for doing so, maybe there should be one somewhere.

examples/cpp/route_guide/route_guide_callback_server.cc
```diff
@@ -228,25 +230,21 @@ class RouteGuideImpl final : public RouteGuide::CallbackService {
...
+          StartRead(&note_);
         }
-        // Didn't write anything, so all done with this note
-        received_notes_->push_back(note_);
-        mu_->unlock();
-        StartRead(&note_);
       }
       RouteNote note_;
       std::mutex* mu_;
       std::vector<RouteNote>* received_notes_;
```
vjpai, line 245: In general these feels like a back way to build this project up, shouldn't it be more like a map or something so that you're not going through the whole thing  every time since you only have to send the notes associated with a particular point?

vjpai (changes requested):

vjpai:
> Apparently @vjpai is still a member, so, lol.

unlimited power star trek gif

vjpai:
> Apparently @vjpai is still a member, so, lol.

pulled back in

examples/cpp/route_guide/route_guide_callback_server.cc
nicolasnoble: If we really wanted to shoehorn a thread there, we could instead go with a semaphore, I guess, instead of a mutex, to do the same as before? But then we're not going to be able to use a nice std construct because they've been introduced in C++20.

As for the lock guard, it may be relevant to avoid keeping the lock for the call to `NextWrite` (or, well, using a scope here... but that's maybe a bit exaggerated imho... dunno)

nicolasnoble:
> > Apparently @vjpai is still a member, so, lol.
> 
> unlimited power star trek gif

Sadly, you don't have any write access, so, no, not that much power. But you're still part of the org as a reader, which makes it possible to tag you, so there's that going for you.

examples/cpp/route_guide/route_guide_callback_server.cc
nicolasnoble: That's... true, I guess. I didn't necessarily want to rewrite the whole thing tho, only fix the incorrect mutex usage :)

vjpai:
> > > Apparently @vjpai is still a member, so, lol.
> > 
> > 
> > unlimited power star trek gif
> 
> Sadly, you don't have any write access, so, no, not that much power. But you're still part of the org as a reader, which makes it possible to tag you, so there's that going for you.


limited power

examples/cpp/route_guide/route_guide_callback_server.cc
```diff
@@ -228,25 +230,21 @@ class RouteGuideImpl final : public RouteGuide::CallbackService {
...
+      std::vector<RouteNote> to_send_notes_;
```
yashykt, line 246: let's add ABSL_GUARDED_BY notations

yashykt: (wherever applicable)

nicolasnoble: Ah, we've polluted the examples with absl already. So that's doable, yes.

nicolasnoble: Right so I had to also switch to using `absl::Mutex` instead of `std::mutex`, since the annotation wouldn't work otherwise.

Since it's only supposed to be an example, I think it's fine to use `absl::Mutex` directly without our typical abstraction layers around it.

yashykt (approved):

yashykt:
You would need to add an exception to examples in check_absl_mutex.sh

nicolasnoble:
Right. We probably don't want to cover the `examples` folder at all.

nicolasnoble:
Those errors aren't related to the example (we're not even building the examples in our tests...). Merging.

PR #29408, opened 2022-04-14
xds_end2end_test: Fix race in security tests

yashykt:
Fix b/226178246

The TLS test infrastructure uses global pointers to keep store the certificate data maps that each test uses. The issue is that these would point to local structures that would go out of scope at the end of a test, but the gRPC channel would not necessarily be done with those structures. In the failing cases, the certificate provider factory was asked to create a certificate provider (because of a cluster update) after the global pointers were cleared out leading to crashes.

The fix is to let the certificate data maps live beyond the life of the tests with synchronized access.

### DIFF
--- a/test/cpp/end2end/xds/xds_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_end2end_test.cc
@@ -41,6 +41,7 @@
 #include "absl/strings/str_format.h"
 #include "absl/strings/str_join.h"
 #include "absl/strings/str_replace.h"
+#include "absl/synchronization/mutex.h"
 #include "absl/types/optional.h"
 
 #include <grpc/grpc.h>
@@ -155,6 +156,22 @@ class FakeCertificateProvider final : public grpc_tls_certificate_provider {
   };
 
   using CertDataMap = std::map<std::string /*cert_name */, CertData>;
+  class CertDataMapWrapper {
+   public:
+    CertDataMap Get() {
+      absl::MutexLock lock(&mu_);
+      return cert_data_map_;
+    }
+
+    void Set(CertDataMap data) {
+      absl::MutexLock lock(&mu_);
+      cert_data_map_ = std::move(data);
+    }
+
+   private:
+    absl::Mutex mu_;
+    CertDataMap cert_data_map_ ABSL_GUARDED_BY(mu_);
+  };
 
   explicit FakeCertificateProvider(CertDataMap cert_data_map)
       : distributor_(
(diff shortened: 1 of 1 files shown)


nicolasnoble (approved):

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -6071,11 +6090,7 @@ class XdsSecurityTest : public XdsEnd2endTest {
     balancer_->ads_service()->SetEdsResource(BuildEdsResource(args));
   }
 
-  void TearDown() override {
-    g_fake1_cert_data_map = nullptr;
-    g_fake2_cert_data_map = nullptr;
-    XdsEnd2endTest::TearDown();
-  }
+  void TearDown() override { XdsEnd2endTest::TearDown(); }
```
markdroth, line 6093: This override isn't needed at all anymore if all it's going to do is delegate to the base class.

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -11218,12 +11061,16 @@ int main(int argc, char** argv) {
   // Workaround Apple CFStream bug
   gpr_setenv("grpc_cfstream", "0");
 #endif
+  grpc::testing::g_fake1_cert_data_map =
+      new grpc::testing::FakeCertificateProvider::CertDataMapWrapper();
```
markdroth, line 11065: No need for dynamic allocation here, since you're deleting the objects before this scope ends.  Just declare a local variable and set the global pointer to point to it.

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -6716,92 +6667,71 @@ TEST_P(XdsSecurityTest, TestMtlsConfigurationWithBothPluginsUpdated) {
...
 
 TEST_P(XdsSecurityTest,
        TestMtlsConfigurationWithIdentityCertificateNameUpdate) {
-  FakeCertificateProvider::CertDataMap fake1_cert_map = {
-      {"", {root_cert_, identity_pair_}},
-      {"bad", {bad_root_cert_, bad_identity_pair_}}};
-  g_fake1_cert_data_map = &fake1_cert_map;
+  g_fake1_cert_data_map->Set({{"", {root_cert_, identity_pair_}},
+                              {"bad", {bad_root_cert_, bad_identity_pair_}}});
+  ;
```
markdroth, line 6687: Please remove.

markdroth (commented):
This looks good!  Only a few minor comments.

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -41,6 +41,7 @@
 #include "absl/strings/str_format.h"
 #include "absl/strings/str_join.h"
 #include "absl/strings/str_replace.h"
+#include "absl/synchronization/mutex.h"
```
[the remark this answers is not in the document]
yashykt: It is already being used in other tests. I actually see it being used in src/core/... @veblush, is this problematic?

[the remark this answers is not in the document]
yashykt: @veblush If this is an issue, let me know and I'll replace it

PR #29411, opened 2022-04-15
HTTP: Use channel arg instead of global to allow PUT requests

yashykt:
The global variable approach did not work so well. Switching to channel args.


ananda1066 (approved):

test/cpp/end2end/xds/xds_end2end_test_lib.cc
```diff
@@ -175,6 +176,11 @@ void XdsEnd2endTest::ServerThread::Serve(grpc_core::Mutex* mu,
     builder.experimental().set_drain_grace_time(
         test_obj_->xds_drain_grace_time_ms_);
     builder.AddListeningPort(server_address, Credentials());
+    // Allow gRPC Core's HTTP server to accept PUT requests for testing
+    // purposes.
+    builder.AddChannelArgument(
+        GRPC_ARG_INTERNAL_ONLY_DO_NOT_USE_UNLESS_YOU_HAVE_PERMISSION_FROM_ALLOW_GRPC_TEAM_ALLOW_BROKEN_PUT_REQUESTS,
```
markdroth, line 182: Instead of hard-coding this here, how about adding a parameter to `CreateBackends()` that will enable this, so that we can use it only in the tests that need it?

src/core/ext/filters/http/server/http_server_filter.h
```diff
@@ -26,10 +26,10 @@
...
 extern const grpc_channel_filter grpc_http_server_filter;
 
-namespace grpc_core {
-// Temporary code that allows servers to accept PUT requests. DO NOT USE WITHOUT
-// PERMISSION.
-void InternalOnlyDoNotUseUnlessYouHavePermissionFromGrpcTeamAllowBrokenPutRequests();
-}  // namespace grpc_core
+// A Temporary channel arg that allows servers to accept PUT requests. DO NOT
+// USE WITHOUT PERMISSION.
+#define GRPC_ARG_INTERNAL_ONLY_DO_NOT_USE_UNLESS_YOU_HAVE_PERMISSION_FROM_ALLOW_GRPC_TEAM_ALLOW_BROKEN_PUT_REQUESTS \
```
markdroth, line 31: We'll probably need some visibility hook on the `grpc_http_filters` BUILD target to allow us to control visibility internally.

ananda1066, line 31: Btw there's a typo in the channel arg - it has the word ALLOW twice.

yashykt: @ananda1066 Can you send a PR/cl for this please?

yashykt: Done! Thanks!

test/cpp/end2end/xds/xds_end2end_test_lib.cc
yashykt: Done!

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -7160,7 +7154,7 @@ class XdsServerSecurityTest : public XdsEnd2endTest {
     builder.AddCertificateProviderPlugin("file_plugin", "file_watcher",
                                          absl::StrJoin(fields, ",\n"));
     InitClient(builder);
-    CreateBackends(1, /*xds_enabled=*/true);
+    CreateBackends(1, /*xds_enabled=*/true, /*allow_put_requests=*/true);
```
markdroth, line 7157: I think we need this only for the RBAC tests, not the other security tests, right?

src/core/ext/filters/http/server/http_server_filter.h
markdroth: I think the way we do this is by adding a visibility tag like the following to the BUILD target:

https://github.com/grpc/grpc/blob/55e870dae61ead220e0c74213410d90f8d21179f/BUILD#L715

We then provide a definition for that tag here:

https://github.com/grpc/grpc/blob/55e870dae61ead220e0c74213410d90f8d21179f/bazel/grpc_build_system.bzl#L103

And we can provide a different definition for the same tag internally.

We should probably add the OSS pieces here as part of this PR.

yashykt: done!

test/cpp/end2end/xds/xds_end2end_test.cc
yashykt: Done!

test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -232,10 +232,12 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType> {
 
     // If use_xds_enabled_server is true, the server will use xDS.
     explicit ServerThread(XdsEnd2endTest* test_obj,
-                          bool use_xds_enabled_server = false)
+                          bool use_xds_enabled_server = false,
+                          bool allow_put_requests = false)
```
markdroth, line 236: Seems like we no longer need this parameter here, in `CreateBackends()`, in `CreateAndStartBackends()`, or in the `BackendServerThread` ctor.

markdroth (approved):
LGTM, with one remaining comment.

markdroth:
Why did you change the repo manager as part of this PR?  Was that intentional?

yashykt:
ah, my master branch got mixed up with https://github.com/grpc/grpc/pull/29435

So, when I merged master into this branch, that commit ended up getting treated as a separate change.

yashykt:
It looks like github did the right thing on merging the PR though, and noticed that change already exists on master.

PR #29422, opened 2022-04-15
ring_hash: trigger internal connection attemps on only one subchannel at a time

markdroth:
I believe this fixes the bug we talked about where C-core would exponentially increase the number of subchannels that it was connecting to at any one time.

I still need to write a test for this.  I think I have an idea of how to do that, but I wanted to send out the code first for an early look.

CC @apolcyn

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
@@ -640,6 +640,13 @@ void RingHash::RingHashSubchannelList::UpdateRingHashConnectivityStateLocked(
   if (start_connection_attempt &&
       !internally_triggered_connection_index_.has_value()) {
     size_t next_index = (index + 1) % num_subchannels();
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_ring_hash_trace)) {
+      gpr_log(GPR_INFO,
+              "[RH %p] triggering internal connection attempt for subchannel "
+              "%p, subchannel_list %p (index %" PRIuPTR " of %" PRIuPTR ")",
+              p, subchannel(next_index)->subchannel(), this, next_index,
+              num_subchannels());
+    }
     internally_triggered_connection_index_ = next_index;
     subchannel(next_index)->subchannel()->AttemptToConnect();
   }
--- a/src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
@@ -148,11 +148,6 @@ class RingHash : public LoadBalancingPolicy {
     void UpdateConnectivityStateLocked(
         grpc_connectivity_state connectivity_state);
 
-    void TriggerInternalConnectionAttempt() {
-      internal_connection_attempt_pending_ = true;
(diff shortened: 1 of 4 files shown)


src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
```diff
@@ -701,11 +724,20 @@ void RingHash::RingHashSubchannelData::ProcessConnectivityChangeLocked(
     }
     p->channel_control_helper()->RequestReresolution();
   }
+  // Update bookkeeping if we were finished an internally triggered
+  // connection attempt.
+  bool internal_connection_attempt_complete =
+      internal_connection_attempt_pending_ &&
```
ejona86, line 730: Hmm... I thought `internal_connection_attempt_pending_` was going to be tied to be specific to an individual subchannel when we were discussing. With the current code, if `internal_connection_attempt_pending_ = true` and the pending connect is on a TF subchannel, and there are 5 subchannels currently CONNECTING because of picks, then won't `internal_connection_attempt_pending_` get cleared as soon as one of the pick-triggered subchannels goes TF? So it will perpetuate the 5 concurrent connection requests. `internal_connection_attempt_pending_` here looks equivalent to only requesting a connection if `connectivity_state == TF`, since this line is equivalent to `bool internal_connection_attempt_complete = (connectivity_state != GRPC_CHANNEL_CONNECTING);`.

I think that is resolved by `int internal_connection_attempt_pending_` where it holds the index of the subchannel that has a queued connection request. Then the resetting code becomes:
```c
bool internal_connection_attempt_complete =
      internal_connection_attempt_pending_ == Index() &&
      connectivity_state != GRPC_CHANNEL_CONNECTING;
```

markdroth: Note that there are two different fields called `internal_connection_attempt_pending_`.  One is in `RingHashSubchannelData`, which is the bit for the individual subchannels.  The other is in `RingHashSubchannelList`, which is essentially a counter of how many subchannels are currently doing an internally triggered connection attempt -- but the counter never goes above 1, so it's a bool instead of an int.

The idea here is to account for internally triggered connection attempts only; we completely ignore any connection attempts triggered by the picker.  Here in `ProcessConnectivityChangeLocked()`, we set `internal_connection_attempt_complete` only if this individual subchannel was doing an internally triggered connection attempt, and in `UpdateRingHashConnectivityStateLocked()` we start a new internally triggered connection attempt only if the current one ended.

That having been said, I think you're right that I could do this same thing less confusingly by just storing the index of the subchannel currently doing the internally triggered connection attempt.  I've changed the code accordingly.

markdroth:
I've now added a test and verified that the test fails without this fix.

src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
ejona86: Oh, I see the there were the two fields now. Okay, that's actually what I expected to see given our conversation. And it did seem to be valid, though the `internal_connection_attempt_complete` middleman does make it less obvious. I had only suggested what I did because I thought you were avoiding the per-subchannel state, but it does turn out easier to verify.

ejona86 (approved):

PR #29423, opened 2022-04-16
Fix Python xDS Interop Client Time Slicing

gnossen:
This fixes b/228743575 and introduces a unit/local e2e test for the Python xDS interop client and server. I suppose it's somewhat similar to our existing "intra-op" tests.

The issue was that there was previously a chance of sending an RPC using old configuration. If a channel is parked [here](https://github.com/grpc/grpc/blob/50df29bdec53b4c18da6ce57b534d935725a61df/src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py#L304) when [the channel config is updated](https://github.com/grpc/grpc/blob/0922756050bc6eced56ff7c18174f62173899f4f/src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py#L351), it will start a single new RPC before its QPS is set to 0. The solution was to just simplify the critical section.

I also spotted an unprotected access that doesn't seem to be causing _this_ issue, but we should nevertheless fix.

### DIFF
--- a/src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
+++ b/src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
@@ -220,11 +220,6 @@ def _on_rpc_done(rpc_id: int, future: grpc.Future, method: str,
                 _global_rpcs_succeeded[method] += 1
         else:
             with _global_lock:
-                _global_rpcs_failed[method] += 1
-        if print_response:
-            if future.code() == grpc.StatusCode.OK:
-                logger.debug("Successful response.")
-            else:
                 logger.debug(f"RPC failed: {call}")
     with _global_lock:
         for watcher in _watchers:
--- a/src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
+++ b/src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
@@ -116,7 +116,6 @@ _global_server = None
 _global_rpcs_started: Mapping[str, int] = collections.defaultdict(int)
 _global_rpcs_succeeded: Mapping[str, int] = collections.defaultdict(int)
 _global_rpcs_failed: Mapping[str, int] = collections.defaultdict(int)
-_global_config_condition = threading.Condition()
 
 # Mapping[method, Mapping[status_code, count]]
 _global_rpc_statuses: Mapping[str, Mapping[int, int]] = collections.defaultdict(
@@ -221,6 +220,11 @@ def _on_rpc_done(rpc_id: int, future: grpc.Future, method: str,
                 _global_rpcs_succeeded[method] += 1
         else:
             with _global_lock:
+                _global_rpcs_failed[method] += 1
+        if print_response:
+            if future.code() == grpc.StatusCode.OK:
+                logger.debug("Successful response.")
+            else:
                 logger.debug(f"RPC failed: {call}")
     with _global_lock:
         for watcher in _watchers:
@@ -256,6 +260,9 @@ class _ChannelConfiguration:
     def __init__(self, method: str, metadata: Sequence[Tuple[str, str]],
                  qps: int, server: str, rpc_timeout_sec: int,
                  print_response: bool, secure_mode: bool):
+        # condition is signalled when a change is made to the config.
+        self.condition = threading.Condition()
+
         self.method = method
         self.metadata = metadata
         self.qps = qps
@@ -267,7 +274,7 @@ class _ChannelConfiguration:
 
 def _run_single_channel(config: _ChannelConfiguration) -> None:
     global _global_rpc_id  # pylint: disable=global-statement
-    with _global_config_condition:
+    with config.condition:
         server = config.server
     channel = None
     if config.secure_mode:
@@ -279,11 +286,10 @@ def _run_single_channel(config: _ChannelConfiguration) -> None:
     with channel:
         stub = test_pb2_grpc.TestServiceStub(channel)
         futures: Dict[int, Tuple[grpc.Future, str]] = {}
-        print_response = False
         while not _stop_event.is_set():
-            with _global_config_condition:
+            with config.condition:
                 if config.qps == 0:
-                    _global_config_condition.wait(
+                    config.condition.wait(
                         timeout=_CONFIG_CHANGE_TIMEOUT.total_seconds())
                     continue
                 else:
@@ -298,7 +304,7 @@ def _run_single_channel(config: _ChannelConfiguration) -> None:
                 _start_rpc(config.method, config.metadata, request_id, stub,
                            float(config.rpc_timeout_sec), futures)
                 print_response = config.print_response
-            _remove_completed_rpcs(futures, print_response)
+            _remove_completed_rpcs(futures, config.print_response)
             logger.debug(f"Currently {len(futures)} in-flight RPCs")
             now = time.time()
             while now < end:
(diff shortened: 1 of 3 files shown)


src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
```diff
@@ -198,7 +198,8 @@ def _on_rpc_done(rpc_id: int, future: grpc.Future, method: str,
                  print_response: bool) -> None:
     exception = future.exception()
     hostname = ""
-    _global_rpc_statuses[method][future.code().value[0]] += 1
+    with _global_lock:
+        _global_rpc_statuses[method][future.code().value[0]] += 1
```
lidizheng, line 202: Good finding.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
```diff
@@ -294,18 +295,17 @@ def _run_single_channel(config: _ChannelConfiguration) -> None:
...
+                    _global_rpc_id += 1
+                    _global_rpcs_started[config.method] += 1
+                start = time.time()
+                end = start + duration_per_query
                 _start_rpc(config.method, config.metadata, request_id, stub,
                            float(config.rpc_timeout_sec), futures)
-            with config.condition:
-                _remove_completed_rpcs(futures, config.print_response)
+                print_response = config.print_response
+            _remove_completed_rpcs(futures, config.print_response)
```
lidizheng, line 308: In this PR, we are making sure the creation of RPCs won't race with the configuration updates. Have we considered in-flight RPCs? In `_remove_completed_rpcs`, we are treating the RPCs with old spec and the RPCs with new spec the same way. The tear might still be a problem for slow-responding RPCs.

There is an alternative of cleaning the in-flight RPCs before making configuration changes. WDYT?

(In the test cases this PR add, adding slow-responding RPCs is not that easy, since it requires server-side changes. I'm okay if we want to defer that test case till future.)

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py
```diff
@@ -0,0 +1,148 @@
...
+        delta = {}
+        for _, method_str in _METHODS:
+            after[method_str] = response.stats_per_method[method_str].result[0]
+            delta[method_str] = after[method_str] - before[method_str]
+        logging.info("Delta: %s", delta)
+        for _, method_str in _METHODS:
+            if method_str == target_method_str:
+                assert delta[method_str] > 0, delta
+            else:
+                assert delta[method_str] == 0, delta
```
lidizheng, line 106: optional: We can also check the status code distribution and other fields in the response. This is the first unit test (end2end unit test) around the Python xDS interop client, we can use it to validate the counting logic is correct. In case, in future, someone else (like me) want to change the statistic logic.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def _dump_streams(process_name: str, stdout: tempfile.TemporaryFile,
+                  stderr: tempfile.TemporaryFile):
+    _dump_stream(process_name, "stdout", stdout)
+    _dump_stream(process_name, "stderr", stderr)
+    sys.stderr.write(f"End {process_name} output.\n")
+
+
+def _test_client(server_port: int, stats_port: int, qps: int,
```
lidizheng, line 77: optional: since we are asserting in this function, we could make it part of `XdsInteropClientTest` to access the `unittest` assertions.

lidizheng (commented):
The unit test (or e2e test) looks great and not easy to build! Thanks for working on it!

Also, we might want to schedule a few Kokoro runs of Authz tests to ensure we fixed it.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
gnossen: > In _remove_completed_rpcs, we are treating the RPCs with old spec and the RPCs with new spec the same way. The tear might still be a problem for slow-responding RPCs.
> There is an alternative of cleaning the in-flight RPCs before making configuration changes. WDYT?

I considered this as a potential cause of the problem and actually implemented the cancellation you're suggesting as in an attempt to fix the bug. But in practice, this is not a problem because all RPCs are created with a timeout defined by the test harness. If we cancel RPCs, then we'll end up with another bad situation -- RPCs with cancellation status codes in the accumulated stats.

If we decide to add this cancellation behavior, it should be added to the spec and to all other test clients.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py
```diff
@@ -0,0 +1,138 @@
...
+_SUBPROCESS_TIMEOUT_SECONDS = 2
+
+
+@contextlib.contextmanager
+def _start_python_with_args(
+    file: str, args: List[str]
+) -> Tuple[subprocess.Popen, tempfile.TemporaryFile, tempfile.TemporaryFile]:
+    with tempfile.TemporaryFile(mode='r') as stdout:
+        with tempfile.TemporaryFile(mode='r') as stderr:
+            proc = subprocess.Popen((sys.executable, file) + tuple(args),
```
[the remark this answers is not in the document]
gnossen: I see no reason to run an imported version of this. This directory is already excluded from google3. I was worried about the same thing, so I checked ahead of time.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py
gnossen: Done.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py
gnossen: Done.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+class XdsInteropClientTest(unittest.TestCase):
+
+    def test_configure_consistenc(self):
+        _, server_port, socket = framework_common.get_socket()
+
+        with _start_python_with_args(
+                _SERVER_PATH,
+            [f"--port={server_port}", f"--maintenance_port={server_port}"
+            ]) as (server, server_stdout, server_stderr):
```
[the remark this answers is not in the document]
gnossen: Adding a newline didn't do anything. This just seems to be the way YAPF wants it.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
lidizheng: Good point about the cancellation. How about delay the configuration until the futures are cleared? In that case, we won't have unwanted status codes in the stats.

(With current design, the same configuration tear problem might apply to other languages as well.)

gnossen: I think this is out of the scope of this PR and the bug that I'm trying to solve here, so I've created [a tracking issue](https://github.com/grpc/grpc/issues/29442) for this discussion.

lidizheng (approved):

gnossen:
[Authz test run](https://sponge2.corp.google.com/1888a447-02e8-483f-9dd8-68eb6ee02b3f) (in-progress)

gnossen:
Three green runs in a row. Going ahead and merging.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py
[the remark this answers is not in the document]
lidizheng: In google3, `sys.executable` was empty, causing this type of subprocess call hard: https://fusion2.corp.google.com/invocations/e33af7b5-bece-4878-a58e-dc5399ae7c9f/targets/%2F%2Fthird_party%2Fpy%2Fgrpc:tests_py3_only%2Finterop%2Fxds_interop_client_test_opensource/log

We might want to revert and cherry-pick. See https://github.com/grpc/grpc/pull/29458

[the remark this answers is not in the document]
gnossen: This wasn't supposed to be picked up by blaze. I did a blaze query for targets under the tests_py3_only directory and saw none. I'll make a google3 change to ensure that this doesn't get picked up, as intended.

PR #29424, opened 2022-04-16
Prevent direct usage of absl::Mutex

yashykt:
Context - https://b.corp.google.com/issues/186685878
absl::Mutex is probably still not safe to use directly, so adding a script to make sure we don't regress till this is fixed.


veblush (approved):
Thank you for this PR!

markdroth (approved):

PR #29428, opened 2022-04-18
subchannel: report IDLE upon existing connection failure and after backoff interval

markdroth:
Prior to this PR, subchannels start in state IDLE, but they never return to that state.  This PR changes subchannels to report IDLE in two additional cases.

The first case is when a subchannel was successfully connected (i.e., in state READY) and the connection fails (e.g., due to a GOAWAY).  Previously, we reported TRANSIENT_FAILURE in this case; we now report IDLE instead.  This avoids failing RPCs in the case where round_robin has only a single address and that server sends a GOAWAY due to (e.g.) max connection age.  In that case, we want to queue RPCs until the first reconnection attempt completes, after which we will transition to either READY or TRANSIENT_FAILURE, depending on whether the attempt was successful.

The second case is when a connection attempt fails, after the backoff delay elapses.  Previously, when a connection attempt failed, we would transition into state TRANSIENT_FAILURE, and we would stay in that state until the next connection attempt started, with no indication of whether or not the backoff delay had elapsed; when the LB policy requested a connection attempt, we would either start the backoff timer (remaining in TRANSIENT_FAILURE until the timer fires, at which point we report CONNECTING) or immediately trigger the connection attempt (reporting CONNECTING), depending on whether backoff delay was needed.  Now, when a connection attempt fails, we report TRANSIENT_FAILURE and immediately start the backoff timer.  When the timer fires, we check whether the LB policy has already requested a connection attempt, in which case we report CONNECTING and immediately trigger the connection attempt, or else we report IDLE.  This provides a signal to the LB policy about when the backoff time has elapsed.

This brings C-core's subchannel connectivity state semantics much closer to those of Java and Go.  It also significantly simplifies the state machine in the subchannel code.


ejona86 (approved):
I actually think I understand the changes. Although if I missed something, it'd probably be pick_first, since I'm less familiar the details there. But the changes still made sense and I don't see any obvious holes.

markdroth:
@yashykt The failures in `//test/core/transport/chttp2:too_many_pings_test` look very similar to the ones I reported in b/229488257: it looks like we're not getting a transparent retry because the transport isn't setting the StreamNetworkState attribute in trailing metadata.  It's not clear to me why this PR is triggering that behavior, but since the symptom looks similar, I'm hoping you can look into that piece.  Thanks!

yashykt:
Hmm, on a quick look at the too_many_pings_test failures, the logs suggest that the keepalive time is not being throttled when a GOAWAY frame with "too_many_pings" is received

markdroth:
Thanks for your help with that, Yash.  Looks like the problem was that the keepalive info was being communicated back up from the transport in the status info, but we weren't passing that status along, since we're now reporting IDLE instead of TF, and status is usually passed only in TF.

I've changed the subchannel code to pass the status along with IDLE for now, since that's a purely internal change, not something visible to the LB policies.  But we'll probably want to find a cleaner way to do this in the long run.

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -308,43 +308,29 @@ class Subchannel::ConnectedSubchannelStateWatcher
...
-                  ? absl::Status(absl::StatusCode::kUnavailable,
-                                 "Subchannel has disconnected.")
-                  : status);
-          c->backoff_begun_ = false;
-          c->backoff_.Reset();
-        }
-        break;
+    // If we're either shutting down or have already seen this
+    // connection failure, do nothing.
+    // TODO(yashkt): Why does the transport sometimes report TF and then
```
yashykt, line 313: discussed on chat

yashykt (approved):
It looks correct but I'm not very familiar with this code, so I'm not confident on the edge cases. I'd be more confident if we had tests.

src/core/ext/filters/client_channel/subchannel.cc
markdroth: I've updated the comment as per our conversation.

markdroth (commented):
I've added a test that I think actually does a decent job of covering this case.  It relies on timing, which is sub-optimal, but at least with our current implementation, it does fail reliably without this PR.

PTAL.

yashykt:
As a check, please try running these tests 100x to check for flakiness

markdroth:
Hmm, and now this seems to have broken the grpclb tests.  Looking into it...

markdroth:
Looks like there was a real bug in grpclb surfaced by this change.  Should be fixed now.

PTAL.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -1420,7 +1445,146 @@ TEST_F(ClientLbEnd2endTest, RoundRobinTransientFailureAtStartup) {
   EXPECT_TRUE(WaitForChannelState(channel.get(), predicate, true));
 }
 
-TEST_F(ClientLbEnd2endTest, RoundRobinSingleReconnect) {
+TEST_F(RoundRobinTest, DoesNotFailRpcsUponDisconnection) {
+  // A connection attempt injector that allows us to control timing.
+  class ConnectionInjector : public ConnectionAttemptInjector {
```
yashykt, line 1450: ooh fancy!

yashykt (approved):

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -201,6 +201,10 @@ void PickFirst::AttemptToConnectUsingLatestUpdateArgsLocked() {
     channel_control_helper()->UpdateState(
         GRPC_CHANNEL_TRANSIENT_FAILURE, status,
         absl::make_unique<TransientFailurePicker>(status));
+    // If there was a previously pending update (which may or may
+    // not have contained the currently selected subchannel), drop
+    // it, so that it doesn't override what we've done here.
+    latest_pending_subchannel_list_.reset();
```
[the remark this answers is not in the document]
markdroth: Yup.  We're already doing the same thing in another case below (line 225), but I just noticed that we were missing it here.

yashykt:
Thanks for adding the tests!

lidizheng:
@markdroth Shall we backport this PR to v1.46.x branch? v1.46.x branch enabled Authz tests, so we might also want to reduce its flakiness.

ejona86:
If authz was still environment-var controlled in v1.46.x then it seems we should just disable the authz tests on that branch. It wasn't ready for normal users in the release so it doesn't need to be tested.

PR #29432, opened 2022-04-18
Add a raise for generate_projects if Python is too old

lidizheng:
Improve the error message.

Related b/229631553.

### DIFF
--- a/tools/buildgen/generate_projects.py
+++ b/tools/buildgen/generate_projects.py
@@ -25,6 +25,9 @@ from typing import Dict, List, Union
 import _utils
 import yaml
 
+if sys.version_info < (3,6):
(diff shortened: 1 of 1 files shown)


tools/buildgen/generate_projects.py
```diff
@@ -25,6 +25,9 @@
 import _utils
 import yaml
 
+if sys.version_info < (3,6):
+    raise RuntimeError('expect Python version > 3.6, but got %s at %s' % (sys.version_info, sys.executable))
```
gnossen, line 29: Nit: `s/expect/Expected`

gnossen (approved):

tools/buildgen/generate_projects.py
lidizheng: Done. Good idea!

PR #29433, opened 2022-04-18
Add a compile Python mechanism for Ubuntu jobs

lidizheng:
Related b/229631553.

The problem is caused by the Python version indeed. Our Kokoro jobs use an image based upon Ubuntu 16, which installs Python3.5 out-of-box.

Why not use deadsnakes/ppa instead of compiling? I attempted this approach, but they only support Ubuntu 18+. They deprecated their support for Ubuntu 16, no pre-compiled packages are available.

Why not use Python 3.10? [Python 3.10 requires OpenSSL > 1.1.1](https://docs.python.org/3/library/ssl.html
). However, our image only have OpenSSL 1.0.2, and the package source don't offer newer versions. This PR intends to introduce a new-enough Python version for our infra code, 3.9 should be a decent choice for several years down the line.


lidizheng:
https://fusion2.corp.google.com/invocations/479fd06e-2385-4b32-93d3-08e862804e97/targets

tools/internal_ci/helper_scripts/prepare_build_linux_rc
```diff
@@ -41,4 +41,29 @@ export DOCKERHUB_ORGANIZATION=grpctesting
...
 
+if [ "${PREPARE_BUILD_INSTALL_DEPS_PYTHON}" == "true" ]
+then
+  # Prepares the build dependencies
+  sudo apt update
+  sudo apt install -y zlib1g-dev libssl-dev jq build-essential libffi-dev wget
+
+  # Downloads and compiles Python 3.9.12
+  pushd /tmp
+  wget -q https://www.python.org/ftp/python/3.9.12/Python-3.9.12.tgz
```
gnossen, line 52: Optional: Copy this to one of our buckets in case this server goes down or the URL changes?

gnossen (approved):
Very nice! Hopefully, this didn't take too long to get it building in CI. That's always a complicated process.

tools/internal_ci/helper_scripts/prepare_build_linux_rc
lidizheng: Good idea! Let me add that.

lidizheng:
Based on the Fusion history of this job, the Python compilation costs us 15-20 minutes (from ~125 to ~140).

lidizheng:
New ad-hoc run passed: https://fusion2.corp.google.com/invocations/db42e631-c22f-4456-b739-92f0c81660b7/targets (105 mins)

tools/internal_ci/helper_scripts/prepare_build_linux_rc
```diff
@@ -41,4 +41,29 @@ export DOCKERHUB_ORGANIZATION=grpctesting
 
 git submodule update --init
 
+if [ "${PREPARE_BUILD_INSTALL_DEPS_PYTHON}" == "true" ]
```
jtattermusch, line 44: We currently don't have the "PREPARE_BUILD_INSTALL_*" section for any of the languages for linux and I don't think we should have one.
For vast majority of our linux builds, we rely on all the necessary dependencies being available in the docker images we have. We only use "PREPARE_BUILD_INSTALL_*" on macos and windows, where we can't use docker.
The problem of these per-language setup snippets is that they 1. add complexity 2. take time (= slow down the builds) 3. increase flakiness (unless a lot of extra care is given).

- Building python from source is slow and we really should not do that at the beginning of any of our linux jobs (there must be a better solution)
- Strictly speaking, the name "PREPARE_BUILD_INSTALL_DEPS_PYTHON" doesn't actually fit your use case - since you're not actually installing dependencies for the grpc-python build, but dependencies for our build/test scripts (in this case for the extract_metadata_from_bazel.py). Since any of these scripts might potentially use type annotations (requires python3.6), it also means that you'd eventually need a slow "build python from source" step for almost all of our linux jobs (and I don't think that's acceptable).

tools/internal_ci/linux/grpc_build_submodule_at_head.sh
```diff
@@ -20,6 +20,7 @@ set -ex
 # change to grpc repo root
 cd $(dirname $0)/../../..
 
+export PREPARE_BUILD_INSTALL_DEPS_PYTHON=true
```
jtattermusch, line 23: as expressed in my other comment, this has actually nothing to do with "python dependencies" and it ends up being an arbitrary hack.

jtattermusch (changes requested):
I don't think this is good solution going forward (see comments).
I understand the desire to avoid reverting https://github.com/grpc/grpc/pull/29254, but if that's the main objective, you could just as well temporarily remove the type annotations from extract_metadata_from_bazel.py (which are nice to have, but strictly speaking optional), which would fix the immediate breakage and would give us time to come up with a proper solution for having a newer python available on our linux workers (this PR isn't what we want AFAIC).

tools/internal_ci/helper_scripts/prepare_build_linux_rc
lidizheng: If customizing the Kokoro image is the practice we want to enforce on Linux jobs, this PR won't make sense.

I do see the conflicts about `PREPARE_BUILD_INSTALL_DEPS_PYTHON`, and open to name changes.

As suggested, we can remove the type annotations to fix the broken tests. I will do that instead and close this PR.

jtattermusch: Thanks for your understanding!

PR #29434, opened 2022-04-18
Remove unmatched grpc_init from *ChannelFromFd methods

drfloob:
These unmatched grpc_init calls were preventing grpc_shutdown from
completing, which notably showed up in the only tests we have that
exercise these methods: `client_interceptors_end2end_tests.cc`.

~These inits appear to be unnecessary, so I removed them. But if I've
missed some reason they're needed, an alternative is to find the right
place to add the corresponding `init_lib.shutdown();` call. I'm not sure
where that would be at this point.~ 

I added matched `init_lib.shutdown()` calls for each fd-based channel creation function, on the off chance these functions are called before any other gRPC surface API function.

See b/175634383

Previously, this test would flake 0.3% of the time (~30 failures of 10,000 runs, confirmed 3 times). The odd bit is that every run - even the passing runs - had shutdown problems. The passing tests would have a gRPC shutdown timeout at 10 seconds because `grpc_shutdown` was never attempted (init refs = 6, one for each of the FD tests). The few flaky runs hung 5 minutes waiting for shutdown, indicating that grpc shutdown was actually occurring ... I'll still have to dig more to figure out why.

However, with this fix, I cannot reproduce the test failures in 20,000 runs, and the passing tests also show gRPC shutting down correctly.

This change reduces test time by 10s for the happy path, and 5 minutes for flaky runs.

@yashykt

### DIFF
--- a/src/cpp/client/create_channel_posix.cc
+++ b/src/cpp/client/create_channel_posix.cc
@@ -34,8 +34,6 @@ class ChannelArguments;
 
 std::shared_ptr<Channel> CreateInsecureChannelFromFd(const std::string& target,
                                                      int fd) {
-  grpc::internal::GrpcLibrary init_lib;
-  init_lib.init();
   grpc_channel_credentials* creds = grpc_insecure_credentials_create();
   auto channel = CreateChannelInternal(
       "", grpc_channel_create_from_fd(target.c_str(), fd, creds, nullptr),
@@ -47,8 +45,6 @@ std::shared_ptr<Channel> CreateInsecureChannelFromFd(const std::string& target,
 
 std::shared_ptr<Channel> CreateCustomInsecureChannelFromFd(
     const std::string& target, int fd, const grpc::ChannelArguments& args) {
-  internal::GrpcLibrary init_lib;
-  init_lib.init();
   grpc_channel_args channel_args;
   args.SetChannelArgs(&channel_args);
   grpc_channel_credentials* creds = grpc_insecure_credentials_create();
@@ -67,8 +63,6 @@ std::shared_ptr<Channel> CreateCustomInsecureChannelWithInterceptorsFromFd(
     std::vector<
         std::unique_ptr<grpc::experimental::ClientInterceptorFactoryInterface>>
         interceptor_creators) {
-  grpc::internal::GrpcLibrary init_lib;
-  init_lib.init();
   grpc_channel_args channel_args;
   args.SetChannelArgs(&channel_args);
   grpc_channel_credentials* creds = grpc_insecure_credentials_create();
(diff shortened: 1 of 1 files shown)


src/cpp/client/create_channel_posix.cc
```diff
@@ -34,8 +34,6 @@ class ChannelArguments;
 
 std::shared_ptr<Channel> CreateInsecureChannelFromFd(const std::string& target,
                                                      int fd) {
-  grpc::internal::GrpcLibrary init_lib;
-  init_lib.init();
   grpc_channel_credentials* creds = grpc_insecure_credentials_create();
   auto channel = CreateChannelInternal(
       "", grpc_channel_create_from_fd(target.c_str(), fd, creds, nullptr),
```
yashykt, line 39: We can add an `init_lib.shutdown();` before the return, and add an comment that that should be safe since `Channel` derives from `GrpcLibraryCodegen` which causes a `grpc_init()` in its constructor.

src/cpp/client/create_channel_posix.cc
```diff
@@ -47,8 +45,6 @@ std::shared_ptr<Channel> CreateInsecureChannelFromFd(const std::string& target,
 
 std::shared_ptr<Channel> CreateCustomInsecureChannelFromFd(
     const std::string& target, int fd, const grpc::ChannelArguments& args) {
-  internal::GrpcLibrary init_lib;
-  init_lib.init();
```
yashykt, line 51: same here

src/cpp/client/create_channel_posix.cc
```diff
@@ -67,8 +63,6 @@ std::shared_ptr<Channel> CreateCustomInsecureChannelWithInterceptorsFromFd(
     std::vector<
         std::unique_ptr<grpc::experimental::ClientInterceptorFactoryInterface>>
         interceptor_creators) {
-  grpc::internal::GrpcLibrary init_lib;
-  init_lib.init();
```
yashykt, line 71: same here

src/cpp/client/create_channel_posix.cc
drfloob: That could work. Is the concern that `InsecureCredentials` construction may require gRPC init?

yashykt: Yes

drfloob: Alright, done.

src/cpp/client/create_channel_posix.cc
drfloob: Done

src/cpp/client/create_channel_posix.cc
drfloob: Done

yashykt (approved):

PR #16083, opened 2018-07-20
Added system roots feature to load roots from OS trust store

tdbhacks:
Added a flag-guarded feature that allows gRPC to load TLS/SSL
roots from the OS trust store. This is the Linux-specific
implementation of such feature.

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/16083)

PR #29436, opened 2022-04-19
FreeBSD system roots implementation

tdbhacks:
This PR is based off [#16083](https://github.com/grpc/grpc/pull/16083), with the implementation being branched from the existing linux-specific logic.

@drfloob

### DIFF
--- a/src/core/lib/security/security_connector/load_system_roots_linux.cc
+++ b/src/core/lib/security/security_connector/load_system_roots_linux.cc
@@ -22,7 +22,7 @@
 
 #include <grpc/slice_buffer.h>
 
-#if defined(GPR_LINUX) || defined(GPR_ANDROID)
+#if defined(GPR_LINUX) || defined(GPR_ANDROID) || defined(GPR_FREEBSD)
 
 #include <dirent.h>
 #include <fcntl.h>
@@ -51,20 +51,26 @@ GPR_GLOBAL_CONFIG_DEFINE_STRING(grpc_system_ssl_roots_dir, "",
 namespace grpc_core {
 namespace {
 
-const char* kLinuxCertFiles[] = {
+#if defined(GPR_LINUX) || defined(GPR_ANDROID)
+const char* kCertFiles[] = {
     "/etc/ssl/certs/ca-certificates.crt", "/etc/pki/tls/certs/ca-bundle.crt",
     "/etc/ssl/ca-bundle.pem", "/etc/pki/tls/cacert.pem",
     "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"};
-const char* kLinuxCertDirectories[] = {
+const char* kCertDirectories[] = {
     "/etc/ssl/certs", "/system/etc/security/cacerts", "/usr/local/share/certs",
     "/etc/pki/tls/certs", "/etc/openssl/certs"};
+#elif defined(GPR_FREEBSD)  // endif GPR_LINUX || GPR_ANDROID
+const char* kCertFiles[] = {"/etc/ssl/cert.pem",
+                            "/usr/local/share/certs/ca-root-nss.crt"};
+const char* kCertDirectories[] = {""};
+#endif                      // GPR_FREEBSD
 
 grpc_slice GetSystemRootCerts() {
   grpc_slice valid_bundle_slice = grpc_empty_slice();
-  size_t num_cert_files_ = GPR_ARRAY_SIZE(kLinuxCertFiles);
+  size_t num_cert_files_ = GPR_ARRAY_SIZE(kCertFiles);
   for (size_t i = 0; i < num_cert_files_; i++) {
     grpc_error_handle error =
-        grpc_load_file(kLinuxCertFiles[i], 1, &valid_bundle_slice);
+        grpc_load_file(kCertFiles[i], 1, &valid_bundle_slice);
     if (error == GRPC_ERROR_NONE) {
       return valid_bundle_slice;
     } else {
@@ -155,8 +161,8 @@ grpc_slice LoadSystemRootCerts() {
     result = GetSystemRootCerts();
   }
   if (GRPC_SLICE_IS_EMPTY(result)) {
-    for (size_t i = 0; i < GPR_ARRAY_SIZE(kLinuxCertDirectories); i++) {
-      result = CreateRootCertsBundle(kLinuxCertDirectories[i]);
+    for (size_t i = 0; i < GPR_ARRAY_SIZE(kCertDirectories); i++) {
+      result = CreateRootCertsBundle(kCertDirectories[i]);
       if (!GRPC_SLICE_IS_EMPTY(result)) {
         break;
       }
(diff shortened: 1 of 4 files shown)


tdbhacks:
Note to reviewers: the related test now runs on both linux and freebsd, but is still named `linux_system_roots_test`. Happy to rename it if that's appropriate, naming suggestions are very welcome!

drfloob:
Thanks for the contribution, @tdbhacks. It's a bit tricky to evaluate since we don't have any test infrastructure for FreeBSD - the platform is entirely community-supported, and we don't have a FreeBSD-community point of contact for gRPC. We're generally open to accepting community patches like this, with the understanding they may break without continued community support (and we won't know when they do break).

This looks useful at first glance, but nearly identical to the linux version (only the paths differ, yes?). I think it would be better to unify the two.

@ZhenLian could you take a look at this as well? Thanks.

tdbhacks:
@drfloob thanks for taking a look!

Indeed, only the paths differ (other than the logic to check directories in `LoadSystemRootCerts()` being dropped, since we only use files here). Wasn't sure if this was worth an independent implementation, I agree that it's essentially duplicate code so I'm open to merging with the linux implementation. We could either just add the FreeBSD paths to the list of linux paths, or we could choose the right set of paths based on platform macros. The latter seems a little less clean, but I don't feel strongly so either option works for me :)

Side note: I'd like to do this for MacOS too, which at first glance seems to use `/etc/ssl/cert.pem` too, either in a different PR or here. Wdyt?

drfloob:
> We could either just add the FreeBSD paths to the list of linux paths, or we could choose the right set of paths based on platform macros.

For security reasons, I think we'd want to be explicit about using the appropriate canonical paths for specific platforms.

Given that Mac is a supported platform in gRPC, I'd recommend a separate PR. There are separate concerns there, and potentially reasons to reject or roll back commits that don't exist for FreeBSD.

tdbhacks:
Sending a separate PR for Mac SGTM.

In the meantime, I've merged the FreeBSD implementation as suggested, it's now choosing the right set of paths based on the platform.

src/core/lib/security/security_connector/load_system_roots_linux.cc
```diff
@@ -51,20 +51,27 @@ GPR_GLOBAL_CONFIG_DEFINE_STRING(grpc_system_ssl_roots_dir, "",
...
+const char* kCertFiles[] = {
     "/etc/ssl/certs/ca-certificates.crt", "/etc/pki/tls/certs/ca-bundle.crt",
     "/etc/ssl/ca-bundle.pem", "/etc/pki/tls/cacert.pem",
     "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"};
-const char* kLinuxCertDirectories[] = {
+const char* kCertDirectories[] = {
     "/etc/ssl/certs", "/system/etc/security/cacerts", "/usr/local/share/certs",
     "/etc/pki/tls/certs", "/etc/openssl/certs"};
+#endif // GPR_LINUX || GPR_ANDROID
+#ifdef GPR_FREEBSD
```
ZhenLian, line 63: What's the difference between using `ifdef GPR_FREEBSD` and `if defined(GPR_FREEBSD)`? 
We probably want to keep the syntax consistent.

ZhenLian (changes requested):
Thanks for making the contributions and sorry about the delay! I was just back from vacation.

The change overall looks good to me, except one minor thing. Let me open the pre-submit tests.
BTW, where is the change related to Mac system? Do you plan to work on it after this is merged?

src/core/lib/security/security_connector/load_system_roots_linux.cc
tdbhacks: Thanks for taking a look! So `ifdef VAL` and `if defined(VAL)` are equivalent, the reason why `if defined(VAL)` is used above is that it allows multiple conditions to be combined into a single if statement (i.e. `ifdef GPR_LINUX || GPR_FREEBSD` would not work). See https://stackoverflow.com/questions/39290019/is-if-defined-macro-equivalent-to-ifdef-macro

Let me know if you'd still like me to change it, happy to do it if you think it's cleaner that way.

re: Mac PR: I haven't sent it yet because I'm wondering whether we should just rename this file, remove the "_linux" suffix, and add the Mac-specific path in here too, for the same reason why @drfloob suggested merging the FreeBSD and Linux implementations. That way we would only have this implementation for all platforms where we support loading system roots, and the fallback implementation for all other platforms. WDYT? If that sounds good to you I'll send in a follow-up PR once this is merged (unless you think it's fine to add both here).

ZhenLian: I see, thanks for the explanation! Would you mind change this to `if defined(VAL)` to make it consistent?

ZhenLian (approved):

ZhenLian:
Let me re-run the test. In the meantime, you could run `clang-format -i THE_CHANGED_FILE` to see anything is changed.

tdbhacks:
Thanks, re-ran clang-format and it appears to have done the trick, not sure why it didn't work last time. I've also changed `ifdef` to `if defined` as suggested, and squashed all my commits.

I'll send a follow-up PR once this is merged to do the same for MacOS. I believe @bdhess spoke with @veblush a few weeks back and he was okay with adding a MacOS implementation based on the same logic and using `/etc/ssl/cert.pem`.

src/core/lib/security/security_connector/load_system_roots_linux.cc
```diff
@@ -51,20 +51,27 @@ GPR_GLOBAL_CONFIG_DEFINE_STRING(grpc_system_ssl_roots_dir, "",
...
+#if defined(GPR_LINUX) || defined(GPR_ANDROID)
+const char* kCertFiles[] = {
     "/etc/ssl/certs/ca-certificates.crt", "/etc/pki/tls/certs/ca-bundle.crt",
     "/etc/ssl/ca-bundle.pem", "/etc/pki/tls/cacert.pem",
     "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"};
-const char* kLinuxCertDirectories[] = {
+const char* kCertDirectories[] = {
     "/etc/ssl/certs", "/system/etc/security/cacerts", "/usr/local/share/certs",
     "/etc/pki/tls/certs", "/etc/openssl/certs"};
+#endif  // GPR_LINUX || GPR_ANDROID
```
markdroth, line 62: Instead of `#endif` and then another independent `#if`, please use `#elif`.  That way, we know for sure that only one of these two definitions will be used.

markdroth (commented):
Just one minor comment, otherwise this looks great!

src/core/lib/security/security_connector/load_system_roots_linux.cc
tdbhacks: Good point, changed to `#elif`.

markdroth:
Test failures look unrelated, so I'll go ahead and merge this.

PR #29446, opened 2022-04-19
Disable layering check for Objective-C

googlewalt:
These targets are currently not layering check clean.  Disabling it
ensures that they will continue to build even when we enable layering
check for Objective-C internally.

@drfloob


dennycd (approved):
prior PR at https://github.com/grpc/grpc/pull/29375

PR #29448, opened 2022-04-19
Add a service to dynamically configure the xDS test server

murgatroid99:
This adds a configuration service to the xDS interop test server, analogous to the service that is already defined for the client. The initial purpose of this service is to enable outlier detection tests in which some servers always respond with errors.

### DIFF
--- a/doc/xds-test-descriptions.md
+++ b/doc/xds-test-descriptions.md
@@ -12,7 +12,7 @@ Server should accept these arguments:
 *   --port=PORT
     *   The port the test server will run on.
 *   --maintenance_port=PORT
-    *   The port for the maintenance server running health, channelz, and admin(CSDS) services.
+    *   The port for the maintenance server running health, channelz, configuration, and admin(CSDS) services.
 *   --secure_mode=BOOLEAN
     *   When set to true it uses XdsServerCredentials with the test server for security test cases.
         In case of secure mode, port and maintenance_port should be different.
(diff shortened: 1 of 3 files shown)


gnossen:
I think technically [this](https://github.com/grpc/grpc-proto/blob/master/grpc/testing/test.proto) is the source of truth for this proto. It can probably be added there as a follow-up though.

gnossen (approved):

src/proto/grpc/testing/messages.proto
```diff
@@ -268,3 +268,12 @@ message ClientConfigureRequest {
 
 // Response for updating a test client's configuration.
 message ClientConfigureResponse {}
+
+// Configurations for a test server
+message ServerConfigureRequest {
+  // The status code with which to respond to all requests
+  EchoStatus response_status = 1;
```
lidizheng, line 275: Should we consider adding the `rpc-behavior` configurations here?

doc/xds-test-descriptions.md
```diff
@@ -93,6 +93,35 @@ The test client changes its behavior right after receiving the
...
+
+```
+message EchoStatus {
+  int32 code = 1;
+  string message = 2;
+}
+
+message ServerConfigureRequest {
+  // The status code with which to respond to all requests
+  EchoStatus response_status = 1;
```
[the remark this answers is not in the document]
murgatroid99: The `EchoStatus` message already exists in `messages.proto`, so I figured that it would be easier to just reuse it.

src/proto/grpc/testing/messages.proto
murgatroid99: We can always do that in a future PR.

lidizheng (approved):

PR #29457, opened 2022-04-20
health check service: rewrite using callback API

markdroth:
This is *so* much simpler to understand and reason about, and it doesn't require a special CQ or a separate thread.


yashykt (commented):
this does look nice!

yashykt (approved):

markdroth:
Yes, there are tests in test/cpp/end2end/health_service_end2end_test.cc.

PR #29458, opened 2022-04-20
Revert "Fix Python xDS Interop Client Time Slicing"

yashykt:
Reverts grpc/grpc#29423
It's causing troubles when importing internally


lidizheng (approved):
Please wait for @gnossen to take a look.

PR #29460, opened 2022-04-20
Revert "Revert "Fix Python xDS Interop Client Time Slicing""

gnossen:
Reverts grpc/grpc#29458


lidizheng (approved):
LGTM as the needed change to fix the breakage is in google3.

PR #29462, opened 2022-04-20
HTTP: Send GOAWAY frame with last stream ID on immediate disconnect

yashykt:
Improves Core's shutdown and retry story. Also helps with b/229488257
Currently, Core servers do not send a GOAWAY frame with last stream ID if the servers want to immediately disconnect, leading to a situation where the clients do not know which RPCs were processed by the server. This makes transparent retry impossible. This PR changes chttp2 to send that GOAWAY frame. Note that Core/C++ server API does not do "graceful" shutdown yet since it has no grace period where it accepts new requests, so this would not totally fix it.

We should also eventually fix the transport ops to have a single method of sending down a goaway with 1) grace period 2) how long to wait for all RPCs to finish.

### DIFF
--- a/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
+++ b/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
@@ -1859,19 +1859,22 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
 
 }  // namespace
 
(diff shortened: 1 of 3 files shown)


markdroth:
I'll let Craig review this.  I'm happy as long as you've verified that this fixes the flake in b/229488257.

yashykt:
Note that this does not fix the flake in b/229488257 but this will aid the final fix

PR #29468, opened 2022-04-21
build_and_run_docker.sh cleanup and simplification

jtattermusch:
The build_and_run_docker.sh is our main entry point for running run_tests.py, task_runner.py (artifacts, packages, distribtests) and bazel under docker.
- Perform some refactoring to improve the structuring and readability of the script (and add a lot of explanatory comments)
- simplify the way reports (ie. the `./reports` dir and XML test reports) and output files (e.g. binaries that are results of the artifact build tasks) are handled (so that we don't require as much postprocessing by the DOCKER_RUN_SCRIPT).
- make some of the well-known kokoro directories (KOKORO_GFILE_DIR, KOKORO_ARTIFACTS_DIR) available from inside the docker container by default, making it simpler to run some special test suites (e.g. bload_diff, memory_diff etc.) under docker.

Overall, this is preparation for making the build_and_run_docker.sh script usable for running tests/experiments under docker locally (without needing to duplicate most of the logic in run_in_docker.sh) and to be able to run bazel under docker much more easily (both locally and on CI). Also it will create more opportunities for simplification of the run_tests.py / task_runner.py stack.

### DIFF
--- a/tools/internal_ci/linux/grpc_bloat_diff.sh
+++ b/tools/internal_ci/linux/grpc_bloat_diff.sh
@@ -22,7 +22,4 @@ source tools/internal_ci/helper_scripts/prepare_build_linux_rc
 
 export DOCKERFILE_DIR=tools/dockerfile/test/cxx_debian11_x64
 export DOCKER_RUN_SCRIPT=tools/internal_ci/linux/grpc_bloat_diff_in_docker.sh
-# The check_on_pr.py needs access to the key to post status on github PRs,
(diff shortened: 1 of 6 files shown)


veblush (approved):

jtattermusch:
Test failures are unrelated (test flakes and ruby on macos artifact build failure)

lidizheng:
This issue might be related to b/231142315 🤔

PR #29469, opened 2022-04-21
Cleanup run_in_docker.sh and corresponding examples.

jtattermusch:
~~Based on https://github.com/grpc/grpc/pull/29468.~~ (rebased)

Cleanup `run_in_docker.sh` script and its examples.
- the script can now reuse the logic in build_and_run_docker.sh, which is good because it reduces the amount of duplication (=easier maintenance) and also makes local test invocations more similar to those on CI (because most codepath in the scripts is now shared).
- some of the example scripts were out of date (e.g. out of date docker image name).


jtattermusch:
CC @lidizheng  @drfloob

veblush (approved):

jtattermusch:
Known failures:
- b/230080881
- other failures are also unrelated since the run_in_docker.sh script and the corresponding examples aren't actually run by CI at all.

PR #29470, opened 2022-04-21
Reduce spam log messages at server under high memory pressure

Vignesh2208:
Only log one message for every 1000 dropped incoming connections due to high memory pressure.

@ctiller

### DIFF
--- a/src/core/lib/iomgr/tcp_server_posix.cc
+++ b/src/core/lib/iomgr/tcp_server_posix.cc
@@ -29,6 +29,7 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <inttypes.h>
 #include <netinet/in.h>
 #include <netinet/tcp.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -221,7 +224,13 @@ static void on_read(void* arg, grpc_error_handle err) {
     }
 
     if (sp->server->memory_quota->IsMemoryPressureHigh()) {
-      gpr_log(GPR_INFO, "Drop incoming connection: high memory pressure");
+      int64_t dropped_connections_count = ++num_dropped_connections;
```
ananda1066, line 227: Don't we have to load the value from the atomic? Please also add Craig to review atomic usage.

Vignesh2208: Sure I can add craig. I don't think it matters because the ++ operator is equivalent to fetch_add and we can probably tolerate race conditions also. I made it explicit now anyway

ananda1066 (approved):

ctiller (approved):

src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -221,7 +224,14 @@ static void on_read(void* arg, grpc_error_handle err) {
     }
 
     if (sp->server->memory_quota->IsMemoryPressureHigh()) {
-      gpr_log(GPR_INFO, "Drop incoming connection: high memory pressure");
+      int64_t dropped_connections_count =
+          num_dropped_connections.fetch_add(1, std::memory_order_relaxed) + 1;
+      if (dropped_connections_count % 1000 == 0) {
```
ctiller, line 229: Change this to == 1 so that we print a message on the first dropped connection

PR #29475, opened 2022-04-21
xDS test server: document rpc-behavior metadata, add hostname option

murgatroid99:
The `sleep-<number>` entry just documents the existing usage in the "timeout" test case. The `fail-on` option is new, and will be used to implement the outlier detection test.

This is an alternative to #29448 for implementing an outlier detection test.

### DIFF
--- a/doc/xds-test-descriptions.md
+++ b/doc/xds-test-descriptions.md
@@ -20,8 +20,12 @@ Server should accept these arguments:
 In addition, when handling requests, if the initial request metadata contains the `rpc-behavior` key, it should modify its handling of the request as follows:
 
  - If the value matches `sleep-<int>`, the server should wait the specified number of seconds before responding.
- - If the value matches `fail=<status code as int>`, the server should respond with the specified status code.
- - A value can have a prefix `hostname=<string>` followed by a space. In that case, the rest of the value should only be applied if the specified hostname matches the server's hostname. 
+ - If the value matches `keep-open`, the server should never respond to the request.
+ - If the value matches `success-on-retry-attempt-<int>`, the server should force an OK response if the value of the `grpc-previous-rpc-attempts` metadata field is equal to the specified number.
+ - If the value matches `error-code-<int>`, the server should respond with the specified status code.
+ - A value can have a prefix `hostname=<string>` followed by a space. In that case, the rest of the value should only be applied if the specified hostname matches the server's hostname.
+
+If a request has multiple `rpc-behavior` metadata values, they should be applied in the order listed.
 
 ## Client
 
--- a/doc/xds-test-descriptions.md
+++ b/doc/xds-test-descriptions.md
@@ -17,6 +17,11 @@ Server should accept these arguments:
     *   When set to true it uses XdsServerCredentials with the test server for security test cases.
         In case of secure mode, port and maintenance_port should be different.
 
+In addition, when handling requests, if the initial request metadata contains the `rpc-behavior` key, it should modify its handling of the request as follows:
+
+ - If the value matches `sleep-<number>`, the server should wait the specified number of seconds before responding.
+ - If the value matches `fail-on=<hostname>`, and the specified hostname matches the server's hostname, the server should respond to the request with an error with the code `ABORTED`.
+
 ## Client
 
 The base behavior of the xDS test client is to send a constant QPS of unary
--- a/doc/xds-test-descriptions.md
+++ b/doc/xds-test-descriptions.md
@@ -21,11 +21,11 @@ In addition, when handling requests, if the initial request metadata contains th
 
  - If the value matches `sleep-<int>`, the server should wait the specified number of seconds before responding.
  - If the value matches `keep-open`, the server should never respond to the request.
- - If the value matches `success-on-retry-attempt-<int>`, the server should force an OK response if the value of the `grpc-previous-rpc-attempts` metadata field is equal to the specified number.
  - If the value matches `error-code-<int>`, the server should respond with the specified status code.
+ - If the value matches `success-on-retry-attempt-<int>`, the server should not apply the `error-code` option if the value of the `grpc-previous-rpc-attempts` metadata field is equal to the specified number.
  - A value can have a prefix `hostname=<string>` followed by a space. In that case, the rest of the value should only be applied if the specified hostname matches the server's hostname.
 
-If a request has multiple `rpc-behavior` metadata values, they should be applied in the order listed.
+The `rpc-behavior` header value can have multiple options separated by commas. In that case, the value should be split by commas and the options should be applied in the order specified. If a request has multiple `rpc-behavior` metadata values, each one should be processed that way in order.
 
 ## Client
 
--- a/doc/xds-test-descriptions.md
+++ b/doc/xds-test-descriptions.md
@@ -19,10 +19,10 @@ Server should accept these arguments:
 
 In addition, when handling requests, if the initial request metadata contains the `rpc-behavior` key, it should modify its handling of the request as follows:
 
- - If the value matches `sleep-<int>`, the server should wait the specified number of seconds before responding.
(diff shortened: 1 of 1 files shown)


doc/xds-test-descriptions.md
```diff
@@ -17,6 +17,11 @@ Server should accept these arguments:
     *   When set to true it uses XdsServerCredentials with the test server for security test cases.
         In case of secure mode, port and maintenance_port should be different.
 
+In addition, when handling requests, if the initial request metadata contains the `rpc-behavior` key, it should modify its handling of the request as follows:
+
+ - If the value matches `sleep-<number>`, the server should wait the specified number of seconds before responding.
+ - If the value matches `fail-on=<hostname>`, and the specified hostname matches the server's hostname, the server should respond to the request with an error with the code `ABORTED`.
```
lidizheng, line 23: Recommend to define the error message , to avoid this code being thrown by other part of the library.

doc/xds-test-descriptions.md
```diff
@@ -17,6 +17,11 @@ Server should accept these arguments:
     *   When set to true it uses XdsServerCredentials with the test server for security test cases.
         In case of secure mode, port and maintenance_port should be different.
 
+In addition, when handling requests, if the initial request metadata contains the `rpc-behavior` key, it should modify its handling of the request as follows:
```
ejona86, line 20: This is a CSV header, so servers should split on `,` and act as if multiple headers were a single string separated by `,`.

Java processes these immediately, so `sleep-10,fail-on=blah` would sleep for 10 seconds before failing if hostname is blah whereas `fail-on=blah,sleep-10` would fail immediately if hostname is blah. I don't know if other languages do that, but it seems easy to implement and easy to define and reasonably powerful.

murgatroid99:
OK, I updated with the various existing options implemented in the Java server. The only new thing is the optional `hostname` prefix.

doc/xds-test-descriptions.md
murgatroid99: I didn't explicitly mention the CSV format because the protocol spec already says that separate metadata entries with the same key are semantically equivalent to a comma-separated list of values in a single metadata entry.

doc/xds-test-descriptions.md
murgatroid99: The Java server currently uses the error message `Rpc failed as per the rpc-behavior header value: <rpc-behavior value>`. I don't think it's important for this to be a specific value that the client or driver can check for, because the call stats just count the error codes.

doc/xds-test-descriptions.md
murgatroid99: I did add a note about applying multiple values in order.

doc/xds-test-descriptions.md
lidizheng: Using the current interop design, you are right, we don't have a way to validate the message content. I guess in future we should take this into account.

ejona86: I don't think our tests should be checking the message like that. I think it should have a "reasonable" message, but am happy with it being implementation-defined.

I would be happy if we had the error descriptions for the interop tests, but that is more about debugging and not having a test assert a particular message.

doc/xds-test-descriptions.md
ejona86: Sure, yeah. Anything that implies that we support multivalue is good.

ejona86: I don't see anything that says this field is a CSV. And I see now we don't say we support multivalue headers. It is just "multiple values" but in a generic sense. We could say that `rpc-behavior` is a comma-separated list and multiple headers must be supported.

The RFC does say how to combine multiple headers into one, but I don't think it would be clear to implementors of this that we actually expect that to work unless we explicitly mention it. The RFC does not say that every header must support the concatenation semantics. It just says that if there are duplicate keys, they can be combined together without impacting their semantics.

doc/xds-test-descriptions.md
```diff
@@ -20,8 +20,12 @@ Server should accept these arguments:
...
 
  - If the value matches `sleep-<int>`, the server should wait the specified number of seconds before responding.
- - If the value matches `fail=<status code as int>`, the server should respond with the specified status code.
- - A value can have a prefix `hostname=<string>` followed by a space. In that case, the rest of the value should only be applied if the specified hostname matches the server's hostname. 
+ - If the value matches `keep-open`, the server should never respond to the request.
+ - If the value matches `success-on-retry-attempt-<int>`, the server should force an OK response if the value of the `grpc-previous-rpc-attempts` metadata field is equal to the specified number.
+ - If the value matches `error-code-<int>`, the server should respond with the specified status code.
+ - A value can have a prefix `hostname=<string>` followed by a space. In that case, the rest of the value should only be applied if the specified hostname matches the server's hostname.
+
+If a request has multiple `rpc-behavior` metadata values, they should be applied in the order listed.
```
ejona86, line 28: I don't think this is sufficient. You really have to specify how to parse a single entry, because that is key-specific. The HTTP spec does not specify the format for a single-value key, even if it was created by concatenating a multi-value key.

> Each `rpc-behavior` header value (as there may be multiple) should be split on `,` and the resulting actions should be processed in-order.

To contrast, we aren't supporting `hostname=hello\,world` or `hostname="bad,name"`, which for some HTTP multi-valued headers would be allowed.

doc/xds-test-descriptions.md
```diff
@@ -20,8 +20,12 @@ Server should accept these arguments:
 In addition, when handling requests, if the initial request metadata contains the `rpc-behavior` key, it should modify its handling of the request as follows:
 
  - If the value matches `sleep-<int>`, the server should wait the specified number of seconds before responding.
- - If the value matches `fail=<status code as int>`, the server should respond with the specified status code.
- - A value can have a prefix `hostname=<string>` followed by a space. In that case, the rest of the value should only be applied if the specified hostname matches the server's hostname. 
+ - If the value matches `keep-open`, the server should never respond to the request.
+ - If the value matches `success-on-retry-attempt-<int>`, the server should force an OK response if the value of the `grpc-previous-rpc-attempts` metadata field is equal to the specified number.
```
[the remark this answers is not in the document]
murgatroid99: I assume that this is supposed to override the behavior of the `error-code` option, and that's what I was trying to communicate with "forces an OK response". Maybe I should just say that explicitly.

doc/xds-test-descriptions.md
murgatroid99: OK, I added a sentence specifying this.

doc/xds-test-descriptions.md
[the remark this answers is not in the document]
murgatroid99: I updated the wording. Is that better?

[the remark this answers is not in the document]
ejona86: I don't think it is "don't do `error-code-`" as much as "don't process any more behaviors." It is a terminal state, essentially. As written now it sounds like if an `error-code-` exists earlier in the list this will retroactively ignore it.

Does something like this work better?
 - If the value matches `sleep-<int>`, the server should wait the specified number of seconds before resuming behavior matching and RPC processing.
 - If the value matches `keep-open`, the server should never respond to the request and behavior matching ends.
 - If the value matches `error-code-<int>`, the server should respond with the specified status code and behavior matching ends.
 - If the value matches `success-on-retry-attempt-<int>` and the value of the `grpc-previous-rpc-attempts` metadata field is equal to the specified number, the normal RPC processing should resume and behavior matching ends.

I think there's some other ways this could be phrased, but I was trying to sort of match the existing style. Some rearrangements could reduce the verbosity, but we've already spent more time on wording this than we were hoping for.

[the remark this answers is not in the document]
murgatroid99: I updated it to use your wording.

ejona86 (approved):

PR #29478, opened 2022-04-21
Fix grpc_sockaddr_to_uri to return URI encoded string. Also see #29323 which has a similar fix for unix abstract addresses.

anicr7:
@markdroth 
@ctiller


binary.dat
```diff
@@ -0,0 +1 @@
+
```
markdroth, line 1: I assume this file was added to the PR by mistake.

markdroth (approved):
Other than one small comment, this looks good to me!

I assume this will need a cherry-pick import.  See go/grpc-cherrypick for details.  Note that you should not merge this PR until you have the cherry-pick CL ready to submit, at which point you should coordinate with the on-duty importer and then get the PR merged and CL submitted at the same time.

test/core/address_utils/sockaddr_utils_fuzzer_test.cc
```diff
@@ -0,0 +1,43 @@
...
+  memset(&address, 0, sizeof(address));
+  memcpy(address.addr, data, size);
+  address.len = size;
+
+  absl::StatusOr<std::string> uri = grpc_sockaddr_to_uri(&address);
+  if (!uri.ok()) return 0;
+  absl::StatusOr<grpc_core::URI> parsed_uri =
+      grpc_core::URI::Parse(uri.value());
+
+  GPR_ASSERT(parsed_uri.ok());
```
ctiller, line 41: Can we go some steps further here (needn't stall this PR) and verify that:

1. `grpc_parse_uri` succeeds
2. `grpc_sock_addr_to_uri` succeeds on the output of the above
3. the two uri strings compare equal

ctiller (approved):
(forgot to approve this... the fuzzer change is optional, but would be really cool)

test/core/address_utils/sockaddr_utils_fuzzer_test.cc
anicr7: Craig, 

This sounds like a good idea. I tried these out locally and I see a lot of failures now. A lot of these failures I feel are majorly due to the test setup. 

Currently we arbitrary allow any data that the fuzzer creates but grpc_parse_uri performs validation when converting the URI back to resolved address. For example for IPV6 addresses the scope has to be valid which while constructing the resolved_address, the test didn't confirm it was as the grpc_resolved_address is just memcpy of the data.

I can change the logic to be more sophisticated here and only proceed with  grpc_parse_uri if the address was truly valid. If there are better ways for me to construct the resolved_address let me know.

binary.dat
anicr7: Removed.

ctiller (approved):

PR #29482, opened 2022-04-22
Fix crash when sockaddr cannot be converted to string

sifmelcara:
`grpc_sockaddr_to_uri()` does not support binder transport URI scheme
(which uses an invalid sa_family (`AF_MAX`) to avoid misuse) at
this moment. Fallback to empty string here to keep the old behavior.

### DIFF
--- a/src/core/ext/filters/client_channel/subchannel.cc
+++ b/src/core/ext/filters/client_channel/subchannel.cc
@@ -676,7 +676,8 @@ Subchannel::Subchannel(SubchannelKey key,
             {GRPC_MAX_CHANNEL_TRACE_EVENT_MEMORY_PER_NODE_DEFAULT, 0,
              INT_MAX}));
     channelz_node_ = MakeRefCounted<channelz::SubchannelNode>(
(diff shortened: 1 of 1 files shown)


src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -676,7 +676,7 @@ Subchannel::Subchannel(SubchannelKey key,
             {GRPC_MAX_CHANNEL_TRACE_EVENT_MEMORY_PER_NODE_DEFAULT, 0,
              INT_MAX}));
     channelz_node_ = MakeRefCounted<channelz::SubchannelNode>(
-        grpc_sockaddr_to_uri(&key_.address()).value(),
+        grpc_sockaddr_to_uri(&key_.address()).value_or(""),
```
markdroth, line 679: The empty string isn't very useful.  Let's make it something like "<unknown address type>".

markdroth: Looks like github munged my comment.  I meant something like `<unknown address type>`.

markdroth (approved):

PR #29483, opened 2022-04-22
Binder transport: Fix server issue when handling many parallel RPC calls

sifmelcara:
When single closure is scheduled multiple times before it is run, it
runs less times than the number of times it is scheduled.

As a result, when server receives multiple RPC calls in a very short
time frame, `accept_stream_locked` is not correctly called for every
RPC call received.

We are working on internal server side stress test to make sure this
kind of error won't happen again.


sifmelcara:
@ctiller After looking at `combiner` and `grpc_closure`'s API I'm not sure if it is intended to not support scheduling the same closure multiple times, would you be able to confirm this?

sifmelcara:
Force pushed. Added a unit test case to verify the fix and also make a small fix for another parallel RPC call race, commit message updated. For some reason we missed this test case before..

LittleCVR (approved):

PR #29485, opened 2022-04-22
Add DNS Server address override to the EventEngine API

drfloob:


include/grpc/event_engine/event_engine.h
```diff
@@ -331,8 +336,11 @@ class EventEngine {
   // de-experimentalize this API.
   virtual bool IsWorkerThread() = 0;
 
-  /// Creates and returns an instance of a DNSResolver.
-  virtual std::unique_ptr<DNSResolver> GetDNSResolver() = 0;
+  /// Creates and returns an instance of a DNSResolver, optionally configured by
+  /// the \a options struct.
+  virtual std::unique_ptr<DNSResolver> GetDNSResolver(
+      const DNSResolver::ResolverOptions& options =
```
markdroth, line 342: @ctiller tends to really dislike default-value parameters.  Might be better to just make this required; it's not that onerous to require the caller to instantiate an empty struct when they call this method.

include/grpc/event_engine/event_engine.h
```diff
@@ -261,6 +261,11 @@ class EventEngine {
     struct LookupTaskHandle {
       intptr_t key[2];
     };
+    /// Optional configuration for DNSResolvers.
+    struct ResolverOptions {
+      /// Override the DNSResolver's default authority.
+      std::string authority;
```
markdroth, line 267: Using the term "authority" here seems weird, because this API is not expressed in terms of "dns:" URIs.  Instead, I suggest calling this "dns_server".

include/grpc/event_engine/event_engine.h
```diff
@@ -261,6 +261,11 @@ class EventEngine {
     struct LookupTaskHandle {
       intptr_t key[2];
     };
+    /// Optional configuration for DNSResolvers.
+    struct ResolverOptions {
+      /// Override the DNSResolver's default authority.
```
markdroth, line 266: We should say something like "if empty, default DNS servers will be used."

Also, do we want to restrict this to being specified as IP:port, or do we want to allow EE impls to be able to support resolving this name using the default DNS servers?  Currently, we support only IP:port.

include/grpc/event_engine/event_engine.h
drfloob: Done

include/grpc/event_engine/event_engine.h
drfloob: Done

include/grpc/event_engine/event_engine.h
drfloob: Good question. I'm in favor of limiting it to IP:port for YAGNI reasons: it avoids the added complexity of supporting a feature that can't be used today. Changing gRPC's URI policy later would technically mean a breaking change for the engine API, but one that most applications will probably never encounter. WDYT?

markdroth (approved):

PR #29495, opened 2022-04-26
Add log urls to junit report.

wanlin31:
This pr adds the feature of providing links to logs in load test Junit reports.

This pr should be merged after https://github.com/grpc/test-infra/pull/306.

### DIFF
--- a/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
+++ b/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
@@ -132,8 +132,8 @@ time ../test-infra/bin/prepare_prebuilt_workers \
 time ../test-infra/bin/runner \
     -i "loadtest_with_prebuilt_workers_${WORKER_POOL_8CORE}.yaml" \
     -i "loadtest_with_prebuilt_workers_${WORKER_POOL_32CORE}.yaml" \
+    -log-url-prefix "${LOG_URL_PREFIX}" \
     -polling-interval 5s \
     -delete-successful-tests \
-    -log-url-prefix "${LOG_URL_PREFIX}" \
     -c "${WORKER_POOL_8CORE}:2" -c "${WORKER_POOL_32CORE}:2" \
     -o "runner/sponge_log.xml"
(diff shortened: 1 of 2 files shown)


tools/internal_ci/linux/grpc_e2e_performance_gke.sh
```diff
@@ -126,11 +126,15 @@ time ../test-infra/bin/prepare_prebuilt_workers \
     -t "${UNIQUE_IDENTIFIER}" \
     -r "${ROOT_DIRECTORY_OF_DOCKERFILES}"
 
+# calculate the prefix of the log URL
```
paulosjca, line 129: I would move this to the environment variable definition section, under WORKER_POOL_32CORE.

I would change the comment to something like "Prefix for log URLs in cnsviewer."

It looks like there is only one version of cnsviewer now, so I would change the URL to use "cnsviewer" instead of "cnsviewer2".

tools/internal_ci/linux/grpc_e2e_performance_gke_experiment.sh
```diff
@@ -124,11 +124,15 @@ time ../test-infra/bin/prepare_prebuilt_workers \
     -t "${UNIQUE_IDENTIFIER}" \
     -r "${ROOT_DIRECTORY_OF_DOCKERFILES}"
 
+# calculate the prefix of the log URL
+LOG_URL_PREFIX="http://cnsviewer2/placer/prod/home/kokoro-dedicated/build_artifacts/${KOKORO_BUILD_ARTIFACTS_SUBDIR}/github/grpc/"
```
paulosjca, line 128: Same comment as in the other file.

tools/internal_ci/linux/grpc_e2e_performance_gke.sh
wanlin31: I have verified that `cnsviewer` link worked okay.

tools/internal_ci/linux/grpc_e2e_performance_gke_experiment.sh
```diff
@@ -130,5 +132,6 @@ time ../test-infra/bin/runner \
     -i "loadtest_with_prebuilt_workers_${WORKER_POOL_32CORE}.yaml" \
     -polling-interval 5s \
     -delete-successful-tests \
+    -log-url-prefix "${LOG_URL_PREFIX}" \
```
paulosjca, line 135: Nit: I would add -log-url-prefix right after -i, to follow the order of the flags in the help. But it is not important.

paulosjca (approved):

PR #27193, opened 2021-08-30
Updated backoff to use absl::Random 

veblush:
https://github.com/grpc/grpc/pull/26479 & internal b/192110891

PR #29496, opened 2022-04-26
Delete BUILD.gn

tamird:
This was erroneously added in https://github.com/grpc/grpc/pull/27193
via bad merge conflict resolution.

The presence of this file makes it impossible for Fuchsia to update
grpc.

@veblush


drfloob:
@tamird Thanks for contributing. Is this a critical bug at all? We generally only do patch releases when they are absolutely essential, it's an onerous process that takes a few days of testing and requires a half dozen people participate. This backport PR is fine to land, but it would not be actionable unless having `BUILD.gn` in the build is a significant problem.

tamird:
As mentioned in the commit message and PR description, the presence of this file in any released version makes that version *unusable* in Fuchsia's build system.

drfloob:
Understood. BUILD.gn existed in the v1.44.0 release as well, which has been out since February. Maybe nobody in the Fuchsia community has attempted to upgrade? The v1.46.0 release will be coming out in a week or two, might it be possible to skip v1.45, or do you need v1.45 for some specific reason?

tamird:
Yes, nobody has attempted to update for some time. We're ok with skipping 1.45.

drfloob:
@tamird Let's make that the plan, then. The tests will not pass due to CI changes that have happened since this release, and we'd need an admin to force-merge it.

nicolasnoble:
Someone called?

PR #29497, opened 2022-04-26
Add script for running msan test locally

sifmelcara:
A dirty workaround that enables running msan locally


ctiller:
@veblush probably best for you to make a call on where to take this

veblush:
It seems a bit too much to me. Maybe we would rather look for other avenue such as using a docker image to leverage msan. But I'd like hear from Jan about this matter.

jtattermusch:
> It seems a bit too much to me. Maybe we would rather look for other avenue such as using a docker image to leverage msan. But I'd like hear from Jan about this matter.

Agreed. I think if we cleanup our bazelrc files, you should be able to run msan locally with much fewer changes.
Also, you can very likely run msan under docker container today (after some time spent figuring out things).

So it's good that you investigated what it takes to run msan locally (which is something that should work),
but we should get other sanitizers to work locally as or (or at least locally under a docker container).

Just to make sure - this is for linux, not Mac, right?

sifmelcara:
> we should get other sanitizers to work locally as or (or at least locally under a docker container).

Back when I wrote this PR, I believe asan/tsan/ubsan already worked locally, but not msan. msan is different than asan/tsan/ubsan because it needs MSan-instrumented standard library to work properly. msan CI job does link with proper MSan-instrumented standard library but local toolchain didn't. (See https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo#introduction).

We can refactor existing toolchain to make it select different C++ standard library to link with when msan is enabled, so that we don't need a separate toolchain. However user will still need to run `build_clang.sh` in this CL to create MSan-instrumented C++ stdlib, unless someone precompile it and host it somewhere.

> Just to make sure - this is for linux, not Mac, right?

Yes this is for Linux, and only tested on Linux

jtattermusch:
> > we should get other sanitizers to work locally as or (or at least locally under a docker container).
> 
> Back when I wrote this PR, I believe asan/tsan/ubsan already worked locally, but not msan. msan is different than asan/tsan/ubsan because it needs MSan-instrumented standard library to work properly. msan CI job does link with proper MSan-instrumented standard library but local toolchain didn't. (See https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo#introduction).
> 
> We can refactor existing toolchain to make it select different C++ standard library to link with when msan is enabled, so that we don't need a separate toolchain. However user will still need to run `build_clang.sh` in this CL to create MSan-instrumented C++ stdlib, unless someone precompile it and host it somewhere.

FYI
go/llvm-docker and the pre-built versions are in https://storage.googleapis.com/clang-builds-stable
This is what we use in our RBE linux image:
https://github.com/grpc/grpc/blob/041b7342183d7380f6580a1566325014153b6999/templates/tools/dockerfile/test/rbe_ubuntu2004/Dockerfile.template#L30


> 
> > Just to make sure - this is for linux, not Mac, right?
> 
> Yes this is for Linux, and only tested on Linux

jtattermusch:
Ad how to run msan locally in a docker sandbox (using the same docker image that's used by RBE):

E.g.
```
tools/bazel --bazelrc=tools/remote_build/linux_docker_sandbox.bazelrc test --config=msan //test/core/alarm_test
```

The general approach is described here: 
https://bazel.build/remote/sandbox

The "--experimental_enable_docker_sandbox" is taken from tools/remote_build/linux_docker_sandbox.bazelrc and the docker image used is the one we use from RBE (since the RBE platform definition provides the docker image to use by default here https://github.com/grpc/grpc/blob/041b7342183d7380f6580a1566325014153b6999/third_party/toolchains/rbe_ubuntu2004/config/BUILD#L45)

Some tweaks might be necessary, but generally this approach works pretty well (and you can use it for most stuff that runs on RBE linux).

PR #29503, opened 2022-04-27
Adding a min progress size argument to grpc_endpoint_read to allow gRPC to use TCP optimizations on the read path

Vignesh2208:
Adds a min progress size argument which lets upper level layers to communicate to the TCP layer and specify the minimum number of bytes required to be read to make meaningful progress.

Requires cherrypick

@markdroth

### DIFF
--- a/src/core/lib/iomgr/tcp_posix.cc
+++ b/src/core/lib/iomgr/tcp_posix.cc
@@ -426,10 +426,6 @@ struct grpc_tcp {
                                       on errors anymore */
   TcpZerocopySendCtx tcp_zerocopy_send_ctx;
   TcpZerocopySendRecord* current_zerocopy_send = nullptr;
-
-  int min_progress_size; /* A hint from upper layers specifying the minimum
-                            number of bytes that need to be read to make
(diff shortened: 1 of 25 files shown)


src/core/lib/iomgr/tcp_posix.cc
```diff
@@ -859,7 +863,8 @@ static bool tcp_do_read(grpc_tcp* tcp, grpc_error_handle* error)
 
 static void maybe_make_read_slices(grpc_tcp* tcp)
     ABSL_EXCLUSIVE_LOCKS_REQUIRED(tcp->read_mu) {
-  if (tcp->incoming_buffer->length == 0 &&
```
ctiller, line 862: Let's separate this piece out into another PR: the API change will be a complicated cherry-pick, and this change may need to roll back - so by splitting them we give ourselves the luxury of an easy rollback if needed - without the need to orchestrate a second cherry-pick.

Vignesh2208: Done

ctiller (approved):

PR #29509, opened 2022-04-28
[WIP] Teach iomgr’s DNSResolver to use EventEngine::ResolvedAddress

drfloob:
See src/core/lib/iomgr/resolve_address.h for the essential changes.

This is a step towards making the iomgr DNS API sufficient to drive a
client channel resolver. In this change, clients will begin handling
EventEngine lookup result types. EventEngine will eventually replace
iomgr, so using EventEngine types now should simplify things a bit
later.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1767,6 +1767,7 @@ grpc_cc_library(
     deps = [
         "gpr_base",
         "iomgr_port",
+        "event_engine_base_hdrs",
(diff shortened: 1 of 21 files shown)


drfloob:
Closing this for now. Feedback was that we should convert all or nothing, so I'll do the larger conversion at a later stage.

PR #29539, opened 2022-04-29
resolver: add more attributes to resolver result

markdroth:
This is a preparatory step toward minimizing the extent to which we expose channel args to the resolver and LB policy APIs.  In subsequent PRs, I will change all of the cases where resolvers currently pass data to the LB policy via channel args to instead use attributes.  Once that's done, I will remove channel args from the resolver result.

Channel args will still be passed *to* the resolver at instantiation time, but I will later change that to be exposed via a smaller API surface using something similar to what we do in EventEngine.


src/core/lib/resolver/server_address.h
```diff
@@ -32,44 +31,74 @@
...
 
-//
-// ServerAddress
-//
-
-// A server address is a grpc_resolved_address with an associated set of
-// channel args.  Any args present here will be merged into the channel
-// args when a subchannel is created for this address.
-class ServerAddress {
+class ResolverAttributeMap {
```
ctiller, line 34: Lets split this into its own file (preferably its own build target too if possible), and let's also give it some unit tests.

src/core/lib/resolver/server_address.h
```diff
@@ -32,44 +31,74 @@
...
+
+  // Removes the attribute from the map.
+  void Remove(const char* key);
+
+  std::string ToString() const;
+
+  bool empty() const { return map_.empty(); }
+
+ private:
+  std::map<const char*, std::unique_ptr<AttributeInterface>> map_;
```
ctiller, line 85: Comparing by static string pointers seems pretty fragile, suggest absl::string_view instead?

src/core/lib/resolver/server_address.h
```diff
@@ -32,44 +31,74 @@
...
+// A server address is an address with two sets of attributes attached,
+// one set that affects subchannel behavior, and another set that is
+// visible only to LB policies.
+// TODO(roth): Remove grpc_channel_args from this class.
+class ServerAddress {
+ public:
+  ServerAddress(
+      const grpc_resolved_address& address, grpc_channel_args* args,
+      ResolverAttributeMap subchannel_attributes = ResolverAttributeMap(),
+      ResolverAttributeMap lb_policy_attributes = ResolverAttributeMap());
```
ctiller, line 101: This is getting to the point where I'd want to suggest a `ServerAddressBuilder` to assist putting these things together.

src/core/lib/resolver/server_address.h
```diff
@@ -32,44 +31,74 @@
...
-  // Attributes are keyed by a C string that is unique by address, not
-  // by value.  All attributes added with the same key must be of the
-  // same type.
   class AttributeInterface {
    public:
     virtual ~AttributeInterface() = default;
 
+    // The type name for this attribute.
+    // There can be only one attribute of a given type in a given AttributeMap.
+    // Attributes are keyed by the address of the string, not the value.
```
ctiller, line 43: Is there an incredibly strong justification for this?

ctiller: Providing a string instance does not solve this problem, and the design doesn't work.

C++ implementations are allowed to (but not required to) merge duplicate string values.
You've recently seen cases where the same string token was given two pointers in one build system.
It would be a valid compile for two instances of `"foo"` in the same binary, coming from two different libraries, to be given the same pointer also.
C++ does not provide the guarantees this code is asking for.

Suggestions to improve the design:
1. mandate a namespace prefix (`grpc.lb_attr1`, `grpc.lb_attr2`, `customer.lb_attr1`) as we do for channel args (indeed the reason we do this for channel args)
2. implement an attribute class that actually provides the guarantees you're looking for (probably internally calling new on some non-zero length data structure to ensure a unique pointer), use that throughout instead of `const char *`.

markdroth: Thanks for pointing this out!

I've sent #29709 to introduce a `UniqueTypeName` API along the lines of your second suggestion and to change the other places in our code where we're using this pattern to use `UniqueTypeName` instead.

I've now merged that PR into this one and changed the resolver attribute API to also use `UniqueTypeName`.

markdroth:
Marking this as draft again while I consider how to rework the approach, as per our discussion today.

PR #29541, opened 2022-04-30
L97: Setuptools entrypoint for generating file

plannigan:
Still just an initial proof of concept that generates files based on proto files and known locations. But figuring that out was the hard part.

@donnadionne


plannigan:
Not quite. It is currently still a POC implementation. I have some local changes that read from `pyproject.toml` to generate code from a single proto file. I should have some more time tomorrow to expand it handle multiple files and write some test cases.

plannigan:
The functionality should now be complete now and ready for review. I still need to add documentation for how to use the feature.

I haven't used Bazel before. I think I added the new test file correctly, but not sure. I could also see an additional automated end to end test that used the sample package, but that might not be necessary.

plannigan:
It looks like there is an issue finding the new dependency [tomli](https://pypi.org/project/tomli/). Maybe I missed something as part of the Bazel config?

tools/distrib/python/grpcio_tools/setup.py
```diff
@@ -293,8 +293,14 @@ def extension_modules():
                      'protobuf>=3.12.0, < 4.0dev',
                      'grpcio>={version}'.format(version=grpc_version.VERSION),
                      'setuptools',
+                     'tomli>=1.2,<3;python_version<"3.11"'
```
lidizheng, line 296: In general, gRPC as a library wants to minimize our dependency. Adding a hard dependency means potential bloat, conflict, duplication. So unless there is a strong reason to do so, we won't add extra dependencies.

That being said, can we consider following two directions?

1. Include a minimum implementation of TOML in grpcio-tools;
2. Consider a wrapper package over grpcio-tools to ship this feature (e.g., grpcio-setuptools).

cc @gnossen

lidizheng:
About the failing Bazel tests, maybe you can add the new dependency in https://github.com/grpc/grpc/blob/master/requirements.bazel.txt?

tools/distrib/python/grpcio_tools/setup.py
plannigan: I understand the concerns about adding a new dependency. There are a few reasons why I think this dependency is acceptable.
* [tomli](https://github.com/hukkin/tomli) is a very small library that keeps bloat in mind. It only supports reading toml files. Writing support is part of a [different library](https://github.com/hukkin/tomli-w).
* The requirement for the dependency goes away in Python 3.11 when the standard library adds `tomllib`. It will take years before it could be removed, but as time goes on, fewer people will be using the external dependency.
* `tomli` is the implementation that will be added to the standard library in Python 3.11. So there won't be a need to a compatibility layer on top or worrying about different implementation details.

Vendor-ing is an option. `tomli` is MIT licensed and the implementation is only 4 files and about 800 lines (including comments & blank lines).

I'd lean away from a wrapper package because `grpcio-tools` already contains `setuptools` specific functionality. Having `setuptools` functionality split between two packages could be confusing to users. So unless the intent would be to move that functionality out of `grpcio-tools` into this new package and introducing more work, I would consider this the least good option.

At the end of the day, I'm most interested in getting this feature accepted. So I'll defer the decision to people that work more closely with the project.

gnossen: Would it be possible to add a minimum version check of 3.11 for this feature and then just depend on the standard library toml implementation. This addresses our concerns about a [leftpad](https://www.davidhaney.io/npm-left-pad-have-we-forgotten-how-to-program/)-style disaster and also provides a well-known path for people who want to use the feature -- they just need to upgrade to 3.11.

If that's not acceptable, then how about making this an optional dependency for interpreters lower than 3.11? Then the burden of installing the toml parser would fall on the user who wants to use this new feature and has no possibility of affecting others.

lidizheng: Adding an extra dependency in the form of `grpcio-tools[toml]` sounds good to me.

gnossen: > I'd be open to making it an extra dependency. That way people that would like to use the functionality can either use 3.11 or install grpcio-tools[toml]

Sounds good to me. Thanks for bearing with us. We're pretty risk averse.

plannigan:
I made the changes to have `tomli` be optional.

The CI is failing, but it doesn't seem to be related to my changes. Does anyone have any ideas.

The next step would be to write documentation. Also, if anyone has comments on the config API, now would be a good time to talk about it in case we want to tweak it some.

plannigan:
I haven't had much time to work on this lately, but do intend to finish this.

PR #29542, opened 2022-04-30
Add NetBSD support (Community-supported)

0-wiz-0:
@donnadionne


drfloob:
@0-wiz-0 thanks for the contribution. NetBSD would be a community-supported platform with this addition, which entails a few things. The gRPC team doesn't have the resources to test or develop on NetBSD, so support may be broken at any time without us knowing, or without us being able to spend cycles fixing it. We're generally ok to allow these sorts of improvements, but that has to be taken into consideration.

As for changes to the PR: Going forward, we request a platform owner to add their contact information to `port_platform.h` so that in the event of a problem, it's clear who to reach out to. You should also update https://github.com/grpc/grpc/blob/master/src/cpp/README.md to add NetBSD to the list of community-supported platforms.

drfloob:
I'd recommend something like:

```
// NetBSD is a community-supported platform. Please contact Jane Doe <developer@example.com> for support.
```

FreeBSD support was added a good long while ago, and it's unfortunate there's no PoC for platform support. This new requirement is partially in response to some challenges there.

0-wiz-0:
I have added a comment as suggested, thanks.

drfloob:
Looks good, thanks for the contribution @0-wiz-0!

drfloob (approved):

PR #29555, opened 2022-05-02
Binder transport: Log endpoint binder object before passing it to client

sifmelcara:
This will help us identify connection establishment related issue.


sifmelcara:
(I created this PR while debugging b/229904921 , I was hoping we can get more info if that happened again. And now it indeed happens again b/241460651 )

sifmelcara:
This will help showing which C++ gRPC server instance is serving the request, when a client try to create new connection to gRPC server

LittleCVR (approved):

sifmelcara:
Build failures look irrelevant

PR #29562, opened 2022-05-03
Second attempt: health check service: rewrite using callback API

markdroth:
Original PR was #29457, reverted in #29559.

The first commit is a pure revert of the previous revert.  The other commits fix problems seen internally upon import.


markdroth:
I think this is now ready for review.

markdroth:
...Or maybe not.  Looks like there's a new windows test failure I need to figure out.

ctiller (approved):

PR #29566, opened 2022-05-03
Remove invalid validation for listener

costinm:
Current envoy proto has the 'address' field marked as required - and the auto-generated proto validators used in Istio will reject listeners without that field.

IMO the field should not be required - but oneof - but I suspect such a change will be difficult to make in the current proto.  

@markdroth

### DIFF
--- a/src/core/ext/xds/xds_listener.cc
+++ b/src/core/ext/xds/xds_listener.cc
@@ -976,10 +976,8 @@ grpc_error_handle LdsResourceParse(
       envoy_config_listener_v3_Listener_api_listener(listener);
   const envoy_config_core_v3_Address* address =
       envoy_config_listener_v3_Listener_address(listener);
-  if (api_listener != nullptr && address != nullptr) {
-    return GRPC_ERROR_CREATE_FROM_STATIC_STRING(
-        "Listener has both address and ApiListener");
-  }
+  // Current envoy proto has address as required, validators will reject if
+  // missing.
   if (api_listener == nullptr && address == nullptr) {
     return GRPC_ERROR_CREATE_FROM_STATIC_STRING(
         "Listener has neither address nor ApiListener");


markdroth:
I don't think this is the right solution.  I've sent envoyproxy/envoy#21132 to remove the PGV annotation in the xds proto file.

costinm:
It may remove it in the XDS proto - but existing control plane ( at least for Istio ) is using the current XDS proto.

I agree the right thing is to not have the field required - but grpc c++ implementation will remain broken (at least with any XDS servers that validate currently) until a release with the new proto will happen, and that also blocks any testing - we only tested with golang which doesn't have this check.

costinm:
Mark: I mean given the release cycles of Envoy, Istio, ASM - the proto change in Envoy plus the ASM change to not generate the (now required) address will take ~6 months where people won't be able to use on grpc c++, only go and java. With our current structure we can't generate invalid XDS, and picking new proto happens with some delay - and getting that into the managed XDS is even longer.

Maybe you can add the check back after the proto is changed and the current XDS implementations are updated ?

markdroth:
Why does it take so long for Istio to pick up the new xDS protos?  That seems like something we should improve.

Trying to fix this on the client seems like a worse fix, because it would mean that users would not be able to use any of the already-released gRPC versions that have this check.  I think we should fix it just once in Istio so that all versions of the gRPC client that are already in use will work.

sanjaypujare:
If this is a blocker for getting our C++ client to work with Istio then we need a quicker solution.

costinm:
Istio is not the only XDS server, and protocol changes - in particular not backward compatible - are expected to take time. 

Right now - with the proto defined by envoy in current releases - grpc c++ is rejecting valid protos and requires invalid one.

After your proto change, on the next release of envoy - grpc will be correct. When the next release gets trough rapid, regular, stable - it will start working. We have pretty clear stability requirements and compat.

sanjaypujare:
In addition to what @costinm said the other gRPC languages - Go and Java - do not reject this payload. Also by removing this validation I don't think we are introducing any regressions or backward incompatible changes.

The issue of users not being able to use an older gRPC version (after using the newer version which has relaxed a check) doesn't sound like an issue to me. Whereas not being able to use gRPC-C++ with Istio is a much bigger issue we need to worry about

markdroth:
Not every xDS server will actually use the PGV annotations for validation.  They're not really a required part of the protocol.

I don't see how the Envoy release cycle has anything to do with this.  The fact that the xDS protos happen to live in the Envoy repo is really just a historical anomaly; they are "owned" by the wider ecosystem, not by Envoy, and they don't really have a strict release cycle.  If you are depending on the Envoy release cycle to pick up changes in the protos, then your process is broken and needs to change.

Note that even if we merged this PR immediately, it would not be available to users until the 1.47 release goes out, so there would still be a release delay here.

That having been said, I will chat with Eric and Doug about this and see if we can come up with a better suggestion.

sanjaypujare:
Not sure where Envoy release cycle was mentioned but the one that matters more is the Istio one. I think they will need to make 2 changes:

- incorporate the latest xds proto
- enhance their implementation so the address being required is enforced for the proxy case but not for the proxyless case (not sure how Istio knows proxy vs proxyless - based on the annotation on the pod?)

markdroth:
Sanjay, Costin mentioned the Envoy release in his last comment above, and that's what I was responding to.

It looks like envoyproxy/envoy#21132 should get merged soon.  I think we should focus on getting this change into Istio.

> enhance their implementation so the address being required is enforced for the proxy case but not for the proxyless case (not sure how Istio knows proxy vs proxyless - based on the annotation on the pod?)

Istio already needs to create a different Listener resource for Envoy and gRPC, because the Envoy listeners are TCP listeners and the gRPC listeners are API listeners.  So I don't see why this will be hard; all they need to do is stop populating the address field in the TCP listener, which they should never have been populating in the first place.

sanjaypujare:
> ...all they need to do is stop populating the address field in the TCP listener, which they should never have been populating in the first place.

You mean stop populating the field in case of the API listener ?

In any case, Istio will not make the change unless a bug is opened in their repo and they are convinced that they need to fix it. @costinm could we do that?

markdroth:
Yes, sorry, I meant in the API listener.

sanjaypujare:
I have opened https://github.com/istio/istio/issues/38914 and hopefully that will be addressed soon.

markdroth:
I've added a comment to https://github.com/istio/istio/issues/38914 about what I think the real issue is here.  However, while we work on getting that resolved, it looks like we will need to go forward as a temporary workaround in the short term.

markdroth:
Closing this in favor of #29841.

PR #29568, opened 2022-05-03
Fix a segfault in client_lb_end2end_test

### LINKED ISSUE #29567: Segfault in client_lb_end2end_test due to absl::string_view(nullptr)

I am working on updating Fedora Linux’s `grpc` package to 1.46.0—currently using `v1.46.0-pre2`. The problem is in gRPC Core.

Fedora Linux Rawhide (development version).

```
gcc (GCC) 12.0.1 20220429 (Red Hat 12.0.1-0)
```

### What did you do?

I compiled `grpc` and ran the core tests.

I don’t expect you to easily reproduce my exact environment, but I can explain what is happening.

The tests pass.

### What did you see instead?

The test `client_lb_end2end_test` fails with `SIGSEGV`. The backtrace is attached:

[client_lb_end2end_test-gdb.log](https://github.com/grpc/grpc/files/8610324/client_lb_end2end_test-gdb.log)

A few key frames are:

```
Thread 1 "client_lb_end2e" received signal SIGSEGV, Segmentation fault.
0x00007ffff71b167d in __strlen_avx2 () from /lib64/libc.so.6
#0  0x00007ffff71b167d in __strlen_avx2 () from /lib64/libc.so.6
No symbol table info available.
#1  0x00007ffff783be7f in std::char_traits<char>::length (__s=0x0) at /usr/include/c++/12/bits/char_traits.h:395
No locals.
#2  std::basic_string_view<char, std::char_traits<char> >::basic_string_view (__str=0x0, this=<optimized out>) at /usr/include/c++/12/string_view:134
No locals.
#3  grpc_core::SubchannelStreamClient::SubchannelStreamClient (this=this@entry=0x7fffec011340, connected_subchannel=..., interested_parties=interested_parties@entry=0xdeafbeef, event_handler=std::unique_ptr<grpc_core::SubchannelStreamClient::CallEventHandler> = {...}, tracer=tracer@entry=0x0) at /builddir/build/BUILD/grpc-1.46.0-pre2/src/core/ext/filters/client_channel/subchannel_stream_client.cc:50
```

The problem begins in `grpc_core::MakeHealthCheckClient`, which forms the `trace` parameter for `grpc_core::MakeOrphanable<SubchannelStreamClient>` as follows:

https://github.com/grpc/grpc/blob/730599bd5b68f12436907cd96ab3fb20579734e7/src/core/ext/filters/client_channel/health/health_check_client.cc#L158-L160

With the trace flag disabled, this is simply `nullptr`. Now `MakeOrphanable` forwards the argument to `SubchannelStreamClient`.

https://github.com/grpc/grpc/blob/730599bd5b68f12436907cd96ab3fb20579734e7/src/core/lib/gprpp/orphanable.h#L69-L72

In the constructor for `SubchannelStreamClient`:

https://github.com/grpc/grpc/blob/730599bd5b68f12436907cd96ab3fb20579734e7/src/core/ext/filters/client_channel/subchannel_stream_client.cc#L42-L69

it is just fine that the `const char*` member `tracer_` is initialized with `nullptr`. There is even a test for this in the constructor body! But the problem is here:

https://github.com/grpc/grpc/blob/730599bd5b68f12436907cd96ab3fb20579734e7/src/core/ext/filters/client_channel/subchannel_stream_client.cc#L50-L53

`ResourceQuotaFromChannelArgs(…)->memory_quota()` returns a `MemoryQuota`, and its `CreateMemoryAllocator` method has the following signature:

https://github.com/grpc/grpc/blob/730599bd5b68f12436907cd96ab3fb20579734e7/src/core/lib/resource_quota/memory_quota.h#L432

Now the `nullptr` is used to construct an `absl::string_view`. I think I have read that this is OK in abseil; if this were `std::string_view`, it would invoke undefined behavior. Either way, at some point thereafter `strlen()` is called on the null pointer held by the `string_view`, and and the segmentation fault ensues.

One possible simple patch is to check for a null `string_view` and replace it with an non-null empty `string_view`:

```diff
From 93a7fe71e51413512fc352900a4d82850eed0488 Mon Sep 17 00:00:00 2001
From: "Benjamin A. Beasley" <code@musicinmybrain.net>
Date: Tue, 3 May 2022 12:59:51 -0400
Subject: [PATCH] Fix a segfault in client_lb_end2end_test

In the SubchannelStreamClient constructor, do not initialize an
absl::string_view with a null pointer; this lead to strlen() being
called on the null pointer. Let the absl::string_view be empty in this
case instead.

Fixes #TBD.
---
 .../ext/filters/client_channel/subchannel_stream_client.cc     | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/src/core/ext/filters/client_channel/subchannel_stream_client.cc b/src/core/ext/filters/client_channel/subchannel_stream_client.cc
index a596e04456..2e23f9a5c4 100644
--- a/src/core/ext/filters/client_channel/subchannel_stream_client.cc
+++ b/src/core/ext/filters/client_channel/subchannel_stream_client.cc
@@ -50,7 +50,8 @@ SubchannelStreamClient::SubchannelStreamClient(
       call_allocator_(
           ResourceQuotaFromChannelArgs(connected_subchannel_->args())
               ->memory_quota()
-              ->CreateMemoryAllocator(tracer)),
+              ->CreateMemoryAllocator(GPR_UNLIKELY(tracer_ != nullptr) ? tracer_
+                                                                       : "")),
       event_handler_(std::move(event_handler)),
       retry_backoff_(
           BackOff::Options()
-- 
2.35.1

```

It’s possible that there is a better or cleaner solution, but this patch fixes the segfault, and I plan to propose it as a PR.

This appears to be the only place where the `CreateMemoryAllocator` or `CreateMemoryOwner` method can be called with a null pointer or a nullptr-initialized `absl::string_view`, unless one is passed as the `peer_string` argument to `grpc_tcp_create()`.

musicinmybrain:
In the `SubchannelStreamClient` constructor, do not initialize an `absl::string_view` with a null pointer; this leads to `strlen()` being called on the null pointer. Let the `absl::string_view` be empty in this case instead.

Fixes #29567.

@markdroth


src/core/ext/filters/client_channel/subchannel_stream_client.cc
```diff
@@ -50,7 +50,8 @@ SubchannelStreamClient::SubchannelStreamClient(
       call_allocator_(
           ResourceQuotaFromChannelArgs(connected_subchannel_->args())
               ->memory_quota()
-              ->CreateMemoryAllocator(tracer)),
+              ->CreateMemoryAllocator(GPR_UNLIKELY(tracer_ != nullptr) ? tracer_
```
markdroth, line 53: No need for the `GPR_UNLIKELY()` here, since this isn't in a performance-sensitive code path.

If the tracer is null, then I suggest making the string something like "SubchannelStreamClient".

markdroth (commented):
Thanks for the contribution!

src/core/ext/filters/client_channel/subchannel_stream_client.cc
musicinmybrain: These are both reasonable suggestions. I’ll follow up with a force-push.

markdroth (approved):
Thanks, this looks good!  I've triggered the tests to run, and I'll merge this once everything looks good there.

nicolasnoble (approved):

markdroth:
The test failures look unrelated to the PR.

musicinmybrain:
Thanks!

PR #29476, opened 2022-04-21
[PSM interop] Introduce isort/pylint to the GKE framework

lidizheng:
This PR introduces the PSM GKE framework to:

* isort
* pylint

This PR also tries to fix the sanity complains if possible, but there were lots of places where the best fix is ambiguous: https://paste.googleplex.com/5988734773231616 (pylint has ~200 complains). In those ambiguous cases, this PR disables pylint, folks, please lift the disable comments during daily coding (boy-scout rule). Here are a few note-worthy ones:

* "arguments-differ"/"signature-differs": The child class overriding the parent class but not respecting parent's signature. Several cases under the "framework" folder.
* "abstract-method": We have abstract class depends on abstract class, where abstract methods weren't necessarily being implemented.
* "unsubscriptable-object":  See https://github.com/PyCQA/pylint/issues/3882

---

Our pytype sanity tests might need more work. Currently, we are using a 2019 release of pytype, which don't support Python > 3.7. On the other hand, the latest pytype don't support Python < 3.7. It requires more delicate testing to sort out what version of pytype + Python suits our infra.

Also, the latest pytype release found more issues with our existing code, the upgrade won't be easy.

PR #29577, opened 2022-05-04
xds-k8s: Fix incorrect type hint in the baseline test

sergiitk:
Minor fix: noticed test_servers should've been a `List[_XdsTestServer]`, but instead of fixing the type I renamed the variable from `test_servers` to `test_server` instead, and grabbed the first element.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/tests/baseline_test.py
+++ b/tools/run_tests/xds_k8s_test_driver/tests/baseline_test.py
@@ -44,16 +44,14 @@ class BaselineTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
         with self.subTest('4_create_forwarding_rule'):
             self.td.create_forwarding_rule(self.server_xds_port)
 
-        test_servers: _XdsTestServer
         with self.subTest('5_start_test_server'):
-            test_servers = self.startTestServers()
+            test_server: _XdsTestServer = self.startTestServers()[0]
 
         with self.subTest('6_add_server_backends_to_backend_service'):
             self.setupServerBackends()
 
-        test_client: _XdsTestClient
         with self.subTest('7_start_test_client'):
-            test_client = self.startTestClient(test_servers[0])
+            test_client: _XdsTestClient = self.startTestClient(test_server)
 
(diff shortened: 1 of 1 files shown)


sergiitk:
Did a manual test run of this single test, all passed.

lidizheng (approved):
Good catch!

tools/run_tests/xds_k8s_test_driver/tests/baseline_test.py
```diff
@@ -44,16 +44,14 @@ def test_traffic_director_grpc_setup(self):
...
-            test_servers = self.startTestServers()
+            test_server: _XdsTestServer = self.startTestServers()[0]
 
         with self.subTest('6_add_server_backends_to_backend_service'):
             self.setupServerBackends()
 
-        test_client: _XdsTestClient
         with self.subTest('7_start_test_client'):
-            test_client = self.startTestClient(test_servers[0])
+            test_client: _XdsTestClient = self.startTestClient(test_server)
```
lidizheng, line 54: Let's lift the declaration two lines up? Variable "test_client" is expected to be used later in other sub tests.

sergiitk: AFAIK `with` doesn't create variable scope (that's python's quirk) - so declarations inside `with` are equivalent to declarations outside.

sergiitk: Found this: https://stackoverflow.com/questions/45100271/scope-of-variable-within-with-statement

lidizheng: No strong opinion here. Asking this because https://github.com/grpc/grpc/pull/29476#discussion_r862011691.

sergiitk: Oh - that comment was referring to breaking it for readability - otherwise the line break made it less readable.

lidizheng: Then I completely misunderstood that 🤦  I thought it's about variable type annotation.

PR #29581, opened 2022-05-05
Change cancellation semantics in the iomgr DNSResolver

drfloob:
DNS requests were previously cancellable, but it was assumed
that the resolution callback would be called in all cases. Now, requests
provide a `bool Cancel(TaskHandle handle)` method that lets callers know
whether cancellation is successful, and if so, the callback will not be
run.

This is in accordance with EventEngine's cancellation semantics, and it
is a step towards migration from iomgr to EventEngine.

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/29581)


markdroth (changes requested):
This looks really good!  Comments are mostly minor; the only really noteworthy things are the one about removing the `DNSRequest` class from the iomgr API and the one about the race condition in the c-ares implementation.

I'd like @apolcyn to look this over as well, in case he can spot any edge cases in the code that I may have missed.

> Reviewed 7 of 20 files at r1, 4 of 14 files at r2, 18 of 18 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29581#-:-N1e-gCo7ScXG_vU18wI:b-ylt9g1)* status: all files reviewed, 11 unresolved discussions (waiting on @drfloob)

___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 347 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dwGK8AD8X_yAdtPoZ:-N1dwGK8AD8X_yAdtPo_:b-n6nnv4) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L347)):*
> ```cpp
>       GRPC_CLOSURE_INIT(&on_dns_lookup_done_, OnDnsLookupDone, this,
>                         grpc_schedule_on_exec_ctx);
>       ares_request_ = std::unique_ptr<grpc_ares_request>(grpc_dns_lookup_ares(
> ```

Are we guaranteed that `grpc_dns_lookup_ares()` will not invoke `on_dns_lookup_done_` in this thread before it returns?  Doing so could cause `OnDnsLookupDone()` to be invoked before `ares_request_` is actually set.

I think we do have this guarantee, because the existing code would have had the same problem, but please check.
___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 386 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1duiCBDBvTz4Pggb6K:-N1duiCBDBvTz4Pggb6L:btaegtz) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L386)):*
> ```cpp
>       GRPC_CARES_TRACE_LOG("AresRequest:%p OnDnsLookupDone", request);
>       // This request is deleted and unregistered upon any exit.
>       auto req_deleter = absl::MakeCleanup([request] { delete request; });
> ```

This can just be:

```
std::unique_ptr<AresRequest> deleter(request);
```
___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 429 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dth5g1q1tYhfuqKh3:-N1dth5g1q1tYhfuqKh4:bzfuche) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L429)):*
> ```cpp
>     // underlying ares_request that the query is performed on
>     std::unique_ptr<grpc_ares_request> ares_request_ ABSL_GUARDED_BY(mu_);
>     bool cancelled_ ABSL_GUARDED_BY(mu_);
> ```

I don't think we need both `cancelled_` and `ran_`, because we don't actually care which one happened first; all we care about is that one of the two has already happened.  So we can instead have just have one bool called `completed_`, and we can do something like this in both `Cancel()` and `OnDnsLookupDone()`:

```
if (completed_) return;
completed_ = true;
```
___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 449 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dxKywE-VIogDRT_JQ:-N1dxKywE-VIogDRT_JR:b-uwmpkr) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L449)):*
> ```cpp
>       std::function<void(absl::StatusOr<std::vector<grpc_resolved_address>>)>
>           on_done) override {
>     auto* request = new AresRequest(name, default_port, interested_parties,
> ```

I think this code will suffer from a race condition, because the `AresRequest` ctor will automatically start the DNS request, and that request could complete in a different thread before this thread acquires the mutex.  This means the completion could remove the entry from `open_requests_` before it is actually added, thus effectively resulting in a memory leak by accumulating entries in `open_requests_` that will never be removed, and possibly resulting in a crash if a subsequent cancellation attempts to access the request from that entry.

I think we need to do one of two things to solve this:

1. Change `AresRequest` to not start the request in the ctor but rather have a separate `Start()` method, just like it was prior to this PR.
2. Change the code here in `ResolveName()` to acquire the mutex before instantiating the `AresRequest` object.

Option 1 is arguably a bit more performant, but this isn't a performance-critical path.  Option 2 is probably simpler, especially since it would eliminate the need for `aba_token_` to be an atomic, since it would be accessed only while holding the mutex.  I'd lean toward option 2, but either one is fine with me.
___
*[`src/core/lib/http/httpcli.h` line 241 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dqGFaAn-iaJIpg79x:-N1dqGFaAn-iaJIpg79y:bqrqkwp) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/http/httpcli.h#L241)):*
> ```objc
>   bool own_endpoint_ ABSL_GUARDED_BY(mu_) = true;
>   bool cancelled_ ABSL_GUARDED_BY(mu_) = false;
>   bool ran_ ABSL_GUARDED_BY(mu_) = false;
> ```

Instead of adding a separate field here, I suggest just using `absl::optional<>` for the `dns_request_` field.  You can unset it when the callback runs.
___
*[`src/core/lib/http/httpcli.h` line 251 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dqWqh4BWAciHoWSqC:-N1dqWqh4BWAciHoWSqD:bqmt2va) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/http/httpcli.h#L251)):*
> ```objc
>   grpc_slice_buffer outgoing_ ABSL_GUARDED_BY(mu_);
>   grpc_error_handle overall_error_ ABSL_GUARDED_BY(mu_) = GRPC_ERROR_NONE;
>   DNSResolver::TaskHandle dns_request_ ABSL_GUARDED_BY(mu_) =
> ```

Suggest renaming this to `dns_request_handle_`.
___
*[`src/core/lib/iomgr/resolve_address.h` line 46 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1ddCXY6_V_luzC_fe9:-N1ddCXZ4kUWxrD30zS1:b-biez38) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/resolve_address.h#L46)):*
> ```objc
>   using TaskHandle = ::grpc_event_engine::experimental::EventEngine::
>       DNSResolver::LookupTaskHandle;
>   static constexpr TaskHandle NULL_HANDLE{0, 0};
> ```

This should be named `kNullHandle`.
___
*[`src/core/lib/iomgr/resolve_address.h` line 50 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1ddeYp-QZqD8TPK5a_:-N1ddeYp-QZqD8TPK5aa:b-vc54o2) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/resolve_address.h#L50)):*
> ```objc
>   // Tracks a single asynchronous DNS resolution attempt. DNS resolution should
>   // begin upon construction.
>   class Request {
> ```

I think we don't need this class in this API at all anymore.  The reason for having it previously was that it basically was the task handle exposed to the caller, but now that we're exposing that as `TaskHandle`, I don't think the caller needs this.

Note that not all of the implementations actually use this (the EE-based iomgr does not use it at all), and the implementations that do use this don't get any benefit from inheritting from this base class, because the implementation contains both the subclass definition and the code that calls the only method on this interface (`Cancel()`), so there's no benefit in defining that method here.
___
*[`src/core/lib/iomgr/resolve_address_posix.cc` line 95 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1djBFp2PHqHCErtGtw:-N1djBFp2PHqHCErtGtx:b8jxoyq) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/resolve_address_posix.cc#L95)):*
> ```cpp
>   // self-deleting class
>   new NativeDNSRequest(name, default_port, std::move(on_done));
>   return DNSResolver::NULL_HANDLE;
> ```

No need for `DNSResolver::` prefix, since we're in that class here.
___
*[`src/core/lib/iomgr/event_engine/resolver.h` line 43 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1djeIS45JjtP0QQiqc:-N1djeIS45JjtP0QQiqd:b5qk2ja) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/event_engine/resolver.h#L43)):*
> ```objc
>   static EventEngineDNSResolver* GetOrCreate();
> 
>   explicit EventEngineDNSResolver(
> ```

This ctor should be private, since it should be used only by `GetOrCreate()`.

That having been said, if you take option 2 from my suggestion below, this ctor would just go away.
___
*[`src/core/lib/iomgr/event_engine/resolver.h` line 59 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dnAHe-gYNlxx7uMu5:-N1dnAHe-gYNlxx7uMu6:b1bgfao) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/event_engine/resolver.h#L59)):*
> ```objc
> 
>  private:
>   std::unique_ptr<EventEngine::DNSResolver> ee_resolver_;
> ```

I don't think we can have just one `EventEngine::DNSResolver` object that we use for all resolution requests, because the client_channel c-ares DNS resolver supports specifying the DNS server to use via the URI authority, but in the EE API, we decided to set that when we instantiate the `DNSResolver` object:

https://github.com/grpc/grpc/blob/a7a960e44619082a3a6faecda50d2131a184eef5/include/grpc/event_engine/event_engine.h#L265

In order to move the c-ares code inside of iomgr, we'll need to do one of two things:

1. Change the iomgr DNS API to require the caller to create a new `DNSResolver` object before it can start any queries (i.e., make the iomgr API more like the EE API).
2. Keep the code here creating a new `EventEngine::DNSResolver` for each individual query (i.e., instantiate the `EventEngine::DNSResolver` when `ResolveName()` is called), as it was doing before this PR.

The second option is probably easier, but either approach is fine with me.

drfloob (commented):
Thanks. I've addressed everything except the EventEngine changes. Given that the `GRPC_USE_EVENT_ENGINE` code does not build on master (the memory allocator changes were never ported), I'm tempted to revert the changes to these EventEngine-specific files and address this code separately.

> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29581#-:-N1eRJKuA_CR5eMptuT8:bqwo8mm)* status: all files reviewed, 11 unresolved discussions (waiting on @drfloob and @markdroth)

___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 347 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dwGK8AD8X_yAdtPoZ:-N1eJH_u4-iO1tEXteJ0:b8aldoc) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L347)):*

Are we guaranteed that `grpc_dns_lookup_ares()` will not invoke `on_dns_lookup_done_` in this thread before it returns?  Doing so could cause `OnDnsLookupDone()` to be invoked before `ares_request_` is actually set.

I think we do have this guarantee, because the existing code would have had the same problem, but please check.
</blockquote></details>

It's a new problem, good catch. Previously a lock was acquired before creating the request. I've added that back.
___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 386 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1duiCBDBvTz4Pggb6K:-N1e3qqMC7xBw__AUYrU:bb6un1j) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L386)):*

This can just be:

```
std::unique_ptr<AresRequest> deleter(request);
```
</blockquote></details>

Done
___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 429 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dth5g1q1tYhfuqKh3:-N1e442KC953XtWNDBC7:br1hgw7) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L429)):*

I don't think we need both `cancelled_` and `ran_`, because we don't actually care which one happened first; all we care about is that one of the two has already happened.  So we can instead have just have one bool called `completed_`, and we can do something like this in both `Cancel()` and `OnDnsLookupDone()`:

```
if (completed_) return;
completed_ = true;
```
</blockquote></details>

Done. If we had left it as is and retained a separate `cancelled_` flag, we'd be able to detect double-cancellation bugs, but double cancellation is UB so your suggestion is valid.
___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 449 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dxKywE-VIogDRT_JQ:-N1eKeqv2CFag2kXw-II:b-utamh9) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L449)):*

I think this code will suffer from a race condition, because the `AresRequest` ctor will automatically start the DNS request, and that request could complete in a different thread before this thread acquires the mutex.  This means the completion could remove the entry from `open_requests_` before it is actually added, thus effectively resulting in a memory leak by accumulating entries in `open_requests_` that will never be removed, and possibly resulting in a crash if a subsequent cancellation attempts to access the request from that entry.

I think we need to do one of two things to solve this:

1. Change `AresRequest` to not start the request in the ctor but rather have a separate `Start()` method, just like it was prior to this PR.
2. Change the code here in `ResolveName()` to acquire the mutex before instantiating the `AresRequest` object.

Option 1 is arguably a bit more performant, but this isn't a performance-critical path.  Option 2 is probably simpler, especially since it would eliminate the need for `aba_token_` to be an atomic, since it would be accessed only while holding the mutex.  I'd lean toward option 2, but either one is fine with me.
</blockquote></details>

Opted for option 2. Good catch.
___
*[`src/core/lib/http/httpcli.h` line 241 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dqGFaAn-iaJIpg79x:-N1e6XiR-vAHW1xCT6bk:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/http/httpcli.h#L241)):*

Instead of adding a separate field here, I suggest just using `absl::optional<>` for the `dns_request_` field.  You can unset it when the callback runs.
</blockquote></details>

Done.
___
*[`src/core/lib/http/httpcli.h` line 251 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dqWqh4BWAciHoWSqC:-N1e52yJD0WQFtSNS8ky:bb6un1j) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/http/httpcli.h#L251)):*

Suggest renaming this to `dns_request_handle_`.
</blockquote></details>

Done
___
*[`src/core/lib/iomgr/resolve_address.h` line 46 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1ddCXY6_V_luzC_fe9:-N1eAJPCEn600N9dgupC:bwppva0) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/resolve_address.h#L46)):*

This should be named `kNullHandle`.
</blockquote></details>

Done. Thanks for the reminder.
___
*[`src/core/lib/iomgr/resolve_address.h` line 50 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1ddeYp-QZqD8TPK5a_:-N1eRGWw1iAhn1cJoPjS:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/resolve_address.h#L50)):*

I think we don't need this class in this API at all anymore.  The reason for having it previously was that it basically was the task handle exposed to the caller, but now that we're exposing that as `TaskHandle`, I don't think the caller needs this.

Note that not all of the implementations actually use this (the EE-based iomgr does not use it at all), and the implementations that do use this don't get any benefit from inheritting from this base class, because the implementation contains both the subclass definition and the code that calls the only method on this interface (`Cancel()`), so there's no benefit in defining that method here.
</blockquote></details>

Done.
___
*[`src/core/lib/iomgr/resolve_address_posix.cc` line 95 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1djBFp2PHqHCErtGtw:-N1eBcqbAJAhVzwl4A5h:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/resolve_address_posix.cc#L95)):*

No need for `DNSResolver::` prefix, since we're in that class here.
</blockquote></details>

Done.

drfloob (commented):
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29581)* status: 9 of 30 files reviewed, 11 unresolved discussions (waiting on @markdroth)

___
*[`src/core/lib/iomgr/event_engine/resolver.h` line 43 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1djeIS45JjtP0QQiqc:-N1eUumS18bAlqNtTKNt:b-onqk6z) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/event_engine/resolver.h#L43)):*

This ctor should be private, since it should be used only by `GetOrCreate()`.

That having been said, if you take option 2 from my suggestion below, this ctor would just go away.
</blockquote></details>

As discussed, I reverted these changes, and I'm going to delete the custom EventEngine-driven iomgr impl separately.
___
*[`src/core/lib/iomgr/event_engine/resolver.h` line 59 at r3](https://reviewable.io/reviews/grpc/grpc/29581#-N1dnAHe-gYNlxx7uMu5:-N1eV2Bo5fwW3gqWHIRR:b-onqk6z) ([raw file](https://github.com/grpc/grpc/blob/f3c1886d24c210360b27382b73c4019f6858dde9/src/core/lib/iomgr/event_engine/resolver.h#L59)):*

I don't think we can have just one `EventEngine::DNSResolver` object that we use for all resolution requests, because the client_channel c-ares DNS resolver supports specifying the DNS server to use via the URI authority, but in the EE API, we decided to set that when we instantiate the `DNSResolver` object:

https://github.com/grpc/grpc/blob/a7a960e44619082a3a6faecda50d2131a184eef5/include/grpc/event_engine/event_engine.h#L265

In order to move the c-ares code inside of iomgr, we'll need to do one of two things:

1. Change the iomgr DNS API to require the caller to create a new `DNSResolver` object before it can start any queries (i.e., make the iomgr API more like the EE API).
2. Keep the code here creating a new `EventEngine::DNSResolver` for each individual query (i.e., instantiate the `EventEngine::DNSResolver` when `ResolveName()` is called), as it was doing before this PR.

The second option is probably easier, but either approach is fine with me.
</blockquote></details>

As discussed, I reverted these changes, and I'm going to delete the custom EventEngine-driven iomgr impl separately.

markdroth (approved):
Just one minor cosmetic comment left, otherwise looks great!

> Reviewed 21 of 21 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29581#-:-N1f-R-ZCYKXdvqzAqdw:bz2ilt1)* status: all files reviewed, 1 unresolved discussion (waiting on @drfloob)

___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 377 at r4](https://reviewable.io/reviews/grpc/grpc/29581#-N1ezsG5Bzq9IbYcBOxp:-N1ezsG5Bzq9IbYcBOxq:b-rn5mjj) ([raw file](https://github.com/grpc/grpc/blob/9569aa2a40121868c416b973e5b2c7a957ad4cd4/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L377)):*
> ```cpp
>           completed_(false),
>           resolver_(resolver),
>           aba_token_(resolver->aba_token_++) {
> ```

Suggest doing this increment in the parent and passing the ABA token in here as a parameter, since that will make the logic a bit easier to understand (e.g., it will be clearer from reading this that the increment is being done while holding the parent's lock).

drfloob (commented):
Excellent, thanks for the review.

> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29581#-:-N1f2P9W3YB-f8Y6r9ef:b-10q6si)* status: all files reviewed, 1 unresolved discussion (waiting on @markdroth)

___
*[`src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc` line 377 at r4](https://reviewable.io/reviews/grpc/grpc/29581#-N1ezsG5Bzq9IbYcBOxp:-N1f2Lox3Q9mS_R6-jJ1:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/9569aa2a40121868c416b973e5b2c7a957ad4cd4/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L377)):*

Suggest doing this increment in the parent and passing the ABA token in here as a parameter, since that will make the logic a bit easier to understand (e.g., it will be clearer from reading this that the increment is being done while holding the parent's lock).
</blockquote></details>

Done.

src/core/lib/http/httpcli.cc
```diff
@@ -410,16 +407,16 @@ void HttpRequest::OnResolved(
...
-    Finish(absl_status_to_grpc_error(addresses_or.status()));
-    return;
-  }
+  dns_request_handle_.reset();
   if (cancelled_) {
     Finish(GRPC_ERROR_CREATE_FROM_STATIC_STRING(
         "cancelled during DNS resolution"));
     return;
   }
+  if (!addresses_or.ok()) {
```
apolcyn, line 416: just curious why this was moved down below the check for `cancelled_`?

src/core/lib/http/httpcli.cc
```diff
@@ -206,22 +202,23 @@ void HttpRequest::Start() {
...
 }
 
 void HttpRequest::Orphan() {
   {
     MutexLock lock(&mu_);
     GPR_ASSERT(!cancelled_);
     cancelled_ = true;
-    dns_request_.reset();  // cancel potentially pending DNS resolution
-    if (connecting_) {
-      // gRPC's TCP connection establishment API doesn't currently have
```
apolcyn, line 219: nit: I think this comment about TCP connect cancellation is still relevant for the `else if (connecting_)` clause.

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
```diff
@@ -358,74 +362,83 @@ class AresClientChannelDNSResolverFactory : public ResolverFactory {
...
 
-    ~AresRequest() override {
+    ~AresRequest() {
       GRPC_CARES_TRACE_LOG("AresRequest:%p dtor ares_request_:%p", this,
                            ares_request_.get());
+      resolver_->UnregisterRequest(task_handle());
     }
 
-    void Start() override {
+    bool Cancel() {
```
apolcyn, line 399: nit: `Cancel() override` ?

apolcyn (commented):
LGTM with some minor comments

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
drfloob: The Request base class is gone, no override required.

src/core/lib/http/httpcli.cc
drfloob: If the request was cancelled, I imagine the caller should not usually be concerned if the request failed.

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
```diff
@@ -473,9 +496,30 @@ class AresDNSResolver : public DNSResolver {
     return default_resolver_->ResolveNameBlocking(name, default_port);
   }
 
+  bool Cancel(TaskHandle handle) override {
+    MutexLock lock(&mu_);
+    if (!open_requests_.contains(handle)) {
+      // Unknown request, possibly completed already, or an invalid handle.
```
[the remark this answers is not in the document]
drfloob: I'll add this in a followup PR.

src/core/lib/http/httpcli.cc
drfloob: I'll add this change in a subsequent PR.

drfloob:
Thanks @apolcyn. I missed Mark's bit where he requested your review as well, I will follow up shortly.

src/core/lib/iomgr/resolve_address.h
```diff
@@ -39,37 +41,40 @@ constexpr int kDefaultSecurePortInt = 443;
...
-  // were previously held while calling Request::Start(). Therefore,
-  // implementations must not invoke \a on_done inline from the call to
-  // Request::Start(). The DNSCallbackExecCtxScheduler utility may help address
-  // this.
-  virtual OrphanablePtr<Request> ResolveName(
+  // were previously held while starting the request. Therefore,
+  // implementations must not invoke \a on_done inline from the call site that
+  // starts the request. The DNSCallbackExecCtxScheduler utility may help
+  // address this.
+  virtual TaskHandle ResolveName(
```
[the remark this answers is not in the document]
drfloob: Excellent find, I was able to write a test for the poll engine that reproduces the lock failure. I'll add a fix as we discussed offline. Thanks Alex!

PR #29589, opened 2022-05-05
Requiring C++14

veblush:
Bumping the version of C++ from 11 to 14 to support C++14. Once this lands, we can use C++14 features throughout gRPC code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,12 +222,12 @@ if (NOT DEFINED CMAKE_C_STANDARD)
   set(CMAKE_C_STANDARD 99)
 endif()
 
-# Add c++11 flags
+# Add c++14 flags
 if (NOT DEFINED CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD 14)
 else()
-  if (CMAKE_CXX_STANDARD LESS 11)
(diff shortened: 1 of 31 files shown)


ctiller:
Do we know why the bloat increase?

veblush:
> Do we know why the bloat increase?

I briefly checked the result and the delta of bloaty test is 8728 bytes for `libgrpc.so` and 8316 bytes for `libgrpc++.so` which ended up exceeding the threshold of `bloat/medium` (16K) slightly. 

From my reading on the `.text` section by comparing disassembled code, I don't think there is any significant regression here. It's mostly coming from the different behavior of compiler. (e.g. c++11 chose to "inline" a certain function but c++14 didn't, or vice versa) Code generation is slightly changed according to the compiler version (maybe some STL code has different code paths depending on the C++ standard), and this seems to have influenced the decision of whether or not to be inlined.

veblush:
[grpc_build_artifacts_multiplatform](https://fusion2.corp.google.com/invocations/2fc1c872-835d-44a9-842b-866764de2b9a): running

ctiller (approved):

markdroth (approved):
Awesome, thanks for doing this!

veblush:
Existing flaky tests:
- Basic Tests C/C++ MacOS: b/TBD
- Bazel RBE UBSAN C/C++: Timeout "test/core/end2end/h2_tls_test@resource_quota_server@poller=poll" b/151212019

PR #29593, opened 2022-05-05
Clean Bazel caches before testing a specific Bazel version

lidizheng:
As title.

CC @gnossen


veblush:
Note that `--cache_test_results=no` works too. Which one is better?

lidizheng:
I'm not sure if this flag ask Bazel to rebuild from scratch or not. Based on the [doc](https://docs.bazel.build/versions/main/command-line-reference.html#flag--cache_test_results) of this flag:

> If set to 'auto', Bazel reruns a test if and only if: (1) Bazel detects changes in the test or its dependencies, (2) the test is marked as external, (3) multiple test runs were requested with --runs_per_test, or(4) the test previously failed. If set to 'yes', Bazel caches all test results except for tests marked as external. If set to 'no', Bazel does not cache any test results.

It didn't mention build. If the binaries were reused or intermediate products from dependencies were reused, we might still not be able to find the hard-bazel-restriction issue.

veblush:
> I'm not sure if this flag ask Bazel to rebuild from scratch or not. Based on the [doc](https://docs.bazel.build/versions/main/command-line-reference.html#flag--cache_test_results) of this flag:
> 
> > If set to 'auto', Bazel reruns a test if and only if: (1) Bazel detects changes in the test or its dependencies, (2) the test is marked as external, (3) multiple test runs were requested with --runs_per_test, or(4) the test previously failed. If set to 'yes', Bazel caches all test results except for tests marked as external. If set to 'no', Bazel does not cache any test results.
> 
> It didn't mention build. If the binaries were reused or intermediate products from dependencies were reused, we might still not be able to find the hard-bazel-version-restriction-in-our-dependency issue.

Fair enough. Then let's use clean command. And please make sure that bazel 3.x test fails with this change. Then we decide to either drop 3.x or fix 3.x build issue.

lidizheng:
(The Bazel Distribtest job in this PR should fail: https://source.cloud.google.com/results/invocations/adce44b2-800f-4bfd-92b5-2c99d6d364fe/targets.)

lidizheng:
```
+ bazel test --test_output=all //:all
Loading:
Loading: 0 packages loaded
Analyzing: 3 targets (0 packages loaded, 0 targets configured)
INFO: Analyzed 3 targets (0 packages loaded, 0 targets configured).
INFO: Found 2 targets and 1 test target...
[0 / 2] [Prepa] BazelWorkspaceStatusAction stable-status.txt
INFO: Elapsed time: 0.361s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
PASSED: //:greeter_test (see /root/.cache/bazel/_bazel_root/054c9929c19606854b88b1d062320e1a/execroot/bazel_cpp_distribtests/bazel-out/k8-fastbuild/testlogs/greeter_test/test.log)
INFO: From Testing //:greeter_test
==================== Test output for //:greeter_test:
+ SERVER_PID=
+ SERVER_TIMEOUT=10
++ mktemp
+ SERVER_OUTPUT=/tmp/tmp.NcyLNXczqL
+ trap cleanup SIGINT SIGTERM EXIT
+ SERVER_PID=27535
+ ./greeter_server
++ await_server
++ TIME=0
++ '[' '!' -s /tmp/tmp.NcyLNXczqL ']'
++ '[' 0 == 10 ']'
++ sleep 1
++ TIME=1
++ '[' '!' -s /tmp/tmp.NcyLNXczqL ']'
++ cat /tmp/tmp.NcyLNXczqL
+ SERVER_ADDRESS=127.0.0.1:45043
++ ./greeter_client --target=127.0.0.1:45043
+ RESPONSE='Greeter received: Hello world'
+ EXPECTED_RESPONSE='Greeter received: Hello world'
+ '[' 'Greeter received: Hello world' '!=' 'Greeter received: Hello world' ']'
+ echo Success.
Success.
+ cleanup
+ '[' -n 27535 ']'
+ kill 27535
//:greeter_test                                                 (cached) PASSED in 1.0s

Executed 0 out of 1 test: 1 test passes.
```

I'm wrong. Cleaning the Bazel won't invalid the test cache...

veblush:
I suspect that this distrib test has its own [WORKSPACE](https://github.com/grpc/grpc/blob/master/test/distrib/bazel/cpp/WORKSPACE) giving an independent bazel workspace so bazel clean and overriding bazel version may not work.

lidizheng:
(Didn't merge master, still expect the Bazel distrib test to fail.)

test/distrib/bazel/test_single_bazel_version.sh
```diff
@@ -51,12 +51,15 @@ FAILED_TESTS=""
...
+# clean the caches and downloaded archives to make sure we build from scratch
+tools/bazel clean --expunge
+tools/bazel build -- //... "${EXCLUDED_TARGETS[@]}" || FAILED_TESTS="${FAILED_TESTS}Build "
+# tools/bazel build -- //:grpc "${EXCLUDED_TARGETS[@]}" || FAILED_TESTS="${FAILED_TESTS}Build "
 
 for TEST_DIRECTORY in "${TEST_DIRECTORIES[@]}"; do
   pushd "test/distrib/bazel/$TEST_DIRECTORY/"
 
-  bazel test --test_output=all //:all || FAILED_TESTS="${FAILED_TESTS}${TEST_DIRECTORY} Distribtest"
+  tools/bazel test --cache_test_results=no --test_output=all //:all || FAILED_TESTS="${FAILED_TESTS}${TEST_DIRECTORY} Distribtest"
```
veblush, line 62: pwd is changed so tools/bazel might not work here.

lidizheng: There is a copied `tools/bazel` at https://github.com/grpc/grpc/blob/7942f30f6301f5baf93ea242d13433349640fc57/test/distrib/bazel/cpp/tools/bazel

lidizheng:
I added a check to verify the Bazel version without `tools/bazel`, it appears that Bazel isn't always picking up the automatic tools/bazel script.

```
+ bazel version
+ grep -q 5.0.0
Extracting Bazel installation...
Starting local Bazel server and connecting to it...
++ bazel version
+ echo 'Incorrect Bazel version! Want=5.0.0 Seen=Build label: 4.2.1
```

lidizheng:
Hmm... Maybe the `bazel clean` needs to happen at per-WORKSPACE level, because the Bazel cache is stored that way?

test/distrib/bazel/test_single_bazel_version.sh
```diff
@@ -51,12 +51,31 @@ FAILED_TESTS=""
...
+BAZEL="$(pwd)/tools/bazel"
+
+# validate the Bazel version
+if ! ($BAZEL version | grep -q "${VERSION}"); then
+  echo "Incorrect Bazel version! Want=${VERSION} Seen=$($BAZEL version)"
+  exit 1
+fi
+
+# clean the caches and downloaded archives to make sure we build from scratch
+$BAZEL clean --expunge
```
jtattermusch, line 65: Requiring "bazel clean" seems speculative. Do we really have a clear evidence that this is necessary?
I'm pretty sure bazel is not stupid and it recognizes that if you change the bazel version, a full rebuild is needed.

test/distrib/bazel/test_single_bazel_version.sh
```diff
@@ -51,12 +51,31 @@ FAILED_TESTS=""
 export OVERRIDE_BAZEL_VERSION="$VERSION"
 # when running under bazel docker image, the workspace is read only.
 export OVERRIDE_BAZEL_WRAPPER_DOWNLOAD_DIR=/tmp
-bazel build -- //... "${EXCLUDED_TARGETS[@]}" || FAILED_TESTS="${FAILED_TESTS}Build "
+
+# pick the Bazel wrapper
```
jtattermusch, line 55: change comment to something more useful.
E.g. # bazel distribtest don't run from the repository root, so we need to explicitly use the `tools/bazel` wrapper to run the right bazel version for us

gnossen:
@lidizheng Can you please catch me up on what led to this change?

> Requiring "bazel clean" seems speculative. Do we really have a clear evidence that this is necessary? I'm pretty sure bazel is not stupid and it recognizes that if you change the bazel version, a full rebuild is needed.

I agree with Jan here. If this is not how Bazel behaves, I'm pretty surprised.

lidizheng:
@jtattermusch @gnossen Feel free to propose a better change to supersede this PR. With the added flags, cleans, the Bazel distributed test still can't reproduce the error:

```
ERROR: Traceback (most recent call last):
	File "/usr/local/google/home/lidiz/src/grpc/test/distrib/bazel/python/WORKSPACE", line 17, column 16, in <toplevel>
		grpc_extra_deps()
	File "/usr/local/google/home/lidiz/.cache/bazel/_bazel_lidiz/afb910f9aa3930959bbd962aee59a5f6/external/com_github_grpc_grpc/bazel/grpc_extra_deps.bzl", line 55, column 26, in grpc_extra_deps
		go_rules_dependencies()
	File "/usr/local/google/home/lidiz/.cache/bazel/_bazel_lidiz/afb910f9aa3930959bbd962aee59a5f6/external/io_bazel_rules_go/go/private/repositories.bzl", line 39, column 23, in go_rules_dependencies
		versions.check(MINIMUM_BAZEL_VERSION, bazel_version = native.bazel_version)
	File "/usr/local/google/home/lidiz/.cache/bazel/_bazel_lidiz/afb910f9aa3930959bbd962aee59a5f6/external/io_bazel_rules_go/go/private/skylib/lib/versions.bzl", line 104, column 13, in _check_bazel_version
		fail("\nCurrent Bazel version is {}, expected at least {}\n".format(
Error in fail:
Current Bazel version is 3.7.2, expected at least 4.2.1
```

In my local runs, I can directly invoke `test/distrib/bazel/run_bazel_distrib_test.sh` to test 5.0.0, 4.2.2 and 3.7.2, then reproduce this error. However, on Kokoro, it's unclear why it won't trigger. In the latest run, Kokoro job picks up 3.7.2 correctly, cleaned, yet no error is generated: https://source.cloud.google.com/results/invocations/8812a16c-4435-4794-9057-a995e4e76182/targets

test/distrib/bazel/test_single_bazel_version.sh
```diff
@@ -51,12 +51,31 @@ FAILED_TESTS=""
...
+
+# test building all targets
+$BAZEL build -- //... "${EXCLUDED_TARGETS[@]}" || FAILED_TESTS="${FAILED_TESTS}Build "
 
 for TEST_DIRECTORY in "${TEST_DIRECTORIES[@]}"; do
   pushd "test/distrib/bazel/$TEST_DIRECTORY/"
 
-  bazel test --test_output=all //:all || FAILED_TESTS="${FAILED_TESTS}${TEST_DIRECTORY} Distribtest"
+  # validate the Bazel version again, since we have a different WORKSPACE file
+  if ! ($BAZEL version | grep -q "${VERSION}"); then
```
gnossen, line 74: This is duplicated. You probably want to factor out a function for this.

lidizheng:
I would probably investigate more later this week, I agree with both of you that Bazel should be smart enough to detect change of environment then invalidate build caches. If build cache is not the culprit of the missing coverage, something else is...

test/distrib/bazel/test_single_bazel_version.sh
```diff
@@ -51,12 +51,31 @@ FAILED_TESTS=""
 export OVERRIDE_BAZEL_VERSION="$VERSION"
 # when running under bazel docker image, the workspace is read only.
 export OVERRIDE_BAZEL_WRAPPER_DOWNLOAD_DIR=/tmp
-bazel build -- //... "${EXCLUDED_TARGETS[@]}" || FAILED_TESTS="${FAILED_TESTS}Build "
+
+# pick the Bazel wrapper
+BAZEL="$(pwd)/tools/bazel"
```
gnossen, line 56: This now creates an undocumented requirement that this script be run from a particular directory. I think it'd be better if we could avoid that.

lidizheng (commented):
Sorry for dragging this PR for 2 weeks. I thought I can quickly find a solution, then patch this PR. But after experiments in https://github.com/grpc/grpc/pull/29751 and https://github.com/grpc/grpc/pull/29669. I don't think I yet to grasp why the downloaded repositories are still cached. Tried:

- Pick a fixed Bazel wrapper
- Ensure the Bazel version is correct
- Clean --expunge
- --cache_test_results=no
- Explicitly set --repository_cache to empty
- Shutdown and reboot Bazel server

None of the above methods made Bazel Distribtests Job fail (and locally it aways fail). I'm going to close this PR, yield the rights of digging into this problem.

PR #29607, opened 2022-05-06
Add an end-to-end test that includes a client-side fork.

bdhess:
@veblush - I tried to boil down the issue I was describing to you into a reproducible test case, and it turns out everything works fine at head, we just had an order-of-operations issue.  This test case covers our usage.

### DIFF
--- a/test/cpp/end2end/BUILD
+++ b/test/cpp/end2end/BUILD
@@ -191,6 +191,30 @@ grpc_cc_binary(
 )
 
 grpc_cc_test(
+    name = "client_fork_test",
+    size = "small",
+    srcs = ["client_fork_test.cc"],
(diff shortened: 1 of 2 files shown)


veblush (approved):
Sorry for the late review!

veblush:
To get this merged, two things needs to be done;

 1. run tools/buildgen/generate_projects.sh
 2. address `test/cpp/end2end/client_fork_test.cc:18:#include <pthread.h>` from `tools/run_tests/sanity/cpp_banned_constructs.sh`. you may want to add a new exception there.

veblush:
and xSAN tests failed so those errors need to be fixed.

drfloob:
I've continued this work in #31035

PR #22444, opened 2020-03-23
Split logs by test case and run client separately for each test

ericgribkoff:
This separates the logs for each test case into separate targets for display on dashboards. This doesn't absolutely require running a new client process for each test case but it seems to make sense, and it makes it much easier tell which logs correspond to which test case when there's not a single client running continuously throughout.

Also added support for specifying a list of test cases which is useful for tracking down/avoiding flaky tests for manual and potentially automated runs.

Please ignore the second commit, which will be reverted and is just to trigger a Kokoro test run on this PR.

PR #29618, opened 2022-05-09
remove duplicate ./reports/ mount for xds tests

jtattermusch:
Tentative fix for b/231142315

Combination of https://github.com/grpc/grpc/pull/29468 and https://github.com/grpc/grpc/pull/22444/commits/d22e89f7fc4ae5d530f9b23e44df17687615c4da has broken the XDS tests.

Even though `export OUTPUT_DIR=reports` was originally introduced to enable uploading for files under reports/ to GCS (and to diplay the sponge_log.xml logs in test UI), it seems that this option is currently unused in all the XDS tests impacted (I checked the test results before the breakage and there seems to be no sponge_log.xml files being uploaded even before). Besides that, setting OUTPUT_DIR=reports seems strictly speaking both unnecessary and incorrect
- https://github.com/grpc/grpc/pull/29468 introduces better mechanisms for obtaining the sponge_log.xml files from inside the docker container
- the "OUTPUT_DIR" technique in the build_and_run_docker.sh script was actually intended for uploading artifacts, not reports (reports are handled automatically).

I think we should really think of having some XDS some tests running on PRs as a smoke test (e.g. run a single noop test inside the XDS test setup), since currently it's way too easy to break them.

### DIFF
--- a/tools/internal_ci/linux/grpc_xds.sh
+++ b/tools/internal_ci/linux/grpc_xds.sh
@@ -22,5 +22,4 @@ source tools/internal_ci/helper_scripts/prepare_build_linux_rc
 
 export DOCKERFILE_DIR=tools/dockerfile/test/bazel
 export DOCKER_RUN_SCRIPT=$BAZEL_SCRIPT
-export OUTPUT_DIR=reports
 exec tools/run_tests/dockerize/build_and_run_docker.sh
--- a/tools/internal_ci/linux/grpc_xds_php.sh
+++ b/tools/internal_ci/linux/grpc_xds_php.sh
@@ -22,5 +22,4 @@ source tools/internal_ci/helper_scripts/prepare_build_linux_rc
 
 export DOCKERFILE_DIR=tools/dockerfile/test/php73_zts_debian11_x64
 export DOCKER_RUN_SCRIPT=tools/internal_ci/linux/grpc_xds_php_test_in_docker.sh
-export OUTPUT_DIR=reports
 exec tools/run_tests/dockerize/build_and_run_docker.sh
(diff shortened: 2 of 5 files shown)


jtattermusch:
@lidizheng  please doublecheck that this change doesn't remove any XDS test results reports unintentionally (I checked and it should be fine, but I don't know XDS tests in depth).

lidizheng:
Triggered xDS tests:
* [grpc_xds](https://fusion2.corp.google.com/invocations/4ac20753-775f-494b-aaba-8b0f6fe4df25/targets)
* [grpc_xds_url_map](http://sponge/0f807e31-65a8-400b-8bc7-bbf2b793660f)
* [grpc_xds_url_map_python](http://sponge/017b1277-086c-4b82-a8ed-55bf0a7d1704)
* [grpc_xds_k8s_lb](http://sponge/dbbb0b96-2a2d-4519-8f99-385c0a473ad2)
* [grpc_xds_k8s_lb_python](http://sponge/ae523a03-1deb-41b4-9a08-49a23b4178df)
* [grpc_xds_k8s_xlang](http://sponge/d3987390-1363-461e-babd-41a23b5a0130)
* [psm-security](http://sponge/e16c967d-3f1b-4176-bfd2-6d3250fe2e1f)
* [psm-security-python](http://sponge/aad44d42-92c4-4943-af74-08b2695d1e92)

sergiitk:
@lidizheng Note that the new framework isn't affected by this. So only [grpc_xds](https://fusion2.corp.google.com/invocations/4ac20753-775f-494b-aaba-8b0f6fe4df25/targets) run is relevant.

sergiitk (approved):
LGTM.

PR #29619, opened 2022-05-09
Upload results from bazel C/C++ ARM64 tests to bigquery

jtattermusch:
as title.

Access to historical data is invaluable for deflaking/flakiness tracking.


jtattermusch:
Adhoc run of the job is green (and it has uploaded results to bigquery correctly).

veblush (approved):

PR #29620, opened 2022-05-09
Make objC InteropTests and MacTests buildable with bazel

jtattermusch:
Without this, building :InteropTest or :MacTests with bazel fails with missing `TestBase.h`.

### DIFF
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -101,6 +101,7 @@ grpc_objc_testing_library(
     hdrs = ["InteropTests/InteropTests.h"],
     deps = [
         ":InteropTestsBlockCallbacks-lib",
+        ":TestBase-lib",
     ],
 )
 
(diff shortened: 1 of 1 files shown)


jtattermusch:
@dennycd  please doublecheck what I'm doing, I'm quite new to objC (and building ObjC with bazel).

dennycd (approved):
Thanks for patching this Jan.  The change looks good to me. 

cc @HannahShiSFB  to make sure there is no conflcting changes from  https://github.com/grpc/grpc-ios/issues/75

src/objective-c/tests/BUILD
```diff
@@ -101,6 +101,7 @@ grpc_objc_testing_library(
     hdrs = ["InteropTests/InteropTests.h"],
     deps = [
         ":InteropTestsBlockCallbacks-lib",
+        ":TestBase-lib",
```
dennycd, line 104: one NIT: we may also add the dep to https://github.com/grpc/grpc/blob/48749f739e7dd91cb69c988d57a89ffe93d8b501/src/objective-c/grpc_objc_internal_library.bzl#L112  (similar to :RemoteTest target) as this seems to be a common base dep for all test tagets. But this can be in a follow up ;  )

jtattermusch:
Adhoc run of grpc_objc_bazel_test: http://sponge/7a7af2a9-e9c3-412b-8783-a2cdd231c2fe

src/objective-c/tests/BUILD
jtattermusch: Let's do as a followup.

PR #29621, opened 2022-05-09
server: per-rpc backend metric reporting

nicolasnoble:
Implement the server API for per-RPC backend metric reporting, as described in grpc/proposal#289.

This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/29621)


nicolasnoble:
Green and ready to go.

nicolasnoble (commented):
I've been preparing a slide deck, yes. I can mark the main entry point as experimental too.

> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621#-:-N1uMaf8AJKHvjyqYJ-P:b-291g1y)* status: all files reviewed, 21 unresolved discussions (waiting on @markdroth and @yashykt)

___
*[`BUILD` line 2942 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o87b29wpfFwnPoj7l:-N1uMqVV83h0yxP-yUrL:brgq1k0) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/BUILD#L2942)):*

Suggest moving this file to src/core/ext/filters/client_channel/lb_policy, since it will eventually need to be exposed as part of the LB policy API when we make that API public.
</blockquote></details>

Sure. Done.
___
*[`BUILD` line 5365 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o8Rw82lsmQuWlqJQI:-N1uQ1qrACqpVAO4S2uY:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/BUILD#L5365)):*

Suggest calling this `grpcpp_orca_interceptor`.
</blockquote></details>

Done.
___
*[`include/grpcpp/call_metric_recorder.h` line 22 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o3KTH3XibH7Dvk2Cq:-N1uQ7m05Mna9fbNnKzm:bg6n1l7) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/call_metric_recorder.h#L22)):*

Does the definition actually need to be in the impl/codegen directory?  I'm not sure if any of our generated code actually needs to refer to any of this.
</blockquote></details>

Uuuuuh you're right. I started needing it there because of ServerContext, but that changed over time. I guess dropping it to basedir/ext should work. Putting it there unless you have a different suggestion.

Shuffled files around.
___
*[`include/grpcpp/server_builder.h` line 409 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o9fEGEK9X2_lMLjIP:-N1uSUkQ5XrAwrHdDjl3:bdvvsp3) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/server_builder.h#L409)):*

I'm not sure whether or not this suggestion makes sense, but just to make sure we've considered it:

Do we actually need a separate list of interceptor creators for this?  If I'm understanding correctly, it seems like the reason for this is to put these "internal" interceptors first.  Could we instead maybe provide some mechanism for ordering the list of interceptors?

@yashykt can probably provide some useful input here, since he has more context than I do on the interceptor APIs.
</blockquote></details>

The discussion I had with Yash is that the ordering problem is complex, and left to the user, in the form of a single "Set" of the whole vector of interceptors.

You're correct in your understanding, but this is precisely to avoid providing an ordering mechanism, as originally designed with the Interceptor API.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 44 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oNHSc4qDsMlFrl5g5:-N1uTa_6Ax0IWvgJPEMy:b-80tlqk) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L44)):*

Let's put this in the `experimental` namespace to start with.  We can graduate it from experimental later once it's proven stable.
</blockquote></details>

Sure.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 44 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o4bSt1nDnPex4U4EN:-N1udATHBX709KbSASSd:b-drtkmw) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L44)):*

How about making this an interface (i.e., all methods would be pure virtual), so that we don't need to expose implementation details here?

For example, we would not need to refer to the `BackendMetricData` struct here at all; that kind of detail would move to the implementation, which would not need to be exposed in public headers.  And I think that would also allow us to pull in the definition of `BackendMetricData`, so we could make it a direct member of the `CallMetricRecorder` impl instead of dynamically allocating it.

I think this would also avoid the need to make `OrcaServerInterceptor` a friend.  Instead, we could expose `CreateSerializedReport()` as a public method on the implementation (not part of the interface at all), and `OrcaServerInterceptor` could directly down-cast to the implementation type.
</blockquote></details>

Myeah, it's a bit of a CPU<-->memory tradeoff here. We'd be eating the cost of the virtual calls every time for every metric to record, instead of eating the cost of the allocation once at the beginning. Not sure which one's better.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 59 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o5HZG1D9EpkYuFd01:-N1uV-Pf8r9X3_aoSIY2:b-fasqu6) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L59)):*

We don't actually support the RPS field, so we don't need this method.
</blockquote></details>

Ah! Removing.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 64 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o5Q1jEqWb-hPuKNCC:-N1uVXeuBOVKqH1cL-_P:biw8y1v) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L64)):*

For the moment, we're still preferring to use `grpc::string_ref` instead of `absl::string_view` in our public APIs, for consistency.  (Ugh.)

Same for `RecordRequestCostMetric()` below.
</blockquote></details>

Right. Done.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 64 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o5ZFI0ilfUdoRtdp1:-N1uWZ0eAFV6tFmG0Sud:b-80tlqk) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L64)):*

I think we can pass the name in by value instead of as a const reference.

Same for `RecordRequestCostMetric()` below.
</blockquote></details>

Sure.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 64 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oJL_d5Ly-hbHNb80T:-N1uXLvq1bRF8eGcvJHo:b-quzkcw) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L64)):*

We need to document the lifetime and ownership semantics for the metric names.

It looks like the current implementation essentially just stores the passed-in `string_view`, which means that the caller is responsible for ensuring that the passed-in strings live at least until the interceptor serializes the proto.  I suspect that the common case here will be that an application will have a fixed set of metric names, in which case this approach is fine: the application can just define global string constants and use them here, so there is no lifetime issue.  However, I don't know if there will be corner cases where the strings will need to be more dynamically generated.

Just as a sanity check, maybe check to see how this is used today in stubby?  If there are ~no cases where the metric names are dynamically generated, then that will confirm that this approach is fine.  But if it turns out that there are a lot of cases where they are dynamically generated, then we might want to consider providing better support for that case.

One simple approach would be to provide an alternate version of this method that creates a copy of the string on the call arena.  Applications that need to pass dynamically generated strings for which they cannot easily manage the lifetime can choose to use that alternate method instead of this one.

If we don't have an easy way to determine how common this is today in stubby, then I think for now we can just assume that this is fine as-is.  If it becomes a problem, we can always add that alternative method later.

This comment also applies to `RecordRequestCostMetric()`.
</blockquote></details>

Right, my assumption was the same as you: the names are probably global constants (which caused me a tiny little bit of pain in the end2end test since the names actually come from the proto we're sending). If we're keeping this in experimental however, this gives us liberty to change it later, and I can simply document it here for now.

Note that the primary factor for me to chose this approach was the existing BackendMetricData that I only moved around, and that was using string_views to function, so I was simply assuming this was where we wanted to be.

I am not sure which Stubby API you're talking about however.
___
*[`include/grpcpp/impl/codegen/server_context.h` line 295 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o44i-6EaraVeRnMDf:-N1uYoYz7K8GJRyjMRy6:b-m0h1be) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/server_context.h#L295)):*

How about having this return a pointer instead of a reference, so that it can return null when the interceptor is not present?  We can have the interceptor create the object and set it on the `ServerContext` at the start of the call.  That way, we don't need to allocate the `CallMetricRecorder` at all when it's not actually going to be used, and there will be a clear error signal to the application if they call this without having first called `RegisterCallMetricLoadReporter()` on the `ServerBuilder`.
</blockquote></details>

My thinking here is that this shouldn't be an illegal usage to call this while the interceptor isn't present. That's also what the Java implementation does: they let the user record metrics even if the interceptor hasn't been registered, and they just dump the data on the floor silently.
___
*[`include/grpcpp/impl/codegen/server_context.h` line 295 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oPdtE7LysqvYuFJz7:-N1uZOVc-xbISE2ruMMt:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/server_context.h#L295)):*

For now, let's call this `ExperimentalGetCallMetricRecorder()`.  We can rename it later once it proves stable.
</blockquote></details>

Done.
___
*[`include/grpcpp/impl/codegen/server_context.h` line 403 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oOvTL2oH4duXb0WN2:-N1uZUtB5q0F0jxL5sR9:b-rn6gbf) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/server_context.h#L403)):*

Doesn't look like this is actually needed.  I don't see anywhere that the factory actually reaches inside of this class, just the interceptor itself.
</blockquote></details>

You're correct, that's leftovers.
___
*[`include/grpcpp/impl/codegen/server_context.h` line 482 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o8oTTA4jSYgiQwn8j:-N1udSb9-ajyjiEOTPuu:bi2buws) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/server_context.h#L482)):*

If possible, it would be nice to allocate this on the call arena instead of creating a separate allocation for it.  I'm thinking maybe we can have the interceptor allocate the object on the call arena at the start of the call and then set this field (which would be a raw pointer instead of a `unique_ptr<>`) to point to it.  We'd presumably also need `ServerContextBase` to explicitly call this object's dtor before the call arena is destroyed.
</blockquote></details>

I can look into that.
___
*[`src/core/ext/filters/common/backend_metric_data.h` line 39 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oAmD64bZHxw1-fqg1:-N1u_1bv3vX_caMiFt2_:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/core/ext/filters/common/backend_metric_data.h#L39)):*

Let's go ahead and remove this pre-existing field, since we don't actually support it.  I probably added this before we decided that we weren't going to support it.
</blockquote></details>

Done.
___
*[`src/cpp/server/orca/call_metric_recorder.cc` line 58 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oShWg3IiD5QqFlP2D:-N1u_K69-R2GihDs_3By:bn985y6) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/call_metric_recorder.cc#L58)):*

It looks like the purpose of this field is to not record backend metric data after the proto has actually been serialized.  But it's not clear to me that we really get benefit from this check, since (a) if we're properly ordering the interceptors, this should never actually happen, and (b) we're not actually providing any API feedback to the application about this error anyway, so the actual behavior would not be any different if we didn't do this.  Given that, I think we can probably just remove it.
</blockquote></details>

I was basically copying the behavior of the Java implementation to be fair :) I was assuming they had a good reason for it, and I didn't really try and question it. I can remove this, yes.
___
*[`src/cpp/server/orca/orca_interceptor.h` line 32 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oBYuQ5lLoMveHSug8:-N1u_jUF5JWRd2LQEgqB:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/orca_interceptor.h#L32)):*

Nit: Please initialize `info_` via an initializer list instead of in the body.
</blockquote></details>

Done.
___
*[`src/cpp/server/orca/orca_interceptor.h` line 49 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oBFob4nJwewWcQO3q:-N1u_kFsCa9ClSU_mSa_:b-dwqrir) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/orca_interceptor.h#L49)):*

Please add `// GRPC_INTERNAL_CPP_ORCA_ORCA_INTERCEPTOR_H`.
</blockquote></details>

Huh? That's supposed to be done automatically by the sanity fixer. I'll let it run again.
___
*[`src/cpp/server/orca/orca_interceptor.cc` line 25 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oCPNq3g6QwO8a2fxW:-N1ucp2GAaCorY_7AcMp:b-e95x85) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/orca_interceptor.cc#L25)):*

Suggest wrapping this in namespace blocks, so that we don't need to explicitly use a prefix each function:

```
namespace grpc {

namespace experimental {

void OrcaServerInterceptor::Intercept(...) { ... }

Interceptor* OrcaServerInterceptorFactory::CreateServerInterceptor(...) { ...}

void OrcaServerInterceptorFactory::Register(...) { ... }

}  // namespace experimental

void RegisterCallMetricLoadReporter(...) { ... }

}  // namespace grpc
```
</blockquote></details>

Yeah I usually dislike doing this. Personal taste, really. My brain works better with the full names. I can still do it if you prefer.
___
*[`src/cpp/server/orca/orca_interceptor.cc` line 36 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oRcz47V6JTCJhJMy-:-N1ua2QY3wt9WJotHoo4:b-1eh3wi) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/orca_interceptor.cc#L36)):*

Not a new problem in this PR, but it looks like we're using the wrong metadata key here.  We're currently using `x-endpoint-load-metrics-bin`, which was the name we originally intended to use, but we subsequently changed the design to just use `endpoint-load-metrics-bin` (no `x-` prefix).  Can you please update that as part of this PR?
</blockquote></details>

Sure, done.
___
*[`include/grpcpp/ext/orca_load_reporter.h` line 30 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oOcS-3zVRUYw2WWNx:-N1uT6OQBjvp6MgP7hlg:b-1eh3wi) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/ext/orca_load_reporter.h#L30)):*

Let's put this in the `experimental` namespace for now.  We can graduate it later when it's proven stable.
</blockquote></details>

Sure, done.
___
*[`include/grpcpp/ext/orca_load_reporter.h` line 30 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oQRAE4TgcdtTAdpC9:-N1uT7u98NDB4-GSLUrI:b-7p92s2) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/ext/orca_load_reporter.h#L30)):*

Can we put this in the same header file as `CallMetricRecorder`?  They're really both part of the same API, so I think it makes sense to keep them together.
</blockquote></details>

Well, my gut feeling was that the `CallMetricRecorder` can be used in the service implementation, and it doesn't need to know about the registration, while the registration happens during server building, which doesn't need to know about the recorder. I don't mind merging the two, but it feels more idiomatic in a "include what you use" way IMHO. I may be too pedantic.

markdroth (changes requested):
This is moving in the right direction!  Let's see what the Java folks say about the two issues that were based on their impl.

> Reviewed 8 of 26 files at r8, 20 of 20 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621#-:-N1y3fjP3dkyXvTsq9O3:bnszb6f)* status: all files reviewed, 9 unresolved discussions (waiting on @ejona86, @nicolasnoble, and @YifeiZhuang)

___
*[`include/grpcpp/server_builder.h` line 409 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o9fEGEK9X2_lMLjIP:-N1y1W32B9-LtoXEDdQW:bopchcc) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/server_builder.h#L409)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

The discussion I had with Yash is that the ordering problem is complex, and left to the user, in the form of a single "Set" of the whole vector of interceptors.

You're correct in your understanding, but this is precisely to avoid providing an ordering mechanism, as originally designed with the Interceptor API.
</blockquote></details>

Okay, sounds good.  Thanks for looking into this!
___
*[`include/grpcpp/impl/codegen/server_context.h` line 295 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o44i-6EaraVeRnMDf:-N1xydk04a6yuOOsDAoL:bb9ty3b) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/server_context.h#L295)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

My thinking here is that this shouldn't be an illegal usage to call this while the interceptor isn't present. That's also what the Java implementation does: they let the user record metrics even if the interceptor hasn't been registered, and they just dump the data on the floor silently.
</blockquote></details>

@ejona86 or @yifeizhuang, was there a reason that Java's API behaves that way?  It seems inefficient to allocate something we don't actually need if we can avoid it (although maybe that's less costly in Java?).  But perhaps more importantly, it also seems bad to not indicate to the application that the data they're providing won't actually be used.

No matter what we decide here, I think we should add a comment here indicating that this will not actually send any of the metric data if `RegisterCallMetricLoadReporter()` was not called on the `ServerBuilder`.
___
*[`include/grpcpp/impl/codegen/server_context.h` line 482 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o8oTTA4jSYgiQwn8j:-N1xtLEi46ShHtIF__wj:bmoo8ev) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/server_context.h#L482)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

I can look into that.
</blockquote></details>

I think it should be as simple as:

```
grpc_core::Arena* arena = grpc_call_get_arena(call_.call);
call_metric_recorder_ = arena->New<experimental::CallMetricRecorder>();
```
___
*[`src/cpp/server/orca/call_metric_recorder.cc` line 58 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oShWg3IiD5QqFlP2D:-N1xzoM2DE08JVmFnA-w:b23mf73) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/call_metric_recorder.cc#L58)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

I was basically copying the behavior of the Java implementation to be fair :) I was assuming they had a good reason for it, and I didn't really try and question it. I can remove this, yes.
</blockquote></details>

@ejona86 or @yifeizhuang, is there a reason that the Java impl tracks when the data has already been serialized and does not add subsequent data to the recorder once that's happened?  Either way, the data won't be used after that point, and we're not returning an error back to the application, so it's not clear what the benefit is of avoiding adding it to the recorder.
___
*[`src/cpp/server/orca/orca_interceptor.h` line 49 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oBFob4nJwewWcQO3q:-N1y-ejk47apA9qMbd1D:b4oecix) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/orca_interceptor.h#L49)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

Huh? That's supposed to be done automatically by the sanity fixer. I'll let it run again.
</blockquote></details>

Doesn't look like it's adding it for some reason.
___
*[`src/cpp/server/orca/orca_interceptor.cc` line 25 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oCPNq3g6QwO8a2fxW:-N1y-qaB1znKAPjtVa4Q:bt771a1) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/orca_interceptor.cc#L25)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

Yeah I usually dislike doing this. Personal taste, really. My brain works better with the full names. I can still do it if you prefer.
</blockquote></details>

Yes, please do.  We use namespaces throughout the code base, so we should be consistent here.

Please make the same change in src/cpp/server/orca/call_metric_recorder.cc.

Thanks!
___
*[`include/grpcpp/ext/orca_load_reporter.h` line 30 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oQRAE4TgcdtTAdpC9:-N1xsvmX2jNSunaRzm8e:bunapsy) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/ext/orca_load_reporter.h#L30)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

Well, my gut feeling was that the `CallMetricRecorder` can be used in the service implementation, and it doesn't need to know about the registration, while the registration happens during server building, which doesn't need to know about the recorder. I don't mind merging the two, but it feels more idiomatic in a "include what you use" way IMHO. I may be too pedantic.
</blockquote></details>

I don't think adding one function definition would really much overhead from an IWYU perspective.  But my concern here is really about making it easy for someone who first discovers this API to understand how it works.  I think that having both parts in the same header file would make that a bit clearer.

That having been said, I'll leave this up to you.
___
*[`include/grpcpp/ext/orca_load_reporter.h` line 32 at r9](https://reviewable.io/reviews/grpc/grpc/29621#-N1y2-_Z5AkfUQdoZXMk:-N1y2-_Z5AkfUQdoZXMl:bodb8le) ([raw file](https://github.com/grpc/grpc/blob/19a06d1644ac71aed37112e75e4431e97e704c7b/include/grpcpp/ext/orca_load_reporter.h#L32)):*
> ```objc
> // call the \a ServerContext::ExperimentalGetCallMetricRecorder() method to
> // retrieve the recorder for the current call.
> void RegisterCallMetricLoadReporter(ServerBuilder*);
> ```

Suggest calling this function `EnableCallMetricRecording()`.  The metrics might not actually have anything to do with load, so having "load" in the name seems confusing.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 44 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o4bSt1nDnPex4U4EN:-N1y0MyGAheiCyATi8Co:bfhqwaa) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L44)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

Myeah, it's a bit of a CPU<-->memory tradeoff here. We'd be eating the cost of the virtual calls every time for every metric to record, instead of eating the cost of the allocation once at the beginning. Not sure which one's better.
</blockquote></details>

You're certainly right that there could be some workloads that do better in either approach.  I was mainly thinking about the code organization advantages, though (e.g., not needing to pull in the definition of `BackendMetricData` and not needing to make `OrcaServerInterceptor` a friend).

I'll leave this up to you.  But if we do keep it this way, we should probably also use the call arena to allocate `BackendMetricData` instead of doing a separate allocation for it.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 64 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oJL_d5Ly-hbHNb80T:-N1y0pZ-0WGnxxN1M_cz:bf4ykmp) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L64)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

Right, my assumption was the same as you: the names are probably global constants (which caused me a tiny little bit of pain in the end2end test since the names actually come from the proto we're sending). If we're keeping this in experimental however, this gives us liberty to change it later, and I can simply document it here for now.

Note that the primary factor for me to chose this approach was the existing BackendMetricData that I only moved around, and that was using string_views to function, so I was simply assuming this was where we wanted to be.

I am not sure which Stubby API you're talking about however.
</blockquote></details>

Okay, sounds like we're on the same page.  For now, let's just add a comment indicating that the caller is responsible for keeping the passed-in strings alive until after trailing metadata is sent.

We can talk offline about investigating the Stubby APIs.

markdroth (changes requested):
Just a quick response to the last round of comments.  I still like the direction this is heading in!

Please let me know if you have any questions. Thanks!

> Reviewed 6 of 6 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621#-:-N1ypyef5W0-Ithe4LVJ:b6bt32q)* status: all files reviewed, 6 unresolved discussions (waiting on @ejona86, @nicolasnoble, and @YifeiZhuang)

___
*[`src/cpp/server/orca/call_metric_recorder.cc` line 58 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oShWg3IiD5QqFlP2D:-N1yo4Mh4HejCgEJZnw0:brkelx6) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/call_metric_recorder.cc#L58)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

My guess is it works in concert with the *other* discussion about allowing data to be recorded despite no interceptor: this allows to preemptively drop data without storing it by constructing a disabled recorder.
</blockquote></details>

I could understand that possibly being the goal here, but unless I'm missing something (which is certainly possible), I don't think this implementation actually works that way.  I don't see any place right now where the code actually constructs a disabled recorder.  It looks to me like the only place we set `disabled_` to true is right here, where we serialize the proto.  But in the case where the interceptor is not being used, this method will never be called.  So I think the only time this will actually be disabled is when both (a) the interceptor is used and (b) we have already serialized the proto into metadata.

In any case, let's see what the Java folks can tell us.
___
*[`include/grpcpp/ext/orca_load_reporter.h` line 30 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oQRAE4TgcdtTAdpC9:-N1yp3or3bHui656YX7Y:b-ucfknh) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/ext/orca_load_reporter.h#L30)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

You may be right about API discovery. I'll move this.
</blockquote></details>

Thanks!
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 64 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oJL_d5Ly-hbHNb80T:-N1ynmTFDoin0aXx2Ykv:b-4kgemb) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L64)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

Yep, I've already altered the comments.
</blockquote></details>

I don't see the comment change.  Maybe I'm looking in the wrong place?  I was expecting to see a comment on these two methods indicating that the application must keep the name string alive until after the trailing metadata is sent.

ejona86:
> @ejona86 or @YifeiZhuang, was there a reason that Java's API behaves that way? It seems inefficient to allocate something we don't actually need if we can avoid it (although maybe that's less costly in Java?). But perhaps more importantly, it also seems bad to not indicate to the application that the data they're providing won't actually be used.

We wanted a snapshot-like behavior while also not making a copy. Since the results won't be seen, it was an internal optimization. The synchronization in the implementation is not precise; it is possible for races. We'd probably need to make the synchronization stronger if it were to cause errors.

"won't actually be used" is also a case that is possible if there is no metric recorder in the Context as well. I agree this stuff could produce errors, and I don't remember how seriously we considered errors in these cases. I think part of the thought was that we really didn't want applications writing error handling for this stuff.

markdroth (commented):
Nico, I chatted offline about this with Eric to better understand why Java's impl is the way it is, and I think I now have a better understanding of what parts of that make sense for C++ and what parts don't.

My main take-aways:
- We completely forgot about synchronization in this PR.  The methods of the `CallMetricRecorder` may be invoked from multiple threads at the same time, so we need to add a mutex inside of `CallMetricRecorder` to guard access to all of the data members.
- In Java, creation of the maps inside of `CallMetricRecorder` is an allocation, so they want to avoid doing that if the data is not going to be used.  That's the main reason for having the `disabled` bit; they dynamically allocate the maps and then do a trick with atomics to snapshot the maps when populating the proto.  That trick won't really work in a non-GC language like C++, so it doesn't help us -- but we also don't need to avoid instantiating the maps in the first place.  So I don't think we need `disabled` for this reason.
- The `ServerContext::ExperimentalGetCallMetricRecorder()` method may also be invoked by the application in multiple threads at the same time, so we need that to be thread-safe.  In Java, that method does not do any allocation, so it's inherently thread-safe.  The `CallMetricRecorder` is allocated by the interceptor at the start of the call.  In `ServerContext::ExperimentalGetCallMetricRecorder()`, if the interceptor was not added, then the `CallMetricRecorder` is null, so they instead return a global singleton instance that is disabled at construction time (this is the second reason for `disabled` in Java), so all of its methods are no-ops.  That approach won't really work that well for C++, because we [can't easily create a global singleton](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables) (there are some work-arounds, but nothing really elegant).  So I think that in order for us to have this always return non-null, we would need to both do a new allocation and add synchronization -- and I'd really prefer to avoid that for performance reasons.
- Eric and I talked about the philosophy behind why we might or might not want `ServerContext::ExperimentalGetCallMetricRecorder()` to indicate to the caller that the interceptor is not in use.  There may be cases where the application wants to know that, such as doing some application-specific error handling if the application expects that the interceptor is configured.  There may also be some cases where the application is okay with the interceptor not being configured, but in that case it may want to avoid doing some expensive work to construct the metric data that it was going to add.  Eric is going to think about maybe adding a second API in Java that does indicate to the caller that the interceptor is not configured.  He does like the idea that some applications may be able to just use the existing API without caring whether or not it's set, but it's not entirely clear how useful that will actually be in practice, and it doesn't seem like something that C++ can easily do without a performance penalty.

So, taking this all into account, I propose the following changes:
- In `CallMetricRecorder`, remove the `disabled_` field, since it doesn't actually help us; the two things that Java is doing with it don't really apply to C++.
- In `CallMetricRecorder`, add a mutex to guard the data members.
- Change the interceptor to allocate the `CallMetricRecorder` at call start time and stash it in the `ServerContext`.
- Change `ServerContext::ExperimentalGetCallMetricRecorder()` to return the `CallMetricRecorder`, which will be null if the interceptor is not running.

If we are sure that the interceptor is run before any other interceptor, then I think we don't need synchronization for the `CallMetricRecorder` pointer inside the `ServerContext`, because it will be set before the application tries to access it.

Please let me know if you have any questions about any of this.  Thanks!

> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621#-:-N1zlLFA9Onn2BgNx-JU:b-k9xdvg)* status: all files reviewed, 6 unresolved discussions (waiting on @ejona86, @nicolasnoble, and @YifeiZhuang)

nicolasnoble (commented):
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621)* status: all files reviewed, 6 unresolved discussions (waiting on @ejona86, @markdroth, @nicolasnoble, and @YifeiZhuang)

___
*[`include/grpcpp/impl/codegen/server_context.h` line 482 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o8oTTA4jSYgiQwn8j:-N2A1s4U23cfNFag70EC:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/server_context.h#L482)):*

I think it should be as simple as:

```
grpc_core::Arena* arena = grpc_call_get_arena(call_.call);
call_metric_recorder_ = arena->New<experimental::CallMetricRecorder>();
```
</blockquote></details>

Done.
___
*[`src/cpp/server/orca/call_metric_recorder.cc` line 58 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oShWg3IiD5QqFlP2D:-N2A1wkY7sdvhNRx1oXu:bbsyj1l) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/call_metric_recorder.cc#L58)):*

I could understand that possibly being the goal here, but unless I'm missing something (which is certainly possible), I don't think this implementation actually works that way.  I don't see any place right now where the code actually constructs a disabled recorder.  It looks to me like the only place we set `disabled_` to true is right here, where we serialize the proto.  But in the case where the interceptor is not being used, this method will never be called.  So I think the only time this will actually be disabled is when both (a) the interceptor is used and (b) we have already serialized the proto into metadata.

In any case, let's see what the Java folks can tell us.
</blockquote></details>

Right, I'm not constructing one right now, but the Java implementation actually does. We could construct one in the case there's no interceptor that is. The only trouble is finding the interceptor exactly. If we get rid of this however then that's not really a matter.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 44 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1o4bSt1nDnPex4U4EN:-N2A2gtQ82hfVsbVBlGi:b787c0i) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L44)):*

You're certainly right that there could be some workloads that do better in either approach.  I was mainly thinking about the code organization advantages, though (e.g., not needing to pull in the definition of `BackendMetricData` and not needing to make `OrcaServerInterceptor` a friend).

I'll leave this up to you.  But if we do keep it this way, we should probably also use the call arena to allocate `BackendMetricData` instead of doing a separate allocation for it.
</blockquote></details>

I guess we can leave it this way for now and alter it if we discover the typical usage case is more expensive this way. I can change to the arena allocator, yes.
___
*[`include/grpcpp/impl/codegen/call_metric_recorder.h` line 64 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oJL_d5Ly-hbHNb80T:-N1zI0fd1M_1Xelj1GvK:bhhdbex) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/include/grpcpp/impl/codegen/call_metric_recorder.h#L64)):*

I don't see the comment change.  Maybe I'm looking in the wrong place?  I was expecting to see a comment on these two methods indicating that the application must keep the name string alive until after the trailing metadata is sent.
</blockquote></details>

Derp. I must've borked it while moving the API around for the other change. I'll re-add the wording.

nicolasnoble (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621#-:-N2FA4NQCyMU18TmSRuP:b-896fix)* status: 20 of 38 files reviewed, 6 unresolved discussions (waiting on @ejona86, @markdroth, @nicolasnoble, and @YifeiZhuang)

nicolasnoble (commented):
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621)* status: 19 of 38 files reviewed, 6 unresolved discussions (waiting on @ejona86, @markdroth, and @YifeiZhuang)

___
*[`src/cpp/server/orca/orca_interceptor.cc` line 25 at r7](https://reviewable.io/reviews/grpc/grpc/29621#-N1oCPNq3g6QwO8a2fxW:-N2OL9BMCZGxAY7g5dkR:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/c8b3c5233f8d32b6f6fa4f20fb917688fd2e5bae/src/cpp/server/orca/orca_interceptor.cc#L25)):*

Yes, please do.  We use namespaces throughout the code base, so we should be consistent here.

Please make the same change in src/cpp/server/orca/call_metric_recorder.cc.

Thanks!
</blockquote></details>

Done.

nicolasnoble (commented):
Everything should've been addressed now.

> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621#-:-N2Oa5xc3wRJ1Cdmya2w:bfq03tg)* status: 33 of 38 files reviewed, 4 unresolved discussions (waiting on @markdroth)

markdroth (approved):
This looks great!

> Reviewed 5 of 5 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/29621#-:-N2RmdkM4d7RiVUKOIo1:b-j45s0z)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nicolasnoble)

___
*[`src/cpp/server/orca/call_metric_recorder.cc` line 75 at r14](https://reviewable.io/reviews/grpc/grpc/29621#-N2OLfYQ4OjiyD9OMZMb:-N2RlaUxBtc4vUza0Gdr:b41trsx) ([raw file](https://github.com/grpc/grpc/blob/412592672622d709aed56c5fd26ab007124f0bc4/src/cpp/server/orca/call_metric_recorder.cc#L75)):*
<details><summary><i>Previously, nicolasnoble (Nicolas Noble) wrote…</i></summary><blockquote>

Oh, you're right, I went a bit too fast here. Fixed.

Also about the -1 value, I only copied your implementation from the oob service here, by the way:

https://github.com/grpc/grpc/blob/master/src/cpp/server/orca/orca_service.cc#L202-L209

I'm guessing it makes sense here to consider "-1" a properly uninitialized value.

</blockquote></details>

Yeah, -1 makes sense, since utilization can actually be 0.

nicolasnoble:
Failures are timeouts, seems like because the tests took too long to run.

PR #29622, opened 2022-05-09
Remove inline execution of timers scheduled in the past

drfloob:
@nicolasnoble


markdroth:
What's the motivation for this change?

drfloob:
I'm closing this in favor of another approach (probably). There's a conversation happening around EventEngine expectations, and the issue of the current iomgr timer implementation potentially running callbacks synchronously/in the same thread.

PR #29632, opened 2022-05-10
doc: add interop-test description for orca

YifeiZhuang:
add cross language readme doc for orca integration test.
Next step: add internal CI for the tests


ejona86:
There's a related discussion to this in https://github.com/grpc/grpc-java/pull/9079 . I think we need to make it a bit more generic. The approach right now is poorly suited to cross-language testing.

doc/interop-test-descriptions.md
```diff
@@ -1069,6 +1069,50 @@ included in that latency measurement (channel teardown semantics differ widely
...
+
+The client verifies that a custom LB policy, which is integrated with ORCA APIs, will receive both 
+per-query and out-of-band metric reports from the backend. 
+
+The client will register the custom LB policy named `test_backend_metrics_load_balancer`, which
+using ORCA APIs already installed a per-query report listener, and an out-of-band report 
+listener (configure load report interval to be 1s). The interop-testing client will run with a
+service config to select the load balancing config (using argument `--service_config_json`), so that 
+it effectively uses this newly registered custom LB policy. A load report reference can be passed 
+from the call to the LB policy through, e.g. CallOptions, to receive metric reports.
```
markdroth, line 1083: How will the test framework interact with this load report reference to verify the resulting load reports?

doc/interop-test-descriptions.md
```diff
@@ -1281,3 +1326,26 @@ Discussion:
...
+
+During test, the server will receive unary requests from the client that each may contain up to two
+test load report, indicating whether it needs to update metrics for the current call or at the OOB server.
+Then the server sets the metrics data, echoing the test load report it just received.
+Specifically:
+1. If the `orca_per_rpc_report` is set in the request, the server test driver will call
+`CallMetricRecorder` to record both utilization and 
+request cost metrics for the current RPC, the metrics data is coped from the received test load 
+report from the request.
+2. And if the `orca_oob_rpeort` is set in the request, the server test driver will call
```
ejona86, line 1349: s/rpeort/report/

doc/interop-test-descriptions.md
```diff
@@ -1069,6 +1069,50 @@ included in that latency measurement (channel teardown semantics differ widely
...
+using ORCA APIs already installed a per-query report listener, and an out-of-band report 
+listener (configure load report interval to be 1s). The interop-testing client will run with a
+service config to select the load balancing config (using argument `--service_config_json`), so that 
+it effectively uses this newly registered custom LB policy. A load report reference can be passed 
+from the call to the LB policy through, e.g. CallOptions, to receive metric reports. 
+The reference contains the load report for the current RPC in the per-query case, or the latest report
+in the out-of-band case. This way, together with server behaviors we can verify the 
+expected metric reports are received.
+
+To perform the test, two new fields are added to the test unary call request data structure to
```
ejona86, line 1088: It looks like this paragraph can be removed. The proto is defined elsewhere.

doc/interop-test-descriptions.md
```diff
@@ -1069,6 +1069,50 @@ included in that latency measurement (channel teardown semantics differ widely
...
+To perform the test, two new fields are added to the test unary call request data structure to 
+communicate with the server:
+1. A `TestLoadReport` type `orca_per_rpc_report` that tells the server to update per-query metrics.
+2. A `TestLoadReport` type `orca_oob_report` that tells the server to update OOB metrics.
+
+Server features:
+* [UnaryCall][]
+* [Backend metrics report][]
+
+#### Per-query metrics reporting
```
ejona86, line 1097: This is written as two different tests. Write it as two separate tests? `orca_per_rpc` and `orca_oob`? Don't follow the soak test style; it looks like it has incorrect formatting and it is a very different test than normal.

The `test_backend_metrics_load_balancer` discussion can just be on one of them and then referenced in the second.

doc/interop-test-descriptions.md
```diff
@@ -1069,6 +1069,50 @@ included in that latency measurement (channel teardown semantics differ widely
...
+Procedures:
+* The client sends a unary request to the server. The call request sets `orca_per_rpc_report` to a
```
ejona86, line 1099: Specify the precise request contents to send

doc/interop-test-descriptions.md
YifeiZhuang: The test framework will fill in the reference with the load report callback API. The test case can then verify the reference has the result. I updated the doc to make it more explicit.

doc/interop-test-descriptions.md
YifeiZhuang: oh, do you suggest to use two separate test cases at the interop-testing client test cases?

doc/interop-test-descriptions.md
markdroth: Don't we have a common test framework for all languages?  If so, how will the test framework fill in the reference for each language-specific client, given that each language will have a different API for this?

ejona86: The test framework doesn't do any assertions (other than assert return code is 0). This is non-xDS interop where we have assertions in each language. The test framework is involved to enable `test_backend_metrics_load_balancer`, but then it is per-language APIs for the test case to interact with that LB policy.

ejona86 (approved):

doc/interop-test-descriptions.md
markdroth: Ah, okay.  In that case, this makes sense.  Thanks!

YifeiZhuang: yea each language should 
* implement the `test_backend_metrics_load_balancer ` in interop test client
* implement interoptest server
* implement test case described in test procedure.

Use

```
./run-test-client.sh --use_tls=false --test_case=orca_per_rpc --service_config_json='{"loadBalancingConfig":[{"test_backend_metrics_load_balancer":{}}]}'
```

to enable the loadbalancer. The client should pass with cross-language server: e.g. a java client and Go server runs:
```
 ./run-test-server.sh --use_tls=false 
```

@markdroth @dfawley If things look good to you, i'll merge this PR and proto change today to unblock java impl.

PR #29633, opened 2022-05-10
xds: add "ignore_resource_deletion" server feature

markdroth:


yashykt:
Sorry for the delay, give me an additional day please. I want to understand this properly.

yashykt (approved):
I think we should also be adding a server side test to test that we continue serving, but I won't block on that. Feel free to put it on my plate too.

markdroth:
Good idea.  I've added the test.  PTAL.

yashykt (approved):

PR #29655, opened 2022-05-11
Properly hinting clangd.

nicolasnoble:
https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd

### DIFF
--- a/compile_flags.txt
+++ b/compile_flags.txt
@@ -1,0 +1,5 @@
+-std=c++14
+-I.
+-Iinclude
+-DDEBUG
+-D_DEBUG


nicolasnoble:
Also, a stopgap until https://github.com/hedronvision/bazel-compile-commands-extractor but I didn't want to spend time on integrating it right now.

ctiller (changes requested):
Actually it looks like clang tidy interacts poorly with this?

nicolasnoble:
Eh, sadface. We'd need to properly keep up a full on .json version instead then. Alright.

PR #29660, opened 2022-05-12
Make stack_tracer more broadly usable

drfloob:
These changes reduce the stack_tracer's dependencies, allowing it to be
added to other targets such as `gpr_base` without dependency cycles.

Why would I want to do this? printf debugging lock problems that only
occur on the CI's Mac VMs, which we cannot access.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -777,6 +777,22 @@ grpc_cc_library(
 )
 
 grpc_cc_library(
+    name = "examine_stack",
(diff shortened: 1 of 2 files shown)


ctiller (approved):

veblush (approved):

drfloob:
XDS and iOS errors are unrelated

PR #29664, opened 2022-05-12
[ObjC] run ios unit tests with bazel

HannahShiSFB:
https://github.com/grpc/grpc-ios/issues/74
to run locally,
OVERRIDE_BAZEL_VERSION=5.1.1 ./tools/run_tests/run_tests.py -l objc -r ios-test-unittests

@dennycd

### DIFF
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -24,12 +24,21 @@ load("@build_bazel_rules_apple//apple:resources.bzl", "apple_resource_bundle")
 load("@build_bazel_rules_apple//apple:ios.bzl", "ios_application", "ios_unit_test")
 load("@build_bazel_rules_apple//apple:macos.bzl", "macos_unit_test")
 load("@build_bazel_rules_apple//apple:tvos.bzl", "tvos_application", "tvos_unit_test")
+load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
 
 licenses(["notice"])
 
 package(default_visibility = ["//visibility:public"])
 
 exports_files(["LICENSE"])
+
+ios_test_runner(
+    name = "ios_test_runner_with_env",
+    test_environment = {
+        "HOST_PORT_LOCAL": "${HOST_PORT_LOCAL}",
+        "HOST_PORT_LOCALSSL": "${HOST_PORT_LOCALSSL}",
+    }
+)
 
 proto_library_objc_wrapper(
     name = "messages_proto",
@@ -200,6 +209,7 @@ ios_unit_test(
         ":NSErrorUnitTests-lib",
         ":RxLibraryUnitTests-lib",
     ],
+    runner = ":ios_test_runner_with_env",
 )
 
 ios_unit_test(
--- a/src/objective-c/tests/UnitTests/GRPCClientTests.m
+++ b/src/objective-c/tests/UnitTests/GRPCClientTests.m
@@ -41,10 +41,10 @@
 // in turn derived from environment variable of the same name.
 #define NSStringize_helper(x) #x
 #define NSStringize(x) @NSStringize_helper(x)
-static NSString *const kHostAddress = NSStringize(HOST_PORT_LOCAL);
+static NSString *kHostAddress = NSStringize(HOST_PORT_LOCAL);
 static NSString *const kPackage = @"grpc.testing";
 static NSString *const kService = @"TestService";
-static NSString *const kRemoteSSLHost = NSStringize(HOST_PORT_REMOTE);
+static NSString *kRemoteSSLHost = NSStringize(HOST_PORT_REMOTE);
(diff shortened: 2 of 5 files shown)


tools/run_tests/run_tests.py
```diff
@@ -1067,7 +1067,7 @@ def test_specs(self):
                 environ=_FORCE_ENVIRON_FOR_WRAPPERS))
         # TODO: replace with run_one_test_bazel.sh when Bazel-Xcode is stable
         out.append(
-            self.config.job_spec(['src/objective-c/tests/run_one_test.sh'],
+            self.config.job_spec(['src/objective-c/tests/run_one_test_bazel.sh'],
```
jtattermusch, line 1070: please see 
https://github.com/grpc/grpc/pull/29665/files#diff-26e0bc5bba55be18a1e28b707c1768be5c3de415944115ec3b4eb006b1d39693R1081

While I think they way of passing HOST_PORT_LOCALSSL and HOST_PORT_LOCAL env variables is useful (since for bazel we could only use static ports until now), I think we should not run bazel test via the run_tests.py script and via the run_one_test_with_bazel.sh

Instead we should use the grpc_objc_bazel_test I just introduced. That test already runs the unitTests target with bazel and does that together with other settings we want (reports in UI, results uploaded in bigquery) and in a way that more aligned with our other bazel test jobs. also it makes it easier to use remote caching in the future.

I think we should focus on improving the grpc_objc_bazel_test job (there's plenty of TODOs and possible improvements) and after new tests are added there, we should then remove them from the run_tests.py suite (which is slower, more complicated and has other problems).

src/objective-c/tests/BUILD
```diff
@@ -24,13 +24,22 @@ load("@build_bazel_rules_apple//apple:resources.bzl", "apple_resource_bundle")
...
 load("@build_bazel_rules_apple//apple:tvos.bzl", "tvos_application", "tvos_unit_test")
+load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
 
 licenses(["notice"])
 
 package(default_visibility = ["//visibility:public"])
 
 exports_files(["LICENSE"])
 
+ios_test_runner(
```
jtattermusch, line 35: note that in bazel we already have this mechanism:
https://github.com/grpc/grpc/blob/adfd009d3a255b825ea91959620c11805418b22b/src/objective-c/tests/BUILD#L60
which is what we are currently using (and the bazel test are passing). I think your approach is better and more general, but we probably shouldn't have two different mechanisms for achieving the same thing.

jtattermusch:
> OVERRIDE_BAZEL_VERSION=5.1.1

+1 to this nice little local workaround for the bazel 4.x issues.

src/objective-c/tests/BUILD
```diff
@@ -24,13 +24,22 @@ load("@build_bazel_rules_apple//apple:resources.bzl", "apple_resource_bundle")
...
+    name = "ios_test_runner_with_env",
+    test_environment = {
```
dennycd, line 37: can we check if using env paramters for ios_unit_test (https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_unit_test-env) would work ? that way we don't have to define extra runner for each test.

src/objective-c/tests/UnitTests/GRPCClientTests.m
```diff
@@ -104,6 +104,9 @@ @interface GRPCClientTests : XCTestCase
 @implementation GRPCClientTests
 
 + (void)setUp {
+  NSDictionary* environment = [[NSProcessInfo processInfo] environment];
```
dennycd, line 107: consider moving these logic into a separate utility function to extract the address.  This would potentially be shared across UnitTest & InteropTest that requires interop server address info.

dennycd (commented):
Thanks Hannah ! Please rebase on latest master and pickup Jan's latest changes on [grpc_objc_bazel_test.sh](https://github.com/grpc/grpc/blob/4976933e6b09706ccc542b7c9c7221cd6c50bd32/tools/internal_ci/macos/grpc_objc_bazel_test.sh#L52) which should already have UnitTest target hooked up to the script. 

For parameterizing interop port value, you can export the env from that script so the port value would be consistent between the interop server and the objc bazel tests.

dennycd:
Also, consider removing UnitTest from Cocoapod test suite as those will be redundant. 

https://github.com/grpc/grpc/blob/4976933e6b09706ccc542b7c9c7221cd6c50bd32/src/objective-c/tests/Podfile#L35 

We will still keep a core set of build test for Cocoapod for release validation etc. thanks

src/objective-c/tests/BUILD
dennycd: looks like we may also pass this in via test_env (https://bazel.build/reference/command-line-reference#flag--test_env)

PR #29668, opened 2022-05-12
xds: don't start resource timer after ADS stream restart if resource is already cached

markdroth:
As per discussion in https://github.com/grpc/proposal/pull/302#discussion_r871689791, we should not start the resource timer upon ADS stream restart if the resource is already cached, because that would defeat the optimization that servers may perform to avoid resending resources already seen by the client in a restarted stream.

Also, when the resource timer does fire, call the watcher's `OnResourceDoesNotExist()` method, not the `OnError()` method.

This change surfaced a bug in the xds resolver where it was failing to initialize the `Resolver::Result::addresses` field, thus resulting in RPCs failing with status UNKNOWN instead of UNAVAILABLE, so I've fixed that as well.  (As a follow-up item, I'm going to consider changes to the `Resolver::Result` API to make it impossible for resolver implementations to leave fields uninitlaized.)


src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -788,9 +789,11 @@ void XdsResolver::StartLocked() {
             grpc_error_std_string(error).c_str());
     std::string error_message;
     grpc_error_get_str(error, GRPC_ERROR_STR_DESCRIPTION, &error_message);
-    Result result;
-    result.service_config = absl::UnavailableError(
+    absl::Status status = absl::UnavailableError(
         absl::StrCat("Failed to create XdsClient: ", error_message));
+    Result result;
+    result.addresses = status;
+    result.service_config = status;
```
yashykt, line 796: can use std::move here?

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -802,10 +805,12 @@ void XdsResolver::StartLocked() {
...
         xds_client_->bootstrap().LookupAuthority(uri_.authority());
     if (authority_config == nullptr) {
-      Result result;
-      result.service_config = absl::UnavailableError(
+      absl::Status status = absl::UnavailableError(
           absl::StrCat("Invalid target URI -- authority not found for ",
                        uri_.authority().c_str()));
+      Result result;
+      result.addresses = status;
+      result.service_config = status;
```
yashykt, line 813: can use std::move here

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -956,11 +961,13 @@ void XdsResolver::OnError(absl::string_view context, absl::Status status) {
   gpr_log(GPR_ERROR, "[xds_resolver %p] received error from XdsClient: %s: %s",
           this, std::string(context).c_str(), status.ToString().c_str());
   if (xds_client_ == nullptr) return;
+  status =
+      absl::UnavailableError(absl::StrCat(context, ": ", status.ToString()));
   Result result;
+  result.addresses = status;
+  result.service_config = status;
```
yashykt, line 968: can use std::move here

src/core/ext/xds/xds_client.cc
```diff
@@ -245,23 +256,23 @@ class XdsClient::ChannelState::AdsCallState
...
+                      name_.authority, type_->type_url(), name_.key)
+                      .c_str());
         }
         auto& authority_state =
             ads_calld_->xds_client()->authority_state_map_[name_.authority];
         ResourceState& state = authority_state.resource_map[type_][name_.key];
         state.meta.client_status = XdsApi::ResourceMetadata::DOES_NOT_EXIST;
-        ads_calld_->xds_client()->NotifyWatchersOnErrorLocked(state.watchers,
-                                                              watcher_error);
+        ads_calld_->xds_client()->NotifyWatchersOnResourceDoesNotExist(
```
yashykt, line 274: why this change?

yashykt (changes requested):

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
markdroth: There's no need for `std::move()` on an `absl::Status`, because the implementation is basically just a ref-counted pointer.  I believe that clang-tidy complains when you try to use `std::move()`.

src/core/ext/xds/xds_client.cc
markdroth: It was a bug that the code was using `OnError()` instead of `OnResourceDoesNotExist()`.  The whole point of the timer is that it's the only way to know that a resource does not exist when it is first requested from the server.

In practice, this probably doesn't make any difference, because the timer is used only when initially subscribing to the resource, which means that there is no previous version cached, and in that situation, the `OnError()` behavior in all of our watchers is basically equivalent to `OnResourceDoesNotExist()`.  But semantically, `OnResourceDoesNotExist()` is correct here.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
yashykt: If it's refcounted, wouldn't std::move help to avoid the unnecessary atomic increment and decrement? I didn't find any information on avoiding std::move() for absl:::Status

markdroth: Looks like this worked, so either I was remembering wrong about clang-tidy not being happy with this, or something has changed since the last time I saw that.

In any case, seems to be fine now.  PTAL.

yashykt (approved):

yashykt (approved):

PR #29670, opened 2022-05-12
Reland DNSResolver cancellation (#29581 and #29631)

drfloob:
This includes a test that deletes the client's `grpc_pollset_set` immediately after cancelling a request.


drfloob:
MacOS CI is fixed: https://source.cloud.google.com/results/invocations/3e49752d-2d31-48d0-9f12-caf90792e32d/targets;query=resolve_address/%2F%2Ftest%2Fcore%2Fiomgr:resolve_address_using_ares_resolver_posix_test/log

Context: previously, it would reliably segfault on MacOS (CI only, not locally) with the following error:

```
E0510 20:43:36.074253000 123145483722752 sync_posix.cc:67]             assertion failed: pthread_mutex_lock(mu) == 0
```

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
```diff
@@ -358,74 +363,86 @@ class AresClientChannelDNSResolverFactory : public ResolverFactory {
...
+          on_resolve_address_done_(std::move(on_resolve_address_done)),
+          completed_(false),
+          resolver_(resolver),
+          aba_token_(aba_token) {
       GRPC_CARES_TRACE_LOG("AresRequest:%p ctor", this);
       GRPC_CLOSURE_INIT(&on_dns_lookup_done_, OnDnsLookupDone, this,
                         grpc_schedule_on_exec_ctx);
+      MutexLock lock(&mu_);
+      pollset_set_ = grpc_pollset_set_create();
+      grpc_pollset_set_add_pollset_set(interested_parties, pollset_set_);
```
apolcyn, line 386: Is this right - should this not be `grpc_pollset_set_add_pollset_set(pollset_set_, interested_parties);` ?

We want fds added to `pollset_set_` to propagate into the callers's `interested_parties_`, which I think we need to change this in order to do?

Otherwise, this might only be working because of backup pollers. E.g. if we do a DNS resolution manually, I suspect it will be slow.

drfloob: Indeed. I dialed the backup poller back to 10s intervals and tried external resolution ... it took 10s. Great catch. Fixed.

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
```diff
@@ -381,9 +382,11 @@ class AresDNSResolver : public DNSResolver {
       GRPC_CLOSURE_INIT(&on_dns_lookup_done_, OnDnsLookupDone, this,
                         grpc_schedule_on_exec_ctx);
       MutexLock lock(&mu_);
+      pollset_set_ = grpc_pollset_set_create();
```
markdroth, line 385: Please initialize this in the initializer list instead of in the body of the ctor.

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
```diff
@@ -434,6 +436,8 @@ class AresDNSResolver : public DNSResolver {
           }
         }
       }
+      grpc_pollset_set_del_pollset_set(request->interested_parties_,
```
markdroth, line 439: The two parameters are in the opposite order here than they are in `Cancel()` (line 411).

I can never remember which order is correct, but I strongly suspect they need to be the same in both places.

test/core/iomgr/resolve_address_test.cc
```diff
@@ -428,6 +428,70 @@ TEST_F(ResolveAddressTest, CancelWithNonResponsiveDNSServer) {
   PollPollsetUntilRequestDone();
 }
 
+// RAII class for pollset and pollset_set creation
+class PSS {
```
markdroth, line 432: Please call this something like `PollsetSetWrapper`.

https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

test/core/iomgr/resolve_address_test.cc
```diff
@@ -428,6 +428,70 @@ TEST_F(ResolveAddressTest, CancelWithNonResponsiveDNSServer) {
...
+
+  gpr_mu* mu_;
+  grpc_pollset* ps_;
+  grpc_pollset_set* pss_;
+};
+
+TEST_F(ResolveAddressTest, DeleteInterestedPartiesAfterCancellation) {
+  // Regression test for race around interested_party deletion after
+  // cancellation.
+  if (std::string(g_resolver_type) != "ares") {
```
markdroth, line 465: Nit: Could use `absl::string_view` instead of `std::string` here to avoid an unnecessary allocation.

markdroth (commented):
This looks really good!  Just a few small comments.

You are probably starting to realize why I am so anxious to see the pollset_set APIs go away.  They are a significant source of bugs. :(

test/core/iomgr/resolve_address_test.cc
drfloob: Done.

test/core/iomgr/resolve_address_test.cc
drfloob: Done

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
drfloob: Done

markdroth (approved):
Looks great!

PR #29676, opened 2022-05-13
Convert ObjC codegen plugin tests to bazel tests.

jtattermusch:
Based on https://github.com/grpc/grpc/pull/29665 (will rebase later).

- turn the original `ios-test-plugintest` and `ios-test-plugin-option-test` tasks into actual bazel test targets.
- the original test scripts were quite messy so I tried to fix what I could (note that they are actually still quite messy and partially broken and also their coverage is poor, but I'm not in the business of fixing that).
- note that the tests only run the protoc binary and the grpc_objective_c_plugin binaries (which have little to do with the actual objC stack) and then look at the generated files with grep etc, so they form a bit of a special category and they actually might be part of the codegen test suite rather than the objC test suite (but I'm leaving them where they are for now to avoid complicating things).

As a followup, the corresponding run_tests.py tests can be removed.
We should doublecheck that this doesn't break the import.

### DIFF
--- a/src/objective-c/tests/PluginTest/plugin_option_test.sh
+++ b/src/objective-c/tests/PluginTest/plugin_option_test.sh
@@ -18,13 +18,10 @@
 
 set -ev
 
-cd $(dirname $0)
-
-# Run the tests server.
-
-ROOT_DIR=../../../..
-PROTOC=$ROOT_DIR/bazel-bin/external/com_google_protobuf/protoc
-PLUGIN=$ROOT_DIR/bazel-bin/src/compiler/grpc_objective_c_plugin
+# protoc and grpc_objective_c_plugin binaries are supplied as "data" in bazel
+PROTOC=./external/com_google_protobuf/protoc
+PLUGIN=./src/compiler/grpc_objective_c_plugin
+WELL_KNOWN_PROTOS_PATH=external/com_google_protobuf/src
 RUNTIME_IMPORT_PREFIX=prefix/dir/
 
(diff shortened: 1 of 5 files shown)


jtattermusch:
Newly added test seem green on the CI: https://source.cloud.google.com/results/invocations/010c85f5-4dbe-4fcd-9c67-f4eb190ed6f5/targets

jtattermusch:
CC @HannahShiSFB

src/objective-c/tests/PluginTest/plugin_option_test.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
dennycd, line 1: would be nice to have a final test step that would take the generated objc stub and run a  bazel build to verify. Perhaps adding another TODO here to note that? thanks

dennycd (approved):
LGTM. looks like grpc_basictests_objc_ios  still had some timeout error.

src/objective-c/tests/PluginTest/plugin_option_test.sh
jtattermusch: sure, but I think the priority should be to migrate as many tests as test we can to bazel BEFORE we start adding new coverage to the existing tests. It's better to have the same test coverage as before with faster and more reliable way of running them (bazel seems to give us that) then having some tests "perfect" while other tests are stuck with a messy and fragile setup.

jtattermusch:
Test failures are unrelated.

PR #29687, opened 2022-05-13
Changed C99 to C11

veblush:
gRPC has been setting C99 explicitly so far (but its core API strictly sticks to C89) but the need to use [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision)) is growing particularly from boringssl which started using multi-threading features such as atomic operations. C11 is already available for all supported compilers (even they default to C17) except MSVC 2017 so gRPC can benefit from this feature as long as it can fallback to C99 for MSVC 2017.


veblush:
[grpc_build_boringssl_at_head](https://fusion2.corp.google.com/invocations/3d80a194-cabe-422d-9c22-0f1e9efff117): running

veblush:
`manylinux2010` python artifact builds fails with this PR so it should be addressed in advance. https://github.com/grpc/grpc/pull/29734

nicolasnoble (approved):

PR #29689, opened 2022-05-13
Fix gRPC C++ init bug

veblush:
Currently gRPC got a segfault with the following code.

```
  grpc::CompletionQueue cq;
  return 0;
}
```

This is because ctor of `CompletionQueue` requires gRPC C++ to be initialized but it's not yet. Although `src/cpp/common/completion_queue_cc.cc` has a static variable, `static internal::GrpcLibraryInitializer g_gli_initializer;` but it isn't necessarily initialized before `CompletionQueue`. (partly ctor of `CompletionQueue` is in the header so a compiler doesn't care of init-ing `g_gli_initializer` before `CompletionQueue`.

To address this, I moved ctors of `CompletionQueue` to its source file and made it call `g_gli_initializer.summon()` to make sure that `g_gli_initializer` is initialized before `CompletionQueue`.

In addition to this, ResourceQuota just got the same change.

### DIFF
--- a/include/grpcpp/impl/codegen/completion_queue.h
+++ b/include/grpcpp/impl/codegen/completion_queue.h
@@ -105,10 +105,7 @@ class CompletionQueue : private grpc::GrpcLibraryCodegen {
  public:
   /// Default constructor. Implicitly creates a \a grpc_completion_queue
   /// instance.
-  CompletionQueue()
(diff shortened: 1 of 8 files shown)


veblush:
@ctiller Please check whether it's desirable for ResourceQuota to have this change.

ctiller (approved):

yashykt (approved):

Issue #26549, opened 2021-06-24
Reorganize the build_artifacts -> build_packages -> distribtest  flow for easy-to-interpret results and simpler troubleshooting.


jtattermusch:
Context https://github.com/grpc/grpc/issues/26546#issuecomment-867480331

Problem statement:
I see folks mis-identifying the rootcause of problems in the artifact - packages - distribtest flow way too often (e.g. https://github.com/grpc/grpc/issues/26546). Basically, I think the issue is that there are multiple jobs (artifact linux, artifact macos, artifact windows, build packages linux, distribtests linux, distribtests macos, distribtests windows) that are grouped/chained together and depend on each other's outcome, but from kokoro perspective, they are mostly presented as individual jobs and that makes finding the links between these jobs far from straightforward.

I think this is something we should fix going forward, otherwise we're going to see confusion and misinterpretation in the future as well. I've already done some work along these lines in #26176, but more work is needed to achieve similar results for other languages.

Potential gains:
- lower test latency (since we won't need to have all the artifacts for all the languages build in a single slow "build_artifacts" job.
- easier to interpret test results, easier to quickly see root causes of problem
- not requiring expert knowledge of the artifact - build packages - distribtest flow in order to be able to diagnose problems
- ability to run distribtest for some languages platform as presubmit tests (e.g. as in #26176). Currently this is not possible since there is now way to display results of a chained job on a github PR (and distribtests are the last link in the build chain).

jtattermusch:
CC @lidizheng, @apolcyn, @stanley-cheung, @veblush

veblush:
@jtattermusch I like this :) Something we need to be aware of is that some language (i.e. ruby) are building Windows artifacts on Linux, which makes this approach a bit complicated.

jtattermusch:
Now with the advent of bazelified RBE tests I think this can be basically considered obsolete.

PR #29699, opened 2022-05-16
Simplify building release packages through per-language grpc_distribtests_* jobs.

jtattermusch:
Progress towards https://github.com/grpc/grpc/issues/26549.

 We already have a set of `grpc_distribtest_{python,php,ruby,csharp}` jobs, that are single-job variants of the build artifact -> build packages -> distribtests workflow. These jobs run both on master an on PRs.

Since these jobs already produce packages for all languages (there are some small exceptions, such as C#'s Grpc.Tools and ruby's grpc.tools gem), it's possible to simply run all these jobs as part of the release, and obtain the packages to upload that way. The advantages:
- running separate job for each language is faster and easier to maintain.
- the desired packages are both build and tested the very first job in the build flow, which makes it easier to navigate between the kokoro jobs we start manually as part of the release (in our current setup one always has to check for the distribtests result which is the 3rd job in the build flow and it takes a lot of clicking to get there)
- we can run the same set of tests on PRs, on master and on the release, which makes things simpler and less error prone.

What's done in this PR:
- create a clone of the grpc_distribtest_* jobs that will live under {linux,windows,macos}/release/ and will started as child jobs of a new parent job (one that will be started manually as part of the release process).
- create a new `release/grpc_collect_all_packages.sh` job that will be used for aggregating the artifacts produced by the first build step (all the grpc_distribtests_* jobs) into a single folder.

This PR doesn't affect correctness of the legacy grpc_build_artifact, grpc_build_packages and grpc_distribtests jobs, since we still need those to exist and to work against already-released branches (to maintain the ability to make patch releases for older 1.46.x, 1.45.x ...)


veblush (approved):

jtattermusch:
Test failures are unrelated since this PR only modifies the distribtest_* scripts and it adds a few more job config (both unrelated to the failures reported by CI).

Issue #29672, opened 2022-05-13
Grpc.Tools: Warnings from generated code in .NET 7


JamesNK:
.NET 7 is adding a build warning for type names with all lowercase characters. This is impacting code generated from Grpc.Tools which aliases namespaces with all lowercase characters, e.g. pb, pbc, grpc, etc

Log output:

> /Users/vstester/GrpcGreeterClient/obj/Debug/net7.0/Protos/Greet.cs(8,7): warning CS8981: The type name 'pb' only contains lower-cased ascii characters. Such names may become reserved for the language. [/Users/vstester/GrpcGreeterClient/GrpcGreeterClient.csproj]
> 
> /Users/vstester/GrpcGreeterClient/obj/Debug/net7.0/Protos/Greet.cs(9,7): warning CS8981: The type name 'pbc' only contains lower-cased ascii characters. Such names may become reserved for the language. [/Users/vstester/GrpcGreeterClient/GrpcGreeterClient.csproj]
> 
> /Users/vstester/GrpcGreeterClient/obj/Debug/net7.0/Protos/Greet.cs(10,7): warning CS8981: The type name 'pbr' only contains lower-cased ascii characters. Such names may become reserved for the language. [/Users/vstester/GrpcGreeterClient/GrpcGreeterClient.csproj]
> 
> /Users/vstester/GrpcGreeterClient/obj/Debug/net7.0/Protos/Greet.cs(11,7): warning CS8981: The type name 'scg' only contains lower-cased ascii characters. Such names may become reserved for the language. [/Users/vstester/GrpcGreeterClient/GrpcGreeterClient.csproj]
> 
> /Users/vstester/GrpcGreeterClient/obj/Debug/net7.0/Protos/GreetGrpc.cs(8,7): warning CS8981: The type name 'grpc' only contains lower-cased ascii characters. Such names may become reserved for the language. [/Users/vstester/GrpcGreeterClient/GrpcGreeterClient.csproj]

Options are to either:
- Update code generation to alias using pascal case. There is no chance these names are ever used by C# and break generated code. However, there is a chance the alias name could conflict with other types names in the app. Probably shouldn't do this.
- Add a pragma suppression. Simplest, easiest. It's unlikely that pb, pbc, pbr, scg, grpc are used by C# in the future.
- Remove aliases and use full namespaces instead. More complicated but safest. There is a mix between aliases and full namespaces so this would make generated code files larger, but more consistent.

aateeque:
[...]

PR #29708, opened 2022-05-17
C#: Suppress CS8981 in generated source

JamesNK:
Fixes https://github.com/grpc/grpc/issues/29672

This is option 2 (see linked issue). It suppresses the new warning in the generated source code. It's the simplest change and I think it's unlikely that any of the namespace aliases would turn into keywords.

@jtattermusch @jskeet

### DIFF
--- a/src/compiler/csharp_generator.cc
+++ b/src/compiler/csharp_generator.cc
@@ -801,7 +801,7 @@ std::string GetServices(const FileDescriptor* file, bool generate_client,
       out.PrintRaw(leading_comments.c_str());
     }
 
-    out.Print("#pragma warning disable 0414, 1591\n");
+    out.Print("#pragma warning disable 0414, 1591, 8981\n");
 
     out.Print("#region Designer generated code\n");
     out.Print("\n");


JamesNK:
Is there generated code still in this repo that needs to be updated? If there is, I don't have the C++ toolchain setup to build the source gen. It would be great if someone - @jtattermusch? - could handle this.

jtattermusch:
> Is there generated code still in this repo that needs to be updated? If there is, I don't have the C++ toolchain setup to build the source gen. It would be great if someone - @jtattermusch? - could handle this.

There is no more generated code in the master branch since we removed src/csharp from it (now lives in the v1.46.x branch).

jtattermusch (approved):
LGTM.

jtattermusch:
Test failures are unrelated (I checked that they're also happening on master).

jtattermusch:
> When can Protobuf compiler changes from [protocolbuffers/protobuf#9981](https://github.com/protocolbuffers/protobuf/pull/9981) be merged? Need both in Grpc.Tools to remove all warnings.

The PR has been merged already, but for the change to land in Grpc.Tools we need one of:
- the protobuf PR is released in the next protobuf release and the grpc/grpc repo upgrades to that version (and then the subsequent Grpc.Tools will have both grpc and protobuf change).
- we backport the protobuf PR to the latest stable protobuf release, protobuf does a patch release (perhaps with some other patches) and grpc/grpc upgrades to that version of protobuf.

PR #29709, opened 2022-05-17
security and subchannel: implement UniqueTypeName and use it in these interfaces

markdroth:
This addresses the problem that @ctiller pointed out in https://github.com/grpc/grpc/pull/29539#discussion_r872555549.


markdroth:
I don't think the template approach is going to work the way we thought it would.  Expanding the template using a `constexpr char[]` actually expands it only once for each unique string instance, and if the compiler de-dups those instances, then it will probably also de-dup the expanded factory class, which means that this isn't giving us any more guarantee than the original code did.

I'm going to change this to require each subclass to explicitly create a static factory instance.  This is a bit more work, but I think it will do what we want.

markdroth:
The remaining test failures look like flakes, so I think this is ready for review.

src/core/lib/gprpp/unique_type_name.h
```diff
@@ -0,0 +1,103 @@
...
+*/
+
+namespace grpc_core {
+
+class UniqueTypeName {
+ public:
+  // Factory class.  There should be a single static instance of this
+  // for each unique type name.  However, this static instance must be
+  // dynamically allocated, since it is not trivially destructible.
+  class Factory {
```
ctiller, line 62: What if we wrote:

```
class Factory {
 public:
  Factory(absl::string_view name) : name_(new std::string(name)) {}
  ~Factory() = default;
  Factory(const Factory&) = delete;
  Factory& operator=(const Factory&) = delete;
  
  UniqueTypeName Create() { return UniqueTypeName(*name_); }
  
 private:
  std::string* name_;
};
```

why?
now we can create factories statically (cleans up the syntax for usage a little)
*and* I think ASAN will let that fly, whilst accidentally creating a factory non-statically will cause an asan error

wdyt?

(if we don't do this, please add the deleted copy/assignment operators here)

ctiller (approved):
left one thing to consider (i think it'd leave a nicer api), but otherwise this looks fine

src/core/lib/gprpp/unique_type_name.h
markdroth: Good idea!  Done.

PR #29715, opened 2022-05-18
Fix checking pointer of pointer in call.cc

lidizheng:
* Fixes macOS xDS fault injection end2end tests

For unknown reason, `call->receiving_buffer_` might be a `nullptr` on macOS causing the test to segfault. The fix in this PR simply patches this assignment, and makes the `nullptr` checking safer.

More info see b/232544809

### DIFF
--- a/src/core/lib/surface/call.cc
+++ b/src/core/lib/surface/call.cc
@@ -1095,9 +1095,9 @@ void FilterStackCall::BatchControl::PostCompletion() {
     error = GRPC_ERROR_NONE;
   }
   if (error != GRPC_ERROR_NONE && op_.recv_message &&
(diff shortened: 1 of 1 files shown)


ctiller:
(drive by note... I'm gonna change a lot of this code in the next ~week in #29637  ... so if there's a test that's currently failing that I can make work again and it's otherwise not urgent, I'd suggest lmk the test and I'll fix it as I go).

lidizheng:
@ctiller That would be great, the flaky test is xds_fault_injection_end2end_test, the Kokoro job is https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:grpc%2Fcore%2Fmaster%2Fmacos%2Fgrpc_basictests_c_cpp.

PR #29722, opened 2022-05-18
xds_k8s_install: Allow to override PYTHON_VERSION for local testing

sergiitk:
While test run environment allows python 3.6, allow local dev to override PYTHON_VERSION for testing.

### DIFF
--- a/tools/internal_ci/linux/grpc_xds_k8s_install_test_driver.sh
+++ b/tools/internal_ci/linux/grpc_xds_k8s_install_test_driver.sh
@@ -16,7 +16,7 @@
 set -eo pipefail
 
 # Constants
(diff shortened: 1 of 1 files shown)


sergiitk:
@lidizheng - Unfortunately. I'm not a big fan. But that's what's available on Kokoro - and py3.6 is pre-provisioned. Of course, we could switch to something else, and install newer python, but that would mean another external dependency introducing extra infra flakes. At some point in the future when benefits from using py3.6 vs infra flakes are diminishing, we should consider alternatives.

lidizheng:
To me, using `pyenv` is fine, we just need to be extra careful about which Python interpreter we are actually using.

PR #29724, opened 2022-05-18
Remove opt MacOS builds from PR presubmit basic tests

drfloob:
In an effort to reduce the load on our MacOS CI pool, @ctiller proposed we disable the Opt builds for the MacOS basic tests. The bazel opt builds will still happen as part of another test in the PR presubmit suite, so opt build coverage is still present (albeit not for CMake, specifically).

This affects the `grpc_basictests_c_cpp.cfg` job, which now runs the following:

```
 ./tools/run_tests/run_tests_matrix.py -f basictests macos corelang --internal_ci -j 1 --inner_jobs 4 --bq_result_table aggregate_results --dry_run
Will run these tests:
  run_tests_c_macos_dbg_native: "bash tools/run_tests/helper_scripts/run_tests_in_workspace.sh -t -j 4 -x ../run_tests/c_macos_dbg_native/sponge_log.xml --report_suite_name c_macos_dbg_native -l c -c dbg --iomgr_platform native --bq_result_table aggregate_results --measure_cpu_costs"
  run_tests_c++_macos_dbg_native: "bash tools/run_tests/helper_scripts/run_tests_in_workspace.sh -t -j 4 -x ../run_tests/cpp_macos_dbg_native/sponge_log.xml --report_suite_name cpp_macos_dbg_native -l c++ -c dbg --iomgr_platform native --bq_result_table aggregate_results --measure_cpu_costs"

--dry_run was used, exiting
```

Previously there were an additional 2: `run_tests_c_macos_opt_native` and `run_tests_c++_macos_opt_native`

### DIFF
--- a/tools/run_tests/run_tests_matrix.py
+++ b/tools/run_tests/run_tests_matrix.py
@@ -190,11 +190,22 @@ def _create_test_jobs(extra_args=[], inner_jobs=_DEFAULT_INNER_JOBS):
                                 ['--report_multi_target'],
                                 inner_jobs=inner_jobs)
 
-    # supported on all platforms.
+    # Linux and Windows core tests
     test_jobs += _generate_jobs(
         languages=['c'],
         configs=['dbg', 'opt'],
-        platforms=['linux', 'macos', 'windows'],
+        platforms=['linux', 'windows'],
+        labels=['basictests', 'corelang'],
+        extra_args=
+        extra_args,  # don't use multi_target report because C has too many test cases
+        inner_jobs=inner_jobs,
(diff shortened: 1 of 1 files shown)


jtattermusch (commented):
I've had something similar in mind, but it was actually along the lines of running the cmake c++ build on PRs in `--build_only` mode, while tests would be run on master.
Instead I'd keep the bazel C++ job on PR for both opt and dbg.

Btw, the PR description is wrong, because you're basically implying that you're removing MacOS opt cmake build from presubmit builds, but with this change you're actually removing it from both presubmit AND continuous runs AFAICT (which seems wrong to me).

PR #29731, opened 2022-05-19
Enable remote cache for selected ObjC bazel tests.

jtattermusch:
This should increase the build speed a lot.


jtattermusch:
Speedup for the bazel_cpp_ios_test job:
BEFORE ~23min AFTER ~15min (https://source.cloud.google.com/results/invocations/94e91205-1c14-4665-820f-e14d8d342ec3)

dennycd (approved):

PR #29740, opened 2022-05-19
Switching to ios test runner for unit tests

dennycd:
### Change Summary

Switching to dedicated ios test runner for bazel-based unit test. Test runner has several upsides  

* This appears to be a default required parameters for [ios_unit_test](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_unit_test), and we observed a fairly large set of simulator launch failure from recent samples as seem in https://github.com/grpc/grpc-ios/issues/94. 
* Test runner allow us to specify exact device and iOS version for the xctest environment which will make our overall test environment more consistent
* test runner are in general more lightweight than ios host since it does not requires loading [full iOS app environment](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_unit_test). This should also help speed up initial launch time for the test  

### Test locally 

```bash
bazel test //src/objective-c/tests:UnitTests --test_filter=CallAPIv2Tests --test_output=all --test_env=GRPC_VERBOSITY=DEBUG 
```
----

https://github.com/grpc/grpc-ios/issues/94

-- 

cc @HannahShiSFB  @jtattermusch

### DIFF
--- a/src/objective-c/grpc_objc_internal_library.bzl
+++ b/src/objective-c/grpc_objc_internal_library.bzl
@@ -31,6 +31,43 @@ load(
     "generate_objc_non_arc_srcs",
     "generate_objc_srcs",
 )
+load("@build_bazel_rules_apple//apple:ios.bzl", "ios_unit_test")
+load(
+    "@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl",
+    "ios_test_runner",
+)
+
+# The default device type for ios objc unit tests
+IOS_UNIT_TEST_DEVICE_TYPE = "iPhone 11"
+
+# The default iOS version for ios objc unit tests
+# IOS_UNIT_TEST_OS_VERSION = "13.3"
+
+def grpc_objc_ios_unit_test(
+        name,
+        deps,
+        env = {}):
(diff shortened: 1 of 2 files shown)


sampajano (approved):
Great improvement! thanks! 😃

dennycd:
All 100 repeated test run pass with 100% success rate after switching to test runner  

INFO: Elapsed time: 1933.977s, Critical Path: 101.73s
INFO: 7217 processes: 5704 remote cache hit, 713 internal, 800 darwin-sandbox.
INFO: Build completed successfully, 7217 total actions

//src/objective-c/tests:UnitTests                                        PASSED in 62.0s
  Stats over 100 runs: max = 62.0s, min = 9.3s, avg = 14.3s, dev = 12.7s

...
Executed 4 out of 4 tests: 4 tests pass.

dennycd:
also manually run separate bazel tests multiple times and verified all test run passes

jtattermusch:
Thanks for the change, it seems that the grpc/core/master/macos/grpc_objc_bazel_test kokoro job is now much more stable.

PR #28148, opened 2021-11-19
[WIP] Use EventEngine::RunAt in Subchannel connection retries

drfloob:
Built on #27513

    Replaces the use of grpc_timer in Subchannel retry logic.

See 3534062a2da78e28c028b07f0a1a6658a1194fa0

@markdroth

PR #29744, opened 2022-05-20
EventEngine::RunAt - Subchannel connection retries

drfloob:
Supersedes https://github.com/grpc/grpc/pull/28148


drfloob:
This is running into deadlock issues due to the underlying iomgr timer executing the callback inline instead of asynchronously in another thread, which happens especially under heavy load. The `now + 2 ms` trick in `IomgrEventEngine` is insufficient, I need to modify the iomgr code to allow for explicitly not executing via the exec_ctx with deadlines in the past.

drfloob:
Deadlocks resolved by https://github.com/grpc/grpc/pull/29755

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -803,8 +806,10 @@ void Subchannel::RequestConnection() {
 void Subchannel::ResetBackoff() {
   MutexLock lock(&mu_);
   backoff_.Reset();
-  if (state_ == GRPC_CHANNEL_TRANSIENT_FAILURE) {
-    grpc_timer_cancel(&retry_timer_);
+  if (state_ == GRPC_CHANNEL_TRANSIENT_FAILURE &&
+      GetDefaultEventEngine()->Cancel(retry_timer_handle_)) {
+    OnRetryTimerLocked();
+    WeakUnref(DEBUG_LOCATION, "ResetBackoff");
```
markdroth, line 812: Not safe to unref while holding the mutex, because if this is the last ref, then the mutex will be destroyed before it can be unlocked, which will cause a use-after-free bug.

(This is particularly unfortunate in this case, because it means that we can't change the weak ref to be owned by the callback, so we need to continue to handle the ref manually.)

drfloob: I changed it to a `ReleasableMutexLock` and released it before unreffing.

markdroth: IMHO, `ReleasableMutexLock` is anti-pattern that we really shouldn't support, because it makes it harder to see where the lock is actually held.  Instead, please do something like this:

```
bool timer_cancelled = false;
{
  MutexLock lock(&mu_);
  backoff_.Reset();
  timer_cancelled =
    state_ == GRPC_CHANNEL_TRANSIENT_FAILURE &&
    GetDefaultEventEngine()->Cancel(retry_timer_handle_);
  if (timer_cancelled) OnRetryTimerLocked();
}
if (timer_cancelled) WeakUnref(DEBUG_LOCATION, "ResetBackoff");
```

ctiller: please see my other comment for a much cleaner way of doing this.

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -961,7 +967,8 @@ void Subchannel::OnConnectingFinishedLocked(grpc_error_handle error) {
     SetConnectivityStateLocked(GRPC_CHANNEL_TRANSIENT_FAILURE,
                                grpc_error_to_absl_status(error));
     WeakRef(DEBUG_LOCATION, "RetryTimer").release();  // Ref held by callback.
-    grpc_timer_init(&retry_timer_, next_attempt_time_, &on_retry_timer_);
+    retry_timer_handle_ =
+        GetDefaultEventEngine()->RunAt(ee_deadline, [this] { OnRetryTimer(); });
```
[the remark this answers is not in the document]
drfloob: Tying a ref's lifetime to the callback's lifetime this way could be a problem if the EventEngine implementation does not delete the callback in a timely manner after execution or cancellation. There are no guarantees in the API about callback destruction times, so I'm hesitant to make this change.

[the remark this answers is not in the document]
drfloob: Per our chat, I added a note to the EventEngine API spec to suggest implementations delete cancelled callbacks ~immediately. It's difficult to enforce, but a conformance test can be added to check for deletion after run/cancellation within some small time bound. I went ahead and implemented your suggestion here.

src/core/ext/filters/client_channel/subchannel.cc
drfloob: Now done more cleanly, thanks.

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -804,13 +804,12 @@ void Subchannel::RequestConnection() {
 }
 
 void Subchannel::ResetBackoff() {
-  ReleasableMutexLock lock(&mu_);
+  auto self = WeakRef(DEBUG_LOCATION, "ResetBackoff");
```
markdroth, line 807: Please add a comment here explaining that we need this weak ref to ensure that the subchannel won't be destroyed until after the mutex is released, because cancelling the timer could delete the last weak ref, which is held by the timer callback.

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -893,12 +892,10 @@ void Subchannel::SetConnectivityStateLocked(grpc_connectivity_state state,
   health_watcher_map_.NotifyLocked(state, status);
 }
 
+// Caller must call WeakUnref after OnRetryTimer returns
```
markdroth, line 895: I don't think we need this comment, since this isn't where anyone will be looking to consider ref-counting.

src/core/ext/filters/client_channel/subchannel.cc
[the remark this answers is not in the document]
markdroth: I like this approach!  I was actually thinking of suggesting something like this when I made my last comment, but I didn't think we had a guarantee of when the callback would be destroyed.  The solution here of taking a new weak ref until after the mutex is released addresses that perfectly.

src/core/ext/filters/client_channel/subchannel.cc
drfloob: Removed

src/core/ext/filters/client_channel/subchannel.cc
drfloob: Done.

drfloob:
*SAN failures: This PR depend on the ExecCtx changes in https://github.com/grpc/grpc/pull/29777

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -804,6 +804,8 @@ void Subchannel::RequestConnection() {
 }
 
 void Subchannel::ResetBackoff() {
+  // Hold a ref to ensure cancellation and subsequent deletion of the closure
+  // does not eliminate the last ref and destroy the Subchannel.
```
markdroth, line 808: ...while holding the mutex.

drfloob: done

markdroth (approved):
Looks great!

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -954,15 +958,19 @@ void Subchannel::OnConnectingFinishedLocked(grpc_error_handle error) {
...
             "subchannel %p %s: connect failed (%s), backing off for %" PRId64
             " ms",
             this, key_.ToString().c_str(), grpc_error_std_string(error).c_str(),
             time_until_next_attempt.millis());
     SetConnectivityStateLocked(GRPC_CHANNEL_TRANSIENT_FAILURE,
                                grpc_error_to_absl_status(error));
-    WeakRef(DEBUG_LOCATION, "RetryTimer").release();  // Ref held by callback.
-    grpc_timer_init(&retry_timer_, next_attempt_time_, &on_retry_timer_);
+    retry_timer_handle_ = GetDefaultEventEngine()->RunAt(
+        ee_deadline, [self = WeakRef(DEBUG_LOCATION, "RetryTimer")] {
```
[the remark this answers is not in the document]
markdroth: Thanks for calling that out.  I did think about that, but this seems okay, since it shouldn't be that hard to write this a different way if we wind up needing to back out the C++14 change for some reason.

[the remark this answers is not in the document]
drfloob: A decent alternative is a few commits back in this PR, feel free to bring it back if you need (in my absence).

PR #29745, opened 2022-05-20
OpenCensus: fixes broken traces exporting caused by a missing EndSpan

epertoso:
It seems like the Census plugin is not ending the span created by `grpc::OpenCensusCallTracer`, therefore tools like Jaeger are unable to relate the spans created by `grpc::OpenCensusCallAttemptTracer` to the previous ones.

@drfloob


test/cpp/ext/filters/census/stats_plugin_end2end_test.cc
```diff
@@ -565,7 +630,58 @@ TEST_F(StatsPluginEnd2EndTest, TestApplicationCensusContextFlows) {
...
+        reinterpret_cast<census_context*>(&app_census_context));
+    context.AddMetadata(kExpectedTraceIdKey,
+                        app_census_context.Span().context().trace_id().ToHex());
+    traces_recorder_->StartRecording();
+    grpc::Status status = stub_->Echo(&context, request, &response);
+    EXPECT_TRUE(status.ok());
+  }
+  ::opencensus::trace::exporter::SpanExporterTestPeer::ExportForTesting();
+  traces_recorder_->StopRecording();
+  auto recorded_spans_ = traces_recorder_->GetSpansAndClearState();
```
markdroth, line 660: This is a local variable, not a data member, so it should not have an `_` suffix.

https://google.github.io/styleguide/cppguide.html#Variable_Names

markdroth (commented):
Just one small nit, otherwise looks really good!

test/cpp/ext/filters/census/stats_plugin_end2end_test.cc
epertoso: Done.

markdroth (approved):
Looks great!

@yashykt I've triggered the tests, but I'll leave it to you to check the results and get this merged.

yashykt:
@epertoso Note that tests are failing. Error message - 
```
test/cpp/ext/filters/census/stats_plugin_end2end_test.cc:660:43: error: no member named 'GetSpansAndClearState' in 'grpc::testing::(anonymous namespace)::ExportedTracesRecorder'
  auto recorded_spans = traces_recorder_->GetSpansAndClearState();
                        ~~~~~~~~~~~~~~~~  ^
test/cpp/ext/filters/census/stats_plugin_end2end_test.cc:671:8: error: variable has incomplete type 'void'
  auto sent_span_data =
       ^
test/cpp/ext/filters/census/stats_plugin_end2end_test.cc:673:8: error: variable has incomplete type 'void'
  auto attempt_span_data =
       ^
test/cpp/ext/filters/census/stats_plugin_end2end_test.cc:677:8: error: variable has incomplete type 'void'
  auto recv_span_data =
       ^
```

epertoso:
My bad, I refactored something and forgot to re-run the tests. They're passing locally now.

epertoso:
Need more work, it seems there's some flakiness related to the exported spans. I'll need more time to see what's happening there.

epertoso:
@yashykt 
I've looked up into why the test is flaky, and that's because the `OpenCensusCallTracer` destructor is called when the `grpc_call` object is destroyed, which may or may not happen before all the spans have been recorded. 

The other tests in the file `stats_plugin_end2end_test.cc` use `absl::SleepFor(absl::Milliseconds(500))` to wait for the destructor to be called.

I'm not a big fan of using `sleep` in multi-threaded tests, but it seems that there is no easy way to register a closure that runs after the call gets destroyed. Do you think it's acceptable to put `SleepFor` before exporting the traces, like the other tests are doing for the stats?

epertoso:
> Hey, yes, I had seen this issue some time back but didn't have time to fix it at that point. I'd say go ahead and add that sleep for now.

Sorry for the late reply, done.

PR #29759, opened 2022-05-21
Fix 64 to 32 bit clang conversion warning

dennycd:
Explicit casting to avoid 64 to 32 bit conversion clang warning ([-Wshorten-64-to-32](https://clang.llvm.org/docs/DiagnosticsReference.html#wshorten-64-to-32))  as seen from gRPC iOS and Firebase iOS SDK (e.g. https://github.com/firebase/firebase-ios-sdk/issues/9790, https://github.com/firebase/firebase-ios-sdk/discussions/9700) . This is observed from MacOS/iOS platform but I believe it also applies to others. thanks 

----

### DIFF
--- a/src/core/ext/filters/fault_injection/fault_injection_filter.cc
+++ b/src/core/ext/filters/fault_injection/fault_injection_filter.cc
@@ -119,7 +119,7 @@ absl::StatusOr<FaultInjectionFilter> FaultInjectionFilter::Create(
 }
 
 FaultInjectionFilter::FaultInjectionFilter(ChannelFilter::Args filter_args)
-    : index_(grpc_channel_stack_filter_instance_number(
+    : index_((int)grpc_channel_stack_filter_instance_number(
           filter_args.channel_stack(),
           filter_args.uninitialized_channel_element())),
       service_config_parser_index_(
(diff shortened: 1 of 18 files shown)


src/core/lib/security/credentials/external/aws_request_signer.cc
```diff
@@ -17,8 +17,40 @@
...
+  #include <openssl_grpc/hmac.h>
+#else
+  #if COCOAPODS==1
+  #include <openssl_grpc/hmac.h>
+#else
+  #if COCOAPODS==1
+  #include <openssl_grpc/hmac.h>
+#else
+  #if COCOAPODS==1
+  #include <openssl_grpc/hmac.h>
```
sampajano, line 30: naive question - why are these look identical/repetitive? 😄

dennycd: updated and removed. These are local cocoapod changes that are not part of the PR

sampajano: ahh i see! makes more sense now! 😄

sampajano (approved):
thanks! :)

markdroth (changes requested):
This doesn't look like the right way to fix this.  First, we should not be using C-style casts in C++ code.  And second, I think that instead of just adding casts to make the compiler problems go away, we should be looking at each of these cases to see what's actually causing the problem.

Just as one example, in the fault injection filter, instead of casting the result of `grpc_channel_stack_filter_instance_number()` to an int, we should change the type of the `index_` field in the filter from `int` to `size_t`.

dennycd:
thanks Mark. will split this into individual case fixes

PR #29763, opened 2022-05-23
doc/interop-test-descriptions.md: Fix organization

ejona86:
The soak tests had the wrong header level and the TODO tests should not
be mixed in with the defined tests.


ejona86:
CC @YifeiZhuang

apolcyn (approved):

PR #29764, opened 2022-05-23
GoogleDefaultCreds: support new-style xDS cluster names

markdroth:


src/core/lib/security/credentials/google_default/google_default_credentials.cc
```diff
@@ -83,6 +84,21 @@ struct metadata_server_detector {
...
 };
+
+namespace {
+
+bool IsXdsNonCfeCluster(const char* xds_cluster) {
+  if (xds_cluster == nullptr) return false;
+  if (absl::StartsWith(xds_cluster, "google_cfe_")) return false;
+  if (!absl::StartsWith(xds_cluster, "xdstp:")) return true;
+  auto uri = grpc_core::URI::Parse(xds_cluster);
+  if (!uri.ok()) return true;  // Shouldn't happen, but assume ALTS.
```
apolcyn, line 95: Since we don't expect this case to happen, can we log an error showing the cluster name and the uri parse error?

apolcyn (approved):

src/core/lib/security/credentials/google_default/google_default_credentials.cc
markdroth: I don't think that log message would ever actually be triggered.  If the cluster name starts with "xdstp:" but is not parseable as a URI, the XdsClient will report the failure to the watcher without ever subscribing to the resource on the wire, so we'll never actually get any endpoints with this kind of cluster name.

ejona86:
Why doesn't this change match against the authority? Isn't that the point of federation?

markdroth:
We could do that if there's a reason to.  My thought was that it wouldn't really buy us much, because it doesn't seem likely that this will accidentally be used with some other authority.  Also, hard-coding it here would make it harder to switch to some other authority in the future if we decided that we wanted to do so, although that's admittedly also not very likely.  So I just erred on the side of doing less work.

If you can think of a compelling reason to add this, I'm certainly open to discussion.

ejona86:
@markdroth, @dfawley, and I discussed this today and discovered the "/envoy.config.cluster.v3.Cluster/google_cfe_" path in the control plane was specified for one specific authority. This had been my assumption and Mark had the opposite assumption/memory (this specific path is independent of the authority, within the context of this one control plane). Mark was closer to this work and had read some documents I didn't know existed, so there were surprises for everyone.

With the knowledge the path was specified for one authority by the control plane, then we all were basically on the same page of being fine both ways but leaning toward checking the authority. So we're going to check the authority.

markdroth:
I've sent #29901 to add the authority check.

PR #29768, opened 2022-05-23
server: add method to expose authority seen by server

markdroth:
Also improve gtest names for test cases in end2end_test.cc.


markdroth:
CC @mohanli-ml

apolcyn (commented):
If I'm correct, this is exposed through [host()](https://github.com/grpc/grpc/blob/0afcc375890eef6de23c8e0a29b6953b3f07b8e0/include/grpcpp/impl/codegen/async_generic_service.h#L44) on `GenericServerContext` (for proxies).

Should the API name be the same here?

ctiller (approved):

PR #29779, opened 2022-05-25
grpc_authz test: longer wait for authz policy reload

jtattermusch:
See internal b/204329811.

Increase the sleep from 2 to 5 seconds.

The grpc_authz tests seems badly designed, this should prevent it from flaking heavily on mac.

### DIFF
--- a/test/core/end2end/tests/grpc_authz.cc
+++ b/test/core/end2end/tests/grpc_authz.cc
@@ -53,8 +53,8 @@ static gpr_timespec five_seconds_from_now(void) {
 
 static void wait_for_policy_reload(void) {
   // Wait for the provider's refresh thread to read the updated files.
-  // TODO: Refactor the tests to use a more reliable mechanism of detecting
-  // policy has been reloaded. See b/204329811
+  // TODO(jtattermusch): Refactor the tests to use a more reliable mechanism of
+  // detecting that the policy has been reloaded. See b/204329811
   gpr_sleep_until(grpc_timeout_seconds_to_deadline(5));
 }
 
--- a/test/core/end2end/tests/grpc_authz.cc
+++ b/test/core/end2end/tests/grpc_authz.cc
@@ -51,6 +51,13 @@ static gpr_timespec five_seconds_from_now(void) {
   return n_seconds_from_now(5);
 }
 
+static void wait_for_policy_reload(void) {
+  // Wait for the provider's refresh thread to read the updated files.
+  // TODO: Refactor the tests to use a more reliable mechanism of detecting
+  // policy has been reloaded. See b/204329811
+  gpr_sleep_until(grpc_timeout_seconds_to_deadline(5));
+}
+
 static void drain_cq(grpc_completion_queue* cq) {
   grpc_event ev;
   do {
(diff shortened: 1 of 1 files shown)


test/core/end2end/tests/grpc_authz.cc
```diff
@@ -51,6 +51,13 @@ static gpr_timespec five_seconds_from_now(void) {
   return n_seconds_from_now(5);
 }
 
+static void wait_for_policy_reload(void) {
+  // Wait for the provider's refresh thread to read the updated files.
+  // TODO: Refactor the tests to use a more reliable mechanism of detecting
```
markdroth, line 56: I don't actually see how this would help.  Even if we had a signal when the policy was reloaded by the background thread, we'd still need to have some timeout by which the test fails if the reload does not happen, or else the test will run forever upon a failure.  I don't see any reasonable way of avoiding this timeout.

markdroth (approved):

test/core/end2end/tests/grpc_authz.cc
jtattermusch: There can be a timeout, but the test should be able to determine whether the reload has actually happened and report it with an appropriate error message.  With the current setup, the error message isn't very useful and it's not clear what has actually happened (looks like not even the author of the test was able to quickly see what's going on - which IMHO is a pretty good signal that the test should be improved).

markdroth: This is an end-to-end test, not a unittest.  It's covering the APIs that would be used by a real application, and those APIs don't provide a way to know when a reload happens, because that's not necessary in those APIs.  I don't think we should expose such a thing here.

I do think we should have separate unit tests that cover more detailed internal behvior, and we could use some sort of dependency-injection mechanism in that kind of test to determine when the reload happens.  But I don't think we should introduce that kind of thing in an end-to-end test.

With regard to making this test more debuggable, I don't think the fact that this took so long to resolve is related to it being hard to debug; I think the real problem was that it got caught between teams (this code is maintained by the security team, not by grpc-team, and they didn't make it enough of a priority to debug this flake).  And to the extent that understanding the test may have been a problem, I think the change you've already made here of introducing a function called "wait_for_policy_reload()" is sufficient to provide future engineers with a hint about what the test is waiting for.

markdroth: Adding a timeout multiplier for Macs might very well be a good idea.  That might be a very good systemic way of addressing all of the flakiness that we're seeing on Macs.

jtattermusch: I'll leave that up to you. I'd prefer to get this PR in ASAP because the macos flakiness is through the roof current and getting the numerous grpc_authz flakes out of the way would make remaining flakes more visible.
I don't own these tests so @ashithasantosh can followup if more than this PR is needed.

markdroth: I will try putting together a PR to increase the multiplier on Macs.

jtattermusch: thanks for the feedback. Feel free to merge as is or take over and modify.

jtattermusch:
Ok, so it looks like grpc_authz are still failing after merging this, likely with the same flakiness rate.
@ashithasantosh  can you please take a look?

ashithasantosh:
Yes, working on it. I am trying to reproduce the error in Mac. And since I haven't made any changes to authz code since November, and the tests starting flaking on April, will also try bisect commands to see if it is an external code change. I will update the internal bug when I make some progress.

PR #29788, opened 2022-05-25
Split the `gpr` target into many small `gpr_*` targets

drfloob:
Replaced with 25 new `gpr_*` targets.

Tested working so far:

```
bazel build :all src/core/... src/compiler/... test/core/... test/cpp/...
```

cc @ctiller


src/core/lib/gpr/assert_internal.h
```diff
@@ -0,0 +1,42 @@
...
+/** abort() the process if x is zero, with rudimentary logging to prevent
+   circular dependencies with gpr_log.
+
+   Intended for internal invariants.  If the error can be recovered from,
+   without the possibility of corruption, or might best be reflected via
+   an exception in a higher-level language, consider returning error code.  */
+#define GPR_ASSERT_INTERNAL(x)                     \
+  do {                                             \
+    if (GPR_UNLIKELY(!(x))) {                      \
+      fprintf(stderr, "assertion failed: %s", #x); \
```
markdroth, line 31: Might be a good idea to have this log the file and line number.

markdroth (commented):
This looks awesome!

src/core/lib/gpr/assert_internal.h
drfloob: Alright, we now have files & line numbers, and the spacing is consistent with normal logging. To keep the dependencies for this internal target low, I haven't attempted to add in time, date, or tid in a cross-platform way.

```
D0830 02:17:02.298932360      16 test_config.cc:138]         test slowdown factor: sanitizer=1, fixture=1, poller=1, total=1
INTERNAL                         time_posix.cc:69]           DO NOT SUBMIT: 42
I0830 02:17:02.299245791      16 ev_epoll1_linux.cc:121]     grpc epoll fd: 3
D0830 02:17:02.299266219      16 ev_posix.cc:141]            Using polling engine: epoll1
D0830 02:17:02.299306255      16 dns_resolver_ares.cc:831]   Using ares dns resolver
```

drfloob:
`gpr` is a publicly depended-upon library, so breaking it up would be problematic. Much as this would improve things for gRPC, it would be a significant breaking change.

PR #29793, opened 2022-05-26
Provide means to control TCP frame sizes in response to high memory pressure

Vignesh2208:
This PR parses the minimum number of bytes that need to be read by the TCP layer to receive a frame that can be fully decrypted. This minimum number of bytes (aka min_progress_size) is conveyed by secure_endpoint to the TCP layer.

Requires Cherrypick.

@ananda1066


ctiller:
Can you separate this into read path and write path changes?
I'd like to land them separately so we can roll them back separately if needed.

Vignesh2208:
> Can you separate this into read path and write path changes? I'd like to land them separately so we can roll them back separately if needed.

Removed the endpoint write frame clipping changes. But I can also add them back. 

The TCP layer right now ignores min_progress_size and the max_frame_size is set to INT_MAX by the transport layer. So it should be safe to land both read and write path changes as well. This PR should have no effect in the actual framing until we  make TCP use min_progress_size and we change max_frame_size at the transport layer

src/core/lib/security/transport/secure_endpoint.cc
```diff
@@ -249,8 +252,17 @@ static void on_read(void* user_data, grpc_error_handle error) {
 
     if (ep->zero_copy_protector != nullptr) {
       // Use zero-copy grpc protector to unprotect.
-      result = tsi_zero_copy_grpc_protector_unprotect(
-          ep->zero_copy_protector, &ep->source_buffer, ep->read_buffer);
+      int last_incomplete_frame_size = 1;
+      // Get the size of the last frame which is not yet fully decrypted.
+      // This estimated is strored in ep->min_progress_size which is
```
ananda1066, line 257: nit: typo

src/core/tsi/alts/zero_copy_frame_protector/alts_zero_copy_grpc_protector.cc
```diff
@@ -218,6 +218,23 @@ static tsi_result alts_zero_copy_grpc_protector_unprotect(
   return TSI_OK;
 }
 
+static tsi_result alts_zero_copy_grpc_protector_unprotect_get_frame_size(
+    tsi_zero_copy_grpc_protector* self, grpc_slice_buffer* protected_slices,
+    grpc_slice_buffer* unprotected_slices, int* last_incomplete_frame_size) {
+  if (self == nullptr || unprotected_slices == nullptr ||
```
ananda1066, line 224: alts_zero_copy_grpc_protector_unprotect already returns INVALID_ARGUMENT if self, protected_slices, or unprotected_slices is nullptr so you shouldn't need to check them here too

src/core/tsi/fake_transport_security.cc
```diff
@@ -465,6 +465,23 @@ static tsi_result fake_zero_copy_grpc_protector_unprotect(
   return TSI_OK;
 }
 
+static tsi_result fake_zero_copy_grpc_protector_unprotect_get_frame_size(
+    tsi_zero_copy_grpc_protector* self, grpc_slice_buffer* protected_slices,
+    grpc_slice_buffer* unprotected_slices, int* last_incomplete_frame_size) {
+  if (self == nullptr || unprotected_slices == nullptr ||
```
ananda1066, line 471: same comment as above

src/core/lib/security/transport/secure_endpoint.cc
```diff
@@ -249,8 +252,17 @@ static void on_read(void* user_data, grpc_error_handle error) {
...
+      // passed to the specify layer to indicate the minimum number of
+      // bytes that need to be read to make meaningful progress. This would
+      // avoid reading of small slices from the network.
+      result = tsi_zero_copy_grpc_protector_unprotect_get_frame_size(
+          ep->zero_copy_protector, &ep->source_buffer, ep->read_buffer,
+          &last_incomplete_frame_size);
+      last_incomplete_frame_size = std::max(1, last_incomplete_frame_size);
+      ep->min_progress_size = result != TSI_OK ? 1 : last_incomplete_frame_size;
     } else {
       // Use frame protector to unprotect.
```
ananda1066, line 267: maybe add a TODO to do this for non-zero copy?

Vignesh2208: Added a TODO

src/core/tsi/fake_transport_security.cc
Vignesh2208: Removed.

src/core/tsi/alts/zero_copy_frame_protector/alts_zero_copy_grpc_protector.cc
Vignesh2208: Fixed.

src/core/tsi/transport_security_grpc.h
```diff
@@ -69,6 +80,10 @@ struct tsi_zero_copy_grpc_protector_vtable {
   tsi_result (*unprotect)(tsi_zero_copy_grpc_protector* self,
                           grpc_slice_buffer* protected_slices,
                           grpc_slice_buffer* unprotected_slices);
+  tsi_result (*unprotect_get_frame_size)(tsi_zero_copy_grpc_protector* self,
```
ctiller, line 83: I'm not loving the name, but don't have a concrete suggestion... maybe `unprotect_and_get_min_progress_size` -- I do explicitly think it should be min_progress_size and not frame_size -- frame_size implies this is only for the body of a frame, but my expectation is that a protocol with a four byte header would set min_progress_size to four prior to seeing the next header.

Vignesh2208: Renamed it now.

ctiller:
this LGTM, but going to let @ZhenLian approve.

src/core/tsi/transport_security_grpc.h
```diff
@@ -69,6 +80,9 @@ struct tsi_zero_copy_grpc_protector_vtable {
   tsi_result (*unprotect)(tsi_zero_copy_grpc_protector* self,
                           grpc_slice_buffer* protected_slices,
                           grpc_slice_buffer* unprotected_slices);
+  tsi_result (*unprotect_and_get_min_progress_size)(
+      tsi_zero_copy_grpc_protector* self, grpc_slice_buffer* protected_slices,
+      grpc_slice_buffer* unprotected_slices, int* min_progress_size);
```
ZhenLian, line 85: This looks good to me overall, but have we considered the possibility to add `int* min_progress_size` directly to the `unprotect` function signature? Because right now the two functions are almost the same, except for the last parameter.
How difficult it is if we change all the internal call-sites of TSI to fill in `min_progress_size`? I think doing that might require some extra work, but is generally beneficial to the code cleanliness in the long run.

Vignesh2208: Yeah Craig and I talked about this. The cherrypick will unfortunately be large if we change  the unprotect method directly and we have to get LGTMs from all the teams/users for landing or rolling back this change. This way, we can incrementally switch each user one by one to use the new method and then after everyone is migrated we can delete the old method and rename this in one go.

ZhenLian: OK, sounds good. I am fine with that approach as well.
It would be good if we could have a TODO for that :)

Vignesh2208: Thanks. Added a TODO

markdroth: I agree with Zhen on this.  I would really prefer to see this parameter added to the existing method rather than cluttering the API with new method variants this way.

It's not clear to me how it helps the import to add a new method anyway.  Either way, you will need to change all existing implementations, whether that's to add a new method or to add a parameter to the existing method.

markdroth (changes requested):

src/core/tsi/transport_security_grpc.h
ctiller: Hey Mark, we're not talking a few callsites here... it's a fairly large change.
The plan here is to introduce the new function, migrate incrementally, and then delete the old function - effectively eliminating the old method (this was mentioned up-thread).

Not doing so for this one would be too risky - it's going to be an LSC internally to get it done as it is.

markdroth: This PR is creating a new method only for the zero-copy frame protector API, which has no callers outside of gRPC.  In codesearch, I see only one reference to `tsi_zero_copy_grpc_protector_unprotect()` internally, and it's the one that the cherry-pick import CL for this PR is already touching.

Vignesh2208: I'll remove this method and do the cherrypick for this one. A similar change has to be made for the non_zero_copy based protector as well that cherrypick would be much larger.

markdroth: Sounds good, thanks!

Another option to consider is adding the parameter but making it an optional parameter, so that there's no need to modify any caller.

Vignesh2208: Craig doesn't like optional parameters I think :) Its probably better to force every caller to use it. I'll make the change. I got confused before with regards to which cherrypick would be large (zero copy vs non-zero-copy). There don't appear to be a lot of callers for zero copy protect/unprotect

Vignesh2208: Renamed the methods now.

markdroth (approved):

PR #29794, opened 2022-05-26
Revert "test: increase test slowdown factor on MacOS"

jtattermusch:
Reverts grpc/grpc#29781

- It seems that the total test jobs runtime on master went from 65mins -> 95mins (which results in the test job timing out now). https://source.cloud.google.com/results/invocations/103096fe-f695-4ef0-881b-39cb4ecc7b57/targets
- the grpc_authz tests are still failing (looks like at the same rate as before).

### DIFF
--- a/test/core/util/test_config.cc
+++ b/test/core/util/test_config.cc
@@ -47,12 +47,6 @@
 int64_t g_fixture_slowdown_factor = 1;
 int64_t g_poller_slowdown_factor = 1;
 
-#if GPR_APPLE
-static const int64_t kPlatformSlowdownFactor = 3;
-#else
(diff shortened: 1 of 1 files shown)


jtattermusch:
See e.g. https://fusion.corp.google.com/projectanalysis/summary/KOKORO/prod:grpc%2Fcore%2Fmaster%2Fmacos%2Fgrpc_bazel_c_cpp_dbg (for the sudden duration increase). For opt config, the same thing has happened.

markdroth (approved):
Yeah, looking at the top Mac flakes I filed bugs for yesterday, only one of them has not seen another flake since this change, and that could just be a coincidence given the flake rate.  So I don't think this change actually helped.

Issue #29560, opened 2022-05-03
Application hangs during shutdown sequence after error connecting to server


marcosatti:
### What version of gRPC and what language are you using?
gRPC 1.45.0, Python

### What operating system (Linux, Windows,...) and version?
Tested on Windows 11, Ubuntu 22.04 Jammy

### What runtime / compiler are you using (e.g. python version or version of gcc)
Python 3.10.4

### What did you do? / What did you see instead?
The application hangs upon cleanup and shutdown. 

In this snippet, the `channel_ready()` call fails (expected) and raises an `asyncio.TimeoutError` which is not caught and so the Python runtime should exit with a failed exit code. 
```
import asyncio
import grpc

async def startup_broker_api_channel():
    _api_channel = grpc.aio.insecure_channel(f"[::1]:65534")
    await asyncio.wait_for(_api_channel.channel_ready(), timeout=3.0)

async def main():
    await startup_broker_api_channel()
  
asyncio.run(main(), debug=True)
```

Instead the runtime hangs and the shutdown never completes, so the process has to be killed. 

Stack trace (in python.exe) indicates cygrpc is doing something (Windows stack trace). CPU cycles are being consumed, so it appears its looping over something continuously.
```
ntoskrnl.exe!KeSynchronizeExecution+0x6776
ntoskrnl.exe!KeLowerIrql+0x16d7
ntoskrnl.exe!KeLowerIrql+0x3589
ntoskrnl.exe!KeWaitForMutexObject+0x234
ntoskrnl.exe!KeQuerySystemTimePrecise+0x131b
ntoskrnl.exe!RtlFindClearBits+0x1634
ntoskrnl.exe!KeSynchronizeExecution+0x3a28
python310.dll!Py_NewReference+0x806
python310.dll!PyEval_EvalFrameDefault+0x2d9d
python310.dll!PyObject_GC_Malloc+0x1045
python310.dll!PyVectorcall_Call+0xb8
cygrpc.cp310-win_amd64.pyd+0x2d55a5
cygrpc.cp310-win_amd64.pyd!PyInit_cygrpc+0x64070
cygrpc.cp310-win_amd64.pyd!PyInit_cygrpc+0x5b4f3
cygrpc.cp310-win_amd64.pyd!PyInit_cygrpc+0x5a823
cygrpc.cp310-win_amd64.pyd!PyInit_cygrpc+0x1323d
python310.dll!PyObject_GetDictPtr+0x286
[...]

PR #29797, opened 2022-05-26
[Aio] Ensure Core channel closes when deallocated

lidizheng:
Fixes https://github.com/grpc/grpc/issues/29560.

### DIFF
--- a/src/python/grpcio_tests/tests_aio/unit/server_interceptor_test.py
+++ b/src/python/grpcio_tests/tests_aio/unit/server_interceptor_test.py
@@ -123,14 +123,14 @@ class _CacheInterceptor(aio.ServerInterceptor):
 
 async def _create_server_stub_pair(
     *interceptors: aio.ServerInterceptor
-) -> Tuple[aio.Server, aio.Channel, test_pb2_grpc.TestServiceStub]:
+) -> Tuple[aio.Server, test_pb2_grpc.TestServiceStub]:
     """Creates a server-stub pair with given interceptors.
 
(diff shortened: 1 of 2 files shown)


src/python/grpcio_tests/tests_aio/unit/server_interceptor_test.py
```diff
@@ -123,14 +123,14 @@ async def wrapper(
 
 async def _create_server_stub_pair(
     *interceptors: aio.ServerInterceptor
-) -> Tuple[aio.Server, test_pb2_grpc.TestServiceStub]:
+) -> Tuple[aio.Server, aio.Channel, test_pb2_grpc.TestServiceStub]:
```
gnossen, line 126: Can you please explain why this is necessary? I don't see the returned channel being used anywhere. Is this about keeping the reference around?

gnossen (approved):

src/python/grpcio_tests/tests_aio/unit/server_interceptor_test.py
lidizheng: Yes. The Python gRPC Stub only keeps the product of `unary_unary` method (Multicallable objects), which depends on the `cygrpc.AioChannel` instead of `grpc.aio.Channel`. Upon this comment, I found that we can inject a reference of Channel into Multicallable object. This behavior is more desired, because we won't break users that wasn't careful enough about their channels and stubs. PR updated.

gnossen (approved):

gnossen (approved):

PR #29798, opened 2022-05-26
Revert "Revert "Fix grpc_sockaddr_to_uri to return URI encoded string. Also see #29323 which has a similar fix for unix abstract addresses.""

anicr7:
Reverts grpc/grpc#29639


markdroth (approved):
Will this require a cherry-pick again?

anicr7:
> Will this require a cherry-pick again?

Yes, it will. Sent out a fix internally for the problematic targets. Will wait for the CL to be submitted before merging this PR.

markdroth:
> Will wait for the CL to be submitted before merging this PR.

That's not the right procedure.  Instead, please get both the CL and the PR ready to submit/merge, and the coodinate with whoever the importer is to do both at the same time.  Otherwise, your changes may be overwritten by an import that is already in flight when you pull the trigger.

anicr7:
> > Will wait for the CL to be submitted before merging this PR.
> 
> That's not the right procedure. Instead, please get both the CL and the PR ready to submit/merge, and the coodinate with whoever the importer is to do both at the same time. Otherwise, your changes may be overwritten by an import that is already in flight when you pull the trigger.

Sorry that is what I meant, I'll wait for the CL to be ready to submitted and then merge+submit them together. I'll sync up with the importer to make sure changes don't get overridden.

PR #29799, opened 2022-05-26
[ObjC] unit tests with bazel

HannahShiSFB:
https://github.com/grpc/grpc-ios/issues/74
to run locally,
OVERRIDE_BAZEL_VERSION=5.1.1 ./tools/run_tests/run_tests.py -l objc -r ios-test-unittests

@dennycd

### DIFF
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -195,10 +195,6 @@ grpc_objc_ios_unit_test(
         ":NSErrorUnitTests-lib",
         ":RxLibraryUnitTests-lib",
     ],
-    env = {
-        "HOST_PORT_LOCAL": "${HOST_PORT_LOCAL}",
-        "HOST_PORT_LOCALSSL": "${HOST_PORT_LOCALSSL}",
-    },
 )
 
 ios_unit_test(
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -195,6 +195,10 @@ grpc_objc_ios_unit_test(
         ":NSErrorUnitTests-lib",
         ":RxLibraryUnitTests-lib",
     ],
+    env = {
+        "HOST_PORT_LOCAL": "${HOST_PORT_LOCAL}",
+        "HOST_PORT_LOCALSSL": "${HOST_PORT_LOCALSSL}",
+    },
 )
 
 ios_unit_test(
--- a/src/objective-c/tests/run_one_test_bazel.sh
+++ b/src/objective-c/tests/run_one_test_bazel.sh
@@ -27,10 +27,6 @@ BAZEL=../../../tools/bazel
 
 INTEROP=../../../bazel-out/darwin-fastbuild/bin/test/cpp/interop/interop_server
 
-[ -d Tests.xcworkspace ] || {
-    ./build_tests.sh
-}
-
 [ -f $INTEROP ] || {
     BAZEL build //test/cpp/interop:interop_server
 }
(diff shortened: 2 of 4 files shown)


src/objective-c/tests/run_one_test_bazel.sh
```diff
@@ -48,4 +48,7 @@ $INTEROP --port=$TLS_PORT --max_send_message_size=8388608 --use_tls &
 
 trap 'kill -9 `jobs -p` ; echo "EXIT TIME:  $(date)"' EXIT
 
-../../../tools/bazel run $SCHEME
+time \
```
jtattermusch, line 51: this change should actually be in grpc_objc_bazel_test.sh as well, because the run_one_test_bazel.sh will probably be removed at some point (once all the tests migrate to grpc_objc_baze_test.sh)

dennycd: https://github.com/grpc/grpc/blob/9c23d7999ce827614195fe2da6653af30cb4298c/tools/internal_ci/macos/grpc_objc_bazel_test.sh#L87  

which currently run MacTest/UnitTest and codegen plugin test

src/objective-c/tests/run_one_test_bazel.sh
```diff
@@ -48,4 +48,7 @@ $INTEROP --port=$TLS_PORT --max_send_message_size=8388608 --use_tls &
...
+    HOST_PORT_LOCALSSL=localhost:$TLS_PORT \
+    HOST_PORT_LOCAL=localhost:$PLAIN_PORT \
```
dennycd, line 53: consider passing this in via [--test_env](https://bazel.build/reference/command-line-reference#flag--test_env) from bazel command line

src/objective-c/tests/BUILD
```diff
@@ -196,6 +196,10 @@ ios_unit_test(
         ":NSErrorUnitTests-lib",
         ":RxLibraryUnitTests-lib",
     ],
+    env = {
```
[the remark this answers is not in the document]
HannahShiSFB: I can remove them and use `—test_env`, but I think it’s better to be explicit that these tests require those environment variables as input.

src/objective-c/tests/run_one_test_bazel.sh
HannahShiSFB: comment addressed

src/objective-c/tests/BUILD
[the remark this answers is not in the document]
dennycd: discussed offline, we can pass in default configs for host addrss here, similar to what we did for preprocessor macros.

src/objective-c/tests/run_one_test_bazel.sh
dennycd: chatted offline, will update grpc_objc_bazel_test with env var as well

dennycd:
script to update with test_env 

https://github.com/grpc/grpc/blob/master/tools/internal_ci/macos/grpc_objc_bazel_test.sh

dennycd (approved):

dennycd:
looks like bazel MacTests failed since it contains InteropTests which has not yet migrated to using the host utils. 

//src/objective-[c/tests](http://c/tests):MacTests                                         FAILED in 180.4s

You may want to rebase on master with https://github.com/grpc/grpc/pull/29800 and that should fix the issue. Other than that, lgmt !

dennycd:
all test passes except objc ios test which is fixed via https://github.com/grpc/grpc/pull/29988 separately due to a c++ build break

PR #29804, opened 2022-05-26
xds-k8s README.md: improve basic setup docs

sergiitk:
Reorder sections, recommend local-dev.cfg instead of grpc-testing to avoid confusion.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/README.md
+++ b/tools/run_tests/xds_k8s_test_driver/README.md
@@ -195,45 +195,6 @@ python -m grpc_tools.protoc --proto_path=../../../ \
 
 # Basic usage
 
-### xDS Baseline Tests
-
-Test suite meant to confirm that basic xDS features work as expected. Executing
-it before other test suites will help to identify whether test failure related
-to specific features under test, or caused by unrelated infrastructure
-disturbances.
-
-The client and server images are created based on Git commit hashes, but not
-every single one of them. It is triggered nightly and per-release. For example,
-the commit we are using below (`d22f93e1ade22a1e026b57210f6fc21f7a3ca0cf`) comes
-from branch `v1.37.x` in `grpc-java` repo.
-
-```shell
-# Help
-python -m tests.baseline_test --help
-python -m tests.baseline_test --helpfull
-
-# Run on grpc-testing cluster
-python -m tests.baseline_test \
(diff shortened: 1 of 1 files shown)


easwars (approved):

sergiitk:
also cc @murgatroid99

tools/run_tests/xds_k8s_test_driver/README.md
```diff
@@ -269,7 +227,44 @@ envsubst < config/local-dev.cfg.example > config/local-dev.cfg
...
+
+Test suite meant to confirm that basic xDS features work as expected. Executing
+it before other test suites will help to identify whether test failure related
+to specific features under test, or caused by unrelated infrastructure
+disturbances.
+
+```shell
+# Help
+python -m tests.baseline_test --help
+python -m tests.baseline_test --helpfull
```
lidizheng, line 244: s/python/python3/g

tools/run_tests/xds_k8s_test_driver/README.md
```diff
@@ -269,7 +227,44 @@ envsubst < config/local-dev.cfg.example > config/local-dev.cfg
...
+
+# Run the baseline test with local-dev.cfg settings
+python -m tests.baseline_test --flagfile="config/local-dev.cfg"
+  
+# Same as above, but using the helper script
+./run.sh tests/baseline_test.py
```
lidizheng, line 250: optional: I would recommend to layout all the command line arguments, so the people onboarding the framework can see what's getting configured here. The `./run.sh` can exist as a shortcut documented as an alternative way of running the tests.

tools/run_tests/xds_k8s_test_driver/README.md
sergiitk: This assumes virtualenv is active, so `python` will point to the correct version.

tools/run_tests/xds_k8s_test_driver/README.md
sergiitk: People onboarding are likely to use `./run.sh` - that's from my experience of a few folks recently. Regarding command arguments, there are a lot of them. That's why I'm suggesting `--help` above.

But I think we could list all *required* arguments. Would it be ok if this is tackled outside of this PR?

lidizheng (approved):

Issue #29667, opened 2022-05-12
[C++] Custom certificate verifier should not depend on presence of root certs


oakad:
TL;DR; I propose to make `grpc_tls_certificate_verifier` object, if set by user, an one and only authority on certificate validity. No messing with root cert/crl sets. Just the output of `grpc_tls_certificate_verifier::Verify` to decide.

This was to some extent discussed before, but the issue still applies, despite the refactoring of TLS code.

Consider a system such as OSX/IOS/Windows and to some extent Android which do not expose user certificates as a bunch of pem blobs in readable location (those easily amount to 90%+ of all client installations on this planet). All those systems provide a readily accessible API to verify peer certificate chains in a system dependent manner. In many cases it is critically important to comply with the system specific method of verification.

Right now we have the implementation referenced below.

The main issue with it: it does all the hard work trying to initialize the "root cert" verification set even if `set_verify_server_certs` is set to false. The check for this options happens rather late, so a lot of cycles are wasted for nothing (as verification will not be relying on those certs anyway).

Nevertheless, we must provide some "root certs" for the current implementation to work at all. Moreover, we must maintain them in valid state, even though we know them to be non-functional.

Additionally, the naming of an option `set_verify_server_certs` is misleading in such scenarios. We don't want to disable verification, we want to delegate it. 

Thus, I think, it makes sense to drop the `set_verify_server_certs` option and instead simply check for presence of explicitly set `grpc_tls_certificate_verifier`. If it is set, no root set need to be populated and verification is fully delegated.

This will also make `grpc_tls_certificate_provider` use case simpler. Right now, in cases where intermediate root certs are delivered via network (xds style thing) the boilerplate of dynamic provider is pretty bad. It's actually easier to put together a custom `X509_STORE_CTX` (or whatever runtime platform offers) explicitly in the verifier than jump around the provider.

https://github.com/grpc/grpc/blob/adfd009d3a255b825ea91959620c11805418b22b/src/core/tsi/ssl_transport_security.cc#L2073-L2117

oakad:
I have rearranged the initialisation flow as following - I think it makes more sense in cases where are no root certs readily available anywhere on the system and custom verifier will be present.

```
--- src/core/lib/security/security_connector/ssl_utils.cc.orig	2021-12-17 06:33:48.000000000 +1100
+++ src/core/lib/security/security_connector/ssl_utils.cc	2022-05-20 19:07:15.000000000 +1000
@@ -426,7 +426,7 @@
[...]

PR #29810, opened 2022-05-27
security: don't require system CA certs if certificate verification is disabled (#29667)

oakad:
Presently, when establishing a TLS connection, gRPC will always attempt to load some CA certificates from the file system or user supplied callback mechanism.

However, on popular commercial platforms (MS Windows, Apple OSX) CA certificates are not readily available to be loaded as files.

Therefore, requirement to always load the root certs creates a problem:

1. When custom, platform specific verifier is present.
2. When security is not an issue, but TLS connection is nevertheless required (various testing situations).

### DIFF
--- a/src/core/tsi/ssl_transport_security.cc
+++ b/src/core/tsi/ssl_transport_security.cc
@@ -2015,9 +2015,6 @@ tsi_result tsi_create_ssl_client_handshaker_factory_with_options(
 
   if (factory == nullptr) return TSI_INVALID_ARGUMENT;
   *factory = nullptr;
-  if (options->pem_root_certs == nullptr && options->root_store == nullptr) {
-    return TSI_INVALID_ARGUMENT;
-  }
 
 #if OPENSSL_VERSION_NUMBER >= 0x10100000
   ssl_context = SSL_CTX_new(TLS_method());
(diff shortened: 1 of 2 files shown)


ZhenLian:
Thanks for making the pull request! I think this is just for illustrative purpose and not meant for a real pull request, so I will close this for now. 
As mentioned in another related issue, we currently don't have enough cycle to work on this, but we will surely add this to our backlog and reconsider this in the future when we need to stabilize our TLS API.

oakad:
"I think this is just for illustrative purpose"

Not nice.
The improvement breaks no existing functionality and improves the client experience.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -2066,27 +2063,6 @@ tsi_result tsi_create_ssl_client_handshaker_factory_with_options(
   }
 
   do {
```
ZhenLian, line 2065: This is irrelevant to this change, but in regards to the pattern the code is using: 
```
do {
  // ...
} while (false);
```
it seems the only reason for using this pattern is to save space for doing `tsi_ssl_handshaker_factory_unref` in multiple places. But I think it sort of affects the readability, so I am not sure what's the best way in normal C++ style guide to handle this case. 

Might need @markdroth 's opinion on this...

ZhenLian (commented):
Sorry, I thought this PR was only for illustration purpose and you wanted me to put together something later. If this is a pull request that can be reviewed, that's even better and can speed things up a bit!

I think it's not strictly true that this PR doesn't break any existing functionality. It does because from an API's perspective, originally we will report an error even when `set_verify` is false and root cert is empty. Now we are changing to make it a valid case. But I generally agree that for TLS credentials, it makes sense to have such changes, especially under the condition when TLS API is still experimental.

I will probably note that this change won't break functionality of SSL credentials, because `skip_server_certificate_verification` is by default set to false. So if a user using old SSL credential doesn't provide any root certs, they will still receive `TSI_INVALID_ARGUMENT`.

We might need some tests to cover this case as well(or, I think some tests might have already been broken...). I will open the pull requests and enable the presubmit checks.

I think @markdroth will also need to take a look at this as this is a breaking API change.

ZhenLian:
@oakad Friendly ping on this...do we still want this to be merged?

oakad:
I have a pretty huge GRPC patch* on my side which I use for a production project with this feature included. So it's clearly, yes, I find it useful and other people may to.

I'm also aware that there's a good chance that patch will need a rework, as I'm still on 1.43 and may not be able to test newer version efficiently.

* The other big feature I've got in there is system certificate store integration for Windows, OSX and Android.

ZhenLian:
@oakad I see, thanks for letting me know! Handing off to @erm-g to continue the work that's needed to be discussed before we can merge this pull request.

matthewstevenson88:
Closing this since it got incorporated into #34859.

PR #29825, opened 2022-05-27
Fix tvos test break and pass interop server env

dennycd:
Update tvOS test scheme to set default env for interop server info. Also updated run test script to properly export test env. 

---

cc @HannahShiSFB


sampajano (approved):
thanks! :)

PR #29832, opened 2022-05-27
Extend RPC timeouts in CSDS tests

lidizheng:
This PR extends the RPC timeouts in CSDS tests to be 10s (the default was 1s). The RPCs are expected to be successful. In normal cases, they will pass within milliseconds. However, recently, we are experiencing timeouts caused by slow macOS workers, hence we need this PR.

### DIFF
--- a/test/cpp/end2end/xds/xds_csds_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_csds_end2end_test.cc
@@ -528,7 +528,8 @@ TEST_P(ClientStatusDiscoveryServiceTest, XdsConfigDumpEndpointError) {
   EdsResourceArgs args({{"locality0", CreateEndpointsForBackends(0, 1)}});
   balancer_->ads_service()->SetEdsResource(BuildEdsResource(args));
   // Ensure the xDS resolver has working configs.
-  CheckRpcSendOk(DEBUG_LOCATION, 1, RpcOptions().set_timeout_ms(kTimeoutMillisecond));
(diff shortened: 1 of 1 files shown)


lidizheng:
Let me take a closer look and see if we should expand this PR to more test cases, or simply extend the default RPC timeouts for xds e2e tests.

lidizheng:
It seems one week ago, people attempted the same approach and caused MSAN and UBSAN to be flaky. Closing this PR for now until we find a direction. Related https://github.com/grpc/grpc/pull/29781, https://github.com/grpc/grpc/pull/29794.

markdroth:
Changing the slowdown factor was reverted.  Increasing the timeouts in the individual tests that were affected is fine.

The problem here is not the approach; it's that you're only fixing one of the tests.  Please fix all of the tests in xds_csds_end2end_test.  Thanks!

lidizheng:
@markdroth PTALA. I updated all the timeout in this file. CSDS unit tests focus on the xDS content, and the availability of the CSDS service. The correctness of them won't be impacted by the increased timeout.

lidizheng:
This is suspicious, because `XdsConfigDumpEndpointDoesNotExist` and `XdsConfigDumpClusterDoesNotExist` have 1000s timeout that should never be reached, and `XdsConfigDumpClusterRequested` expects to see DEADLINE_EXCEEDED (the subscribed resource is not available, hence the REQUESTED state).

Other tests can increase the RPC timeout without problem, the above ones can't. I guess more investigation is needed.

PR #27793, opened 2021-10-21
Initial support for Haiku

korli:
This provides the baseline needed to build gRPC on Haiku systems.

At least one project has started using on gRPC as a dependency.

@drfloob

PR #29833, opened 2022-05-29
Fix c-ares Bazel build under Haiku

Tevo45:
This contains fixes for the c-ares Bazel BUILD file under Haiku, as well as a backport of https://github.com/c-ares/c-ares/commit/7586c5f (I haven't looked into it, but perhaps it might be a better idea to bump c-ares to a new enough version instead?); they are necessary for my Bazel Haiku port, as mentioned on https://github.com/bazelbuild/bazel/issues/12522#issuecomment-1121321856.

### DIFF
--- a/bazel/grpc_deps.bzl
+++ b/bazel/grpc_deps.bzl
@@ -289,6 +289,8 @@ def grpc_deps():
                 "https://storage.googleapis.com/grpc-bazel-mirror/github.com/c-ares/c-ares/archive/6654436a307a5a686b008c1d4c93b0085da6e6d8.tar.gz",
                 "https://github.com/c-ares/c-ares/archive/6654436a307a5a686b008c1d4c93b0085da6e6d8.tar.gz",
             ],
+            patches = ["@com_github_grpc_grpc//third_party:cares/cares_haiku.patch"],
+            patch_args = ["-p1"],
         )
 
     if "com_google_absl" not in native.existing_rules():
--- a/third_party/cares/BUILD
+++ b/third_party/cares/BUILD
@@ -6,5 +6,6 @@ exports_files([
     "config_freebsd/ares_config.h",
     "config_linux/ares_config.h",
     "config_openbsd/ares_config.h",
+    "config_haiku/ares_config.h",
     "config_windows/ares_config.h",
 ])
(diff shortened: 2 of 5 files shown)


Tevo45:
Seems patch misplaced a few lines on `grpc_deps.bzl` while importing the diff anyhow. Should be fine now, sorry for that

jtattermusch:
I looked up Haiku (https://en.wikipedia.org/wiki/Haiku_(operating_system)) and it doesn't seem like we'd want to take on the maintenance burden for such system. If there was a one line-fix somewhere to get things working on Haiku, we could accept it but this change is way too large and complex for something that we don't want to maintain and that seems to only have a very niche user base.

Tevo45:
>I looked up Haiku (https://en.wikipedia.org/wiki/Haiku_(operating_system)) and it doesn't seem like we'd want to take on the maintenance burden for such system.

#27793 introduces Haiku support on gRPC and the impression I've got from that was that that the maintenance burden for community platforms falls entirely within the responsibility of the corresponding community. Korli appointed himself as the Haiku maintainer there, but I'd be willing to be around and take care of my Bazel-related changes as well, if that seems reasonable.

>way too large and complex

506 lines goes to [third_party/cares/config_haiku/ares_config.h](https://github.com/grpc/grpc/pull/29833/commits/c35d129caf207c9b461db5c520a2fd0efb39cbbf#diff-c32b3687fbdd8e59a0e1834bde923ea31fc809f26363aa3b4832ef499ac037e5), which is autogenerated by c-ares' configure script and has to be in-tree since the BUILD file for c-ares doesn't know how to generate that (that's just following what gRPC already does for every other platform). 42 more lines go to a patch adding a couple of defines to c-ares so it builds correctly; that is already on upstream, but gRPC's Bazel build uses an older version of c-ares. The remaining 11 lines are just adding the patch and appropriate definitions to the build system.

The actual functional changes are little despite the apparent size, and the maintenance burden seems negligible perhaps beside someone having to remove the patch when c-ares finally gets bumped (and perhaps updating the configuration header as well; I'd be willing to be around for that, but not doing so wouldn't affect any other platform in any case). Perhaps one could get Bazel to apply the c-ares patch only when building for Haiku, then there would definitely be no functional changes on other platforms, if that seems more acceptable anyhow.

PR #29841, opened 2022-05-31
xDS: Workaround to get gRPC clients working with istio

yashykt:
Listener protos had an incorrect validation where it expected `Listener.address` to be always present. This does not work so well when clients are trying to decide whether the LDS resource is for a client or a server. This was fixed in https://github.com/envoyproxy/envoy/pull/21132 but till istio performs validation according to the previously incorrect validation rule, we need this workaround so as to be able to work with istio.

### DIFF
--- a/test/cpp/end2end/xds/xds_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_end2end_test.cc
@@ -1082,19 +1082,21 @@ TEST_P(XdsEnabledServerTest, BadLdsUpdateNoApiListenerNorAddress) {
       ::testing::HasSubstr("Listener has neither address nor ApiListener"));
 }
 
-TEST_P(XdsEnabledServerTest, BadLdsUpdateBothApiListenerAndAddress) {
-  Listener listener = default_server_listener_;
-  listener.mutable_api_listener();
-  SetServerListenerNameAndRouteConfiguration(balancer_.get(), listener,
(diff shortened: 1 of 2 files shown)


markdroth (approved):

test/cpp/end2end/xds/xds_end2end_test.cc
```diff
@@ -1082,19 +1082,21 @@ TEST_P(XdsEnabledServerTest, BadLdsUpdateNoApiListenerNorAddress) {
...
-  backends_[0]->Start();
-  const auto response_state = WaitForLdsNack(DEBUG_LOCATION);
-  ASSERT_TRUE(response_state.has_value()) << "timed out waiting for NACK";
-  EXPECT_THAT(
-      response_state->error_message,
-      ::testing::HasSubstr("Listener has both address and ApiListener"));
-}
+// TODO(roth): Re-enable the following test once
+// github.com/istio/istio/issues/38914 is resolved.
+// TEST_P(XdsEnabledServerTest, BadLdsUpdateBothApiListenerAndAddress) {
```
markdroth, line 1087: Can you prepend `DISABLED_` on the test name instead of commenting it out?  That way we at least keep it building.

sanjaypujare:
Thanks for this. I support this will be ported to 1.47?

PR #29849, opened 2022-05-31
Modify ContextList to include the relative start position and size of a traced stream within the output buffer.

Vignesh2208:
This change is required as we internally move to a new API to export timestamps associated with traced RPCs.

@ctiller


yashykt:
A quick side note - when I had written some of this, we didn't have access to STL yet

ctiller:
am I blocking this work?

ctiller:
Assuming this can be closed

Issue #28387, opened 2021-12-20
[Python][macOS] 3.10 wheel is mislabeled as universal2


DriveAthelete:
```
(venv) grigoryvorobiev@MacBook-Pro-Grigory pythonProject1 % python main.py                    
Traceback (most recent call last):
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/main.py", line 1, in <module>
    from google.cloud import translate
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/google/cloud/translate/__init__.py", line 17, in <module>
    from google.cloud.translate_v3.services.translation_service.client import (
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/google/cloud/translate_v3/__init__.py", line 17, in <module>
    from .services.translation_service import TranslationServiceClient
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/google/cloud/translate_v3/services/translation_service/__init__.py", line 16, in <module>
    from .client import TranslationServiceClient
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/google/cloud/translate_v3/services/translation_service/client.py", line 25, in <module>
    from google.api_core import gapic_v1  # type: ignore
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/google/api_core/gapic_v1/__init__.py", line 16, in <module>
    from google.api_core.gapic_v1 import config
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/google/api_core/gapic_v1/config.py", line 23, in <module>
    import grpc
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/grpc/__init__.py", line 22, in <module>
    from grpc import _compression
  File "/Users/grigoryvorobiev/PycharmProjects/pythonProject1/venv/lib/python3.10/site-packages/grpc/_compression.py", line 15, in <module>
    from grpc._cython import cygrpc
[...]

PR #29857, opened 2022-06-01
Set Correct Platform Tag in Wheels on Mac OS with Python 3.10

gnossen:
Intended to fix https://github.com/grpc/grpc/issues/28387

After much experimentation (seriously, look at the commit history), I found that the "ARCHFLAGS" approach just wasn't working. Instead, I simply rename the artifact. This has been tested to work on an M1 mac running with Rosetta.

This is a _hack_. Long term, we'll need to figure out why `distutils` is not generating the proper platform and fix that. We'll also want to generate both `arm64` artifacts and a proper `universal2` artifact containing both x64 and arm64 artifacts. However, this will unbreak M1 users in the short term.

Side note: This PR also updates the Python 3.10 patch version installed on MacOS.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -427,6 +427,7 @@ if 'darwin' in sys.platform:
     os.environ['ARCHFLAGS'] = '-arch {}'.format(platform.machine())
     sys.stderr.write("AAAAAAAAAAAAAAAAAAAA ARCHFLAGS: '{}'\n".format(os.environ['ARCHFLAGS']))
     sys.stderr.flush()
+    sys.exit(1)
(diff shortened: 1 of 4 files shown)


sergiitk (approved):

eapolinario:
@gnossen , should we expect a new release of `grpcio` (and the other grpc libraries) in order to pick up the change in https://github.com/grpc/grpc/pull/29857?

gnossen:
Yes. I'm expecting expecting a release candidate later today (pacific time) and 1.47.0 will follow in a couple of days.

PR #29859, opened 2022-06-01
Enable authorization logs for debugging

ashithasantosh:
Enabling authorization logs for authorization test targets to help with debugging MacOS failures.


ashithasantosh:
Looks like I am able to reproduce the error and view authz logs in the kokoro fails. Let me attempt to debug without having to submit this PR.

PR #29861, opened 2022-06-01
[C++/Doc] gRPC Objects with static storage duration

veblush:


doc/cpp/static-grpc-object.md
```diff
@@ -0,0 +1,38 @@
+# gRPC Objects with static storage duration
+
+In general, it's discouraged to define gRPC objects with static storage
```
markdroth, line 3: How about instead of "discouraged", we say "not supported"?

markdroth (commented):
Is this the right place for this doc?  Do we maybe want to put this somewhere on the grpc.io site?

doc/cpp/static-grpc-object.md
```diff
@@ -0,0 +1,38 @@
...
+The following code may crash because `cq`, which is a `grpc::CompleitionQueue`
+type is defined as a global static variable which is going to be initialized
+before `main()`.
+To initialize `grpc::CompletionQueue` type object, it needs fully initialized
+gRPC system but it may or may not happen at that point.
+In addition to that, when `cq` is destroyed, it also needs alive gRPC system
+which may be terminiated already.
+
+```
+static grpc::CompletionQueue cq;
```
ctiller, line 22: Can I write:
```
static std::unique_ptr<grpc::CompletionQueue> cq;

int main() {
  cq.reset(new grpc::CompletionQueue());
  cq->Shutdown();
}
```

veblush: This code is for a bad example (cq created before main()) but cq in your code will be created in main(). Do you show it as a solution?

doc/cpp/static-grpc-object.md
veblush: SGTM.

veblush:
> Is this the right place for this doc? Do we maybe want to put this somewhere on the grpc.io site?

I looked around to find the right place for this topic but I couldn't find any so I picked this directory because there are other slightly advanced topics. e.g. pending_api_cleanups.md in the same directory was published at https://grpc.github.io/grpc/cpp/md_doc_cpp_pending_api_cleanups.html so this doc will be published there as well.

[grpc.io](https://grpc.io/docs/languages/cpp/) looks like a place for the quick start resource (quick start & basic tutorial). We may want to have a dedicated place there to cover various intermediate & advanced topics.

doc/cpp/static-grpc-object.md
ctiller: But it will be destroyed after main... I posit this mainly to figure out the bounds of what we'd support.
For example, I've certainly encouraged folks to have a singleton channel client side to do most of their requests - as it tends to avoid the anti-pattern of creating a new channel per request (which crops up frequently!).

veblush: Oh your code here is for a solution, right?

ctiller:
am I blocking this work?

Issue #22883, opened 2020-05-06
RPC failure status should include all useful information about the failure


markdroth:
See discussion in #21308.

markdroth:
Another interesting case we should consider: If the resolver never returns a result and RPCs fail due to DEADLINE_EXCEEDED, the error message should indicate that the resolver never returned a result.

apolcyn:
> Another interesting case we should consider: If the resolver never returns a result and RPCs fail due to DEADLINE_EXCEEDED, the error message should indicate that the resolver never returned a result.

I agree, I also think this "waiting to acquire resource X" would be helpful debug info for a lot of other things that are currently buried in logs. For example, there are a number of interesting things that an RPC can "stall" on:
* RPC can be stalled because the channel's LB policy hasn't picked a connection for the RPC
  - the connection-pick might be incomplete because:
     a) name resolution hasn't completed
     b) load balancer isn't responding (if relevant to the LB policy in use)
     c) connections are all in "connecting" state.
     ... etc.
* RPC can get stuck waiting for a slot in it's http/2 connection's "max concurrent streams" queue.
* RPC can have an active send message that it hasn't yet written to the wire because it's waiting for space in the outgoing HTTP/2 flow control window
.....etc.

If these things could represented e.g. with some counters, and be accounted for on a per-RPC basis, then they could then be tied back to each RPC's context and be surfaced up to users either in an RPC's error message or perhaps alongside it - IMO that would useful to debug why some RPC's fail and potentially also why some RPC's are slow (e.g. determining that some specific high latency RPC took a long time because it was spending all of it's time waiting for name resolution to finish).

I'm not sure if perhaps the census context could be leveraged for these kinds of things....

markdroth:
This issue is only about things that cause RPCs to fail, so it would not include things like flow control.  Those things are better diagnosed by (e.g.) the application noticing that a send_message op takes too long to complete.

I don't think we need to use census for this, and I don't think we need a new API; I think it's sufficient to just include the relevant information in the status message we return along with the status code (typically UNAVAILABLE for these cases).

But regardless of what API we use to return the information, the important thing is to structure our internals such that all of the information is available at the point at which the status is being generated.

markdroth:
Another example: When a wait_for_ready RPC hits its deadline, the status message should include the error from the last LB pick.

markdroth:
[...]

PR #29867, opened 2022-06-02
LB policy API: add status to subchannel connectivity state notifications

markdroth:
Built on #28339.

Another small step toward https://github.com/grpc/grpc/issues/22883.


src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -656,7 +657,15 @@ class ClientChannel::SubchannelWrapper : public SubchannelInterface {
       // Ignore update if the parent WatcherWrapper has been replaced
       // since this callback was scheduled.
       if (watcher_ != nullptr) {
-        watcher_->OnConnectivityStateChange(state_change.state);
+        // Propagate status only in state TF.
+        // We specifically want to avoid propagating the status for
```
yashykt, line 661: What would happen if propagate this anyway?

src/core/ext/filters/client_channel/lb_policy/outlier_detection/outlier_detection.cc
```diff
@@ -176,27 +176,34 @@ class OutlierDetectionLb : public LoadBalancingPolicy {
 
       void Eject() {
         ejected_ = true;
-        if (last_seen_state_.has_value() &&
-            *last_seen_state_ != GRPC_CHANNEL_TRANSIENT_FAILURE) {
-          watcher_->OnConnectivityStateChange(GRPC_CHANNEL_TRANSIENT_FAILURE);
+        if (last_seen_state_.has_value()) {
```
yashykt, line 179: is this change for nor checking the last_seen_state value intentional?

yashykt (approved):

src/core/ext/filters/client_channel/client_channel.cc
markdroth: I don't know that it would necessarily break anything, but it would definitely not be the intended semantic that we want to expose to the LB policy API.  The intended semantic is that a non-OK status is returned only in state TRANSIENT_FAILURE.

The only reason that this is an issue is the special-case hack that we added in #29428 to have the real subchannel return a non-OK status in IDLE state so that it has a way of returning the keepalive info.  We don't want to expose that hack to the subchannel interface that is used by the LB policies.

src/core/ext/filters/client_channel/lb_policy/outlier_detection/outlier_detection.cc
markdroth: Yes.  The outlier detection policy indicates an ejected subchannel by reporting TRANSIENT_FAILURE for that subchannel instead of the real connectivity state.  This means that it's possible to have "duplicate" TRANSIENT_FAILURE notifications if we eject a subchannel whose real state was already TRANSIENT_FAILURE or if we uneject a subchannel whose real state is TRANSIENT_FAILURE.

Previously, we were avoiding sending these duplicate notifications by checking whether the real state was already TRANSIENT_FAILURE.  However, now that we're sending the status along with the connectivity state, the "duplicate" notifications aren't actually duplicates anymore, because even though the connectivity state isn't changing, the status is.  And we don't want to avoid updating the status, because that will lead to incorrect RPC failure status messages -- e.g., we might report "subchannel ejected by outlier detection" when the real problem is actually a connectivity issue, or vice versa.  To avoid that, I've changed the code here to send the notifications on ejection and unejection even if the connectivity state is not changing.

PR #29868, opened 2022-06-02
Backport "Support for prebuilt Ruby binary on x64-mingw-ucrt platform (#29684)" to 1.47.x

apolcyn:
* Add support for prebuilt Ruby binary on x64-mingw-ucrt platform, which is used for Windows RubyInstaller Ruby 3.1+


johnnyshields:
Thank you!

PR #28898, opened 2022-02-16
Status on

veblush:
@nicolasnoble

PR #29869, opened 2022-06-02
Enable GRPC_ERROR_IS_ABSEIL_STATUS

veblush:
Roll-forward #28898

### DIFF
--- a/include/grpc/impl/codegen/port_platform.h
+++ b/include/grpc/impl/codegen/port_platform.h
@@ -49,10 +49,11 @@
 #endif  // GPR_ABSEIL_SYNC
 
 /*
- * Defines GRPC_ERROR_IS_ABSEIL_STATUS to use absl::Status for grpc_error_handle
+ * Defines GRPC_ERROR_IS_NOT_ABSEIL_STATUS to not use absl::Status for
+ * grpc_error_handle. This is a temporary knob for migration process.
  */
-#ifndef GRPC_ERROR_IS_ABSEIL_STATUS
-// #define GRPC_ERROR_IS_ABSEIL_STATUS 1
+#ifndef GRPC_ERROR_IS_NOT_ABSEIL_STATUS
+#define GRPC_ERROR_IS_ABSEIL_STATUS 1
 #endif
 
 /* Get windows.h included everywhere (we need it) */
--- a/BUILD
+++ b/BUILD
@@ -144,8 +144,8 @@ config_setting(
 )
 
 config_setting(
-    name = "use_abseil_status",
-    values = {"define": "use_abseil_status=true"},
+    name = "disable_use_abseil_status",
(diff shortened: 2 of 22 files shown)


veblush:
[grpc_build_artifacts_multiplatform](https://fusion2.corp.google.com/invocations/b58a262f-55e9-4d88-acdb-9da440ab7171/targets): running

include/grpc/impl/codegen/port_platform.h
```diff
@@ -52,7 +52,7 @@
  * Defines GRPC_ERROR_IS_ABSEIL_STATUS to use absl::Status for grpc_error_handle
  */
 #ifndef GRPC_ERROR_IS_ABSEIL_STATUS
```
ctiller, line 54: Shouldn't we just remove this?

veblush: That will be done sometime after this change is proven safe.

ctiller: Should we leave the configuration in BUILD then, to be able to quickly toggle it?

ctiller:
Noting the large bloat increase: this change is likely to be requested rolled back due to a size increase on mobile platforms.

include/grpc/impl/codegen/port_platform.h
veblush: Yeah, that's a good idea. I just reintroduced it again.

veblush:
> Noting the large bloat increase: this change is likely to be requested rolled back due to a size increase on mobile platforms.

Good point. And I briefly checked with the internal iOS sample app. (OSSDemo.app) 
Build without and without this change;
- 17,333,514 bytes
- 17,596,426 bytes (1.5% increase)

Actual size increase is smaller than 8.0% so it'd be okay to have it. But it's worth investigation.

veblush:
From 
- [Bloaty test result](https://gist.github.com/veblush/c481995bf4a951645994debd4a3c7279)
- [Bloaty test result with -fno-exceptions](https://gist.github.com/veblush/359abe14c9ce60fa2dd35391095502ef)

It seems that having C++ object instead of the pointer to the struct inherently increases the code size because
- It needs to generate C++ exception stack unwinding because of `absl::Status`. Once exception is disabled (with -fno-exceptions), size increase decreased from +383Ki to +261Ki.
- Checking if error is none adds more code for `absl::Status` because it needs to create OkStatus() object and call Equal function. This overhead will be mitigated by https://github.com/grpc/grpc/pull/29937.

markdroth (approved):
LGTM, but please get Craig's approval as well before merging.

ctiller (approved):
Approved on the basis that there'll be a cherry pick in that we can roll back separately to the normal flow: I expect some pushback on the bloat issue.

PR #29875, opened 2022-06-02
[ObjC]fix 64 to 32 bit clang conversion warning in src/core/lib/transport

HannahShiSFB:

fix 64 to 32 bit clang conversion warning in src/core/lib/transport

@dennycd

### DIFF
--- a/src/core/lib/transport/metadata_batch.h
+++ b/src/core/lib/transport/metadata_batch.h
@@ -498,7 +498,7 @@ class ParseHelper {
     return ParsedMetadata<Container>(
         trait,
         ParseValueToMemento<typename Trait::MementoType, Trait::ParseMemento>(),
-        transport_size_);
+        static_cast<uint32_t>(transport_size_));
   }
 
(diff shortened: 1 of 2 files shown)


src/core/lib/transport/metadata_batch.h
```diff
@@ -488,7 +488,7 @@ class ParseHelper {
     return ParsedMetadata<Container>(
         trait,
         ParseValueToMemento<typename Trait::MementoType, Trait::ParseMemento>(),
-        transport_size_);
+        (uint32_t)transport_size_);
```
ctiller, line 491: please no C style casts

HannahShiSFB:
> Will there be a test in place once this is done to ensure no regressions?

We plan to make these warnings as errors after these are done so regressions will break the build

src/core/lib/transport/metadata_batch.h
HannahShiSFB: comment addressed

ctiller (approved):

PR #29877, opened 2022-06-02
[ObjC]fix 64 to 32 bit clang conversion warning in src/core/lib

HannahShiSFB:

fix 64 to 32 bit clang conversion warning in src/core/lib

@dennycd

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
@@ -477,9 +477,9 @@ void PriorityLb::ChoosePriorityLocked() {
     }
   }
   // Did not find any child in CONNECTING, delegate to last child.
-  SetCurrentPriorityLocked(config_->priorities().size() - 1,
-                           /*deactivate_lower_priorities=*/false,
-                           "no usable children");
+  SetCurrentPriorityLocked(
+      static_cast<int32_t>(config_->priorities().size() - 1),
+      /*deactivate_lower_priorities=*/false, "no usable children");
 }
 
 void PriorityLb::SetCurrentPriorityLocked(int32_t priority,
--- a/src/core/lib/gprpp/bitset.h
+++ b/src/core/lib/gprpp/bitset.h
@@ -157,7 +157,7 @@ class BitSet {
   ToInt() const {
     Int result = 0;
     for (size_t i = 0; i < kTotalBits; i++) {
-      if (is_set(i)) result |= (Int(1) << i);
+      if (is_set(static_cast<int>(i))) result |= (Int(1) << i);
     }
     return result;
   }
--- a/src/core/lib/gprpp/bitset.h
+++ b/src/core/lib/gprpp/bitset.h
@@ -156,8 +156,8 @@ class BitSet {
                           Int>::type
   ToInt() const {
     Int result = 0;
-    for (size_t i = 0; i < kTotalBits; i++) {
-      if (is_set(static_cast<int>(i))) result |= (Int(1) << i);
+    for (int i = 0; i < kTotalBits; i++) {
+      if (is_set(i)) result |= (Int(1) << i);
     }
     return result;
   }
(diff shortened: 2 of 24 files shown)


src/core/lib/gprpp/bitset.h
```diff
@@ -163,7 +163,7 @@ class BitSet {
                           Int>::type
   ToInt() const {
     Int result = 0;
-    for (size_t i = 0; i < kTotalBits; i++) {
+    for (int i = 0; i < kTotalBits; i++) {
```
markdroth, line 166: @ctiller, can you please provide a sanity check that this is safe?

src/core/lib/surface/channel.h
```diff
@@ -145,7 +145,7 @@ class Channel : public RefCounted<Channel>,
 
   int TestOnlyRegisteredCalls() {
     MutexLock lock(&registration_table_.mu);
-    return registration_table_.map.size();
+    return static_cast<int>(registration_table_.map.size());
```
markdroth, line 148: Instead of doing a cast here, how about changing this function to return `size_t`?

src/core/lib/gprpp/bitset.h
ctiller: eh... it's *probably* safe but weird enough that I'm curious to see the warning/error that triggered the change in case something systemically is causing these odd conversions.

src/core/lib/surface/channel.h
HannahShiSFB: It can be done, that’ll require changes in the call stack in `ChannelTestPeer::registered_calls` and `End2endTest.ManyStubs`

markdroth: That sounds fine.

src/core/lib/gprpp/bitset.h
HannahShiSFB: changed to 
for (size_t i = 0; i < kTotalBits; i++) {
      if (is_set(static_cast<int>(i))) result |= (Int(1) << i);
    }

dennycd:
please rebase on latest master for some of the build breaks seen in objc ios test. thanks

src/core/lib/surface/channel.h
HannahShiSFB: comment addressed

src/core/lib/gprpp/bitset.h
```diff
@@ -154,7 +154,7 @@ class BitSet {
   ToInt() const {
     Int result = 0;
     for (size_t i = 0; i < kTotalBits; i++) {
```
ctiller, line 156: would it be better to just say `int i` here, and make kTotalBits an int too?

HannahShiSFB: addressed

HannahShiSFB: changed back to size_t. I tried to change to int, but it turned out that more warnings showed, more castings needed.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -477,9 +477,9 @@ void PriorityLb::ChoosePriorityLocked() {
     }
   }
   // Did not find any child in CONNECTING, delegate to last child.
-  SetCurrentPriorityLocked(config_->priorities().size() - 1,
-                           /*deactivate_lower_priorities=*/false,
-                           "no usable children");
+  SetCurrentPriorityLocked(
+      static_cast<int32_t>(config_->priorities().size() - 1),
```
markdroth, line 481: This looks completely wrong.  Using `size_t` is correct here.

markdroth (commented):
This was not approved by anyone in the C-core team.  I don't think this should have been merged.

I see several casts here that don't look right.

sampajano:
> This was not approved by anyone in the C-core team. I don't think this should have been merged.
> 
> I see several casts here that don't look right.

Ahh i see. Apologize for the premature merge. I'm still learning the process :)

Will revert now and work with Hannah to re-send this PR with c-core reviews.

Issue #29848, opened 2022-05-31
_watch_connectivity_state uses an incorrect error message


genericmoniker:
### Is your feature request related to a problem? Please describe.

My application makes many calls via gRPC. Sometimes in my logs I see stack traces like this:

```
Traceback (most recent call last):
File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
self.run()
File "/usr/lib64/python3.6/threading.py", line 864, in run
self._target(*self._args, **self._kwargs)
File "/opt/lib64/python3.6/site-packages/grpc/_channel.py", line 914, in _poll_connectivity
time.time() + 0.2)
File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 487, in grpc._cython.cygrpc.Channel.watch_connectivity_state
File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 349, in grpc._cython.cygrpc._watch_connectivity_state
File "src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi", line 357, in grpc._cython.cygrpc._watch_connectivity_state
ValueError: Cannot invoke RPC: Channel closed!
```
This tells me that I have a problem, but isn't very helpful in narrowing down where it is since the stack trace is entirely grpcio and standard library threading.

### Describe the solution you'd like

It would help narrow down the problem if the exception message could include the gRPC "method" being called. Maybe there is other information available that would also be helpful?

### Additional context

I also run into this in unit testing, where I usually try to mock out any actual RPCs. Occasionally a patch will "miss" and a real RPC will be attempted and I'll see a "ValueError: Cannot invoke RPC: Channel closed!" exception. It isn't always repeatable, though, and I'm using the xdist plugin to pytest which runs the tests in parallel. So I'll see dots for each test as it runs, and a stack trace like the above in the middle of the dots. If I run w/o parallel tests to print each test as it runs, often the problem doesn't happen again (or at least after a few attempts). Again it would help to know which RPC is being attempted.

gnossen:
I agree that this error message isn't great. In fact, even the "Cannot invoke RPC" message seems to be wrong. Unless I'm mistaken, it looks like this code path is only reachable from [`Channel.subscribe`](https://grpc.github.io/grpc/python/_modules/grpc.html#Channel.subscribe). I think the error message should probably be something more like "Unable to monitor channel state: channel closed". That's one thing that should definitely be fixed.

[...]

PR #29882, opened 2022-06-02
Fix Channel.subscribe error message

gnossen:
Fixes https://github.com/grpc/grpc/issues/29848

### DIFF
--- a/src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi
+++ b/src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi
@@ -382,7 +382,7 @@ cdef object _watch_connectivity_state(
           state.c_connectivity_completion_queue, <cpython.PyObject *>tag)
       state.connectivity_due.add(tag)
     else:
-      raise ValueError('Cannot invoke RPC: %s' % state.closed_reason)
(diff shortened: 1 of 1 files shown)


src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi
```diff
@@ -382,7 +382,7 @@ cdef object _watch_connectivity_state(
           state.c_connectivity_completion_queue, <cpython.PyObject *>tag)
       state.connectivity_due.add(tag)
     else:
-      raise ValueError('Cannot invoke RPC: %s' % state.closed_reason)
+      raise ValueError('Cannot monitor channel state: %s' % state.closed_reason)
```
lidizheng, line 385: Should we update the type of the exception to be `RuntimeError`?

lidizheng (approved):
Thanks for the fix.

src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi
gnossen: While I would agree if this were new code, someone could be depending on the particular type of this exception, so this could be considered a behavior change. Leaving this as it is for now.

PR #29905, opened 2022-06-03
Revert "Revert "EventEngine::RunAt - Subchannel connection retries""

Vignesh2208:
Reland event engine timer usage after fixes for internal breakages. The failures were due to missing ExecCtx and un-initialized grpc during subchannel destruction.

@drfloob


src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -952,11 +957,28 @@ void Subchannel::OnConnectingFinishedLocked(grpc_error_handle error) {
             time_until_next_attempt.millis());
     SetConnectivityStateLocked(GRPC_CHANNEL_TRANSIENT_FAILURE,
                                grpc_error_to_absl_status(error));
+    // Ensure grpc remains initialized when the timer callback runs. The
+    // subchannel could get deleted when the timer callback is run and
+    // subchannel deletion may destroye some pollset_sets which required grpc
+    // to be initialized.
+    grpc_init();
```
markdroth, line 964: Instead of doing this surrounding only the timer callback, I think we should do this init in the subchannel's ctor and do shutdown in the subchannel's dtor.  That will make the code simpler by having only one place where we need to do shutdown, and it will also catch edge cases other than timer callbacks, which we've run into in the past.

markdroth, line 964: Note that adding grpc init and shutdown to the subchannel code has been tried in the past but had to be reverted.  I don't know what the exact problem with this was in the past, but I think this is the right solution.  So I suggest running a TGP on this change to see if there are any problems, and if there are, we should figure out how to address them to make this work properly.

markdroth (changes requested):

src/core/ext/filters/client_channel/subchannel.cc
Vignesh2208: We already tried and failed to do this last year. It caused lot of flakes to c# tests (b/198765101) because c# runtime exists before grpc_shutdown and unregisters a logging function which causes subsequent gpr_log function calls to fail at the time of grpc_shutdown. So had to revert that PR: https://github.com/grpc/grpc/pull/27004

Vignesh2208: On the other hand, I just remembered we deleted C#, I'll update it

Vignesh2208: TGP passed for this PR (cl/453263832). I think we can give it a shot and land it

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -784,10 +799,15 @@ void Subchannel::RequestConnection() {
 }
 
 void Subchannel::ResetBackoff() {
+  // Hold a ref to ensure cancellation and subsequent deletion of the closure
+  // does not eliminate the last ref and destroy the Subchannel before the
+  // method returns.
+  auto self = WeakRef(DEBUG_LOCATION, "ResetBackoff");
```
markdroth, line 805: I don't think this is needed anymore.

markdroth (commented):
One small thing left to clean up here.  Otherwise, looks good.

src/core/ext/filters/client_channel/subchannel.cc
Vignesh2208: Wouldn't removing this possibly lead to a situation where OnRetryTimerLocked method after the subchannel is destroyed ? EventEngine::Cancel may end up deleting the subchannel

markdroth: Oh, you're right.  Sorry, I was misreading this.

markdroth (approved):

PR #29908, opened 2022-06-04
XdsClient reverts back to calling watcher's OnError() when resource timer fires

yousukseung:


markdroth:
I followed up on the internal email thread.  I don't think this is the right solution.

PR #29922, opened 2022-06-06
google-c2p resolver: use federation if enabled via env var

markdroth:
I know we're not ready to merge this yet, but I wanted to get the change ready and have it reviewed, so that it's ready to go when we're ready to make the change.


apolcyn (commented):
@markdroth what do you think about guarding this under the `GRPC_EXPERIMENTAL_XDS_FEDERATION` env var?

We could then merge this risk-free and start getting integration test coverage.

apolcyn:
FYI I've manually tested in and end-to-end setup that this works, when `GRPC_EXPERIMENTAL_XDS_FEDERATION` is set true.

markdroth:
I've changed this to trigger based on the federation env var, and I've merged in your changes from #31471, but in a slightly different way.  PTAL.

apolcyn (approved):

PR #29927, opened 2022-06-07
Fixing use after free bug due to invocation gprc_call_get_peer after grpc_call_cancel

Vignesh2208:
Fixes b/222144056, b/227951292

@markdroth

### DIFF
--- a/src/core/lib/surface/call.cc
+++ b/src/core/lib/surface/call.cc
@@ -770,6 +770,16 @@ void FilterStackCall::CancelWithError(grpc_error_handle error) {
     GRPC_ERROR_UNREF(error);
     return;
   }
+  // (b/222144056): During normal operation, call->peer_string is made to point
+  // to a string/memory allocated by the transport. After the above ExecCtx
+  // flush due to grpc_call_cancel, the transport may be destroyed. When the
+  // transport gets destroyed, it frees the allocated peer_string. However,
+  // call->peer_string still continues to point to that free'ed memory location
+  // and may cause a heap-use after free error if there's an attempt to access
+  // it again using grpc_call_get_peer API call. By setting it to nullptr here
+  // we prevent such errors by forcing any grpc_call_get_peer invocation
+  // after a call cancellation to return nullptr.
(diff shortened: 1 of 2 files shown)


markdroth:
Let me re-ask the same question I asked in #29069:

Do we know that this problem exists only during cancellation? Is there also a window in normal call finishing during which the transport is destroyed before the `grpc_call` object? If so, the same thing could happen in that case.

markdroth:
It was done without making a copy for performance reasons.  The peer string is returned to the surface from the transport as part of the completion of the send_initial_metadata op on the client[1] or the recv_initial_metadata op on the server.  However, we don't want to make a copy of the string whenever those ops occur, since we don't actually need the value unless the application actually asks for it, which isn't that common.  So the value we return to the surface is basically just a pointer to a string owned by the transport rather than being a copy.  We make a copy only when the application actually asks for the value.

IIRC, the reason it's atomic is that we didn't want to have to acquire a lock when the application accesses the value.  I'm not sure how much performance overhead a lock would actually impose here, though, since this lock would be uncontended most of the time.

Getting back to the main question, though... I guess what I'd like to know here is how exactly this happens in cancellation?  If I better understood that use-case, I might be able to help reason through whether this is likely to occur in non-cancellation cases.  In principle, I don't see how this can happen, because once a transport is chosen, the transport should stay at the bottom of the call stack until the call stack is destroyed, and that doesn't happen until the grpc_call object itself is destroyed.  Is it possible that the problem here is not that the transport is being destroyed but rather that the transport is destroying the peer string before the last ref to the transport goes away?  If so, can we fix the problem in the transport instead?

If I'm wrong about the above and there are actually cases where the transport goes away before the call does, then making this string ref-counted would work, although it might be a bit of overkill.  Another option might be that if we can get a signal up to the surface when the transport is being destroyed, we can set the peer string pointer back to null at that point.


[1] Side note: When retries are in use, the value may not actually be correct on the client until after the recv_initial_metadata op completes; if we ever do a gRPC 2.0, one of the things I'd like us to change is to not make the value available at all until after recv_initial_metadata is done.

PR #29934, opened 2022-06-07
Disable flaky grpc_authz test case on mac

jtattermusch:
- based on historical data, this seems to be the test case that is responsible for vast majority of flakiness (see https://source.cloud.google.com/results/invocations/ce30bd5a-e0ba-4c02-8e8b-7c54e2b243da/targets/%2F%2Ftest%2Fcore%2Fend2end:h2_full_no_retry_test@grpc_authz/log for reference) 
- fix the test label printed to the logs (so actually we want to disable `test_file_watcher_recovers_from_failure`)

### DIFF
--- a/test/core/end2end/tests/grpc_authz.cc
+++ b/test/core/end2end/tests/grpc_authz.cc
@@ -668,7 +668,7 @@ static void test_file_watcher_recovers_from_failure(
   grpc_channel_args server_args = {GPR_ARRAY_SIZE(args), args};
 
   grpc_end2end_test_fixture f = begin_test(
-      config, "test_file_watcher_valid_policy_reload", nullptr, &server_args);
+      config, "test_file_watcher_recovers_from_failure", nullptr, &server_args);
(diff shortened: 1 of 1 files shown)


jtattermusch:
This time the actually `test_file_watcher_valid_policy_reload` test case has failed: 
https://source.cloud.google.com/results/invocations/5ed60307-4c02-4633-8a14-dc9f95f579bc/targets/%2F%2Ftest%2Fcore%2Fend2end:h2_fd_test@grpc_authz/log

ashithasantosh:
Thank you for updating the test name. I will close my PR (https://github.com/grpc/grpc/pull/29920/files) then.

https://screenshot.googleplex.com/9do2iJQkQyCbpG3.png
The tests fail with deadline exceeded error and the delay is usually in grpc_load_file block. Somehow it takes several seconds for load file to complete. I have some more runs to try out, will update the bug soon with more details.

ashithasantosh (approved):

PR #29955, opened 2022-06-08
Add ExecCtx to subchannel destructor

Vignesh2208:
Fixes internal crashes.

@ctiller

### DIFF
--- a/src/core/ext/filters/client_channel/subchannel.cc
+++ b/src/core/ext/filters/client_channel/subchannel.cc
@@ -697,6 +697,7 @@ Subchannel::Subchannel(SubchannelKey key,
 }
 
 Subchannel::~Subchannel() {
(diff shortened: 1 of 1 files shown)


markdroth:
I'm not sure what crash you're seeing, but this really does not look like the right fix to me.  The subchannel should not be destroyed in any context in which an ExecCtx is not already present on the stack.  If it is, then we should fix that problem instead of adding a duplicate ExecCtx here.

Vignesh2208:
> I'm not sure what crash you're seeing, but this really does not look like the right fix to me. The subchannel should not be destroyed in any context in which an ExecCtx is not already present on the stack. If it is, then we should fix that problem instead of adding a duplicate ExecCtx here.

Yes this is a bad idea. closing the PR

PR #29956, opened 2022-06-08
Add ring_hash_lb to trace logs for Python/C++ xds interop clients

lidizheng:
This PR allows us to better debug issues related to ring_hash_lb (like b/231690283).

However, this flag won't be automatically backported to older releases, and the flags don't include all xDS features (like fault injection). In future, we might want a more controllable way of updating what kinds of logs we are generating. For example, documenting this step somewhere.

PR #29958, opened 2022-06-08
xDS Interop: Add ring_hash_lb traces

yashykt:
Context - https://b.corp.google.com/issues/231690283


markdroth:
Looks like this is a dup of #29956.

PR #29925, opened 2022-06-06
xds-k8s: Fix the issue with Java PSM security tests skipped

sergiitk:
1. Fixes the issue with Java PSM security tests accidentally skipped because Java was missing from the list of languages, ref https://github.com/grpc/grpc/pull/28978
2. Invert the logic of `is_supported` methods, making them normally open
3. Make languages an `enum.Flag` to avoid accidental typos when listing the languages
4. Rename `XdsKubernetesTestCase.isSupported` to `XdsKubernetesTestCase.is_supported` to be consistent with `XdsUrlMapTestCase.is_supported`
5. Add extra logging

PR #29960, opened 2022-06-08
Add supported Node version ranges in xDS k8s url_map tests

murgatroid99:
This adds is_supported implementations for most of the url_map tests that didn't already have them. The exception is metadata_filter_test because it doesn't use any specific client features.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/tests/url_map/csds_test.py
+++ b/tools/run_tests/xds_k8s_test_driver/tests/url_map/csds_test.py
@@ -19,6 +19,7 @@ from absl.testing import absltest
 
 from framework import xds_url_map_testcase
 from framework.test_app import client_app
+from framework.helpers import skips
 
 # Type aliases
 HostRule = xds_url_map_testcase.HostRule
@@ -28,6 +29,7 @@ DumpedXdsConfig = xds_url_map_testcase.DumpedXdsConfig
 RpcTypeUnaryCall = xds_url_map_testcase.RpcTypeUnaryCall
 RpcTypeEmptyCall = xds_url_map_testcase.RpcTypeEmptyCall
 XdsTestClient = client_app.XdsTestClient
+_Lang = skips.Lang
 
 logger = logging.getLogger(__name__)
 flags.adopt_module_key_flags(xds_url_map_testcase)
@@ -36,6 +38,12 @@ _NUM_RPCS = 50
 
 
 class TestBasicCsds(xds_url_map_testcase.XdsUrlMapTestCase):
+
+    @staticmethod
+    def is_supported(config: skips.TestConfig) -> bool:
+        if config.client_lang == _Lang.NODE:
+            return not config.version_lt('v1.5.x')
+        return True
 
     @staticmethod
     def url_map_change(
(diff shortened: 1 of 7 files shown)


tools/run_tests/xds_k8s_test_driver/tests/url_map/timeout_test.py
```diff
@@ -81,6 +81,8 @@ class TestTimeoutInRouteRule(_BaseXdsTimeOutTestCase):
 
     @staticmethod
     def is_supported(config: skips.TestConfig) -> bool:
+        if config.client_lang == skips.Lang.NODE:
+            return not config.version_lt('v1.4.x')
```
sergiitk, line 85: See @lidizheng's comment below. Does it mean javascript also supports what Lidi's talking about?

sergiitk (approved):
LGTM

tools/run_tests/xds_k8s_test_driver/tests/url_map/csds_test.py
```diff
@@ -37,6 +39,12 @@
 
 class TestBasicCsds(xds_url_map_testcase.XdsUrlMapTestCase):
 
+    @staticmethod
+    def is_supported(config: skips.TestConfig) -> bool:
+        if config.client_lang == _Lang.NODE:
+            return not config.version_lt('v1.5.x')
```
lidizheng, line 45: How about:

```py
return config.version_gte('v1.5.x')
```

tools/run_tests/xds_k8s_test_driver/tests/url_map/timeout_test.py
murgatroid99: That comment is on a `server_lang` check, because only the Java server handles the specified metadata properly. This is a `client_lang` check, so it's a separate issue.

tools/run_tests/xds_k8s_test_driver/tests/url_map/csds_test.py
murgatroid99: I wrote it this way to be consistent with the checks @sergiitk added in #29925.

lidizheng: If we can simplify the condition, I would recommend to do it. But it's not a hard blocker in any way.

tools/run_tests/xds_k8s_test_driver/tests/url_map/timeout_test.py
sergiitk: I've noticed this was a `server_lang` check, and that's what surprised me. I know sometimes other languages that have no server-side support use java. So my question is 

```
        # TODO(lidiz) either add support for rpc-behavior to other languages, or we
        # should always use Java server as backend.
```

Does it mean JS server side supports this?

tools/run_tests/xds_k8s_test_driver/tests/url_map/csds_test.py
sergiitk: Yea, I'm fine one way or another. I swapped to `not config.version_lt` expression everywhere so when I was changing the logic from normally-closed to normally open. This was

a) To emphasise that "is_supported _not_ when version less than ..."
b) if check fails for some reason (like version doesn't parse), and the check return `False`, then we still run it.

tools/run_tests/xds_k8s_test_driver/tests/url_map/timeout_test.py
murgatroid99: No, there is no JS server implementation. The code I added here is a `client_lang` check. It doesn't say anything about the server side. It's just a check for the JS version that has the client side implementation of this feature, like the checks in all of the other tests.

PR #29965, opened 2022-06-09
xDS Istop Interop: Add differentiation on the protocol being used

yashykt:
Changes -
1. Figure out whether a port uses xDS (builder and creds) / TLS creds / insecure creds based on the command line flags. (Note that we do not listen on http/tcp specific ports at the moment since we do not care about those.)
2. `ForwardEcho` handler - Use `XdsCreds` when the scheme is `xds`, insecure creds when the scheme is `grpc` (not adding support for tls at the moment) and fail all other schemes. (In the future, we will probably forward requests with all other schemes handling to something else.)

### DIFF
--- a/test/cpp/interop/istio_echo_server.cc
+++ b/test/cpp/interop/istio_echo_server.cc
@@ -20,6 +20,7 @@
 #include <condition_variable>
 #include <deque>
 #include <map>
+#include <memory>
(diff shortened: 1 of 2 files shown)


veblush (approved):

yashykt:
Thanks for the review!

PR #29970, opened 2022-06-09
Status: Use move instead of copy for std::string in constructors

yashykt:


ctiller:
Please write a gRFC for this public API change.
Also, please note that deploying this change makes it harder for us to migrate to `absl::Status` in the long term (which takes a `string_view` for these arguments), and so it seems likely this change is undesirable.

yashykt:
> Please write a gRFC for this public API change.
Hmm, I didn't think this would need a gRFC.

> Also, please note that deploying this change makes it harder for us to migrate to absl::Status in the long term (which takes a string_view for these arguments), and so it seems likely this change is undesirable.
good point! Thanks!

ctiller:
I'd take the stance that if it's anything more than a comment it certainly requires a gRFC (note the nuances in this relatively simple example) - and that some comment changes likely also warrant a gRFC written.

PR #29972, opened 2022-06-09
[xDS interop] Fix the affinity test support range

lidizheng:
As title.

This matches the description in https://github.com/grpc/grpc/blob/master/doc/grpc_xds_features.md.

Related https://github.com/grpc/grpc-go/pull/5411 and https://github.com/grpc/grpc-go/pull/5410

This PR can't be tested since we don't have k8s-lb jobs for v1.39.x and v1.40.x branches.


sergiitk (approved):

lidizheng:
Thanks for the review!

PR #29976, opened 2022-06-09
Tcp connect cancellation implementation for posix

Vignesh2208:
This implementation relies on using sharded hashmaps to track pending connections. The shard count can be easily changed if contention is observed.

@ctiller


src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -62,8 +63,37 @@ struct async_connect {
   grpc_endpoint** ep;
   grpc_closure* closure;
   grpc_channel_args* channel_args;
+  int64_t connection_handle;
+  bool connect_cancelled;
 };
 
+namespace {
+gpr_once g_tcp_client_posix_init = GPR_ONCE_INIT;
+size_t g_pending_connection_shards = 0;
```
markdroth, line 72: How about something like this to minimize the number of globals needed:

```
struct ConnectionShard {
  grpc_core::Mutex mu;
  absl::flat_hash_map<int64_t, async_connect*> pending_connections ABSL_GUARDED_BY(&mu);
};
std::vector<ConnectionShard>* g_connection_shards = nullptr;
```

This can also be initialized more simply:

```
size_t num_shards = std::max(2 * gpr_cpu_num_cores(), 1u);
g_connection_shards = new std::vector<ConnectionShard>(num_shards);
```

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -138,6 +168,7 @@ static void on_writable(void* acp, grpc_error_handle error) {
   grpc_closure* closure = ac->closure;
   std::string addr_str = ac->addr_str;
   grpc_fd* fd;
+  bool connect_cancelled = false;
```
markdroth, line 171: Might as well just declare this on line 184, where it actually gets used.

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -342,13 +410,61 @@ static int64_t tcp_connect(grpc_closure* closure, grpc_endpoint** ep,
...
+  gpr_mu_lock(&ac->mu);
+  bool connection_cancel_success = (ac->fd != nullptr);
+  if (connection_cancel_success) {
+    // Connection is still pending. The on_writable callback hasn't executed
+    // yet because ac->fd != nullptr.
+    ac->connect_cancelled = true;
+    // Force on_writable to run as soon as possible and gracefully terminate.
+    // When the on_writable callback runs, it would detect that a connect
+    // cancellation is requested and take the necessary actions.
+    grpc_fd_set_writable(ac->fd);
```
markdroth, line 455: Could we trigger this by simply shutting down the fd?

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -190,6 +190,45 @@ void test_fails(void) {
   gpr_log(GPR_ERROR, "---- finished test_fails() ----");
 }
 
+void test_connect_cancellation_succeeds(void) {
```
markdroth, line 193: How about also testing the case where the cancellation fails because the connection attempt is already complete?

markdroth (commented):
Thanks for doing this!

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Yes thats a good idea. Added this in lines 198 and 145

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: Updated to use this struct.

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: Fixed.

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: Yes I think that should also work. I updated it. But the thread calling grpc_tcp_cancel_connect would need to have an ExecCtx now. Is that okay ? Otherwise we can revert back to grpc_fd_set_writable

Vignesh2208:
> Looks great!
> 
> One more thing I just noticed: Before merging, please add a comment in tcp_client.h clarifying that if cancellation is successful, the callback will not be invoked.
> 
> Thanks!

Thanks added it now.

dennycd:
@Vignesh2208  this one may have caused our ios cronet test crash, see [grpc-ios/#105](https://github.com/grpc/grpc-ios/issues/105) for more details, thanks !

PR #29981, opened 2022-06-10
Update CMakeLists.txt to tell MSVC using utf8 lang/c++

hongnod:
Tell MSVC to build grpc using utf-8


drfloob:
@hongnod CMakeLists.txt is a generated file. The change needs to be made in `templates/CMakeLists.txt.template`, then the build files need to be regenerated using ./tools/buildgen/generate_projects.sh.

hongnod:
@drfloob I see thansk！

PR #29989, opened 2022-06-12
Add a flag to control usage of min_progress_size in TCP for receive buffer allocation

Vignesh2208:
Add a flag to control usage of min_progress_size in TCP for receive buffer allocation. This could play an important role if the receiver is under high memory pressure.


ctiller:
Let's add something to tcp_posix_test.cc to prove that this works.

Vignesh2208:
> Let's add something to tcp_posix_test.cc to prove that this works.

Added a min_progress_size to grpc_endpoint_read in tcp_posix_test.cc

test/core/iomgr/tcp_posix_test.cc
```diff
@@ -295,7 +298,7 @@ static void large_read_test(size_t slice_size) {
   GRPC_CLOSURE_INIT(&state.read_cb, read_cb, &state, grpc_schedule_on_exec_ctx);
 
   grpc_endpoint_read(ep, &state.incoming, &state.read_cb, /*urgent=*/false,
-                     /*min_progress_size=*/1);
+                     /*min_progress_size=*/min_progress_size);
```
ctiller, line 301: is there some kind of assert that we can drop in here to prove that min_progress_size is respected>

Vignesh2208: I added an assert now. In the process, I discovered bugs in the way I was handling min_progress_size in tcp_posix.cc. It required some non-trivial amount of change (included in the latest commit). Could you ptal.

ctiller (approved):

PR #29990, opened 2022-06-12
Change secure endpoint write code to use max_frame_size to control encrypted frame sizes at the sender.

Vignesh2208:
Change secure endpoint write code to use max_frame_size to control encrypted frame sizes at the sender.


src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1014,14 +1020,26 @@ static void write_action_begin_locked(void* gt,
 
 static void write_action(void* gt, grpc_error_handle /*error*/) {
   GPR_TIMER_SCOPE("write_action", 0);
+  static bool kEnablePeerStateBasedFraming =
+      GPR_GLOBAL_CONFIG_GET(grpc_experimental_enable_peer_state_based_framing);
   grpc_chttp2_transport* t = static_cast<grpc_chttp2_transport*>(gt);
   void* cl = t->cl;
   t->cl = nullptr;
+  // If grpc_experimental_enable_peer_state_based_framing is set to true,
+  // choose max_frame_size as 2 * max http2 frame size of peer. If peer is under
```
ananda1066, line 1029: just curious, is the 2x factor arbitrary?

src/core/lib/security/transport/secure_endpoint.cc
```diff
@@ -410,8 +411,26 @@ static void endpoint_write(grpc_endpoint* secure_ep, grpc_slice_buffer* slices,
...
+      // Break the input slices into chunks of size = max_frame_size and call
+      // tsi_zero_copy_grpc_protector_protect on each chunk. This ensures that
+      // the protector cannot create frames larger than the specified
+      // max_frame_size.
+      while (slices->length > static_cast<size_t>(max_frame_size) &&
+             result == TSI_OK) {
+        grpc_slice_buffer_move_first(slices,
+                                     static_cast<size_t>(max_frame_size),
+                                     &ep->tmp_frame_size_clip_buffer);
+        result = tsi_zero_copy_grpc_protector_protect(
```
ananda1066, line 426: The answer is probably yes, but just to clarify - the new protected chunk will be appended to output buffer?

src/core/lib/security/transport/secure_endpoint.cc
```diff
@@ -410,8 +411,26 @@ static void endpoint_write(grpc_endpoint* secure_ep, grpc_slice_buffer* slices,
 
     if (ep->zero_copy_protector != nullptr) {
       // Use zero-copy grpc protector to protect.
-      result = tsi_zero_copy_grpc_protector_protect(ep->zero_copy_protector,
-                                                    slices, &ep->output_buffer);
+      grpc_slice_buffer_reset_and_unref_internal(
```
ananda1066, line 414: What is the reason to  reset / unref at the start of the if statement instead of at the end?

src/core/lib/security/transport/secure_endpoint.cc
```diff
@@ -143,7 +145,7 @@ struct secure_endpoint {
   grpc_core::MemoryAllocator::Reservation self_reservation;
   std::atomic<bool> has_posted_reclaimer;
   int min_progress_size;
-
+  grpc_slice_buffer tmp_frame_size_clip_buffer;
```
ananda1066, line 148: I think this var should be renamed to something clearer or less specific.. something like protector_staging_buffer maybe?

ananda1066 (changes requested):

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
Vignesh2208: yeah, the 2x is arbitrary right now. We will have to see how it performs. Making it a value < 1 might cause the sender to send very small frames - which would cause some overhead at the sender.

src/core/lib/security/transport/secure_endpoint.cc
Vignesh2208: Yeah the implementations we have only append to the output buffer. They don't reset the output buffer before appending.

src/core/lib/security/transport/secure_endpoint.cc
Vignesh2208: Good catch. Moved it to the end of the if statement.

src/core/lib/security/transport/secure_endpoint.cc
Vignesh2208: Done.

ananda1066:
Are the test failures unrelated?

ctiller (approved):

ctiller (commented):
Just thinking about this a little more... can we add something to secure_endpoint_test.cc to assert that this is working?

Vignesh2208:
> Just thinking about this a little more... can we add something to secure_endpoint_test.cc to assert that this is working?

secure_endpoint_test.cc calls grpc_endpoint_tests(...) defined in endpoint_tests.cc. I configured the max_frame_size argument in the grpc_endpoint_write API call made in that file. The tests there basically write some data over one endpoint in a connected endpoint pair and verify that the data read at the other endpoint equals the data that was written

ctiller:
Could we maybe inject an endpoint underneath the secure endpoint and assert the max write size is respected?

Vignesh2208:
> Could we maybe inject an endpoint underneath the secure endpoint and assert the max write size is respected?

 I added an intercept_endpoint to secure_endpoint_test now. The intercept endpoint will intercept every write, parse the frame size and verify that the protected data size does not exceed max_frame_size. It will then pass the slices down to tcp endpoint.

ctiller (approved):

ananda1066 (approved):

PR #29992, opened 2022-06-13
Change log level for skipping fork handlers to GPR_INFO

ederst:
This basically reverts https://github.com/grpc/grpc/pull/28566 completely which was already partly reverted with https://github.com/grpc/grpc/pull/29232, and addresses the unanswered question of @vakker:

> What about the E0506 14:52:21.476434491 1299177 fork_posix.cc:76] Other threads are currently calling into gRPC, skipping fork() handlers messages (from [here](https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/fork_posix.cc#L76))? They were both changed at the same time to ERROR from INFO, and this is still causing log clutter.

It seems that some projects using grpc experience log spamming due to this, for example the ones that I know of:
* https://github.com/ray-project/ray/issues/22518#issuecomment-1120827112
* Pulumi with grpc version 1.46.x

### DIFF
--- a/src/core/lib/iomgr/fork_posix.cc
+++ b/src/core/lib/iomgr/fork_posix.cc
@@ -73,7 +73,7 @@ void grpc_prefork() {
     return;
   }
   if (!grpc_core::Fork::BlockExecCtx()) {
-    gpr_log(GPR_ERROR,
+    gpr_log(GPR_INFO,
             "Other threads are currently calling into gRPC, skipping fork() "
             "handlers");
     return;


drfloob (approved):

drfloob:
CC @gnossen

Stephan-Walter:
Hi,

we experience also the described log spamming. Unfortunately no new package is available at PyPi until now. 

Is there a timeline until when we will see the fix also on PyPi? Otherwise we would have to check to workaround the problem, what I would prefer to avoid.

ederst:
I think the fix should be available since 1.48.x. At least the commit shows up there and we use this version without experiencing log spamming.

Stephan-Walter:
Ok, but it seems, that this has never reached [PyPi](https://pypi.org/project/grpcio/) for some reason.

Stephan-Walter:
Just to let you know. 1.48.1 is available since September 2nd and fix the log spam issue as expected.

So from my point of view the problem is now also solved for the Python users.

PR #29997, opened 2022-06-13
[gRPC/iOS] Interop test deflake

dennycd:
Deflake objc interop test by switching to runner-based ios test (https://github.com/grpc/grpc/pull/29740). Also splitting interop local and remote test suites so we can analyze and examine them separately 

---
cc @HannahShiSFB


sampajano (approved):
thanks!

PR #30001, opened 2022-06-13
round_robin: update status upon each new connection failure

markdroth:
Also makes the same change in ring_hash, although the bug isn't actually visible there.

Built on #29993.

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
@@ -269,6 +269,12 @@ class RingHash : public LoadBalancingPolicy {
     // The index of the subchannel currently doing an internally
     // triggered connection attempt, if any.
     absl::optional<size_t> internally_triggered_connection_index_;
+
+    // TODO(roth): If we ever change the helper UpdateState() API to not
+    // need the status reported for TRANSIENT_FAILURE state (because
+    // it's not currently actually used for anything outside of the picker),
+    // then we will no longer need this data member.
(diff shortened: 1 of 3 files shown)


src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
```diff
@@ -646,8 +652,17 @@ void RingHash::RingHashSubchannelList::UpdateRingHashConnectivityStateLocked(
...
+  // In TRANSIENT_FAILURE, report the last reported failure.
+  // Otherwise, report OK.
+  if (state == GRPC_CHANNEL_TRANSIENT_FAILURE) {
+    if (!status.ok()) {
+      last_failure_ = absl::UnavailableError(absl::StrCat(
+          "no reachable subchannels; last error: ", status.ToString()));
+    }
+    status = last_failure_;
+  } else {
+    status = absl::OkStatus();
```
yashykt, line 664: can a shutdown state reach here?

yashykt (approved):

src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
markdroth: No, subchannels never report SHUTDOWN.  The underlying ConnectedSubchannel does report SHUTDOWN, but in that case, the subchannel reports IDLE.

https://github.com/grpc/grpc/blob/86e282ba9792da83a11ac1ec03074ccc92e6d1aa/src/core/ext/filters/client_channel/subchannel.cc#L331

PR #30004, opened 2022-06-13
Call: Send cancel op down the stack even when no ops are sent

yashykt:
Fixes b/204470117

When we create a call, depending on the configuration of the stack and the call, we might take a few references on the call such as - `deadline_timer` and `chttp2`.

In the case that we unref the call without sending any ops, these refs don't get cleaned up, and we need to wait for those refs to 
 get cleaned up on their own, for example, waiting for the deadline timer to fire which would send a cancel op and let go of the refs. Even though this is not a leak, this still results in an unnecessary memory buildup.

To avoid letting these refs lie around unnecessarily, we need to send down a cancel op even when ops are not started.

The specific case where we ran into this issue was with hijacking interceptors on direct channels.

### DIFF
--- a/src/core/lib/surface/call.cc
+++ b/src/core/lib/surface/call.cc
@@ -367,7 +367,6 @@ class FilterStackCall final : public Call {
   bool received_initial_metadata_ = false;
   bool receiving_message_ = false;
   bool requested_final_op_ = false;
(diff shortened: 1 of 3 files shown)


ctiller (approved):

yashykt:
Thanks for reviewing!

PR #30013, opened 2022-06-15
EventEngine::RunAt(): Weighted Target LB Policy removal timer

drfloob:
There may be a pre-existing race here if the timer callback is fired while the DelayedRemovalTimer is being orphaned. A timer mutex would solve it, but I'm not sure it's necessary.

Potential sequence of events:
T1: Orphan: and `timer_handle_.has_value() = true`, if block entered.
T2: OnTimerLocked: `timer_handle.reset()` succeeds
T1: Orphan continued: attempts to call `GetDefaultEventEngine()->Cancel(*timer_handle_);`


src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
```diff
@@ -444,44 +445,32 @@ void WeightedTargetLb::UpdateStateLocked() {
 WeightedTargetLb::WeightedChild::DelayedRemovalTimer::DelayedRemovalTimer(
     RefCountedPtr<WeightedTargetLb::WeightedChild> weighted_child)
     : weighted_child_(std::move(weighted_child)) {
-  GRPC_CLOSURE_INIT(&on_timer_, OnTimer, this, nullptr);
-  Ref().release();
-  grpc_timer_init(&timer_, ExecCtx::Get()->Now() + kChildRetentionInterval,
-                  &on_timer_);
+  timer_handle_ = GetDefaultEventEngine()->RunAt(
+      absl::Now() + kChildRetentionInterval, [self = Ref()] {
```
markdroth, line 449: Doesn't this lambda need to be `mutable` in order to do the `std::move()` into the nested lambda below?

markdroth (commented):
I don't follow your comment about the pre-existing race.  In what sequence of events would this race occur?

src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
drfloob: Yes, you're right. I've replaced it with `self = Ref()`. Even though the function will only be called once, making the arg mutable feels like the wrong choice.

drfloob:
I've updated the description to explain the race. Thanks for the review.

src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
drfloob: Done. As discussed, the fact that [this behavior](https://godbolt.org/z/5sn1dzhcE) sometimes surprises folks is why I prefer to not use mutable, but I'm fine to use here it and move on given that it's not a problem in this case.

markdroth (approved):

PR #30034, opened 2022-06-16
doc: interop-tests descrption about orca_oob test server lock

YifeiZhuang:


doc/interop-test-descriptions.md
```diff
@@ -1143,10 +1147,15 @@ Procedures:
...
     }
     ```
-The call carries a reference to receive the load report, e.g. using CallOptions.
-The reference will be passed to the custom LB policy as part of the `OrcaOobReportListener` API.
-* Client asserts that, after 1.5 second, the latest OOB load report received is equal to the test load report.
-* Client sends another unary call to the server. The call request sets `orca_oob_report` to a 
+The test then blocks until the first response is received. This means the server lock is acquired.
+It is important that if the stream is completed or terminated with error at this time, 
+the test aborts and fails.
+* Client samples the latest OOB load report by doing empty unary calls. The call
```
markdroth, line 1153: Instead of having to make additional unary calls to do this, how about simply having the LB policy invoke an injected callback that the client provides at the time that it registers the LB policy?  This is essentially what we do for C++ in our end-to-end tests:

https://github.com/grpc/grpc/blob/d9f64437b09d97bd9aebc6bfa46d5a61e7477c41/test/cpp/end2end/client_lb_end2end_test.cc#L2579

This way, there's no polling needed; instead, we get a callback when each OOB backend metric report is received.

doc/interop-test-descriptions.md
```diff
@@ -1159,8 +1168,12 @@ different test load report.
...
     }
     ```
-The call still carries a reference to receive the load report.
-* Client asserts that, after 1.5 second, the latest OOB load report received is equal to the new test load report.
+The call then blocks until the second response is received. If the stream is completed or terminated 
+with error at this time, the test aborts and fails.
+* Similarly, client then keeps trying to check the latest OOB load report received by using 
+an empty unary call, waiting 1 second between each retry, until it is equal to the test load report.
+If retry exceeds 5 times limit the test fails.
+* Client asserts the streaming call is successful.
```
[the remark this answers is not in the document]
YifeiZhuang: Client completes the call first

doc/interop-test-descriptions.md
YifeiZhuang: Yea no polling has its advantage. It avoids sleeps between each polling, and processes immediately after callback is invoked. 
Part of the reason java does polling is because it saves some complexity of extra injection, both implementation and description here. In java can re-use what we already have (using callOptions)for per-query test so it would be easier to reason and consistent.
In my opinion either way would work, what about I make the doc less specific and allow each language to implement their verification.  The timeout should be the same tho.

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
+Server reports backend metrics data in per-query case, echoing metrics data from the unary call request.
 
 Using ORCA APIs we install the per-query metrics reporting server interceptor, so that it can attach 
 metrics per RPC.
-Also using ORCA APIs we register the `OpenRCAService` implementation to the server, so that it can
-report metrics periodically. The minimum report interval in the ORCA service is set to 1 sec.
 
-During test, the server will receive unary requests from the client that each may contain up to two
-test load report, indicating whether it needs to update metrics for the current call or at the OOB server.
+During `orca_per_query` test, the server will receive unary requests from the client that may contain a
```
ejona86, line 1414: Don't specify which test. The tests get to choose which features they use. Just say "If `SimpleRequest.orca_per_rpc_report` is set, the server will add the metric data from `orca_per_rpc_report` to the RPC using the language's CallMetricRecorder." You can still say something about "Both utilization and request cost metrics should be copied."

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
-metrics, the metrics data is coped from the received test load report from the request.
+
+#### Report out-of-band metrics
+Server reports backend metrics data in out-of-band case, echoing metrics data from full duplex
+call request.
+
+Using ORCA APIs we register the `OpenRCAService` implementation to the server, so that it can
+report metrics periodically. The minimum report interval in the ORCA service is set to 1 sec.
+
+The server implementation should use a lock or similar mechanism to allow only one client to perform the `orca_oob` test at a time
```
ejona86, line 1429: s/perform the `orca_oob` test/control the server's out-of-band reports/

Remove the orca_oob reference below as well.

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
+in order to not interfere with other clients' tests.
+
+During `orca_oob` test, the server receives full duplex requests from the client that contains a
```
ejona86, line 1432: "test load report" isn't obviously orca_oob_report. Combining the two sentences seems to help:
> When `StreamingOutputCallRequest.orca_oob_report` is set a the request of FullDuplexCall, ...

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
+test load report, indicating whether it needs to update metrics at the OOB server.
+When the `orca_oob_report` is set in the request, the server test driver will block until
+it acquires the oob lock. If the stream wins the lock, the server calls `MetricRecorder` to first clear
```
ejona86, line 1435: "When the stream acquires". There's no "if"; it is only a question of "when." "If" should make you wonder, "what if this doesn't happen?" A stream "winning" the lock makes it sound like winning a race or the like, and definitely sounds like it can lose.

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
+
+The server implementation should use a lock or similar mechanism to allow only one client to perform the `orca_oob` test at a time
+in order to not interfere with other clients' tests.
+
+During `orca_oob` test, the server receives full duplex requests from the client that contains a
+test load report, indicating whether it needs to update metrics at the OOB server.
+When the `orca_oob_report` is set in the request, the server test driver will block until
+it acquires the oob lock. If the stream wins the lock, the server calls `MetricRecorder` to first clear 
+all the previous metrics data, and then record the utilization 
+metrics, the metrics data is coped from the received test load report from the request. This stream
```
ejona86, line 1437: s/utilization metrics, .../ utilization metrics from `orca_oob_report`./

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
+keeps holding the lock for all the subsequent messages, and it always echoes the test report received from the
```
ejona86, line 1438: I don't know what this "always echos" is talking about.

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
+
+During `orca_oob` test, the server receives full duplex requests from the client that contains a
+test load report, indicating whether it needs to update metrics at the OOB server.
+When the `orca_oob_report` is set in the request, the server test driver will block until
+it acquires the oob lock. If the stream wins the lock, the server calls `MetricRecorder` to first clear 
+all the previous metrics data, and then record the utilization 
+metrics, the metrics data is coped from the received test load report from the request. This stream
+keeps holding the lock for all the subsequent messages, and it always echoes the test report received from the 
+request.
+It is important that whenever the stream is closed, either successfully completed or terminated with error,
```
ejona86, line 1440: s/It is important that whenever/Whenever/

All of what is defined here is important. If any one thing is messed up in this section, the test will be inaccurate.

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1404,38 @@ fully communicate metadata.
...
+
+Using ORCA APIs we register the `OpenRCAService` implementation to the server, so that it can
+report metrics periodically. The minimum report interval in the ORCA service is set to 1 sec.
+
+The server implementation should use a lock or similar mechanism to allow only one client to perform the `orca_oob` test at a time
+in order to not interfere with other clients' tests.
+
+During `orca_oob` test, the server receives full duplex requests from the client that contains a
+test load report, indicating whether it needs to update metrics at the OOB server.
+When the `orca_oob_report` is set in the request, the server test driver will block until
```
ejona86, line 1434: s/request/request and the stream has not previously acquired the lock/

doc/interop-test-descriptions.md
```diff
@@ -1125,13 +1125,17 @@ registered custom LB policy. A load report reference can be passed from the call
 through, e.g. CallOptions, to receive metric reports.
 The test framework will fill in the reference with the latest load report from the report listener.
 This way, together with server behaviors we can verify the expected metric reports are received.
+Note that this test client includes multiple verification steps that requires the server to keep the states.
```
ejona86, line 1128: This is all part of "Backend Metrics Report," so doesn't need to be said.

doc/interop-test-descriptions.md
```diff
@@ -1125,13 +1125,17 @@ registered custom LB policy. A load report reference can be passed from the call
...
 * [UnaryCall][]
+* [FullDuplexCall][]
 * [Backend Metrics Report][]
 
 Procedures:
-* Client sends a unary call to the server. The call request sets `orca_oob_report` to a test load report.
+1. Client starts a bidi-streaming call to the server. In the first message the request sets `orca_oob_report` to a test load report.
     ```
     {
       orca_oob_report:{
```
ejona86, line 1141: It needs to request a response. Ditto below.

doc/interop-test-descriptions.md
```diff
@@ -1143,10 +1147,14 @@ Procedures:
...
     }
     ```
-The call carries a reference to receive the load report, e.g. using CallOptions.
-The reference will be passed to the custom LB policy as part of the `OrcaOobReportListener` API.
-* Client asserts that, after 1.5 second, the latest OOB load report received is equal to the test load report.
-* Client sends another unary call to the server. The call request sets `orca_oob_report` to a 
+The test then blocks until the first response is received. This means the server lock is acquired.
+It is important that if the stream is completed or terminated with error at this time, 
+the test aborts and fails. 
+2. Client verifies (timeout=5s) that the latest OOB load report received is equal to the test load report.
```
ejona86, line 1153: View the file in github (three dots at the top-right, you get a drop-down, then "View file"). The formatting is pretty broken here. I think you need to indent some of the previous lines, but I'm not certain.

doc/interop-test-descriptions.md
```diff
@@ -1143,10 +1147,14 @@ Procedures:
       }
     }
     ```
-The call carries a reference to receive the load report, e.g. using CallOptions.
-The reference will be passed to the custom LB policy as part of the `OrcaOobReportListener` API.
-* Client asserts that, after 1.5 second, the latest OOB load report received is equal to the test load report.
-* Client sends another unary call to the server. The call request sets `orca_oob_report` to a 
+The test then blocks until the first response is received. This means the server lock is acquired.
```
ejona86, line 1150: Take a look at ping_pong for how simply we try to express the test process. Terseness is helpful because it makes it much easier to verify a test implementation is correct.

This can just be a step (2) "Client waits for a reply." Stuff like "this means" should be removed. The "it is important" was important to the Java implementation, and all implementations should consider stuff like that, but isn't a good fit for discussing here. If anything, just say "Client asserts it receives a reply."

Yes, we were careful on the implementation side of things, but that was really about me hating bad error codes. How that is done in each language and what turns out to be important in each language will vary.

doc/interop-test-descriptions.md
```diff
@@ -1143,10 +1147,14 @@ Procedures:
...
+It is important that if the stream is completed or terminated with error at this time, 
+the test aborts and fails. 
+2. Client verifies (timeout=5s) that the latest OOB load report received is equal to the test load report.
+To do verification, client may inject a callback to the custom LB policy, or poll the result by
+doing empty unary calls that carries a reference, e.g. using CallOptions, that will be filled in
+by the custom LB policy as part of the `OrcaOobReportListener` API.
+3. Client then sends another request on the stream to the server. The request sets `orca_oob_report` to a
```
ejona86, line 1157: Delete "The request sets `orca_oob_report` to a different test load report." We're not explaining why the test is written the way it is. We're saying what the test has to do. "You do things things, and you have verified the behavior is correct."

doc/interop-test-descriptions.md
```diff
@@ -1159,8 +1167,11 @@ different test load report.
       }
     }
     ```
-The call still carries a reference to receive the load report.
-* Client asserts that, after 1.5 second, the latest OOB load report received is equal to the new test load report.
+The call then blocks until the second response is received. If the stream is completed or terminated 
+with error at this time, the test aborts and fails.
+4. Similar to step 2, client then verifies (timeout=5s) that the latest OOB load report received is
+equal to the new test load report.
+5. Client completes the call, and asserts the streaming call is successful.
```
ejona86, line 1174: s/completes the call/half closes the stream/

doc/interop-test-descriptions.md
ejona86, line 1153: I think this will normally fail, because "latest" might be out-of-date. I see it as more "Client waits up to 5 seconds to receive a OOB load report that matches the requested load report in step 1. To wait for load report..."

doc/interop-test-descriptions.md
ejona86: This was marked resolved, but it doesn't seem resolved. This comment is talking about https://github.com/grpc/grpc-java/blob/69abec5a2881afe5525f45ffdbd8f4fb23cc53ff/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java#L1798

doc/interop-test-descriptions.md
```diff
@@ -1391,22 +1402,20 @@ fully communicate metadata.
...
+
+If `SimpleRequest.orca_per_rpc_report` is set in unary call, the server will add
+the metric data from `orca_per_rpc_report` to the RPC using the language's 
+CallMetricRecorder.
+
+If `SimpleRequest.orca_oob_report` is set in fullDuplexCall call, the server 
+will first clear all the previous metrics data, and then add utilization metrics
+from `orca_oob_report` to the `OpenRCAService`.
+The server implementation should use a lock or similar mechanism to allow only
+one client to control the server's out-of-band reports.
```
ejona86, line 1421: s/reports/reports until the end of the RPC/

doc/interop-test-descriptions.md
ejona86: No, not in text. It is part of the message; you can include it in the message snippet. And it is also necessary in step 3, unless you remove the "After getting a response" in step 4.

ejona86 (approved):

markdroth (approved):

PR #30041, opened 2022-06-17
EventEngine::RunAfter: Rls LB Policy

drfloob:


src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1116,40 +1116,45 @@ LoadBalancingPolicy::PickResult RlsLb::Picker::Pick(PickArgs args) {
...
                     self->entry_->lb_policy_.get(), self->entry_.get(),
                     self->entry_->is_shutdown_
                         ? "(shut down)"
-                        : self->entry_->lru_iterator_->ToString().c_str(),
-                    self->armed_);
+                        : self->entry_->lru_iterator_->ToString().c_str());
           }
-          bool cancelled = !self->armed_;
-          self->armed_ = false;
-          if (cancelled) return;
```
yijiem, line 1152: Do we need to honor the early return here if we have already cancelled the timer (through `Orphan()`)? We could check `if (self->backoff_timer_task_handle_.has_value())` provided if it hasn't been reset above.

drfloob: Good question. Previously `OnBackoffTimer` would be called whether the timer was cancelled or not, hence the need to check the condition inside the callback. Now, `OnBackoffTimer` is only called if the timer has _not_ been cancelled. The logic for what to do on cancellation was moved to line 1131. So I think it's correct as is.

drfloob: I see what you're saying now, the `cancelled` value in the callback is conflated with the concept of being orphaned. That callback had different behavior when orphaned, regardless of the timer cancellation status. Fixed.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1419,35 +1425,43 @@ void RlsLb::Cache::ResetAllBackoff() {
...
+  }
+  lb_policy_->work_serializer()->Run(
+      [this]() {
+        RefCountedPtr<RlsLb> lb_policy(lb_policy_);
         if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_rls_trace)) {
-          gpr_log(GPR_INFO, "[rlslb %p] cache cleanup timer fired (%s)",
-                  cache->lb_policy_, StatusToString(error).c_str());
+          gpr_log(GPR_INFO, "[rlslb %p] cache cleanup timer fired", lb_policy_);
         }
-        if (error == absl::CancelledError()) return;
```
yijiem, line 1434: Is there a similar issue here, i.e. the logic of the timer being cancelled is conflated with shutdown?

drfloob: Indeed, fixed. Great catch!

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1419,35 +1425,41 @@ void RlsLb::Cache::ResetAllBackoff() {
...
   map_.clear();
   lru_list_.clear();
-  grpc_timer_cancel(&cleanup_timer_);
+  if (cleanup_timer_handle_.has_value() &&
+      lb_policy_->channel_control_helper()->GetEventEngine()->Cancel(
+          *cleanup_timer_handle_)) {
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_rls_trace)) {
+      gpr_log(GPR_INFO, "[rlslb %p] cache cleanup timer canceled", lb_policy_);
+    }
+    cleanup_timer_handle_.reset();
```
yijiem, line 1434: Maybe reset the handle regardless of whether `EventEngine::Cancel` returns true or false? I think technically we want to reset it when `EventEngine::Cancel` returns false.

yijiem (approved):

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
drfloob: Done. I'm not sure that's necessary, but it doesn't hurt. Shutdown only appears to be called after RlsLb's `ShutdownLocked`, which means the callback will return early on `if (lb_policy->is_shutdown_) return;` regardless. The handle-reset in Cache::Shutdown is probably unnecessary, but safe, and conveys intent.

yijiem: oh, sorry then I missed that! Thanks for the explanation!

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1116,40 +1116,47 @@ LoadBalancingPolicy::PickResult RlsLb::Picker::Pick(PickArgs args) {
 RlsLb::Cache::Entry::BackoffTimer::BackoffTimer(RefCountedPtr<Entry> entry,
                                                 Timestamp backoff_time)
     : entry_(std::move(entry)) {
-  GRPC_CLOSURE_INIT(&backoff_timer_callback_, OnBackoffTimer, this, nullptr);
-  Ref(DEBUG_LOCATION, "BackoffTimer").release();
-  grpc_timer_init(&backoff_timer_, backoff_time, &backoff_timer_callback_);
+  backoff_timer_task_handle_ =
+      entry_->lb_policy_->channel_control_helper()->GetEventEngine()->RunAfter(
+          backoff_time - Timestamp::Now(),
```
markdroth, line 1121: Need to instantiate an `ExecCtx` and `ApplicationCallbackExecCtx` here.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1419,35 +1425,40 @@ void RlsLb::Cache::ResetAllBackoff() {
...
   map_.clear();
   lru_list_.clear();
-  grpc_timer_cancel(&cleanup_timer_);
+  if (cleanup_timer_handle_.has_value() &&
+      lb_policy_->channel_control_helper()->GetEventEngine()->Cancel(
+          *cleanup_timer_handle_)) {
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_rls_trace)) {
+      gpr_log(GPR_INFO, "[rlslb %p] cache cleanup timer canceled", lb_policy_);
+    }
+    lb_policy_->Unref(DEBUG_LOCATION, "Shutdown");
```
markdroth, line 1434: Let's have the lambda capture the ref, like we do everywhere else, so that we don't need to manually unref in this case.

I think if we do that, then we don't need to care whether or not the cancellation succeeds.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1419,35 +1425,40 @@ void RlsLb::Cache::ResetAllBackoff() {
...
+  cleanup_timer_handle_.reset();
 }
 
-void RlsLb::Cache::OnCleanupTimer(void* arg, grpc_error_handle error) {
-  Cache* cache = static_cast<Cache*>(arg);
-  cache->lb_policy_->work_serializer()->Run(
-      [cache, error]() {
-        RefCountedPtr<RlsLb> lb_policy(cache->lb_policy_);
+void RlsLb::Cache::OnCleanupTimer() {
+  lb_policy_->work_serializer()->Run(
```
markdroth, line 1440: As above, please move the enterance into the `WorkSerializer` to be inline in the lambda (line 1374), so that we can just move the captured ref into it instead of doing this manual ref-tracking.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1361,11 +1368,10 @@ RlsLb::Cache::Entry::OnRlsResponseLocked(
 //
 
 RlsLb::Cache::Cache(RlsLb* lb_policy) : lb_policy_(lb_policy) {
-  Timestamp now = Timestamp::Now();
   lb_policy_->Ref(DEBUG_LOCATION, "CacheCleanupTimer").release();
-  GRPC_CLOSURE_INIT(&timer_callback_, OnCleanupTimer, this, nullptr);
-  grpc_timer_init(&cleanup_timer_, now + kCacheCleanupTimerInterval,
-                  &timer_callback_);
+  cleanup_timer_handle_ =
```
markdroth, line 1372: Suggest moving this code into a separate private method called `StartCleanupTimer()`, so that we can just call it from both here and from the end of `OnCleanupTimer()` (line 1460).

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1361,11 +1368,10 @@ RlsLb::Cache::Entry::OnRlsResponseLocked(
...
+      lb_policy->channel_control_helper()->GetEventEngine()->RunAfter(
+          kCacheCleanupTimerInterval, [this] { OnCleanupTimer(); });
```
markdroth, line 1374: As above, need to declare `ExecCtx` and `ApplicationCallbackExecCtx` here.

markdroth (commented):
This looks good overall, but there are a couple of issues.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1103,40 +1106,41 @@ LoadBalancingPolicy::PickResult RlsLb::Picker::Pick(PickArgs args) {
...
+      backoff_time - ExecCtx::Get()->Now(),
+      [self = Ref(DEBUG_LOCATION, "BackoffTimer")] { self->OnBackoffTimer(); });
 }
 
 void RlsLb::Cache::Entry::BackoffTimer::Orphan() {
-  if (armed_) {
-    armed_ = false;
-    grpc_timer_cancel(&backoff_timer_);
+  if (backoff_timer_task_handle_.has_value()) {
+    if (GetDefaultEventEngine()->Cancel(*backoff_timer_task_handle_)) {
```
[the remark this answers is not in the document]
markdroth: Sorry, nevermind this comment -- I think this was a saved draft from a much earlier draft of this PR that I had never sent and didn't realize was still there.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
drfloob: This one is a behavioral change, but I think that's probably an improvement. The ref will now be taken earlier when scheduling a timer, rather than inside the work serializer callback. And if we don't explicitly release the ref before restarting the next timer, a ref is now held continuously between timer call execution and scheduling of the next timer, when previously the ref count would have dropped. I imagine if this is a significant change, we would have seen flakes in the previous implementation when trying to ref an orphaned lb_policy from inside the work serializer.

markdroth: I would prefer to use the same pattern we're using everywhere else of capturing the ref in the callback lambda, since that allows the compiler to enfore ref ownership.  I understand that it means that we'll take a new ref and then drop the previous one every time the timer fires, but this isn't performance-sensitive code, so that seems like a good trade-off.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
drfloob: Good catch, thanks.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -1103,40 +1106,41 @@ LoadBalancingPolicy::PickResult RlsLb::Picker::Pick(PickArgs args) {
...
 }
 
-void RlsLb::Cache::Entry::BackoffTimer::OnBackoffTimer(
-    void* arg, grpc_error_handle /*error*/) {
-  auto* self = static_cast<BackoffTimer*>(arg);
-  self->entry_->lb_policy_->work_serializer()->Run(
-      [self]() {
-        RefCountedPtr<BackoffTimer> backoff_timer(self);
+void RlsLb::Cache::Entry::BackoffTimer::OnBackoffTimer() {
+  entry_->lb_policy_->work_serializer()->Run(
```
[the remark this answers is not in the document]
drfloob: Fixed.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
drfloob: Done

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
drfloob: Done

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
drfloob: Done

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -539,8 +542,8 @@ class RlsLb : public LoadBalancingPolicy {
     std::list<RequestKey> lru_list_ ABSL_GUARDED_BY(&RlsLb::mu_);
     std::unordered_map<RequestKey, OrphanablePtr<Entry>, absl::Hash<RequestKey>>
         map_ ABSL_GUARDED_BY(&RlsLb::mu_);
-    grpc_timer cleanup_timer_;
-    grpc_closure timer_callback_;
+    absl::optional<EventEngine::TaskHandle> cleanup_timer_handle_
+        ABSL_GUARDED_BY(&RlsLb::mu_);
```
markdroth, line 546: I don't think this needs to be guarded by the mutex.  It will be accessed only within the `WorkSerializer`.

(It does currently happen to also always be accessed while holding the mutex, but I think that's essentially a coincidence -- it's an implementation detail that could change over time.)

I think we can just remove this lock annotation.

markdroth (approved):
This looks really good!  Just a couple of minor comments, then feel free to merge.

I looked carefully at all of the call trees that access the task handles to make sure they're being synchronized properly, since the compiler can't always do that for us due to the `ABSL_NO_THREAD_SAFETY_ANALYSIS` annotation we need to use on various `Orphan()` methods.  Everything looks good to me.

Thanks!

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
drfloob: Good point, done.

src/core/ext/filters/client_channel/lb_policy/rls/rls.cc
```diff
@@ -531,6 +531,9 @@ class RlsLb : public LoadBalancingPolicy {
     void MaybeShrinkSize(size_t bytes)
         ABSL_EXCLUSIVE_LOCKS_REQUIRED(&RlsLb::mu_);
 
+    // Shared logic for starting the cleanup timer
+    void StartCleanupTimer() ABSL_EXCLUSIVE_LOCKS_REQUIRED(&RlsLb::mu_);
```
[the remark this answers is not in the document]
drfloob: Done

drfloob:
Thanks for the thorough review, Mark!

drfloob:
iOS built tool failure is unrelated.

PR #30043, opened 2022-06-17
EventEngine::RunAfter: GrpcLb

drfloob:


src/core/lib/backoff/backoff.h
```diff
@@ -39,6 +39,9 @@ class BackOff {
   /// Returns the time at which the next attempt should start.
   Timestamp NextAttemptTime();
 
+  /// Returns the amount of time to wait before the next attempt should start.
+  Duration NextAttemptWaitTime();
```
markdroth, line 43: I assume this is leftover from an earlier draft of this PR.  Please remove.

src/core/lib/load_balancing/lb_policy.cc
```diff
@@ -72,15 +72,16 @@ LoadBalancingPolicy::PickResult LoadBalancingPolicy::QueuePicker::Pick(
   //    ExitIdleLocked().
   if (!exit_idle_called_ && parent_ != nullptr) {
     exit_idle_called_ = true;
-    auto* parent = parent_->Ref().release();  // ref held by lambda.
+    auto* parent = parent_->Ref(DEBUG_LOCATION, "QueuePicker")
```
markdroth, line 75: Suggest moving this change to a separate PR.

Note that we can probably change the `ExecCtx::Run()` below to instead use `EventEngine::Run()`, and then we can capture the ref in the lambda, just like we're doing everywhere else.

markdroth (commented):
This looks good overall!  Just a few minor comments.

src/core/lib/load_balancing/lb_policy.cc
drfloob: Removed. It was helpful in debugging, I thought it would be harmless to leave in, but I'll file a separate PR.

drfloob: https://github.com/grpc/grpc/pull/31836

markdroth (approved):

PR #30045, opened 2022-06-17
EventEngine::RunAfter: Priority LB

drfloob:


yijiem (approved):
Looks good! Thanks AJ! I think I found an issue in my PR for PollingResolver by looking at this change. Will fix it later.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -522,46 +516,48 @@ PriorityLb::ChildPriority::DeactivationTimer::DeactivationTimer(
...
       gpr_log(GPR_INFO, "[priority_lb %p] child %s (%p): reactivating",
               child_priority_->priority_policy_.get(),
               child_priority_->name_.c_str(), child_priority_.get());
     }
-    timer_pending_ = false;
-    grpc_timer_cancel(&timer_);
+    if (!child_priority_->priority_policy_->channel_control_helper()
+             ->GetEventEngine()
+             ->Cancel(*timer_handle_)) {
+      timer_handle_.reset();
```
markdroth, line 543: I think we want to do this regardless of whether the cancellation succeeds, because even if cancellation fails and the callback has not run yet, we don't want the callback to delete the child.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -580,53 +576,53 @@ PriorityLb::ChildPriority::FailoverTimer::FailoverTimer(
...
               "[priority_lb %p] child %s (%p): cancelling failover timer",
               child_priority_->priority_policy_.get(),
               child_priority_->name_.c_str(), child_priority_.get());
     }
-    timer_pending_ = false;
-    grpc_timer_cancel(&timer_);
+    if (!child_priority_->priority_policy_->channel_control_helper()
+             ->GetEventEngine()
+             ->Cancel(*timer_handle_)) {
+      timer_handle_.reset();
```
markdroth, line 605: Same here: We need to reset the timer handle even if the cancellation fails, because if the timer has fired but the callback has not yet run, we don't want the callback to report TRANSIENT_FAILURE.

markdroth (commented):
I think there are a couple of bugs in this one.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
drfloob: Good catch. I think could still be an ordering issue here if the callback fires after cancellation fails, but before the timer handle is reset. I've added a fix here and below, which ensures that the handle is reset before the cancellation attempt is made.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -537,17 +537,19 @@ void PriorityLb::ChildPriority::DeactivationTimer::Orphan() {
               child_priority_->priority_policy_.get(),
               child_priority_->name_.c_str(), child_priority_.get());
     }
-    if (!child_priority_->priority_policy_->channel_control_helper()
-             ->GetEventEngine()
-             ->Cancel(*timer_handle_)) {
-      timer_handle_.reset();
-    }
+    auto tmp_handle = *timer_handle_;
```
markdroth, line 540: I don't think we need this temp variable.  This code runs in the `WorkSerializer`, as does the callback, so it's impossible for the callback to run between the call to `Cancel()` and resetting `timer_handle_`.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
```diff
@@ -599,25 +600,25 @@ void PriorityLb::ChildPriority::FailoverTimer::Orphan() {
...
```
markdroth, line 603: Same here.

src/core/ext/filters/client_channel/lb_policy/priority/priority.cc
drfloob: Good point, fixed.

markdroth (approved):
Looks great!

PR #30058, opened 2022-06-18
Update authz tests to use callback

ashithasantosh:
Replaces sleep times in test with callbacks. The implementation adds a callback to the FileWatcherAuthorizationPolicyProvider, which is executed only when the file contents change. Now, the test can wait for the callback to be executed, instead of waiting for X seconds for reload to complete. 

This helps with flaky tests in MacOS, which occurs due to IO delays (grpc_load_file takes approx 9 seconds to complete). Due to IO delays, reload doesn't complete and the test uses the previous authorization policy to make decisions, thereby accepting the RPC request instead of denying it.


ashithasantosh:
The C++ end2end tests still use hard-coded wait times. 
https://github.com/grpc/grpc/blob/master/test/cpp/end2end/grpc_authz_end2end_test.cc#L797

Since the callback is used for testing purpose, I think we should avoid exposing it in C++ or C-core APIs. @markdroth What do you think?

ashithasantosh:
> The C++ end2end tests still use hard-coded wait times. https://github.com/grpc/grpc/blob/master/test/cpp/end2end/grpc_authz_end2end_test.cc#L797
> 
> Since the callback is used for testing purpose, I think we should avoid exposing it in C++ or C-core APIs. @markdroth What do you think?

Updated C++ end2end tests too to use callback instead of hard-coded wait times. The callback API is not exposed to C-core or C++ APIs.

src/core/lib/security/authorization/grpc_authorization_policy_provider.cc
```diff
@@ -117,6 +119,15 @@ FileWatcherAuthorizationPolicyProvider::FileWatcherAuthorizationPolicyProvider(
         return;
       }
       absl::Status status = provider->ForceUpdate();
+      {
+        MutexLock lock(&provider->mu_);
```
markdroth, line 123: We're already grabbing this lock inside of `ForceUpdate()`, so it seems inefficient to grab it again here.  Instead, I suggest moving this code inside of `ForceUpdate()` so that we're only grabbing the lock once.  See my comment below for suggested code.

test/core/security/grpc_authorization_policy_provider_test.cc
```diff
@@ -164,6 +186,24 @@ TEST(AuthorizationPolicyProviderTest, FileWatcherRecoversFromFailure) {
   auto provider = FileWatcherAuthorizationPolicyProvider::Create(
       tmp_authz_policy->name(), /*refresh_interval_sec=*/1);
   ASSERT_TRUE(provider.ok());
+  absl::Notification on_first_reload_done;
+  absl::Notification on_second_reload_done;
+  bool first_reload = true;
+  std::function<void(grpc_status_code code, const char* error_details)>
+      callback = [&on_first_reload_done, &on_second_reload_done, &first_reload](
```
markdroth, line 193: As mentioned elsewhere, instead of having one callback deal with all of this state, let's just reset the callback after the first reload is done.

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -15,6 +15,8 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "absl/synchronization/notification.h"
```
markdroth, line 18: As mentioned elsewhere, please use `gpr_event` instead of `absl::Notification`.

test/core/end2end/tests/grpc_authz.cc
```diff
@@ -546,6 +540,16 @@ static void test_file_watcher_valid_policy_reload(
...
+  absl::Notification on_reload_done;
+  std::function<void(grpc_status_code code, const char* error_details)>
+      callback = [&on_reload_done](grpc_status_code status,
+                                   const char* error_details) {
+        GPR_ASSERT(GRPC_STATUS_OK == status);
+        GPR_ASSERT(nullptr == error_details);
+        on_reload_done.Notify();
+      };
+  dynamic_cast<grpc_core::FileWatcherAuthorizationPolicyProvider*>(provider)
+      ->SetCallbackForTesting(std::move(callback));
```
markdroth, line 552: To be safe, I think we should reset the callback to null at the end of each test, to make sure that the lambda does not get called again after we return from this function, because the state captured by the lambda will no longer be valid then.

Same thing in all tests.

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -752,7 +779,27 @@ TEST_F(GrpcAuthzEnd2EndTest, FileWatcherRecoversFromFailure) {
       "  ]"
       "}";
   grpc_core::testing::TmpFile tmp_policy(policy);
-  InitServer(CreateFileWatcherAuthzPolicyProvider(tmp_policy.name(), 1));
+  auto provider = CreateFileWatcherAuthzPolicyProvider(tmp_policy.name(), 1);
+  absl::Notification on_first_reload_done;
+  absl::Notification on_second_reload_done;
+  bool first_reload = true;
+  std::function<void(grpc_status_code code, const char* error_details)>
+      callback = [&on_first_reload_done, &on_second_reload_done, &first_reload](
```
markdroth, line 787: As mentioned elsewhere, please reset the callback after the first reload, so that we don't need all of this state machinery.

src/core/lib/security/authorization/grpc_authorization_policy_provider.h
```diff
@@ -83,6 +85,9 @@ class FileWatcherAuthorizationPolicyProvider
                                          unsigned int refresh_interval_sec,
                                          absl::Status* status);
 
+  void SetCallbackForTesting(
+      std::function<void(grpc_status_code code, const char* error_details)> cb);
```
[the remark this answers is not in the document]
ashithasantosh: Done.

src/core/lib/security/authorization/grpc_authorization_policy_provider.cc
```diff
@@ -105,6 +105,8 @@ FileWatcherAuthorizationPolicyProvider::FileWatcherAuthorizationPolicyProvider(
   if (!status->ok()) {
     return;
   }
+  // Callback is not executed on initial file read.
+  execute_cb_ = false;
```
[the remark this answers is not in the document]
ashithasantosh: During first read, when file_contents_ variable gets the contents in file path, execute_cb_ is set to true, and callback gets executed on first read. The previous intent was callback should only get executed when contents in the file path change.
I updated the callback based on your recommendation and removed execute_cb_ member variable.

src/core/lib/security/authorization/grpc_authorization_policy_provider.h
```diff
@@ -101,7 +106,12 @@ class FileWatcherAuthorizationPolicyProvider
   std::unique_ptr<Thread> refresh_thread_;
   gpr_event shutdown_event_;
 
+  bool execute_cb_ = false;
```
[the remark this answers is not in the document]
ashithasantosh: Since this variable is only used by the refresh thread, I didn't use a mutex here.

src/core/lib/security/authorization/grpc_authorization_policy_provider.cc
```diff
@@ -140,6 +159,7 @@ absl::Status FileWatcherAuthorizationPolicyProvider::ForceUpdate() {
     return absl::OkStatus();
   }
   file_contents_ = std::move(*file_contents);
+  execute_cb_ = true;
```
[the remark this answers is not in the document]
ashithasantosh: Yes, I agree we should keep the callback general. Thank you for the recommendation :). Updated the code.

test/core/end2end/tests/grpc_authz.cc
```diff
@@ -15,6 +15,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#include "absl/synchronization/notification.h"
```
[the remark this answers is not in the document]
ashithasantosh: Done.

src/core/lib/security/authorization/grpc_authorization_policy_provider.cc
ashithasantosh: Done.

test/core/end2end/tests/grpc_authz.cc
```diff
@@ -661,6 +675,25 @@ static void test_file_watcher_recovers_from_failure(
           tmp_policy.name().c_str(), /*refresh_interval_sec=*/1, &code,
           &error_details);
   GPR_ASSERT(GRPC_STATUS_OK == code);
+  absl::Notification on_first_reload_done;
+  absl::Notification on_second_reload_done;
+  bool first_reload = true;
+  std::function<void(grpc_status_code code, const char* error_details)>
+      callback = [&on_first_reload_done, &on_second_reload_done, &first_reload](
```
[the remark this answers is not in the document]
ashithasantosh: Done.

test/core/security/grpc_authorization_policy_provider_test.cc
```diff
@@ -19,6 +19,8 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "absl/synchronization/notification.h"
```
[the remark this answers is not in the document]
ashithasantosh: Done.

test/core/security/grpc_authorization_policy_provider_test.cc
ashithasantosh: Done.

test/cpp/end2end/grpc_authz_end2end_test.cc
ashithasantosh: Done.

test/cpp/end2end/grpc_authz_end2end_test.cc
ashithasantosh: Done.

test/core/end2end/tests/grpc_authz.cc
```diff
@@ -641,11 +633,28 @@ static void test_file_watcher_invalid_policy_skip_reload(
...
   grpc_authorization_policy_provider_release(provider);
   test_allow_authorized_request(f);
+  gpr_event on_reload_done;
+  gpr_event_init(&on_reload_done);
+  std::function<void(bool contents_changed, absl::Status status)> callback =
+      [&on_reload_done](bool contents_changed, absl::Status status) {
+        if (contents_changed) {
+          GPR_ASSERT(absl::StatusCode::kInvalidArgument == status.code());
+          GPR_ASSERT(
+              status.message().compare("\"name\" field is not present.") == 0);
```
markdroth, line 643: Instead of `string_view.compare(expected) == 0`, please just say `string_view == expected`.

Same thing throughout.

markdroth (approved):
This looks great!  Just one minor cosmetic comment remaining; feel free to merge after fixing.

ashithasantosh:
Thank you for the review @markdroth :)

PR #30061, opened 2022-06-19
Revert "Revert "Tcp connect cancellation implementation for posix""

Vignesh2208:
Reverts grpc/grpc#30037. Fix ios breakages by adding grpc_tcp_client_global_init() to iomgr_posix_cfstream.cc

### DIFF
--- a/src/core/lib/iomgr/iomgr_posix.cc
+++ b/src/core/lib/iomgr/iomgr_posix.cc
@@ -80,6 +80,7 @@ void grpc_set_default_iomgr_platform() {
   grpc_set_pollset_vtable(&grpc_posix_pollset_vtable);
   grpc_set_pollset_set_vtable(&grpc_posix_pollset_set_vtable);
   grpc_core::SetDNSResolver(grpc_core::NativeDNSResolver::GetOrCreate());
+  grpc_tcp_client_global_init();
   grpc_set_iomgr_platform_vtable(&vtable);
 }
 
--- a/src/core/lib/iomgr/iomgr_posix_cfstream.cc
+++ b/src/core/lib/iomgr/iomgr_posix_cfstream.cc
@@ -177,6 +177,7 @@ void grpc_set_default_iomgr_platform() {
     grpc_set_pollset_set_vtable(&grpc_apple_pollset_set_vtable);
     grpc_set_iomgr_platform_vtable(&apple_vtable);
   }
+  grpc_tcp_client_global_init();
   grpc_set_timer_impl(&grpc_generic_timer_vtable);
   grpc_core::SetDNSResolver(grpc_core::NativeDNSResolver::GetOrCreate());
 }
--- a/src/core/lib/iomgr/tcp_client_posix.cc
+++ b/src/core/lib/iomgr/tcp_client_posix.cc
@@ -27,6 +27,7 @@
 #include <string.h>
 #include <unistd.h>
 
+#include "absl/container/flat_hash_map.h"
 #include "absl/strings/str_cat.h"
 
 #include <grpc/support/alloc.h>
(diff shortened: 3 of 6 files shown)


src/core/lib/iomgr/iomgr_posix.cc
```diff
@@ -80,6 +80,7 @@ void grpc_set_default_iomgr_platform() {
   grpc_set_pollset_vtable(&grpc_posix_pollset_vtable);
   grpc_set_pollset_set_vtable(&grpc_posix_pollset_set_vtable);
   grpc_core::SetDNSResolver(grpc_core::NativeDNSResolver::GetOrCreate());
+  grpc_tcp_client_global_init();
```
drfloob, line 83: There's a separate spot for initialization, see comments below.

src/core/lib/iomgr/iomgr_posix_cfstream.cc
```diff
@@ -177,6 +177,7 @@ void grpc_set_default_iomgr_platform() {
     grpc_set_pollset_set_vtable(&grpc_apple_pollset_set_vtable);
     grpc_set_iomgr_platform_vtable(&apple_vtable);
   }
+  grpc_tcp_client_global_init();
```
drfloob, line 180: I don't think this is needed if you take my other suggestion.

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -62,8 +63,33 @@ struct async_connect {
...
+};
+
+namespace {
+
+gpr_once g_tcp_client_posix_init = GPR_ONCE_INIT;
+std::vector<ConnectionShard>* g_connection_shards = nullptr;
+std::atomic<int64_t> g_connection_id{1};
+
+void do_tcp_client_global_init(void) {
+  size_t num_shards = std::max(2 * gpr_cpu_num_cores(), 1u);
```
drfloob, line 83: I think this can be done in an implementation of `grpc_iomgr_platform_vtable.init`, or maybe `grpc_tcp_posix_init`. A new global init is redundant.

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -208,6 +241,14 @@ static void on_writable(void* acp, grpc_error_handle error) {
   }
 
 finish:
+  if (!connect_cancelled) {
+    int shard_number = ac->connection_handle % (*g_connection_shards).size();
+    struct ConnectionShard* shard = &(*g_connection_shards)[shard_number];
+    {
+      grpc_core::MutexLock lock(&shard->mu);
```
drfloob, line 248: I'm curious about the performance impact of taking another lock on every connection and every write. Have you investigated this?

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: iomgr.init implementations get invoked with every grpc_init invoked after a successful grpc_shutdown. I dont think this function implementation can be moved to a iomgr.init implementations because this needs to be executed exactly once per process. Also the global variables referenced here are static and thus cannot be referenced from another file. So I think some grpc_tcp_client_global_init function is needed

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: Yeah good point. This lock is not acquired on every write though. The on_writable callback gets executed only once per connection after the connect() succeeds. So the lock will be acquired once before initiating the connection attempt and once after the connection attempt succeeds/fails aynchronously. I dont anticipate contention because we already have a global lock in grpc_fd_create() implementations and haven't observed any major complaints

Vignesh2208: Maybe we can add a flag to control the number of minimum number of shards ?

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: I could call grpc_tcp_client_global_init() from tcp_posix_init(). but I think I still need to add grpc_tcp_client_global_init in iomgr_posix_cfstream.cc. Does that work ?

markdroth: I thought about commenting about this in my original review, but I decided not to bother, because the existing iomgr code already uses a lot of the `grpc_init()` infrastructure, and all of this code will go away when iomgr does.  But since AJ brought it up, here's what I was going to suggest:

```
std::vector<ConnectionShard>* GetConnectionShards() {
  static auto* shards = new std::vector<ConnectionShard>(std::max(2 * gpr_cpu_num_cores(), 1u));
  return shards;
}

std::atomic<int64_t>* GetConnectionId() {
  static auto* id = new std::atomic<int64_t>{1};
  return id;
}
```

Then change all code that accesses `g_connection_shards` and `g_connection_id` to instead use `GetConnectionShards()` and `GetConnectionId()`, respectively.

This avoids all need for explicit initialization.  However, note that when you have a function that returns a static variable like this, the compiler implicitly adds a lock around the variable's initialization, and I'm not sure if that would have any performance impact.

I think we should just do whatever's most expedient here.  We are trying to make this code go away, after all.

markdroth (approved):

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: Thanks for pointing this out. Since we avoid the global lock around the static variable initialization with the current approach, I am leaning towards keeping it. W.r.t performance impact, I ran the connection scaling benchmarks (maintained by @ananda1066) which establish many channels and send RPCs over them. I did not observe any regressions there.

So I'll go ahead and merge this PR for now.

PR #30065, opened 2022-06-20
Suport OS400 system

zheddie:
This pull request is to make grpcio support on IBM i operating system. thanks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,8 @@ if 'linux' in sys.platform:
     CARES_INCLUDE += (os.path.join('third_party', 'cares', 'config_linux'),)
 if 'openbsd' in sys.platform:
     CARES_INCLUDE += (os.path.join('third_party', 'cares', 'config_openbsd'),)
+if 'os400' in sys.platform:
+    CARES_INCLUDE += (os.path.join('third_party', 'cares', 'config_os400'),)
 RE2_INCLUDE = (os.path.join('third_party', 're2'),)
 SSL_INCLUDE = (os.path.join('third_party', 'boringssl-with-bazel', 'src',
                             'include'),)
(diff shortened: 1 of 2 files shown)


drfloob:
The code you modified is in c-ares, a third-party library that gRPC consumes. We don't maintain it. You'll want to make these changes upstream; the code you changed will be overwritten when we update the c-ares dependency.

drfloob:
Apologies, my mistake, gRPC does maintain this bit of c-ares code. Let me take another look here.

drfloob:
I'm surprised to hear someone wants to run gRPC on it, I'm curious how you're planning to use it. I understand if you can't share details.

Regardless, for unsupported platforms like OS400, we don't have the testing infrastructure or human resources to vet or maintain the code for it. Functionality may be broken at any time and we wouldn't notice, and similarly if the platform code harms some other officially supported platforms we'll likely remove it if a maintainer can't be contacted. If you'd like to support OS400, we'd ask you to add personal contact information for a maintainer, such as seen [here](https://github.com/grpc/grpc/blob/92f60f9201641c11d13caeaede6dcde069ae6c8e/include/grpc/impl/codegen/port_platform.h#L439-L440).

As far as I can tell, this is missing some essential configuration in `third_party/cares/BUILD` and `cares.BUILD`. How did you test this?

zheddie:
Thanks for your effort on this PR. You may already noticed that we would try our best to use platform specific macros or indicators to make the code changes affecting our systems only.   The goal of our changes for os400 is to make our customers easy to setup/install ML frameworks , such as ML Server. 

I understand that you do not have any resources for os400, so, we would make sure the changes are accurate for our system. You can help to review to make sure it does not break any other platforms. 

Yes, if there's any code changes need in cares, we would use separate PR to make them works on os400.  By far, the changes in this PR could make the cares works. Of course, we may have future PRs to fix other functional issues.

zheddie:
personal information were added with one new commit. thanks.

zheddie:
@drfloob more comments on it? thanks.

zheddie:
@drfloob need your review . thanks.

drfloob:
> As for the BUILD and cares.BUILD, I guess they are used for your test purpose build.

bazel is gRPC's main supported build system, both of these files are essential for the bazel build. Are you trying to add python-only support for OS400? And what is your build command/process that turns these changes into the binary you ultimately want built?

zheddie:
Sorry for late response , looks like I missed the note of the update.
Yes, currently we only focus on the python support of grpc. the command we used to do the build is `python setup.py install`.

drfloob:
If new platform support is going to be added, it'll have to be for all languages. Otherwise we'll be tracking a more complicated patchwork of support, and we want to be in that situation. See this doc for how platforms are tracked: https://github.com/grpc/grpc/tree/master/src/cpp. The assumption here is that wrapped languages (including Python) build upon a C-library foundation of support, listed in that doc.

PR #30077, opened 2022-06-21
TSI: return handshaker error message for inclusion in RPC failure status

markdroth:
Another step toward #22883.


apolcyn:
Looks like there's an issue with the PR as is? TSAN is reporting a data race in `//test/core/tsi/alts/handshaker:alts_concurrent_connectivity_test`

src/core/tsi/transport_security_interface.h
```diff
@@ -482,11 +484,14 @@ typedef void (*tsi_handshaker_on_next_done_cb)(
...
-    size_t* bytes_to_send_size, tsi_handshaker_result** handshaker_result,
-    tsi_handshaker_on_next_done_cb cb, void* user_data);
+tsi_result tsi_handshaker_next(tsi_handshaker* self,
+                               const unsigned char* received_bytes,
+                               size_t received_bytes_size,
+                               const unsigned char** bytes_to_send,
+                               size_t* bytes_to_send_size,
+                               tsi_handshaker_result** handshaker_result,
+                               tsi_handshaker_on_next_done_cb cb,
+                               void* user_data, std::string* error = nullptr);
```
ZhenLian, line 494: Maybe add a new line of comment explaining the meaning of `error`?

src/core/lib/security/transport/security_handshaker.cc
```diff
@@ -132,6 +132,7 @@ class SecurityHandshaker : public Handshaker {
   RefCountedPtr<grpc_auth_context> auth_context_;
   tsi_handshaker_result* handshaker_result_ = nullptr;
   size_t max_frame_size_ = 0;
+  std::string tsi_handshake_error_;
```
ZhenLian, line 135: Can I know what will happen if TLS resumption happens? Will it re-use the the same `SecurityHandshaker` object, or create a new one? 
I don't think it will cause any problems here, as I think `SecurityHandshaker` will always be alive throughout the lifetime of  `tsi_handshaker_next`. Just asking a question off the top of my head while I was reading through the code :)

ZhenLian (approved):
Thanks for the change! It looks great!

src/core/tsi/transport_security_interface.h
markdroth: Done.

src/core/lib/security/transport/security_handshaker.cc
markdroth: SSL session caching and resumption is handled in the SSL TSI implementation.  It's not visible to the code here at all.

markdroth (commented):
@apolcyn The TSAN bug should be fixed now.

markdroth:
@apolcyn Still waiting for your review on this.

@ZhenLian I've filed #30228 to track a follow-up item, which is writing better tests that we generate the right RPC failure status messages for TSI failures.

src/core/tsi/alts/handshaker/alts_handshaker_client.cc
```diff
@@ -204,25 +207,34 @@ void alts_handshaker_client_handle_response(alts_handshaker_client* c,
   if (handshaker == nullptr) {
     gpr_log(GPR_ERROR,
             "handshaker is nullptr in alts_tsi_handshaker_handle_response()");
-    handle_response_done(client, TSI_INTERNAL_ERROR, nullptr, 0, nullptr);
+    handle_response_done(
```
apolcyn, line 210: nit: doesn't need to happen in this PR this since this PR is still an improvement, but we could probably further improve these error messages by including context from the `on_status_received` callback (which shows the ALTS RPC status code and details).

apolcyn (approved):

src/core/tsi/alts/handshaker/alts_handshaker_client.cc
markdroth: I wanted to do something like that, but the current code does not appear to wait for the status to be received before it returns the handshaker result, so it would have been a more significant change to do that.  I think you have an existing TODO in here about using the ALTS RPC status, so we can deal with this as part of that.

PR #30079, opened 2022-06-22
xDS: fix endpoint weight defaults

markdroth:
For some inexplicable reason, we were defaulting to an endpoint weight of 500 instead of 1.  See https://github.com/envoyproxy/envoy/pull/21813 for context.

Also, this removes some unnecessary pickiness about endpoint weights in the ring_hash code.  We were previously ignoring any endpoint with weight 0, which should never happen; now instead, we ignore the weight but still use the endpoint.

### DIFF
--- a/test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc
@@ -368,6 +368,52 @@ TEST_P(RingHashTest, NoHashPolicy) {
               ::testing::DoubleNear(kDistribution50Percent, kErrorTolerance));
 }
 
+// Tests that we observe endpoint weights.
+TEST_P(RingHashTest, EndpointWeights) {
+  CreateAndStartBackends(3);
+  const double kDistribution50Percent = 0.5;
+  const double kDistribution25Percent = 0.25;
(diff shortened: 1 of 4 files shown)


markdroth:
Hold off on reviewing -- I want to add a test here.

yashykt (approved):

test/cpp/end2end/xds/xds_ring_hash_end2end_test.cc
```diff
@@ -368,6 +368,52 @@ TEST_P(RingHashTest, NoHashPolicy) {
...
+  EdsResourceArgs args(
+      {{"locality0",
+        {CreateEndpoint(0, ::envoy::config::endpoint::v3::HealthStatus::UNKNOWN,
+                        0),
+         CreateEndpoint(1, ::envoy::config::endpoint::v3::HealthStatus::UNKNOWN,
+                        1),
+         CreateEndpoint(2, ::envoy::config::endpoint::v3::HealthStatus::UNKNOWN,
+                        2)}}});
+  balancer_->ads_service()->SetEdsResource(BuildEdsResource(args));
+  // TODO(donnadionne): remove extended timeout after ring creation
```
yashykt, line 398: was this copied from somewhere?

PR #30091, opened 2022-06-23
XdsSecurityTest flakiness: Add broken pipe to expected error messages

yashykt:
Fixes b/237007226
Such a status was seen on https://source.cloud.google.com/results/invocations/b34339ca-b15a-4c53-b38a-2124bff54e1c/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_end2end_test@poller%3Dpoll/log for example

### DIFF
--- a/test/cpp/end2end/xds/xds_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_end2end_test.cc
@@ -384,7 +384,7 @@ class XdsSecurityTest : public XdsEnd2endTest {
                     "(UNKNOWN: Failed to connect to remote host: Connection "
                     "(refused|reset by peer)|UNAVAILABLE: Failed to connect "
                     "to remote host: FD shutdown|UNKNOWN: Tls handshake failed|"
-                    "UNAVAILABLE: Socket closed)"));
+                    "UNAVAILABLE: Socket closed|UNAVAILABLE: Broken pipe)"));
           }
         });
         Status status = SendRpc();


markdroth:
Why are we seeing "broken pipe"?  Shouldn't that occur only if the server is not doing a graceful shutdown, where the RPC is still in flight when it shuts down?

yashykt:
> Why are we seeing "broken pipe"? Shouldn't that occur only if the server is not doing a graceful shutdown, where the RPC is still in flight when it shuts down?

I believe so, and given that we don't have graceful shutdown, that seems to be expected.

markdroth:
Can we change the server to do a graceful shutdown in this test, using the `StopListeningAndSendGoaways()` method I added to the xDS test framework in #29786?

yashykt:
I can probably change the framework to alternate between two backends, so that we expect the channel to move to a different backend on an update. That should remove the broken pipe and socket closed errors. We would still have `connection refused`\`failed to connect` type errors though while the new backend is coming up.

markdroth (approved):
I'm okay with this PR as-is, since it will fix the immediate problem.  But in the long term, I would really like to see us fix this test framework to ensure that we don't fail RPCs in situations where they really shouldn't be failing.

markdroth:
What I'm saying is that the "broken pipe" situation should never actually occur in practice due to graceful shutdown.

PR #30093, opened 2022-06-23
Fix Race Condition During grpcio-tests Install on CI

gnossen:
Last Thursday around noon, we started seeing the following error across all platforms at a high rate when invoking `python setup.py test_lite` from the `grpcio_tests` directory:

```
error: [WinError 183] Cannot create a file when that file already exists: 'T:\\altsrc\\github\\grpc\\workspace_python_windows_opt_native\\src\\python\\grpcio_tests\\.eggs\\protobuf-3.20.1-py3.8-win-amd64.egg'
```

I was able to track the problem down to [this line](https://github.com/grpc/grpc/blob/17811f67de52c90718176747c8a011ab3dee236b/src/python/grpcio_tests/commands.py#L136). For some reason, when installing this package (which is internal only), we actually install dependencies when commands besides `python setup.py install` are chosen. This was the case here. We were previously explicitly leaving out the `protobuf` dependency of `grpcio-tools` [here](https://github.com/grpc/grpc/blob/17811f67de52c90718176747c8a011ab3dee236b/tools/run_tests/helper_scripts/build_python.sh#L167) so that dependencies weren't installed until `python setup.py test_lite` was called.

Unforunately, this breakage was _not_ hermetic. That is, it doesn't seem to be associated with any particular commit in source control. Further, I was unable to correlate the onset of this breakage with any of the pypi packages in our transitive closure. I also was unable to track down any changes to Kokoro config that could have contributed.

The issue occurs on 3.7, 3.8, and 3.9 but seemingly not on 3.6. I would not be surprised if this is a bug in the `setuptools` package (though updating that dependency did not help). It has received less attention lately due to [PEP 517](https://peps.python.org/pep-0517/). We probably need to switch to a PEP 517 compliant build system soon to make sure that we don't fall prey to any further bit rot.

### DIFF
--- a/src/python/grpcio_tests/commands.py
+++ b/src/python/grpcio_tests/commands.py
@@ -102,7 +102,7 @@ class TestLite(setuptools.Command):
 
     def _add_eggs_to_path(self):
         """Fetch install and test requirements"""
-        # self.distribution.fetch_build_eggs(self.distribution.install_requires)
+        self.distribution.fetch_build_eggs(self.distribution.install_requires)
 
 
 class TestPy3Only(setuptools.Command):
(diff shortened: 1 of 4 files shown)


sergiitk (approved):
18 commits to make a 10-line change — classic example why dev's productivity can't be measured in the lines of code 😄

PR #30095, opened 2022-06-23
[gRPC/iOS] Remove and cleanup CoreTest iOS test

dennycd:
CoreTests no longer valid as we will be refactoring iOS/MacOS event engine impl out of grpc c-core.  Current CoreTests also does not contain any actual test case (mainly build test) so the removal does not reduce any existing test coverage for us. 

----
cc @HannahShiSFB 

https://github.com/grpc/grpc-ios/issues/89


sampajano (approved):
Great!😃

PR #30096, opened 2022-06-24
[gRPC/iOS] Remove libuv pod dependnecy from iOS cocoapod build

dennycd:
Libuv is no longer used in iOS build for event engine impl. Removing it from cocoapod build targets. 

---
cc @drfloob  @HannahShiSFB 

https://github.com/grpc/grpc-ios/issues/106


sampajano (approved):
Very nice? 😃

PR #30097, opened 2022-06-24
Set uses polling to true for oracle_event_engine_posix_test

Vignesh2208:
This is required because the oracle event engine uses poll syscall. If uses_polling is set to False, there will be an attempt to run this test with GRPC_POLL_STRATEGY=none which will hijack the poll c-wrapper causing the test to fail.

### DIFF
--- a/build_autogenerated.yaml
+++ b/build_autogenerated.yaml
@@ -7113,7 +7113,6 @@ targets:
   - linux
   - posix
   - mac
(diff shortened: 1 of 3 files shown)


ctiller:
Let's add a TODO to figure out some more reasonable design here.

ctiller (approved):

PR #30104, opened 2022-06-24
BinderTransport: Support Intent URI

sifmelcara:
This commit adds an new overload of `CreateBinderChannel`, that accepts
    an intent URI (which should be able to be parsed by
    https://developer.android.com/reference/android/content/Intent#parseUri(java.lang.String,%20int) )
    for specifying the component to connect to.

    Later we will deprecate the old APIs that accepts `package_name`,
    `class_name`, and `action_name` separately. Intent URI seems to be a
    better and more flexible way to specify a component to connect to.

    `grpc.binder.custom_android_intent_action_name` channel arg is deprecated now.
    User should use URI to specify custom action instead.

### DIFF
--- a/src/core/ext/transport/binder/client/channel_create.cc
+++ b/src/core/ext/transport/binder/client/channel_create.cc
@@ -35,6 +35,7 @@
 #include <grpc/support/port_platform.h>
 
 #include "absl/memory/memory.h"
+#include "absl/strings/substitute.h"
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
 
@@ -59,10 +60,6 @@ namespace {
 // grpc.io.action.BIND is the standard action name for binding to binder
 // transport server.
 const char* kStandardActionName = "grpc.io.action.BIND";
-
-// grpc::ChannelArguments key for specifying custom action name.
-const char* kCustomIntentActionNameArgKey =
-    "grpc.binder.custom_android_intent_action_name";
 }  // namespace
 
 namespace grpc {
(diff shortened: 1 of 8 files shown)


src/core/ext/transport/binder/client/channel_create.cc
```diff
@@ -154,6 +154,65 @@ std::shared_ptr<grpc::Channel> CreateCustomBinderChannel(
...
+
+  auto channel = CreateChannelInternal(
+      "", grpc::internal::CreateClientBinderChannelImpl(new_args),
+      std::vector<
+          std::unique_ptr<experimental::ClientInterceptorFactoryInterface>>());
+
+  grpc_channel_args_destroy(new_args);
+
+  return channel;
+}
```
sifmelcara, line 214: Mostly just the copy of the implementation of the old version of API (which is also in this file). We probably will deprecate the other version of API later so we simply make a copy here instead of refactor them to make them share code.

src/core/ext/transport/binder/java/io/grpc/binder/cpp/NativeConnectionHelper.java
```diff
@@ -30,12 +30,19 @@ final class NativeConnectionHelper {
   // Maps connection id to GrpcBinderConnection instances
   static Map<String, GrpcBinderConnection> s = new HashMap<>();
```
LittleCVR, line 31: Rename `s` to a more meaningful name.

LittleCVR (approved):
nits

src/core/ext/transport/binder/java/io/grpc/binder/cpp/NativeConnectionHelper.java
sifmelcara: Done

PR #30072, opened 2022-06-21
Update win RBE docker image to VS2019

jtattermusch:
- update the RBE docker image (to contain VS2019 instead of VS2015)
- regenerate RBE toolchain configuration (to use VS2019)

Basically followed steps from go/rbe-windows-user-guide

I also tried to upgrade the docker image to use VS2017 (= the version we're currently using for our cmake build on windows) but I wasn't able to make that work. IMHO the setup where we use VS2017 for our cmake build (which takes care of checking the compatibility with VS2017) and we use newer VS2019 for RBE builds actually makes sense.

PR #30108, opened 2022-06-24
Update rbe_windows2019

veblush:
Let's try

### DIFF
--- a/third_party/toolchains/rbe_win2019/Dockerfile
+++ b/third_party/toolchains/rbe_win2019/Dockerfile
@@ -1,94 +1,79 @@
-# This dockerfile is taken from go/rbe-windows-user-guide
-# (including the fix --compilation_mode=dbg)
-
-# This Dockerfile creates an image that:
-# - Has the correct MTU setting for networking from inside the container to work.
-# - Has Visual Studio 2015 Build Tools installed.
-# - Has msys2 + git, curl, zip, unzip installed.
-# - Has Python 2.7 installed.
-# TODO(jsharpe): Consider replacing "ADD $URI $DEST" with "Invoke-WebRequest -Method Get -Uri $URI -OutFile $DEST"
-# Use the latest Windows Server Core image.
-#
-# WARNING: What's the `:ltsc2019` about?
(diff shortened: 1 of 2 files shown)


third_party/toolchains/rbe_win2019/Dockerfile
```diff
@@ -1,94 +1,79 @@
...
-RUN Start-Process msiexec.exe -ArgumentList \"/i C:\\TEMP\\7z.msi /qn /norestart /log C:\\TEMP\\7z_install_log.txt\" -wait
-RUN $oldpath = (Get-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH).path; \
-  $newpath = \"$oldpath;C:\Program Files\7-Zip\"; \
-  Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH -Value $newPath
-# Extract msys2
-ADD http://repo.msys2.org/distrib/x86_64/msys2-base-x86_64-20181211.tar.xz C:\\TEMP\\msys2.tar.xz
-RUN 7z x C:\TEMP\msys2.tar.xz -oC:\TEMP\msys2.tar
-RUN 7z x C:\TEMP\msys2.tar -oC:\tools
-# Start building the actual image
+# This Dockerfile creates an image that has the following:
```
jtattermusch, line 1: According to my investigation, just upgrading the docker image won't be enough:
- the `https://github.com/grpc/grpc/tree/master/third_party/toolchains/bazel_0.26.0_rbe_windows` needs to be upgraded as well (as it contains the paths to the VC++ tools etc - so it basically has the VS version hardcoded in it).
- looks like since we're using bazel4.x currently, we'll need to use VS2019 (see comment in the dockerfile).

jtattermusch: I created https://github.com/grpc/grpc/pull/30072, but still not quite there.

PR #30123, opened 2022-06-27
Using string rep for status-time

veblush:
Using a string representation of time when storing as a payload of status to avoid dumping its internal representation.


veblush:
To unblock the upgrade, I'm merging this first. Adding fuzzer tests sounds worth it so let me do it separately.

malkia:
This maybe expensive on Windows, as it ends up calling absl::LocalTimeZone() https://github.com/abseil/abseil-cpp/issues/1760#issuecomment-2495670125

Possibly some caching needs to be done, with a TTL?

PR #30124, opened 2022-06-27
xds-k8s: Remove skips.version_lt(), and use only skips.version_gte()

sergiitk:
A minor refactoring.

cc @temawi


gnossen (approved):

sergiitk:
I spot-tested this locally, have enough confidence to not run full jobs for a change as trivial as this.

PR #30126, opened 2022-06-27
Update protobuf to 3.21.2

yashykt:
Automatic script to update protobuf did not work, so performed the update manually.

Changes -
1. Updated protobuf submodule to 3.21.2
2. Updated bazel dependencies
3. Updated check_submodules.sh
4. Modified protobuf.patch to use "4.21.2" because of the python protobuf breaking change
5. Modified bazel files to use ":well_known_type_protos" instead of ":well_known_protos" since the latter has been deprecated and removed in 3.21.2.
6. Regenerated code
7. Regenerated UPB files


yashykt:
I suspect that the automatic script did not work due to the two breaking changes -
- Python proto breaking change leading to version change from 3.19.4 to 4.21.2
- Removal of ":well_known_types" target

yashykt:
Protobuf v3.21.2 seems broken for Python. The target `'@com_google_protobuf//:protobuf_python'` is not publicly visible.

jtattermusch:
> Automatic script to update protobuf did not work, so performed the update manually.

What didn't work? Let's fix the automation - we're going to need to fix it anyway so we might as well do it right now.

yashykt:
> What didn't work? Let's fix the automation - we're going to need to fix it anyway so we might as well do it right now.

The errors weren't verbose enough. I'm thinking of skipping this protobuf version upgrade for the release given the issues.

PR #30128, opened 2022-06-28
Add std::shared_ptr support to ChannelArgs, and precondition ChannelArgs with a default EventEngine

drfloob:
Any class that is derived from `std::enable_shared_from_this` can now be stored as a shared_ptr inside ChannelArgs.

This is a step towards using ChannelArgs as the primary means of
accessing EventEngine instances in gRPC-core. If not explicitly provided
by the application, a default EventEngine will populated into
ChannelArgs during preconditioning.


src/core/lib/channel/channel_args.h
```diff
@@ -86,6 +88,32 @@ struct ChannelArgTypeTraits<
   };
 };
 
+// Specialization for shared_ptr
```
ctiller, line 91: I'd still rather get us to a place that we don't need to write `GetObjectRef<std::shared_ptr<T>>`...

As I see it the big reason we can't is that we can't go from `GetObjectRef<EventEngine>` to the conclusion that we need to wrap in a std::shared_ptr... so maybe we just give ourselves that hint:

```
template <typename T> struct WrapInSharedPtr : std::false_type {};
template <> struct WrapInSharedPtr<EventEngine> : std::true_type {};
```

and I suspect the rest of this shakes out so that SetObject/GetObject/GetObjectRef do as we'd like them to do.

drfloob: This PR adds general shared_ptr support to ChannelArgs, not just for EventEngine. I'd like to land it generally if possible. We could special-case EventEngine here if we wanted to add the same special cases for any other type we want to add as a shared_ptr.

I think the bigger blocker was `GetObject<T>` where we expect to return a raw pointer `p` in most cases, but `p->get()` when a shared_ptr is stored in channel_args. I can take another look.

ctiller (approved):

PR #30003, opened 2022-06-13
tell MSVC to build grpc with utf-8

hongnod:
fix #29979

PR #30129, opened 2022-06-28
tell MSVC to build grpc with utf-8

hongnod:
backport #30003


yashykt:
The right way to do this would be to cherry-pick the merged commit from master into the 1.47.x branch

Issue #8451, opened 2016-10-21
[C#] Client channel socket leaks unless read stream drained explicitly


kkm000:
This is still happening in v1.0.0, and I suspect has always been there.

I am using bidirectional streaming RPCs to talk to server. The server sends an exactly known number of messages back. Therefore  I am not reading the response stream to EOF, but rather end the call as soon as I get the data on the client.

If the call is disposed by the client without first reading the server stream to EOF, channel will not close its socket on shutdown, and a warning message is reported: _W1020 18:40:55.407708 Grpc.Core.Channel Channel shutdown was called but there are still 1 active calls for that channel._

In my debugging case, the server returns exactly one message. I can trace the problem as far as this. The socket handle is locked via a call to `ChannelSafeHandle.DangerousAddRef()` without a matching release on call end. `AsyncCallBase.ReleaseResourcesIfPossible()` fails to call ReleaseResources() when called from `AsyncCall.HandleFinished()`, because the member flag `readingDone` is never set to `true` in this scenario.

I do not understand the logic of the code to tell if simply setting `readingDone = true` before [the call to `ReleaseResourcesIfPossible()` from `HandleFinished()`](https://github.com/grpc/grpc/blob/v1.0.0/src/csharp/Grpc.Core/Internal/AsyncCall.cs#L498) is all that is missing. Maybe also creating a `streamingReadTcs` is required to complete possible pending reads? Is the call actually cancelled (in the gRPC sense) when `Dispose()` is called?

I work around the bug by calling `MoveNext` on the calls' reader in a loop until it returns `false`, which, in my case, it immediately does. But this a less than fantastic workaround.

jtattermusch:
Do you have a code snippet that replicates this?

What platform are you on? Also, from your description it seems it's about a call on the client side, is that right?

Btw, for testing various orderings in which operation completions can arrive to a call, we have this test:
https://github.com/grpc/grpc/blob/9f4e828343a78a24988e638cf7ac865cf3947cbe/src/csharp/Grpc.Core.Tests/Internal/AsyncCallTest.cs  (it runs the call handling logic against a mocked-out C core API and verifies that the behavior is as expected).
Would you be able to express your scenario as a test case there?

jtattermusch:
To understand this correctly, you basically read X (a known number) of responses on the client side and then you stop reading because you know that's all the messages that will be sent from the server?

I think if you are expecting X responses, it is a good practice to make sure that's the end of the stream and that there hasn't been an error:
```
[...]

PR #30134, opened 2022-06-28
Bugfix/issue 8451

tonydnewell:
Fix for issue [8451](https://github.com/grpc/grpc/issues/8451) - Client channel socket leaks unless read stream drained explicitly
Sets readingDone to true when client cancels or disposes of the call.
Tests have been added to check that the resources are cleaned up correctly.
@jtattermusch

### DIFF
--- a/src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
+++ b/src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
@@ -180,6 +180,122 @@ namespace Grpc.Core.Tests
         }
 
         [Test]
+        public async Task ServerStreamingCall_DisposeCallBeforeAllRead()
+        {
+            // Test to reproduce https://github.com/grpc/grpc/issues/8451
+            // "Client channel socket leaks unless read stream drained explicitly"
+            // This version of the test disposes the call before the response stream is
+            // fully read.
+
+            helper.ServerStreamingHandler = new ServerStreamingServerMethod<string, string>(async (request, responseStream, context) => {
+                try
+                {
+                    // send responses to the client with a short delay between each response
+                    foreach (string resp in request.Split(new[] { ' ' }))
+                    {
+                        await responseStream.WriteAsync(resp);
+                        await Task.Delay(50);
+                    }
+                    context.ResponseTrailers.Add("xyz", "");
+                }
+                catch (RpcException ex)
+                {
+                    // server may get an exception when the client cancels
+                    Assert.AreEqual(StatusCode.Cancelled, ex.Status.StatusCode);
+                }
+            });
+
+            using (var call = Calls.AsyncServerStreamingCall(helper.CreateServerStreamingCall(), "A B C D E"))
+            {
+                Assert.AreEqual(1, channel.GetCallReferenceCount());
+
+                try
+                {
+                    while (await call.ResponseStream.MoveNext())
+                    {
+                        Assert.AreEqual("A", call.ResponseStream.Current);
+                        // To reproduce issue #8451 - leave the loop early and ignore other responses
+                        break;
+                    }
+                }
+                catch (RpcException ex)
+                {
+                    Assert.Fail("Unexpected exception thrown {}", ex);
+                }
+            }
+            
+            // short delay to allow callbacks to fire
+            await Task.Delay(500);
+
+            // Resources for the call should have been cleaned up and thus no reference
+            // to the call in the channel. This would fail with a count of 1 before the
+            // fix for #8451.
+            Assert.AreEqual(0, channel.GetCallReferenceCount());
+        }
+
+        [Test]
+        public async Task ServerStreamingCall_CancelCallBeforeAllRead()
+        {
+            // Test to reproduce https://github.com/grpc/grpc/issues/8451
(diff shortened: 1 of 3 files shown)


jtattermusch:
Looks like the may be relevant:
https://source.cloud.google.com/results/invocations/3370cd09-f4a7-4738-a0b4-c83d18a37246/targets/github%2Fgrpc%2Frun_tests%2Fcsharp_linux_opt_native%2Fcsharp.mono.Grpc.IntegrationTesting.InteropClientServerTest/tests
(looks like in this context, setting readingDone=true might be making the internal call logic think it's not a streaming call?)

https://source.cloud.google.com/results/invocations/7aaefdf9-aee3-4287-94d5-658de9ef4145/targets/github%2Fgrpc%2Frun_tests%2Fcsharp_macos_opt_native%2Fcsharp.coreclr.Math.Tests.MathClientServerTest/tests
(not sure what could be causing this - I'll need to look into it).

src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
```diff
@@ -179,6 +179,122 @@ public async Task ServerStreamingCall_CancelClientSideRead()
...
+                }
+            });
+
+            using (var call = Calls.AsyncServerStreamingCall(helper.CreateServerStreamingCall(), "A B C D E"))
+            {
+                Assert.AreEqual(1, channel.GetCallReferenceCount());
+
+                try
+                {
+                    while (await call.ResponseStream.MoveNext())
```
jtattermusch, line 214: nit: the while() { do_sth_once; break;) is equivalent to  just invoking call.ResponseStream.MoveNext().
I think just calling MoveNext once without a loop and adding an explanatory comment makes the test shorter and the test logic easier to read.
I don't feel strongly about this though so leaving the resolution up to you.

src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
```diff
@@ -179,6 +179,122 @@ public async Task ServerStreamingCall_CancelClientSideRead()
...
+                    {
+                        await responseStream.WriteAsync(resp);
+                        await Task.Delay(50);
+                    }
+                    context.ResponseTrailers.Add("xyz", "");
+                }
+                catch (RpcException ex)
+                {
+                    // server may get an exception when the client cancels
+                    Assert.AreEqual(StatusCode.Cancelled, ex.Status.StatusCode);
```
jtattermusch, line 204: Checking the status code in the server side handler isn't necessary (you can just eat the exception or simply let the handler throw, the grpc glue code will handle that. 
Also, the status  code it's really a thing on the server side (as the server sends status code to the client, not the other way round).
So let's remove the try catch block?

src/csharp/Grpc.Core.Tests/Internal/AsyncCallTest.cs
```diff
@@ -472,6 +472,113 @@ public void ServerStreaming_MoreResponses_Success()
...
+            fakeCall.ReceivedResponseHeadersCallback.OnReceivedResponseHeaders(true, new Metadata());
+            Assert.AreEqual(0, asyncCall.ResponseHeadersAsync.Result.Count);
+
+            var readTask1 = responseStream.MoveNext();
+            fakeCall.ReceivedMessageCallback.OnReceivedMessage(true, CreateResponsePayload());
+            Assert.IsTrue(readTask1.Result);
+            Assert.AreEqual("response1", responseStream.Current);
+
+            // Cancel the call
+            asyncCall.Cancel();
```
jtattermusch, line 528: I thought the idea here would be to test an "inflight" read
(MoveNext has been called, but the OnReceivedMessage callback hasn't been called yet when OnReceivedStatusOnClient is received)
Generally speaking the OnReceivedStatusOnClient callback can be delivered earlier than the last (or last few) OnReceivedMessage callbacks, since there can be a race between delivering the events on two different threads in the threadpool.

So I think you might wanna add one more test to test the "inflight" read scenario.

src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
```diff
@@ -179,6 +179,122 @@ public async Task ServerStreamingCall_CancelClientSideRead()
...
+            Assert.AreEqual(1, channel.GetCallReferenceCount());
+
+            try
+            {
+                while (await call.ResponseStream.MoveNext())
+                {
+                    Assert.AreEqual("A", call.ResponseStream.Current);
+                    // To reproduce issue #8451 - cancel the call and leave
+                    // the loop early and ignore other responses
+                    tokenSource.Cancel();
```
jtattermusch, line 277: I think you should also test the scenario when you cancel before the first read and see if the call gets cleanup up properly anyway (I think that may reveal some extra issues).

jtattermusch (commented):
Thanks for the PR, I added some comments.

tonydnewell:
I can reproduce some of the "Basic Tests C# Linux" test failures - but only on Linux and only intermittently (I don't have a mac).  I expect there is a race condition with the callbacks  so I need to re-examine this fix and come up with a more robust solution.

tonydnewell:
Closing as I have created a different solution

Issue #28153, opened 2021-11-19
Possible memory leak of grpc_ssl_credentials (grpc_call leak if response stream not consumed by client)


tomes92:
### What version of gRPC and what language are you using?

gRPC C# v2.42.0

### What operating system (Linux, Windows,...) and version?

Windows 10 Pro 19043 x64

### What did you do?

Disposal of a gRPC client , created with a root certificates string, doesn't seem to free this string in unmanaged memory. The code below repeatedly creates and disposes a gRPC client. No running gRPC service is required and a random root certificates string can be used to reproduce this memory leak on my system.

Proto file:

```proto
syntax = "proto3";

package communication;

service MessageService {
	rpc MessagePipeline(stream Message) returns (stream Message);
}

message Message {
}
```

C# code:

```csharp
using Communication;
using Grpc.Core;
using System;
using System.Linq;
using System.Threading;

namespace Client
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var client = new MessageClient())
            {
                client.Start();

                Console.ReadKey();
            }
        }
    }

    class MessageClient : IDisposable
    {
        const int reconnectDelay = 1000; //ms

        bool bDisposed;

        Thread connectThread;
        Channel channel;

        public void Dispose()
        {
            if (bDisposed)
                return;

            bDisposed = true;

            if (connectThread != null)
            {
                if (connectThread.ThreadState != ThreadState.Unstarted)
                    connectThread.Join();

                connectThread = null;
            }
        }

        public bool Start()
        {
            connectThread = new Thread(new ThreadStart(RunConnect));
            connectThread.Name = "Connect thread";
            connectThread.Start();

            return true;
        }

        #region Connection

        /*internal static string GetCertificatePem(X509Certificate2 cert)
        {
            var builder = new StringBuilder();
            builder.AppendLine("-----BEGIN CERTIFICATE-----");
            builder.AppendLine(Convert.ToBase64String(cert.RawData, Base64FormattingOptions.InsertLineBreaks));
            builder.AppendLine("-----END CERTIFICATE-----");

            return builder.ToString();
        }

        internal static string GetRootCertificates()
        {
            StringBuilder builder = new StringBuilder();

            using (var store = new X509Store(StoreName.Root))
            {
                store.Open(OpenFlags.ReadOnly);

                foreach (X509Certificate2 mCert in store.Certificates)
                {
                    builder.AppendLine(
[...]

PR #30141, opened 2022-06-29
Fix memory leak - streaming call, no server

tonydnewell:
Fix for https://github.com/grpc/grpc/issues/28153

When the server is unavailable and the client creates a server streaming call (e.g. a duplex call) the connection to the server can not be established. Internally the OnReceivedResponseHeaders() callback is called with success == false. This 'false' case was never handled so the call was left in a state where the resources where never released.

### DIFF
--- a/src/csharp/Grpc.Core.Tests/Internal/AsyncCallTest.cs
+++ b/src/csharp/Grpc.Core.Tests/Internal/AsyncCallTest.cs
@@ -494,6 +494,20 @@ namespace Grpc.Core.Internal.Tests
         }
 
         [Test]
+        public void ServerStreaming_ServerResponseFailureDoesntLeakResources()
+        {
+            // Test for https://github.com/grpc/grpc/issues/28153
+            // Memory leak if resources not released when server unavailable
+            asyncCall.StartServerStreamingCall("request1");
+
+            // error reading response headers - possibly because server unavailable
+            fakeCall.ReceivedResponseHeadersCallback.OnReceivedResponseHeaders(false, new Metadata());
+            fakeCall.ReceivedStatusOnClientCallback.OnReceivedStatusOnClient(true, new ClientSideStatus(new Status(StatusCode.Unavailable, ""), new Metadata()));
+
+            Assert.AreEqual(0, channel.GetCallReferenceCount());
+        }
+
+        [Test]
         public void DuplexStreaming_NoRequestNoResponse_Success1()
         {
             asyncCall.StartDuplexStreamingCall();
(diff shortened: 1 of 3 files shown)


src/csharp/Grpc.Core/Channel.cs
```diff
@@ -223,6 +223,17 @@ protected override async Task ShutdownAsyncCore()
...
 
             var activeCallCount = activeCallCounter.Count;
+
+            // Part of fix for https://github.com/grpc/grpc/issues/28153 memory leak
+            if (activeCallCount > 0)
+            {
+                // There may be a race condition when disposing of a call and then
+                // immediately closing the channel. This reduces the chance although
+                // not completely eliminating it. It gives callbacks a chance to happen.
+                await Task.Delay(200);
```
jtattermusch, line 233: this seems very arbitrary. We can't have a random delay in the shutdown procedure like this.
Is this `. only about avoiding the warning message to be printed (while the call is disposed properly) or 2. is that actually needed for the call to be disposed at all?
If it's 1., I think we can live with the message being printed if the channel is abandoned too early. If it's 2. that's a bigger problems since we can't have the call sometimes ending up disposed properly and sometimes not.

src/csharp/Grpc.Core.Tests/Internal/AsyncCallTest.cs
```diff
@@ -493,6 +493,20 @@ public void ServerStreaming_StartCallFailureDoesntLeakResources()
...
 
+        [Test]
+        public void ServerStreaming_ServerResponseFailureDoesntLeakResources()
+        {
+            // Test for https://github.com/grpc/grpc/issues/28153
+            // Memory leak if resources not released when server unavailable
+            asyncCall.StartServerStreamingCall("request1");
+
+            // error reading response headers - possibly because server unavailable
+            fakeCall.ReceivedResponseHeadersCallback.OnReceivedResponseHeaders(false, new Metadata());
```
jtattermusch, line 504: since OnReceivedResponseHeaders can be triggered anytime after the call has started, the user can be attempting to read a message while that happens (e.g just before or just after) - it would be good to test those scenarios and make sure the expected behavior doesn't change.

Basically the ordering of OnReceivedResponseHeader(), OnReceivedStatusOnClient() and MoveNext() happening is fully at random (the two events can be delivered in any order and MoveNext() is invoked by the user so we have no control over when exactly it happens).

src/csharp/Grpc.Core.Tests/Internal/AsyncCallTest.cs
```diff
@@ -493,6 +493,20 @@ public void ServerStreaming_StartCallFailureDoesntLeakResources()
             Assert.IsTrue(fakeCall.IsDisposed);
         }
 
+        [Test]
+        public void ServerStreaming_ServerResponseFailureDoesntLeakResources()
```
jtattermusch, line 497: nit: rename to 
`ServerStreaming_OnReceivedResponseHeadersFailureDoesntLeakResources()` to be more specific?

PR #30154, opened 2022-06-30
Remove ExecCtx requirements from the EventEngine conformance test suite

drfloob:


Vignesh2208:
Is this not needed anymore because of the timer fork ?

Vignesh2208 (approved):

drfloob:
Partially. It started as an artifact of the libuv engine. When I created `IomgrEventEngine`, we decided all callbacks would need to instantiate their own ExecCtx (and not rely on the engine having one). Then removing the reliance on `grpc_timer_*` removed the rest of the need. The timer fork uses its own thread pool and calls the callbacks directly.

PR #30172, opened 2022-07-01
client channel: add tests for authority override precedence

markdroth:


test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -602,6 +605,78 @@ TEST_F(ClientLbEnd2endTest, ChannelIdleness) {
   EXPECT_EQ(channel->GetState(false), GRPC_CHANNEL_READY);
 }
 
+TEST_F(ClientLbEnd2endTest, AuthorityOverrideOnChannel) {
```
ctiller, line 608: It's probably useful to have some end2end tests around this, but it strikes me that we could easily add some unit tests here and improve code health:

suppose we introduce in client_channel.cc:
```
ChannelArgs MakeSubchannelArgs(
    const ChannelArgs& channel_args,
    const ChannelArgs& address_args,
    RefCountedPtr<SubchannelPoolInterface> subchannel_pool) {
    return
        args.UnionWith(address.args())
            .SetObject(chand_->subchannel_pool_)
            // If we haven't already set the default authority arg, add it from
            // the channel.
            // If we haven't already set the default authority arg (i.e., it
            // was not explicitly set by the application nor overridden by
            // the resolver), add it from the channel's default.
            .SetIfUnset(GRPC_ARG_DEFAULT_AUTHORITY, chand_->default_authority_)
            // Remove channel args that should not affect subchannel uniqueness.
            .Remove(GRPC_ARG_HEALTH_CHECK_SERVICE_NAME)
            .Remove(GRPC_ARG_INHIBIT_HEALTH_CHECKING)
            .Remove(GRPC_ARG_CHANNELZ_CHANNEL_NODE);
}
```

then we could have some really direct tests in client_channel_test.cc that verify the properties we want:

```
EXPECT_EQ(MakeSubchannelArgs(ChannelArgs().Set(DEFAULT_AUTHORITY, "foo"),
                             ChannelArgs().Set(DEFAULT_AUTHORITY, "bar"),
                             nullptr),
          ChannelArgs().Set(DEFAULT_AUTHORITY, "foo"));
```

and likely a dozen other tests that verify things we really care about.

ctiller (changes requested):
This seems like a really good opportunity to add some unit tests for properties we care about.

test/cpp/end2end/client_lb_end2end_test.cc
markdroth: Done.

ctiller (approved):

PR #30174, opened 2022-07-01
ObjC: Optional support to skip methods with types without prefixes.

thomasvl:
The existing optional support for objc_class_prefix only checks the file
defining the service, but the types for the messages can come from other files.
This allows those methods to be skipped if the other files didn't have a
objc_class_prefix set.

@donnadionne
@veblush

### DIFF
--- a/src/compiler/objective_c_generator.cc
+++ b/src/compiler/objective_c_generator.cc
@@ -156,6 +156,8 @@ inline map< ::std::string, ::std::string> GetMethodVars(
 }
 
 void PrintMethodDeclarations(Printer* printer, const MethodDescriptor* method) {
+  if (!ShouldIncludeMethod(method)) return;
+
   map< ::std::string, ::std::string> vars = GetMethodVars(method);
 
   PrintProtoRpcDeclarationAsPragma(printer, method, vars);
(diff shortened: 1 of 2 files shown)


src/compiler/objective_c_generator.cc
```diff
@@ -38,6 +38,15 @@ using ::std::set;
 namespace grpc_objective_c_generator {
 namespace {
 
+inline bool ShouldIncludeMethod(const MethodDescriptor* method) {
```
dennycd, line 41: NIT: consider moving this inline util to [objective_c_generator_helpers.h](https://github.com/grpc/grpc/blob/master/src/compiler/objective_c_generator_helpers.h)

dennycd (approved):
rest LGTM, thanks for patching this

src/compiler/objective_c_generator.cc
thomasvl: Done

thomasvl:
The failures don't seem related to the actual change.

dennycd:
iOS ObjC build break unrelated and investigating via  (https://github.com/grpc/grpc-ios/issues/111).

PR #30186, opened 2022-07-03
Update helloworld.proto

preranam05:


markdroth:
I'm not sure what this PR is for, but we don't need a new RPC method added to this service.

PR #30215, opened 2022-07-06
gtestify test/core/resource_quota

veblush:


ctiller (approved):
Approved with hesitation: It's unclear to me that this change is making anything better.

veblush:
> Approved with hesitation: It's unclear to me that this change is making anything better.

I agree that this particular test doesn't benefit but I'd like to do for consistency.

PR #30231, opened 2022-07-07
[gRPC/iOS] Test fix preventing dangling callback after test case done

dennycd:
Prevents response handler callback after test case exit which raises global exception that would terminate our test runs. 
 - Properly deallocate RMTService object in each test teardown 
 - Checking service obj state w/ weak ptr to ensure we don't continue execution if the object is already gone. 

### Test 

Run existing InteropTest suite . 

------


sampajano (approved):
Thanks :)

PR #30236, opened 2022-07-07
Include an invalid config to custom LB test

temawi:
This will verify that the configuration for a missing LB gets skipped
in favor of the valid one.


temawi:
@sergiitk

PR #30238, opened 2022-07-07
xds interop: Afinity tests supported in C++ after 1.48

yashykt:
https://github.com/grpc/grpc/pull/29959 fixed affinity tests for Core and v1.48 picks it up


sergiitk (approved):
Do we need to do affinity feature to https://github.com/grpc/grpc/blob/master/doc/grpc_xds_features.md?

markdroth (changes requested):
We have supported ring_hash since 1.40, and we advertise that fact.  If we have bugs in earlier versions, we should back-port fixes instead of just not testing them.

PR #30240, opened 2022-07-07
Changed num_threads based on slowness factor for xSAN

veblush:
Details: b/236048892 

Recently the duration of `test/cpp/end2end:thread_stress_test` kept increasing and got close to the limit of 30min under mSAN particularly so I changed this test to use less threads based on `grpc_test_slowdown_factor`.

### DIFF
--- a/test/cpp/end2end/thread_stress_test.cc
+++ b/test/cpp/end2end/thread_stress_test.cc
@@ -43,7 +43,13 @@
 using grpc::testing::EchoRequest;
 using grpc::testing::EchoResponse;
 
+#if defined(__APPLE__)
+// Use less # of threads on Mac because its test machines are less powerful
+// to finish the test on time. (context: b/185231823)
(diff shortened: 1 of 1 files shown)


veblush:
`//test/cpp/end2end:thread_stress_test@poller=poll` spent ~8.5 minutes on End2endTest/0 which usually took over 20 minutes.

ctiller (commented):
How many threads does this give TSAN?

veblush:
> How many threads does this give TSAN?

TSAN has slowness factor 5, so it will have 60 threads (300 / 5).

ctiller:
Ok... let's maybe assert this so that future generations don't mess it up... TSAN needs at least 10-ish threads to be effective here.

jtattermusch (approved):
LGTM once you add an assert that num_threads is larger than e.g. 10 (as Craig has suggested).

PR #30243, opened 2022-07-08
xds interop test: temporarily disable failover test

YifeiZhuang:
Blast radius of the test is too high. 
To mitigate, disable until we validate and fix failover test case

Ref b/238226704

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/tests/failover_test.py
+++ b/tools/run_tests/xds_k8s_test_driver/tests/failover_test.py
@@ -18,6 +18,7 @@ from absl import flags
 from absl.testing import absltest
 
 from framework import xds_k8s_testcase
+from framework.helpers import skips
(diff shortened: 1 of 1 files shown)


tools/run_tests/xds_k8s_test_driver/tests/failover_test.py
```diff
@@ -33,6 +33,11 @@ class FailoverTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
     REPLICA_COUNT = 3
     MAX_RATE_PER_ENDPOINT = 100
 
+    @staticmethod
+    def is_supported(config: skips.TestConfig) -> bool:
```
sergiitk, line 37: Skips import is missing. Let me see if I can fix that for you so we can merge this today.

sergiitk (approved):
self five

Self approve!

PR #30246, opened 2022-07-08
Re-Reland: Add SRV and TXT record lookup methods to the iomgr API

drfloob:
Solves an MSAN issue around Orphaning while holding a lock.


drfloob:
Confirmed that releasing the lock before unreffing solves the issue.

1,500 MSAN runs of the problematic code: 5% failure rate (75). 2000 runs with the lock-release fix: 0 errors.

markdroth (approved):

PR #30249, opened 2022-07-08
Allow Gradle to use more memory when building Java interop

ejona86:
Should fix "Expiring Daemon because JVM heap space is exhausted".

https://github.com/grpc/grpc-java/pull/9269 probably pushed the build
over the edge, but there's been evidence via flakes for a good while
that we've been reaching the limit.

b/238438006


ejona86:
CC @sergiitk

gnossen (approved):

PR #30250, opened 2022-07-08
Add outlier detection xDS interop test using k8s interop framework

murgatroid99:
I tested this manually in the staging environment with the C++ client and the Java server.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
+++ b/tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
@@ -29,11 +29,21 @@ RpcTypeEmptyCall = xds_url_map_testcase.RpcTypeEmptyCall
 _XdsTestServer = xds_k8s_testcase.XdsTestServer
 _XdsTestClient = xds_k8s_testcase.XdsTestClient
 
+# Testing consts
+_QPS = 100
+_REPLICA_COUNT = 5
 
 class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
 
+    @staticmethod
+    def is_supported(config: skips.TestConfig) -> bool:
+        if config.client_lang == _Lang.CPP:
+            return not config.version_lt('v1.48.x')
+        if config.client_lang == _Lang.NODE:
+            return not config.version_lt('v1.6.x')
+        return False
+
     def test_outlier_detection(self) -> None:
-        REPLICA_COUNT = 5
 
         with self.subTest('00_create_health_check'):
             self.td.create_health_check()
@@ -59,14 +69,14 @@ class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
 
         test_servers: List[_XdsTestServer]
         with self.subTest('05_start_test_servers'):
-            test_servers = self.startTestServers(replica_count=REPLICA_COUNT)
+            test_servers = self.startTestServers(replica_count=_REPLICA_COUNT)
 
         with self.subTest('06_add_server_backends_to_backend_services'):
             self.setupServerBackends()
 
         test_client: _XdsTestClient
         with self.subTest('07_start_test_client'):
-            test_client = self.startTestClient(test_servers[0], qps=100)
+            test_client = self.startTestClient(test_servers[0], qps=_QPS)
 
         with self.subTest('08_test_client_xds_config_exists'):
             self.assertXdsConfigExists(test_client)
--- a/tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
+++ b/tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
@@ -19,8 +19,6 @@ from absl.testing import absltest
 
 from framework import xds_k8s_testcase
 from framework import xds_url_map_testcase
-from framework.infrastructure import k8s
-from framework.test_app import server_app
 
 logger = logging.getLogger(__name__)
 flags.adopt_module_key_flags(xds_k8s_testcase)
@@ -76,7 +74,7 @@ class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
         with self.subTest('09_test_servers_received_rpcs_from_test_client'):
             self.assertRpcsEventuallyGoToGivenServers(test_client, test_servers)
 
-        rpc_types = [RpcTypeUnaryCall]
+        rpc_types = (RpcTypeUnaryCall, )
         with self.subTest('10_chosen_server_removed_by_outlier_detection'):
             test_client.update_config.configure(
                 rpc_types=rpc_types,
--- a/tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
+++ b/tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
@@ -19,6 +19,7 @@ from absl.testing import absltest
 
 from framework import xds_k8s_testcase
 from framework import xds_url_map_testcase
+from framework.helpers import skips
 
 logger = logging.getLogger(__name__)
 flags.adopt_module_key_flags(xds_k8s_testcase)
@@ -28,6 +29,7 @@ RpcTypeUnaryCall = xds_url_map_testcase.RpcTypeUnaryCall
 RpcTypeEmptyCall = xds_url_map_testcase.RpcTypeEmptyCall
 _XdsTestServer = xds_k8s_testcase.XdsTestServer
 _XdsTestClient = xds_k8s_testcase.XdsTestClient
+_Lang = skips.Lang
 
 # Testing consts
 _QPS = 100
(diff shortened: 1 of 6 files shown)


tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,95 @@
...
+import logging
+from typing import List
+
+from absl import flags
+from absl.testing import absltest
+
+from framework import xds_k8s_testcase
+from framework import xds_url_map_testcase
+from framework.infrastructure import k8s
+from framework.test_app import server_app
```
sergiitk, line 23: These are unused.

```suggestion
```

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,95 @@
...
+RpcTypeUnaryCall = xds_url_map_testcase.RpcTypeUnaryCall
+RpcTypeEmptyCall = xds_url_map_testcase.RpcTypeEmptyCall
+_XdsTestServer = xds_k8s_testcase.XdsTestServer
+_XdsTestClient = xds_k8s_testcase.XdsTestClient
+
+
+class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
+
+    def test_outlier_detection(self) -> None:
+        REPLICA_COUNT = 5
```
sergiitk, line 38: Move this to the module or class level. Method-level variables should be lowercase.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,95 @@
...
+        test_servers: List[_XdsTestServer]
+        with self.subTest('05_start_test_servers'):
+            test_servers = self.startTestServers(replica_count=REPLICA_COUNT)
+
+        with self.subTest('06_add_server_backends_to_backend_services'):
+            self.setupServerBackends()
+
+        test_client: _XdsTestClient
+        with self.subTest('07_start_test_client'):
+            test_client = self.startTestClient(test_servers[0], qps=100)
```
sergiitk, line 71: Is there a reason why qps is 100, and not the default 25? If yes, let's leave a comment.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+        with self.subTest('08_test_client_xds_config_exists'):
+            self.assertXdsConfigExists(test_client)
+
+        with self.subTest('09_test_servers_received_rpcs_from_test_client'):
+            self.assertRpcsEventuallyGoToGivenServers(test_client, test_servers)
+
+        rpc_types = [RpcTypeUnaryCall]
```
sergiitk, line 79: nit: if the list is not going to be modified, prefer a tuple
```suggestion
        rpc_types = (RpcTypeUnaryCall, )
```
```suggestion
        rpc_types = [RpcTypeUnaryCall]
```

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,95 @@
...
+flags.adopt_module_key_flags(xds_k8s_testcase)
+
+# Type aliases
+RpcTypeUnaryCall = xds_url_map_testcase.RpcTypeUnaryCall
+RpcTypeEmptyCall = xds_url_map_testcase.RpcTypeEmptyCall
+_XdsTestServer = xds_k8s_testcase.XdsTestServer
+_XdsTestClient = xds_k8s_testcase.XdsTestClient
+
+
+class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
```
sergiitk, line 35: Do we need to add `is_supported`?

sergiitk (commented):
Looks good, with minor comments. When/if addressed, I'll give it a test run.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
murgatroid99: We discussed this in the PR for the test spec for this test: #29623.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
murgatroid99: Added

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
sergiitk: Ah, right. Ok.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -29,11 +29,21 @@
...
 _XdsTestClient = xds_k8s_testcase.XdsTestClient
 
+# Testing consts
+_QPS = 100
+_REPLICA_COUNT = 5
 
 class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
 
+    @staticmethod
+    def is_supported(config: skips.TestConfig) -> bool:
```
sergiitk, line 40: Missing an import and an alias for _Lang - this will break things unless fixed.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -29,11 +29,21 @@
...
+        if config.client_lang == _Lang.CPP:
+            return not config.version_lt('v1.48.x')
```
sergiitk, line 44: `version_lt` is no longer available, please use `version_gte`

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,106 @@
...
+RpcTypeEmptyCall = xds_url_map_testcase.RpcTypeEmptyCall
+_XdsTestServer = xds_k8s_testcase.XdsTestServer
+_XdsTestClient = xds_k8s_testcase.XdsTestClient
+
+# Testing consts
+_QPS = 100
+_REPLICA_COUNT = 5
+
+
+class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
```
sergiitk, line 37: Would be amazing to add a brief comment here and link https://github.com/grpc/grpc/blob/master/doc/xds-test-descriptions.md#outlier_detection.

I know we don't do this in other places, but it's a good time to start being more descriptive in our tests.

The body of the comment could be the same as in the test spec:

> This test verifies that the client applies the outlier detection configuration and temporarily drops traffic to a server that fails requests.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @staticmethod
+    def is_supported(config: skips.TestConfig) -> bool:
+        if config.server_lang != _Lang.JAVA:
+            return False
+        if config.client_lang == _Lang.CPP:
```
sergiitk, line 43: Do you know if Python gets automatic support now that CPP is supported?

sergiitk (changes requested):

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
murgatroid99: Fixed.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
murgatroid99: Fixed.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
murgatroid99: Added

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
murgatroid99: It probably does. I modified the check to include Python.

tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class OutlierDetectionTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
+    '''
+    Implementation of https://github.com/grpc/grpc/blob/master/doc/xds-test-descriptions.md#outlier_detection
+
+    This test verifies that the client applies the outlier detection
+    configuration and temporarily drops traffic to a server that fails
+    requests.
+    '''
```
sergiitk, line 46:
```suggestion
    """
    Implementation of https://github.com/grpc/grpc/blob/master/doc/xds-test-descriptions.md#outlier_detection

    This test verifies that the client applies the outlier detection
    configuration and temporarily drops traffic to a server that fails
    requests.
    """
```

sergiitk (approved):
Thank you! This is good to go once my docstring fix is applied.

Issue #30247, opened 2022-07-08
Update SymbolTable to new name DefPool


elharo:
See https://critique.corp.google.com/cl/456999727 for necessary changes

nikzayn:
Can I take this up?

yashykt:
Please open a pull request for this

yashykt:
Looks like this was done in https://github.com/grpc/grpc/pull/35208

PR #30258, opened 2022-07-09
Updated SymbolTable to DefPool (#30247)

nikzayn:
In this PR, I have updated the naming convection of SymbolTable to DefPool.

Please do check if there are any other changes required from my end for this.

### DIFF
--- a/src/core/ext/xds/xds_api.cc
+++ b/src/core/ext/xds/xds_api.cc
@@ -82,7 +82,7 @@ XdsApi::XdsApi(XdsClient* client, TraceFlag* tracer,
                const XdsBootstrap::Node* node,
                const CertificateProviderStore::PluginDefinitionMap*
                    certificate_provider_definition_map,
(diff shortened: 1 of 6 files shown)


markdroth:
You'll need to fix the CLA on your commit before we can look at this.

veblush:
I think we should do along with a regular upb update and the further update is currently blocked by https://github.com/protocolbuffers/upb/issues/684.

yashykt:
Sounds like this is no longer blocked on upb, but is now blocked on protobuf, since we need to upgrade the two at the same time.

yashykt:
looks like the protobuf update is postponed to jan of next year

yashykt:
Sounds like we want to have this change, but there are a bunch of upgrades that are needed before this goes through, and when those upgrades are made, we'll need to make many more changes. So, given that this PR by itself is not gonna make it through, we are gonna close it.

PR #30267, opened 2022-07-11
Removed GRPC_ERROR_IS_ABSEIL_STATUS

veblush:
This PR simply removed `GRPC_ERROR_IS_ABSEIL_STATUS` and removed all #ifdef blocks using it. This needs to be cherry-picked. Once this gets merged, the literal `grpc_error_handle` will be replaced with `absl::Status` along with old error macros and functions such as `GRPC_ERROR_IS_NONE`.


src/core/lib/iomgr/error_internal.h
```diff
@@ -28,39 +28,4 @@
 
 #include "src/core/lib/iomgr/error.h"
 
-#ifndef GRPC_ERROR_IS_ABSEIL_STATUS
```
markdroth, line 31: Seems like there's nothing actually left in this file.  Might as well remove it.

veblush: Yep, done!

PR #30270, opened 2022-07-12
Adding lock and reset call_ in ~ClientContext to avoid race condition

LittleCVR:
Adding lock and reset call_ in ~ClientContext to avoid race condition

### DIFF
--- a/src/cpp/client/client_context.cc
+++ b/src/cpp/client/client_context.cc
@@ -77,8 +77,12 @@ ClientContext::ClientContext()
 }
 
 ClientContext::~ClientContext() {
-  if (call_) {
-    grpc_call_unref(call_);
+  {
+    internal::MutexLock lock(&mu_);
+    if (call_) {
+      grpc_call_unref(call_);
+      call_ = nullptr;
+    }
   }
   g_client_callbacks->Destructor(this);
 }
--- a/src/cpp/client/client_context.cc
+++ b/src/cpp/client/client_context.cc
@@ -77,12 +77,9 @@ ClientContext::ClientContext()
 }
 
 ClientContext::~ClientContext() {
-  {
-    internal::MutexLock lock(&mu_);
-    if (call_) {
-      grpc_call_unref(call_);
-      call_ = nullptr;
-    }
+  if (call_) {
+    grpc_call_unref(call_);
+    call_ = nullptr;
   }
   g_client_callbacks->Destructor(this);
 }


src/cpp/client/client_context.cc
```diff
@@ -77,8 +77,12 @@ ClientContext::ClientContext()
 }
 
 ClientContext::~ClientContext() {
-  if (call_) {
-    grpc_call_unref(call_);
+  {
+    internal::MutexLock lock(&mu_);
```
veblush, line 81: Is this lock required? Usually holding a lock in ctor or dtor isn't typical.

LittleCVR: I think this is required, as the test still crashes without holding the lock. Theoretically, someone can access `call_` after `grpc_call_unref()` but before `call_` being set to nullptr, which causes trouble.

veblush: Accessing `client_context` while destroying it sounds wrong to me. Would you check why your test case crash without holding a lock? There could be a bug in your test code.

LittleCVR: It seems that the test code:
1. calls `ClientContext::TryCancel()` to speed up the test tear down, without calling `TryCancel()` the test still works but slower.
2. will not call `ClientContext::TryCancel()` if `ClientContext` has been "fully destructed", because the test code stops tracking `ClientContext` in its `GlobalCallbacks::Destructor()` (which `~ClientContext()` calls as its last statement).

IIUC, the race will be exposed when `TryCancel()` and `~ClientContext()` are called in different thread, to be more specific, when `~ClientContext()` was entered but `call_` hasn't been set to nullptr.

It looks to me that `ClientContext` is supposed to support multithreads environment and it should itself avoid such crash, or is the test code doing anything wrong here?

ctiller: I think the test code is wrong here. An object should not be used once its destructor is called, and it's the callers responsibility to ensure that.

veblush: Setting `call_` to nullptr in the dtor of ClientContext would be reasonable to address the case where the global hook might want to access it in its callback but using the lock doesn't sound right in this case because we don't support case where a ClientContext instance is being destroyed and one of its methods is being called by different threads, which is a race condition.

LittleCVR: Thanks for replying!

@ctiller I do agree we shouldn't use the object once its destructor is called to some extent. But it is gRPC who offers the `GlobalCallbacks` class and its `Destructor()` function here for people to access the object even after its destructor is called (but not yet finished). The `Destructor()` function is given the `ClientContext` object as its argument. What kind of operation does gRPC expect the user to do in the `Destructor()` function if no one should use the object once its destructor is called? (See below paragraph for our use case :)

@veblush The test code is trying to adopt an easy way to "cancel all other calls if a particular `ClientContext` is going to be destroyed". The test code is not randomly calling `TryCancel()` in some random thread in this case, it keeps track of all `ClientContext` when `DefaultConstructor()` is called, and removes them in `Destructor()`. It carefully uses a mutex to ensure that if a `ClientContext` has been fully destroyed, `TryCancel()` will not be called. If this is not enough, then sounds like the `GlobalCallbacks` is not a good candidate to achieve such goal (cancel other calls when a particular `ClientContext` is going to be destroyed). Is there a suggested way to do that? Of course we can go all the way back to our context manager and add a function like "CancelAllCallsForTesting()`, but `GlobalCallbacks` looks like a good and simple shortcut that we really want to use :)

LittleCVR: We've removed our `GlobalCallbacks` which cancels all other RPC calls. It increases the testing time but it's acceptable. We don't need this anymore. The lock is removed.

veblush:
It looks good to me. Craig, would you take a look at this?

PR #30276, opened 2022-07-12
Changing memory benchmark driver's flags

nancylucy01:
Changed the memory benchmark's flags to benchmark_name, size, and scenario_config. This was done to make the driver more configurable for running more benchmarks. Also edited the calling infrastructure (the python script) accordingly.

Tests Passed:
Tests considered passed if value sent to server and client was correct. Checked using print statements

Ran script with size= 0, 10, 1000, 10000
Benchmark_name != "call"
Scenario_config = minstack, secure, resource_quota

Same PR on cloned respository with previous comments: [Link](https://github.com/nancylucy01/grpc/pull/1)


ctiller (approved):

test/core/memory_usage/memory_usage_test.cc
```diff
@@ -32,9 +35,13 @@
...
-ABSL_FLAG(int, warmup, 100, "Warmup iterations");
-ABSL_FLAG(int, benchmark, 1000, "Benchmark iterations");
-ABSL_FLAG(bool, minstack, false, "Use minimal stack");
+ABSL_FLAG(std::string, benchmark_name, "call", "Which benchmark to run");
+ABSL_FLAG(int, size, 50000, "Number of channels/calls");
+ABSL_FLAG(std::string, scenario_config, "insecure",
+          "Possible Values: minstack (Use minimal stack), resource_quota, "
+          "secure (Use SSL credentials on server)");
+ABSL_FLAG(bool, memory_profiling, false,
+          "Run memory profiling");  // not connected to anything yet
```
ananda1066, line 44: Could we add a TODO here as well?

test/core/memory_usage/memory_usage_test.cc
```diff
@@ -72,26 +79,55 @@ int main(int argc, char** argv) {
...
-                  absl::StrCat("--benchmark=", absl::GetFlag(FLAGS_benchmark)),
-                  absl::StrCat("--minstack=", absl::GetFlag(FLAGS_minstack))});
-
-  /* wait for completion */
-  if ((status = cli.Join()) != 0) {
-    printf("client failed with: %d", status);
-    return 1;
+
+  /* Set configurations */
+  if (absl::GetFlag(FLAGS_size) == 0) absl::SetFlag(&FLAGS_size, 50000);
```
ananda1066, line 84: How about we remove this line and just say that if size flag is not set, we default to 50000 (which is the default of the flag anyway)?

test/core/memory_usage/memory_usage_test.cc
```diff
@@ -72,26 +79,55 @@ int main(int argc, char** argv) {
...
+  // TODO(chennancy): add in resource quota parameter setting later
+  const std::map<std::string /*scenario*/, ScenarioArgs> scenarios = {
+      {"secure", {/*client=*/{}, /*server=*/{"--secure"}}},
+      {"resource_quota", {/*client=*/{}, /*server=*/{"--secure"}}},
+      {"minstack", {/*client=*/{"--minstack"}, /*server=*/{"--minstack"}}},
+      {"insecure", {{}, {}}},
+  };
+  auto it_scenario = scenarios.find(absl::GetFlag(FLAGS_scenario_config));
+  if (it_scenario == scenarios.end()) {
+    return 3;
```
ananda1066, line 99: Please leave a comment on what this error means - that no viable scenario could be found? (and maybe a printf statement as well)

test/core/memory_usage/memory_usage_test.cc
```diff
@@ -72,26 +79,55 @@ int main(int argc, char** argv) {
...
+  }
+
+  /* per-call memory usage benchmark */
+  if (absl::GetFlag(FLAGS_benchmark_name) == "call") {
+    /* start the server */
+    std::vector<std::string> server_flags = {
+        absl::StrCat(root, "/memory_usage_server",
+                     gpr_subprocess_binary_extension()),
+        "--bind", grpc_core::JoinHostPort("::", port)};
+    absl::c_move(it_scenario->second.server, std::back_inserter(server_flags));
```
ananda1066, line 109: let's change the c_move to std::move for easier readability (we tend to use std::move)

test/core/memory_usage/memory_usage_test.cc
```diff
@@ -72,26 +79,55 @@ int main(int argc, char** argv) {
...
+    Subprocess svr(server_flags);
+
+    /* start the client */
+    std::vector<std::string> client_flags = {
+        absl::StrCat(root, "/memory_usage_client",
+                     gpr_subprocess_binary_extension()),
+        "--target", grpc_core::JoinHostPort("127.0.0.1", port),
+        absl::StrCat("--warmup=", 10000),
+        absl::StrCat("--benchmark=", absl::GetFlag(FLAGS_size))};
+    absl::c_move(it_scenario->second.client, std::back_inserter(client_flags));
```
ananda1066, line 119: same comment as above

test/core/memory_usage/memory_usage_test.cc
ananda1066: I think we can also rewrite this as something like `server_flags.emplace_back(std::move(it_scenario->second.server))` for readability

test/core/memory_usage/memory_usage_test.cc
nancylucy01: Sounds good, changed it

test/core/memory_usage/memory_usage_test.cc
nancylucy01: Alright, seems reasonable now that we changed the default of the flag to 50000, I'll update the documentation accordingly.

test/core/memory_usage/memory_usage_test.cc
nancylucy01: same response as above

test/core/memory_usage/memory_usage_test.cc
nancylucy01: Added a print statement saying "No scenario matching the name could be found"

ananda1066 (approved):

PR #30277, opened 2022-07-12
XdsClient: refcount tracing improvements

markdroth:
@ctiller You may be interested in the changes to channel_args.h.


yashykt (commented):
non-blocking comment: I think I saw a message somewhere that SourceLocation is now available

src/core/ext/xds/xds_client.cc
```diff
@@ -1417,10 +1417,10 @@ void XdsClient::Orphan() {
 }
 
 RefCountedPtr<XdsClient::ChannelState> XdsClient::GetOrCreateChannelStateLocked(
-    const XdsBootstrap::XdsServer& server) {
+    const XdsBootstrap::XdsServer& server, const char* reason) {
```
yashykt, line 1420: check if absl::string_view is better here

yashykt (approved):

src/core/ext/xds/xds_client.cc
markdroth: In this case, the only place this value will be used is in `gpr_log()`, which takes `const char*`, so using `absl::string_view` would require dynamic allocation (`std::string(reason).c_str()`) every time we log the string.

PR #30280, opened 2022-07-12
Make GetDefaultEventEngine return a shared_ptr

drfloob:
This works around valgrind memory leaks by giving EventEngines a fixed
lifetime. We eventually want ref-counted EventEngines internally, so this is 
a step in the right direction as well.


include/grpc/event_engine/event_engine.h
```diff
@@ -423,6 +423,10 @@ class EventEngine {
   /// closure may need to be destroyed for things to progress (e.g., if a
   /// closure holds a ref to some ref-counted object).
   virtual bool Cancel(TaskHandle handle) = 0;
+
+  /// Return a shared_ptr to this instance.
+  /// The instance must already be managed by a shared_ptr.
+  std::shared_ptr<EventEngine> GetSharedPtr() { return shared_from_this(); }
```
ctiller, line 429: Unclear whether adding a method that's already there is worth it?

(I can always do `ee->shared_from_this();` and since that's part of std it's maybe more likely to be clear what's going on?)

test/core/event_engine/default_engine_methods.cc
```diff
@@ -0,0 +1,107 @@
...
+    }
+    TaskHandle RunAfter(Duration when,
+                        absl::AnyInvocable<void()> closure) override {
+      return {-1, -1};
+    }
+    bool Cancel(TaskHandle handle) override { return false; };
+  };
+};
+
+TEST_F(DefaultEngineTest, SharedPtrGlobalEventEngineLifetimesAreValid) {
```
ctiller, line 74: let's add a test with a dozen-ish threads doing `GetDefaultEventEngine().reset();` in a loop for a few seconds

include/grpc/event_engine/event_engine.h
drfloob: Good point, removed.

test/core/event_engine/default_engine_methods.cc
drfloob: Added, and it found a race. I reworked the default engine storage using weak_ptrs, it's a bit cleaner now.

drfloob:
@ctiller there's a issue with promise/activity contexts, wherein activities are being run on different threads than those on which they were created. I confirmed this is happening with the channel idle filters. Promise Contexts are all thread_local, so they aren't being found. I think we may need a non-thread-local Context type. Best I can imagine is a fork of `MakeActivity`, maybe `MakeThreadLocalActivity`, with the default being non-thread-local, but I could use your advice here.

ctiller (approved):

src/core/lib/event_engine/default_event_engine.h
```diff
@@ -17,16 +17,25 @@
...
 #include <grpc/support/port_platform.h>
 
+#include <memory>
+
 #include <grpc/event_engine/event_engine.h>
 
+#include "src/core/lib/promise/context.h"
+
+namespace grpc_core {
+template <>
```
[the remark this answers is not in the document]
drfloob: I had the same concern regarding the `IWYU pragma: keep`, but no great alternative ideas. Happy to move it when we think of something better.

src/core/lib/event_engine/event_engine.cc
```diff
@@ -0,0 +1,80 @@
+// Copyright 2021 The gRPC Authors
```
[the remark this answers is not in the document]
drfloob: Stowaway removed. I think it snuck back in on a merge conflict.

drfloob:
This does not require a cherrypick, confirmed via preparing the cherrypick (and via grep)

PR #30285, opened 2022-07-13
Python examples: change localhost to 127.0.0.1

### LINKED ISSUE #30282: Extremely slow (>30 seconds) responses for Python examples
grpcio: 1.47.0
grpcio-tools: 1.47.0
Python: 3.10.4

Ubuntu 22.04 LTS

Python 3.10.4

### What did you do?
Follow the instructions on the gRPC Python quickstart https://grpc.io/docs/languages/python/quickstart/

1. Download the example
```
# Clone the repository to get the example code:
git clone -b v1.46.3 --depth 1 --shallow-submodules https://github.com/grpc/grpc
# Navigate to the "hello, world" Python example:
cd grpc/examples/python/helloworld
```

2. Run a gRPC application
Run the server
```
python greeter_server.py
```
From another terminal, run the client:
```
python greeter_client.py
```
A quick response message `Greeter client received: Hello, you!` from the server within less than a second

### What did you see instead?
A very slow response message that responds after more than 30 seconds.

### Possible fix
The problem appears to be with the client which connects to  `localhost:50051` as it takes too long to resolve. When I change it to `127.0.0.1:50051` the response is near instant as expected. 
https://github.com/grpc/grpc/blob/24171b3b57e9453c9d91f249eecf3cccd4000d95/examples/python/helloworld/greeter_client.py#L25-L32

I have tested with other examples as well and I'm facing the same issue there too.

I have tested with both v.1.46.3 and with the latest commit as of writing this issue -> 24171b3b57e9453c9d91f249eecf3cccd4000d95

alielbashir:
This PR changes all `localhost` mentions to `127.0.0.1` in the Python examples, which speeds up requests by > 30 seconds on some machines.

Resolves #30282

### DIFF
--- a/examples/python/async_streaming/client.py
+++ b/examples/python/async_streaming/client.py
@@ -108,7 +108,7 @@ def process_call(executor: ThreadPoolExecutor, channel: grpc.Channel,
 
 def run():
     executor = ThreadPoolExecutor()
-    with grpc.insecure_channel("localhost:50051") as channel:
+    with grpc.insecure_channel("127.0.0.1:50051") as channel:
         future = executor.submit(process_call, executor, channel,
                                  "555-0100-XXXX")
         future.result()
--- a/examples/python/auth/async_customized_auth_client.py
+++ b/examples/python/auth/async_customized_auth_client.py
@@ -26,7 +26,7 @@ helloworld_pb2, helloworld_pb2_grpc = grpc.protos_and_services(
 _LOGGER = logging.getLogger(__name__)
 _LOGGER.setLevel(logging.INFO)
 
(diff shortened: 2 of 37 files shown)


gnossen:
`localhost` also isn't specific to IPV4 or IPV6, whereas `127.0.0.1` certainly is. The vast majority of (posix) environments include `localhost` in `/etc/hosts`, so the lookup should be nearly instantaneous for the majority of users. Cases where it takes tens of seconds to look up `localhost` seem highly pathological, so I would need some strong evidence from the community (i.e. more user complaints) to make this change.

PR #30305, opened 2022-07-15
Basic callback client and server for new perchannel benchmark

nancylucy01:
Added basic callback client and server files, and changed the benchmark driver to be able to call them if benchmark_name is "channel"
Currently running the per channel benchmark without the scenario flags since secure flag doesn't exist in the call client benchmark and it needs to for the driver to be able to pass the correct flags to both the per channel and per call benchmark. Will be fixed in a different PR


test/core/memory_usage/callback_client.cc
```diff
@@ -68,6 +66,31 @@ class ClientCallbackImpl {
...
+
+    CallParams* params = new CallParams();
+
+    auto callback = [params](const grpc::Status& status) {
+      if (!status.ok()) {
+        gpr_log(GPR_ERROR, "ServerQuit RPC failed.");
+        delete params;
+        return;
+      }
+      gpr_log(GPR_INFO, "ServerQuit RPC succeeded.");
```
krishnacx, line 89: We can simplify to:

```
if(!status.ok)
{
 gpr_log(GPR_ERROR, "ServerQuit RPC failed.");
}
else
{
 gpr_log(GPR_INFO, "ServerQuit RPC succeeded.");
}
delete params;
        return;
```

test/core/memory_usage/callback_server.cc
```diff
@@ -0,0 +1,109 @@
...
+  // Set the authentication mechanism.
+  std::shared_ptr<grpc::ServerCredentials> creds =
+      grpc::InsecureServerCredentials();
+  if (absl::GetFlag(FLAGS_secure)) {
+    gpr_log(GPR_INFO, "Supposed to be secure, is not yet");
+    // TODO (chennancy) Add in secure credentials
+  }
+  builder.AddListeningPort(server_address, creds);
+
+  // Register "service" as the instance through which we'll communicate with
```
ananda1066, line 96: I don't think this comment is needed.

test/core/memory_usage/callback_server.cc
```diff
@@ -0,0 +1,109 @@
...
+  // clients.
+  builder.RegisterService(&callback_server);
+
+  signal(SIGINT, sigint_handler);
```
ananda1066, line 100: Perhaps add a comment on why this is needed? I think per_call has a comment that you can reference

test/core/memory_usage/memory_usage_test.cc
```diff
@@ -125,6 +128,31 @@ int main(int argc, char** argv) {
...
+    /* Per-channel benchmark*/
+    /* start the server */
+    std::vector<std::string> server_flags = {
+        absl::StrCat(root, "/memory_usage_callback_server",
+                     gpr_subprocess_binary_extension()),
+        "--bind", grpc_core::JoinHostPort("::", port)};
+    Subprocess svr(server_flags);
+
+    // Wait one second before starting client to avoid race conditions
+    gpr_sleep_until(grpc_timeout_seconds_to_deadline(1));
```
ananda1066, line 143: Is this needed?

test/core/memory_usage/callback_client.cc
```diff
@@ -0,0 +1,108 @@
...
+#include "test/core/util/test_config.h"
+
+ABSL_FLAG(std::string, target, "localhost:443", "Target host:port");
+ABSL_FLAG(bool, secure, false, "Use SSL Credentials");
+
+void UnaryCall(std::shared_ptr<grpc::ChannelCredentials> creds) {
+  // Create a channel to the server and a stub
+  std::shared_ptr<grpc::Channel> channel =
+      CreateChannel(absl::GetFlag(FLAGS_target), creds);
+  std::unique_ptr<grpc::testing::BenchmarkService::Stub> stub_ =
```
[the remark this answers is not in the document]
nancylucy01: Thanks, I renamed the stub and removed the whitespaces according to the guide.

test/core/memory_usage/callback_client.cc
```diff
@@ -0,0 +1,108 @@
...
+#include "test/core/util/port.h"
+#include "test/core/util/subprocess.h"
+#include "test/core/util/test_config.h"
+
+ABSL_FLAG(std::string, target, "localhost:443", "Target host:port");
+ABSL_FLAG(bool, secure, false, "Use SSL Credentials");
+
+void UnaryCall(std::shared_ptr<grpc::ChannelCredentials> creds) {
+  // Create a channel to the server and a stub
+  std::shared_ptr<grpc::Channel> channel =
```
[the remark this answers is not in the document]
nancylucy01: I didnt realize that was an issue, changed it, thank you!

test/core/memory_usage/callback_client.cc
```diff
@@ -0,0 +1,108 @@
...
+
+  struct CallParams {
+    grpc::ClientContext context;
+    grpc::testing::SimpleRequest request;
+    grpc::testing::SimpleResponse response;
+  };
+
+  CallParams* params = new CallParams();
+
+  auto callback = [](const grpc::Status& status) {
```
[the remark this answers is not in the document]
nancylucy01: Is this what you meant? I moved it within the parameters of UnaryCall

test/core/memory_usage/callback_server.cc
```diff
@@ -0,0 +1,109 @@
...
+int main(int argc, char** argv) {
+  absl::ParseCommandLine(argc, argv);
+  char* fake_argv[1];
+
+  GPR_ASSERT(argc >= 1);
+  fake_argv[0] = argv[0];
+  grpc::testing::TestEnvironment env(&argc, argv);
+
+  grpc_init();
+  std::string server_address = absl::GetFlag(FLAGS_bind);
```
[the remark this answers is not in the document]
nancylucy01: Added in the edge case, but if that edge case happens, how does the client know which port to target?

test/core/memory_usage/callback_server.cc
nancylucy01: Removed it

test/core/memory_usage/callback_server.cc
nancylucy01: Added the per_call comment "We have some sort of deadlock, so let's not exit gracefully for now."
Also added a TODO for graceful shutdown

test/core/memory_usage/memory_usage_test.cc
nancylucy01: Based on running the test 200 times for each, without sleep had an 10% RPC failure rate, while with sleep had an 0% failure rate. I assume its from the client sending the RPC before the server is set up.

test/core/memory_usage/callback_client.cc
nancylucy01: Thanks, simplified it

test/core/memory_usage/callback_server.cc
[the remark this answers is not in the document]
ananda1066: Great question; I don't think it will know - perhaps we log an error and return here instead?

test/core/memory_usage/memory_usage_test.cc
ananda1066: Ok sounds good, we should expand on the comment to explain what the possible race condition is then

test/core/memory_usage/callback_client.cc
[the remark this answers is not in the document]
ananda1066: Yup, thanks!

ananda1066 (changes requested):

test/core/memory_usage/callback_server.cc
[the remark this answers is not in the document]
nancylucy01: Sounds good, I added the edge case to both client and server

test/core/memory_usage/memory_usage_test.cc
nancylucy01: Alright, changed the comment

ananda1066 (approved):

test/core/memory_usage/callback_client.cc
```diff
@@ -0,0 +1,106 @@
...
+  std::unique_ptr<grpc::testing::BenchmarkService::Stub> stub =
+      grpc::testing::BenchmarkService::NewStub(channel);
+
+  // Start a call.
+  struct CallParams {
+    grpc::ClientContext context;
+    grpc::testing::SimpleRequest request;
+    grpc::testing::SimpleResponse response;
+  };
+  CallParams* params = new CallParams();
```
ctiller, line 80: Can you comment on the ownership of params (what deletes it and where)

test/core/memory_usage/memory_usage_test.cc
```diff
@@ -25,15 +25,18 @@
...
 #include <grpc/support/string_util.h>
+#include <grpc/support/time.h>
 
 #include "src/core/lib/gpr/string.h"
 #include "src/core/lib/gprpp/host_port.h"
 #include "test/core/util/port.h"
 #include "test/core/util/subprocess.h"
+#include "test/core/util/test_config.h"
 
 ABSL_FLAG(std::string, benchmark_name, "call", "Which benchmark to run");
```
ctiller, line 41: We're not testing the `channel` branch in CI with this change.

Suggestion:
1. rename `benchmark_name` to `benchmark_names`, default it to `"call,channel"`
2. change main() to be:
```
int main(int argc, char**argv) {
  // code to ~line 103
  auto benchmarks = absl::StrSplit(absl::GetFlag(FLAGS_benchmark_name));
  for (const auto& benchmark : benchmarks) {
    int r = RunBenchmark(benchmark, ...); // ... is whatever context we need to pass down
    if (r != 0) return r;
  }
}
```
3. Write the new RunBenchmark function:
```
int RunBenchmark(absl::string_view benchmark, ...) {
  if (benchmark == "call") return RunCallBenchmark(...);
  else if (benchmark == "channel") return RunChannelBenchmark(...);
  else { log("that's not a benchmark"); return 1; }
}
```

ctiller (commented):
a couple of nits, and one substantive thing: we're not testing this code and we need to be
i left a comment on a suggestion on how to restructure, although other restructurings are likely possible

test/core/memory_usage/callback_client.cc
ananda1066: Hm IIRC, if we delete the params in the callback, the connection closes; where would be an appropriate place to delete them? Perhaps we remove the struct and avoid heap-allocated memory here?

test/core/memory_usage/memory_usage_test.cc
nancylucy01: Would it be better to do this or to have the calling python script call the benchmark again but with benchmark_name="channel"? Since right now we only have one size flag since we weren't accounting for multiple benchmarks to be run at once. Also I'm not sure if making it run multiple benchmarks within the driver will cause issues for the python script when working with the output.

nancylucy01: What is required for the CI to test the channel branch?

test/core/memory_usage/callback_client.cc
```diff
@@ -0,0 +1,106 @@
...
+  if (absl::GetFlag(FLAGS_secure)) {
+    // TODO (chennancy) Add in secure credentials
+    gpr_log(GPR_INFO, "Supposed to be secure, is not yet");
+  }
+
+  // Create a channel to the server and a stub
+  std::shared_ptr<grpc::Channel> channel =
+      CreateChannel(absl::GetFlag(FLAGS_target), creds);
+  std::unique_ptr<grpc::testing::BenchmarkService::Stub> stub =
+      grpc::testing::BenchmarkService::NewStub(channel);
```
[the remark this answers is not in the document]
nancylucy01: Great suggestion! I changed it in the next PR

test/core/memory_usage/memory_usage_test.cc
ctiller: the `grpc_cc_test` will invoke the binary with no arguments - we can tweak it to add some arguments, but that will be clumsy.

Another way to approach this might be:
1. Split out the RunCallBenchmark, RunChannelBenchmark functions into a new library (or libraries)
2. have the main.cc file call these
3. have a call_test.cc, channel_test.cc program that just calls the appropriate benchmark driver

nancylucy01: I think I'll go with the first suggestion and just change size to sizes so that if benchmark_names="call, channel" and sizes =  "50000, 10000" then call will run with 50000 calls and channel will run with 10000 channels. Or perhaps it might be better to do benchmark="call_50000, channel_10000"?

nancylucy01: Discussed over call, going with first suggestion and changing default size to smaller. Different sizes not needed as the default is just for the CI to test both, and normally it will be called with only one benchmark.

ctiller (approved):

PR #30307, opened 2022-07-15
Detect invalid UTF-8 characters

ralphchung:
This PR aims to fix the bug [b/208085307](http://b/208085307).

The test case fails because the input fed to the `grpc_core::JsonReader::Parse()` is invalid, but `grpc_core::JsonReader::Parse()` cannot detect that.

This change references largely from Table 3-7 in http://www.unicode.org/versions/Unicode6.2.0/ch03.pdf which shows the range of valid UTF-8 characters.


ctiller:
Please add relevant test cases to json_test.cc

src/core/lib/json/json_reader.cc
```diff
@@ -137,29 +138,44 @@ class JsonReader {
...
+      return false;
+    }
+    utf8_first_byte_ = c;
+  } else if (utf8_bytes_remaining_ == 1) {
+    if ((c & 0xc0) != 0x80) {
+      return false;
+    }
+    --utf8_bytes_remaining_;
+  } else if (utf8_bytes_remaining_ == 2) {
+    /// Reference: https://www.unicode.org/versions/Unicode14.0.0/ch03.pdf
```
Vignesh2208, line 160: Please add the list of UTF-8 characters this is supposed to filter out directly inside the comment itself and then provide the link to the document.

ralphchung:
@ctiller @Vignesh2208 
Please have another review. Thanks!

Vignesh2208:
> @ctiller @Vignesh2208 Please have another review. Thanks!

Thanks for making the change. lgtm

Vignesh2208 (approved):

ctiller (approved):

PR #30309, opened 2022-07-15
xds interop: choose correct cluster in grpc_xds_k8s_lb_python.sh

sergiitk:
Discovered it while investigating the failover test issue. For some reason, python LB tests are using the cluster used for PSM Security tests - and also the same cluster for primary and secondary roles.

cc @YifeiZhuang


sergiitk:
Test run:  https://source.cloud.google.com/results/invocations/85169720-a90f-4b94-986b-41018f52dd9a

gnossen (approved):

PR #30320, opened 2022-07-18
Extend ENOBUFS handling logic to prevent deadlocks when ENOBUFS is returned due to filled up queues at the network interface

Vignesh2208:
If ENOBUFS is returned when zero copy is enabled and no un-acked zero-copy records are present at the time of ENOBUFS generation, then EPOLLERR would not get generated and the connection would deadlock because the write thread would never wake up. Detect this condition and prevent the write thread from going to an indefinite sleep in that case.

### DIFF
--- a/src/core/lib/iomgr/tcp_posix.cc
+++ b/src/core/lib/iomgr/tcp_posix.cc
@@ -478,7 +478,7 @@ namespace {
 int GetRLimitMemLockMax() {
   static int kRlimitMemLock = []() -> int {
     struct rlimit limit;
-    if (getrlimit(RLIMIT_MEMLOCK, &limit)) {
+    if (getrlimit(RLIMIT_MEMLOCK, &limit) != 0) {
       return -1;
     }
     return static_cast<int>(limit.rlim_max);


arjunroy (changes requested):
Fundamentally, not having enough buffer space on a non-blocking socket will return EAGAIN, not ENOBUFS. So we do not have to add all this logic to reduce send size.

ENOBUFS from sendmsg() has three possible causes that jump out at me:
1) Not enough optmem
2) TCP fastopen not working out at the start of a connection
3) msg_controllen value too large (ie. it's a 64 bit value in the struct but it will fail if set to > INT_MAX).

Neither of these would suggest reducing the size of the send.
Does the "cont" logic, in isolation, address the problem this patch is trying to fix?

soheilhy:
My understanding is similar to what Arjun mentioned. If we actually run of wmem we will just do partial write or `EAGAIN`.

Vignesh2208:
> My understanding is similar to what Arjun mentioned. If we actually run of wmem we will just do partial write or `EAGAIN`.

I looked at this a bit more. It seems RLIMIT_MEMLOCK is a limit which controls the max number of pages that can be pinned to RAM: https://elixir.bootlin.com/linux/latest/source/net/core/skbuff.c#L1137

Its used in the tcp zero-copy path and from my understanding of the kernel code, it looks like it can return ENOBUFS if the number of pages a process is trying to pin to RAM through zero-copy exceeds this limit. 

I removed the max bytes per sendmsg logic and added an error statement to warn the user now.

src/core/lib/iomgr/tcp_posix.cc
```diff
@@ -133,6 +134,9 @@ class TcpZerocopySendRecord {
   // Indicates whether all underlying data has been sent or not.
   bool AllSlicesSent() { return out_offset_.slice_idx == buf_.count; }
 
+  // Returns the total length of the underlying data;
```
arjunroy, line 137: Clarify this comment. "This is the total length sent via the successful sendmsg(MSG_ZEROCOPY) call that created this TcpZerocopySendRecord".

src/core/lib/iomgr/tcp_posix.cc
```diff
@@ -381,10 +385,15 @@ class TcpZerocopySendCtx {
   //
   // Please refer to the STATE TRANSITION DIAGRAM below for more details.
   //
-  bool UpdateZeroCopyOMemStateAfterSend(bool seen_enobuf) {
+  bool UpdateZeroCopyOMemStateAfterSend(bool seen_enobuf, bool& cont) {
```
arjunroy, line 388: See comment below. Can we just elide this parameter by returning "true" in any scenario where we would have set this to true?

src/core/lib/iomgr/tcp_posix.cc
```diff
@@ -381,10 +385,15 @@ class TcpZerocopySendCtx {
...
     MutexLock guard(&lock_);
     is_in_write_ = false;
+    cont = false;
     if (seen_enobuf) {
+      if (ctx_lookup_.size() == 1) {
+        // There is no un-acked z-copy record. Retry the sendmsg again.
```
arjunroy, line 394: I wouldn't use acked in this context since it might be confused for TCP Acks, which this is not. Also, say why we are retrying the sendmsg().

Something like: "There are no other pending TCP Tx0cp completions, and thus we will not receive Tx0cp EPOLLERR events since this sendmsg() failed. Therefore, set the socket as writeable now so we retry the send in a timely manner."

arjunroy (changes requested):
Generally speaking I'm ok with the scope of this patch now, but did have some requests.

PR #30340, opened 2022-07-19
interop: Add v1.48.0 to interop testing for C++, php, ruby and python

yashykt:
Not adding csharp right now since that was erroring out @jtattermusch


jtattermusch (approved):
LGTM once the adhoc interop matrix test is green (looks like it is already).

Note that 1.47.1 tag should also be added.
I'll make sure of updating the docs so that "csharp" is no longer listed in the instructions.

PR #30370, opened 2022-07-21
Remove empty function grpc_cq_global_init and grpc_executor_global_init

panzhongxian:
This function hasn't been call anywhere except in `do_basic_init()`.

Some emtpy functions are nessesary:

- Empty virtual destructor
- Empty functions used for placeholder, like `grpc_lb_policy_grpclb_shutdown()`
- Empty functions according to different macro definitions, like `gpr_precise_clock_init()`

Unlike those empty function, the `grpc_cq_global_init()` seems useless. The same as function `grpc_executor_global_init()`

Can it be removed. Or keep it for compatibility: since it is declared in header file and not static, some one may have used it?

### DIFF
--- a/src/core/lib/debug/stats.cc
+++ b/src/core/lib/debug/stats.cc
@@ -31,7 +31,6 @@
 
 #include <grpc/support/alloc.h>
 #include <grpc/support/cpu.h>
-#include <grpc/support/sync.h>
 
 grpc_stats_data* grpc_stats_per_cpu_storage = nullptr;
 static size_t g_num_cores;
--- a/src/core/lib/debug/stats.cc
+++ b/src/core/lib/debug/stats.cc
@@ -35,14 +35,11 @@
 
 grpc_stats_data* grpc_stats_per_cpu_storage = nullptr;
 static size_t g_num_cores;
-static gpr_once g_once = GPR_ONCE_INIT;
 
 void grpc_stats_init(void) {
-  gpr_once_init(&g_once, []() {
-    g_num_cores = gpr_cpu_num_cores();
-    grpc_stats_per_cpu_storage = static_cast<grpc_stats_data*>(
-        gpr_zalloc(sizeof(grpc_stats_data) * g_num_cores));
-  });
+  g_num_cores = gpr_cpu_num_cores();
+  grpc_stats_per_cpu_storage = static_cast<grpc_stats_data*>(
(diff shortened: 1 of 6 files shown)


ctiller:
So long as the tests work out this looks good to me. We could probably do away with the gpr_time_precise bits now if someone went through and converted the channelz stuff to use std::chrono::steady_clock instead of cycle counts.

drfloob:
I didn't find any internal uses of these either, looks like this deletion was made safe about a year ago with https://github.com/grpc/grpc/pull/26999. 

@panzhongxian there are some sanity check failures that need to be addressed (IWYU and clang format), please fix those and we can move forward.

panzhongxian:
Hi, @drfloob  I saw the sanity checks failure. But I executed the following command to format code file using the docker:

```bash
tools/distrib/clang_format_code.sh
```

No file changed. Are the versions and configs of clang-format in docker and pipeline same?

ctiller:
You'll want to run tools/distrib/iwyu.sh followed by
tools/distrib/clang_format_code.sh.
The first will insert some header files but mess up formatting, the second
will then restore the formatting.

ctiller:
I guess... the other way to do it is to allow our pr-autofix action to run in your repo, and it'll send you a PR on top of yours with the relevant corrections. You'll need to go to Actions on your fork to enable it.

ctiller:
(oh and you'll probably need to repush)

panzhongxian:
> I guess... the other way to do it is to allow our pr-autofix action to run in your repo, and it'll send you a PR on top of yours with the relevant corrections. You'll need to go to Actions on your fork to enable it.

@ctiller Thanks for your direction.

panzhongxian:
By the way, is it reasonable moving `grpc_stats_init()` lambda function into `do_basic_init()`?  `grpc_stats_init()` is called only once and no corresponding destroy function or shutdown fucntion.

ctiller:
It looks like the sanity tests have broken again... let's get that fixed again (i think this change seems reasonable still)

I'd recommend that further changes get broken out into a different PR - it'd be good to get this landed.

panzhongxian:
> It looks like the sanity tests have broken again... let's get that fixed again (i think this change seems reasonable still)
> 
> I'd recommend that further changes get broken out into a different PR - it'd be good to get this landed.

Thanks. I'll follow your recommendation if there were new changes.

ctiller (approved):

yashykt (approved):

Issue #27502, opened 2021-09-28
Grpc.Tools signals libprotobuf warnings as errors


codami:
### What version of gRPC and what language are you using?
C#
Google.Protobuf: 3.18.0
Grpc.AspNetCore: 2.39.0
Grpc.Tools: 2.40.0

### What operating system (Linux, Windows,...) and version?
Linux & Windows

### What did you do?
I was building a .NET Class Library (gRPC Server side code) out of a proto file, using "dotnet build" and the _Protobuf_CoreCompile task failed after a couple of errors like this one:
`[libprotobuf WARNING T : error : \src\github\grpc\workspace_protoc_windows_x64\third_party\protobuf\src\google\protobuf\compiler\csharp\csharp_enum.cc:74] Duplicate enum value Short (originally MY_VIEW_SHORT) in MyView; adding underscore to distinguish
`

### What did you expect to see?
I would expect to see warnings at most, since the libprotobuf output was just a warning.

### What did you see instead?
The build failed because of errors.

After starting the command manually I have noticed that the original output from the libprotobuf is this one:
`[libprotobuf WARNING T:\src\github\grpc\workspace_protoc_windows_x64\third_party\protobuf\src\google\protobuf\compiler\csharp\csharp_enum.cc:74] Duplicate enum value Short (originally MY_VIEW_SHORT) in MyView; adding underscore to distinguish`

A closer look at the Grpc.Tools implementation shows that the warning above is converted into error by this matching regex: https://github.com/grpc/grpc/blob/ab6beb3f686857d687bf72e80b1fbaf7d3c3e4a4/src/csharp/Grpc.Tools/ProtoCompile.cs#L214

tonydnewell:
I've created a PR to fix this

PR #30371, opened 2022-07-21
issue 27502 - parse warning from libprotobuf

tonydnewell:
Fix for https://github.com/grpc/grpc/issues/27502

The warnings and errors logged by some of the plugins such as the csharp plugin are in a different format from other warnings etc logged by the protobuf compiler (the plugins use GOOGLE_LOG to log message).

The parsing code in Grpc.Tools has been modified to correctly parse these so that warnings are no longer handled as errors.

The unit tests have been extended and fixed (they did not notice if a test case did not match the expected log level).

### DIFF
--- a/src/csharp/Grpc.Tools/ProtoCompile.cs
+++ b/src/csharp/Grpc.Tools/ProtoCompile.cs
@@ -206,6 +206,63 @@ namespace Grpc.Tools
                 }
             },
 
+            // Example warning from plugins that use GOOGLE_LOG
+            // [libprotobuf WARNING T:\altsrc\..\csharp_enum.cc:74] Duplicate enum value Work (originally Work) in PhoneType; adding underscore to distinguish
+            new ErrorListFilter
+            {
+                Pattern = new Regex(
+                    pattern: "^\\[.+? WARNING (?'FILENAME'.+?):(?'LINE'\\d+?)\\] ?(?'TEXT'.*)",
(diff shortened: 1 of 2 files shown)


src/csharp/Grpc.Tools/ProtoCompile.cs
```diff
@@ -206,6 +206,63 @@ public class ProtoCompile : ToolTask
...
+                Pattern = new Regex(
+                    pattern: "^\\[.+? WARNING (?'FILENAME'.+?):(?'LINE'\\d+?)\\] ?(?'TEXT'.*)",
+                    options: RegexOptions.Compiled | RegexOptions.IgnoreCase,
+                    matchTimeout: s_regexTimeout),
+                LogAction = (log, match) =>
+                {
+                    // The filename and line logged by the plugins may not be useful to the
+                    // end user as they are not the location in the proto file but rather
+                    // in the source code for the plugin. Log them anyway as they may help in
+                    // diagnositics.
```
jskeet, line 222: Nit: diagnositics => diagnostics, here and below

jskeet (approved):
I'm not massively familiar with this code, but assuming the new tests failed before and pass now, this LGTM.

src/csharp/Grpc.Tools/ProtoCompile.cs
tonydnewell: fixed typos

jtattermusch (approved):
LGTM.

PR #30376, opened 2022-07-21
[objc] Migrate interop remote test target to test runner

dennycd:
Migrate last ios test to grpc_objc_ios_unit_test using test runner which has lower failure rate of simulator launch on kokoro nodes. 

### Change Summary 

* Update InteropTestsRemote to using grpc_objc_ios_unit_test. 
   - Clean up and removal of ios_host targets since it is no longer in use. 

* Bumping up retry count to 10 for network flakes. 
   - Since our current timeout wait is very short  (3 seconds), the overall timeout in worst scenario (30 seconds) would still be shorter than our original 64 seconds test timeout. This should allow us to accomodate most kokoro flakes for objc bazel tests.   

### Test & Verify 

objc bazel test pass green for interop remote test

---

cc @HannahShiSFB  @jtattermusch

### DIFF
--- a/tools/internal_ci/macos/grpc_objc_bazel_test.sh
+++ b/tools/internal_ci/macos/grpc_objc_bazel_test.sh
@@ -100,7 +100,7 @@ objc_bazel_tests/bazel_wrapper \
   $BAZEL_FLAGS \
   --test_env HOST_PORT_LOCAL=localhost:$PLAIN_PORT \
   --test_env HOST_PORT_LOCALSSL=localhost:$TLS_PORT \
-  --test_env FLAKE_TEST_REPEATS=3 \
+  --test_env FLAKE_TEST_REPEATS=10 \
   -- \
   "${EXAMPLE_TARGETS[@]}" \
   "${TEST_TARGETS[@]}"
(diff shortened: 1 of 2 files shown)


sampajano (approved):
thanks :)

tools/internal_ci/macos/grpc_objc_bazel_test.sh
```diff
@@ -100,7 +100,7 @@ objc_bazel_tests/bazel_wrapper \
   $BAZEL_FLAGS \
   --test_env HOST_PORT_LOCAL=localhost:$PLAIN_PORT \
   --test_env HOST_PORT_LOCALSSL=localhost:$TLS_PORT \
-  --test_env FLAKE_TEST_REPEATS=3 \
+  --test_env FLAKE_TEST_REPEATS=10 \
```
jtattermusch, line 103: 10 retries for a test seems crazy high. Why would you ever need so many retries for a test? If you really need that many retries it means that the test itself is probably garbage and adds almost no value.
We should focus on how to make individual invocations of the test more reliable.

jtattermusch: Let's revert this line.

jtattermusch:
@dennycd  I think these should be the next steps:
- revert the FLAKY_TEST_REPEATS line change as per my comment
- get https://github.com/grpc/grpc/pull/30869 merged first
- re-test this PR (Ideally do ~50 runs in an ephemeral kokoro build)
- if things are looking good, get this PR merged.

jtattermusch:
Let's merge https://github.com/grpc/grpc/pull/30938 instead.

PR #30379, opened 2022-07-21
Fix long http2 header values

ralphchung:
This PR aims to fix the bug [b/230976696](http://b/230976696).

The test case fails due to the assertion failure https://github.com/grpc/grpc/blob/0536a75167cd4df5298f788df4b640c9f791c973/src/core/ext/transport/chttp2/transport/hpack_encoder_table.cc#L28

Therefore, this PR prevents callers from making this assertion fail.


test/core/transport/chttp2/hpack_encoder_test.cc
```diff
@@ -219,6 +221,152 @@ TEST(HpackEncoderTest, TestBasicHeaders) {
   delete g_compressor;
 }
 
+grpc_slice EncodeHeaderIntoBytes(
```
ralphchung, line 224: This helper function mainly copies from https://github.com/grpc/grpc/blob/0536a75167cd4df5298f788df4b640c9f791c973/test/core/transport/chttp2/hpack_encoder_test.cc#L157-L158

I am not sure if I should reuse it or do what I have here.

ctiller: If we're introducing a new function that's largely the same, let's do the work to merge the two.

src/core/ext/transport/chttp2/transport/hpack_encoder.cc
```diff
@@ -431,13 +431,27 @@ void HPackCompressor::Framer::Encode(HttpSchemeMetadata,
...
+    EmitLitHdrWithBinaryStringKeyNotIdx(
+        Slice::FromStaticString(GrpcTraceBinMetadata::key()), slice.Ref());
+  }
   EncodeIndexedKeyWithBinaryValue(&compressor_->grpc_trace_bin_index_,
-                                  "grpc-trace-bin", slice.Ref());
+                                  GrpcTraceBinMetadata::key(), slice.Ref());
 }
 
 void HPackCompressor::Framer::Encode(GrpcTagsBinMetadata, const Slice& slice) {
+  const size_t table_entry_size = GrpcTagsBinMetadata::key().size() +
```
ctiller, line 446: Let's add a function (in hpack_constants.h maybe, next to EntriesForBytes)... something like SizeForEntry(size_t key_length, size_t value_length)

src/core/ext/transport/chttp2/transport/hpack_encoder.cc
```diff
@@ -431,13 +431,27 @@ void HPackCompressor::Framer::Encode(HttpSchemeMetadata,
...
+                                  slice.size() +
```
ctiller, line 447: Seems like a `HPackCompressor::Framer::EncodeRepeatingSliceValue(absl::string_view key, const Slice& slice)` would be useful here, so we could reduce these Encode methods to one liners.

(repeating or some such word to indicate that hpack compression should be employed if possible)

It might be worth adding a third argument of `max_compress_length` or some such so that we can control the maximum hpack table usage per header field - some fields we know are more important than others and we'd be more willing to take memory for them.

src/core/ext/transport/chttp2/transport/hpack_encoder_table.h
```diff
@@ -32,7 +34,10 @@ class HPackEncoderTable {
  public:
   HPackEncoderTable() : elem_size_(hpack_constants::kInitialTableEntries) {}
 
-  static constexpr size_t MaxEntrySize() { return 65535; }
+  using EntrySize = uint16_t;
```
ctiller, line 37: using decl should go before the constructor, at the top of the public section

ralphchung:
@ctiller Please have another review. Thanks!

test/core/transport/chttp2/hpack_encoder_test.cc
```diff
@@ -219,6 +237,117 @@ TEST(HpackEncoderTest, TestBasicHeaders) {
   delete g_compressor;
 }
 
+MATCHER(HasLiteralHeaderFieldNewNameFlagIncrementalIndexing, "") {
+  constexpr size_t kHttp2FrameHeaderSize = 9u;
+  constexpr uint8_t kLiteralHeaderFieldNewNameFlagIncrementalIndexing = 0x40;
```
Vignesh2208, line 242: Can you add a brief comment about why it needs to be 0x40

Vignesh2208:
Looks good. just one minor nit! Thanks

Vignesh2208 (changes requested):
Please comment

test/core/transport/chttp2/hpack_encoder_test.cc
ralphchung: I feel like the long variable name should be somehow self-explanatory. But it should be ok to add comments here.

ralphchung:
> Please comment

Done

src/core/ext/transport/chttp2/transport/hpack_encoder.cc
```diff
@@ -521,6 +521,18 @@ void HPackCompressor::Framer::EncodeIndexedKeyWithBinaryValue(
   }
 }
 
+void HPackCompressor::Framer::EncodeRepeatingSliceValue(
+    const absl::string_view& key, const Slice& slice, uint32_t* index,
+    std::size_t max_compression_size) {
```
ctiller, line 526: nit... we usually just write size_t

test/core/transport/chttp2/hpack_encoder_test.cc
```diff
@@ -152,43 +154,59 @@ static void CrashOnAppendError(absl::string_view, const grpc_core::Slice&) {
...
+
+  grpc_slice ret = grpc_slice_merge(output.slices, output.count);
   grpc_slice_buffer_destroy_internal(&output);
 
+  return ret;
+}
+
+/* verify that the output generated by encoding the stream matches the
+   hexstring passed in */
+static void verify(const verify_params params, const char* expected,
```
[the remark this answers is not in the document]
ctiller: Also, use_true_binary_metadata seems never to be used here, so maybe we could eliminate verify_params and just pass a `bool eof`

ctiller (approved):

Vignesh2208 (approved):

PR #30381, opened 2022-07-21
[objc] Better test reporting with current test case instance

dennycd:
Improve xctest reporting by passing in current test case instance to GRPCTestRunWithFlakeRepeats block. 

### Change Summary 

* Adding additional testCase parameter to repeat run block. The instance is made available to both testAssert and waiterBlock so that in scenarios of test failure, failure message logs are reported & associated with the current test case instance 

### Test & Verify 

build and test pass green for interop local, ssl and remote  

----

cc @HannahShiSFB


sampajano (approved):
Thanks :)

PR #30383, opened 2022-07-22
Forking tcp socket utils for posix event engine endpoints

Vignesh2208:
Forks required functions from: 

- src/core/lib/iomgr/socket_utils_posix{.h,.cc}
- src/core/lib/iomgr/socket_utils_common_posix.cc
- src/core/lib/address_utils/sockaddr_utils{.h,.cc}
- test/core/address_utils/sockaddr_utils_test.cc


src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
```diff
@@ -0,0 +1,851 @@
...
+namespace posix_engine {
+
+using ::grpc_event_engine::experimental::EndpointConfig;
+using ::grpc_event_engine::experimental::EventEngine;
+
+namespace {
+
+int AdjustValue(int default_value, int min_value, int max_value,
+                absl::optional<int> actual_value) {
+  if (!actual_value.has_value() || *actual_value < min_value ||
```
ctiller, line 59: Would this cause actual problems if we changed to Clamp(x.value_or(default_value), min_value, max_value)
-- I realize that we have default value out of range for many of the examples here, but let's audit to make sure that matters

src/core/lib/event_engine/posix_engine/tcp_socket_utils.h
```diff
@@ -0,0 +1,294 @@
...
+// Converts a EventEngine::ResolvedAddress into a newly-allocated human-readable
+// string.
+//
+// Currently, only the AF_INET, AF_INET6, and AF_UNIX families are recognized.
+// If the normalize flag is enabled, ::ffff:0.0.0.0/96 IPv6 addresses are
+// displayed as plain IPv4.
+absl::StatusOr<std::string> SockaddrToString(
+    const EventEngine::ResolvedAddress* resolved_addr, bool normalize);
+
+class PosixSocketWrapper {
```
ctiller, line 159: Should we get RCVLOWAT in here?

ctiller (approved):

src/core/lib/event_engine/posix_engine/tcp_socket_utils.h
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
Vignesh2208: I checked this. I don't think we should be using a Clamp here. Some parameters such as keep_alive_time_ms is set to 0  (default) to disable it if its not included in the EndpointConfig. Clamping this parameter would result in keep_alive_time_ms getting enabled and set a value of 1ms when that wasn't the original intention. The same is true for a few other parameters as well. For these parameters, a value of 0 is interpreted differently compared to positive values.

PR #30384, opened 2022-07-22
Add an option to rebuild core configuration in tests.

yousukseung:

### DIFF
--- a/src/core/lib/config/core_configuration.h
+++ b/src/core/lib/config/core_configuration.h
@@ -106,16 +106,19 @@ class CoreConfiguration {
     delete config_.exchange(p, std::memory_order_release);
   }
 
+  // Rebuild core configuration with currently registered builders.
+  // Tests can call this to restore core configuration before
+  // BuildSpecialConfiguration() was called.
+  static void RebuildConfiguration() {
+    delete config_.exchange(nullptr, std::memory_order_acquire);
+    BuildNewAndMaybeSet();
+  }
+
   // Attach a registration function globally.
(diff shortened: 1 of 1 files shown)


ctiller:
This seems likely to introduce cases whereby we leak configuration between tests and inadvertently create ordering issues.

yousukseung:
> 

Actually I'm adding this to fix an order-dependent test (b/238890770).
I updated a bit to add a function to rebuild, let me know if this one looks OK.

To provide some context, I think there's no way to restore core configuration back after calling BuildSpecialConfiguration(), and after my recent CL we now have a test file with tests that both require the original config and that use BuildSpecialConfiguration(), which is order-dependent.

ctiller:
We use CoreConfiguration::RunWithSpecialConfiguration to handle this
scenario in a few places.

yousukseung:
> RunWithSpecialConfiguration

Yes, I'm aware but not sure if that can help here, the test I am fixing already does what RunWithSpecialConfiguration() does (reset-build-test-reset).

In the above test file, there are tests that require builders that we register to ResolverRegistry during initialization, and other tests that override. It only passes if the latter ones run after the former in the same process, but this is not always guaranteed.

yousukseung:
Craig, I am instead adding a function that rebuilds with registered builders. Also removed CallRegisteredBuilders() that was declared but never defined, but if what I'm adding is what this was going to do I can rename the new function to this. Let me know.

yousukseung:
I found a fix that does not require this fix, closing.

PR #30389, opened 2022-07-22
Windows IOCP implementation for the WindowsEventEngine

drfloob:
This is a partial fork of the windows iomgr code - specifically the IOCP and Socket pieces - with some improved architecture and encapsulation. And the start of a WindowsEventEngine.

Once this code is used in a gRPC TCP context, I imagine a few issues will shake out. Also, getting sanitizers set up with MSVC will take a bit of work (see a commit referencing abseil and MSVC bugs to hack around).

I forked the IomgrEventEngine's posix poller interfaces in the hope of negotiating compatibility between the platforms, but the interfaces diverged a fair bit, and I'm doubtful we'll be able to use these "pollers" generically in the same TCP code. Reunification might not happen, and that's probably fine, we'll see how similar the TCP code looks once it's fleshed out.

I also extracted the IomgrEventEngine's timer piece into a separate component, usable by both engines.


src/core/ext/filters/load_reporting/server_load_reporting_filter.cc
```diff
@@ -136,7 +137,7 @@ std::string GetCensusSafeClientIpString(
     }
     return client_ip;
   } else {
-    GPR_UNREACHABLE_CODE();
+    GPR_UNREACHABLE_CODE(abort());
```
ctiller, line 140: curious as to what changed to need this

src/core/lib/event_engine/windows/event_poller.h
```diff
@@ -0,0 +1,100 @@
...
+  // Force set a writable event on the underlying socket.
+  virtual void SetWritable() = 0;
+  // Returns true if the WrappedSocket has been shutdown.
+  virtual bool IsShutdown() = 0;
+  virtual ~WrappedSocket() = default;
+};
+
+// A low-level poller that watches for socket events and dispatches
+// notifications.
+class EventPoller {
```
ctiller, line 67: Likely the split we want is:

class Poller {
 public:
  Shutdown(); 
  Work(Duration);
  Kick();
};

for posix: readable/writable notifications go into a PosixPoller type that we implement from each poller interface
for windows: IOCP just exports a Watch method, because we don't need the indirection

src/core/lib/event_engine/windows/iocp.cc
```diff
@@ -0,0 +1,146 @@
...
+        absl::StrFormat("IOCP::%p: Received no completions", this));
+  }
+  GPR_ASSERT(completion_key && overlapped);
+  if (overlapped == &kick_overlap_) {
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_event_engine_trace)) {
+      gpr_log(GPR_DEBUG, "IOCP::%p kicked", this);
+    }
+    outstanding_kicks_.fetch_sub(1);
+    if (completion_key == (ULONG_PTR)&kick_token_) {
+      return absl::AbortedError(
```
ctiller, line 92: note that kicks are going to end up being on the critical path, so we'll probably want to return without allocating here

src/core/lib/event_engine/windows/iocp.cc
```diff
@@ -0,0 +1,146 @@
...
+  BOOL success = GetQueuedCompletionStatus(
+      iocp_handle_, &bytes, &completion_key, &overlapped,
+      static_cast<DWORD>(
+          std::chrono::duration_cast<std::chrono::milliseconds>(timeout)
+              .count()));
+  if (success == 0 && overlapped == NULL) {
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_event_engine_trace)) {
+      gpr_log(GPR_DEBUG, "IOCP::%p deadline exceeded", this);
+    }
+    return absl::DeadlineExceededError(
```
ctiller, line 82: note that deadline exceeded will eventually happen for every timer expiry, so we probably don't want to be allocating here

src/core/ext/filters/load_reporting/server_load_reporting_filter.cc
drfloob: I haven't investigated what's different from our CI per se, but this is what I get building master with MSVC 14.32 (Visual Studio 2022).

```
ERROR: C:/users/hork/proj/grpc/BUILD:4534:16: Compiling src/core/ext/filters/load_reporting/server_load_reporting_filter.cc failed: (Exit 2): cl.exe failed: error executing command C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.32.31326\bin\HostX64\x64\cl.exe /nologo /DCOMPILER_MSVC /DNOMINMAX /D_WIN32_WINNT=0x0601 /D_CRT_SECURE_NO_DEPRECATE ... (remaining 43 arguments skipped)
src/core/ext/filters/load_reporting/server_load_reporting_filter.cc(139): warning C4003: not enough arguments for function-like macro invocation 'GPR_UNREACHABLE_CODE'
src/core/ext/filters/load_reporting/server_load_reporting_filter.cc(139): error C2143: syntax error: missing ')' before ','
src/core/ext/filters/load_reporting/server_load_reporting_filter.cc(139): error C2660: 'gpr_unreachable_code': function does not take 0 arguments
```

I think it _should_ be failing today. Not sure why it isn't.

src/core/lib/event_engine/windows/iocp.cc
drfloob: Good catch. This now returns a (copy of a) static const status.

src/core/lib/event_engine/windows/iocp.cc
drfloob: Good catch. This now returns a (copy of a) static const status.

src/core/lib/event_engine/windows/event_poller.h
drfloob: Partially done. I broke out the interface for the Windows code, but will do the Posix pollers after the poll fork PR lands

drfloob: Resolved in https://github.com/grpc/grpc/pull/30413

src/core/lib/event_engine/default_event_engine_factory.cc
```diff
@@ -31,3 +46,5 @@ std::unique_ptr<EventEngine> DefaultEventEngineFactory() {
```
ctiller, line 46: Let's rename `IomgrEventEngine` to `PosixEventEngine`

src/core/lib/event_engine/iomgr_engine/posix_poller.h
```diff
@@ -0,0 +1,40 @@
...
+#include "src/core/lib/event_engine/poller.h"
+
+namespace grpc_event_engine {
+namespace iomgr_engine {
+
+// Interface for posix pollers
+class PosixPoller : public experimental::Poller {
+ public:
+  virtual ~Poller() = default;
+  virtual absl::Status Work(grpc_core::Duration timeout) = 0;
```
ctiller, line 28: Remove this and Kick, they're from `experimental::Poller`

src/core/lib/event_engine/windows/iocp.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include "src/core/lib/event_engine/poller.h"
+#include "src/core/lib/event_engine/windows/socket.h"
+
+namespace grpc_event_engine {
+namespace experimental {
+
+class IOCP final : public Poller {
+ public:
+  explicit IOCP(EventEngine* event_engine) noexcept;
```
ctiller, line 33: Having to pass an entire `EventEngine` in here makes it harder to test this in isolation. Is there a more restrictive interface we can pass here, and offer a fake in tests?

test/core/event_engine/windows/iocp_test.cc
```diff
@@ -0,0 +1,379 @@
...
+using ::grpc_event_engine::experimental::Poller;
+using ::grpc_event_engine::experimental::SelfDeletingClosure;
+using ::grpc_event_engine::experimental::WindowsEventEngine;
+using ::grpc_event_engine::experimental::WinSocket;
+}  // namespace
+
+class IOCPTest : public testing::Test {};
+
+TEST_F(IOCPTest, ClientReceivesNotificationOfServerSend) {
+  auto engine = absl::make_unique<WindowsEventEngine>();
```
ctiller, line 54: I'd like to see this test not reference `WindowsEventEngine` - as it stands if there's a bug I must debug the entirety of the `WindowsEventEngine` system, rather than just the slice that I might suspect to be faulty.

test/core/event_engine/windows/self_deleting_closure.h
```diff
@@ -0,0 +1,45 @@
...
+#include <grpc/support/port_platform.h>
+
+#include "absl/functional/any_invocable.h"
+
+#include <grpc/event_engine/event_engine.h>
+
+namespace grpc_event_engine {
+namespace experimental {
+
+class SelfDeletingClosure : public EventEngine::Closure {
```
ctiller, line 26: This is probably something we want in test/core/event_engine/util?

src/core/lib/event_engine/windows/socket.h
```diff
@@ -0,0 +1,115 @@
...
+  // Accessor method for underlying socket
+  SOCKET socket();
+
+ private:
+  void NotifyOnReady(OpState& info, EventEngine::Closure* callback);
+
+  SOCKET socket_;
+  grpc_core::Mutex mu_;
+  bool is_shutdown_ ABSL_GUARDED_BY(mu_) = false;
+  EventEngine* event_engine_;
```
ctiller, line 102: Is there a more restrictive interface we could offer here?

src/core/lib/event_engine/windows/socket.h
```diff
@@ -0,0 +1,115 @@
+// Copyright 2022 gRPC authors.
```
ctiller, line 1: suggest renaming to win_socket.{h,cc} to match the class name

src/core/lib/event_engine/windows/socket.h
```diff
@@ -0,0 +1,115 @@
...
+
+#include <grpc/event_engine/event_engine.h>
+
+#include "src/core/lib/event_engine/socket_notifier.h"
+#include "src/core/lib/gprpp/sync.h"
+
+namespace grpc_event_engine {
+namespace experimental {
+
+class WinSocket final : public SocketNotifier {
```
ctiller, line 32: Can we get an isolated test for this type?

ctiller (changes requested):

src/core/lib/event_engine/default_event_engine_factory.cc
drfloob: Done

src/core/lib/event_engine/iomgr_engine/posix_poller.h
drfloob: I've removed the file, it was speculative and Vignesh is working on the Posix version at the moment.

test/core/event_engine/windows/self_deleting_closure.h
drfloob: Migrated to test/core/event_engine/common_closures.h and added tests (now with more custom destructors!)

src/core/lib/event_engine/windows/socket.h
drfloob: Those are in `test/core/event_engine/windows/win_socket_test`.

src/core/lib/event_engine/windows/socket.h
drfloob: Done

src/core/lib/event_engine/windows/socket.h
drfloob: Yes, as we discussed, I was considering a subset of the EventEngine API, something like 

```
class Executor {
  virtual ~Executor() = default;
  virtual void Run(Closure* closure) = 0;
  virtual void Run(absl::AnyInvocable<void()> closure) = 0;  
};
```

The idea would be, though, that some EventEngine implementation would be plugged into it. i.e., `class MyEventEngine : public EventEngine, Executor { ... };`. So it doesn't make much difference until there are alternative implementations to plug into it.

drfloob: I take that back, I'll extract the common Executor functionality and keep it separate, and have the engines take ownership of one.

src/core/lib/event_engine/windows/iocp.h
drfloob: I extracted an Executor interface, it's a much simpler thing now.

test/core/event_engine/windows/iocp_test.cc
drfloob: Done, this now relies on a ThreadedExecutor (identical to the one used inside the WindowsEventEngine, but it does not have to be)

src/core/lib/event_engine/windows/socket.h
drfloob: Done.

src/core/lib/event_engine/windows/win_socket.cc
```diff
@@ -0,0 +1,201 @@
...
+}
+
+void WinSocket::NotifyOnReady(OpState& info, EventEngine::Closure* closure) {
+  grpc_core::MutexLock lock(&mu_);
+  if (IsShutdown()) {
+    info.SetError(WSAESHUTDOWN);
+    executor_->Run(closure);
+    return;
+  };
+  if (info.has_pending_iocp_) {
```
ctiller, line 88: nowadays I'm getting into the habit of writing `if (absl::exchange(info.has_pending_iocp_, false)) {` for this pattern to more clearly & concisely label that it's a state edge

ctiller (approved):

src/core/lib/event_engine/windows/win_socket.cc
drfloob: Done.

PR #30390, opened 2022-07-22
Adding methods to get memory usage of server for the perchannel benchmark

nancylucy01:
Added two functions related to getting server memory usage: 

GetBeforeSnapshot for getting the memory usage of the server process before the server is created. This is an RPC since the server program itself has to record it's initial memory usage as the client process starts up after the server process. The RPC then allows the client process to get this number from the server process.

GetMemUsage is replacing getrusage within memstat.h as getrusage only allows us to get the memory usage of the calling process or it's children. GetMemUsage allows us to get the memory usage of any process as long as we know the pid. We switched it since it allows us to get the peak server process without sending an RPC to the server. Sending an RPC to the server asking for it's memory usage was a concern since when resource quota is an option later, the server may be overwhelmed by channels/calls and won't be able to respond to a RPC request for it's memory usage. GetMemUsage allows us to get the server process's memory within the client process using pid, therefore avoiding this issue.

Tested by: Checking if previous method of Snapshot to get Memstat, and GetMemUsage returned the same number for memory usage.


ctiller (commented):
Can we move the existing call profiler to use this mechanism to get server memory usage too, just to prove that it's consistent and working as we complete the rest of the code?

test/core/memory_usage/memstats.h
```diff
@@ -18,15 +18,53 @@
...
+  stat_stream >> temp_pid >> comm >> state >> ppid >> pgrp >> session >>
+      tty_nr >> tpgid >> flags >> minflt >> cminflt >> majflt >> cmajflt >>
+      utime >> stime >> cutime >> cstime >> priority >> nice >> O >>
+      itrealvalue >> starttime >> vsize >> rss;
+  stat_stream.close();
+
+  // pid does not connect to an existing process
+  if (state == "") {
+    printf("PID does not exist");
+    return 0;
```
nancylucy01, line 56: Is returning 0 enough? The previous code aborted when getrusage wasn't a success, but I figured the memory usage of a non-existing process is technically 0.

ctiller: we should think about how this will be used: it's true that a non-existing process uses no memory, but it's also true that if we can't find a PID being used for a test, that test is probably busted. I'd `GPR_ASSERT(state != "");`

test/core/memory_usage/memstats.h
```diff
@@ -18,15 +18,53 @@
...
+#include <fstream>
+#include <iostream>
+
+#include "absl/strings/str_cat.h"
+
 // IWYU pragma: no_include <bits/types/struct_rusage.h>
 
+// Get the memory usage of either the calling process or another process using
+// the pid
+long GetMemUsage(absl::optional<int> pid = absl::nullopt) {
```
ctiller, line 30: free functions in headers need to be inline or they'll produce ODR violations, however this function is probably larger than I'd like to see inlined... soo let's leave a function prototype here and move the implementation into memstats.cc.

ctiller (approved):

ctiller (changes requested):
sorry meant to request changes a moment ago (see comments on the code)

test/core/memory_usage/memstats.h
nancylucy01: Makes sense, changed it

test/core/memory_usage/memstats.h
nancylucy01: I see, moved the implementation into memstats.cc

ctiller (approved):
lgtm assuming tests pass

PR #28873, opened 2022-02-14
Fix fetching CXX variable in setup.py

kacper-ka:
While building the *Python* package for a *poky*-based custom distribution I have stumbled upon an issue caused by the mechanism of subprocess invocation.

`setup.py` in `check_linker_need_libatomic` function verifies whether inclusion of `atomic` library is necessary. It is done by invoking a C++ compiler in a subprocess and the compiler's name is read from `CXX` environmental variable. In a *bitbake* managed build context this variable holds not only compiler's executable but various compiler flags as well, in my case the contents of this variable are:
```
arm-poky-linux-gnueabi-g++ -mthumb -mfpu=neon -mfloat-abi=hard -mcpu=cortex-a7 -fstack-protector-strong -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security -Wdate-time --sysroot=<redacted>/build/tmp/work/cortexa7t2hf-neon-poky-linux-gnueabi/grpcio/1.43.0-r0/recipe-sysroot
```

Proposed solution of the issue is to split the contents of the variable and take only the first element of the resulting list. Possibly it would be better to preserve all the flags though I don't think it is necessary as we only need to know whether `-latomic` switch is required (I don't think the outcome depends on CPU model, instruction set or floating-point ABI).

This PR resolves #25362

PR #30076, opened 2022-06-21
Use cpp consistently in check_linker_need_libatomic

fedej:
`check_linker_need_libatomic` will always fail its second check, because `cxx` is now a list. This affects piwheel builds since 1.46.0. 
```
2022-05-20T21:01:38,827   Running command python setup.py egg_info
2022-05-20T21:01:42,599   Traceback (most recent call last):
2022-05-20T21:01:42,601     File "<string>", line 2, in <module>
2022-05-20T21:01:42,601     File "<pip-setuptools-caller>", line 34, in <module>
2022-05-20T21:01:42,602     File "/tmp/pip-wheel-6bjraj6j/grpcio_27b0a215ff714c6f93177ad66c3f3c03/setup.py", line 257, in <module>
2022-05-20T21:01:42,602       if check_linker_need_libatomic():
2022-05-20T21:01:42,603     File "/tmp/pip-wheel-6bjraj6j/grpcio_27b0a215ff714c6f93177ad66c3f3c03/setup.py", line 213, in check_linker_need_libatomic
2022-05-20T21:01:42,604       cpp_test = subprocess.Popen(
2022-05-20T21:01:42,604     File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
2022-05-20T21:01:42,606       self._execute_child(args, executable, preexec_fn, close_fds,
2022-05-20T21:01:42,607     File "/usr/lib/python3.9/subprocess.py", line 1698, in _execute_child
2022-05-20T21:01:42,609       and os.path.dirname(executable)
2022-05-20T21:01:42,610     File "/usr/lib/python3.9/posixpath.py", line 152, in dirname
2022-05-20T21:01:42,628       p = os.fspath(p)
2022-05-20T21:01:42,629   TypeError: expected str, bytes or os.PathLike object, not list
2022-05-20T21:01:42,824   ERROR: [present-rich] python setup.py egg_info exited with 1
2022-05-20T21:01:42,863   [bold magenta]full command[/]: [blue]/usr/bin/python3 -c '
```

[Full log](https://www.piwheels.org/logs/0000/0826/8202.txt)

PR #30401, opened 2022-07-24
Fix usage of `cxx` variable in libatomic check

lopsided98:
#28873 made the `cxx` variable into a list, but only updated the first usage. This causes build failures on armv6l, which requires libatomic.

cc @lidizheng @kacper-ka


fedej:
I think [this PR](https://github.com/grpc/grpc/pull/30076)  is solving the same issue

PR #30404, opened 2022-07-25
Fix possible integer overflow in `to_seconds_from_sub_second_time()`

ralphchung:
This PR aims to fix [b/231828673](http://b/231828673).

It is possible to experience integer overflow when the input is small (negatively). Actually unit test has test cases for that. In that case, integer overflow happens, but the behavior is still good. In other words, integer overflow does not cause problems.

This PR tries to deal with integer overflow with simpler implementation.

### DIFF
--- a/src/core/lib/gpr/time.cc
+++ b/src/core/lib/gpr/time.cc
@@ -77,16 +77,19 @@ static gpr_timespec to_seconds_from_sub_second_time(int64_t time_in_units,
   } else if (time_in_units == INT64_MIN) {
     out = gpr_inf_past(type);
   } else {
-    if (time_in_units >= 0) {
-      out.tv_sec = time_in_units / units_per_sec;
-    } else {
-      out.tv_sec = (-((units_per_sec - 1) - (time_in_units + units_per_sec)) /
-                    units_per_sec) -
-                   1;
-    }
+    out.tv_sec = time_in_units / units_per_sec;
+    /// Our current usage does not seem to cause precision issues. Division for
+    /// integers should be enough instead of converting them to doubles for
+    /// better precision.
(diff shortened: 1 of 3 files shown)


src/core/lib/gpr/time.cc
```diff
@@ -77,16 +77,19 @@ static gpr_timespec to_seconds_from_sub_second_time(int64_t time_in_units,
...
   } else {
-    if (time_in_units >= 0) {
-      out.tv_sec = time_in_units / units_per_sec;
-    } else {
-      out.tv_sec = (-((units_per_sec - 1) - (time_in_units + units_per_sec)) /
-                    units_per_sec) -
-                   1;
-    }
+    out.tv_sec = time_in_units / units_per_sec;
+    /// Our current usage does not seem to cause precision issues. Division for
```
ctiller, line 81: I don't understand why this comment was added

ctiller (changes requested):
Can you go ahead and add the relevant fuzzing example to the corpus?

src/core/lib/gpr/time.cc
ralphchung: My understanding is we should treat every number in the chain of operations `(time_in_units - (out.tv_sec * units_per_sec)) * (GPR_NS_PER_SEC / units_per_sec)` as floating numbers to avoid losing precision from division with remainders. I did not do that since current usage does not seem to cause that.

ctiller: Oh I think we're ok here because `GPR_NS_PER_SEC % units_per_sec == 0` always... let's drop in a `GPR_DEBUG_ASSERT` to verify that.

ctiller (approved):

Issue #25950, opened 2021-04-11
Grpc.Tools is not passing OutputOptions or AdditionalProtocArguments to protoc.exe


StingyJack:
### What version of gRPC and what language are you using?
Grpc.Tools 2.36.4

### What operating system (Linux, Windows,...) and version?
Windows 10 + Linux (current, idk)

### What runtime / compiler are you using (e.g. python version or version of gcc)
Whatever Grpc.Tools uses

### What did you do?
I have included 3 proto files in a VS project. Two are compiled and are simple message payloads. The third is linked from here and compiled by another proj file. 
``` xml
 <ItemGroup>
    <Protobuf Include="Protos\example_data_payloads.proto" GrpcServices="None" ProtoCompile="true" OutputOptions="csharp_opt=file_extension=.g.cs"/>
    <Protobuf Include="Protos\example_data_schema.proto" GrpcServices="None" ProtoCompile="true" OutputOptions="file_extension=.g.cs"/>
    <Protobuf Include="Protos\example_data_stack.proto" GrpcServices="None" ProtoCompile="false" />
  </ItemGroup>
```
For the two that are proto compiled in this project, I am trying to fix the file extension to note that the file has been code generated so it can be excluded from test coverage reports. OutputOptions is described [here](https://github.com/grpc/grpc/blob/master/src/csharp/BUILD-INTEGRATION.md) and I've tried a few variations of it but the file is still generated as a .cs file. 

### What did you expect to see?
the code file generated with a .g.cs extension

### What did you see instead?
just the file with a .cs extension in the obj/ directory

The proto file for the first item is 
``` protobuf
syntax = "proto3";
option csharp_namespace = "Core.Schema";
package example_data_schema;

message ExampleData {
	bytes id = 1;
	string name  = 3;
	int64 code = 5;
	string description = 7;
}
```
the code is otherwise generated correctly.  I would really prefer the exclusion mechanism to be part of the compiled output, like a [GeneratedCode] attribute, since analyzers may be working from the compiled output and not using source analysis.  The current conventions used by the proto generation are not any of the common ways to designate code thats been generated.

jtattermusch:
related https://github.com/grpc/grpc/issues/25930

StingyJack:
Neither `AdditionalProtocArguments` or `OutputOptions` are being passed along to protoc.exe. 

I updated my proj file to 
``` xml
[...]

PR #30410, opened 2022-07-26
Fix OutputOptions and GrpcOutputOptions (issue #25950)

tonydnewell:
Fix for OutputOptions and GrpcOutputOptions being ignored - issue https://github.com/grpc/grpc/issues/25950

The Grpc.Tools msbuild XML needed updating to include using the user-supplied options.


jtattermusch:
CC @JamesNK

jtattermusch (approved):
LGTM, thanks for the fix.

We should think about how to improve the current test suite for Grpc.Tools so that we can reliably test changes like this and we don't have to rely on manual testing (which is unrealiable and timeconsuming).

jtattermusch:
CC @apolcyn

PR #30413, opened 2022-07-26
Introduce the EventEngine internal Poller interface

drfloob:
Introducing it first allows us to develop against it in parallel.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -2242,8 +2242,12 @@ grpc_cc_library(
     ],
     external_deps = [
         "absl/container:inlined_vector",
+        "absl/status:statusor",
     ],
-    deps = [],
+    deps = [
+        "gpr_platform",
+        "grpc_base",
+    ],
 )
 
 grpc_cc_library(
--- a/BUILD
+++ b/BUILD
@@ -2236,6 +2236,17 @@ grpc_cc_library(
 )
 
 grpc_cc_library(
+    name = "event_engine_poller",
+    hdrs = [
+        "src/core/lib/event_engine/poller.h",
+    ],
+    external_deps = [
+        "absl/container:inlined_vector",
+    ],
+    deps = [],
+)
+
+grpc_cc_library(
     name = "iomgr_ee_timer",
     srcs = [
         "src/core/lib/event_engine/iomgr_engine/timer.cc",
--- a/BUILD
+++ b/BUILD
@@ -3093,6 +3093,8 @@ grpc_cc_library(
     ],
     external_deps = [
         "absl/memory",
+        "absl/status",
+        "absl/status:statusor",
         "absl/strings",
     ],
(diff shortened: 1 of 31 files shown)


BUILD
```diff
@@ -2235,6 +2235,17 @@ grpc_cc_library(
     deps = ["gpr_base"],
 )
 
+grpc_cc_library(
+    name = "event_engine_poller",
+    hdrs = [
+        "src/core/lib/event_engine/poller.h",
+    ],
+    external_deps = [
+        "absl/container:inlined_vector",
```
Vignesh2208, line 2244: I think absl statusor should  also be added as a dependency

drfloob: Fixed.

Vignesh2208 (approved):

src/core/lib/event_engine/poller.h
```diff
@@ -0,0 +1,53 @@
...
+#include "absl/container/inlined_vector.h"
+#include "absl/status/statusor.h"
+
+#include <grpc/event_engine/event_engine.h>
+
+namespace grpc_event_engine {
+namespace experimental {
+
+namespace {
+constexpr size_t kInitialSize = 5;
```
ctiller, line 28: move this into `class Poller` - generally we don't want anonymous namespaces in headers anyway

src/core/lib/event_engine/poller.h
```diff
@@ -0,0 +1,53 @@
...
+ public:
+  using Events = absl::InlinedVector<EventEngine::Closure*, kInitialSize>;
+  virtual ~Poller() = default;
+  // Poll once for events, returning a collection of Closures to be executed.
+  //
+  // Returns:
+  //  * absl::AbortedError if it was Kicked.
+  //  * absl::DeadlineExceeded if timeout occurred
+  //  * A collection of closures to execute, otherwise
+  virtual absl::StatusOr<Events> Work(EventEngine::Duration timeout) = 0;
```
ctiller, line 45: AbortedError and DeadlineExceeded both require a string which must be copied - and consequently an allocation made to hold that string. Since this is fast path and both kicks and deadline exceeded are on that fast path, let's instead define this as:

```
using Events = ...;
struct DeadlineExceeded {};
struct Kicked {};
using WorkResult = absl::variant<Events, DeadlineExceeded, Kicked>;
virtual WorkResult Work(Duration timeout) = 0;
```

drfloob: Done

src/core/lib/event_engine/poller.h
drfloob: Done. I left a note about maybe needing to tune this eventually, 5 is arbitrary. For IOCP, I expect a vector of size 1.

ctiller (approved):

PR #30414, opened 2022-07-26
Remove `g_glip` and `class GrpcLibraryInterface`

ralphchung:

This PR aims to remove the pointer `g_glip` and `class GrpcLibraryInterface`.

Details:
* `grpc::GrpcLibraryCodegen` is moved to `grpc::GrpcLibrary` to replace the need for pointer `g_glip`.
* `grpc::GrpcLibraryInterface` is removed.


include/grpcpp/grpc_library.h
```diff
@@ -16,13 +16,24 @@
...
-#endif  // GRPCXX_IMPL_CODEGEN_GRPC_LIBRARY_H
+/// Classes that require gRPC to be initialized should inherit from this class.
+class GrpcLibrary {
+ public:
+  explicit GrpcLibrary(bool call_grpc_init = true);
+  virtual ~GrpcLibrary();
+
+ private:
+  bool grpc_init_called_;
+  static internal::GrpcLibrary internal_grpc_library;
```
Vignesh2208, line 34: Why is there a static grpc::internal::GrpcLibrary ? It seems like its methods only call grpc_init() and grpc_shutdown(). Why not call them directly in grpc::GrpcLibrary constructor and destructor respectively ?

Vignesh2208 (changes requested):
Looks mostly good to me.

include/grpcpp/grpc_library.h
```diff
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright 2018 gRPC authors.
+ * Copyright 2022 gRPC authors.
```
Vignesh2208, line 3: Please switch to newer style C++ header/copyright comments in this the file (e.g, src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc)

ctiller:
Just noting here: this will need an internal cherry-pick, as there are internal usages of this header.

include/grpcpp/grpc_library.h
ralphchung: Done

ctiller (approved):

BUILD
```diff
@@ -6700,6 +6699,7 @@ grpc_cc_library(
     tags = ["nofixdeps"],
     visibility = ["@grpc:public"],
     deps = [
+        "grpc++_base",
```
ralphchung, line 6702: I think this is a better place to bring the dependency given that we are removing `grpc++_codegen_base` and `grpc++_codegen_base_src` in the future.

Also, this change makes the blaze rule in Google internal not need to add the dependency.

Vignesh2208 (approved):

Vignesh2208 (changes requested):
Actually there seem to be some bazel ASAN failures. Can you verify ?

include/grpcpp/grpc_library.h
```diff
@@ -0,0 +1,43 @@
+// Copyright 2022 The gRPC Authors
```
ctiller, line 1: Lets move this to impl/... I don't think it should be public api

include/grpcpp/impl/grpc_library.h
```diff
@@ -24,25 +24,34 @@
...
-class GrpcLibrary final : public GrpcLibraryInterface {
+
+/// Classes that require gRPC to be initialized should inherit from this class.
+class GrpcLibrary {
  public:
-  void init() override { grpc_init(); }
-  void shutdown() override { grpc_shutdown(); }
+  explicit GrpcLibrary(bool call_grpc_init = true) : grpc_init_called_(false) {
+    if (call_grpc_init) {
+      grpc_init();
```
ralphchung, line 37: It seems calling `grpc_core::initInternally()` here might experience race condition. @ctiller do you have any suggestion on this?

ralphchung: @ctiller suggested use `grpc_init()` instead.

Vignesh2208 (approved):

PR #30417, opened 2022-07-26
Revert "Remove empty function grpc_cq_global_init and grpc_executor_global_init"

markdroth:
Reverts grpc/grpc#30370 due to internal breakage.

### DIFF
--- a/src/core/lib/debug/stats.cc
+++ b/src/core/lib/debug/stats.cc
@@ -31,14 +31,18 @@
 
 #include <grpc/support/alloc.h>
 #include <grpc/support/cpu.h>
+#include <grpc/support/sync.h>
 
 grpc_stats_data* grpc_stats_per_cpu_storage = nullptr;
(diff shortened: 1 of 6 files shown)


ctiller (approved):

panzhongxian:
Hi. @markdroth I'm sorry that the PR caused breakage.
Could you show me more info about the internal breakage? Are there some repositories have been calling the empty functions? Or the changes is incorrect itself?

ctiller:
Looked into this a little... it looks like we have some direct callers of grpc_stats_init internally which is kind of yuck.
I've got a good handle on how to fix that though, so I'll likely post a relanding in a few minutes.

panzhongxian:
> Looked into this a little... it looks like we have some direct callers of grpc_stats_init internally which is kind of yuck. I've got a good handle on how to fix that though, so I'll likely post a relanding in a few minutes.

Got it. Thanks.

PR #30418, opened 2022-07-27
outlier_detection LB: reset state when OD is disabled

markdroth:
Updates the OD policy as per https://github.com/grpc/proposal/pull/315.

Also fixes the xds_cluster_resolver policy to not recreate itself if only OD fields change in its config, and fixes a bug where the CDS update was incorrectly failing to check the OD fields to detect a new config.

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/outlier_detection/outlier_detection.h
+++ b/src/core/ext/filters/client_channel/lb_policy/outlier_detection/outlier_detection.h
@@ -30,7 +30,7 @@ namespace grpc_core {
 bool XdsOutlierDetectionEnabled();
 
 struct OutlierDetectionConfig {
-  Duration interval = Duration::Infinity();
+  Duration interval = Duration::Seconds(10);
   Duration base_ejection_time = Duration::Milliseconds(30000);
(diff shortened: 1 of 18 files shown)


src/core/ext/filters/client_channel/lb_policy/outlier_detection/outlier_detection.h
```diff
@@ -30,7 +30,7 @@ namespace grpc_core {
 bool XdsOutlierDetectionEnabled();
 
 struct OutlierDetectionConfig {
-  Duration interval = Duration::Infinity();
+  Duration interval = Duration::Seconds(10);
```
yashykt, line 33: Was this a bug? Is the default 10 seconds?

markdroth: Yeah, the default is supposed to be 10s.  We were failing to default properly, and there was some logic previously to disable the policy if the interval was infinite, which we were triggering via this default.

yashykt:
looks mostly good to me. I'm gonna take another pass at the proposal and the code to see if I spot anything.

yashykt (approved):

PR #30419, opened 2022-07-27
Migrating posix event pollers to use new event poller interface

Vignesh2208:
The PR also improves testing of the epoll1 and poll pollers and fixes some bugs found with the improved tests.


src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
```diff
@@ -487,24 +506,26 @@ void Epoll1EventHandle::SetWritable() { write_closure_->SetReady(); }
...
-absl::Status Epoll1Poller::Work(grpc_core::Timestamp deadline,
-                                std::vector<EventHandle*>& pending_events) {
+Poller::WorkResult Epoll1Poller::Work(EventEngine::Duration timeout) {
+  Poller::Events pending_events;
+  grpc_core::Timestamp deadline = ToTimestamp(timeout);
   if (g_epoll_set_.cursor == g_epoll_set_.num_events) {
-    auto status = DoEpollWait(deadline);
-    if (!status.ok()) {
-      return status;
+    if (DoEpollWait(deadline) == 0) {
```
drfloob, line 513: This converts a timeout to a deadline, then `DoEpollWait` converts the deadline back to a timeout. I suggest changing the signature to `DoEpollWait(EventEngine::Duration timeout)`.

Please review and use this: https://github.com/grpc/grpc/pull/30427. I forked it off from my IOCP poller branch, I think it will work nicely here.

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
```diff
@@ -206,6 +215,14 @@ void ForkPollerListRemovePoller(Epoll1Poller* poller) {
   }
 }
 
+grpc_core::Timestamp ToTimestamp(EventEngine::Duration when) {
```
drfloob, line 218: I believe this can be deleted now. And if it's necessary elsewhere, I suggest moving it to the time_util.

src/core/lib/event_engine/iomgr_engine/iomgr_engine_closure.h
```diff
@@ -43,14 +44,39 @@ class IomgrEngineClosure final
...
+      Ref();
+    }
     cb_(absl::exchange(status_, absl::OkStatus()));
-    if (!is_permanent_) {
+    // For the ref taken at the beginning of this function. If it is a temporary
+    // closure, it will get deleted immediately.
+    Unref();
+  }
+
+  // Ref/Unref methods should only be called on permanent closures.
```
drfloob, line 58: I don't see why this is necessary yet. We chatted a bit offline, but I'm not convinced this needs a ref-counted closure type (and the overhead, both in client usage complexity and added atomic operations). Once in context of TCP code it might be more obvious, but I don't see it yet. I think the caller will know when its connections are closed, and can safely delete its permanent closures.

That said, maybe @ctiller has a better picture.

ctiller: i'd move the ref counting into the thing that owns this instance if refcounting is needed

ctiller: (but its unclear to me why this is needed)

Vignesh2208: when moving to the vector<Closure*> return type for Work(...), the idea is to have a permanent closure associated with each EventHandle. A vector of such permanent closures will be returned from Work(..). Running this closures will execute any pending actions on the corresponding handles. But running these closures may trigger callbacks which orphan the handles that own them. When the owning handles are orphaned, they delete these closures and this can happen while the closure is still being run causing a use-after-free bug. The Ref/Unref methods help prevent this.

ctiller: right, but can't that ref be held by the containing type...

when we start:
void ContainingType::ScheduleThing() { Ref(); ScheduleClosure(&closure_); }
void SpecialClosure::Run() { containing_type_->Unref(); }

Vignesh2208: Sure it would have to be a new Closure type I think. I dont see how we can safely do containing_type_->Unref() inside the any-invocable (cb_) executed inside IomgrEngineClosure::Run(). That might end up deleting the containing_type_ while IomgrEngineClosure::Run() is still executing. Is it okay if I create new closure types internal to epoll1 and poll poller ?

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
Vignesh2208: Done.

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
Vignesh2208: Deleted this. I didn't need to use it elsewhere so its not added to time_util

src/core/lib/event_engine/iomgr_engine/iomgr_engine_closure.h
Vignesh2208: Updated the PR to use special closure types for epoll1 and poll pollers. Removed Ref/Unref methods from IomgrEngineClosure

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
```diff
@@ -77,17 +91,24 @@ class Epoll1EventHandle : public EventHandle {
     write_closure_->InitEvent();
     error_closure_->InitEvent();
     pending_actions_ = 0;
+    exec_actions_closure_ = std::make_unique<Epoll1ExecActionsClosure>(this);
```
drfloob, line 94: I don't think the custom Closure type is necessary, nor the forward declaration / circular dependency. If you use something like a basic lambda-wrapped closure, which we've written a few times now, I think this gets simpler. For example:

```
class Epoll1EventHandle {
...
 private:
  AnyInvocableClosure exec_actions_closure_;
};
```

Then in your constructor
`Epoll1EventHandle(...) : exec_actions_closure_([this] { ExecutePendingActions(); Unref(); } )`

Or

`Epoll1EventHandle(...) : exec_actions_closure_([this] { DoCallback(); } )`

I extracted some common, tested closure types I've hacked together a few times (including in my poller) and added them here https://github.com/grpc/grpc/pull/30438. Please review, and we can both use them.

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
```diff
@@ -51,8 +50,9 @@
 #include "src/core/lib/event_engine/iomgr_engine/wakeup_fd_posix.h"
 #include "src/core/lib/event_engine/iomgr_engine/wakeup_fd_posix_default.h"
 #include "src/core/lib/gprpp/fork.h"
-#include "src/core/lib/gprpp/time.h"
 
+using ::grpc_event_engine::experimental::EventEngine;
```
drfloob, line 54: Using statements should be inside the inner-most namespace.

src/core/lib/event_engine/iomgr_engine/ev_poll_posix.cc
```diff
@@ -71,6 +73,19 @@ static const int kPolloutCheck = POLLOUT | POLLHUP | POLLERR;
 namespace grpc_event_engine {
 namespace iomgr_engine {
 
+class PollEventHandle;
```
drfloob, line 76: Same suggestion here about simpler closure relationships.

src/core/lib/event_engine/iomgr_engine/ev_poll_posix.h
```diff
@@ -42,8 +42,8 @@ class PollPoller : public EventPoller {
   PollPoller(Scheduler* scheduler, bool use_phony_poll);
   EventHandle* CreateHandle(int fd, absl::string_view name,
                             bool track_err) override;
-  absl::Status Work(grpc_core::Timestamp deadline,
-                    std::vector<EventHandle*>& pending_events) override;
+  Poller::WorkResult Work(
+      grpc_event_engine::experimental::EventEngine::Duration timeout) override;
```
drfloob, line 46: The namespace qualification is redundant here.

drfloob (changes requested):

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
Vignesh2208: Done. I used AnyInvocableClosure now.

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
Vignesh2208: Done.

src/core/lib/event_engine/iomgr_engine/ev_poll_posix.cc
Vignesh2208: Fixed.

src/core/lib/event_engine/iomgr_engine/ev_poll_posix.h
Vignesh2208: The poller definition is inside grpc_event_engine::iomgr_engine namespace whereas EventEngine is inside grpc_event_engine::experimental namespace. So I think the namespace qualification is required here.

drfloob: ah, my mistake. The `grpc_event_engine` piece is redundant, but getting rid of it, `experimental::EventEngine::Duration` looks a bit weird. I think it's fine as is.

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
```diff
@@ -41,33 +42,38 @@
...
  public:
   Epoll1EventHandle(int fd, Epoll1Poller* poller)
       : fd_(fd),
         pending_actions_(0),
         list_(),
         poller_(poller),
+        exec_actions_closure_([this]() {
+          ExecutePendingActions();
+          // Unref - for the Ref taken in SetPendingActions.
+          Unref();
```
drfloob, line 75: I am still curious if this is going to be a source of performance issues, doing multiple atomic increment and decrements for every I/O callback. Not sure what can be done about it at this point, thought. @ctiller thoughts?

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.h
```diff
@@ -47,17 +47,18 @@ class Epoll1Poller : public EventPoller {
...
   void Kick() override;
   Scheduler* GetScheduler() { return scheduler_; }
   void Shutdown() override;
   ~Epoll1Poller() override;
 
  private:
-  absl::Status ProcessEpollEvents(int max_epoll_events_to_handle,
-                                  std::vector<EventHandle*>& pending_events);
-  absl::Status DoEpollWait(grpc_core::Timestamp deadline);
+  bool ProcessEpollEvents(int max_epoll_events_to_handle,
```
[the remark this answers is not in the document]
Vignesh2208: Done.

src/core/lib/event_engine/iomgr_engine/ev_poll_posix.cc
```diff
@@ -211,13 +227,13 @@ class PollEventHandle : public EventHandle {
   bool pollhup_;
   int watch_mask_;
   absl::Status shutdown_error_;
+  experimental::AnyInvocableClosure exec_actions_closure_;
```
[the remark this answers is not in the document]
Vignesh2208: Fixed. Added a using statement now.

src/core/lib/event_engine/iomgr_engine/ev_epoll1_linux.cc
Vignesh2208: It turns out we dont need a Ref/Unref for epoll1 poller because there is no danger of use after free. This is because when an Epoll1EventHandle is orphaned, its not deleted/free'ed but rather just returned to a pool/cache of EventHandles from where it can be re-used. So we don't need to Ref/Unref there and I removed it.

Vignesh2208: @hork, @ctiller  I did discover a bug in Epoll1EventHandle implementation which required converting the Epoll1EventHandle's pending_actions_ variable from int ---> atomic<int>. Could you ptal ? Thanks!

Vignesh2208: I added some rigorous testing in test/core/event_engine/iomgr_event_engine/event_poller_posix_test.cc and it caught some bugs. Those fixes are also included now in the PR with comments explaining the issues.

Vignesh2208:
Friendly ping on this one :) I have fixed some bugs and updated per our discussion

src/core/lib/event_engine/posix_engine/posix_engine_closure.h
```diff
@@ -31,36 +31,42 @@ namespace posix_engine {
...
+  ~PosixEngineClosure() final = default;
   void SetStatus(absl::Status status) { status_ = status; }
   void Run() override {
-    cb_(absl::exchange(status_, absl::OkStatus()));
-    if (!is_permanent_) {
+    // We need to read the is_permanent_ variable before executing the
+    // enclosed callback. This is because a permanent closure may delete this
+    // object within the callback itself and thus reading this variable after
+    // the callback execution is not safe.
+    if ABSL_PREDICT_FALSE (!is_permanent_) {
```
ctiller, line 50: please make this look like an actual if... `if (ABSL_PREDICT_FALSE(...))`

src/core/lib/event_engine/iomgr_engine/ev_poll_posix.cc
```diff
@@ -0,0 +1,858 @@
+// Copyright 2022 The gRPC Authors
```
ctiller, line 1: merge error?

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
```diff
@@ -487,24 +523,25 @@ void Epoll1EventHandle::SetWritable() { write_closure_->SetReady(); }
...
+    if (ProcessEpollEvents(
+            was_kicked_ ? INT_MAX : MAX_EPOLL_EVENTS_HANDLED_PER_ITERATION,
+            pending_events)) {
       was_kicked_ = false;
     }
-    return status;
+    if (pending_events.empty()) {
+      return Poller::Kicked{};
+    }
+    return pending_events;
```
drfloob, line 544: nit: it seems non-intuitive that the poller could be kicked, but Work returns a result that is not Poller::Kicked(). I think it's worth documenting.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.h
```diff
@@ -42,26 +40,35 @@ namespace posix_engine {
...
+  // Process the epoll events found by DoEpollWait() function.
+  // - g_epoll_set.cursor points to the index of the first event to be processed
+  // - This function then processes up-to max_epoll_events_to_handle and
+  //   updates the g_epoll_set.cursor.
+  // It returns true, it there was a Kick that forced invocation of this
+  // function. It also returns the list of closures to run to take action
+  // on file descriptors that became readable/writable.
+  bool ProcessEpollEvents(int max_epoll_events_to_handle,
+                          Poller::Events& pending_events);
+  int DoEpollWait(
```
drfloob, line 66: nit: please document the meaning of the return type with its declaration.

src/core/lib/event_engine/posix_engine/ev_poll_posix.cc
```diff
@@ -87,20 +92,32 @@ class PollEventHandle : public EventHandle {
...
         watch_mask_(-1),
         shutdown_error_(absl::OkStatus()),
+        exec_actions_closure_([this]() { ExecutePendingActions(); }),
         on_done_(nullptr),
-        read_closure_(reinterpret_cast<IomgrEngineClosure*>(kClosureNotReady)),
+        read_closure_(reinterpret_cast<PosixEngineClosure*>(kClosureNotReady)),
         write_closure_(
-            reinterpret_cast<IomgrEngineClosure*>(kClosureNotReady)) {
+            reinterpret_cast<PosixEngineClosure*>(kClosureNotReady)) {
+    poller_handles_list_.next = nullptr;
```
drfloob, line 100: nit: I'd recommend having a `HandleList` handle its own member initialization, either via member default values or via a constructor.

For example:
```
  class HandlesList {
   public:
    explicit HandlesList(PollEventHandle* handle) : handle(handle) {}
    PollEventHandle* handle;
    PollEventHandle* next = nullptr;
    PollEventHandle* prev = nullptr;
  };
```

test/core/event_engine/posix/event_poller_posix_test.cc
```diff
@@ -513,7 +544,196 @@ TEST_P(EventPollerTest, TestEventPollerHandleChange) {
...
+  Worker(Scheduler* scheduler, PosixEventPoller* poller, int num_handles,
+         int num_wakeups_per_handle)
+      : scheduler_(scheduler), poller_(poller) {
+    handles_.reserve(num_handles);
+    for (int i = 0; i < num_handles; i++) {
+      handles_.push_back(
+          new WakeupFdHandle(num_wakeups_per_handle, scheduler_, poller_));
+    }
+  }
+  void Ref() { ref_count_.fetch_add(1, std::memory_order_relaxed); }
```
drfloob, line 675: nit: You can use our internal RefCounted types here, no need to reinvent.

test/core/event_engine/posix/event_poller_posix_test.cc
```diff
@@ -513,7 +544,196 @@ TEST_P(EventPollerTest, TestEventPollerHandleChange) {
...
+      promise.Set(true);
+    }
+  }
+
+  void Start() {
+    // Start executing Work(..).
+    scheduler_->Run([this]() { Work(); });
+  }
+
+  void Wait() { EXPECT_TRUE(promise.Get()); }
```
drfloob, line 687: suggestion: You can use `promise.WaitWithTimeout` to time-box this test. By default, tests run for 5 minutes, which is a long time if you're deadlocked within a second or two.

test/core/event_engine/posix/event_poller_posix_test.cc
```diff
@@ -513,7 +544,196 @@ TEST_P(EventPollerTest, TestEventPollerHandleChange) {
...
+                                      : absl::InternalError("Spurious Wakeup");
+        case EINTR:
+          continue;
+        default:
+          return absl::Status(absl::StatusCode::kInternal,
+                              absl::StrCat("read: ", strerror(errno)));
+      }
+    }
+  }
+  void Ref() { ref_count_.fetch_add(1, std::memory_order_relaxed); }
```
drfloob, line 633: nit: You can use our internal RefCounted types here, no need to reinvent.

drfloob (changes requested):
Mostly looks good to me, I left just a few suggestions and style fixes.

test/core/event_engine/posix/event_poller_posix_test.cc
Vignesh2208: Done. I used DualRefCounted now.

test/core/event_engine/posix/event_poller_posix_test.cc
Vignesh2208: Done.

test/core/event_engine/posix/event_poller_posix_test.cc
Vignesh2208: Done. I used DualRefCounted now.

src/core/lib/event_engine/posix_engine/ev_poll_posix.cc
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.h
Vignesh2208: Done.

src/core/lib/event_engine/iomgr_engine/ev_poll_posix.cc
Vignesh2208: I dont see a merge error. Not sure what you meant.

src/core/lib/event_engine/posix_engine/posix_engine_closure.h
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
Vignesh2208: Added a comment to Epoll1Poller::Work

drfloob (approved):

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
```diff
@@ -76,57 +83,96 @@ class Epoll1EventHandle : public EventHandle {
...
+    // parallel and read the pending_actions_ variable. So we need to use
+    // atomics to manipulate pending_actions_ variable.
+
+    // Load pending_actions_ into curr.
+    int curr = pending_actions_.load(std::memory_order_acquire);
+    while (true) {
+      // Check whether pending_actions_ == curr. If so set pending_actions_ to
+      // curr | pending_actions. Otherwise update curr to latest value of
+      // pending_actions_ and retry.
+      if (pending_actions_.compare_exchange_strong(curr, curr | pending_actions,
```
ctiller, line 120: Could we eliminate this by breaking our pending actions into three: read, write, and error - and having a different closure for each?

Vignesh2208: Yes I used 3 atomic variables in the latest commit. It eliminates that loop.

Vignesh2208:
Friendly ping on this one :)

Vignesh2208:
@ctiller Friendly ping on this one :) Thanks for taking a look

ctiller (approved):
I think this is good for now.

CC @drfloob for the rest of this

I think looking at this I think we need another iteration on the poller interface (now that we have `Executor` and some commitment to doing work stealing)

I'd suggest we make the following change to Work:

```
virtual absl::Status Work(Duration timeout, absl::FunctionRef<void()> call_before_scheduling_work) = 0;
```

Notes:
- no longer return a vector of closures, instead just execute them directly on an `Executor` (this type alleviates most of my concerns regarding circular dependencies and testability of pollers in isolation)
- *but* as we know we want to schedule the next work before scheduling read/write closures, pass down a helper function to allow that (`absl::FunctionRef` is guaranteed not to allocate, but in exchange can't be stored - it's perfect for this use case).

drfloob:
> no longer return a vector of closures, instead just execute them directly on an Executor

Excellent, I agree, I had started the IOCP implementation with that design, using a subset of the EventEngine API as the executor (which brought about some of your testing concerns, understandably).

> pass down a helper function to allow [some relative ordering of callback execution]

This function's job is essentially to make the Poller schedule itself (anonymously) for async execution, and it cannot block because the read/write callbacks need to be scheduled too, hopefully quickly. Depending on how that's all implemented, read/write callbacks and this self-scheduling function may still be executed in any relative order. I think that's probably fine, the goal here is to ensure we're polling quickly, not to guarantee any specific ordering. But if you're suggesting we need a strong relative ordering, that should be made clear here.

Side note: I'm hoping we can find a stricter API than "execute this anonymous code in this specific order", but I think that was the best I came up with as well back when I did the IOCP poller. I'll give it a think.

PR #30427, opened 2022-07-27
Time conversion utility for EventEngine::Durations

drfloob:

### DIFF
--- a/src/core/lib/event_engine/time_util.cc
+++ b/src/core/lib/event_engine/time_util.cc
@@ -1,0 +1,30 @@
+// Copyright 2022 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/core/lib/event_engine/time_util.cc
```diff
@@ -0,0 +1,30 @@
...
+
+#include <grpc/event_engine/event_engine.h>
+
+namespace grpc_event_engine {
+namespace experimental {
+
+int64_t Milliseconds(EventEngine::Duration d) {
+  return std::chrono::duration_cast<std::chrono::milliseconds>(d).count();
+}
+
```
Vignesh2208, line 28: Can we also add a method to convert to grpc_core::Timestamp ?

drfloob: See my comments on your PR. I don't think it's necessary, but it can be added here, sure.

Vignesh2208 (approved):

Issue #29854, opened 2022-06-01
Grpc.Core.Internal.AsyncCall<>.UnaryCall hangs (malformed call status?)


zhenomansh:
**Ver 1.45** (1.46 code is the same)
The symptom is hanging in Grpc.Core.Internal.AsyncCall<>.**UnaryCall** on return unaryResponseTcs.Task.GetAwaiter().GetResult(); line.

The stack trace example is:

>   [Waiting on Async Operation, double-click or press enter to view Async Call Stacks]
>   Grpc.Core.dll!Grpc.Core.Internal.AsyncCall<GenericRequest, GenericResponse>.**UnaryCall**(GenericRequest msg) Line 116 C#
>   Grpc.Core.dll!Grpc.Core.Calls.BlockingUnaryCall<GenericRequest, GenericResponse>(Grpc.Core.CallInvocationDetails<GenericRequest, GenericResponse> call, GenericRequest req) Line 25 C#
>   Grpc.Core.dll!Grpc.Core.DefaultCallInvoker.BlockingUnaryCall<GenericRequest, GenericResponse>(Grpc.Core.Method<GenericRequest, GenericResponse> method, string host, Grpc.Core.CallOptions options, GenericRequest request) Line 16 C#
>   Grpc.Core.Api.dll!Grpc.Core.Interceptors.InterceptingCallInvoker.BlockingUnaryCall.AnonymousMethod__3_0<GenericRequest, MGenericResponse>(GenericRequest req, Grpc.Core.Interceptors.ClientInterceptorContext<GenericRequest, GenericResponse> ctx) Line 51 C#
>   Grpc.Core.Api.dll!Grpc.Core.ClientBase.ClientBaseConfiguration.ClientBaseConfigurationInterceptor.BlockingUnaryCall<GenericRequest, GenericResponse>(GenericRequest request, Grpc.Core.Interceptors.ClientInterceptorContext<GenericRequest, GenericResponse> context, Grpc.Core.Interceptors.Interceptor.BlockingUnaryCallContinuation<GenericRequest, GenericResponse> continuation) Line 174 C#
>   Grpc.Core.Api.dll!Grpc.Core.Interceptors.InterceptingCallInvoker.BlockingUnaryCall<GenericRequest, GenericResponse>(Grpc.Core.Method<GenericRequest, GenericResponse> method, string host, Grpc.Core.CallOptions options, GenericRequest request) Line 48 C#

[...]

PR #30429, opened 2022-07-28
Fix hang in AsyncCall.UnaryCall

tonydnewell:
Fix for https://github.com/grpc/grpc/issues/29854

This fixes the hang in AsyncCall.UnaryCall() when an exception in thrown.

I was unable to reproduce the original corrupt status code. Looking at the marshalling code it is unclear
how the UTF-8 decoding would fail.

There is no obvious way to create a unit test witha mock failure and the failure happens at quite a low level. I was able to test locally by temporarily modifying AsyncCall.UnaryCall() and explicitly throwing an exception and checking the behaviour of both the fixed and non-fixed code.


src/csharp/Grpc.Core/Internal/AsyncCall.cs
```diff
@@ -122,6 +122,9 @@ public TResponse UnaryCall(TRequest msg)
                             catch (Exception e)
                             {
                                 Logger.Error(e, "Exception occurred while invoking completion delegate.");
+                                unaryResponseTcs.SetException(new RpcException(
```
jtattermusch, line 125: I think there are problems with this fix:
-  since in https://github.com/grpc/grpc/issues/29854, ctx.GetReceivedStatusOnClient() throws an exception, the `HandleUnaryResponse()` (which is responsible for cleaning up after the call) never gets invoked, so there is high likelyhood that the call won't get cleaned up properly and it will leak.
- in principle, if the message detail in GetReceivedStatusOnClient can get corrupt (and stop being unmarshallable as UTF-8),  the exact same thing can likely happen for other call types than just syncUnary (e.g. async Unary and potentially some of the streaming ones as well). Rather than fixing the same problem in multiple places, it would be better to make sure we understand why exactly does `ctx.GetReceivedStatusOnClient()` fail and whether that's caused by a bug in C-core or not.

tonydnewell:
Too out of date with branch. Will create another PR

PR #30431, opened 2022-07-28
XdsClient: convert xDS bootstrap code to use new JSON parsing API

markdroth:
Built on #27976.


src/core/ext/xds/xds_bootstrap.h
```diff
@@ -83,23 +95,33 @@ class XdsBootstrap {
...
-  static std::unique_ptr<XdsBootstrap> Create(absl::string_view json_string,
-                                              grpc_error_handle* error);
+  static absl::StatusOr<XdsBootstrap> Create(absl::string_view json_string);
 
   // Do not instantiate directly -- use Create() above instead.
-  XdsBootstrap(Json json, grpc_error_handle* error);
+  XdsBootstrap() = default;
+
+  // Copyable.
+  XdsBootstrap(const XdsBootstrap&);
```
ctiller, line 109: Do we need copyable? (the copy constructors for the maps and vectors are adding a fair amount of bloat here)

markdroth: No, we don't.  Removed.

src/core/ext/xds/xds_bootstrap_grpc.cc
```diff
@@ -36,380 +36,285 @@
...
+   public:
+    bool IsEnabled(absl::string_view key) const override {
+      if (key == "federation") return XdsFederationEnabled();
+      return true;
+    }
+  };
+  auto bootstrap_json = LoadFromJson<BootstrapJson>(*json, XdsJsonArgs());
+  if (!bootstrap_json.ok()) return bootstrap_json.status();
+  std::vector<XdsServer> servers;
+  for (auto& server : bootstrap_json->servers) {
```
ctiller, line 284: I'm certain I'm missing something, but any way we could make this `servers = std::move(bootstrap_json->servers);`?

markdroth: I've changed it to avoid the need for this entirely by making XdsBootstrap a cleaner interface.  The change was a little invasive, but I think it makes things cleaner anyway.

ctiller (approved):

PR #30436, opened 2022-07-28
service config: fix portability bug

markdroth:
This fixes a portability CBF that was introduced in #30321.


markdroth:
Reverting this in favor of https://github.com/grpc/grpc/pull/30437.

PR #30439, opened 2022-07-28
[ObjC] run ios CronetTest from bazel

HannahShiSFB:
It's part of grpc pod migration.  https://github.com/grpc/grpc-ios/issues/104

matching the CronetTest framework artifact as in our current cocoapod builds.
will delete the temporary code for pod build when migration finish. 

How to run:
1. Start port server with tools/run_tests/start_port_server.py
2. OVERRIDE_BAZEL_VERSION=5.2.0 SCHEME=CronetTests src/objective-c/tests/run_one_test_bazel.sh 

@dennycd

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -704,6 +704,13 @@ grpc_cc_library(
 )
 
 grpc_cc_library(
+    name = "grpc_cronet_hdrs",
+    hdrs = [
+        "include/grpc/grpc_cronet.h",
+    ],
+)
+
+grpc_cc_library(
     name = "tchar",
     srcs = [
         "src/core/lib/gprpp/tchar.cc",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -88,3 +88,32 @@ pip_install(
     name = "grpc_python_dependencies",
     requirements = "@com_github_grpc_grpc//:requirements.bazel.txt",
 )
+
+http_archive(
+    name = "build_bazel_rules_swift",
+    sha256 = "12057b7aa904467284eee640de5e33853e51d8e31aae50b3fb25d2823d51c6b8",
+    url = "https://github.com/bazelbuild/rules_swift/releases/download/1.0.0/rules_swift.1.0.0.tar.gz",
+)
+
+http_archive(
+    name = "rules_pods",
+    urls = ["https://github.com/pinterest/PodToBUILD/releases/download/4.1.0-412495/PodToBUILD.zip"],
+)
+
+load(
+    "@build_bazel_rules_swift//swift:repositories.bzl",
+    "swift_rules_dependencies",
+)
+load(
+    "@rules_pods//BazelExtensions:workspace.bzl",
+    "new_pod_repository",
+)
+
+swift_rules_dependencies()
+
+new_pod_repository(
+    name = "CronetFramework",
+    is_dynamic_framework = True,
+    podspec_url = "https://raw.githubusercontent.com/CocoaPods/Specs/master/Specs/2/e/1/CronetFramework/0.0.5/CronetFramework.podspec.json",
+    url = "https://storage.googleapis.com/grpc-precompiled-binaries/cronet/Cronet.framework-v0.0.5.zip",
+)
--- a/src/objective-c/BUILD
+++ b/src/objective-c/BUILD
@@ -240,6 +240,18 @@ grpc_objc_library(
 )
 
 grpc_objc_library(
+    name = "grpc_objc_client_core_cronet_testing",
+    srcs = glob(["GRPCClient/private/GRPCCore/GRPCCoreCronet/*.m"]),
+    hdrs = glob(["GRPCClient/private/GRPCCore/GRPCCoreCronet/*.h"]),
+    deps = [
+        ":grpc_objc_client_core_internal_testing",
+        "//:grpc_cronet_hdrs",
+        "//src/core/ext/transport/cronet:grpc_transport_cronet_client_secure",
+        "@CronetFramework",
+    ],
+)
+
+grpc_objc_library(
     name = "proto_objc_rpc_internal_testing",
     srcs = [
         "ProtoRPC/ProtoRPCLegacy.m",
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -22,6 +22,7 @@ load(
     "local_objc_grpc_library",
     "proto_library_objc_wrapper",
 )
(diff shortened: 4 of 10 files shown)


BUILD
```diff
@@ -565,6 +565,13 @@ grpc_cc_library(
     ],
 )
 
+grpc_cc_library(
```
dennycd, line 568: see if we could move this into src/objective-c/BUILD if only used by iOS/

WORKSPACE
```diff
@@ -88,3 +88,32 @@ pip_install(
...
+    "swift_rules_dependencies",
+)
+load(
+    "@rules_pods//BazelExtensions:workspace.bzl",
+    "new_pod_repository",
+)
+
+swift_rules_dependencies()
+
+new_pod_repository(
```
dennycd, line 114: pls add a summary of this target (e.g. matching the cronet framework artifact as in our current cocoapod builds etc. )

WORKSPACE
```diff
@@ -88,3 +88,32 @@ pip_install(
     name = "grpc_python_dependencies",
     requirements = "@com_github_grpc_grpc//:requirements.bazel.txt",
 )
+
+http_archive(
+    name = "build_bazel_rules_swift",
```
dennycd, line 93: comments on why this part is needed ?

src/objective-c/tests/BUILD
```diff
@@ -220,6 +242,13 @@ ios_unit_test(
     ],
 )
 
+grpc_objc_ios_unit_test(
+    name = "CronetTests",
```
dennycd, line 246: plx specify which existing cocoapod test this is replacing with

dennycd: https://github.com/grpc/grpc/blob/009dadbb743f4575b60997fb723919eba0bcb00a/src/objective-c/tests/Podfile#L44

dennycd (commented):
plx check and see if bazel cronet passed in kororo ci , thanks !

BUILD
HannahShiSFB: Bazel BUILD cannot include files in parent directories.
I can take a look if this header can be moved to `src/objective-c`, but that may mean grpc API changes

WORKSPACE
HannahShiSFB: It’s needed by PodToBUILD which converts Cronet pod to bazel build.
https://bazel.build/migrate/cocoapods

test/core/end2end/generate_tests.bzl
```diff
@@ -429,6 +429,9 @@ def grpc_end2end_tests():
             "//test/core/compression:args_utils",
             "//:grpc_http_filters",
         ],
+        visibility = [
+            "//src/objective-c/tests:__subpackages__",
```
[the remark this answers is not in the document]
HannahShiSFB: e2e code is called from cornet test cases, therefore this dependency

src/objective-c/tests/CronetTests/CronetUnitTests.mm
```diff
@@ -39,7 +39,11 @@
 #import "test/core/end2end/data/ssl_test_data.h"
 #import "test/core/util/test_config.h"
 
-#import <openssl_grpc/ssl.h>
+#if COCOAPODS
```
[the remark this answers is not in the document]
HannahShiSFB: will delete the temporary code for pod build when bazel migration finish.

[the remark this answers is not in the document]
HannahShiSFB: unfortunately,  cocoapod install scripts only modified src/core/ not obj-c/.

src/objective-c/tests/BUILD
HannahShiSFB: > https://github.com/grpc/grpc/blob/009dadbb743f4575b60997fb723919eba0bcb00a/src/objective-c/tests/Podfile#L44

exactly

BUILD
dennycd: thanks, ok for now.

WORKSPACE
dennycd: sg. https://github.com/grpc/grpc-ios/issues/104 will track the task to introduce bazel based CronetFramework build so we can remove this cocoapod dependency in our bazel tool tree.

src/objective-c/BUILD
```diff
@@ -239,6 +239,18 @@ grpc_objc_library(
     ],
 )
 
+grpc_objc_library(
+    name = "grpc_objc_client_core_cronet_testing",
+    srcs = glob(["GRPCClient/private/GRPCCore/GRPCCoreCronet/*.m"]),
```
dennycd, line 244: Consider refactor this out to a separate target e.g. GRPCCoreCronet similar to how we structure our Cocoapod (https://github.com/grpc/grpc/blob/27e5b4e1877840b58c92d5d853c20ab713d19187/gRPC.podspec#L150) and have the test target depends on it instead

HannahShiSFB: There is already a `grpc_objc_client` at line 118.
The testing targets are configured differently from prod targets, I can take a look if they can be consolidated in later PRs.

dennycd: sg, let's do that in a follow up PR . The equivalent Pod test target is 

https://github.com/grpc/grpc/blob/ac585b8f7e78ea9d698bcc6c0019a336332b57e7/src/objective-c/tests/Podfile#L44

dennycd (approved):
chatted offline, we will do a follow up refactor of GRPCCoreCronet lib target

jtattermusch:
The cronet tests are fully broken on master: https://source.cloud.google.com/results/invocations/7801e1fc-29f7-404b-8373-96424cffc1a1/targets/%2F%2Fsrc%2Fobjective-c%2Ftests:CronetTests/log
please fix this ASAP.
For newly introduced test suite, there need to be more testing than just a single run on presumbit (which seems to have passed in this case) - what the flakiness if you do e.g. 100 runs locally?

dennycd:
I believe the regression is not related to this PR or caused by test flake. this one landed on 9/27 and the bazel cronet test is consistently green until https://github.com/grpc/grpc/pull/29598. Failure first seen in https://source.cloud.google.com/results/invocations/3f8719e3-5f05-4577-b9a8-c820adbeab60/targets/grpc%2Fcore%2Fpull_request%2Fmacos%2Fgrpc_objc_bazel_test/log

PR #30448, opened 2022-08-01
[ObjC]fix c++14 errors in bazel objC

HannahShiSFB:
It's part of grpc pod migration. 
to support #30439, since c-core changed to c++14.
https://github.com/grpc/grpc-ios/issues/104


dennycd (approved):
lgtm. Maybe more overview on the changes ( ObjC++ and c++14 compiler flag)

PR #30449, opened 2022-08-01
Allow callback functions to return TSI_ASYNC without causing errors.

zeromath:
This change allows the callback function `OnHandshakeNextDoneGrpcWrapper` to return `TSI_ASYNC` without causing any error. This is needed to run asynchronous operations multiple times.

The model is as following: when the callback function `OnHandshakeNextDoneGrpcWrapper` is called, we need to return `TSI_ASYNC` to inform gRPC to wait for another call of `OnHandshakeNextDoneGrpcWrapper` later so that certain gRPC operations can be carried out. This model is needed internally.

@markdroth @ctiller @ZhenLian @matthewstevenson88

### DIFF
--- a/src/core/lib/security/transport/security_handshaker.cc
+++ b/src/core/lib/security/transport/security_handshaker.cc
@@ -386,6 +386,9 @@ grpc_error_handle SecurityHandshaker::OnHandshakeNextDoneLocked(
         /*urgent=*/true, /*min_progress_size=*/1);
     return error;
   }
+  if (result == TSI_ASYNC) {
+    return GRPC_ERROR_NONE;
+  }
   if (result != TSI_OK) {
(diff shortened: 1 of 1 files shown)


markdroth:
I don't understand why this is necessary.  If the TSI handshaker knows that it needs to do more async work, why doesn't it just do that work directly before it invokes the callback?  Why is this hop into the gRPC security handshaker necessary or useful?

markdroth:
As we discussed, the problem you're seeing here is a bug, and we should diagnose and fix that bug instead of changing the TSI API this way.

zeromath:
Per offline discussion and experimenting, this is **not** a bug in gRPC or TSI and this change is no longer needed.

PR #30450, opened 2022-08-01
Remove `include/grpcpp/impl/codegen/async_generic_service.h`

ralphchung:
This PR is a part of [b/196639718](http://b/196639718). It tries to move the content of `include/grpcpp/impl/codegen/async_generic_service.h` to `include/grpcpp/generic/async_generic_service.h`. We do not remove anything to achieve backward compatibility. The file `include/grpcpp/impl/codegen/async_generic_service.h` should be removed in the future.


include/grpcpp/impl/codegen/async_generic_service.h
```diff
@@ -19,119 +19,8 @@
...
-        [this](grpc::CallbackServerContext* ctx) {
-          return CreateReactor(static_cast<GenericCallbackServerContext*>(ctx));
-        });
-  }
-
-  grpc::Server* server_{nullptr};
-};
-
-}  // namespace grpc
+#include <grpcpp/impl/codegen/async_generic_service.h>
```
ctiller, line 24: Shouldn't this file be removed instead of including itself?

Vignesh2208 (approved):

PR #30451, opened 2022-08-01
xDS interop: set default socket timeout to 60

sergiitk:
`kubernetes` library does not provide a way to configure the default socket timeout that will be used with `urllib3` it uses under the hood. And `urllib3` default socket timeout is infinity.

This PR sets the default socket timeout using python's `socket.setdefaulttimeout()` to 60 seconds.

This affects `urllib3` directly,  and therefore `kubernetes`.
The changes is also picked up by the `google-api-python-client`, which does not use `urllib3` (it uses `httplib2`), but [respectes](https://googleapis.github.io/google-api-python-client/docs/epy/googleapiclient.http-module.html#build_http) `socket.setdefaulttimeout()`.

ref b/240068800

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/bin/cleanup/cleanup.py
+++ b/tools/run_tests/xds_k8s_test_driver/bin/cleanup/cleanup.py
@@ -336,6 +336,9 @@ def main(argv):
         raise app.UsageError('Too many command-line arguments.')
     load_keep_config()
 
+    # Must be called before KubernetesApiManager or GcpApiManager init.
+    xds_flags.set_socket_default_timeout_from_flag()
+
     project: str = xds_flags.PROJECT.value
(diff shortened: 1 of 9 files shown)


sergiitk:
- [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/99f7bad5-d93e-48fa-a1a1-46b2ce1beba8)
- [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/0e3143c5-99f0-42e9-a84e-06962e391d8b)
- [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/d25dacd5-df36-4a27-897b-1fdffb4db358)

gnossen (approved):

PR #30452, opened 2022-08-01
Add enablePrometheus annotation.

wanlin31:
The PR adds the enablePrometheus annotation to load tests that are
part of PSM data plan performance tests. This annotation enables
all PSM related tests to obtain data from Prometehus, even for the
regular tests.


gnossen (approved):

tools/internal_ci/linux/grpc_psm_performance_gke_experiment.sh
```diff
@@ -100,6 +100,7 @@ psmBuildConfigs() {
         -a ci_gitCommit="${GRPC_GITREF}" \
         -a ci_gitCommit_java="${GRPC_JAVA_GITREF}" \
         -a ci_gitActualCommit="${KOKORO_GIT_COMMIT}" \
+        -a enablePrometheus="true" \
```
paulosjca, line 103: true does not need quotation marks in this case.

wanlin31: Right, with or without the quotation it produce the same results in test-infra. Let me remove it.

paulosjca (approved):

PR #30454, opened 2022-08-01
unimplemented error on parse failure confused me

cove:
This response confused me due to https://github.com/dotnet/aspnetcore/issues/39317 where the body of a response went missing, but I got back an unimplemented error.

### DIFF
--- a/include/grpcpp/impl/codegen/client_unary_call.h
+++ b/include/grpcpp/impl/codegen/client_unary_call.h
@@ -88,6 +88,8 @@ class BlockingUnaryCallImpl {
     // TODO(yashykt): If deserialization fails, but the status received is OK,
     // then it might be a good idea to change the status to something better
     // than StatusCode::UNIMPLEMENTED to reflect this.
+    // TODO(cove): This confused me due to this issue:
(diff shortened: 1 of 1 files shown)


yashykt:
Note that is C++ and the linked issue is .NET

markdroth:
I don't have all of the context on the C# problem here, but I don't think it makes sense to add a comment in a C++ header file about a C# issue.

cove:
Sorry let me explain the C# connection here for context better: 

- I have a server that's written in C# and is using MSFT's gRPC impl
- My client is using this gRPC impl in C++
- I ran into this issue cited in to the TODO on the C++ side _and_ a bug on the C# server side at the same time, which resulted in an unimplemented error from the C++ gRPC client when it was implemented on the server side. So took some time to track down what was going on, a better error message from the gRPC client would have helped.

markdroth:
Okay, thanks, that's useful info.  We should definitely have a bug to track improving the error message in this case, but I don't think we need a comment pointing to the C# issue.

@yashykt, do we have an existing bug tracking this?  If not, let's create one.

PR #30467, opened 2022-08-02
service config: use new JSON API

markdroth:


src/core/ext/filters/rbac/rbac_service_config_parser.cc
```diff
@@ -18,587 +18,657 @@
...
-                              std::vector<grpc_error_handle>* error_list) {
-  std::string regex;
-  ParseJsonObjectField(regex_matcher_json, "regex", &regex, error_list);
-  return regex;
-}
+struct RbacConfig {
+  struct RbacPolicy {
+    struct Rules {
+      struct Policy {
+        struct CidrRange {
```
ctiller, line 42: The deep nesting of these structs makes it hard to reason about this code without tooling (to answer the 'where the heck am I' question)

Suggest either:
1. flattening the structure (pulling the inner structs up to the top level namespace and giving them longer names)
or
2. moving all methods out-of-line below the structure definitions, so that the overall structure can be grokked on ~1 page of code, and the methods carry location information (in the form of `RbacConfig::RbacPolicy::Rules::Policy::CidrRange::JsonLoader()` for instance)

ctiller (changes requested):
Looks good... there's a readability issue I think we should tackle prior to submission though. Left some thoughts on alternatives, but happy to discuss further.

src/core/ext/filters/rbac/rbac_service_config_parser.cc
markdroth: Done.  I went with the second option.

ctiller (approved):

PR #30473, opened 2022-08-02
[WIP] EventEngine Forkables

drfloob:
A (currently) `pthread_atfork`-based fork support mechanism, allowing EventEngines - or any other object that wants to implement the `Forkable` interface - respond to forks.

### DIFF
--- a/src/core/lib/event_engine/forkable.h
+++ b/src/core/lib/event_engine/forkable.h
@@ -1,0 +1,57 @@
+// Copyright 2022 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef GRPC_CORE_LIB_EVENT_ENGINE_FORKABLE_H
+#define GRPC_CORE_LIB_EVENT_ENGINE_FORKABLE_H
+
+#include <grpc/support/port_platform.h>
+
+namespace grpc_event_engine {
+namespace experimental {
+
+// Register fork handlers with the system, enabling fork support.
+//
+// This provides pthread-based support for fork events. Any objects that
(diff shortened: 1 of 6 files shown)


ctiller:
I think this has the right structure so far

src/core/lib/event_engine/forkable.h
```diff
@@ -0,0 +1,60 @@
...
+// Global callback for pthread_atfork's *parent argument
+void PostforkParent();
+// Global callback for pthread_atfork's *child argument
+void PostforkChild();
+
+// An interface to be implemented by EventEngines that wish to have managed fork
+// support.
+class Forkable {
+ public:
+  virtual ~Forkable() = default;
```
ctiller, line 42: prefer `protected: ~Forkable() = default;` here -- there should be nothing that deletes a forkable directly, so hiding the destructor prevents that, at which point we don't need the virtual destructor, saving one word per forkable.

drfloob: Done

drfloob:
Reping, this is ready to go. The few test failures are infra issues, or otherwise things I fixed upstream (JSON obj loader).

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -76,6 +82,17 @@ bool PosixEventEngine::Cancel(EventEngine::TaskHandle handle) {
   return r;
 }
 
+void PosixEventEngine::PrepareFork() {
```
ctiller, line 85: Given this isn't being used, maybe we apply some YAGNI and leave the forkable stuff only on the timer_manager_?

ctiller (approved):
Two suggestions which are subjective enough that I'll leave it to your discretion. Looks good.

src/core/lib/event_engine/posix_engine/posix_engine.cc
drfloob: SGTM, done. I also took care of the other thing we talked about, Forkable self-management on (de)construction.

gnossen:
@drfloob Seems that this PR still has not resolved the issue seen at https://github.com/apache/beam/issues/22533#issuecomment-1208723352. I'll try to figure out what the delta is between this PR and the fork interop tests, which this PR seems to have fixed.

ctiller:
Thanks Richard - I think we should land this regardless FWIW as it's a
strict improvement on the state of the codebase from where we were.

drfloob:
@gnossen that is quite odd. I don't believe we've introduced any other new threads, I'll double check.

@ctiller agreed, let's merge this. The test failures are all infrastructure issues.

PR #30477, opened 2022-08-03
Istio Echo: Add version string

yashykt:
Add version string in the echo response message. This is to allow using the same image for the example at https://istio.io/v1.12/blog/2021/proxyless-grpc/#in-your-kubernetes-deployment.

I also noticed a major bug where the xDS server was not serving anything earlier.


sanjaypujare:
> ...
> I also noticed a major bug where the xDS server was not serving anything earlier.

Meaning an Istio bug (bug #?) or a bug in grpc (which was not handling the response)?

veblush (approved):

PR #30481, opened 2022-08-03
Also allow passing in channel args as part of the check_peer call. 

anicr7:
This allows the security connectors to receive args that might be added by handshakers. Currently none of the connectors use the args, hence they are marked as unused. This is required for changes internally for security connectors related to Directpath

@markdroth


markdroth (approved):
Looks great!

Before merging this, please create a cherry-pick import CL, get all necessary approvals, and then coordinate with the importer to merge this PR and submit the CL at the same time.

PR #30484, opened 2022-08-03
Reenable fork tests for all versions but 3.7

gnossen:


gnossen:
Looks like this is now failing as expected:

- https://source.cloud.google.com/results/invocations/a88589ae-91ee-4556-b159-9b32add794e8/targets

PR #30502, opened 2022-08-04
BUILD: move mpscq to its own BUILD target so it can be used without gpr

markdroth:
I had to tag this with "nofixdeps" to get it to stop removing the "gpr_codegen" dependency, without which it won't build.  Is there a better way to do this?


ctiller:
yeah... things are getting a bit message up here.

mpscq includes log.h for GPR_ASSERT (since GPR_ASSERT depends on being able to call gpr_log)... fix_build_deps has gone and grabbed one of the header-only targets we have for that which is wrong (but hard to teach it not to).

The correct fix here (and something that should happen prior to this landing) is to break the circular dependency between gpr_base and mpscq - I'd suggest ejecting anything from gpr_base that depends on mpscq into their own targets, and then having mpscq depend on gpr_base for the log stuff.

ctiller:
alternatively: don't use GPR_ASSERT and really break that dependency

markdroth:
I think we do want to use `GPR_ASSERT()` here.  And since the goal of this PR was to be able to use the mpscq code without depending on all of gpr_base, I guess this isn't going to actually work.

Is it feasible to break gpr_base up into individual targets?  If so, I could take a crack at that.

ctiller:
It's definitely feasible... @drfloob had a first attempt about a month ago. I'd maybe let @ralphchung land #30518 first - as that'll likely help this anyway.

I've periodically thought that it might be worth us defining a single header isolated assert for lower level components... something like:
```
#define GPR_LOW_LEVEL_ASSERT(x) if (!(x)) { abort(); }
```
we'd trade terrible debug-ability here with a lighter dependency story, which might be ok for something like mpscq.

PR #30505, opened 2022-08-04
xDS Interop: C++ xDS Authz supported after 1.47

yashykt:


sergiitk (approved):

sergiitk:
cc @markdroth

sergiitk:
We probably should update https://github.com/grpc/grpc/blob/master/doc/grpc_xds_features.md too.

PR #30507, opened 2022-08-04
Add channel memory benchmark into memory_diff.py

nancylucy01:
Adding channel benchmark flags into memory_diff.py so that the channel memory benchmark will run per PR like the call memory benchmark.
Minstack results for channel benchmark isn't being run with minstack, as flags currently aren't matching up between the channel benchmark and call benchmark

### DIFF
--- a/tools/profiling/memory/memory_diff.py
+++ b/tools/profiling/memory/memory_diff.py
@@ -117,7 +117,7 @@ else:
                     text += '{}: {}\n'.format(key, cur[key])
                 else:
                     text += '{}: {} -> {}\n'.format(key, old[key], cur[key])
-                    if key is 'client call' or 'server call':
+                    if key == 'client call' or 'server call':
                         call_diff_size += cur[key] - old[key]
                     else:
                         channel_diff_size += cur[key] - old[key]
--- a/tools/profiling/memory/memory_diff.py
+++ b/tools/profiling/memory/memory_diff.py
@@ -47,6 +47,10 @@ _INTERESTING = {
         (rb'client call memory usage: ([0-9\.]+) bytes per call', float),
     'server call':
         (rb'server call memory usage: ([0-9\.]+) bytes per call', float),
+    'client channel':
+        (rb'client channel memory usage: ([0-9\.]+) bytes per channel', float),
+    'server channel':
+        (rb'server channel memory usage: ([0-9\.]+) bytes per channel', float),
 }
 
 _SCENARIOS = {
@@ -54,7 +58,10 @@ _SCENARIOS = {
     'minstack': ['--scenario_config=minstack'],
 }
 
-
+_BENCHMARKS = {
+    'call': ['--benchmark_names=call','--size=50000'],
+    'channel': ['--benchmark_names=channel','--size=20000'],
+}
 def _run():
     """Build with Bazel, then run, and extract interesting lines from the output."""
     subprocess.check_call([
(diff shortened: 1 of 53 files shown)


tools/profiling/memory/memory_diff.py
```diff
@@ -97,23 +103,28 @@ def _run():
...
-    for key, value in sorted(cur.items()):
+    for key, value in cur.items():
         text += '{}: {}\n'.format(key, value)
 else:
     print(cur, old)
-    diff_size = 0
+    call_diff_size = 0
+    channel_diff_size = 0
     for scenario in _SCENARIOS.keys():
-        for key, value in sorted(_INTERESTING.items()):
```
nancylucy01, line 106: Want to confirm that it's fine to remove sorted? when it's sorted, the results are in a confusing order with channel and call results being mixed between each other.

tools/profiling/memory/memory_diff.py
```diff
@@ -47,36 +47,42 @@
...
 
 _SCENARIOS = {
     'default': [],
     'minstack': ['--scenario_config=minstack'],
 }
 
-
+_BENCHMARKS = {
+    'call': ['--benchmark_names=call','--size=50000'],
+    'channel': ['--benchmark_names=channel','--size=20000'],
```
nancylucy01, line 63: meant to be 10000, not 20000. Will update with a commit once tests finish running

tools/profiling/memory/memory_diff.py
ctiller: so the slightly scary thing is that python doesn't guarantee an ordering on these... maybe we can call them `call/client`, `call/server`, `channel/client`, `channel/server` or the like instead of the current `client call`

nancylucy01: Oh, was not aware of that, thanks. Changed how we call them to call/client

tools/profiling/memory/memory_diff.py
nancylucy01: Updated it!

tools/profiling/memory/memory_diff.py
```diff
@@ -101,19 +112,28 @@ def _run():
...
+                        call_diff_size += cur[key] - old[key]
+                    else:
+                        channel_diff_size += cur[key] - old[key]
+
+    print("CALL_DIFF_SIZE: %f" % call_diff_size)
+    print("CHANNEL_DIFF_SIZE: %f" % channel_diff_size)
+    check_on_pr.label_increase_decrease_on_pr('per-call-memory', call_diff_size,
+                                              64)
+    check_on_pr.label_increase_decrease_on_pr('per-channel-memory', channel_diff_size,
+                                              1000)
```
nancylucy01, line 135: Chose 1000 since the range from running it 10+ times was 17333-17758 for server, and 37012-37330 for client so max 743 difference so far. Is a range of 1000 too much?

tools/profiling/memory/memory_diff.py
```diff
@@ -101,19 +112,28 @@ def _run():
...
+                        call_diff_size += cur[key] - old[key]
+                    else:
+                        channel_diff_size += cur[key] - old[key]
+
+    print("CALL_DIFF_SIZE: %f" % call_diff_size)
+    print("CHANNEL_DIFF_SIZE: %f" % channel_diff_size)
+    check_on_pr.label_increase_decrease_on_pr('per-call-memory', call_diff_size,
+                                              64)
+    check_on_pr.label_increase_decrease_on_pr('per-channel-memory',
+                                              channel_diff_size, 1000)
```
ctiller, line 135: what's the noise level we see in real runs? (like can we run 10 times and check the output - this value should be about 2x that)

nancylucy01: Chose 1000 since the range from running it 10+ times was 17333-17758 for server, and 37012-37330 for client so max 743 difference so far. Is a range of 1000 too much?

tools/profiling/memory/memory_diff.py
nancylucy01: Check discussion thread https://github.com/grpc/grpc/pull/30507#discussion_r938940875

tools/profiling/memory/memory_diff.py
ctiller: oh, then we're probably ok... seems likely we'll want to up the channel count until that drops maybe... let's see

ctiller (approved):

tools/profiling/memory/memory_diff.py
nancylucy01: Unfortunately my cloudtop can't run it with a higher channel count than 27000 before the UnaryCalls start to fail.

PR #30516, opened 2022-08-05
Allow storing security context extension in grpc_auth_context

anicr7:
This is currently unused, but will be used internally for passing required references that can be tied to auth contexts.

@markdroth

### DIFF
--- a/src/core/lib/security/context/security_context.h
+++ b/src/core/lib/security/context/security_context.h
@@ -24,6 +24,7 @@
 #include <stddef.h>
 
 #include <utility>
+#include <memory>
 
 #include "absl/strings/string_view.h"
 
--- a/src/core/lib/security/context/security_context.h
+++ b/src/core/lib/security/context/security_context.h
@@ -64,7 +64,7 @@ struct grpc_auth_context
     : public grpc_core::RefCounted<grpc_auth_context,
                                    grpc_core::NonPolymorphicRefCount> {
  public:
-  // Base class for all extensions to inherit form.
+  // Base class for all extensions to inherit from.
   class Extension {
    public:
     virtual ~Extension() = default;
--- a/src/core/lib/security/context/security_context.h
+++ b/src/core/lib/security/context/security_context.h
@@ -41,16 +41,6 @@
 
 extern grpc_core::DebugOnlyTraceFlag grpc_trace_auth_context_refcount;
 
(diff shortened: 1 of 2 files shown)


src/core/lib/security/context/security_context.h
```diff
@@ -41,6 +41,16 @@
 
 extern grpc_core::DebugOnlyTraceFlag grpc_trace_auth_context_refcount;
 
+/* --- grpc_security_context_extension ---
+
+   Extension to the security context that may be set in a filter and accessed
+   later by a higher level method on a grpc_call object. */
+
+struct grpc_security_context_extension {
```
markdroth, line 49: This C type is actually for a different purpose, and I don't think we should use it here.  You can move this back down to its original location, below `grpc_auth_context`.

For this PR, let's instead introduce an idiomatic C++ type, like this:

```
struct grpc_auth_context
    : public grpc_core::RefCounted<grpc_auth_context,
                                   grpc_core::NonPolymorphicRefCount> {
 public:
  // A base class for extensions to inherit from.
  class Extension {
   public:
    virtual ~Extension() = default;
  };

  void set_extension(std::unique_ptr<Extension> extension) {
    extension_ = std::move(extension);
  }

  // ...

 private:
  // ...
  std::unique_ptr<Extension> extension_;
};
```

Then you can define a subclass of `grpc_auth_context::Extension` that contains whatever you need it to.

markdroth (commented):
Please add a test in test/core/security/auth_context_test.cc for this.  It's probably sufficient to just create a dynamically allocated extension and add it to the auth context, since all we really need to check is that the extension will be destroyed when the auth context is destroyed, and ASAN will fail if that's not the case.

src/core/lib/security/context/security_context.h
```diff
@@ -64,6 +64,11 @@ struct grpc_auth_context
     : public grpc_core::RefCounted<grpc_auth_context,
                                    grpc_core::NonPolymorphicRefCount> {
  public:
+  // Base class for all extensions to inherit form.
```
markdroth, line 67: s/form/from/

markdroth (approved):
Looks great!

src/core/lib/security/context/security_context.h
anicr7: Fixed typo.

PR #30518, opened 2022-08-05
Build Target Reorganization Part 1

ralphchung:

```
┌───────┐┌───────────────┐        
│grpc++ ││grpc++_unsecure│        
└┬─────┬┘└┬─────────┬────┘        
┌▽───┐┌▽──▽───────┐┌▽────────────┐
│grpc││grpc++_base││grpc_unsecure│
└┬───┘└┬──────────┘└┬────────────┘
┌▽─────▽────────────▽┐            
│grpc_base           │            
└────────────────────┘            
```

* (Covered in this PR) `gpr_base` is merged into `gpr`. The `gpr` here itself is nothing but depending on `gpr_base`. I don't think we need to separate them.
* `grpc++_internals` and `grpc++_internal_hdrs_only` are merged into `grpc++`. The reason is similar. `grpc++` does not contain much stuff but `grpc++_internals`.
* (Covered in this PR) `gpr_codegen` is merged to `gpr` since the files in `gpr_codegen` should be removed in the future and be moved to `gpr`.
* `grpc_codegen` is merged to `grpc_base` with the similar reason as above.
* `grpc++_codegen_base` and `grpc++_codegen_base_src` are merged to `grpc++_base` with the similar reason as above.
* `grpc_secure` is merged to `grpc` since `grpc` is by default secure. We do not need to separate them.
* `grpc++_base_unsecure` is merged either into `grpc++_base` or into `grpc++_unsecure`.

This PR removes `gpr_base` and merges `gpr_codegen` into `gpr` without removing `gpr_codegen` for temporary compatibility. It should be removed in the following PRs.


BUILD
```diff
@@ -965,13 +964,14 @@ grpc_cc_library(
     ],
     language = "c++",
     public_hdrs = GPR_PUBLIC_HDRS,
-    tags = ["nofixdeps"],
-    visibility = ["@grpc:alt_gpr_base_legacy"],
+    tags = [
+        "avoid_dep",
```
ctiller, line 968: Suspect we should remove the avoid_dep here... could you try and see whether fix_build_deps stays rational?

ctiller (approved):
Please do a trial import and see if there's any problems there before submitting.

Vignesh2208 (approved):
If the goal is to eliminate gpr_base entirely, that will most definitely require a cherrypick since gpr_base is used in some places internally as well. Please confirm! Thanks!

BUILD
ralphchung: I tried, and I did not find anything change because of this tag.

ctiller (approved):

Issue #17761, opened 2019-01-18
Cancelling a stream via a cancellation token can lead to incorrect results


jskeet:
### What version of gRPC and what language are you using?

1.17.1, C# 
 
### What operating system (Linux, Windows, …) and version?

Windows 
 
### What runtime / compiler are you using (e.g. python version or version of gcc)

.NET Core SDK 2.1.402
 
### What did you do?

Create a small but non-empty table in a Spanner database, then run the code below, providing the project ID, Spanner instance ID, Spanner database ID and table name.

```csharp
using Google.Cloud.Spanner.Common.V1;
using Google.Cloud.Spanner.V1;
using Grpc.Core;
using System;
using System.Threading;
using System.Threading.Tasks;

class Program
{
    static async Task Main(string[] args)
    {
        if (args.Length != 4)
        {
            Console.WriteLine($"Required arguments: <project ID> <instance ID> <database ID> <table name>");
            return;
        }
        TaskScheduler.UnobservedTaskException += (sender, e) => Console.WriteLine(e.Exception);
        var dbName = new DatabaseName(args[0], args[1], args[2]);
        var client = SpannerClient.Create();
        var grpcClient = client.GrpcClient;

        var session = await grpcClient.CreateSessionAsync(new CreateSessionRequest { DatabaseAsDatabaseName = dbName });

        try
        {
            var request = new ExecuteSqlRequest
            {
                SessionAsSessionName = session.SessionName,
                Sql = $"SELECT * from {args[3]}",
            };

            using (var streamingCall = grpcClient.ExecuteStreamingSql(request))
            {
                var cts = new CancellationTokenSource();
                cts.Cancel();
                // Comment out the following line to see cancellation work properly
                await streamingCall.ResponseHeadersAsync;
                int count = 0;
                while (await streamingCall.ResponseStream.MoveNext(cts.Token))
                {
                    Console.WriteLine($"Read response {++count}");
                }
                Log("Done");
            }
        }
        catch (RpcException e)
        {
            Log($"Caught RPC exception of status {e.StatusCode}");
        }
    }

    static void Log(string message) =>
        Console.WriteLine($"{DateTime.UtcNow:HH:mm:ss.FFFFFF}: {message}");
}
```

### What did you expect to see?

Either a cancellation exception *or* successfully reading a single row, if the cancellation isn't observed in time. Either of those is fine. 
 
### What did you see instead?
 
[...]

Issue #18234, opened 2019-03-04
Inconsistent behavior of RECV_STATUS_ON_CLIENT  and  RECV_MESSAGE  operation when a call gets cancelled


jtattermusch:
This is the root cause of https://github.com/grpc/grpc/issues/17761 (which can happen in other languages than C# as well).

The scenario:
1 .client starts server streaming call and the server responds to the call "quickly" by sending a small response and the call status. Both get immediately sent on the wire and they are also received on the client side.
2. client requests a "late" call cancellation, but in the meantime the response message and OK status have already been received from the server (which can happen in practice, there's always an inherent "race" between client requesting cancellation and the server sending responses).
3. The pending RECV_STATUS_ON_CLIENT operation returns an OK status because that's what was received on the wire.
4. an operation RECV_MESSAGE on the client side will return success==false, and won't return the received message (but we know that internally, the message was read just fine, because we were able to read the OK status coming right after it).

The problem is the inconsistency between behavior or RECV_STATUS_ON_CLIENT  and RECV_MESSAGE.  The received status is reported as OK, but the received message is thrown away at the same time.

Proposed behavior (both are fine)
option 1: the call reports an error status and the response message is thrown away 
option 2: the call reports OK status and the cached message is returned normally (it must have been cached because we were able to read the OK status).

The current behavior is something in-between and it is hard to come up with reasonable wrapping logic to make things behave as expected in the C# layer without reimplementing a lot of C core logic (which we definitely don't want).

jtattermusch:
@AspirinSJL  can you take a look?

jtattermusch:
My repro and unsuccessful attempts to fix in the C# layer are here https://github.com/grpc/grpc/pull/18124

AspirinSJL:
The received message is thrown away because the client has cancelled the call, right?

Looks like the problem is due to the different sources to determine the results of RECV_STATUS_ON_CLIENT and RECV_MESSAGE. 

I think option 1 is more reasonable. We should overwrite the status to "cancelled" in this case because that reflects the user's intent. 

I will try to fix this early this week.

jtattermusch:
The problem with option 1 is that the RECV_STATUS_ON_CLIENT operation is always pending since the start of a server streaming call, so it could happen that the OK status being returned by the surface API wil race with client trying to request cancellation.

Option 2 would be some something along "we know that we already reported the call's status as OK, so we should not throw away the received message despite the users' attempt to cancel - because the call has in fact already finished."

[...]

PR #30523, opened 2022-08-08
Test to verify fix for #17761

tonydnewell:
Verify fix for https://github.com/grpc/grpc/issues/17761

Scenario: the client cancels the request after the server has already successfully returned the results but before the client starts reading.

Expected behaviour: the client sees a Cancelled status.

### DIFF
--- a/src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
+++ b/src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
@@ -189,5 +189,41 @@ namespace Grpc.Core.Tests
 
             using (obj) {}
         }
+
+        [Test]
+        public async Task ServerStreamingCall_CancelAfterServerHasFinished()
+        {
+            // Test to verify the fix for https://github.com/grpc/grpc/issues/17761 that was
+            // fixed in c-core https://github.com/grpc/grpc/issues/18234
+            // Scenario: the client cancels the request after the server has already successfully
+            // returned the results but before the client starts reading.
+            // Expected behaviour is that the client sees a Cancelled status.
+
+            helper.ServerStreamingHandler = new ServerStreamingServerMethod<string, string>(async (request, responseStream, context) => {
+                await responseStream.WriteAsync("abc");
+            });
+
+            var cts = new CancellationTokenSource();
+            var call = Calls.AsyncServerStreamingCall(helper.CreateServerStreamingCall(new CallOptions(cancellationToken: cts.Token)), "request1");
+            // make sure the response and status sent by the server is received on the client side
+            await Task.Delay(2000);
+
+            // cancel the call before actually reading the response
+            cts.Cancel();
+            var moveNextTask = call.ResponseStream.MoveNext();
+
+            try
+            {
+                // cannot use Assert.ThrowsAsync because it uses Task.Wait and would deadlock.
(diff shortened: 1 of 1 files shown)


src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
```diff
@@ -189,5 +189,41 @@ public void CanDisposeDefaultCancellationRegistration()
 
             using (obj) {}
         }
+
+        [Test]
+        public async Task ServerStreamingCall_CancelAfterServerHasFinished()
+        {
+            // Test to verify the fix for https://github.com/grpc/grpc/issues/17761 that was
+            // fixed in c-core https://github.com/grpc/grpc/issues/18234
```
jtattermusch, line 197: https://github.com/grpc/grpc/issues/18234 is an issue and from reading its comment thread it's not clear to me which was the PR that fixed the problem. If https://github.com/grpc/grpc/issues/18234 has really been fixed, it would be good to point out with PR fixed it (and close it as such).

src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
```diff
@@ -189,5 +189,41 @@ public void CanDisposeDefaultCancellationRegistration()
...
+            // Scenario: the client cancels the request after the server has already successfully
+            // returned the results but before the client starts reading.
+            // Expected behaviour is that the client sees a Cancelled status.
```
jtattermusch, line 200: expected behavior is that we see Cancelled status AND no messages are received? (at least it looks like that's what you're asserting).

src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
```diff
@@ -189,5 +189,41 @@ public void CanDisposeDefaultCancellationRegistration()
 
             using (obj) {}
         }
+
+        [Test]
+        public async Task ServerStreamingCall_CancelAfterServerHasFinished()
```
jtattermusch, line 194: I suggest running the test many times (e.g. 5000) to make sure it's not flaky
(you can use run_tests.py -n 5000).

tonydnewell: Ran 5000 times OK

src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
tonydnewell: Probably wasn't https://github.com/grpc/grpc/issues/18234 but was fixed sometime between tags v1.30.0 - June 2020 (broken) and v1.31.0-pre1 -July 2020 (fixed).  I haven't managed to narrow down with a git bisect as my dev environment is too different from the older builds.

src/csharp/Grpc.Core.Tests/CallCancellationTest.cs
tonydnewell: Updated the comments clarify the expected behaviour

jtattermusch (approved):
LGTM.

jtattermusch:
re-running the tests.

jtattermusch:
@veblush  do you have the permission to merge to the 1.46.x branch? The test failures are unrelated.

PR #30539, opened 2022-08-08
Fork iomgr traced buffer list implementation for posix event engine

Vignesh2208:
This code will be used in upcoming PRs for TCP event engine endpoints


src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,305 @@
...
+  memset(info, 0, sizeof(*info));
+  info->length = offsetof(tcp_info, length);
+  return getsockopt(fd, IPPROTO_TCP, TCP_INFO, info, &(info->length));
+}
+}  // namespace.
+
+void TracedBuffer::AddNewEntry(TracedBuffer** head, uint32_t seq_no, int fd,
+                               void* arg) {
+  GPR_DEBUG_ASSERT(head != nullptr);
+  TracedBuffer* new_elem = new TracedBuffer(seq_no, arg);
```
drfloob, line 202: This static method is effectively a conjoined TracedBuffer constructor and a list append operation. I'd recommend splitting them up and protect the interfaces a bit better via encapsulation.

Consider something like:

```
std::list<TracedBuffer> traced_buffer list;
traced_buffer_list.emplace_front(seq_no, fd, arg);
```

... or similar with pointers if heap allocation is necessary here.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
```diff
@@ -0,0 +1,155 @@
...
+
+#include "src/core/lib/event_engine/posix_engine/internal_errqueue.h"
+#include "src/core/lib/iomgr/port.h"
+
+namespace grpc_event_engine {
+namespace posix_engine {
+
+struct ConnectionMetrics {
+  /* Delivery rate in Bytes/s. */
+  absl::optional<uint64_t> delivery_rate;
```
drfloob, line 34: Are there any subsets of these values that will always be present or absent in unison? If so, then a single `absl::has_value<SpecificXYZMetrics>(thing)` is going to be nicer than a handful of `absl::has_value<uint64_t>(x)` calls (and maybe save a few cycles).

Are there any that will always be present? If so, they don't need to be optional.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
```diff
@@ -0,0 +1,155 @@
...
+
+// TracedBuffer is a class to keep track of timestamps for a specific buffer in
+// the TCP layer. We are only tracking timestamps for Linux kernels and hence
+// this class would only be used by Linux platforms. For all other platforms,
+// TracedBuffer would be an empty class.
+// The timestamps collected are according to Timestamps declared above A
+// TracedBuffer list is kept track of using the head element of the list. If
+// *the head element of the list is nullptr, then the list is empty.
+#ifdef GRPC_LINUX_ERRQUEUE
+class TracedBuffer {
```
drfloob, line 108: As with most things, this needs tests.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,305 @@
...
+  constexpr int64_t cmsg_hdr_len = CMSG_ALIGN(sizeof(struct cmsghdr));
+  const int64_t len = opt_stats->cmsg_len - cmsg_hdr_len;
+  int64_t offset = 0;
+
+  while (offset < len) {
+    const auto* attr = reinterpret_cast<const nlattr*>(data + offset);
+    const void* val = data + offset + NLA_HDRLEN;
+    switch (attr->nla_type) {
+      case TCP_NLA_BUSY: {
+        metrics->busy_usec.emplace(ReadUnaligned<uint64_t>(val));
```
drfloob, line 111: assignment  (`=`) should be sufficient.

drfloob (changes requested):
Good start! Please let me know if you have any questions.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,305 @@
...
+  }
+  ptr->next_ = new_elem;
+}
+
+void TracedBuffer::ProcessTimestamp(TracedBuffer** head,
+                                    struct sock_extended_err* serr,
+                                    struct cmsghdr* opt_stats,
+                                    struct scm_timestamping* tss) {
+  GPR_DEBUG_ASSERT(head != nullptr);
+  TracedBuffer* elem = *head;
```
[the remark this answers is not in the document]
Vignesh2208: Updated to use std::list<TracedBuffer*>. Still doing heap allocations to avoid copying of TracedBuffer objects each time the list is popped

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,305 @@
...
+      case TCP_NLA_SND_SSTHRESH: {
+        metrics->snd_ssthresh.emplace(ReadUnaligned<uint32_t>(val));
+        break;
+      }
+    }
+    offset += NLA_ALIGN(attr->nla_len);
+  }
+}
+
+int GetSocketTcpInfo(tcp_info* info, int fd) {
```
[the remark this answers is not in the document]
Vignesh2208: I moved it to internal_errqueue{.h,.cc} files

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: I updated the method now to use lists. However it doesn't make sense to have two methods here because the TracedBuffer type is not used externally by the TCP code. So all the manipulations need to be done only within the static methods.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
Vignesh2208: Unfortunately I dont think thats true. Some kernel versions may support some values while others dont.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
Vignesh2208: I added a traced_buffer_list_test now.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: Fixed.

Vignesh2208 (commented):
Thanks. Updated per suggestions.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
drfloob: This is still a bit C-ish, I think you're missing an opportunity here to do away with leaked internal details via some better type design. Consider something like:

```
class TracedBufferList {
 public:
  void AddEntry(int32_t seq_no, int fd, void* arg);
  void ProcessTimestamp(struct sock_extended_err* serr,
                               struct cmsghdr* opt_stats,
                               struct scm_timestamping* tss);
  Shutdown();

 private:
  std::list<TracedBuffer> buffer_list;
}
```

There are no more static methods that operate on an object's internals, no more passing around pointers to pointers to heads of lists.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
drfloob: I'm wondering instead if a collection of named metrics would be better. You'd then be able to iterate across key-values, and it'll only contain the data that's actually available.

test/core/event_engine/posix/traced_buffer_list_test.cc
```diff
@@ -0,0 +1,126 @@
...
+  ASSERT_GT(ts->info.length, 0);
+  int* done = reinterpret_cast<int*>(arg);
+  *done = 1;
+}
+
+}  // namespace
+
+// Tests that all TracedBuffer elements in the list are flushed out on shutdown.
+// Also tests that arg is passed correctly.
+TEST(BufferListTest, Testshutdownflusheslist) {
```
drfloob, line 58: Test names are conventionally written in CamelCase.

test/core/event_engine/posix/traced_buffer_list_test.cc
```diff
@@ -0,0 +1,126 @@
...
+  }
+  TracedBuffer::Shutdown(traced_buffers, nullptr, absl::OkStatus());
+  for (auto i = 0; i < NUM_ELEM; i++) {
+    ASSERT_EQ(verifier_called[i], 1);
+  }
+  ASSERT_TRUE(traced_buffers.empty());
+}
+
+// Tests that the timestamp verifier is called on an ACK timestamp.
+TEST(BufferListTest, Testverifiercalledonack) {
```
drfloob, line 75: Test names are conventionally written in CamelCase.

test/core/event_engine/posix/traced_buffer_list_test.cc
```diff
@@ -0,0 +1,126 @@
...
+  int verifier_called = 0;
+  TracedBuffer::AddNewEntry(traced_buffers, 213, 0, &verifier_called);
+  TracedBuffer::ProcessTimestamp(traced_buffers, &serr, nullptr, &tss);
+  ASSERT_EQ(verifier_called, 1);
+  ASSERT_TRUE(traced_buffers.empty());
+  TracedBuffer::Shutdown(traced_buffers, nullptr, absl::OkStatus());
+}
+
+// Tests that shutdown can be called repeatedly.
+TEST(BufferListTest, Testrepeatedshutdown) {
```
drfloob, line 93: Test names are conventionally written in CamelCase.

test/core/event_engine/posix/traced_buffer_list_test.cc
```diff
@@ -0,0 +1,126 @@
...
+  TcpSetWriteTimestampsCallback(TestVerifierCalledOnAckVerifier);
+  std::list<TracedBuffer*> traced_buffers;
+  int verifier_called = 0;
+
+  TracedBuffer::AddNewEntry(traced_buffers, 213, 0, &verifier_called);
+  TracedBuffer::ProcessTimestamp(traced_buffers, &serr, nullptr, &tss);
+  ASSERT_EQ(verifier_called, 1);
+  ASSERT_TRUE(traced_buffers.empty());
+  TracedBuffer::Shutdown(traced_buffers, nullptr, absl::OkStatus());
+  TracedBuffer::Shutdown(traced_buffers, nullptr, absl::OkStatus());
```
drfloob, line 109: What behavior is this testing? Is there anything that can be asserted?

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  traced_buffers.push_back(new_elem);
+}
+
+void TracedBuffer::ProcessTimestamp(std::list<TracedBuffer*>& traced_buffers,
+                                    struct sock_extended_err* serr,
+                                    struct cmsghdr* opt_stats,
+                                    struct scm_timestamping* tss) {
+  auto it = traced_buffers.begin();
```
drfloob, line 217: This can be `auto& it`

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,289 @@
...
+    TracedBuffer* elem = (*it);
+    // The byte number refers to the sequence number of the last byte which this
+    // timestamp relates to.
+    if (serr->ee_data >= elem->seq_no_) {
+      switch (serr->ee_info) {
+        case SCM_TSTAMP_SCHED:
+          FillGprFromTimestamp(&(elem->ts_.scheduled_time.time), &(tss->ts[0]));
+          ExtractOptStatsFromCmsg(&(elem->ts_.scheduled_time.metrics),
+                                  opt_stats);
+          it++;
```
drfloob, line 228: According to the style guide, prefer pre-increment wherever possible. https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,289 @@
...
+          break;
+        case SCM_TSTAMP_ACK:
+          FillGprFromTimestamp(&(elem->ts_.acked_time.time), &(tss->ts[0]));
+          ExtractOptStatsFromCmsg(&(elem->ts_.acked_time.metrics), opt_stats);
+          // Got all timestamps. Do the callback and free this TracedBuffer. The
+          // thing below can be passed by value if we don't want the restriction
+          // on the lifetime.
+          g_timestamps_callback(elem->arg_, &(elem->ts_), absl::OkStatus());
+          it = traced_buffers.erase(it);
+          delete static_cast<TracedBuffer*>(elem);
```
drfloob, line 243: `elem` is already defined as a `TracedBuffer*`.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,289 @@
...
+        metrics->snd_ssthresh = ReadUnaligned<uint32_t>(val);
+        break;
+      }
+    }
+    offset += NLA_ALIGN(attr->nla_len);
+  }
+}
+}  // namespace.
+
+void TracedBuffer::AddNewEntry(std::list<TracedBuffer*>& traced_buffers,
```
drfloob, line 196: This can be a std::forward_list, actually, since it's never going backwards. Saves some memory.

And if that type was encapsulated inside a `TraceBufferList` class, you'd have a much smaller set of places where this change would need to be made.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: Done. Updated to this design.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
Vignesh2208: I feel that would require storing a string or some other type key and a value for each unique metric which might be expensive.

test/core/event_engine/posix/traced_buffer_list_test.cc
Vignesh2208: Fixed.

test/core/event_engine/posix/traced_buffer_list_test.cc
Vignesh2208: Fixed.

test/core/event_engine/posix/traced_buffer_list_test.cc
Vignesh2208: Fixed.

test/core/event_engine/posix/traced_buffer_list_test.cc
Vignesh2208: Added some asserts now.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: I dont think it can be forward_list because that type doesn't seem to support appending new entries to back of the list.  We have to do that to ensure sequence numbers are processed in order.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: This is giving a "Non-const lvalue reference to type '_List_iterator<...>' cannot bind to a temporary of type" compile time error

Vignesh2208:
Thanks for the review. Updated per suggestions.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
drfloob: I see, thanks. It's probably worth documenting that the list needs to be processed in a FIFO manner.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
```diff
@@ -0,0 +1,179 @@
...
+    void* arg_;       /* The arg to pass to timestamps_callback */
+    Timestamps ts_;   /* The timestamps corresponding to this buffer */
+  };
+  std::list<TracedBuffer*> buffer_list_;
+};
+
+#else  /* GRPC_LINUX_ERRQUEUE */
+class TracedBufferList {
+ public:
+  // Add a new entry in the TracedBuffer list pointed to by head. Also saves
```
drfloob, line 155: nit: these comments don't describe this class's operations. Would suggest removing them all from this no-op class, documenting that it's no-op, and leaving the comments on the supported implementation.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
drfloob: As discussed offline, this is fine as is. Keys could be enum classes to save some space, but I'm happy to not bikeshed this any further.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
drfloob: Looks good!

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -0,0 +1,289 @@
...
+        break;
+      }
+    }
+    offset += NLA_ALIGN(attr->nla_len);
+  }
+}
+}  // namespace.
+
+void TracedBufferList::AddNewEntry(int32_t seq_no, int fd, void* arg) {
+  TracedBuffer* new_elem = new TracedBuffer(seq_no, arg);
```
drfloob, line 201: nit: you can let the list manage the lifetimes by using a `std::list<TracedBuffer>` (non-pointer). `list::emplace` uses placement-new under the hood, and it would allow us to avoid having to call `delete` manually (and avoid us having to debug leaks when we failed to do so).

drfloob (commented):
This is looking good! I have just a few final suggestions for improvements, PTAL.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/traced_buffer_list.h
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: Done.

Vignesh2208:
Fixed per your suggestions.

drfloob (approved):

PR #30556, opened 2022-08-10
Reland: Add std::shared_ptr support to ChannelArgs, and precondition ChannelArgs with a default EventEngine

drfloob:
Relands https://github.com/grpc/grpc/pull/30128

```
Any class that is derived from std::enable_shared_from_this can now be 
stored as a shared_ptr inside ChannelArgs.

This is a step towards using ChannelArgs as the primary means of
accessing EventEngine instances in gRPC-core. If not explicitly provided
by the application, a default EventEngine will populated into
ChannelArgs during preconditioning.
```


src/core/lib/channel/channel_args.h
```diff
@@ -84,14 +85,14 @@ using RefType = absl::remove_cvref_t<decltype(*std::declval<T>().Ref())>;
 template <typename T>
 struct ChannelArgTypeTraits<
     T, absl::enable_if_t<
-           std::is_base_of<RefCounted<channel_args_detail::RefType<T>>,
-                           channel_args_detail::RefType<T>>::value ||
+           std::is_base_of<RefCounted<channel_args_detail::RefType<T> >,
```
ctiller, line 88: did clang-format do this?

ctiller (approved):

src/core/lib/channel/channel_args.h
drfloob: That may have been my IDE, I made some changes there. But clang_format_code doesn't fix it, interestingly.

drfloob:
This currently depends on https://github.com/grpc/grpc/pull/30619 landing first, but I believe I will split this PR in two: the shared_ptr ChannelArgs piece can land now, and the EventEngine preconditioning can land after #30619

PR #30557, opened 2022-08-11
Start to buildfix OpenBSD.

3405691582:
Some minor changes for the platform:

* `sys/types.h` needs to be included before `netinet/tcp.h` to provide
  definitions for `u_int8_t`, `u_int16_t`, and `u_int32_t`.

* cares has an OpenBSD-specific config header but is not actually hooked
  up in cares.BUILD. Remedy that.

### DIFF
--- a/src/core/lib/iomgr/sockaddr_utils_posix.cc
+++ b/src/core/lib/iomgr/sockaddr_utils_posix.cc
@@ -23,6 +23,8 @@
 #ifdef GRPC_POSIX_SOCKET_UTILS_COMMON
 
 #include "src/core/lib/iomgr/socket_utils.h"
+// sys/types.h must precede netinet/tcp.h for compatibility.
+#include <sys/types.h>
 #ifdef GRPC_LINUX_TCP_H
 #include <linux/tcp.h>
 #else
@@ -31,7 +33,6 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/socket.h>
-#include <sys/types.h>
 #include <unistd.h>
 
 #include <string>
--- a/third_party/cares/cares.BUILD
+++ b/third_party/cares/cares.BUILD
@@ -88,6 +88,11 @@ config_setting(
     values = {"cpu": "watchos_arm64_32"}
 )
 
+config_setting(
+    name = "openbsd",
+    values = {"cpu": "openbsd"},
+)
+
 copy_file(
     name = "ares_build_h",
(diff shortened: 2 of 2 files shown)


yashykt:
The BUILD change seems correct to me. Can you provide some more details/references for the header order change?

3405691582:
On this platform, `netinet/tcp.h` refers to types `u_int8_t`, `u_int16_t`, and `u_int32_t`. These types are provided by `sys/types.h` and must come before the include of `netinet/tcp.h`. 

This should not adversely impact any existing platforms since on those platforms the inclusion of `sys/types.h` is not necessary to precede `netinet/tcp.h`.

drfloob:
That sounds like a bug in the platform's `netinet/tcp.h` implementation. If it requires `sys/types.h`, it should probably include it. I found a [few](http://fxr.watson.org/fxr/source/netinet/tcp.h) [implementations](https://sites.uclouvain.be/SystInfo/usr/include/netinet/tcp.h.html) that do so.

3405691582:
I don't know if it makes sense to say that because a platform header doesn't compile on its own, that means that it's a platform bug and that making a header reordering tweak is not advisable. 

Ultimately, gRPC doesn't build on the platform as-is. Is this a rejection for the pr or would you prefer I add additional preprocessor guards to scope a different inclusion order for the platform?

drfloob:
> making a header reordering tweak is not advisable.

I have no objection to getting gRPC working on the platform, the PR looks fine. I'm just calling out the deficiency we're working around here. If there's a legitimate reason the header isn't included (i.e. to support alternative implementations of `u_int8_t` and its kin), that's fine, but I doubt it.

drfloob (approved):

src/core/lib/iomgr/sockaddr_utils_posix.cc
```diff
@@ -23,6 +23,7 @@
 #ifdef GRPC_POSIX_SOCKET_UTILS_COMMON
 
 #include "src/core/lib/iomgr/socket_utils.h"
+#include <sys/types.h>
```
yashykt, line 26: If such a heading order is necessary, let's add a comment around that effect. I'm pretty sure someone will go in and change the order again otherwise

3405691582: Done.

yashykt (approved):

drfloob:
All of the errors are unrelated. Thanks for the patch!

PR #30559, opened 2022-08-11
Tcp endpoint implementation for posix event engine.

Vignesh2208:
Fork of iomgr tcp endpoint code located in src/core/lib/iomgr/tcp_posix.cc. It includes a posix_endpoint_test which creates pairs of connected endpoints and writes random data at one end and verifies data read at the other end. It also tests these operations with zero-copy enabled and disabled and for both set of pollers: epoll and poll.


Vignesh2208:
The full posix event engine endpoint impl. Its kind of a big PR but its mostly just copy-paste. Could you ptal at your convenience. Thanks!

include/grpc/event_engine/slice_buffer.h
```diff
@@ -67,6 +68,11 @@ class SliceBuffer {
     return *this;
   }
 
+  // Swap the contents of this SliceBuffer with the contents of another.
```
drfloob, line 71: `///`

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,410 @@
...
+namespace {
+
+constexpr int kMinMessageSize = 1024;
+constexpr int kMaxMessageSize = 8192;
+constexpr int kNumConnections = 10;
+constexpr int kNumExchangedMessages = 100;
+std::atomic<int> g_num_active_connections{0};
+
+// Returns a random message with bounded length.
+std::string GetNextSendMessage() {
```
drfloob, line 71: This is copy-pasted from another test. Please deduplicate.

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,410 @@
...
+  }
+  tmp_s.reserve(len);
+  for (int i = 0; i < len; ++i) {
+    tmp_s += alphanum[rand() % (sizeof(alphanum) - 1)];
+  }
+  return tmp_s;
+}
+
+std::shared_ptr<EventEngine> GetPosixEE() {
+  static std::shared_ptr<EventEngine> posix_ee =
```
drfloob, line 95: This will keep an EventEngine alive through the end of the process, which can result in leaked threads under TSAN. I've seen it flake in similar situations, I'm not entirely sure why it isn't consistently failing.

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,410 @@
...
+      std::make_shared<PosixEventEngine>();
+  return posix_ee;
+}
+
+EventEngine* GetOracleEE() {
+  static EventEngine* oracle_ee = new PosixOracleEventEngine();
+  return oracle_ee;
+}
+
+class TestScheduler : public Scheduler {
```
drfloob, line 105: This is copy-pasted from another test. Please deduplicate.

test/core/event_engine/test_suite/event_engine_test_utils.cc
```diff
@@ -108,7 +119,10 @@ absl::Status SendValidatePayload(std::string data, Endpoint* send_endpoint,
   // Wait for async read to complete.
   GPR_ASSERT(read_promise.Get() == true);
   // Check if data written == data read
-  if (data != ExtractSliceBufferIntoString(&read_slice_buf)) {
+  std::string data_read = ExtractSliceBufferIntoString(&read_store_buf);
+  if (data != data_read) {
+    std::cout << "Data written = " << data << std::endl;
```
drfloob, line 124: why not use `gpr_log`? If you need to use iostream for some reason, then `std:cerr` should be used for error text.

test/core/event_engine/test_suite/BUILD
```diff
@@ -124,6 +124,7 @@ grpc_cc_library(
         "cpu:10",
         "no_windows",
     ],
+    visibility = ["@grpc:public"],
```
drfloob, line 127: This is too broad, no? Can it be `visibility = "tests"` or `visibility = ["//test:__subpackages__"]`?

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,410 @@
...
+  grpc_core::ChannelArgs args;
+  auto quota = grpc_core::ResourceQuota::Default();
+  args = args.Set(GRPC_ARG_RESOURCE_QUOTA, quota);
+  if (is_zero_copy_enabled) {
+    args = args.Set(GRPC_ARG_TCP_TX_ZEROCOPY_ENABLED, 1);
+    args = args.Set(GRPC_ARG_TCP_TX_ZEROCOPY_SEND_BYTES_THRESHOLD,
+                    kMinMessageSize);
+  }
+  ChannelArgsEndpointConfig config(args);
+  auto status = GetOracleEE()->CreateListener(
```
drfloob, line 150: Per a totw I linked you to earlier, I suggest:

```
auto listener = oracle_ee->CreateListener(...)
ASSERT_TRUE(listener.ok());
listener->Bind(...);
listener->Start(...);
```

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,410 @@
...
+  EventEngine* engine_;
+};
+
+std::list<std::tuple<std::unique_ptr<EventEngine::Endpoint>,
+                     std::unique_ptr<EventEngine::Endpoint>>>
+CreateConnectedEndpoints(PosixEventPoller* poller, bool is_zero_copy_enabled,
+                         int num_connections) {
+  EXPECT_NE(GetOracleEE(), nullptr);
+  EXPECT_NE(GetPosixEE(), nullptr);
+  EXPECT_NE(poller, nullptr);
```
drfloob, line 124: If the argument type was `const PosixEvenPoller& poller`, the compiler would enforce this constraint (modulo some funny business).

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,410 @@
...
+    int one = 1;
+    int flags;
+
+    client_fd = socket(AF_INET6, SOCK_STREAM, 0);
+    setsockopt(client_fd, SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one));
+    // Make fd non-blocking.
+    flags = fcntl(client_fd, F_GETFL, 0);
+    EXPECT_EQ(fcntl(client_fd, F_SETFL, flags | O_NONBLOCK), 0);
+
+    if (connect(client_fd,
```
drfloob, line 172: This appears to be the third copy of this code in the codebase. Please deduplicate.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1495 @@
...
+typedef GRPC_MSG_IOVLEN_TYPE msg_iovlen_type;
+#else
+typedef size_t msg_iovlen_type;
+#endif
+#define MAX_READ_IOVEC 64
+
+namespace grpc_event_engine {
+namespace posix_engine {
+
+using ::grpc_event_engine::experimental::EventEngine;
```
drfloob, line 76: these should be in the inner-most namespace.

drfloob (commented):
There's a lot to go through here, so I'll add my initial comments to unblock your progress, and continue to review.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,239 @@
...
+#include "src/core/lib/event_engine/posix_engine/tcp_socket_utils.h"
+#include "src/core/lib/event_engine/posix_engine/traced_buffer_list.h"
+#include "src/core/lib/iomgr/port.h"
+
+namespace grpc_event_engine {
+namespace posix_engine {
+
+#ifdef GRPC_POSIX_SOCKET_TCP
+
+using ::grpc_event_engine::experimental::EndpointConfig;
```
drfloob, line 37: using statements should not be in headers unless the intent is to expose these as public types owned by this file.  https://google.github.io/styleguide/cppguide.html#Aliases

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,239 @@
...
+using ::grpc_event_engine::experimental::EventEngine;
+using ::grpc_event_engine::experimental::SliceBuffer;
+
+class TcpZerocopySendCtx;
```
drfloob, line 41: This does not appear to require a forward declaration. Did I miss a circular dependency?

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,239 @@
...
+class TcpZerocopySendRecord;
```
drfloob, line 42: same

BUILD
```diff
@@ -2749,6 +2749,32 @@ grpc_cc_library(
     ],
 )
 
+grpc_cc_library(
+    name = "posix_event_engine_endpoint",
```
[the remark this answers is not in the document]
ctiller: please don't... rolling these things into single targets invites circular dependencies between layers over time that make the codebase incrementally hard to reason about. Always prefer creating the smallest target set that you can: our ambition should be one header, one cc file per bazel target.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1495 @@
...
+  //      |       |  v       sendmsg returns ENOBUFS               |
+  //      +-----> OPEN  ------------->>-------------------------> FULL
+  //                ^                                              |
+  //                |                                              |
+  //                | sendmsg completes                            |
+  //                +----<<---------- CHECK <-------<<-------------+
+  //                                        Tx-zero copy succeeds and there is
+  //                                        an active sendmsg
+  //
+  enum class OMemState : int8_t {
```
ctiller, line 419: idk what o means... please give this a full name

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1495 @@
...
+    OPEN,   // Everything is clear and omem is not full.
```
ctiller, line 420: kOpen,...

BUILD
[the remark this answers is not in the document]
drfloob: Don't we want the endpoint code to be hidden within the PosixEventEngine, though? Why would you want it to be reused outside the engine? I consider them an atomic unit, so maybe we don't need a separate header for it.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: I moved this to event_engine_test_utils{.h,.cc} files now

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: I removed this. I added a WaitForPendingTasks(std::shared_ptr<EventEngine> engine) function to event_engine_test_utils{.h,.cc} that spins until only one reference to the event engine shared ptr exists. This ensures safe shutdown.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: I created posix_engine_test_utils{.h,.cc} files and moved it there.

test/core/event_engine/test_suite/event_engine_test_utils.cc
Vignesh2208: Modified to use gpr_log.

test/core/event_engine/test_suite/BUILD
Vignesh2208: Fixed.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: I couldn't use ASSERT_TRUE because the function returns a non void value. But I renamed the variable.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: I think its better left as a pointer because its also captured inside a closure in line 133.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: Moved it to posix_engine_test_utils{.h,.cc} files into a function called ConnectToServerOrDie

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
Vignesh2208: Removed them.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
Vignesh2208: I moved the entire class definition over to the header file now.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
Vignesh2208: I moved the entire class definition of TcpZerocopySendRecord over to the header file now.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
Vignesh2208: It stands for optmem - I updated the name and also added a brief description in the comment above the enum definition.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc

BUILD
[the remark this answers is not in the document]
Vignesh2208: I also think it would be better to have a separate target for endpoint. It would allow us to even reuse the same endpoint with different event engines if needed.

Vignesh2208:
I made the first round of edits. Could you ptal! Thanks!

BUILD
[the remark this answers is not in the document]
drfloob: Alrighty, I withdraw my request here.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1135 @@
...
+
+  // TODO(ctiller): Check if supported by OS.
+  // TODO(ctiller): Allow some adjustments instead of hardcoding things.
+
+  static constexpr int kRcvLowatMax = 16 * 1024 * 1024;
+  static constexpr int kRcvLowatThreshold = 16 * 1024;
+
+  int remaining = std::min(static_cast<int>(incoming_buffer_->Length()),
+                           min_progress_size_);
+  remaining = std::min(remaining, kRcvLowatMax);
```
drfloob, line 363: I believe we can use the initializer list overload here: `std::min({1,2,3,4})`

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1135 @@
...
+    traced_buffer_mu_.Lock();
+    traced_buffers_.AddNewEntry(static_cast<uint32_t>(bytes_counter_ + length),
+                                fd_, outgoing_buffer_arg_);
+    traced_buffer_mu_.Unlock();
+    outgoing_buffer_arg_ = nullptr;
+  }
+  return true;
+}
+
+#else
```
drfloob, line 733: I believe this goes with the `GRPC_LINUX_ERRQUEUE` 230 lines earlier. Please comment here.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1135 @@
...
+  int one = 1;
+  if (setsockopt(fd_, SOL_TCP, TCP_INQ, &one, sizeof(one)) == 0) {
+    inq_capable_ = true;
+  } else {
+    gpr_log(GPR_DEBUG, "cannot set inq fd=%d errno=%d", fd_, errno);
+    inq_capable_ = false;
+  }
+#else
+  inq_capable_ = false;
+#endif /* GRPC_HAVE_TCP_INQ */
```
drfloob, line 1090: Please be consistent with comment styles, I'd suggest `#endif  // GRPC_HAVE_TCP_INQ` as is used elsewhere in this file.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+// the event engine is alive for the lifetime of the endpoint.
+std::unique_ptr<PosixEndpoint> CreatePosixEndpoint(
+    EventHandle* handle, PosixEngineClosure* on_shutdown,
+    std::shared_ptr<grpc_event_engine::experimental::EventEngine> engine,
+    const grpc_event_engine::experimental::EndpointConfig& config);
+
+}  // namespace posix_engine
+}  // namespace grpc_event_engine
+
+#endif  // GRPC_CORE_LIB_EVENT_ENGINE_POSIX_ENGINE_POSIX_ENDPOINT_H
```
drfloob, line 637: missing newline at the end of this file. Did the sanitizer pick that up?

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+#include "src/core/lib/gprpp/dual_ref_counted.h"
+#include "src/core/lib/gprpp/global_config.h"
+#include "src/core/lib/gprpp/notification.h"
+#include "src/core/lib/iomgr/port.h"
+#include "test/core/event_engine/posix/posix_engine_test_utils.h"
+#include "test/core/event_engine/test_suite/event_engine_test_utils.h"
+#include "test/core/event_engine/test_suite/oracle_event_engine_posix.h"
+#include "test/core/util/port.h"
+
+GPR_GLOBAL_CONFIG_DECLARE_STRING(grpc_poll_strategy);
```
drfloob, line 48: Is this safe to re-declare? I think this should live the source files somewhere, not in the tests. question for @ctiller

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+using Listener = ::grpc_event_engine::experimental::EventEngine::Listener;
+using namespace std::chrono_literals;
+
+constexpr int kMinMessageSize = 1024;
+constexpr int kNumConnections = 10;
+constexpr int kNumExchangedMessages = 100;
+std::atomic<int> g_num_active_connections{0};
+
+EventEngine* GetOracleEE() {
+  static EventEngine* oracle_ee = new PosixOracleEventEngine();
```
drfloob, line 73: Shouldn't TSAN complain about threads leaked off the end of the process if this engine is never destroyed?

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+using Endpoint = ::grpc_event_engine::experimental::EventEngine::Endpoint;
+using Listener = ::grpc_event_engine::experimental::EventEngine::Listener;
+using namespace std::chrono_literals;
+
+constexpr int kMinMessageSize = 1024;
+constexpr int kNumConnections = 10;
+constexpr int kNumExchangedMessages = 100;
+std::atomic<int> g_num_active_connections{0};
+
+EventEngine* GetOracleEE() {
```
drfloob, line 72: I believe this will make tests non-hermetic, since the oracle is being reused. I recommend a new oracle for every test.

test/core/event_engine/posix/posix_endpoint_test.cc
drfloob: Are you concerned about the object's lifetime, or the capture itself? You can capture it by reference for the same cost, I believe ... `[&poller] { poller.Kick(); }`. Regarding lifetimes, I don't think pointer or reference matters here since ownership is not transferred ... the caller must ensure it's alive until the callback is done.

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  static EventEngine* oracle_ee = new PosixOracleEventEngine();
+  return oracle_ee;
+}
+
+std::list<std::tuple<std::unique_ptr<EventEngine::Endpoint>,
```
drfloob, line 77: nit: consider a return struct instead, so accessing its members is easier to read.

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+    delete server_signal;
+    server_signal = new grpc_core::Notification();
+  }
+  delete server_signal;
+  return connections;
+}
+
+}  // namespace
+
+class TestParam {
```
drfloob, line 149: This looks like a struct to me, it has no owned behaviors, does not mutate its input, and exposes all of its state.

```
struct TestParam {
  std::string poller;
  bool is_zero_copy_enabled;
};
```

constructor optional.

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+    ASSERT_TRUE(result == Poller::WorkResult::kOk ||
+                result == Poller::WorkResult::kKicked);
+    // Corresponds to the Ref taken for the current instantiation. If the
+    // result was Poller::WorkResult::kKicked, then the next work instantiation
+    // would not have been scheduled and the poll_again callback would have
+    // been deleted.
+    Unref();
+  }
+  std::shared_ptr<EventEngine> engine_;
+  PosixEventPoller* poller_;
```
drfloob, line 206: poller is unowned here, right?

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  PosixEventPoller* PosixPoller() { return poller_; }
+
+ private:
+  PosixEventPoller* poller_;
+  std::unique_ptr<TestScheduler> scheduler_;
+  std::shared_ptr<EventEngine> posix_ee_;
+};
+
+TEST_P(PosixEndpointTest, ConnectExchangeBidiDataTransferTest) {
+  if (PosixPoller() == nullptr) {
```
drfloob, line 245: What is/are the valid scenario(s) in which this test should pass but the poller == nullptr? Should this be an assert?

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  }
+  Worker* worker = new Worker(GetPosixEE(), PosixPoller());
+  worker->Start();
+  {
+    auto connections = CreateConnectedEndpoints(
+        PosixPoller(), GetParam().IsZeroCopyEnabled(), 1, GetPosixEE());
+    auto it = connections.begin();
+    auto client_endpoint = std::move(std::get<0>(*it));
+    auto server_endpoint = std::move(std::get<1>(*it));
+    EXPECT_TRUE(client_endpoint != nullptr);
```
drfloob, line 256: EXPECT_NE

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+// Create  N connections and exchange and verify random number of messages over
+// each connection in parallel.
+TEST_P(PosixEndpointTest, MultipleIPv6ConnectionsToOneOracleListenerTest) {
+  if (PosixPoller() == nullptr) {
+    return;
+  }
+  Worker* worker = new Worker(GetPosixEE(), PosixPoller());
+  worker->Start();
+  auto connections =
+      CreateConnectedEndpoints(PosixPoller(), GetParam().IsZeroCopyEnabled(),
```
drfloob, line 287: Calling PosixPoller again creates a new instance, right?

test/core/event_engine/posix/posix_endpoint_test.cc
```diff
@@ -0,0 +1,352 @@
...
+  // Create one thread for each connection. For each connection, create
+  // 2 more worker threads: to exchange and verify bi-directional data transfer.
+  threads.reserve(kNumConnections);
+  for (int i = 0; i < kNumConnections; i++) {
+    // For each connection, simulate a parallel bi-directional data transfer.
+    // All bi-directional transfers are run in parallel across all connections.
+    auto it = connections.begin();
+    auto client_endpoint = std::move(std::get<0>(*it));
+    auto server_endpoint = std::move(std::get<1>(*it));
+    EXPECT_TRUE(client_endpoint != nullptr);
```
drfloob, line 299: EXPECT_NE

test/core/event_engine/posix/posix_engine_test_utils.cc
```diff
@@ -0,0 +1,59 @@
...
+#include "test/core/event_engine/posix/posix_engine_test_utils.h"
+
+#include <fcntl.h>
+#include <poll.h>
+
+namespace grpc_event_engine {
+namespace posix_engine {
+
+using ResolvedAddress =
+    grpc_event_engine::experimental::EventEngine::ResolvedAddress;
```
drfloob, line 24: please use full namespace qualification in using statements. Did the linter not catch this?

test/core/event_engine/test_suite/BUILD
drfloob: This does not appear to be fixed.

test/core/event_engine/test_suite/event_engine_test_utils.cc
```diff
@@ -44,6 +45,45 @@ using Listener = ::grpc_event_engine::experimental::EventEngine::Listener;
...
+  tmp_s.reserve(len);
+  for (int i = 0; i < len; ++i) {
+    tmp_s += alphanum[rand() % (sizeof(alphanum) - 1)];
+  }
+  return tmp_s;
+}
+
+// Waits until the use_count of the event engine shared_ptr has reached 1
+// and returns.
+void WaitForPendingTasks(std::shared_ptr<EventEngine>&& engine) {
```
drfloob, line 81: This presumes that all tasks own a shared_ptr to this EventEngine, which even in this PR is not the case. I'd suggest a rename to `WaitForSingleOwner` or some such, with the caveat that some callbacks _may still be running_ when this reaches its condition.

test/core/event_engine/test_suite/event_engine_test_utils.cc
```diff
@@ -44,6 +45,45 @@ using Listener = ::grpc_event_engine::experimental::EventEngine::Listener;
...
+    len = dis(gen);
+  }
+  tmp_s.reserve(len);
+  for (int i = 0; i < len; ++i) {
+    tmp_s += alphanum[rand() % (sizeof(alphanum) - 1)];
+  }
+  return tmp_s;
+}
+
+// Waits until the use_count of the event engine shared_ptr has reached 1
```
drfloob, line 79: This same commentary is in the header, so it's redundant here.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: PosixPoller will be nullptr if the specified poller (epoll/poll) is not supported on this platform. Were trying both pollers epoll and poll in these tests. Some platforms like MacOS dont support epoll pollers.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: No it just returns the same pointer. There is one poller instructed per test.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: Yes its not owner by worker. Rather its owned by the test which creates an instance of Worker

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+  bool AllSendRecordsEmpty() {
+    absl::MutexLock guard(&lock_);
+    return free_send_records_size_ == max_sends_;
+  }
+
+  bool Enabled() const { return enabled_; }
+
+  void SetEnabled(bool enabled) {
+    GPR_DEBUG_ASSERT(!enabled || !MemoryLimited());
+    enabled_ = enabled;
```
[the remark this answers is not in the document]
Vignesh2208: No its set only once at the beginning in the PosixEndpointImpl constructor. Every other access is a read. So it doesn't need to be lock protected.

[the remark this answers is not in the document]
drfloob: I see, that's fine then. You could improve the type's encapsulation by making this field's initialization part of the `TcpZerocopySendCtx` constructor (along with the setsockopt logic that controls it). Just a suggestion.

test/core/event_engine/posix/posix_endpoint_test.cc
drfloob: Doesn't the build system already enforce that epoll1 tests do not run on mac? The BUILD file shows that `posix_endpoint_test` is already configured with `uses_polling=True`, which means you'll have test targets like `posix_endpoint_test@poller=poll` and `posix_endpoint_test@poller=epoll1`. I'd suggest using the build system's tooling to enforce this rather than reinventing that logic in the test setup itself.

test/core/event_engine/posix/posix_endpoint_test.cc
drfloob: It's probably worth documenting the ownership and lifetime expectations since the type system isn't going to enforce it.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: The intent is to test the functionality of GetDefaultPoller in line 218 as well. It reads the poll_strategy environment variable and returns the appropriate poller. In the test setup, we override the poll_strategy environment variable and in line 220 we check whether the name of the poller returned by GetDefaultPoller matches the set environment variable.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+      grpc_event_engine::experimental::SliceBuffer& slices_to_send) {
+    AssertEmpty();
+    out_offset_.slice_idx = 0;
+    out_offset_.byte_idx = 0;
+    buf_.Swap(slices_to_send);
+    Ref();
+  }
+
+  // References: 1 reference per sendmsg(), and 1 for the tcp_write().
+  void Ref() { ref_.fetch_add(1, std::memory_order_relaxed); }
```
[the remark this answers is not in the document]
Vignesh2208: Actually for this particular object, we cant use the RefCounted types because the Unref operation doesn't delete the object. Rather it adds the object to a free_list maintained by TcpSendZeroCopyCtx object: https://github.com/grpc/grpc/pull/30559/files#diff-480c81a437473d5161dac9218be290d75ce42b686c1405a88e756c9fa8084a0aR756

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+ public:
+  TcpZerocopySendRecord() { buf_.Clear(); };
+
+  ~TcpZerocopySendRecord() { AssertEmpty(); }
+
+  // Given the slices that we wish to send, and the current offset into the
+  //   slice buffer (indicating which have already been sent), populate an iovec
+  //   array that will be used for a zerocopy enabled sendmsg().
+  msg_iovlen_type PopulateIovs(size_t* unwind_slice_idx,
+                               size_t* unwind_byte_idx, size_t* sending_length,
```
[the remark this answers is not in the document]
Vignesh2208: Added comments now.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: Added comment now.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: I simplified the testing logic per our offline discussion. But I'd prefer to have the nullptr check still to handle some corner cases. For e.g poller might be epoll, but the system may not support creating epoll_fds. In that case, GetDefaultPoller will return nullptr.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: Fixed it now in the latest commit to use a reference. I couldn't use a const reference though because the function calls poller.Kick() which cannot be marked const.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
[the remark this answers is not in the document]
Vignesh2208: Added more detailed comments.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+    return false;
+  }
+
+ private:
+  struct OutgoingOffset {
+    size_t slice_idx = 0;
+    size_t byte_idx = 0;
+  };
+
+  void AssertEmpty() {
```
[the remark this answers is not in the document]
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+  // enabled. This can result in several sendmsg() calls to flush all of the
+  // data to wire. Each sendmsg() takes a reference on the
+  // TcpZerocopySendRecord, and corresponds to a single sequence number.
+  // ReleaseSendRecord releases a reference on TcpZerocopySendRecord for a
+  // single sequence number. This is called either when we receive the relevant
+  // error queue notification (saying that we can discard the underlying
+  // buffers for this sendmsg()) is received from the kernel - or, in case
+  // sendmsg() was unsuccessful to begin with.
+  TcpZerocopySendRecord* ReleaseSendRecord(uint32_t seq) {
+    absl::MutexLock guard(&lock_);
```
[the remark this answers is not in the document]
Vignesh2208: Done.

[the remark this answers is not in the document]
Vignesh2208: Done.

test/core/event_engine/posix/posix_endpoint_test.cc
Vignesh2208: Fixed.

test/core/event_engine/posix/posix_engine_test_utils.cc
Vignesh2208: Fixed.

test/core/event_engine/test_suite/event_engine_test_utils.cc
Vignesh2208: Renamed to WaitForSingleOwner

test/core/event_engine/test_suite/event_engine_test_utils.cc
Vignesh2208: Removed it.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+                           int additional_flags);
+#ifdef GRPC_LINUX_ERRQUEUE
+  // Reads \a cmsg to process zerocopy control messages.
+  bool ProcessErrors();
+  void ProcessZerocopy(struct cmsghdr* cmsg);
+  // Reads \a cmsg to derive timestamps from the control messages.
+  struct cmsghdr* ProcessTimestamp(msghdr* msg, struct cmsghdr* cmsg);
+#endif
+  absl::Mutex read_mu_;
+  absl::Mutex traced_buffer_mu_;
```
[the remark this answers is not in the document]
Vignesh2208: Added in the latest commit

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+                           ssize_t* sent_length, int* saved_errno,
+                           int additional_flags);
+#ifdef GRPC_LINUX_ERRQUEUE
+  // Reads \a cmsg to process zerocopy control messages.
+  bool ProcessErrors();
+  void ProcessZerocopy(struct cmsghdr* cmsg);
+  // Reads \a cmsg to derive timestamps from the control messages.
+  struct cmsghdr* ProcessTimestamp(msghdr* msg, struct cmsghdr* cmsg);
+#endif
+  absl::Mutex read_mu_;
```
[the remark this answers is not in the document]
Vignesh2208: Added in the latest commit

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+  bool WriteWithTimestamps(struct msghdr* msg, size_t sending_length,
+                           ssize_t* sent_length, int* saved_errno,
+                           int additional_flags);
+#ifdef GRPC_LINUX_ERRQUEUE
+  // Reads \a cmsg to process zerocopy control messages.
+  bool ProcessErrors();
+  void ProcessZerocopy(struct cmsghdr* cmsg);
+  // Reads \a cmsg to derive timestamps from the control messages.
+  struct cmsghdr* ProcessTimestamp(msghdr* msg, struct cmsghdr* cmsg);
+#endif
```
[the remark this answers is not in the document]
Vignesh2208: Fixed in the latest commit

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+  bool TcpFlushZerocopy(TcpZerocopySendRecord* record, absl::Status& status);
+  bool TcpFlush(absl::Status& status);
+  void TcpShutdownTracedBufferList();
+  void UnrefMaybePutZerocopySendRecord(TcpZerocopySendRecord* record);
+  void ZerocopyDisableAndWaitForRemaining();
+  bool WriteWithTimestamps(struct msghdr* msg, size_t sending_length,
+                           ssize_t* sent_length, int* saved_errno,
+                           int additional_flags);
+#ifdef GRPC_LINUX_ERRQUEUE
+  // Reads \a cmsg to process zerocopy control messages.
```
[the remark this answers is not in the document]
Vignesh2208: Fixed in the latest commit

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -0,0 +1,637 @@
...
+  }
+  const grpc_event_engine::experimental::EventEngine::ResolvedAddress&
+  GetLocalAddress() const {
+    return local_address_;
+  }
+
+  void MaybeShutdown(absl::Status why);
+
+ private:
+  void Ref() { ref_count_.fetch_add(1, std::memory_order_relaxed); }
```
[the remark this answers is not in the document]
Vignesh2208: Yes. Replaced it to use grpc_core::RefCounted

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
Vignesh2208: Modified to use this in the latest commit

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
Vignesh2208: Add a comment.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
Vignesh2208: Fixed it now.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1135 @@
...
+  if (poller_->CanTrackErrors()) {
+    Ref();
+    handle_->NotifyOnError(on_error_);
+  }
+}
+
+std::unique_ptr<PosixEndpoint> CreatePosixEndpoint(
+    EventHandle* handle, PosixEngineClosure* on_shutdown,
+    std::shared_ptr<EventEngine> engine, const EndpointConfig& config) {
+  GPR_ASSERT(handle != nullptr);
```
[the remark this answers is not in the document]
Vignesh2208: Handle is created by the poller and its ownership is transferred to the endpoint because it doesnt need to be kept around by the code which created it. I updated the comment in CreatePosixEndpoint in the header file to make this clear. So I think we can keep the handle as a pointer.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1135 @@
...
+  GPR_ASSERT(engine != nullptr);
```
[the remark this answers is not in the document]
Vignesh2208: Removed it.

Vignesh2208:
Fixed your previous round of comments. Could you ptal. Thanks!

include/grpc/event_engine/slice_buffer.h
```diff
@@ -88,6 +94,17 @@ class SliceBuffer {
     grpc_slice_buffer_move_first_into_buffer(&slice_buffer_, n, dst);
   }
 
+  /// Removes/deletes the last n bytes in the SliceBuffer and add it to the
+  /// other SliceBuffer
+  void RemoveLastNBytesIntoSliceBuffer(size_t n, SliceBuffer& other) {
```
drfloob, line 99: nit: would rename this to `MoveLastNBytesIntoSliceBuffer` to match the next method, which does the same thing on the first N bytes.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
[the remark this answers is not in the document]
drfloob: Gotcha, thanks.

drfloob (approved):

include/grpc/event_engine/slice_buffer.h
Vignesh2208: Renamed.

PR #30560, opened 2022-08-11
implementation of library in ansi C language

franalta:
I don't know to who ask this request, however I'll make it: it's possible make a library in ansi C language? Actually only core shared is in C language


yashykt:
It's not clear what this PR is trying to achieve.

jtattermusch:
This PR looks like a mistake when filing an issue. Closing.

jtattermusch:
We don't officially support a C-only implementation of gRPC. This is a topic that has been asked many times before, so I suggest googling for existing answers to this question and what the available options are.

PR #30562, opened 2022-08-11
Drop Python 3.6 From Distribtests

gnossen:
- CentOS has been deprecated and the latest version does not support 3.7 or later, so dropping it.
- OpenSuse's latest version also seemingly does not support 3.7, so dropping it as well.
- Ubuntu 18.04 also only supports up to 3.6, so updating to 20.04.


gnossen:
[Manual triggered run](https://sponge2.corp.google.com/f2a91c68-de14-4515-a1f3-b40d45ed020b)

gnossen:
[And another](https://sponge2.corp.google.com/6d71d6eb-ac43-415d-a8c0-1f8d3460f687)

gnossen:
Manual trigger is green. Looks like we're good to go.

yashykt (approved):

yashykt:
should we use ubuntu 22 instead?

gnossen:
> should we use ubuntu 22 instead?

My thinking was that I was only updating the _oldest_ version. I think we should actually have both. But I'll save adding 22 as well for a follow-up.

PR #30570, opened 2022-08-12
Build Target Reorganization Part 2

ralphchung:

```
┌───────┐┌───────────────┐        
│grpc++ ││grpc++_unsecure│        
└┬─────┬┘└┬─────────┬────┘        
┌▽───┐┌▽──▽───────┐┌▽────────────┐
│grpc││grpc++_base││grpc_unsecure│
└┬───┘└┬──────────┘└┬────────────┘
┌▽─────▽────────────▽┐            
│grpc_base           │            
└────────────────────┘            
```

* (Covered in #30518 ) `gpr_base` is merged into `gpr`. The `gpr` here itself is nothing but depending on `gpr_base`. I don't think we need to separate them.
* `grpc++_internals` and `grpc++_internal_hdrs_only` are merged into `grpc++`. The reason is similar. `grpc++` does not contain much stuff but `grpc++_internals`.
* (Covered in #30518 ) `gpr_codegen` is merged to `gpr` since the files in `gpr_codegen` should be removed in the future and be moved to `gpr`.
* (Covered in this PR) `grpc_codegen` is merged to `grpc_base` with the similar reason as above.
* `grpc++_codegen_base` and `grpc++_codegen_base_src` are merged to `grpc++_base` with the similar reason as above.
* (Covered in this PR) `grpc_secure` is merged to `grpc` since `grpc` is by default secure. We do not need to separate them.
* `grpc++_base_unsecure` is merged either into `grpc++_base` or into `grpc++_unsecure`.

This PR merges `grpc_secure` and `grpc_codegen` to `grpc` without removing `grpc_codegen` for temporary compatibility. It should be removed in the following PRs.


BUILD
```diff
@@ -468,7 +474,6 @@ grpc_cc_library(
             "//conditions:default": GRPC_XDS_TARGETS,
         },
     ],
-    standalone = True,
```
ctiller, line 471: what does this change?

ralphchung: Nothing.

I found the parameter is unused in both Github and google3.

https://github.com/grpc/grpc/blob/master/bazel/grpc_build_system.bzl#L145
https://source.corp.google.com/piper///depot/google3/third_party/grpc/bazel/grpc_build_system.bzl;l=657

ctiller (approved):

PR #30571, opened 2022-08-12
GCP Observability: Add plugin registry API

yashykt:
Just adds the scaffolding for now in interest of doing it in a piece-wise manner. Proves that the integration with the stackdriver exporter works after https://github.com/grpc/grpc/pull/30567

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -7262,7 +7262,7 @@ grpc_cc_library(
         "include/grpcpp/opencensus.h",
     ],
     tags = ["nofixdeps"],
-    visibility = ["@grpc:public"],
+    visibility = ["//test:__subpackages__"],
     deps = [
         "gpr",
         "grpc_opencensus_plugin",
--- a/BUILD
+++ b/BUILD
@@ -7247,6 +7247,28 @@ grpc_cc_library(
     ],
 )
 
+# This is an EXPERIMENTAL target subject to change.
+grpc_cc_library(
+    name = "grpcpp_opencensus_gcp_observability_plugin",
+    srcs = [
+        "src/cpp/ext/filters/census/gcp_observability.cc",
+    ],
+    external_deps = [
+        "opencensus-trace-stackdriver_exporter",
+        "opencensus-stats-stackdriver_exporter",
+    ],
+    language = "c++",
+    public_hdrs = [
+        "include/grpcpp/opencensus.h",
+    ],
+    tags = ["nofixdeps"],
+    visibility = ["@grpc:public"],
+    deps = [
+        "gpr",
+        "grpc_opencensus_plugin",
+    ],
+)
+
 grpc_cc_library(
     name = "json",
     srcs = [
--- a/BUILD
+++ b/BUILD
@@ -7249,18 +7249,19 @@ grpc_cc_library(
 
 # This is an EXPERIMENTAL target subject to change.
 grpc_cc_library(
-    name = "grpcpp_opencensus_gcp_observability_plugin",
+    name = "grpcpp_gcp_observability",
     srcs = [
-        "src/cpp/ext/filters/census/gcp_observability.cc",
+        "src/cpp/ext/gcp_observability/gcp_observability.cc",
+    ],
+    hdrs = [
+        "src/cpp/ext/gcp_observability/gcp_observability.h",
     ],
     external_deps = [
+        "opencensus-trace",
         "opencensus-trace-stackdriver_exporter",
         "opencensus-stats-stackdriver_exporter",
     ],
     language = "c++",
-    public_hdrs = [
-        "include/grpcpp/opencensus.h",
-    ],
     tags = ["nofixdeps"],
     visibility = ["//test:__subpackages__"],
     deps = [
--- a/include/grpcpp/opencensus.h
+++ b/include/grpcpp/opencensus.h
@@ -42,6 +42,13 @@ void RegisterOpenCensusViewsForExport();
 // Returns the tracing Span for the current RPC.
 ::opencensus::trace::Span GetSpanFromServerContext(ServerContext* context);
 
+namespace experimental {
+// Setup gRPC's OpenCensus plugin to export monitoring and trace information to
+// stackdriver. Users of this API should not call `RegisterOpenCensusPlugin()`
+// or `RegisterOpenCensusViewsForExport()` in addition to this call.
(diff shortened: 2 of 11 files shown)


include/grpcpp/opencensus.h
```diff
@@ -42,6 +42,13 @@ void RegisterOpenCensusViewsForExport();
 // Returns the tracing Span for the current RPC.
 ::opencensus::trace::Span GetSpanFromServerContext(ServerContext* context);
 
+namespace experimental {
+// Setup gRPC's OpenCensus plugin to export monitoring and trace information to
+// stackdriver. Users of this API should not call `RegisterOpenCensusPlugin()`
+// or `RegisterOpenCensusViewsForExport()` in addition to this call.
+void RegisterOpenCensusPluginForGcpObservability();
```
sanjaypujare, line 49: Is this what you will expose to the end users? In Java (and Go) we just have something like `grpcInit()` that users call which initializes everything for o11y. Will your `grpcInit` be implemented later?

BUILD
```diff
@@ -7247,6 +7247,28 @@ grpc_cc_library(
     ],
 )
 
+# This is an EXPERIMENTAL target subject to change.
+grpc_cc_library(
+    name = "grpcpp_opencensus_gcp_observability_plugin",
```
sanjaypujare, line 7252: I know we have called this observability code as plugin in various places but it is not really based on any plugin arch nor is it pluggable AFAICS. Again the name as opencensus in it whereas I thought we are going to hide that

sanjaypujare: As discussed offline: the goal of observability is to hide opencensus (or OTel in future) as implementation details but the C++ app developer is only aware of the grpc observability API.

src/cpp/ext/filters/census/gcp_observability.cc
```diff
@@ -0,0 +1,56 @@
...
+constexpr uint32_t kMaxAnnotations = 128;
+constexpr uint32_t kMaxMessageEvents = 128;
+constexpr uint32_t kMaxLinks = 128;
+}  // namespace
+
+void RegisterOpenCensusPluginForGcpObservability() {
+  // TODO(yashykt): Add code for gRPC config parsing
+  grpc::RegisterOpenCensusPlugin();
+  grpc::RegisterOpenCensusViewsForExport();
+  // TODO(yashykt: Get probability from config
```
sanjaypujare, line 41: And also whether to enable/disable traces.

src/cpp/ext/filters/census/gcp_observability.cc
```diff
@@ -0,0 +1,56 @@
...
+  opencensus::trace::TraceConfig::SetCurrentTraceParams(
+      {kMaxAttributes, kMaxAnnotations, kMaxMessageEvents, kMaxLinks,
+       opencensus::trace::ProbabilitySampler(1.0)});
+  opencensus::exporters::trace::StackdriverOptions trace_opts;
+  // TODO(yashykt): Set up project ID based on config
+  opencensus::exporters::trace::StackdriverExporter::Register(
+      std::move(trace_opts));
+  opencensus::exporters::stats::StackdriverOptions stats_opts;
+  // TODO(yashykt): Set up project ID based on config
```
sanjaypujare, line 50: And also only if stats (metrics) is enabled

src/cpp/ext/filters/census/gcp_observability.cc
```diff
@@ -0,0 +1,56 @@
...
+// TODO(yashykt): These constants are currently derived from the example at
+// https://cloud.google.com/traffic-director/docs/observability-proxyless#c++.
+// We might want these to be configurable.
+constexpr uint32_t kMaxAttributes = 128;
+constexpr uint32_t kMaxAnnotations = 128;
+constexpr uint32_t kMaxMessageEvents = 128;
+constexpr uint32_t kMaxLinks = 128;
+}  // namespace
+
+void RegisterOpenCensusPluginForGcpObservability() {
```
sanjaypujare, line 37: This should be named as `GcpObservability_grpcInit` or such

sanjaypujare (changes requested):
Added comments requesting changes

include/grpcpp/opencensus.h
yashykt: Moved to `GcpObservabiltyInit`

BUILD
yashykt: Renamed to `grpcpp_gcp_observability`

src/cpp/ext/filters/census/gcp_observability.cc
yashykt: Done

src/cpp/ext/filters/census/gcp_observability.cc
yashykt: Added TODO

src/cpp/ext/filters/census/gcp_observability.cc
yashykt: Added TODO

src/cpp/ext/gcp_observability/gcp_observability.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include <grpcpp/opencensus.h>
+
+namespace grpc {
+namespace experimental {
+
+namespace {
+// TODO(yashykt): These constants are currently derived from the example at
+// https://cloud.google.com/traffic-director/docs/observability-proxyless#c++.
+// We might want these to be configurable.
```
sanjaypujare, line 38: @DNVindhya we don't configure these in Java right? Why are there max values for attrs, annotations, events etc? Do we set them to the same values i.e. 128 for all 3?

sanjaypujare (approved):
LGTM (with one clarifying question added)

src/cpp/ext/gcp_observability/gcp_observability.cc
DNVindhya: Yes, in Java we only configure the trace sampler based on Observability Config and for rest we use [default](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/config/TraceParams.java#L50) values.

sanjaypujare: So only events value is 128 and the other 2 are 32 each. Ideally we should have the same default values for all languages but right now it's not a blocker for this PR

PR #30572, opened 2022-08-12
Ensure iomgr does not track Forkable threads

drfloob:
New uses of grpc_core::Thread should not be tracked by iomgr. It would
be nice to enforce this, but that can come later. This should unblock
python releases.


gnossen:
We have confirmation from the Beam team that this fixes the issue. Thanks for all the work on this!

drfloob:
All the test failures are unrelated, this is ready to merge.

gnossen (approved):

PR #30589, opened 2022-08-15
Add a fixed minimum connection timeout for testing

drfloob:
This decouples the backoff reconnect timeout from the connection timeout when fixed-length timeouts are set in the end2end tests. If we can do away with the fixed timeouts altogether, then that solution can probably supersede this one.


drfloob:
Closed in favor of #30590

PR #30590, opened 2022-08-15
[fixit] Core End2End Tests: Increase the connect timeout

yashykt:
In the case of tls fixtures, the 1 second connect timeout does not seem to be enough for some configurations.

So, instead of the testing arg "grpc.testing.fixed_reconnect_backoff_ms" which uses the same value for both the backoff and the connection timeout, use the other args to separate out the two.

### DIFF
--- a/test/core/end2end/goaway_server_test.cc
+++ b/test/core/end2end/goaway_server_test.cc
@@ -247,27 +247,23 @@ int main(int argc, char** argv) {
 
   std::string addr;
 
-  grpc_arg arg_array[] = {
-      grpc_channel_arg_integer_create(
-          const_cast<char*>(GRPC_ARG_INITIAL_RECONNECT_BACKOFF_MS), 1000),
-      grpc_channel_arg_integer_create(
-          const_cast<char*>(GRPC_ARG_MAX_RECONNECT_BACKOFF_MS), 1000),
-      grpc_channel_arg_integer_create(
-          const_cast<char*>(GRPC_ARG_MIN_RECONNECT_BACKOFF_MS), 5000),
-      /* When this test brings down server1 and then brings up server2,
-       * the targetted server port number changes, and the client channel
-       * needs to re-resolve to pick this up. This test requires that
-       * happen within 10 seconds, but gRPC's DNS resolvers rate limit
-       * resolution attempts to at most once every 30 seconds by default.
-       * So we tweak it for this test. */
-      grpc_channel_arg_integer_create(
-          const_cast<char*>(GRPC_ARG_DNS_MIN_TIME_BETWEEN_RESOLUTIONS_MS),
-          1000)};
-  grpc_channel_args client_args = {GPR_ARRAY_SIZE(arg_array), arg_array};
+  auto client_args =
+      grpc_core::ChannelArgs()
+          .Set(GRPC_ARG_INITIAL_RECONNECT_BACKOFF_MS, 1000)
+          .Set(GRPC_ARG_MAX_RECONNECT_BACKOFF_MS, 1000)
+          .Set(GRPC_ARG_MIN_RECONNECT_BACKOFF_MS, 5000)
+          /* When this test brings down server1 and then brings up server2,
+           * the targetted server port number changes, and the client channel
+           * needs to re-resolve to pick this up. This test requires that
+           * happen within 10 seconds, but gRPC's DNS resolvers rate limit
+           * resolution attempts to at most once every 30 seconds by default.
+           * So we tweak it for this test. */
+          .Set(GRPC_ARG_DNS_MIN_TIME_BETWEEN_RESOLUTIONS_MS, 1000)
+          .ToC();
 
   /* create a channel that picks first amongst the servers */
   grpc_channel_credentials* creds = grpc_insecure_credentials_create();
-  grpc_channel* chan = grpc_channel_create("test", creds, &client_args);
+  grpc_channel* chan = grpc_channel_create("test", creds, client_args.get());
   grpc_channel_credentials_release(creds);
   /* and an initial call to them */
   grpc_slice host = grpc_slice_from_static_string("127.0.0.1");
--- a/test/core/end2end/goaway_server_test.cc
+++ b/test/core/end2end/goaway_server_test.cc
@@ -247,24 +247,23 @@ int main(int argc, char** argv) {
 
   std::string addr;
 
-  grpc_channel_args client_args;
-  grpc_arg arg_array[2];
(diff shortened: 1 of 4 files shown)


test/core/end2end/tests/connectivity.cc
```diff
@@ -65,14 +66,18 @@ static void test_connectivity(grpc_end2end_test_config config) {
   grpc_core::CqVerifier cqv(f.cq);
   child_events ce;
 
-  grpc_channel_args client_args;
-  grpc_arg arg_array[1];
+  grpc_arg arg_array[] = {
```
drfloob, line 69: These values are ignored if `fixed_reconnect_backoff_ms` is set.

test/core/end2end/tests/connectivity.cc
```diff
@@ -65,14 +66,18 @@ static void test_connectivity(grpc_end2end_test_config config) {
...
+      grpc_channel_arg_integer_create(
+          const_cast<char*>(GRPC_ARG_INITIAL_RECONNECT_BACKOFF_MS), 1000),
+      grpc_channel_arg_integer_create(
+          const_cast<char*>(GRPC_ARG_MAX_RECONNECT_BACKOFF_MS), 1000),
+      grpc_channel_arg_integer_create(
+          const_cast<char*>(GRPC_ARG_MIN_RECONNECT_BACKOFF_MS), 5000)};
   arg_array[0].type = GRPC_ARG_INTEGER;
```
drfloob, line 76: This overwrites the first value in the fixed list.

test/core/end2end/tests/connectivity.cc
```diff
@@ -135,9 +140,9 @@ static void test_connectivity(grpc_end2end_test_config config) {
      READY is reached */
   while (state != GRPC_CHANNEL_READY) {
     grpc_channel_watch_connectivity_state(
-        f.client, state, grpc_timeout_seconds_to_deadline(3), f.cq, tag(4));
+        f.client, state, grpc_timeout_seconds_to_deadline(10), f.cq, tag(4));
```
drfloob, line 143: With respect to the flakiness bug we're solving, this change made no difference in my testing.

test/core/end2end/tests/connectivity.cc
```diff
@@ -116,7 +121,7 @@ static void test_connectivity(grpc_end2end_test_config config) {
 
   /* quickly followed by a transition to TRANSIENT_FAILURE */
   grpc_channel_watch_connectivity_state(f.client, GRPC_CHANNEL_CONNECTING,
-                                        grpc_timeout_seconds_to_deadline(3),
+                                        grpc_timeout_seconds_to_deadline(10),
```
drfloob, line 124: With respect to the flakiness bug we're solving, this change made no difference in my testing.

drfloob (commented):
Something like this could work! If the plan is to do away with the fixed test values, I'm thinking we should remove their handling from the subchannel code as well. And this same fix would need to be done in a few tests - I prototyped a fix in max_connection_idle, but I think there are 4 or 5.

test/core/end2end/tests/connectivity.cc
yashykt: I increased since we increased the timeout. Anything above 5 should be fine here imo

test/core/end2end/tests/connectivity.cc
yashykt: I increased since we increased the timeout. Anything above 5 should be fine here imo

test/core/end2end/tests/connectivity.cc
yashykt: lol, no wonder my tests were failing, fixed

test/core/end2end/tests/connectivity.cc
yashykt: fixed

yashykt:
> Something like this could work! If the plan is to do away with the fixed test values, I'm thinking we should remove their handling from the subchannel code as well. And this same fix would need to be done in a few tests - I prototyped a fix in max_connection_idle, but I think there are 4 or 5.

About removing the handling from the subchannel code, I have seen code in prod use this test arg unfortunately.

drfloob:
> About removing the handling from the subchannel code, I have seen code in prod use this test arg unfortunately.

I can't think of a better reason to remove it. Using `grpc.testing.fixed_reconnect_backoff_ms` in production really shouldn't be a thing. If the functionality is impossible to remove, then we should provide a legitimate prod path to accomplish it. But we can do that separately.

yashykt:
> I can't think of a better reason to remove it. Using grpc.testing.fixed_reconnect_backoff_ms in production really shouldn't be a thing. If the functionality is impossible to remove, then we should provide a legitimate prod path to accomplish it. But we can do that separately.

I agree. All I'm saying is that there is a bit of legwork there to be done, and a potential breaking change announcement.

test/core/end2end/goaway_server_test.cc
```diff
@@ -247,24 +247,23 @@ int main(int argc, char** argv) {
...
-   * happen within 10 seconds, but gRPC's DNS resolvers rate limit
-   * resolution attempts to at most once every 30 seconds by default.
-   * So we tweak it for this test. */
-  arg_array[1].type = GRPC_ARG_INTEGER;
-  arg_array[1].key =
-      const_cast<char*>(GRPC_ARG_DNS_MIN_TIME_BETWEEN_RESOLUTIONS_MS);
-  arg_array[1].value.integer = 1000;
-  client_args.args = arg_array;
-  client_args.num_args = 2;
+  grpc_arg arg_array[] = {
```
ctiller, line 250: going to just suggest:

```
auto channel_args = grpc_core::ChannelArgs()
  .Set(GRPC_ARG_INITIAL_RECONNECT_BACKOFF_MS, 1000)
  .Set(GRPC_ARG_MAX_RECONNECT_BACKOFF_MS, 1000)
  .Set(GRPC_ARG_MIN_RECONNECT_BACKOFF_MS, 5000)
  .Set(GRPC_ARG_DNS_MIN_TIME_BETWEEN_RESOLUTIONS_MS, 1000)
  .ToC();
  
// ...
// use as: client_args.get()
```

ctiller: (and possibly packing it all up into a helper in test_utils.h)

yashykt: Accepted suggestion to use new API. 

Pushing back on the suggestion to move it to a helper, since this configuration has an impact on these tests (they are testing connectivity itself).

ctiller (approved):

PR #30593, opened 2022-08-16
[fixit] Increase timeout value for XDS Core End2End Test

ralphchung:
The current low timeout value causes test flakiness. It was suggested that the timeout value should be in seconds.

This PR changes the timeout value in `GlobalXdsClientTest.MultipleChannelsShareXdsClient` and `GlobalXdsClientTest.MultipleChannelsShareXdsClientWithResourceUpdateAfterOneChannelGoesAway` from 100 ms to 1 sec.


yashykt (approved):
I would prefer a 5 sec timeout, but this is fine too if the flakiness has disappeared

PR #30594, opened 2022-08-16
xDS interop: collect pod logs

sergiitk:
- Added support for pod log collection. To enable, set `--collect_app_logs` flag, and specify `--log_dir`.
- Added support and helpers for operating on the `--log_dir` (natively provided by absl)
- Added support for `--follow` to `bin/run_test_server.py` and `bin/run_test_client.py` to follow pod logs printed to stdout
- Moved `PortForwarder` from k8s.py to its own file

The collection itself will be enabled per-suite in https://github.com/grpc/grpc/pull/30735.


sergiitk:
- [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/638059f3-e055-4666-9d99-4b2f3612c0a8)
- [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/1cf5b625-630d-45e0-be62-35d218fd0139)
- [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/be6e1a8c-6022-4d56-a432-f1c88b0d9902)

tools/run_tests/xds_k8s_test_driver/framework/xds_flags.py
```diff
@@ -127,6 +127,13 @@
     default=False,
     help="Force resource cleanup, even if not created by this test run")
 
+COLLECT_APP_LOGS = flags.DEFINE_bool(
+    'collect_app_logs',
+    default=False,
```
gnossen, line 132: Why not turn this on by default?

tools/run_tests/xds_k8s_test_driver/framework/test_app/runners/base_runner.py
```diff
@@ -16,15 +16,55 @@
...
+            raise FileNotFoundError('Log collection is not enabled.')
+        return self._logs_subdir
+
+    @property
+    def log_stop_event(self) -> threading.Event:
+        if not self.should_collect_logs:
+            raise ValueError('Log collection is not enabled.')
+        return self._log_stop_event
+
+    def stop_logging_if_needed(self):
```
gnossen, line 65: Nit: `maybe_stop_logging` is a more established naming pattern.

tools/run_tests/xds_k8s_test_driver/framework/infrastructure/k8s_internal/k8s_log_collector.py
```diff
@@ -0,0 +1,129 @@
...
+import threading
+from typing import Any, Callable, Optional, TextIO
+
+from kubernetes import client
+from kubernetes.watch import watch
+
+logger = logging.getLogger(__name__)
+
+
+class PodLogCollector(threading.Thread):
```
gnossen, line 26: There are [a bunch of landmines](https://stackoverflow.com/questions/660961/overriding-python-threading-thread-run) when subclassing `threading.Thread` rather than just passing a callable to the `threading.Thread` constructor, but it looks like you've managed to avoid them all. No complaints on my end as long you've confirmed that it's all working as intended.

gnossen (approved):
Awesome work! This is going to make troubleshooting issues so much easier!

tools/run_tests/xds_k8s_test_driver/framework/xds_flags.py
sergiitk: I don't think we need this in most cases on local dev. It'll write to "/tmp` by default (this is an absl thing). It won't make sense for pre-provisioned pods (like url map tests) either.
Also, `--log_dir` is different per each environment (kokoro/staging/local), so I want setting `collect_app_logs` to be a conscious choice.
And there's always flagfiles to set this flags if needed.

tools/run_tests/xds_k8s_test_driver/framework/infrastructure/k8s_internal/k8s_log_collector.py
sergiitk: Yea, I had an impression this was the case, but couldn't find any strong advices against it. Thanks for noticing! I'll give it another looks, just in case. Won't be a difficult refactoring to not extend a thread, but manage it as a field.

sergiitk:
Latest test run with the collection not enabled (regression check):
- [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/ab7dcb3c-4ff7-4260-90e6-8edd6830dc37)
- [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/b80645c3-76de-4042-a6aa-9c222a50a822)
- [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/763446d7-f86f-49b1-96ec-c9e0b53d2f51)

sergiitk:
Python jobs:
- [grpc/core/master/linux/grpc_xds_k8s_lb_python](https://source.cloud.google.com/results/invocations/87766b6d-4e45-4c1d-bccf-7e664b2303f8)
- [grpc/core/master/linux/grpc_xds_url_map_python](https://source.cloud.google.com/results/invocations/fd90b972-c205-4867-8abb-b2112dd181d2)
- [grpc/core/master/linux/psm-security-python](https://source.cloud.google.com/results/invocations/98d3e679-ec8a-40bd-9f36-88179747b0d6)

PR #30608, opened 2022-08-16
Fix forkable globals

drfloob:

### DIFF
--- a/src/core/lib/event_engine/forkable.cc
+++ b/src/core/lib/event_engine/forkable.cc
@@ -22,15 +22,17 @@
 
 #include "absl/container/flat_hash_set.h"
 
+#include "src/core/lib/gprpp/no_destruct.h"
 #include "src/core/lib/gprpp/sync.h"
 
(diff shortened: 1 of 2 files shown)


src/core/lib/event_engine/forkable.cc
```diff
@@ -22,55 +22,57 @@
...
 
 namespace grpc_event_engine {
 namespace experimental {
 
 namespace {
-grpc_core::Mutex g_mu;
-bool g_registered{false};
-absl::flat_hash_set<Forkable*> g_forkables;
+grpc_core::NoDestruct<grpc_core::Mutex> g_mu;
+grpc_core::NoDestruct<bool> g_registered ABSL_GUARDED_BY(g_mu)(false);
```
ctiller, line 33: I think this NoDestruct is not needed

drfloob: Done

ctiller (approved):

PR #30620, opened 2022-08-17
[fixit] Fix flakiness for test case ChannelzSamplerTest.SimpleTest by reducing the chance of port conflict

ralphchung:
This test case seems to only fail when it is run in parallel. After digging into it, I found this test case opens a subprocess as a gRPC server using the hard-coded address and port number which are probably the cause of the flakiness.

This PR fixes the issue by always assigning an unused port number to the `server_address`.

### DIFF
--- a/test/cpp/util/channelz_sampler_test.cc
+++ b/test/cpp/util/channelz_sampler_test.cc
@@ -26,6 +26,7 @@
 #include <string>
 #include <thread>
 
+#include "absl/strings/str_cat.h"
 #include "gtest/gtest.h"
 
 #include <grpc/grpc.h>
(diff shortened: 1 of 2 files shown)


test/cpp/util/channelz_sampler_test.cc
```diff
@@ -164,6 +165,12 @@ TEST(ChannelzSamplerTest, SimpleTest) {
   client_thread_2.join();
 }
 
+int GenerateUniuquePortNumber() {
```
ctiller, line 168: Prefer to use `grpc_pick_unused_port_or_die()`

ctiller (approved):
Please update the PR description before submission

PR #30621, opened 2022-08-17
XdsSecurityTest: Use a timeout of 5000ms for RPCs

yashykt:
Backing out the change from https://github.com/grpc/grpc/pull/30609 to avoid invalidating the tests that might have been depending on the previous timeout value

### DIFF
--- a/test/cpp/end2end/xds/xds_end2end_test_lib.cc
+++ b/test/cpp/end2end/xds/xds_end2end_test_lib.cc
@@ -1002,7 +1002,7 @@ size_t XdsEnd2endTest::WaitForAllBackends(
 absl::optional<AdsServiceImpl::ResponseState> XdsEnd2endTest::WaitForNack(
     const grpc_core::DebugLocation& debug_location,
     std::function<absl::optional<AdsServiceImpl::ResponseState>()> get_state,
-    StatusCode expected_status) {
+    const RpcOptions& rpc_options, StatusCode expected_status) {
   absl::optional<AdsServiceImpl::ResponseState> response_state;
   auto deadline = absl::Now() + absl::Seconds(30);
   auto continue_predicate = [&]() {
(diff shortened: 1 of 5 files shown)


test/cpp/end2end/xds/xds_end2end_test_lib.cc
```diff
@@ -1014,7 +1014,9 @@ absl::optional<AdsServiceImpl::ResponseState> XdsEnd2endTest::WaitForNack(
            response_state->state != AdsServiceImpl::ResponseState::NACKED;
   };
   do {
-    const Status status = SendRpc();
+    // Some tests for example the XdsSecurityTests need a higher timeout from
+    // the default 1000ms for some builds.
+    const Status status = SendRpc(RpcOptions().set_timeout_ms(5000));
```
markdroth, line 1019: Please add an optional parameter to this method to pass in the `RpcOptions` and then override it only in the affected tests, rather than hard-coding this default for all tests.

markdroth (commented):
Thanks, I like this approach much better!

test/cpp/end2end/xds/xds_end2end_test_lib.cc
yashykt: I don't mind doing that but, I couldn't think of a reason where a `WaitForNack()` test would expect a Deadline Exceeded

yashykt: but from a futuristic perspective, if some test ends up requiring a RpcOptions() then yes, it does make sense to add that argument now itself.

yashykt: Done!

test/cpp/end2end/xds/xds_end2end_test_lib.cc
```diff
@@ -1014,7 +1014,9 @@ absl::optional<AdsServiceImpl::ResponseState> XdsEnd2endTest::WaitForNack(
            response_state->state != AdsServiceImpl::ResponseState::NACKED;
   };
   do {
-    const Status status = SendRpc();
+    // Some tests for example the XdsSecurityTests need a higher timeout from
```
markdroth, line 1017: I think this comment is no longer necessary.

markdroth (approved):
Looks great!

PR #30623, opened 2022-08-17
FaultInjection: Fix random number generation

yashykt:
`rand()` is not thread-safe, so instead use `absl::InsecureBitGen` and `absl::Uniform` to generate random numbers.


test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
```diff
@@ -204,7 +204,13 @@ TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageAbortViaHeaders) {
 
 TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageDelay) {
   CreateAndStartBackends(1);
+  const uint32_t kFixedDelayMilliseconds = 100000;
+#ifndef NDEBUG
```
markdroth, line 208: I'm really not keen on putting `#ifndef`s in the middle of individual tests.  Can we not just increase the timeout even for opt builds?

markdroth (commented):
This looks really good!  Just one comment to address.

test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
yashykt: Done

test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
```diff
@@ -204,11 +204,11 @@ TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageAbortViaHeaders) {
 
 TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageDelay) {
   CreateAndStartBackends(1);
-  const uint32_t kRpcTimeoutMilliseconds = grpc_test_slowdown_factor() * 3000;
+  const uint32_t kRpcTimeoutMilliseconds = 10000;
```
markdroth, line 207: Don't we still want to use `grpc_test_slowdown_factor()` here, with whatever timeout value we choose?

test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
```diff
@@ -250,11 +250,11 @@ TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageDelay) {
 TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageDelayViaHeaders) {
   CreateAndStartBackends(1);
   const uint32_t kFixedDelayMilliseconds = 100000;
-  const uint32_t kRpcTimeoutMilliseconds = grpc_test_slowdown_factor() * 3000;
+  const uint32_t kRpcTimeoutMilliseconds = 10000;
```
markdroth, line 253: Same here.

test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
yashykt: This is used in RpcOptions, and RpcOptions has it's own grpc_test_slowdown_factor() multiplier -
https://github.com/grpc/grpc/blob/2a7286a67c71e2e0c0eee5e49ba1a719fcabb3a8/test/cpp/end2end/xds/xds_end2end_test_lib.cc#L419
https://github.com/grpc/grpc/blob/2a7286a67c71e2e0c0eee5e49ba1a719fcabb3a8/test/core/util/test_config.cc#L81

test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
yashykt: see above

test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
markdroth: Ah, good catch!

markdroth (approved):

yashykt:
Thanks for reviewing!

PR #30628, opened 2022-08-18
[fixit] Remove absl::Now() from sensitive EventEngine timer_test

drfloob:
absl::Now() is not guaranteed to be monotonic, and we are seeing some
potential time travel at different frequencies on different machines. I've
replaced the most sensitive pieces with std::chrono::steady_clock to see
if that fixes the flakes we are seeing.


ctiller (approved):

drfloob:
The 2 failures are unrelated.

PR #30629, opened 2022-08-18
Fix StatusCancelledWithoutStartingRecvTrailingMetadata flake in client_lb_end2end_test

ananda1066:
Flake in `ClientLbInterceptTrailingMetadataTest.StatusCancelledWithoutStartingRecvTrailingMetadata` due to race between stream cancellation and check for trailing metadata. Added conditional var to wait on stream cancellation.

Tested with 100k runs, no flakes.

### DIFF
--- a/test/cpp/end2end/client_lb_end2end_test.cc
+++ b/test/cpp/end2end/client_lb_end2end_test.cc
@@ -2251,14 +2251,14 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
     return std::move(load_report_);
   }
 
+  // Returns true if received callback within deadline.
   bool WaitForLbCallback() {
     grpc::internal::MutexLock lock(&mu_);
-    bool timed_out = false;
     while (!trailer_intercepted_) {
-      timed_out = cond_.WaitWithTimeout(&mu_, absl::Seconds(3));
+      if (cond_.WaitWithTimeout(&mu_, absl::Seconds(3))) return false;
     }
     trailer_intercepted_ = false;
-    return timed_out;
+    return true;
   }
 
  private:
@@ -2343,7 +2343,7 @@ TEST_F(ClientLbInterceptTrailingMetadataTest,
     ctx.TryCancel();
   }
   // Wait for stream to be cancelled.
-  ASSERT_FALSE(WaitForLbCallback());
+  ASSERT_TRUE(WaitForLbCallback());
   // Check status seen by LB policy.
   EXPECT_EQ(1, num_trailers_intercepted());
   absl::Status status_seen_by_lb = last_status();
--- a/test/cpp/end2end/client_lb_end2end_test.cc
+++ b/test/cpp/end2end/client_lb_end2end_test.cc
@@ -2251,12 +2251,14 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
     return std::move(load_report_);
   }
 
-  void WaitForLbCallback() {
+  bool WaitForLbCallback() {
     grpc::internal::MutexLock lock(&mu_);
+    bool timed_out = false;
     while (!trailer_intercepted_) {
-      cond_.WaitWithTimeout(&mu_, absl::Seconds(3));
+      timed_out = cond_.WaitWithTimeout(&mu_, absl::Seconds(3));
     }
     trailer_intercepted_ = false;
+    return timed_out;
   }
 
  private:
@@ -2341,7 +2343,7 @@ TEST_F(ClientLbInterceptTrailingMetadataTest,
     ctx.TryCancel();
   }
   // Wait for stream to be cancelled.
-  WaitForLbCallback();
+  ASSERT_FALSE(WaitForLbCallback());
   // Check status seen by LB policy.
   EXPECT_EQ(1, num_trailers_intercepted());
   absl::Status status_seen_by_lb = last_status();
--- a/test/cpp/end2end/client_lb_end2end_test.cc
+++ b/test/cpp/end2end/client_lb_end2end_test.cc
@@ -2231,9 +2231,9 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
 
   static void TearDownTestCase() { grpc_shutdown(); }
 
-  int trailers_intercepted() {
+  int num_trailers_intercepted() {
     grpc::internal::MutexLock lock(&mu_);
-    return trailers_intercepted_;
+    return num_trailers_intercepted_;
   }
 
   absl::Status last_status() {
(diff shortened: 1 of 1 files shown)


test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -2328,6 +2333,11 @@ TEST_F(ClientLbInterceptTrailingMetadataTest,
     auto stream = stub->BidiStream(&ctx);
     ctx.TryCancel();
   }
+  {
+    // Wait for stream to be cancelled.
+    grpc::internal::MutexLock lock(&mu_);
+    cond_.WaitWithTimeout(&mu_, absl::Seconds(3));
```
ctiller, line 2339: condition variables can wake spuriously so it's best to add a loop with a variable that becomes signalled:

```
while (!trailers_intercepted()) {
  cond_.WaitWithTimeout(&mu_, absl::Seconds(3));
}
```

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -2264,10 +2267,12 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
       self->load_report_ =
           BackendMetricDataToOrcaLoadReport(*backend_metric_data);
     }
+    self->Signal();
   }
 
+  void Signal() { cond_.Signal(); }
```
markdroth, line 2273: This method isn't necessary.  Instead, line 2270 above can just do `self->cond_.Signal()`.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -2251,6 +2251,9 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
     return std::move(load_report_);
   }
 
+  grpc::internal::Mutex mu_;
```
markdroth, line 2254: Please make these data members private, and add a method to wait for the next event:

```
void WaitForLbCallback() {
  grpc::internal::MutexLock lock(&mu_);
  cond_.WaitWithTimeout(&mu_, absl::Seconds(3));
}
```

markdroth (commented):
Thanks for doing this!

test/cpp/end2end/client_lb_end2end_test.cc
markdroth: Thinking about this further, a more general solution might be to have a queue of LB callback data and then have a method to get the latest result within some specified timeout.  For example:

```
public:
  absl::optional<LbTrailingMetadataCallbackData> GetNextTrailingMetadataCallbackData(absl::Duration timeout) {
    grpc::internal::MutexLock lock(&mu_);
    if (queue_.empty()) {
      grpc::internal::CondVar cv;
      cv_ = cv;
      if (!cv.WaitWithTimeout(&mu_, timeout)) return absl::nullopt;
    }
    auto result = std::move(queue_.front());
    queue_.pop_front();
    return result;
  }

 private:
  struct LbTrailingMetadataCallbackData { ... };

  static void ReportTrailerIntercepted(
      const grpc_core::TrailingMetadataArgsSeen& args_seen) {
    LbTrailingMetadataCallbackData entry = // ...convert from args_seen ...
    ClientLbInterceptTrailingMetadataTest* self = current_test_instance_;
    grpc::internal::MutexLock lock(&self->mu_);
    queue_.push_back(std::move(entry));
    if (cv_ != nullptr) {
      cv_->Signal();
      cv_ = nullptr;
    }
  }

  grpc::internal::Mutex mu_;
  grpc::internal::CondVar* cv_ ABSL_GUARDED_BY(&mu_) = nullptr;
  std::deque<LbTrailingMetadataCallbackData> queue_ ABSL_GUARDED_BY(&mu_);
```

This way, we're not reusing the same `CondVar` for multiple LB callbacks.

ananda1066: Hm so I have run the test many times and I'm pretty sure this test case is the only place where this race is happening - I'm wondering if we need to add this complexity? I updated with the previous review comment and the test is passing in 100k runs

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Done.

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Ah I saw this pattern in some other places and was wondering why it was needed - done

test/cpp/end2end/client_lb_end2end_test.cc
markdroth: I was thinking that using the same `CondVar` for multiple LB callbacks might cause problems.  If it's not causing problems for any of the other tests, then this seems fine as-is.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -2251,24 +2251,36 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
     return std::move(load_report_);
   }
 
+  void WaitForLbCallback() {
+    grpc::internal::MutexLock lock(&mu_);
+    while (!trailer_intercepted_) {
+      cond_.WaitWithTimeout(&mu_, absl::Seconds(3));
```
markdroth, line 2257: I think the `WaitForLbCallback()` method should return a bool, which should be false if `WaitWithTimeout()` returns false.  That way, the caller can say `ASSERT_TRUE(WaitForLbCallback())`, and the test will fail if no callback arrives on time.

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Yeah, this is the only test in the suite that has this issue, I haven't seen it cause any issues - will run with blase once more to make sure.

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Oh I didn't notice WaitWithTimeout() returned a bool - done, thanks!

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Ok ran with blase again - no failures for this test suite.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -2251,8 +2251,15 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
     return std::move(load_report_);
   }
 
-  grpc::internal::Mutex mu_;
-  grpc::internal::CondVar cond_;
+  bool WaitForLbCallback() {
+    grpc::internal::MutexLock lock(&mu_);
+    bool timed_out = false;
```
markdroth, line 2256: I don't think we need this bool.  If the wait times out, then we know that there wasn't a callback, so there's no need to set `trailer_intercepted_` to false, which means that we can return immediately.

```
while (!trailer_intercepted_) {
  if (!cond_.WaitWithTimeout(&mu_, absl::Seconds(3))) return false;
}
trailer_intercepted_ = false;
return true;
```

markdroth (approved):
Just one minor comment left, otherwise looks great!

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: done

PR #30631, opened 2022-08-18
[fixit] Increase sleep time for test `ProxyEnd2endTest.RpcDeadlineExpires` to reduce flakiness

ralphchung:
This PR increases the sleep time to increase the possibility of exceeding deadlines. Not sure if this is a good solution.

In one of the failure runs, the difference between the actual timestamp returning from RPC call and the specified deadline is measured 63.497 ms.

### DIFF
--- a/test/cpp/end2end/end2end_test.cc
+++ b/test/cpp/end2end/end2end_test.cc
@@ -1561,15 +1561,15 @@ TEST_P(ProxyEnd2endTest, RpcDeadlineExpires) {
   EchoResponse response;
   request.set_message("Hello");
   request.mutable_param()->set_skip_cancelled_check(true);
-  // Let server sleep for 400 ms first to guarantee expiry.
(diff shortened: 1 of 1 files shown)


test/cpp/end2end/end2end_test.cc
```diff
@@ -1561,15 +1561,15 @@ TEST_P(ProxyEnd2endTest, RpcDeadlineExpires) {
   EchoResponse response;
   request.set_message("Hello");
   request.mutable_param()->set_skip_cancelled_check(true);
-  // Let server sleep for 40 ms first to guarantee expiry.
-  // 40 ms might seem a bit extreme but the timer manager would have been just
+  // Let server sleep for 400 ms first to guarantee expiry.
```
ctiller, line 1564: Let's make it 4 seconds

ctiller (approved):

PR #30636, opened 2022-08-19
xDS interop: Use ref server in py tests when set_not_serving needed

sergiitk:
In python tests that require set_not_serving server RPC, override
the python server with the reference server (Java) because
the python server doesn't yet support set_not_serving RPC.

Ref https://github.com/grpc/grpc/issues/30635.

Issue #30635, opened 2022-08-19
Implement SetServing and SetNotServing in Python xDS Interop Server


gnossen:
This is currently causing failover tests to fail:

```
[  FAILED  ] FailoverTest.test_failover
======================================================================
ERROR: test_failover (__main__.FailoverTest)
FailoverTest.test_failover
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/work/grpc/tools/run_tests/xds_k8s_test_driver/tests/failover_test.py", line 105, in test_failover
    default_test_servers[0].set_not_serving()
  File "/tmp/work/grpc/tools/run_tests/xds_k8s_test_driver/framework/test_app/server_app.py", line 82, in set_not_serving
    self.update_health_service_client.set_not_serving()
  File "/tmp/work/grpc/tools/run_tests/xds_k8s_test_driver/framework/rpc/grpc_testing.py", line 129, in set_not_serving
    log_level=logging.INFO)
  File "/tmp/work/grpc/tools/run_tests/xds_k8s_test_driver/framework/rpc/grpc.py", line 54, in call_unary_with_deadline
    return rpc_callable(req, **call_kwargs)
  File "/tmp/work/grpc/tools/run_tests/xds_k8s_test_driver/venv/lib/python3.7/site-packages/grpc/_channel.py", line 946, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/tmp/work/grpc/tools/run_tests/xds_k8s_test_driver/venv/lib/python3.7/site-packages/grpc/_channel.py", line 849, in _end_unary_response_blocking
    raise _InactiveRpcError(state)
grpc._channel._InactiveRpcError: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.UNIMPLEMENTED
	details = "Method not found!"
	debug_error_string = "{"created":"@1660819003.995318194","description":"Error received from peer ipv4:10.40.1.215:8080","file":"src/core/lib/surface/call.cc","file_line":966,"grpc_message":"Method not found!","grpc_status":12}"
>
```

PR #27096, opened 2021-08-24
use fixed c++ server image for python tests

ericgribkoff:

PR #30637, opened 2022-08-19
xDS interop: Python LB tests build and use the python server

sergiitk:
Undoes https://github.com/grpc/grpc/pull/27096.
While we lost context why py tests were used pinned cpp server,
we think this is due to lack of support of the set_not_serving RPC
in the python server, see https://github.com/grpc/grpc/issues/30635.

This RPC is only used in two tests, and for them we added a
temporary override of the test server to the reference Java server,
see https://github.com/grpc/grpc/pull/30636.

All other LB tests should work with the python server just fine.

### DIFF
--- a/tools/internal_ci/linux/grpc_xds_k8s_lb_python.sh
+++ b/tools/internal_ci/linux/grpc_xds_k8s_lb_python.sh
@@ -120,7 +120,7 @@ run_test() {
     --kube_context="${KUBE_CONTEXT}" \
     --secondary_kube_context="${SECONDARY_KUBE_CONTEXT}" \
     --client_image="${CLIENT_IMAGE_NAME}:${GIT_COMMIT}" \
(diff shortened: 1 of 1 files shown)


sergiitk:
Test run executed using  the driver from the https://github.com/grpc/grpc/pull/30636:

- [grpc/core/master/linux/grpc_xds_k8s_lb_python](https://source.cloud.google.com/results/invocations/499c1284-223a-4b41-abea-24fc43f6f74b)

sergiitk:
Missed commit ref in the server args. Re-running the test:
- [grpc/core/master/linux/grpc_xds_k8s_lb_python](https://source.cloud.google.com/results/invocations/e3f7a9f6-1ee6-4c66-b223-c1ca36e2d371)

sergiitk:
Passes now.

gnossen (approved):

PR #30649, opened 2022-08-19
xds_routing_e2e_test: fix XdsRetryPolicyMaxBackOff flake

markdroth:


yashykt (approved):

test/cpp/end2end/xds/xds_routing_end2end_test.cc
```diff
@@ -2289,6 +2289,10 @@ TEST_P(LdsRdsTest, XdsRetryPolicyMaxBackOff) {
   max_interval->set_seconds(1 * grpc_test_slowdown_factor());
   max_interval->set_nanos(0);
   SetRouteConfiguration(balancer_.get(), new_route_config);
+  // Send an initial RPC to make sure we get connected (we don't want
```
yashykt, line 2292: is a similar fix needed at other places?

markdroth: If so, we'll find that out when we investigate those other flakes.

PR #30664, opened 2022-08-19
[fixit] Solve the flakiness for test case `End2endTest.ClientCancelsBidi`

ralphchung:
The reason why this test case is flaky is that occasionally `OnRead()` is called before `OnCancel()` from the server's perspective. As a result, this line https://github.com/grpc/grpc/blob/master/test/cpp/end2end/test_service_impl.cc#L592 fails. This PR tries to deal with it by adding some synchronization.

From the semantics in the test case, I think the function `TryCancel()` in https://github.com/grpc/grpc/blob/master/test/cpp/end2end/end2end_test.cc#L1313 does not guarantee an RPC call will be cancelled. Expecting the RPC will always be cancelled sounds a bit strange to me.


test/cpp/end2end/test_service_impl.cc
```diff
@@ -631,6 +636,8 @@ CallbackTestServiceImpl::BidiStream(CallbackServerContext* context) {
     bool setup_done_{false};
     std::thread finish_thread_;
     bool client_try_cancel_ = false;
+    std::promise<void> cancel_notifier_;
```
ctiller, line 639: prefer `absl::Notification` here, iirc `std::{promise,future}` are banned in various places

ctiller: We've been using `absl::Notification` in the server code for at about a year, so if that's the case there's more to clean up. I think that doc is being overly conservative here though.

markdroth: @veblush, can you confirm what the current state is of the absl synchronization library?  Is it actually okay to use gRPC at this point?

ctiller (approved):

ctiller:
Going ahead and merging: if absl::Notification is indeed not ready, then we have sufficient other code to clean up.

PR #30669, opened 2022-08-20
[fixit] Tcp: Fix error handling

yashykt:
I was slightly unsure of this so creating a draft PR to ask for comments.

Iiuc, we should be reading from the error queue till we get an EAGAIN. If we don't, we risk losing events for Edge-Triggered cases, though I'm not sure if it can actually happen.

### DIFF
--- a/src/core/lib/iomgr/tcp_posix.cc
+++ b/src/core/lib/iomgr/tcp_posix.cc
@@ -1313,7 +1313,6 @@ struct cmsghdr* process_timestamp(grpc_tcp* tcp, msghdr* msg,
  * messages from the queue.
  */
 static bool process_errors(grpc_tcp* tcp) {
-  bool processed_err = false;
   struct iovec iov;
   iov.iov_base = nullptr;
   iov.iov_len = 0;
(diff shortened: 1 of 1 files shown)


ctiller:
I think this looks right, but I'd appreciate a look from @arjunroy and @soheilhy

soheilhy:
Good question. This is needed because there are two conditions on which we can get `EPOLLERR`:
1) There is something on the error queue.
2) There is an actual socket error.

When we receive `EPOLLERR` we try to `recvmsg(MSG_ERRQUEUE)` until we see `EAGAIN`.  If we read any messages from the error queue, we are done. Otherwise, we need to rearm read and write handlers in order to check for (2).  In TCP, you must read and/or write to see the actual socket error.  Reading `getsockopt(SO_ERROR)` will (stupidly) clear the socket error, and you get into a deadlock.

ctiller:
Ok so based on the note from @soheilhy above my read is that the original function is correct and we need to instead be ensuring that we're always arming read and write paths post calling `process_errors`.

yashykt:
So, after discussing with @soheilhy, turns out that EPOLLERR does not get affected by EPOLLET, so even if by chance we don't consume all possible notifications right now, we would still get another EPOLLER, and we would get to them later.

PR #30690, opened 2022-08-22
[fixit] Deflake xds_outlier_detection_end2end_test

yousukseung:


test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
```diff
@@ -38,6 +38,13 @@ class OutlierDetectionTest : public XdsEnd2endTest {
     return absl::StrCat(ipv6_only_ ? "[::1]" : "127.0.0.1", ":",
                         backends_[index]->port(), "_0");
   }
+  void set_test_interval(absl::Duration interval) { test_interval_ = interval; }
```
markdroth, line 41: I don't understand why this machinery is needed, especially since this setter method is never actually being called anywhere.  Storing this value in the base class makes the individual tests less hermetic and doesn't seem to actually add any value.  I think it would be better to remove this and just directly use the appropriate values in the individual tests.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
```diff
@@ -102,27 +109,28 @@ TEST_P(OutlierDetectionTest, SuccessRateEjectionAndUnejection) {
...
-  EXPECT_EQ(100, backends_[1]->backend_service()->request_count());
-  // Let base ejection period pass and see that we are no longer ejecting, rpcs
-  // going to their expectedly hashed backends.
-  gpr_sleep_until(grpc_timeout_milliseconds_to_deadline(2000));
-  ResetBackendCounters();
-  CheckRpcSendOk(DEBUG_LOCATION, 100, rpc_options);
-  CheckRpcSendOk(DEBUG_LOCATION, 100, rpc_options1);
-  EXPECT_EQ(100, backends_[0]->backend_service()->request_count());
-  EXPECT_EQ(100, backends_[1]->backend_service()->request_count());
+
```
markdroth, line 119: Please don't add blank lines within individual tests.

Same thing throughout.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
```diff
@@ -194,7 +202,8 @@ TEST_P(OutlierDetectionTest, SuccessRateMaxPercent) {
                           .set_server_expected_error(StatusCode::CANCELLED));
   CheckRpcSendOk(DEBUG_LOCATION, 1, rpc_options2);
   CheckRpcSendOk(DEBUG_LOCATION, 1, rpc_options3);
-  gpr_sleep_until(grpc_timeout_milliseconds_to_deadline(100));
+  gpr_sleep_until(grpc_timeout_milliseconds_to_deadline(
```
markdroth, line 205: Since the problem here is that the timing isn't lining up properly, I think we should try to remove as many of these sleeps as possible.  We should be able to replace almost all of them with `WaitForBackend()` calls to detect when the ejection actually occurs.  I think the only cases where that won't work are the ones where no ejection is happening, so those cases will probably still need sleeps, but those should also be less prone to flakiness.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: Removed.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: Done.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: I'm not sure if we can wait for backends elsewhere easily. FYI, I did not see time flakes in 10K runs otherwise.

- These tests check for no ejection, hence sleep seems to be the only option.
SuccessRateStdevFactor
SuccessRateEnforcementPercentage
SuccessRateMinimumHosts
SuccessRateRequestVolume
FailurePercentageThreshold
FailurePercentageEnforcementPercentage                                                                                                                                       
FailurePercentageMinimumHosts
FailurePercentageRequestVolume
SuccessRateAndFailurePercentageBothDisabled
SuccessRateAndFailurePercentageEjectionPolicyDisabled                 

- These tests should wait for a subset of backends to be ejected, which is not easy to change. 
SuccessRateMaxPercent (one of {0, 1})
FailurePercentageMaxPercentage (one of {0, 1})
SuccessRateAndFailurePercentage (two of {0, 1, 2, 3})

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
markdroth: Please also remove the `get_test_interval()` method and the `test_interval_` data member.  All of this should be local within the individual tests, not in the test framework.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
```diff
@@ -194,7 +199,8 @@ TEST_P(OutlierDetectionTest, SuccessRateMaxPercent) {
...
+      10 * get_test_interval() / absl::Milliseconds(1)));
```
markdroth, line 203: Do we need to actually sleep for 10x the interval in all of these tests in order to make them pass reliably?  If it's possible to use a lower multiplier without introducing flakiness, that seems preferable, since it will allow the tests to finish faster.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: I guess I was not clear in saying not easy. Yes, above works in one hand, but these tests should also ensure if not more than expected # of backends switched, so I feel using a timeout can be justified for the same reason we use it in the first group of tests to ensure we have proper coverage for false positives.

I considered adding something similar to above followed by a timeout, but then found this does not help with extra complexity as long as the total max timeout value is the same.

WDYT?

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: I did not have time to test individual tests with different values but x10 was the first multiplier that passed after ~50k runs with 100% after trying x2 and x4. Let me try x4~x8 again and update.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
markdroth: After the above code, you can still send 100 RPCs to each of the 3 remaining backends and make sure they all get to the right places.  That should ensure that no other backends were incorrectly ejected.

yousukseung: I see that works since we know a round of update happened here, unlike the first group of tests. Will update.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: Done.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: Done.

test/cpp/end2end/xds/xds_outlier_detection_end2end_test.cc
yousukseung: Done.

markdroth (approved):
Looks great!

PR #30695, opened 2022-08-22
Fix PickFirstTest flake in client_lb_end2end_test

ananda1066:
Add delay to shutdown for `PickFirstTest.CheckStateBeforeStartWatch` to make sure server is properly shut down before being restarted. Otherwise, the test sometimes has broken pipe failures if the server is in some middle state.

Verified with blase - no failures after change.

### DIFF
--- a/test/cpp/end2end/client_lb_end2end_test.cc
+++ b/test/cpp/end2end/client_lb_end2end_test.cc
@@ -2281,13 +2281,8 @@ class ClientLbInterceptTrailingMetadataTest : public ClientLbEnd2endTest {
   static ClientLbInterceptTrailingMetadataTest* current_test_instance_;
   int num_trailers_intercepted_ = 0;
   bool trailer_intercepted_ = false;
-<<<<<<< HEAD
   grpc_core::Mutex mu_;
   grpc_core::CondVar cond_;
-=======
-  grpc::internal::Mutex mu_;
-  grpc::internal::CondVar cond_;
->>>>>>> 609ebbf43d65b191dff5f41f5ffbc9ca58656847
   absl::Status last_status_;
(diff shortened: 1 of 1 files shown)


test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -429,7 +429,7 @@ class ClientLbEnd2endTest : public ::testing::Test {
       cond_.Signal();
     }
 
-    void Shutdown() {
+    void Shutdown(int timeout_ms = 0) {
```
markdroth, line 432: It doesn't look like this parameter is actually being used, which means that this PR is actually a no-op.

Also, even if you did use this parameter to pass to `server_->Shutdown()` on line 435 instead of the hard-coded timeout (which I assume is what you intended to do), I don't think that would actually solve the problem.  Unfortunately, we have a long-standing known issue where there is not actually any API to do a true graceful shutdown on a C++ server.  Even with a non-zero timeout, the server still shuts down immediately from the perspective of the transport (i.e., the transport sends a GOAWAY and then immediately closes the connection), which means that the client may think the call was actually started on the server (and thus not eligible for a transparent retry), so when the connection is closed, the client fails the call with "Broken pipe" instead of being transparently retried.

Normally when we see this, we fix it by changing the test to call `StopListeningAndSendGoaways()` (an internal API we added to the test) instead of `Shutdown()`, so that it does do a graceful shutdown at the transport level.  In this case, though, we'll probably need something a little more complicated, because we need to restart the server immediately after.

I'll put together a separate PR for this.

markdroth (changes requested):

ananda1066:
https://github.com/grpc/grpc/pull/30698 properly fixes flake

PR #30701, opened 2022-08-22
Add DebugLocation to CqVerifier

drfloob:
Most tests call Verify multiple times. This will help us identify which
of those calls failed.


yashykt (approved):

ctiller:
1. don't we get that from the call stack?
2. consider the `SourceLocation location = SourceLocation()` stuff on Expect - we can do this for most platforms without needing to type it out repeatedly (this could be a change only in cq_verifier.{h,cc})

drfloob:
Closed in favor of #30702

PR #30702, opened 2022-08-22
[fixit] Add source location to CqVerifier output

drfloob:
Supersedes #30701

### DIFF
--- a/test/core/end2end/cq_verifier.cc
+++ b/test/core/end2end/cq_verifier.cc
@@ -187,26 +187,28 @@ std::string CqVerifier::ToString() const {
   return absl::StrJoin(expectations, "\n");
 }
 
-void CqVerifier::FailNoEventReceived() const {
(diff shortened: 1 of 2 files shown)


yashykt (approved):

ctiller (approved):
I'd prefer that we got better information out of callstacks at crash time, but very well :)

drfloob:
> I'd prefer that we got better information out of callstacks at crash time, but very well :)

Agreed. Sometimes abseil's failure signal handler doesn't produce much of value. For example, an MSAN bug I was looking at today ended with:

```
*** SIGABRT received at time=1661138474 on cpu 1 ***
PC: @     0x7ff94e642438  (unknown)  raise
    @          0x11dcf39        352  absl::lts_20220623::WriteFailureInfo()
    @          0x11db480        576  absl::lts_20220623::AbslFailureSignalHandler()
    @           0xad206b  (unknown)  SignalAction()
```

This might be useful as well.

PR #30717, opened 2022-08-23
xDS interop: fix an issue with secondary zone namespaces not cleaned

sergiitk:
All alternative server runners except the failover test reuse the primary server runners' namespace. Failover test is using the secondary cluster, and manages its own namespace there. `reuse_namespace` disables namespace cleanup, and in this case it was set to `True` incorrectly.


sergiitk:
- [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/92fb5428-ae41-457f-8d28-d598c79214bb)

gnossen (approved):

PR #30730, opened 2022-08-24
ConnectionAttemptInjector: fix tsan failures

markdroth:
This fixes a `ThreadSanitizer: data race on vptr (ctor/dtor vs virtual call)` failure caused by `ConnectionAttemptInjector` having a vtable.  The fix is to combine the functionality of both subclasses into a single class so that no inheritence is needed.


yashykt (approved):
Can you post a link to the tsan failure please

markdroth:
Sure, the tsan failure was:

 https://source.cloud.google.com/results/invocations/185848da-7f61-45d2-93fa-d64baeeb7673/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_ring_hash_end2end_test@poller%3Dpoll/log

PR #30732, opened 2022-08-24
Adjust rpc timeouts in xds tests to reduce Deadline exceeded errors in xds msan tests

Vignesh2208:
Attempt to fix following flakes by increasing rpc timeout:

- https://source.cloud.google.com/results/invocations/d3af675e-6074-4517-9c94-c95a1eb51ed0/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_routing_end2end_test@poller%3Dpoll/tests
- https://source.cloud.google.com/results/invocations/02483217-05b5-4182-852f-1f3aeedaf392/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_routing_end2end_test@poller%3Dpoll/tests
- https://source.cloud.google.com/results/invocations/71d7aafc-4db2-44a9-8ec4-618641952fcd/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_cluster_end2end_test@poller%3Dpoll/tests
- https://source.cloud.google.com/results/invocations/8a5ca189-c538-4854-8e1e-5d67ed5d70cb/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_ring_hash_end2end_test@poller%3Depoll1/tests
- https://source.cloud.google.com/results/invocations/7a626d1e-bcd1-46b1-b229-55dbec1480f0/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_ring_hash_end2end_test@poller%3Dpoll/tests
- https://source.cloud.google.com/results/invocations/749dc2b6-3505-4917-b8f4-32df925d1547/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_ring_hash_end2end_test@poller%3Dpoll/tests
- https://source.cloud.google.com/results/invocations/ab209970-6322-404c-b904-c03d15d6943a/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_routing_end2end_test@poller%3Depoll1/tests

PR #30733, opened 2022-08-24
[fixit] bump RPC timeout for flaky XdsRoutingWeightedClusterUpdateWeights test

drfloob:
This has failed once in the past year under MSAN.


drfloob:
Covered by #30732

PR #30734, opened 2022-08-24
[fixit] Change the default XDS test timeout for MSAN builds

drfloob:
Flakes in MSAN builds of XDS end2end tests have dominated the flaky 
dashboard. This change bumps the default RPC timeout to 2 seconds 
for MSAN builds, which should help. This has been done on a test-by-test 
basis otherwise, and we continue to see multiple Deadline Exceeded 
errors in XDS end2end tests every day.

### DIFF
--- a/test/cpp/end2end/xds/xds_end2end_test_lib.h
+++ b/test/cpp/end2end/xds/xds_end2end_test_lib.h
@@ -40,6 +40,7 @@
 #include "src/proto/grpc/testing/echo.grpc.pb.h"
 #include "src/proto/grpc/testing/xds/v3/http_connection_manager.grpc.pb.h"
 #include "src/proto/grpc/testing/xds/v3/http_filter_rbac.grpc.pb.h"
+#include "test/core/util/build.h"
 #include "test/core/util/port.h"
 #include "test/cpp/end2end/counted_service.h"
 #include "test/cpp/end2end/test_service_impl.h"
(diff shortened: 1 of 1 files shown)


test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -720,7 +721,7 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType> {
     // Dev note: If a timeout or Deadline Exceeded error is occurring in an XDS
     // end2end test, consider changing that test's timeout instead of this
     // global default.
-    int timeout_ms = 1000;
+    int timeout_ms = BuiltUnderMsan() ? 2000 : 1000;
```
ctiller, line 724: should this be `grpc_test_slowdown_factor()*1000`?

drfloob: We can, but it's more of a shotgun approach. That'd be setting expected slowdowns of:
* 4x for MSAN
* 5x for UBSAN (which is not a problem for XDS tests)
* 3x for ASAN (which is not a problem for XDS tests)
* 5x for TSAN (which is not a problem for XDS tests)

markdroth: The test framework uses `grpc_timeout_milliseconds_to_deadline()` on the value of `timeout_ms`:

https://github.com/grpc/grpc/blob/03b6b01043f778e76b1f1380f5619067a7c5a42a/test/cpp/end2end/xds/xds_end2end_test_lib.cc#L419

And `grpc_timeout_milliseconds_to_deadline()` already multiplies by `grpc_test_slowdown_factor()`:

https://github.com/grpc/grpc/blob/93fbacb5343ea8cf993f779cee0847aed107bdc7/test/core/util/test_config.cc#L90

If we multiplied by that factor here, we'd be applying that multiplier twice, which I don't think we want.

markdroth (commented):
I think I'd still prefer to set the timeouts in the individual tests.  It's a little more work to play whack-a-mole with it, but it's much safer and results in more hermetic tests.

drfloob:
I see, I'll close the PR. We can consider changing the test slowdown factor if MSAN timeouts are an issue generally, because if that's not the case, it's odd that XDS would be differently affected.

PR #30741, opened 2022-08-24
Fix PickFirstTest.PendingUpdateAndSelectedSubchannelFails flake in client_lb_end2end_test

ananda1066:
Add hold to test case to delay connection failure until test condition succeeds.

No failures in 100k runs.

### DIFF
--- a/test/cpp/end2end/client_lb_end2end_test.cc
+++ b/test/cpp/end2end/client_lb_end2end_test.cc
@@ -1312,7 +1312,7 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
   auto stub = BuildStub(channel);
   // Create a number of servers, but only start 1 of them.
   CreateServers(2);
-  StartServer(0);
+  StartServers(2);
   // Initially resolve to first server and make sure it connects.
   gpr_log(GPR_INFO, "Phase 1: Connect to first server.");
   response_generator.SetNextResolution({servers_[0]->port_});
@@ -1321,11 +1321,7 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
   ConnectionAttemptInjector injector;
   auto hold = injector.AddHold(servers_[1]->port_);
   // Send a resolution update with the remaining servers, none of which are
-  // running yet, so the update will stay pending.  Note that it's important
-  // to have multiple servers here, or else the test will be flaky; with only
-  // one server, the pending subchannel list has already gone into
-  // TRANSIENT_FAILURE due to hitting the end of the list by the time we
-  // check the state.
+  // running yet, so the update will stay pending.
   gpr_log(GPR_INFO,
           "Phase 2: Resolver update pointing to remaining "
           "(not started) servers.");
@@ -1337,7 +1333,6 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
   hold->Wait();
   // RPCs will continue to be sent to the first server.
   CheckRpcSendOk(DEBUG_LOCATION, stub);
-  hold->Resume();
   // Now stop the first server, so that the current subchannel list
   // fails.  This should cause us to immediately swap over to the
   // pending list, even though it's not yet connected.  The state should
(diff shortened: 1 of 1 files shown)


test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -1311,13 +1311,15 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
...
   StartServer(0);
   // Initially resolve to first server and make sure it connects.
   gpr_log(GPR_INFO, "Phase 1: Connect to first server.");
   response_generator.SetNextResolution({servers_[0]->port_});
   CheckRpcSendOk(DEBUG_LOCATION, stub, true /* wait_for_ready */);
   EXPECT_EQ(channel->GetState(false), GRPC_CHANNEL_READY);
+  ConnectionAttemptInjector injector;
+  auto hold = injector.AddHold(servers_[1]->port_);
   // Send a resolution update with the remaining servers, none of which are
   // running yet, so the update will stay pending.  Note that it's important
```
markdroth, line 1324: The last sentence of this comment can be removed now.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -1328,8 +1330,14 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
...
           "(not started) servers.");
   response_generator.SetNextResolution(GetServersPorts(1 /* start_index */));
+  // Add hold before connection attempt to ensure RPCs will be sent to first
+  // server. Otherwise, pending subchannel list might already have gone into
+  // TRANSIENT_FAILURE due to hitting the end of the server list by the time
+  // we check the state.
+  hold->Wait();
   // RPCs will continue to be sent to the first server.
   CheckRpcSendOk(DEBUG_LOCATION, stub);
+  hold->Resume();
```
markdroth, line 1340: I don't think we want to resume this until after we check the channel's state on line 1353.  Or to say this another way, I think this should replace the `StartServer(1)` call on line 1358.

Also, I think we can probably change the check on line 1353 to accept only CONNECTING, not TRANSIENT_FAILURE.  I suspect that this PR will also fix the flake described in the TODO on line 1349 (and that TODO comment can go away now).

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -1311,13 +1311,15 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
       BuildChannel("", response_generator);  // pick_first is the default.
   auto stub = BuildStub(channel);
   // Create a number of servers, but only start 1 of them.
-  CreateServers(10);
+  CreateServers(2);
```
markdroth, line 1314: I think you can replace these two lines with just `StartServers(2)`.  Now that we're using the connection attempt injector, we can start both servers from the get-go, and we can just release the hold when we're ready for the second server to get connected.

markdroth (changes requested):
Thanks for doing this!

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Oops, guess that change didn't make it into the PR. Done.

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Done

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: Done

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -1346,18 +1339,13 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
...
-  // and into the LB policies.
   EXPECT_THAT(channel->GetState(false),
-              ::testing::AnyOf(GRPC_CHANNEL_CONNECTING,
-                               GRPC_CHANNEL_TRANSIENT_FAILURE));
-  // Now start the second server.
-  gpr_log(GPR_INFO, "Phase 4: Starting second server.");
-  StartServer(1);
-  // The channel should go to READY state and RPCs should go to the
-  // second server.
+              ::testing::AnyOf(GRPC_CHANNEL_CONNECTING));
```
markdroth, line 1343: Don't need `AnyOf()` here anymore.  Instead, you can just say `EXPECT_EQ(channel->GetState(false), GRPC_CHANNEL_CONNECTING)`.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -1346,18 +1339,13 @@ TEST_F(PickFirstTest, PendingUpdateAndSelectedSubchannelFails) {
...
+  // Resume connection attempt to second server now that first server is down.
+  // The channel should go to READY state and RPCs should go to the second
+  // server.
+  hold->Resume();
+  gpr_log(GPR_INFO, "Phase 4: Connect to second server.");
```
markdroth, line 1348: Please move this above the call to `hold->Resume()` and have the log message say "Phase 4: Resuming connection attempt to second server."  I think that will make the log output make more sense when debugging this test in the future.

markdroth (approved):
Please verify with the latest changes that the test is still non-flaky.  If so, looks great!

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: done.

test/cpp/end2end/client_lb_end2end_test.cc
ananda1066: done

PR #30749, opened 2022-08-25
end2end tests: apply test slowdown factor in various places where it was missed

markdroth:


src/core/lib/gprpp/time.h
```diff
@@ -200,6 +200,16 @@ class Duration {
     }
     return *this;
   }
+  Duration& operator*=(int64_t multiplier) {
+    if (millis_ == std::numeric_limits<int64_t>::max()) {
+      *this = multiplier < 0 ? NegativeInfinity() : Infinity();
+    } else if (millis_ == std::numeric_limits<int64_t>::min()) {
+      *this = multiplier < 0 ? Infinity() : NegativeInfinity();
+    } else {
+      millis_ *= multiplier;
```
ctiller, line 209: Need to deal with overflow here...

either explicitly (I probably made a mistake):
```
} else if (multiplier > 0 && millis_ > std::numeric_limits<int64_t>::max() / multiplier) {
  *this = Infinity();
} else if (multiplier < 0 && millis_ > std::numeric_limits<int64_t>::max() / -multiplier) {
  *this = NegativeInfinity();
} else if (multiplier > 0 && millis_ < std::numeric_limits<int64_t>::min() / multiplier) {
  *this = NegativeInfinity();
} else if (multiplier < 0 && millis_ < std::numeric_limits<int64_t>::min() / -multiplier) {
  *this = Infinity();
}
```

or leveraging the code for double and replacing this method with:
```
Duration& operator*=(int64_t multiplier) { *this = *this * static_cast<double>(multiplier); }
```

markdroth: Good catch, fixed.

test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
```diff
@@ -225,8 +225,8 @@ TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageDelay) {
   auto* delay_percentage = http_fault.mutable_delay()->mutable_percentage();
   delay_percentage->set_numerator(kDelayPercentagePerHundred);
   delay_percentage->set_denominator(FractionalPercent::HUNDRED);
-  auto* fixed_delay = http_fault.mutable_delay()->mutable_fixed_delay();
-  fixed_delay->set_seconds(kFixedDelaySeconds);
+  SetProtoDuration(grpc_core::Duration::Seconds(kFixedDelaySeconds),
```
yashykt, line 228: kFixedDelaySeconds in this test and in other tests in this file is set to 100seconds. If we scale that, it will go up to 500seconds under UBSAN. I suggest changing 100seconds to 20seconds. That should still give us the DEADLINE_EXCEEDED errors that we want

markdroth: Good point.  Fixed.

In the process, I also found a few more places where we needed to take the slowdown factor into account.

yashykt (approved):

src/core/lib/gprpp/time.h
```diff
@@ -200,6 +200,7 @@ class Duration {
     }
     return *this;
   }
+  Duration& operator*=(double multiplier);
```
drfloob, line 203: consider exercising this in `gprpp/time_test.cc`

test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -747,6 +748,11 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType> {
       return *this;
     }
 
+    RpcOptions& set_timeout(grpc_core::Duration timeout) {
```
drfloob, line 751: Why not replace the `set_timeout_ms(int rpc_timeout_ms)` version with this? Are both needed?

drfloob (approved):
LGTM with small suggestions.

src/core/lib/gprpp/time.h
markdroth: Done.

test/cpp/end2end/xds/xds_end2end_test_lib.h
markdroth: Yeah, they're both useful.  The vast majority of the tests just hard-code a millisecond value as an integer, and it makes the test less verbose to use this form.

PR #30751, opened 2022-08-25
TimerManager: Periodically check for thread exits

drfloob:
I identified a subtle bug in some other work, but have not figured out how to poke it reliably outside that work (The `std::shared_ptr<EventEngine> GetDefaultEventEngine` change, which should not affect TimerManager behavior).

Some thread ordering is resulting in a hang on TimerManager shutdown, since timer threads are exiting and the cv.Wait on TimerManager destruction is not being woken up.

I believe this invalidates a claimed guarantee about timer checks under contention, so we might want to reword that comment. There may also be a different fix that ensures proper ordering and cv signaling under all circumstances, but this PR works around it without much extra overhead.


drfloob:
This solution only hid a problem of circular dependencies. Even in successful runs, the timeout-wait would loop indefinitely, but TSAN did not catch the leaked thread in that situation.

PR #30755, opened 2022-08-25
Add a provision to allow specification of separate set of channel args for the channel to grpclb server

Vignesh2208:
This allows overriding some specific set of channel args instead of sharing all channel args with the parent channel.

### DIFF
--- a/include/grpc/impl/codegen/grpc_types.h
+++ b/include/grpc/impl/codegen/grpc_types.h
@@ -368,6 +368,10 @@ typedef struct {
    balancer before using fallback backend addresses from the resolver.
    If 0, enter fallback mode immediately. Default value is 10000. */
 #define GRPC_ARG_GRPCLB_FALLBACK_TIMEOUT_MS "grpc.grpclb_fallback_timeout_ms"
+/* The set of channel args to be passed to the channel established to the grpclb
+ * server. If unset, the same channel args of the parent channel will be used
+ * for the channel to the grpclb server. */
+#define GRPC_ARG_GRPCLB_CHANNEL_ARGS "grpc.grpclb_channel_args"
 /* Timeout in milliseconds to wait for the child of a specific priority to
    complete its initial connection attempt before the priority LB policy fails
    over to the next priority. Default value is 10 seconds. */
--- a/include/grpc/impl/codegen/grpc_types.h
+++ b/include/grpc/impl/codegen/grpc_types.h
@@ -371,8 +371,7 @@ typedef struct {
 /* Channel args to be used for the control-plane channel created to the grpclb
  * load balancers. This is a pointer arg whose value is a grpc_channel_args
  * object. If unset, most channel args from the parent channel will be
- * propagated to the grpclb channel. It is expected to be a pointer of type
- * grpc_channel_args */
+ * propagated to the grpclb channel. */
 #define GRPC_ARG_GRPCLB_CHANNEL_ARGS "grpc.grpclb_channel_args"
 /* Timeout in milliseconds to wait for the child of a specific priority to
    complete its initial connection attempt before the priority LB policy fails
--- a/include/grpc/impl/codegen/grpc_types.h
+++ b/include/grpc/impl/codegen/grpc_types.h
@@ -368,9 +368,11 @@ typedef struct {
    balancer before using fallback backend addresses from the resolver.
    If 0, enter fallback mode immediately. Default value is 10000. */
 #define GRPC_ARG_GRPCLB_FALLBACK_TIMEOUT_MS "grpc.grpclb_fallback_timeout_ms"
-/* The set of channel args to be passed to the channel established to the grpclb
- * server. If unset, the same channel args of the parent channel will be used
- * for the channel to the grpclb server. */
+/* Channel args to be used for the control-plane channel created to the grpclb
+ * load balancers. This is a pointer arg whose value is a grpc_channel_args
+ * object. If unset, most channel args from the parent channel will be
+ * propagated to the grpclb channel. It is expected to be a pointer of type
+ * grpc_channel_args */
 #define GRPC_ARG_GRPCLB_CHANNEL_ARGS "grpc.grpclb_channel_args"
 /* Timeout in milliseconds to wait for the child of a specific priority to
    complete its initial connection attempt before the priority LB policy fails
--- a/include/grpc/impl/codegen/grpc_types.h
+++ b/include/grpc/impl/codegen/grpc_types.h
@@ -368,11 +368,12 @@ typedef struct {
    balancer before using fallback backend addresses from the resolver.
    If 0, enter fallback mode immediately. Default value is 10000. */
 #define GRPC_ARG_GRPCLB_FALLBACK_TIMEOUT_MS "grpc.grpclb_fallback_timeout_ms"
-/* Channel args to be used for the control-plane channel created to the grpclb
- * load balancers. This is a pointer arg whose value is a grpc_channel_args
- * object. If unset, most channel args from the parent channel will be
- * propagated to the grpclb channel. */
-#define GRPC_ARG_GRPCLB_CHANNEL_ARGS "grpc.grpclb_channel_args"
+/* Experimental Arg. Channel args to be used for the control-plane channel
+ * created to the grpclb load balancers. This is a pointer arg whose value is a
+ * grpc_channel_args object. If unset, most channel args from the parent channel
+ * will be propagated to the grpclb channel. */
+#define GRPC_ARG_EXPERIMENTAL_GRPCLB_CHANNEL_ARGS \
+  "grpc.experimental.grpclb_channel_args"
 /* Timeout in milliseconds to wait for the child of a specific priority to
    complete its initial connection attempt before the priority LB policy fails
    over to the next priority. Default value is 10 seconds. */
--- a/include/grpcpp/support/channel_arguments.h
+++ b/include/grpcpp/support/channel_arguments.h
@@ -70,6 +70,9 @@ class ChannelArguments {
   /// the resolver.
   void SetGrpclbFallbackTimeout(int fallback_timeout);
 
+  /// Set the channel args to be used for the channel to grpclb server.
+  void SetGrpclbChannelArgs(const ChannelArguments& grpclb_channel_args);
+
   /// Set a mutator for the underlying socket.
   void SetSocketMutator(grpc_socket_mutator* mutator);
 
--- a/include/grpcpp/support/channel_arguments.h
+++ b/include/grpcpp/support/channel_arguments.h
@@ -70,9 +70,6 @@ class ChannelArguments {
   /// the resolver.
   void SetGrpclbFallbackTimeout(int fallback_timeout);
 
-  /// Set the channel args to be used for the channel to grpclb server.
-  void SetGrpclbChannelArgs(const ChannelArguments& grpclb_channel_args);
-
   /// Set a mutator for the underlying socket.
   void SetSocketMutator(grpc_socket_mutator* mutator);
 
--- a/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
@@ -1360,7 +1360,8 @@ ChannelArgs BuildBalancerChannelArgs(
     const ChannelArgs& args) {
   ChannelArgs grpclb_channel_args;
   const grpc_channel_args* lb_channel_specific_args =
-      args.GetPointer<grpc_channel_args>(GRPC_ARG_GRPCLB_CHANNEL_ARGS);
+      args.GetPointer<grpc_channel_args>(
+          GRPC_ARG_EXPERIMENTAL_GRPCLB_CHANNEL_ARGS);
   if (lb_channel_specific_args != nullptr) {
     grpclb_channel_args = ChannelArgs::FromC(lb_channel_specific_args);
   } else {
--- a/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
@@ -1358,18 +1358,11 @@ ServerAddressList ExtractBalancerAddresses(const ChannelArgs& args) {
 ChannelArgs BuildBalancerChannelArgs(
     FakeResolverResponseGenerator* response_generator,
     const ChannelArgs& args) {
-  // Create channel args for channel credentials that does not contain bearer
-  // token credentials.
-  auto* channel_credentials = args.GetObject<grpc_channel_credentials>();
-  GPR_ASSERT(channel_credentials != nullptr);
-  RefCountedPtr<grpc_channel_credentials> creds_sans_call_creds =
-      channel_credentials->duplicate_without_call_credentials();
-  GPR_ASSERT(creds_sans_call_creds != nullptr);
-
   ChannelArgs grpclb_channel_args;
-  if (args.Contains(GRPC_ARG_GRPCLB_CHANNEL_ARGS)) {
-    grpclb_channel_args = ChannelArgs::FromC(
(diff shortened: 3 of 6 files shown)


include/grpc/impl/codegen/grpc_types.h
```diff
@@ -368,6 +368,10 @@ typedef struct {
    balancer before using fallback backend addresses from the resolver.
    If 0, enter fallback mode immediately. Default value is 10000. */
 #define GRPC_ARG_GRPCLB_FALLBACK_TIMEOUT_MS "grpc.grpclb_fallback_timeout_ms"
+/* The set of channel args to be passed to the channel established to the grpclb
```
markdroth, line 371: This comment is a little misleading in that it implies that either exactly this set of channel args or exactly the set from the parent channel will be used, when in reality the grpclb LB policy will make some changes to the channel args either way.  I suggest the following wording instead:

"""
Channel args to be used for the control-plane channel created to the grpclb load balancers.  This is a pointer arg whose value is a grpc_channel_args object.  If unset, most channel args from the parent channel will be propagated to the grpclb channel.
"""

include/grpcpp/support/channel_arguments.h
```diff
@@ -70,6 +70,9 @@ class ChannelArguments {
   /// the resolver.
   void SetGrpclbFallbackTimeout(int fallback_timeout);
 
+  /// Set the channel args to be used for the channel to grpclb server.
+  void SetGrpclbChannelArgs(const ChannelArguments& grpclb_channel_args);
```
markdroth, line 74: Let's not add a first-class API for this.  That would require a gRFC, and I don't think it's something we want anyway, since this is an edge case, not really a first-class knob that we're providing.

I think it's fine for users `SetPointer()` or `SetPointerWithVtable()` for this arg.

src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
```diff
@@ -1365,36 +1365,48 @@ ChannelArgs BuildBalancerChannelArgs(
...
-      // args of the parent channel.
-      .Remove(GRPC_SSL_TARGET_NAME_OVERRIDE_ARG)
-      // Don't want to pass down channelz node from parent; the balancer
-      // channel will get its own.
-      .Remove(GRPC_ARG_CHANNELZ_CHANNEL_NODE)
-      // Remove the channel args for channel credentials and replace it
-      // with a version that does not contain call credentials. The loadbalancer
-      // is not necessarily trusted to handle bearer token credentials.
-      .Remove(GRPC_ARG_CHANNEL_CREDENTIALS)
+
```
markdroth, line 1368: Please remove unnecessary blank lines within functions.

Same on line 1408.

src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
```diff
@@ -1365,36 +1365,48 @@ ChannelArgs BuildBalancerChannelArgs(
...
+  ChannelArgs grpclb_channel_args;
+  if (args.Contains(GRPC_ARG_GRPCLB_CHANNEL_ARGS)) {
```
markdroth, line 1370: Suggest doing the lookup only once:

```
const grpc_channel_args* args_from_channel_args =
    args.GetPointer<grpc_channel_args>(GRPC_ARG_GRPCLB_CHANNEL_ARGS);
if (args_from_channel_args != nullptr) {
  grpclb_channel_args = ChannelArgs::FromC(args_from_channel_args);
} else {
  ...
}
```

src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
```diff
@@ -1365,36 +1365,48 @@ ChannelArgs BuildBalancerChannelArgs(
...
+            // channel will get its own.
+            .Remove(GRPC_ARG_CHANNELZ_CHANNEL_NODE)
+            // Remove the channel args for channel credentials and replace it
+            // with a version that does not contain call credentials. The
+            // loadbalancer is not necessarily trusted to handle bearer token
+            // credentials.
+            .Remove(GRPC_ARG_CHANNEL_CREDENTIALS);
+  }
+
+  return grpclb_channel_args
```
markdroth, line 1409: Suggest moving the code from lines 1361-1367 down here, since this is where we're using the objects created there.

src/cpp/common/channel_arguments.cc
```diff
@@ -29,11 +29,35 @@
...
 #include "src/core/lib/iomgr/exec_ctx.h"
 #include "src/core/lib/iomgr/socket_mutator.h"
 
 namespace grpc {
 
+static const grpc_arg_pointer_vtable* GrpcChannelArgsVTable() {
+  static const grpc_arg_pointer_vtable tbl = {
+      // copy
+      [](void* p) -> void* {
+        return p == nullptr
```
markdroth, line 42: No need to check for nullptr.  `grpc_channel_args_copy()` accepts a null argument.

src/cpp/common/channel_arguments.cc
```diff
@@ -29,11 +29,35 @@
...
+                   ? nullptr
+                   : grpc_channel_args_copy(static_cast<grpc_channel_args*>(p));
+      },
+      // destroy
+      [](void* p) {
+        if (p != nullptr) {
```
markdroth, line 48: Same here, `grpc_channel_args_destroy()` already checks for null.

markdroth (commented):
Please also add a test to test/cpp/end2end/grpclb_end2end_test.cc that shows an example of actually using this.  Thanks!

include/grpc/impl/codegen/grpc_types.h
Vignesh2208: Fixed.

include/grpcpp/support/channel_arguments.h
Vignesh2208: I modified SetPointer to specially check for this channel arg. I think asking the user to use SetPointerWithVtable might introduce some bugs if they dont pass the correct Vtable

src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
Vignesh2208: Fixed.

src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
Vignesh2208: Fixed.

src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
Vignesh2208: Done.

src/cpp/common/channel_arguments.cc
Vignesh2208: Removed.

src/cpp/common/channel_arguments.cc
Vignesh2208: Removed.

Vignesh2208:
> Please also add a test to test/cpp/end2end/grpclb_end2end_test.cc that shows an example of actually using this. Thanks!

Thanks for the review. I also updated grpclb_end2end_test to send a special user-agent string to the loadbalancer using this channel arg and verify the received user-agent string at the loadbalancer. Could you let me know if this works. Thanks!

test/cpp/end2end/grpclb_end2end_test.cc
```diff
@@ -198,6 +210,12 @@ class BalancerServiceImpl : public BalancerService {
       if (serverlist_done_) goto done;
     }
     {
+      // The loadbalancer should see a dummy user agent because it was
+      // specifically configured at the client using
+      // GRPC_ARG_GRPCLB_CHANNEL_ARGS
+      auto it = context->client_metadata().find("user-agent");
+      EXPECT_TRUE(it != context->client_metadata().end());
```
markdroth, line 217: This one should be `ASSERT_TRUE()`, because if the expectation is false, the next line will cause a crash.

test/cpp/end2end/grpclb_end2end_test.cc
```diff
@@ -406,17 +424,29 @@ class GrpclbEnd2endTest : public ::testing::Test {
...
     if (!expected_targets.empty()) {
       args.SetString(GRPC_ARG_FAKE_SECURITY_EXPECTED_TARGETS, expected_targets);
+      grpclb_channel_args = grpclb_channel_args.Set(
+          GRPC_ARG_FAKE_SECURITY_EXPECTED_TARGETS, expected_targets);
     }
     if (subchannel_cache_delay_ms > 0) {
       args.SetInt(GRPC_ARG_GRPCLB_SUBCHANNEL_CACHE_INTERVAL_MS,
                   subchannel_cache_delay_ms * grpc_test_slowdown_factor());
     }
+    auto grpclb_c_channel_args = grpclb_channel_args.ToC().release();
```
markdroth, line 446: If you don't call `.release()` here, then you don't need to explicitly call `grpc_channel_args_destroy()` on it later.  And you can pass `grpclb_c_channel_args.get()` to `args.SetPointerWithVtable()`.

include/grpc/impl/codegen/grpc_types.h
```diff
@@ -368,9 +368,11 @@ typedef struct {
    balancer before using fallback backend addresses from the resolver.
    If 0, enter fallback mode immediately. Default value is 10000. */
 #define GRPC_ARG_GRPCLB_FALLBACK_TIMEOUT_MS "grpc.grpclb_fallback_timeout_ms"
-/* The set of channel args to be passed to the channel established to the grpclb
- * server. If unset, the same channel args of the parent channel will be used
- * for the channel to the grpclb server. */
+/* Channel args to be used for the control-plane channel created to the grpclb
+ * load balancers. This is a pointer arg whose value is a grpc_channel_args
+ * object. If unset, most channel args from the parent channel will be
+ * propagated to the grpclb channel. It is expected to be a pointer of type
```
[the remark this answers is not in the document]
Vignesh2208: Removed.

src/cpp/common/channel_arguments.cc
```diff
@@ -213,6 +202,12 @@ void ChannelArguments::SetInt(const std::string& key, int value) {
 }
 
 void ChannelArguments::SetPointer(const std::string& key, void* value) {
+  if (key == GRPC_ARG_GRPCLB_CHANNEL_ARGS) {
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

test/cpp/end2end/grpclb_end2end_test.cc
Vignesh2208: Replaced with GPR_ASSERT since ASSERT_TRUE could only be used in functions returning void.

test/cpp/end2end/grpclb_end2end_test.cc
Vignesh2208: Fixed.

test/cpp/end2end/grpclb_end2end_test.cc
```diff
@@ -103,22 +103,42 @@ using BalancerService = CountedService<LoadBalancer::Service>;
...
+// A test user agent string sent by the client only to the grpclb loadbalancer.
 // The backend should not see this user-agent string.
-constexpr char kGrpclbSpecificUserAgentString[] =
-    "grpc-grpclb-dummy-user-agent";
+constexpr char kGrpclbSpecificUserAgentString[] = "grpc-grpclb-test-user-agent";
+
+//  VTable for grpc_channel_args type used to specify channel args for channels
+//  to grpclb load balancers.
+const grpc_arg_pointer_vtable* GrpcChannelArgsVTable() {
+  static const grpc_arg_pointer_vtable tbl = {
```
markdroth, line 113: I think you can just move this static declaration to line 467.  I don't need we need it in a separate `GrpcChannelArgsVTable()` function.

test/cpp/end2end/grpclb_end2end_test.cc
```diff
@@ -103,22 +103,42 @@ using BalancerService = CountedService<LoadBalancer::Service>;
...
 
   Status Echo(ServerContext* context, const EchoRequest* request,
               EchoResponse* response) override {
-    // The backend should not see a dummy user agent configured at the client
+    // The backend should not see a test user agent configured at the client
     // using GRPC_ARG_GRPCLB_CHANNEL_ARGS.
     auto it = context->client_metadata().find("user-agent");
     if (it != context->client_metadata().end()) {
-      EXPECT_FALSE(it->second.starts_with(kGrpclbSpecificUserAgentString));
+      GPR_ASSERT(!it->second.starts_with(kGrpclbSpecificUserAgentString));
```
markdroth, line 141: We generally don't want to use `GPR_ASSERT()` in tests, because if the assertion fails, it will cause the entire process to exit rather than just failing the test and continuing on to the next test.

This line doesn't need to be an assertion in the first place, because the code after this will continue to work fine even if this expectation fails.  Please put this back to using `EXPECT_FALSE()`.

test/cpp/end2end/grpclb_end2end_test.cc
```diff
@@ -210,12 +230,12 @@ class BalancerServiceImpl : public BalancerService {
...
     }
     {
-      // The loadbalancer should see a dummy user agent because it was
+      // The loadbalancer should see a test user agent because it was
       // specifically configured at the client using
       // GRPC_ARG_GRPCLB_CHANNEL_ARGS
       auto it = context->client_metadata().find("user-agent");
-      EXPECT_TRUE(it != context->client_metadata().end());
-      EXPECT_TRUE(it->second.starts_with(kGrpclbSpecificUserAgentString));
+      GPR_ASSERT((it != context->client_metadata().end()));
```
markdroth, line 237: Here I suggest saying:

```
EXPECT_TRUE(it != context->client_metadata().end());
if (it != context->client_metadata().end()) {
  EXPECT_THAT(it->second, ::testing::StartsWith(kGrpclbSpecificUserAgentString));
}
```

test/cpp/common/channel_arguments_test.cc
```diff
@@ -253,6 +277,44 @@ TEST_F(ChannelArgumentsTest, SetUserAgentPrefix) {
   EXPECT_TRUE(found);
 }
 
+TEST_F(ChannelArgumentsTest, SetGrpclbChannelArgs) {
```
markdroth, line 280: I don't think we need this test here, since we're not actually doing any special handling for this particular arg in `ChannelArguments`.

I think all the changes in this file should be reverted.

include/grpc/impl/codegen/grpc_types.h
```diff
@@ -368,6 +368,11 @@ typedef struct {
    balancer before using fallback backend addresses from the resolver.
    If 0, enter fallback mode immediately. Default value is 10000. */
 #define GRPC_ARG_GRPCLB_FALLBACK_TIMEOUT_MS "grpc.grpclb_fallback_timeout_ms"
+/* Channel args to be used for the control-plane channel created to the grpclb
```
yashykt, line 371: Might want to call this experimental

Vignesh2208: Done.

test/cpp/common/channel_arguments_test.cc
Vignesh2208: Done.

test/cpp/end2end/grpclb_end2end_test.cc
Vignesh2208: Done.

test/cpp/end2end/grpclb_end2end_test.cc
Vignesh2208: Done.

test/cpp/end2end/grpclb_end2end_test.cc
Vignesh2208: Done.

markdroth (approved):
Looks great!

PR #30760, opened 2022-08-26
If `-infile` is not readable, bail out and show an error!

csobrinho:
Without this, `grpc_cli` will connect, send the metadata and then completely stall until it is killed leading to believe it is the server that is stalled instead of `grpc_cli` that won't send the message.

(release notes: yes)


veblush (approved):

yashykt (approved):

csobrinho:
Hi @veblush and @yashykt. The PR was approved but it seems like two tests failed. What should I do next? Can the presubmit be run again? Thanks

PR #30780, opened 2022-08-29
xDSFaultInjection test: Change number of RPCs

yashykt:
The current `ComputeIdealNumRpcs` will result in 625 RPCs being sent which is a weird number for a range of 0 to 100. Choosing a multiple of 100 makes more sense for a uniform distribution in my opinion. 

`XdsFaultInjectionAlwaysDelayPercentageAbortSwitchDenominator` on the other hand has a range of 0 to 1,000,000 and a good number would probably be a million to get close to the mean but that might be too much, so leaving it alone since that has not been flaking.

### DIFF
--- a/test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_fault_injection_end2end_test.cc
@@ -149,7 +149,9 @@ TEST_P(FaultInjectionTest, XdsFaultInjectionPercentageAbort) {
   const uint32_t kAbortPercentagePerHundred = 50;
   const double kAbortRate = kAbortPercentagePerHundred / 100.0;
   const double kErrorTolerance = 0.1;
-  const size_t kNumRpcs = ComputeIdealNumRpcs(kAbortRate, kErrorTolerance);
+  // Choose a number of RPCs that's a multiple of 100 for a higher chance of
+  // reaching the mean.
+  const size_t kNumRpcs = 500;
   // Create an EDS resource
   EdsResourceArgs args({{"locality0", CreateEndpointsForBackends()}});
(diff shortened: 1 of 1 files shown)


markdroth:
Two questions:
1. Has this test actually been failing since #30749 was merged?  I don't remember seeing it on any of the flakiness reports since then.  Or do we have evidence that it will still flake if run on RBE for long enough?
2. I don't understand the mathematical logic here.  Why does lowering the number of attempts (from 625 to 500) decrease the flakiness?  I would think that the more attempts we give it, the more likely we are to hit the right probability.  Do we have evidence that this change will actually decrease the flakiness?

yashykt:
> Has this test actually been failing since https://github.com/grpc/grpc/pull/30749 was merged? I don't remember seeing it on any of the flakiness reports since then. Or do we have evidence that it will still flake if run on RBE for long enough?

I saw a failure on `XdsFaultInjectionPercentageAbort` which should not be getting affected by deadlines. Unfortunately, I haven't been able to reproduce this on RBE. https://source.cloud.google.com/results/invocations/f6a52dc7-94f3-420b-830a-04efdc33e451/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_fault_injection_end2end_test@poller%3Depoll1/log

> I don't understand the mathematical logic here. Why does lowering the number of attempts (from 625 to 500) decrease the flakiness? I would think that the more attempts we give it, the more likely we are to hit the right probability. Do we have evidence that this change will actually decrease the flakiness?

My intuition is - For a uniform distribution, each element in the range has an equal chance of getting hit, so if we really want to get close to the mean, we need to choose a multiple of the denominator (100).

PR #30784, opened 2022-08-29
subchannel: include IP address in status messages

markdroth:
Another step toward #22883.

### DIFF
--- a/src/core/ext/filters/client_channel/subchannel.cc
+++ b/src/core/ext/filters/client_channel/subchannel.cc
@@ -847,10 +847,10 @@ void Subchannel::SetConnectivityStateLocked(grpc_connectivity_state state,
                                         ": ", status.message()));
     status.ForEachPayload(
         [this](absl::string_view key, const absl::Cord& value)
-            // Want to use ABSL_EXCLUSIVE_LOCKS_REQUIRED(&mu_) here,
-            // but that won't work, because we can't pass the lock
-            // annotation through absl::Status::ForEachPayload().
-            ABSL_NO_THREAD_SAFETY_ANALYSIS { status_.SetPayload(key, value); });
+        // Want to use ABSL_EXCLUSIVE_LOCKS_REQUIRED(&mu_) here,
+        // but that won't work, because we can't pass the lock
+        // annotation through absl::Status::ForEachPayload().
+        ABSL_NO_THREAD_SAFETY_ANALYSIS { status_.SetPayload(key, value); });
   }
   if (channelz_node_ != nullptr) {
     channelz_node_->UpdateConnectivityState(state);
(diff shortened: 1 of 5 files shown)


src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -838,7 +837,18 @@ const char* SubchannelConnectivityStateChangeString(
...
+    status_ = status;
+  } else {
+    // Augment status message to include IP address.
+    status_ = absl::Status(status.code(),
+                           absl::StrCat(grpc_sockaddr_to_uri(&key_.address())
+                                            .value_or("<unknown address type>"),
+                                        ": ", status.message()));
+    status.ForEachPayload(
+        [this](absl::string_view key, const absl::Cord& value)
+            ABSL_NO_THREAD_SAFETY_ANALYSIS { status_.SetPayload(key, value); });
```
yashykt, line 850: is ABSL_NO_THREAD_SAFETY_ANALYSIS needed here?
Can you add a comment for why it's needed

markdroth: Done.

yashykt: Does `ABSL_EXCLUSIVE_LOCKS_REQUIRED` not work either?

yashykt (approved):

src/core/ext/filters/client_channel/subchannel.cc
markdroth: No, it doesn't.  That's why I said in the comment that it won't work.

The problem is that if we mark the lambda as `ABSL_EXCLUSIVE_LOCKS_REQUIRED(&mu_)`, the compiler complains, because the lambda is being invoked by `absl::Status::ForEachPayload()`, which is not itself marked as `ABSL_EXCLUSIVE_LOCKS_REQUIRED(&Subchannel::mu_)` -- and it cannot be marked as such, because (a) the absl code doesn't know anything about the gRPC `Subchannel` class, and (b) `mu_` is a private member of `Subchannel`, so the absl code wouldn't have access to it anyway.

We've run into a similar pattern before with `Orphanable`, `InternallyRefCounted<>`, and `DualRefCounted<>`, where the `Orphan()` method gets invoked via a smart pointer being reset or destroyed.  The smart pointer methods are not specific to the class with the `Orphan()` method, so they can't be marked with the right lock annotation, which means we can't use `ABSL_EXCLUSIVE_LOCKS_REQUIRED()` on the `Orphan()` method.  So the only option is to use `ABSL_NO_THREAD_SAFETY_ANALYSIS`.  Here are a couple of examples:

https://github.com/grpc/grpc/blob/cbed2cef570ee6ffdc3c7f86056dbfcb8375a15b/src/core/ext/filters/client_channel/lb_policy/rls/rls.cc#L269
https://github.com/grpc/grpc/blob/cbed2cef570ee6ffdc3c7f86056dbfcb8375a15b/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L157

PR #30787, opened 2022-08-29
Stop dotnet tests in the OSS benchmark.

wanlin31:
https://github.com/grpc/grpc-dotnet/pull/1803 has updated the required Dotnet version to be 7.0, however, the images in the test-infra repository is set as 6.0, this mismatch causing the failure when building the Dotnet images for OSS benchmark Donet tests, causing CI to fail. This PR stops the Dotnet tests in the OSS benchmarks.

Dotnet version in Docker image is still set to 6.0 in these locations:

https://github.com/grpc/test-infra/blob/master/containers/init/build/dotnet/Dockerfile
https://github.com/grpc/test-infra/blob/master/containers/pre_built_workers/dotnet/Dockerfile
https://github.com/grpc/test-infra/blob/master/containers/runtime/dotnet/Dockerfile


paulosjca (approved):

jtattermusch (approved):
LGTM.

PR #30789, opened 2022-08-30
client_channel, client_auth: rewrite disallowed status codes from the control plane

markdroth:
This implements [gRFC A54](https://github.com/grpc/proposal/blob/master/A54-restrict-control-plane-status-codes.md).

### DIFF
--- a/src/core/lib/channel/status_util.cc
+++ b/src/core/lib/channel/status_util.cc
@@ -22,6 +22,8 @@
 
 #include <string.h>
 
+#include "absl/strings/str_cat.h"
+
 #include "src/core/lib/gpr/useful.h"
 
 struct status_string_entry {
@@ -124,9 +126,9 @@ absl::Status MaybeRewriteIllegalStatusCode(const absl::Status& status,
     case absl::StatusCode::kAborted:
     case absl::StatusCode::kOutOfRange:
     case absl::StatusCode::kDataLoss: {
-      return absl::InternalError(absl::StrCat(
-          "illegal status code from ", source, "; original status: ",
-          status.ToString()));
+      return absl::InternalError(
+          absl::StrCat("illegal status code from ", source,
+                       "; original status: ", status.ToString()));
     }
     default:
       return status;
--- a/src/core/lib/channel/status_util.cc
+++ b/src/core/lib/channel/status_util.cc
@@ -109,3 +109,28 @@ bool grpc_status_code_from_int(int status_int, grpc_status_code* status) {
   *status = static_cast<grpc_status_code>(status_int);
   return true;
 }
+
+namespace grpc_core {
+
+absl::Status MaybeRewriteIllegalStatusCode(const absl::Status& status,
+                                           absl::string_view source) {
+  switch (status.code()) {
+    // The set of disallowed codes, as per
+    // https://github.com/grpc/proposal/blob/master/A54-restrict-control-plane-status-codes.md.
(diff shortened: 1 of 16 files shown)


src/core/lib/security/transport/client_auth_filter.cc
```diff
@@ -157,7 +160,15 @@ ArenaPromise<absl::StatusOr<CallArgs>> ClientAuthFilter::GetCallCredsMetadata(
 
   auto client_initial_metadata = std::move(call_args.client_initial_metadata);
   return TrySeq(
-      creds->GetRequestMetadata(std::move(client_initial_metadata), &args_),
+      Seq(creds->GetRequestMetadata(std::move(client_initial_metadata), &args_),
```
yashykt, line 163: I see, so we need to check all filters where a per-channel logic is affecting an RPC's status. Have you checked all filters? 
Is there a place where we can document this for new filter writers?

yashykt:
I'll leave it to @ctiller to review the filter changes

yashykt (approved):

src/core/lib/security/transport/client_auth_filter.cc
markdroth: The goal here is not to check all filters.  The [gRFC](https://github.com/grpc/proposal/blob/master/A54-restrict-control-plane-status-codes.md#proposal) specifies 4 individual cases that need to be covered, and those are the ones covered in this PR.

src/core/lib/channel/status_util.cc
```diff
@@ -109,3 +111,28 @@ bool grpc_status_code_from_int(int status_int, grpc_status_code* status) {
...
+    // https://github.com/grpc/proposal/blob/master/A54-restrict-control-plane-status-codes.md.
+    case absl::StatusCode::kInvalidArgument:
+    case absl::StatusCode::kNotFound:
+    case absl::StatusCode::kAlreadyExists:
+    case absl::StatusCode::kFailedPrecondition:
+    case absl::StatusCode::kAborted:
+    case absl::StatusCode::kOutOfRange:
+    case absl::StatusCode::kDataLoss: {
+      return absl::InternalError(
+          absl::StrCat("illegal status code from ", source,
```
ctiller, line 130: nit: capitalize `I` in illegal

src/core/lib/channel/status_util.cc
```diff
@@ -109,3 +111,28 @@ bool grpc_status_code_from_int(int status_int, grpc_status_code* status) {
   *status = static_cast<grpc_status_code>(status_int);
   return true;
 }
+
+namespace grpc_core {
+
+absl::Status MaybeRewriteIllegalStatusCode(const absl::Status& status,
```
ctiller, line 117: If we changed this signature to:

```
absl::Status MaybeRewriteIllegalStatusCode(absl::Status status, absl::string_view source)
```
we could std::move the status into this function in the majority of (all?) cases, and then use the RVO to make this rewriting essentially free even if there's an error message attached to a valid status.

ctiller (approved):
It's probably worth addressing the performance comment I made

src/core/lib/channel/status_util.cc
markdroth: Done.

src/core/lib/channel/status_util.cc
markdroth: Done.

PR #30797, opened 2022-08-30
Build target reorganization Part 3

ralphchung:

```
┌──────┐┌───────────────┐          
│grpc++││grpc++_insecure│          
└┬─────┘└─────┬─────────┘          
┌▽──────────┐┌▽───────────────────┐
│grpc++_base││grpc++_base_unsecure│
└┬────┬─────┘└┬─┬─────────────────┘
┌▽───┐│       │┌▽────────────┐     
│grpc││       ││grpc_unsecure│     
└┬───┘│       │└┬────────────┘     
┌▽────▽───────▽─▽┐                 
│grpc_base       │                 
└────────────────┘      
```

* (Covered in #30518 ) `gpr_base` is merged into `gpr`. The `gpr` here itself is nothing but depending on `gpr_base`. I don't think we need to separate them.
* (Covered in this PR) `grpc++_internals` and `grpc++_internal_hdrs_only` are merged into `grpc++_base`.
* (Covered in #30518 ) `gpr_codegen` is merged to `gpr` since the files in `gpr_codegen` should be removed in the future and be moved to `gpr`.
* (Covered in #30570) `grpc_codegen` is merged to `grpc_base` with the similar reason as above.
* (Covered in this PR) `grpc++_codegen_base` and `grpc++_codegen_base_src` are merged to `grpc++_base` and `grpc++_base_unsecure` with the similar reason as above.
* (Covered in #30570) `grpc_secure` is merged to `grpc` since `grpc` is by default secure. We do not need to separate them.

This PR does not remove targets `grpc++_codegen_base` and `grpc++_codegen_base_src` for temporary compatibility. They should be removed in the future.


BUILD
```diff
@@ -623,7 +665,15 @@ grpc_cc_library(
         "src/cpp/common/tls_credentials_options.cc",
         "src/cpp/server/insecure_server_credentials.cc",
         "src/cpp/server/secure_server_credentials.cc",
-    ],
+    ] + select(
```
ctiller, line 668: This needs to be select_deps - our tooling for imports can't support selects here

ctiller:
The additional 60kb code size for this refactoring needs an explanation of why (it may/may not be ok - idk).

BUILD
ralphchung: I see.

ctiller (approved):

PR #30802, opened 2022-08-31
Replace `absl::exchange()` with `std::exchange()`

ralphchung:
Since we are at C++14 now, we can start using `std::exchange()`.


ctiller (approved):

ctiller:
Probably worth updating core_banned_functions.py too to avoid backsliding

ralphchung:
> Probably worth updating core_banned_functions.py too to avoid backsliding

Updated

PR #30809, opened 2022-08-31
client_channel: allow LB policy to communicate update errors to resolver

### LINKED ISSUE #30803: Pick First load balancer does not retry to re-resolve if the first address resolution has 0 addresses

C++
v1.46.4
v1.36.4

Linux

gcc 7

### What did you do?

gRPC was attempting to connect using the Pick First load balancer and received a list of 0 addresses from the cARES resolver.

When trying to reproduce the issue consistently we passed an empty address list along so that we would fall into the block handling the case when there are 0 subchannels in AttemptToConnectUsingLatestUpdateArgsLocked in both [v1.36.4](https://github.com/grpc/grpc/blob/v1.36.4/src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc#L192) and [v1.46.4](https://github.com/grpc/grpc/blob/v1.46.4/src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc#L191).

gRPC automatically retrying address resolution to continue attempting to connect even when it receives 0 addresses.

### What did you see instead?

_On gRPC v1.36.4_ the following logs before gRPC stopped attempting to connect:
```
2022-04-12_04:43:14.45451 I0412 04:43:14.454464756 5638 pick_first.cc:147] Pick First 0x3bf5400 created.
2022-04-12_04:43:14.45455 I0412 04:43:14.454521631 5638 pick_first.cc:266] Pick First 0x3bf5400 received update with 0 addresses
2022-04-12_04:43:14.45461 I0412 04:43:14.454584964 5638 subchannel_list.h:363] [pick_first 0x3bf5400] Creating subchannel list 0x3c5a6c0 for 0 subchannels
```

When reproducing the issue on v1.46.4 gRPC stopped attempting to connect as well.

Relevant [gRPC mailing list email chain](https://groups.google.com/g/grpc-io/c/P-b8NQEkTr0/m/Eu9uygMKAgAJ) with @markdroth

markdroth:
Fixes #30803.

This fixes a bug where, when the resolver returns an empty address list or an error for addresses on the first resolution attempt, we put the channel in TRANSIENT_FAILURE, but we never do any re-resolutions, so the channel never recovers from that state, even when valid DNS data becomes available.

@ejona86 and @dfawley and I had already agreed on the desired way to handle problems like this, but I hadn't implemented it yet, since I (mistakenly) believed that there was no case in which we'd actually encounter this problem in practice.  So this PR fills in that missing piece of the client_channel architecture, which is to provide a mechanism whereby the LB policy can return some feedback to the resolver to indicate whether the resolver result was accepted.  The resolver can then use this feedback to determine whether to go into backoff.

The specific changes here are:
- The resolver result can now include an optional `result_health_callback`.  If non-null, it will be invoked by the channel with the resulting status of the result from the LB policy.
- The LB policy's `UpdateLocked()` method now returns a status, which can be non-OK in the case that the LB policy rejects the update.  All of the leaf LB policies have been changed to return non-OK when they get an empty address list or an error for the addresses.  (Note: There are some edge cases where a policy may not know for sure that there is a problem with the config until after `UpdateLocked()` has returned, so there will need to be some follow-up changes to address those cases.)
- In `PollingResolver`, we return a `result_health_callback` with every result, and we use that to determine whether to schedule another attempt or reset the backoff state.

@daichij, can you please try this out and see if it fixes the problem for you?


apolcyn (commented):
Main general comment, can we simplify this change by making `ReportResult` return an absl::Status, and have resolvers read that?

Since `result_health_callback` is only ever called inline from `ReportResult`, I don't think we we would lose anything.

markdroth:
Even though the code in this PR always invokes the callback synchronously, we'll need it to be async in the future to support some of the use-cases where the LB policies don't know whether or not the update is good at the time that `UpdateLocked()` is called.  That's why the API is designed this way.

apolcyn:
>  Even though the code in this PR always invokes the callback synchronously, we'll need it to be async in the future to support some of the use-cases where the LB policies don't know whether or not the update is good at the time that UpdateLocked() is called. That's why the API is designed this way.

Makes sense. For these async cases, would it be simpler to handle them by adding another method to the `Resolver` API?

The complexity I'm thinking about is, by storing `result_health_callback` in the client channel, we'll be adding another external ref to `resolver_`, which seems to be a little abnormal for `OrphanablePtr` usage.

markdroth:
The callback is created and owned by the resolver, so it's an internally-created ref, which is exactly the intended usage pattern for `InternallyRefCounted<>` objects like the resolver.

apolcyn:
> The callback is created and owned by the resolver, so it's an internally-created ref, which is exactly the intended usage pattern for InternallyRefCounted<> objects like the resolver.

I'm a little confused. When the client channel invokes the callback asynchronously from `ReportResult`, it will be holding a second ref on the Result via the callback. So, for example, `resolver_.reset()` will no longer be enough to ensure eventual destruction, the client channel will also need to clear it's `result_health_callback`.

markdroth:
The channel isn't holding a ref on the resolver; it's holding a callback, and the callback is holding a ref.  The callback is opaque from the perspective of the channel; it's created by the resolver, and only the resolver knows that it's holding a ref.

To say this another way, I think there are two conceptually independent things going on here:

1. The resolver is expecting a callback and needs the callback to hold a ref to itself to ensure that it is not destroyed before the callback runs.  This is exactly the use-case that `InternallyRefCounted<>` was designed for.  (Internally, see go/grpc-c-core-ref-counted-types#slide=id.g2fcb01e5a8_0_91; this is case 1 as described there.)
2. The channel is holding a callback from the resolver.  You're right that the channel needs to be sure to eventually delete this callback, but that's true of basically any callback ever taken by any piece of code.  From the channel's perspective, the callback is opaque; it has no idea that the channel is holding a ref to the resolver, and it does not care.

So I really don't think there's anything abnormal here; I think both of these uses are exactly the normal way of handling things from the perspective of each object.  The fact that the object holding the callback (and thus transitively holding an internal ref to the resolver) happens to be the same object holding the external ref on the resolver is essentially just an implementation detail; I don't think it actually changes any of the principles here.

In addition, note that there are functional reasons why we want the callback to actually be associated with an individual resolver result.  There are some potential future use-cases where the resolver may be getting data from a control plane and needs to provide feedback to the control plane as to whether that individual result was accepted or not.

src/core/lib/resolver/resolver.h
```diff
@@ -67,6 +69,14 @@ class Resolver : public InternallyRefCounted<Resolver> {
     // TODO(roth): Before making this a public API, figure out a way to
     // avoid exposing channel args this way.
     ChannelArgs args;
+    // If non-null, this callback will be invoked when the LB policy has
```
yashykt, line 72: Could we have cases where there is a race between the scheduling of the health callback and `ReportResult`? I'm thinking of a possibility where a callback gets delayed due to scheduling through something like an ExecCtx.

markdroth: I don't see how that could cause a race.  If the resolver chooses to return a callback, it cannot return another result until the callback is actually invoked, including whatever scheduling delay there is.  From the perspective of the resolver, it can't tell the difference between scheduling delay and the channel choosing to not invoke the callback yet for other reasons.

src/core/lib/resolver/resolver.h
```diff
@@ -67,6 +69,14 @@ class Resolver : public InternallyRefCounted<Resolver> {
...
+    // processed the result.  The status value passed to the callback
+    // indicates whether the LB policy accepted the update.  For polling
+    // resolvers, if the reported status is non-OK, then the resolver
+    // should put itself into backoff to retry the resolution later.
+    // The resolver impl must not call ResultHandler::ReportResult()
+    // again until after this callback has been invoked.
+    std::function<void(absl::Status)> result_health_callback;
```
yashykt, line 79: The TODO on line 89 can go away now

src/core/lib/resolver/resolver.h
yashykt: Is there any additional synchronization requirement for the result_health_callback (things like locks/work serializer should be held or not held)?
Let's also document why `ReportResult` cannot return this status synchronously

src/core/lib/resolver/resolver.h
markdroth: Done.

src/core/lib/resolver/resolver.h
markdroth: Updated comments.

yashykt (approved):

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -126,44 +130,67 @@ void PollingResolver::OnRequestCompleteLocked(Result result) {
...
+    result.result_health_callback = [self =
+                                         std::move(self)](absl::Status status) {
+      self->GetResultStatus(std::move(status));
+    };
     result_handler_->ReportResult(std::move(result));
   }
   Unref(DEBUG_LOCATION, "OnRequestComplete");
 }
 
+void PollingResolver::GetResultStatus(absl::Status status) {
```
apolcyn, line 161: Do we have a guarantee that an LB policy's `UpdateLocked` method won't be called between the time that `ReportResult` is called and `GetResultStatus` is called?

(from looking at the code, I think this won't happen, but I'm not sure if we have a more formal guarantee)

apolcyn (approved):

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
markdroth: I'm not sure I understand this question.  The LB policy's `UpdateLocked()` method will *always* be called between when the resolver calls `ReportResult()` and when the result-health callback is invoked.  It has to be that way, because the result of the `UpdateLocked()` call is what determines the status to be passed to the result-health callback.

apolcyn: Sorry! I meant to ask what's our guarantee that an LB policy's `UpdateLocked` method won't cause `RequestReResolutionLocked` to be invoked synchronously, i.e. before the health check callback is invoked?

markdroth: Ah, okay -- good question!

It looks like this PR did actually add such a case in pick_first.  I could change pick_first to avoid that, but in the future, when we have more async cases for this, it probably won't be possible to do so.  So instead, I've changed `PollingResolver` to handle that case properly, basically by deferring the re-resolution until after the result-health callback has been invoked.

Thanks for catching this!

daichij:
@markdroth Back-porting this change appears to have solved our issue!

markdroth:
Super, I'm glad to hear it!

PR #30811, opened 2022-08-31
retry: fix TSAN failure in BatchData dtor

markdroth:
Once the BatchData dtor unrefs the owning call, the call's arena (in which the BatchData object is allocated) may be destroyed, so it is no longer safe to access data members of BatchData, such as cleaning up a RefCountedPtr<>.

This also explains a previous TSAN problem for which a workaround was added in #29344, so that workaround is no longer needed.

### DIFF
--- a/src/core/ext/filters/client_channel/retry_filter.cc
+++ b/src/core/ext/filters/client_channel/retry_filter.cc
@@ -1726,13 +1726,6 @@ void RetryFilter::CallData::CallAttempt::BatchData::RunClosuresForCompletedCall(
 
 void RetryFilter::CallData::CallAttempt::BatchData::RecvTrailingMetadataReady(
     void* arg, grpc_error_handle error) {
-  // Add a ref to the outer call stack. This works around a TSAN reported
-  // failure that we do not understand, but since we expect the lifetime of
-  // this code to be relatively short we think it's OK to work around for now.
-  // TSAN failure:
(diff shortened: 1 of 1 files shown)


src/core/ext/filters/client_channel/retry_filter.cc
```diff
@@ -1353,11 +1357,12 @@ RetryFilter::CallData::CallAttempt::BatchData::BatchData(
 RetryFilter::CallData::CallAttempt::BatchData::~BatchData() {
   if (GRPC_TRACE_FLAG_ENABLED(grpc_retry_trace)) {
     gpr_log(GPR_INFO, "chand=%p calld=%p attempt=%p: destroying batch %p",
-            call_attempt_->calld_->chand_, call_attempt_->calld_,
-            call_attempt_.get(), this);
+            call_attempt_->calld_->chand_, call_attempt_->calld_, call_attempt_,
+            this);
   }
+  CallAttempt* call_attempt = call_attempt_;
   GRPC_CALL_STACK_UNREF(call_attempt_->calld_->owning_call_, "Retry BatchData");
```
ctiller, line 1364: nit:
```
  CallAttempt* call_attempt = std::exchange(call_attempt_, nullptr);
  GRPC_CALL_STACK_UNREF(call_attempt->calld_->owning_call_, "Retry BatchData");
```
just to make sure there's no subsequent use in future versions.

ctiller (approved):

src/core/ext/filters/client_channel/retry_filter.cc
markdroth: Done.

PR #30818, opened 2022-09-01
Support Python 3.11

gnossen:
- Git made breaking change related to file ownership, so this had to be silenced.
- There were a couple of places where 3.6 was still referenced. Those have been removed.


gnossen:
CC @xuanwang-amos

tools/dockerfile/grpc_artifact_python_manylinux2014_x86/Dockerfile
```diff
@@ -42,4 +43,7 @@ RUN curl -sSL -o ccache.tar.gz https://github.com/ccache/ccache/releases/downloa
     && cmake -DCMAKE_BUILD_TYPE=Release -DZSTD_FROM_INTERNET=ON -DHIREDIS_FROM_INTERNET=ON .. \
     && make -j4 && make install \
     && cd ../.. \
-    && rm -rf ccache-4.5.1 ccache.tar.gz
\ No newline at end of file
+    && rm -rf ccache-4.5.1 ccache.tar.gz
+
+
+RUN git config --global --add safe.directory /var/local/jenkins/grpc
```
sergiitk, line 49: I recommend moving it above the ccache install, and group with other `pip install` commands to reduce the number of layers: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

```
RUN /opt/python/cp38-cp38/bin/pip install --upgrade cython \
    && /opt/python/cp39-cp39/bin/pip install --upgrade cython \
    && /opt/python/cp310-cp310/bin/pip install --upgrade cython \
    && /opt/python/cp311-cp311/bin/pip install --upgrade cython \
    && git config --global --add safe.directory /var/local/jenkins/grpc
```

tools/dockerfile/grpc_artifact_python_musllinux_1_1_x64/Dockerfile
```diff
@@ -35,3 +36,5 @@ RUN curl -sSL -o ccache.tar.gz https://github.com/ccache/ccache/releases/downloa
     && make -j4 && make install \
     && cd ../.. \
     && rm -rf ccache-4.5.1 ccache.tar.gz
+
+RUN git config --global --add safe.directory /var/local/jenkins/grpc
```
sergiitk, line 40: How do we know it's `/var/local/jenkins/grpc`? Is this $HOME? Should we then use the env var?

sergiitk (approved):
Non-blocking questions/suggestions.

tools/dockerfile/grpc_artifact_python_musllinux_1_1_x64/Dockerfile
gnossen: It's not home. As far as I know, there is no env var provided for this. It's simply hardcoded in the docker images. We could set an env var in `run_tests.py`, but I think that refactor is out of the scope of this PR.

tools/dockerfile/grpc_artifact_python_manylinux2014_x86/Dockerfile
gnossen: Ack.

PR #30819, opened 2022-09-01
XdsClient: don't ignore resources after an invalid Resource wrapper

markdroth:
Previously, if we encountered a resource wrapped in a `Resource` wrapper message, we would stop processing the response at that point -- i.e., we would ignore any subsequent resources in the response that might still have been valid.  This PR fixes the behavior so that we no longer ignore subsequent resources.  It also uses the resource name from the `Resource` wrapper in NACK messages in cases where we previously did not report the resource name in the NACK message.


yashykt (approved):
Can you make sure that other languages do the same please?

markdroth:
I have no reason to believe that any other language has this bug, but I mentioned it in the chat for them to check.

PR #30821, opened 2022-09-01
Performant thread-safe Work Queue

drfloob:


src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,238 @@
...
+
+  // The managed items in the queue
+  std::deque<Storage> elements_ ABSL_GUARDED_BY(mu_);
+  // The most recently enqueued element. This is reserved from work stealing
+  absl::optional<T> most_recent_element_
+      ABSL_GUARDED_BY(most_recent_element_lock_);
+  absl::base_internal::SpinLock ABSL_ACQUIRED_AFTER(mu_)
+      most_recent_element_lock_;
+  // TODO(hork): consider ABSL_CACHELINE_ALIGNED
+  std::atomic<grpc_core::Timestamp> most_recent_element_enqueue_timestamp_{
```
ctiller, line 224: this should definitely be `std::atomic<int64_t>` - there's no specialization for Timestamp, so this will end up adding a mutex

@soheilhy I'm assuming this is meant for some kind of fairness ordering, and on that assumption likely the millisecond granularity provided by `Timestamp` isn't right

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,238 @@
...
+    }
+    most_recent_element_enqueue_timestamp_.store(kInvalidTimestamp,
+                                                 std::memory_order_relaxed);
+    absl::optional<T> tmp = std::exchange(most_recent_element_, absl::nullopt);
+    most_recent_element_lock_.Unlock();
+    return tmp;
+  }
+
+  // The managed items in the queue
+  std::deque<Storage> elements_ ABSL_GUARDED_BY(mu_);
```
ctiller, line 217: Note that `std::deque` has a minimum bucket size of 4KB, meaning we're likely to be consuming that much extra per thread - is it warranted?

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,238 @@
...
+namespace grpc_event_engine {
+namespace experimental {
+
+// A fast work queue based lightly on an internal Google implementation.
+//
+// This uses atomics and lightweight spinlocks to access the most recent element
+// in the queue, making it fast for LIFO operations. Accessing the oldest (next)
+// element requires taking a mutex lock.
+template <typename T>
+class WorkQueue {
```
ctiller, line 44: Any chance we could make this non-templated?

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,238 @@
...
+  // Returns whether the queue is empty
+  bool Empty() const {
+    return (most_recent_element_enqueue_timestamp_.load(
+                std::memory_order_relaxed) == kInvalidTimestamp &&
+            oldest_enqueued_timestamp_.load(std::memory_order_relaxed) ==
+                kInvalidTimestamp);
+  }
+
+  // Returns the number of elements in the queue
+  // TODO(hork): this is an expensive method. consider caching the size in an
```
ctiller, line 84: Probably not worth caching: we shouldn't be calling much outside of tests or some very infrequent stats collection.

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,238 @@
...
+    most_recent_element_lock_.Unlock();
+    return tmp;
+  }
+
+  // The managed items in the queue
+  std::deque<Storage> elements_ ABSL_GUARDED_BY(mu_);
+  // The most recently enqueued element. This is reserved from work stealing
+  absl::optional<T> most_recent_element_
+      ABSL_GUARDED_BY(most_recent_element_lock_);
+  absl::base_internal::SpinLock ABSL_ACQUIRED_AFTER(mu_)
```
ctiller, line 221: I'm moderately worried about `SpinLock` safety outside of kernel schedulers we control - especially since there's a Lock() (and not TryLock()) in Add.

@soheilhy what's your intuition about how necessary this is, and could we perhaps just use a `Mutex` here?

src/core/lib/event_engine/workqueue.h
drfloob: It should be an `std::atomic<int64_t>`, thanks. I'll use millis for it.

This variable is mostly being used as an inexpensive check to see if the most-recent-element is populated, an atomic bool could do that job. The one time that its Timestamp value is actually used is when populating the oldest enqueued timestamp. And that feature is useful for gathering stats, and potentially for smarter auto-scaling heuristics.

src/core/lib/event_engine/workqueue.h
drfloob: We can. I made it templated in case we wanted to support AnyInvocable and Closure* types. We could specialize the queue to one or the other, or both with overloads/separate methods. Today's thread pool only uses AnyInvocables, so we could start there.

Is there any particular reason you'd want to avoid templates in this code, specifically?

src/core/lib/event_engine/workqueue.h
drfloob: ack, I replaced it with a comment about the behavior.

src/core/lib/event_engine/workqueue.h
drfloob: (I'm also quite interested in @soheilhy 's take, but to add ...)

This was a first pass, and I wanted to discuss it. I'm getting some benchmarks set up, and I planned to swap this with a Mutex to see the effect. I'd prefer not to rely on anything in absl's `base_internal` namespace as well, but I think we can consider tolerating it if the performance benefit is significant. We also have gpr_spinlock, which I could wrap in something that plays well with absl lockable types.

soheilhy: This used to be an atomic, but after adding timestamps we ended up adding a spinlock. It really depends on how this is used. You can start with `Mutex` and switch back to `SpinLock` later.

src/core/lib/event_engine/workqueue.h
soheilhy: Internally, I have implemented a linked ring-buffer.

I suspect we won't use this class internally and use the executors we have, right?

src/core/lib/event_engine/workqueue.h
drfloob: Using atomic<int64_t> made no significant performance difference in my tests (see the last 2 commits with benchmark results).

src/core/lib/event_engine/workqueue.h
drfloob: Some quick benchmark results on my machine using a Mutex instead of a Spinlock:

single-element tests: 66% decrease in performance (~3.6M/s to ~1.2M/s)
multiple-element tests: 22% decrease in performance (~1.1M/s to 860k/s)

```
------------------------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------------
BM_WorkQueueIntptrPopFront/1           807 ns          806 ns       865773 items_per_second=1.24034M/s
BM_WorkQueueIntptrPopFront/8          8991 ns         8989 ns        75894 items_per_second=889.984k/s
BM_WorkQueueIntptrPopFront/64        75695 ns        75674 ns         9436 items_per_second=845.735k/s
BM_WorkQueueIntptrPopFront/512      589618 ns       589446 ns         1162 items_per_second=868.612k/s
BM_WorkQueueIntptrPopFront/4096    4857182 ns      4855613 ns          145 items_per_second=843.56k/s
I0902 19:37:09.771871994      12 test_config.cc:194]         TestEnvironment ends
```

drfloob: correction: this was a debug build. There is effectively no difference in opt builds, ~12M/s for all sizes > 1. I'll write some multithreaded microbenchmarks.

src/core/lib/event_engine/workqueue.h
ctiller: Internally in the data center I expect we won't use this class.
We will for any non data center apps (mobile, other small devices).

drfloob:
Update: I've added some multi-threaded benchmarks that exercise sparsely-populated queues, where there are N threads, N queues, various queue fill rates, and various steal attempt counts from all threads accessing all queues.

Summary of 10 threads polling 10 work queues:
* at 2% fill (expected 98% steal miss), WorkQueue can do 20x to 50x compared to std::deque alone. There are some diminishing returns that scale with queue volume and steal attempts relative to thread count (not shown).
* at 10% fill, WorkQueue has a flat profile, with about 20x performance improvement over std::deque
* at 50% fill, WorkQueue performs about 5x better.

At 128 threads (on a 128 core machine):
* the std::deque benchmark consistently performs worse than the same tests with 10 threads. 
* WorkQueue continues to scale up, but not linearly. There may be a sweet spot to tune.

We have discussed some options to further improve this, but I'm happy with the performance right now.

src/core/lib/event_engine/workqueue.h
drfloob: Let's keep the int64_t atomic as you suggested, it may be contributing to the opt-build performance wins we're seeing now.

src/core/lib/event_engine/workqueue.h
drfloob: Performance update below. Spinlock has been removed.

src/core/lib/event_engine/workqueue.h
drfloob: Most of the 4k would be unused in our current use cases, so I can see about a ring buffer implementation. From what I've seen so far though, I'd expect under 10 threads in the pool for the current workloads, so < 50Kb added memory with the global queue included. That's well within the "medium bloat" labeling.

src/core/lib/event_engine/workqueue.h
drfloob: I specialized the queue for Closure & AnyInvocable. There's some asymmetry in dealing with the two types given that Closure* callbacks are not owned by the queue, but AnyInvocables are, so the common usage is that closures _must_ be executed or a leak will happen (or if the caller knows it's an Closure-wrapped AnyInvocable, then they can delete it instead).

src/core/lib/event_engine/workqueue.cc
```diff
@@ -0,0 +1,208 @@
...
+  if (oldest_enqueued_timestamp_.load(std::memory_order_relaxed) !=
+      kInvalidTimestamp) {
+    absl::optional<EventEngine::Closure*> t = TryLockAndPop(/*front=*/false);
+    if (t.has_value()) return *t;
+  }
+  return absl::nullopt;
+}
+
+void WorkQueue::Add(EventEngine::Closure* closure) {
+  grpc_core::ExecCtx exec_ctx;
```
ctiller, line 117: is the exec_ctx to get current time?

drfloob: It is, that's its only purpose.

test/core/event_engine/work_queue_test.cc
```diff
@@ -0,0 +1,176 @@
...
+  ASSERT_GT(youngest_ts, oldest_ts);
+}
+
+TEST(WorkQueueTest, ThreadedStress) {
+  WorkQueue queue;
+  constexpr int thd_count = 33;
+  constexpr int element_count_per_thd = 3333;
+  std::vector<std::thread> threads;
+  threads.reserve(thd_count);
+  AnyInvocableClosure closure([] {});
```
ctiller, line 150: lets have this closure increment a counter or some such thing, so we can assert that all thd_count*element_count_per_thd closures are scheduled

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,126 @@
...
+//
+// This uses atomics to access the most recent element in the queue, making it
+// fast for LIFO operations. Accessing the oldest (next) element requires taking
+// a mutex lock.
+class WorkQueue {
+ public:
+  // comparable to Timestamp::milliseconds_after_process_epoch()
+  static const int64_t kInvalidTimestamp = -1;
+
+  class Storage {
```
ctiller, line 47: this seems like it could be private (along with AddInternal) with each of the Add methods being wrappers.

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,126 @@
...
+    EventEngine::Closure* closure_ = nullptr;
+    int64_t enqueued_ = kInvalidTimestamp;
+  };
+
+  WorkQueue() = default;
+  // Returns whether the queue is empty
+  bool Empty() const;
+  // Returns the number of elements in the queue
+  // This method locks the queue, be mindful of performance when using it.
+  size_t Size();
```
ctiller, line 81: What's the use case?

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,126 @@
...
+  // Returns the Timestamp of when the most recently-added element was
+  // enqueued.
+  grpc_core::Timestamp OldestEnqueuedTimestamp() const;
```
ctiller, line 84: What's the use case for making this public?

src/core/lib/event_engine/workqueue.h
```diff
@@ -0,0 +1,126 @@
...
+  // Returns the next (oldest) element from the queue, or nullopt if empty
+  absl::optional<EventEngine::Closure*> PopFront() ABSL_LOCKS_EXCLUDED(mu_);
```
ctiller, line 86: Would recommend dropping optional here... pointers are already nullable, so we can just use nullptr for empty.

test/core/event_engine/work_queue_test.cc
```diff
@@ -0,0 +1,176 @@
+// Copyright 2022 The gRPC Authors
```
ctiller, line 1: please either rename this file to workqueue_test.cc or rename the header/impl to work_queue.{h,cc}

test/core/event_engine/work_queue_test.cc
```diff
@@ -0,0 +1,176 @@
...
+  constexpr int thd_count = 33;
+  constexpr int element_count_per_thd = 3333;
+  std::vector<std::thread> threads;
+  threads.reserve(thd_count);
+  AnyInvocableClosure closure([] {});
+  for (int i = 0; i < thd_count; i++) {
+    threads.emplace_back([&] {
+      int cnt = 0;
+      do {
+        queue.Add(&closure);
```
ctiller, line 155: Let's instead make a:
```
class TestClosure : public Closure {
 public:
  void Run() { delete this; }
};
```

and here do `queue.Add(new TestClosure());`

and when we're popping, go ahead and just run the closure we pop:

```
while (cnt < element_count_per_thd) {
  if (auto* c = queue.PopFront()) c->Run();
}
```

in this way we leverage ASAN to tell us whether we've got any bugs where closures can get aliased internally and returned twice or such (or not returned)

ctiller:
This looks really fantastically close, very happy with how it's shaping up.

ctiller (changes requested):

ctiller:
One further question... am I right to assume that (in the absense of threading) this type is expected to behave exactly as `std::deque<Closure*>` ... ie that:
```
WorkQueue::Add(Closure*) maps to deque::push_back
WorkQueue::PopFront maps to deque::pop_front
WorkQueue::PopBack maps to deque::pop_back
WorkQueue::Empty maps to deque::empty
WorkQueue::Size maps to deque::size
```

if this is the case, I'd highly recommend writing a quick fuzzer in the style of https://github.com/grpc/grpc/blob/master/test/core/gprpp/chunked_vector_fuzzer.cc or https://github.com/grpc/grpc/blob/master/test/core/avl/avl_fuzzer.cc - such a fuzzer could apply random push/pop combinations and ensure that the mapping listed above matches the observable behavior of each type (and ensure that those assumptions continue to hold in the future)

src/core/lib/event_engine/workqueue.cc
ctiller: as discussed, let's get https://github.com/grpc/grpc/pull/30815 in first and switch this to use that.

test/core/event_engine/work_queue_test.cc
drfloob: Done.

src/core/lib/event_engine/workqueue.h
drfloob: Potential future work, helping queue-owners tune behavior based on queue lag, and provides an inexpensive hook for stats collection / observability (more specific to the usage than a general "max queue age" stat that could be done internally).

src/core/lib/event_engine/workqueue.h
drfloob: Potential future work, helping queue-owners tune behavior based on queue depth. Given the expense, it may be best to remove it. An efficient implementation may be useful for prioritizing certain queues during work stealing, or deciding when to spin up more worker threads. Should we remove it as YAGNI?

src/core/lib/event_engine/workqueue.h
drfloob: Done

test/core/event_engine/work_queue_test.cc
drfloob: Good idea, done. ASAN passed fine.

src/core/lib/event_engine/workqueue.h
drfloob: Good catch, fixed.

drfloob:
Thanks for the tips! I've added a fuzzer, and ran it for 10 minutes @ 42 jobs @ 42 workers. Please let me know if you have any suggestion for how we can improve the fuzzer.

test/core/event_engine/work_queue/corpora/a02dfa6454c3047ba1eb04274980c636d1e56875
```diff
@@ -0,0 +1,96 @@
+actions {
```
ctiller, line 1: I'd probably drop these for now, and just check in one 'empty' file - the fuzzing backends will keep a big corpus for us, so this mostly just increases the number of bytes we need to check out to get to building.

We used to check in big corpora to github and I don't think I've seen any benefit.

test/core/event_engine/work_queue/work_queue_fuzzer.cc
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+          work_queue_.Add(CreateInvocable(action.add().key()));
+          deque_.push_back(CreateClosureWrappedInvocable(action.add().key()));
+        }
+      } break;
+      case work_queue_fuzzer::Action::kPopFront: {
+        // pop and discard front closure, executing to ensure memory freeing
+        auto* wq_c = work_queue_.PopFront();
+        if (wq_c == nullptr) {
+          assert(deque_.empty());
```
ctiller, line 52: use GPR_ASSERT or something that doesn't disappear in opt builds (since we'll also fuzz in opt builds)

ctiller: or do the abort like on 73

test/core/event_engine/work_queue/work_queue_fuzzer.cc
```diff
@@ -0,0 +1,131 @@
...
+      auto* dq_c = deque_.back();
+      deque_.pop_back();
+      wq_c->Run();
+      dq_c->Run();
+    }
+  }
+
+  WorkQueue work_queue_;
+  std::deque<EventEngine::Closure*> deque_;
+  absl::optional<int> last_executed_key_;
```
ctiller, line 121: Took me a minute to figure out how this works, probably deserves a comment.

ctiller (approved):
a few nits, but looks good

src/core/lib/event_engine/workqueue.h
ctiller: up to you

test/core/event_engine/work_queue/work_queue_fuzzer.cc
drfloob: Agreed, comment added.

test/core/event_engine/work_queue/work_queue_fuzzer.cc
drfloob: I replaced both instances with a variation of the abort.

test/core/event_engine/work_queue/corpora/a02dfa6454c3047ba1eb04274980c636d1e56875
drfloob: Nuked.

src/core/lib/event_engine/workqueue.h
drfloob: Also nuked, it's not terribly complicated to add back.

drfloob:
Thanks Craig. I'll wait on https://github.com/grpc/grpc/pull/30815 and do make that change before merging.

PR #30823, opened 2022-09-02
XdsClient: add unit test and fix watcher notification bugs

markdroth:
This adds a unit test for XdsClient and fixes several watcher-notification bugs found in the process.  Specifically:
- When an ADS stream fails or an xDS channel reports a connectivity failure, report an error only to the watchers for resources being subscribed to on that particular channel, not to watchers on other channels.
- Cache the error status for the channel, so that if a new watcher is started after the channel reports the error, we can immediately report that error to the new watcher.
- If a resource is NACKed and has not been previously cached, or does not exist, report that fact to any new watcher that may be started later.
- If a resource in an ADS response is unparseable but is wrapped in a `Resource` wrapper, we do know its name, so record the validation failure in the cache and report it to the watchers.


yashykt:
> Cache the error status for the channel, so that if a new watcher is started after the channel reports the error, we can immediately report that error to the new watcher.
If a resource is NACKed and has not been previously cached, or does not exist, report that fact to any new watcher that may be started later.

What's the current (pre-fix) behavior?

markdroth:
> > Cache the error status for the channel, so that if a new watcher is started after the channel reports the error, we can immediately report that error to the new watcher.
> > If a resource is NACKed and has not been previously cached, or does not exist, report that fact to any new watcher that may be started later.
> 
> What's the current (pre-fix) behavior?

We were failing to report anything to these watchers.

src/core/ext/xds/xds_client.cc
```diff
@@ -522,27 +518,53 @@ void XdsClient::ChannelState::UnsubscribeLocked(const XdsResourceType* type,
...
-            status.ToString().c_str());
-    xds_client_->NotifyOnErrorLocked(absl::UnavailableError(
-        absl::StrCat("xds channel in TRANSIENT_FAILURE, connectivity error: ",
-                     status.ToString())));
+void XdsClient::ChannelState::SetChannelStatusLocked(absl::Status status) {
+  if (shutting_down_) return;
+  status = absl::Status(
+      status.code(), absl::StrCat("xDS channel for server ", server_.server_uri,
+                                  ": ", status.message()));
+  gpr_log(GPR_INFO, "[xds_client %p] %s", xds_client(),
```
yashykt, line 534: shouldn't this be protected by a trace flag?

src/core/ext/xds/xds_client.cc
```diff
@@ -522,27 +518,53 @@ void XdsClient::ChannelState::UnsubscribeLocked(const XdsResourceType* type,
...
-    // Notify all watchers of error.
-    gpr_log(GPR_INFO,
-            "[xds_client %p] xds channel for server %s in "
-            "state TRANSIENT_FAILURE: %s",
-            xds_client(), server_.server_uri.c_str(),
-            status.ToString().c_str());
-    xds_client_->NotifyOnErrorLocked(absl::UnavailableError(
-        absl::StrCat("xds channel in TRANSIENT_FAILURE, connectivity error: ",
-                     status.ToString())));
+void XdsClient::ChannelState::SetChannelStatusLocked(absl::Status status) {
```
yashykt, line 529: Might be a dumb question - Why isn't there a single queue for the error notifications and actual updates?

yashykt (commented):
Haven't yet read the tests and the fake xds transport

yashykt:
> We were failing to report anything to these watchers.
Would it have been eventually reported again? Is there a case where we would be hung forever, if there were no more updates possibly?

src/core/ext/xds/xds_client.cc
markdroth: There is.  All of the notifications go through the same `WorkSerializer`.

src/core/ext/xds/xds_client.cc
markdroth: Nope.  This is a message we want to log even if tracing is disabled, because it's an unusual situation that is important to log.

Note that even before this PR, this message was already being logged without a tracer enabled when the channel reported TRANSIENT_FAILURE.  The only thing new in this PR is that it will now also be logged when the ADS stream is closed.

markdroth (commented):
> Would it have been eventually reported again? Is there a case where we would be hung forever, if there were no more updates possibly?

Prior to this PR, the newly started watchers did not receive the previously-seen error, so they would not get any new notification until a new event occurred, whether that be an error or a valid update or a does-not-exist.  But there was no guarantee that a new event would ever occur within any particular interval, so they could indeed have been hung indefinitely.

test/core/xds/xds_transport_fake.h
```diff
@@ -0,0 +1,180 @@
...
+
+    bool HaveMessageFromClient();
+    absl::optional<std::string> WaitForMessageFromClient(
+        absl::Duration timeout);
+
+    void SendMessageToClient(absl::string_view payload);
+    void MaybeSendStatusToClient(absl::Status status);
+
+   private:
+    class RefCountedEventHandler : public RefCounted<RefCountedEventHandler> {
```
yashykt, line 72: would it make sense to just make `EventHandler` as a Refcounted class

yashykt (approved):
I think this is a good change. Are there more structural changes planned here? If so, can I request a design doc kind of thing which talks about where we want to be structurally finally

test/core/xds/xds_transport_fake.h
markdroth: In the production code, it doesn't actually need to be ref-counted; the ref-counting is needed only here in the test code.  I don't think it makes sense to add complexity to the production code for something we need only in the test.

PR #30828, opened 2022-09-02
Update to the event engine poller interface

Vignesh2208:
The Work(..) method of the interface is updated to take an absl::AnyInvocable callback as argument instead of returning a vector of closures.


src/core/lib/event_engine/poller.h
```diff
@@ -30,20 +30,23 @@ namespace experimental {
 // Work(...).
 class Poller {
  public:
-  // This initial vector size may need to be tuned
-  using Events = absl::InlinedVector<EventEngine::Closure*, 5>;
+  struct Ok {};
   struct DeadlineExceeded {};
   struct Kicked {};
-  using WorkResult = absl::variant<Events, DeadlineExceeded, Kicked>;
```
ctiller, line 37:
```
enum class WorkResult { kOk, kDeadlineExceeded, kKicked };
```

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
```diff
@@ -542,8 +541,14 @@ Poller::WorkResult Epoll1Poller::Work(EventEngine::Duration timeout) {
     if (pending_events.empty()) {
       return Poller::Kicked{};
     }
-    return pending_events;
   }
+  // Invoke the provided callback.
+  call_before_processing_events();
+  // Process all pending events inline.
+  for (auto& it : pending_events) {
+    (*it).ExecutePendingActions();
```
ctiller, line 549: it->ExecutePendingActions();

ctiller (changes requested):

src/core/lib/event_engine/poller.h
```diff
@@ -30,20 +30,23 @@ namespace experimental {
...
-  //  * absl::DeadlineExceeded if timeout occurred
-  //  * A collection of closures to execute, otherwise
-  virtual WorkResult Work(EventEngine::Duration timeout) = 0;
+  //  * Poller::Kicked if it was Kicked.
+  //  * Poller::DeadlineExceeded if timeout occurred
+  //  * Poller::Ok, otherwise indicating that the callback function was invoked
+  //    and some events were processed.
+  virtual WorkResult Work(
+      EventEngine::Duration timeout,
+      absl::FunctionRef<void()> call_before_processing_events) = 0;
```
drfloob, line 49: Let's call the callback something like `poll_again` to make it self-documenting.

src/core/lib/event_engine/poller.h
```diff
@@ -30,20 +30,23 @@ namespace experimental {
...
+  struct Ok {};
   struct DeadlineExceeded {};
   struct Kicked {};
-  using WorkResult = absl::variant<Events, DeadlineExceeded, Kicked>;
+  using WorkResult = absl::variant<Ok, DeadlineExceeded, Kicked>;
 
   virtual ~Poller() = default;
-  // Poll once for events, returning a collection of Closures to be executed.
+  // Poll once for events and process received events. The callback function
+  // is expected to be invoked prior to processing received events.
```
drfloob, line 40: Please be explicit about what the callback's responsibility is, which is to schedule another thread that calls Work. And because of that, I think it would be better advice to just say "The implementation should run the next polling cycle as quickly as possible to ensure continuous polling." And that may not mean calling it immediately!

For example, say this self-scheduling callback uses the same executor that's used inside the Work implementation to schedule read/write callbacks. If it's a LIFO executor, calling that function first may mean that the polling Work thread runs *after* read/write callbacks are processed. I'd let the implementation decide the best course of events, and otherwise make the API contract clear about whether it's a blocking or non-blocking call. If it's a blocking call, then the Work implementation has more control of the actual ordering, so I'd recommend that.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
Vignesh2208: Done.

src/core/lib/event_engine/poller.h
Vignesh2208: Done.

src/core/lib/event_engine/poller.h
Vignesh2208: Done.

Vignesh2208:
Friendly ping on this one :)

src/core/lib/event_engine/posix_engine/ev_poll_posix.cc
```diff
@@ -811,11 +810,18 @@ Poller::WorkResult PollPoller::Work(EventEngine::Duration timeout) {
...
   if (pending_events.empty()) {
     if (was_kicked_ext) {
-      return Poller::Kicked{};
+      return Poller::WorkResult::kKicked;
     }
-    return Poller::DeadlineExceeded{};
+    return Poller::WorkResult::kDeadlineExceeded;
+  }
+  // Invoke the provided callback.
+  poll_again();
```
drfloob, line 818: Given that a scheduler (async task executor) is already available here, why not make `poll_again` synchronous and put it on the scheduler? See the IOCP implementation, which would already prefer that. This simplifies the callback a bit, and groups the concurrency logic a bit better (visually, at least).

If you go that route, poll_again would either need to be an AnyInvocable or Closure, or the docstring would need to make it known that the function being referenced must outlive the call to `Work`. I'd prefer the former, which lets the type system enforce the constraint.

Vignesh2208: Done.

test/core/event_engine/posix/event_poller_posix_test.cc
```diff
@@ -673,18 +677,20 @@ class Worker : public grpc_core::DualRefCounted<Worker> {
 
  private:
   void Work() {
-    auto result = g_event_poller->Work(24h, [this]() {
+    auto result = g_event_poller->Work(24h, [self = Ref()]() {
```
drfloob, line 680: This lgtm. I'm not sure if we could avoid the ref churn here, but at least now this guarantees a living object until all Workers are finished.

drfloob (approved):

test/core/event_engine/posix/event_poller_posix_test.cc
Vignesh2208: Yes thats true

drfloob (approved):

src/core/lib/event_engine/poller.h
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/ev_poll_posix.cc
```diff
@@ -559,15 +560,13 @@ uint32_t PollEventHandle::BeginPollLocked(uint32_t read_mask,
   return mask;
 }
 
-EventEngine::Closure* PollEventHandle::EndPollLocked(int got_read,
-                                                     int got_write) {
-  EventEngine::Closure* closure = nullptr;
+bool PollEventHandle::EndPollLocked(int got_read, int got_write) {
```
ctiller, line 563: bool got_read/bool got_write? (let's get this bit fixed... it's been a while since we were c89!)

src/core/lib/event_engine/posix_engine/ev_poll_posix.cc
```diff
@@ -811,11 +810,18 @@ Poller::WorkResult PollPoller::Work(EventEngine::Duration timeout) {
...
+  }
+  // Schedule the provided callback.
+  scheduler_->Run(std::move(poll_again));
+  // Process all pending events inline.
+  for (auto& it : pending_events) {
+    it->ExecutePendingActions();
   }
-  return pending_events;
+  return Poller::WorkResult::kOk;
+  ;
```
ctiller, line 824: spurious ;

src/core/lib/event_engine/poller.h
```diff
@@ -30,20 +29,22 @@ namespace experimental {
...
-  //  * absl::AbortedError if it was Kicked.
-  //  * absl::DeadlineExceeded if timeout occurred
-  //  * A collection of closures to execute, otherwise
-  virtual WorkResult Work(EventEngine::Duration timeout) = 0;
+  //  * Poller::WorkResult::kKicked if it was Kicked.
+  //  * Poller::WorkResult::kDeadlineExceeded if timeout occurred
+  //  * Poller::WorkResult::kOk, otherwise indicating that the callback function
+  //  was schduled to run asynchonously and some events were processed.
+  virtual WorkResult Work(EventEngine::Duration timeout,
+                          absl::AnyInvocable<void()> poll_again) = 0;
```
ctiller, line 47: suggest making this `absl::FunctionRef<void()> schedule_poll_again`
and calling it directly in implementations

right now we'd probably write `Work(... [this] { Run(Poll); });` and that's fine, but it would also allow us to:
1. use `Closure` instead of `AnyInvocable` if that makes sense
2. mutate & see stack variables at the `Work` call site - which might be useful to save an allocation down the track

src/core/lib/event_engine/posix_engine/ev_poll_posix.cc
Vignesh2208: Oops fixed it now.

src/core/lib/event_engine/posix_engine/ev_poll_posix.cc
Vignesh2208: Removed.

src/core/lib/event_engine/poller.h
Vignesh2208: Updated to this model in the latest commit

Vignesh2208:
Updated per your comments. Could you ptal. Thanks!

test/core/event_engine/posix/event_poller_posix_test.cc
```diff
@@ -376,14 +380,8 @@ void WaitAndShutdown(server* sv, client* cl) {
...
     gpr_mu_unlock(&g_mu);
-    result = g_event_poller->Work(24h);
-    if (absl::holds_alternative<Poller::Events>(result)) {
-      auto pending_events = absl::get<Poller::Events>(result);
-      for (auto it = pending_events.begin(); it != pending_events.end(); ++it) {
-        (*it)->Run();
-      }
-      pending_events.clear();
-    }
+    result = g_event_poller->Work(24h, []() {});
```
ctiller, line 383: Maybe make these callbacks assert called/notcalled?

ctiller (approved):

test/core/event_engine/posix/event_poller_posix_test.cc
Vignesh2208: I think that could make the test flaky since there is a possibility that the test can finish in one iteration and sometimes it might require more than one iteration of Work

PR #30829, opened 2022-09-03
Cpp msbuild integration notes

apenn-msft:


yashykt:
@jtattermusch Can you take a look please?

drfloob:
That comment about being "in the works" is misleading, it was made 4 years ago, and the C# library is now in maintenance mode. @jtattermusch I'm curious if you know the history of the original intention there.

Otherwise, this PR describes how to use an unofficial external project that we don't help maintain (vcproj), so I don't think this documentation should be kept here. For instance, if some future vcproj changes require an update to these instructions, the official repo shouldn't be in the position of having to make changes accommodate them  - especially given we very likely won't notice the need to update them.

jtattermusch:
I agree with @drfloob. The historical context is that while Grpc.Tools was originally intended to be able to peform codegen even for Visual Studio C++ projects (not just C#). But AFAIK that isn't really being used much in practice and the Grpc.Tools usage for C# is much more important for us.
Since as @drfloob  has pointed out, we don't really maintain the vcpkg rules for gRPC C++ and we also don't test Grpc.Tools with C++, we shouldn't be adding a documentation here.

PR #30835, opened 2022-09-04
patch tests to use passed CMAKE_CXX_STANDARD

h-vetinari:
gRPC compiles against abseil, which has an ABI-dependence on the C++ standard used to compile, but [allows](https://github.com/grpc/grpc/blob/v1.48.1/CMakeLists.txt#L230-L236) overriding this with `CMAKE_CXX_STANDARD`. Compiling the cpp example should do the same.


veblush:
Thanks for the PR! and this looks good to me. @jtattermusch would you give this another lgtm?

jtattermusch (approved):
LGTM.

veblush (approved):

PR #30506, opened 2022-08-04
XdsClient: access certificate provider data via dependency injection

markdroth:
This makes it easier to use XdsClient without pulling in so much gRPC-specific code.

PR #30854, opened 2022-09-06
XdsClient: remove now-unnecessary XdsCertificateProviderPluginMapInterface

markdroth:
I added this interface in #30506, but it is no longer needed with the approach taken in #30759.

### DIFF
--- a/src/core/ext/xds/xds_bootstrap_grpc.cc
+++ b/src/core/ext/xds/xds_bootstrap_grpc.cc
@@ -378,7 +378,7 @@ grpc_error_handle GrpcXdsBootstrap::ParseCertificateProvider(
         config_json = it->second;
       }
     } else {
-      // "config" is an optional field, so use an empty JSON object.
+      // "config" is an optional field, so default to an empty JSON object.
       config_json = Json::Object();
     }
     // Try to instantiate the provider.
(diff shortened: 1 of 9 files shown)


src/core/ext/xds/xds_bootstrap_grpc.cc
```diff
@@ -461,8 +464,17 @@ std::string GrpcXdsBootstrap::ToString() const {
     parts.push_back("      },\n");
   }
   parts.push_back("}");
```
yashykt, line 466: This seems to want a '\n' at the end

src/core/ext/xds/xds_bootstrap_grpc.cc
```diff
@@ -376,24 +367,36 @@ grpc_error_handle GrpcXdsBootstrap::ParseCertificateProvider(
...
-        }
+        config_json = it->second;
       }
     } else {
-      // "config" is an optional field, so create an empty JSON object.
-      absl::Status status = certificate_provider_plugin_map_->AddPlugin(
-          instance_name, plugin_name, Json::Object());
-      if (!status.ok()) {
-        error_list.push_back(absl_status_to_grpc_error(status));
+      // "config" is an optional field, so use an empty JSON object.
```
yashykt, line 381: maybe a better comment would be - "config" is an optional field that defaults to an empty JSON object.

yashykt (approved):
Approving but minor comments

src/core/ext/xds/xds_bootstrap_grpc.cc
markdroth: Good catch, fixed.

PR #30855, opened 2022-09-06
Add GoLang master to supported Outlier Detection versions

zasweq:
@sergiitk


tools/run_tests/xds_k8s_test_driver/tests/outlier_detection_test.py
```diff
@@ -68,6 +68,8 @@ def is_supported(config: skips.TestConfig) -> bool:
             return config.version_gte('v1.49.x')
         if config.client_lang == _Lang.NODE:
             return config.version_gte('v1.6.x')
+        if config.client_lang == _Lang.GO:
+            return config.version_gte('master')
```
sergiitk, line 72:
```suggestion
        if config.client_lang == _Lang.GO:
            # TODO(zasweq): Update when the feature makes in a version branch.
            return config.version_gte('master')
```

sergiitk (approved):
Added a todo so we don't forget to update `master` to the version branch, once the feature is released.

PR #30869, opened 2022-09-07
Reduce ObjC bazel test flakiness

jtattermusch:
- according to my experiments, the flakiness we've been seeing with `//src/objective-c/tests:Interop*` targets seems to be linked to the CFStream bug on MacOS (b/133182964), and setting `GRPC_CFSTREAM_RUN_LOOP=1` does seem to reduce flakiness by a lot.
- reverting some questionable changes from https://github.com/grpc/grpc/pull/30287
    - the local interop tests should be very reliable (especially for the local ones there's nothing really anything in the path besides a bug that could cause flakiness) so enabling retries is only papering over the real problem, especially when we're enabling retries not for specific carefully selected tests, but as a blanket solution.
    - enabling retries for tests by default seems dangerous, especially in the scenario where the tests don't print any info about how many retries have been made for a particular testcase (which this PR fixes by at least adding a log line).
    - for remote interop tests, timeout of 3seconds seems too optimistic if we really want to have highly reliable tests.

We can look into removing the need for GRPC_CFSTREAM_RUN_LOOP=1 in the future (perhaps upgrading to newer MacOS version will help), but for now, getting tests more reliable is the priority.

### DIFF
--- a/src/objective-c/tests/Common/TestUtils.m
+++ b/src/objective-c/tests/Common/TestUtils.m
@@ -26,10 +26,10 @@
 #define NSStringize(x) @NSStringize_helper(x)
 
 // Default test flake repeat counts
-static const NSUInteger kGRPCDefaultTestFlakeRepeats = 2;
+static const NSUInteger kGRPCDefaultTestFlakeRepeats = 1;
 
 // Default interop local test timeout.
-const NSTimeInterval GRPCInteropTestTimeoutDefault = 3.0;
+const NSTimeInterval GRPCInteropTestTimeoutDefault = 15.0;
 
 NSString *GRPCGetLocalInteropTestServerAddressPlainText() {
   static NSString *address;
@@ -125,6 +125,10 @@ BOOL GRPCTestRunWithFlakeRepeats(XCTestCase *testCase, GRPCTestRunBlock testBloc
 
     if (result == XCTWaiterResultCompleted && assertionSuccess) {
       return YES;
+    }
+
+    if (!isLastRun) {
+      NSLog(@"test attempt %@ failed, will retry.", NSStringize(runs));
     }
     runs += 1;
   }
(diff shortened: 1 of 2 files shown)


jtattermusch:
The data:
Based on an adhoc run with --runs_per_test=50 (50 runs per test targets),
- I only got 1 failure for the InteropTestsLocalCleartext, which is much better than any previous attempts.
https://source.cloud.google.com/results/invocations/21114c8e-8ef9-44df-b58a-beb57d58a174/targets/%2F%2Fsrc%2Fobjective-c%2Ftests:InteropTestsLocalCleartext/tests
- All the failures I got for InteropTestsRemote and MacTests are instances of `InteropTestsRemote testKeepaliveWithV2API` - but from the data it looks that this particular test case is actually flaky by itself, independently of other test cases - confirmed by even local runs of it (I'll try to address in a separate PR).
e.g. https://source.cloud.google.com/results/invocations/21114c8e-8ef9-44df-b58a-beb57d58a174/targets/%2F%2Fsrc%2Fobjective-c%2Ftests:InteropTestsRemote/log

jtattermusch:
Based on my latest experiments, this will likely be superseded by https://github.com/grpc/grpc/pull/30897 and https://github.com/grpc/grpc/pull/30896 (together with switching the ObjC bazel job to macos Monterey)

dennycd (approved):

jtattermusch:
> Based on my latest experiments, this will likely be superseded by #30897 and #30896 (together with switching the ObjC bazel job to macos Monterey)

Superseded by PRs mentioned above and by switching the objc bazel test job to run on MacOS Monterey.

jtattermusch:
We can reopen this PR if the GRPC_CFSTREAM_RUN_LOOP=1 proves necessary.

PR #30874, opened 2022-09-07
Fix flakiness in EventEngineClientTest.ConnectToNonExistentListenerTest

Vignesh2208:
Replaces constant port number with a dynamically chosen unused port number and eliminates chances of port collisions.


drfloob (approved):

veblush (approved):
Thanks for the quick fix!

PR #30889, opened 2022-09-08
GcpObservability: Code for reading config from env var

yashykt:


test/cpp/ext/gcp/observability_config_test.cc
```diff
@@ -65,6 +69,135 @@ TEST(GcpObservabilityConfigJsonParsingTest, Defaults) {
   EXPECT_TRUE(config.project_id.empty());
 }
 
+TEST(GcpEnvParsingTest, NoEnvironmentVariableSet) {
+  auto config = GcpObservabilityConfig::ReadFromEnv();
```
veblush, line 73: Maybe resetting the environment variable would be required in case where the test is running while the variable is set.

veblush (approved):

test/cpp/ext/gcp/observability_config_test.cc
yashykt: I always unset the env var at the end of a test that sets it

PR #30844, opened 2022-09-05
Upgrade bazel version to bazel 5.3 (repo-wide)

jtattermusch:
- use a more up-to-date bazel version
- fixes objC bazel problem with python3 mentioned here: https://github.com/grpc/grpc/pull/29599

PR #30896, opened 2022-09-08
upgrade rules_apple and apple_support bazel deps

jtattermusch:
Followup for https://github.com/grpc/grpc/pull/30844.

Newer bazel rules for iOS tests - newer rules potentially means fewer infrastructure problems with ObjC bazel tests.

The update is also required to make the ObjC bazel test compatible with the new Kokoro MacOS monterey workers (that only have python3.9+ installed), without it the tests fail with `solve AttributeError: module 'plistlib' has no attribute 'readPlist' (plistlib.readPlist was removed in Python 3.9).

### DIFF
--- a/bazel/grpc_deps.bzl
+++ b/bazel/grpc_deps.bzl
@@ -376,20 +376,20 @@ def grpc_deps():
     if "build_bazel_rules_apple" not in native.existing_rules():
         http_archive(
             name = "build_bazel_rules_apple",
-            sha256 = "0052d452af7742c8f3a4e0929763388a66403de363775db7e90adecb2ba4944b",
+            sha256 = "a19cf84dd060eda50be9ba5b0eca88377e0306ffbc1cc059df6a6947e48ac61a",
             urls = [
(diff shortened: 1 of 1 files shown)


jtattermusch:
The ObjC test failure is a known one, seen many times on master.

jtattermusch:
Adhoc test result showing that with this change, the ObjC bazel tests are passing on a MacOS monterey kokoro worker: 
https://source.cloud.google.com/results/invocations/9471ec7d-c974-4df7-8259-468ab0c99603/targets

(currently we're on macos mojave).

veblush (approved):

dennycd (approved):
thanks for updating this

veblush:
This turned out that this requires Bazel 5.x or later but we need to support Bazel 4.x until Bazel 6.0 is released. So we need to roll this back.

jtattermusch:
> This turned out that this requires Bazel 5.x or later but we need to support Bazel 4.x until Bazel 6.0 is released. So we need to roll this back.

That's odd. We have distribtest that test with bazel 4.x and they didn't report any problems, so I don't understand what this issue was. https://source.cloud.google.com/results/invocations/ce23c06c-443a-4f77-b0ee-510a68c2bc71/log

Issue #10, opened 2015-01-13
Remove libevent submodule


ctiller:
It's no longer used.

PR #30901, opened 2022-09-08
xDS interop: Handle the edge case when rand deployment_id is all nums

sergiitk:
Fun edge case: when `rand_string()` happen to generate numbers only, yaml interprets `deployment_id` label value as an integer, but k8s expects label values to be strings.

K8s responds with a barely readable 400 Bad Request error: `ReadString: expects \" or n, but found 9, error found in #10 byte of ...|ent_id`.

Prepending deployment name forces `deployment_id` into a string, as well as it's just a better description.

ref b/245709149

cc @zasweq

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/framework/test_app/runners/k8s/k8s_base_runner.py
+++ b/tools/run_tests/xds_k8s_test_driver/framework/test_app/runners/k8s/k8s_base_runner.py
@@ -225,12 +225,28 @@ class KubernetesBaseRunner(base_runner.BaseRunner):
         return resource
 
     def _create_deployment(self, template, **kwargs) -> k8s.V1Deployment:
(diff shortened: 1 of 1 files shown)


sergiitk:
- [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/550beca6-90c2-4eee-b64c-cd627f959eaf)
- [grpc/core/master/linux/psm-security](https://source.cloud.google.com/results/invocations/418a00ef-fc2f-4483-bdc8-3f0bded8a332)
- [grpc/core/master/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/3b5e2aa2-1704-44bb-afab-b8574d4dca7c)

gnossen (approved):

PR #30930, opened 2022-09-09
Support alias rule

veblush:
Updated the script to find the right dependencies where some paths use `alias` rules.

### DIFF
--- a/tools/buildgen/extract_metadata_from_bazel_xml.py
+++ b/tools/buildgen/extract_metadata_from_bazel_xml.py
@@ -123,6 +123,7 @@ def _rule_dict_from_xml_node(rule_xml_node):
         'generator_function': None,
         'size': None,
         'flaky': False,
+        'actual': None,
     }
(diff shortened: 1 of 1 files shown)


tools/buildgen/extract_metadata_from_bazel_xml.py
```diff
@@ -462,6 +468,12 @@ def update_test_metadata_with_transitive_metadata(
             lib_dict['language'] = 'c++'
 
 
+def _get_actual_rule(bazel_rules, rule):
+    while rule != None and rule['actual'] != None:
```
jtattermusch, line 472: `!= None` is bad python afaik. `is not None` is better.

Actually `while rule and rule['actual']:` should be sufficient.

jtattermusch, line 472: add doc comment. sth like "Follows rule aliases until the actual rule is resolved"

PR #30931, opened 2022-09-09
subchannel: provide a non-racy mechanism for updating a data producer

markdroth:
This fixes a race spotted in our code by @dfawley, which can happen when two different channels are sharing the same subchannel and they both try to add an ORCA data producer at the same time in different threads.  The scenario is as follows:

1. Thread 1 [calls `subchannel->GetDataProducer()`](https://github.com/grpc/grpc/blob/ebcda5f0c1de9e0dddf320ab40180ab5a8a9dbaf/src/core/ext/filters/client_channel/lb_policy/oob_backend_metric.cc#L390) and gets null.
2. Thread 2 makes the same call and also gets null.
3. Thread 1 [instantiates a new producer](https://github.com/grpc/grpc/blob/ebcda5f0c1de9e0dddf320ab40180ab5a8a9dbaf/src/core/ext/filters/client_channel/lb_policy/oob_backend_metric.cc#L393), which [calls `subchannel->AddDataProducer()`](https://github.com/grpc/grpc/blob/ebcda5f0c1de9e0dddf320ab40180ab5a8a9dbaf/src/core/ext/filters/client_channel/lb_policy/oob_backend_metric.cc#L293), which succeeds.
4. Thread 2 now instantiates another new producer, which also calls `subchannel->AddDataProducer()`.  This will trigger [an assertion because the entry is already present](https://github.com/grpc/grpc/blob/ebcda5f0c1de9e0dddf320ab40180ab5a8a9dbaf/src/core/ext/filters/client_channel/subchannel.cc#L792).

The fix is to replace the separate `GetDataProducer()` and `AddDataProducer()` methods with a single `GetOrAddDataProducer()` method, which allows the ORCA data producer to do both the check and the add in a single step.


src/core/ext/filters/client_channel/lb_policy/oob_backend_metric.cc
```diff
@@ -384,14 +384,23 @@ OrcaWatcher::~OrcaWatcher() {
...
+          created = true;
+        }
+      });
+  // If we just created the producer, start it.
+  // This needs to be done outside of the lambda passed to
+  // GetOrAddDataProducer() to avoid deadlocking by re-acquiring the
+  // subchannel lock while already holding it.
+  if (created) producer_->Start(subchannel->Ref());
   // Register ourself with the producer.
   producer_->AddWatcher(this);
```
dfawley, line 405: Does the same race exist if the second initializer calls `AddWatcher` on the producer before the first one calls `Start`?

markdroth: I don't think so.  The producer has its own mutex, which it acquires inside of `AddWatcher()`.

dfawley: Are you sure?  If `AddWatcher` is called before `Start`, couldn't you lose a watch?  `MaybeStartStreamLocked` would be a nop in `AddWatcher` and I see nothing in `Start` that looks for existing watches.

Maybe `AddWatcher` needs to optionally start if it hasn't been started yet?

markdroth: I think this would already be handled properly.  Note that `Start()` does not actually start the ORCA stream; what it starts is the connectivity state watch on the subchannel.  When the connectivity state watch reports READY, then if the watcher map is non-empty, we start the ORCA stream:

https://github.com/grpc/grpc/blob/bcd8c991e62b4d6326e6dc61ff1aa3bea2b3f35a/src/core/ext/filters/client_channel/lb_policy/oob_backend_metric.cc#L371

So I don't think it matters if `AddWatcher()` is called before `Start()`.  The only distinction that does matter is whether `AddWatcher()` is called before or after we see READY state, and I think the code is already handling this properly.

dfawley: Ah, I see; thanks for explaining.

yashykt (approved):

Issue #30933, opened 2022-09-10
Ruby 1.49.0.pre1 - grpc getting gRPC dynamic C core load error on x64-mingw-ucrt


johnnyshields:
@apolcyn I've found a critical issue with grpc Ruby [1.49.0.pre1](https://rubygems.org/gems/grpc/versions/1.49.0.pre1) on a clean install of Windows / Ruby 3.1 x64-mingw-ucrt. The symptom is the error `Couldn't find or load gRPC's dynamic C core (LoadError)` when loading the gem.

I've found that renaming `grpc_c.64-ucrt.ruby` to `grpc_c.64-msvcrt.ruby` fixes the issue. Note that the `-ucrt` gem comes with the following binaries (msvcrt are empty):

```
grpc_c.32-msvcrt.ruby - 0 KB
grpc_c.64-msvcrt.ruby - 0 KB
grpc_c.64-ucrt.ruby - 9,650 KB
```

(This means that in the builds I've made so far, it's actually been loading the msvcrt binary on ucrt the whole time. Oops 🤦‍♂️)

It would be ideal to fix this before the full release of 1.49.0.

```
C:/ruby31/bin/ruby.exe -x C:\workspace\app\bin\bundle exec C:\ruby31\bin\ruby.exe C:/workspace/app/bin/rails server -b 127.0.0.1 -p 3000 -e development
C:/ruby31/lib/ruby/gems/3.1.0/gems/bootsnap-1.12.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require': Couldn't find or load gRPC's dynamic C core (LoadError)
        from C:/ruby31/lib/ruby/gems/3.1.0/gems/bootsnap-1.12.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
        from C:/ruby31/lib/ruby/gems/3.1.0/gems/grpc-1.49.0.pre1-x64-mingw-ucrt/src/ruby/lib/grpc/grpc.rb:20:in `<main>'
        from C:/ruby31/lib/ruby/gems/3.1.0/gems/grpc-1.49.0.pre1-x64-mingw-ucrt/src/ruby/lib/grpc.rb:19:in `require_relative'
        from C:/ruby31/lib/ruby/gems/3.1.0/gems/grpc-1.49.0.pre1-x64-mingw-ucrt/src/ruby/lib/grpc.rb:19:in `<main>'
[...]

PR #30934, opened 2022-09-10
v1.49.x -- Ruby: Output a single binary file per platform-specific gem

johnnyshields:
In #30933, I discovered the Ruby UCRT gem attempting to load the MSVCRT binary, and failing because that binary file is a zero-byte placeholder. Specifically, the gem root dir looks like this:

```
grpc_c.32-msvcrt.ruby - 0 KB
grpc_c.64-msvcrt.ruby - 0 KB
grpc_c.64-ucrt.ruby - 9,650 KB
```

It would be possible to fix the reference so the UCRT binary is loaded (probably a C build flag issue), *however*...

Rather than having 3 binaries files (1 real + 2 placeholders), the most elegant solution is to only create 1 binary file for each platform-specific, named `grpc_c.ruby`, which is:
- On Windows: the correct binary file for the given Windows architecture
- On Linux/Mac: a 0-byte placeholder

For reference, the UCRT gem root dir will now look like this:

```
grpc_c.ruby - 9,650 KB
```

This solution unambiguously avoids loading the wrong binary, and also gets rid of a lot of cruft in the build code at the same time. Obviously we'll never need more than one `grpc_c` binary for any given platform gem.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -38,7 +38,7 @@ Rake::ExtensionTask.new('grpc_c', spec) do |ext|
       |file| file.start_with?(
         "src/ruby/bin/", "src/ruby/ext/", "src/ruby/lib/", "src/ruby/pb/")
     }
-    spec.files += %w( etc/roots.pem grpc_c.32-msvcrt.ruby grpc_c.64-msvcrt.ruby grpc_c.64-ucrt.ruby )
+    spec.files += ['etc/roots.pem', GRPC_RUBY_BINARY]
   end
 end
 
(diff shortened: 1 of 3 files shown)


johnnyshields:
@apolcyn ready for your review.

johnnyshields:
@apolcyn any update on this?

apolcyn:
Thanks for the investigation, @johnnyshields.

This PR looks reasonable to me. But given the size of this PR, and the fragility of the windows build, I'm a little apprehensive to merge this on 1.49. We'd need to release this change straight into a 1.49.1 patch release, without a pre-release, so we'd have more risk of breaking the whole windows build.

So I'd prefer to fix the build issue, which seems to not be setting the [macro correctly](https://github.com/grpc/grpc/blob/a7a89917b5c387f95bdf6804274fc6ee95e06850/src/ruby/ext/grpc/rb_loader.c#L27), if possible. That said if it's very hard, I'm fine with this as a fall back.

Do you have any idea what's wrong there? The current `extconf.rb` code looks right w.r.t. detection of the ucrt platform, and I checked that rake-compiler-dock does fill in `x64-mingw-ucrt` as the RUBY_PLATFORM when running that code.

johnnyshields:
@apolcyn unfortunately I've exhausted my bandwidth I can spend investigating this issue. I'm fine to get this either in 1.49.1 or 1.50.0. I believe it is actually possible to cut a pre-release of a tiny version `v1.49.1-pre1` and I'm glad to test when you have a gem cut either way.

apolcyn:
I think I've found the cflags issue - opened https://github.com/grpc/grpc/pull/31051. I'm backporting that and aiming to get that into 1.49.1

johnnyshields:
@apolcyn thank you. I'm closing this PR (1.49.x backport) however will leave the PR for master open.

PR #30954, opened 2022-09-13
Added insecure channels counter

AlienCookie:
@ctiller


src/core/lib/surface/channel.cc
```diff
@@ -121,6 +121,18 @@ absl::StatusOr<RefCountedPtr<Channel>> Channel::CreateWithBuilder(
     return status;
   }
 
+  auto auth_context = channel_args.GetObject<grpc_auth_context>();
```
ctiller, line 124: Suspect we're missing a header here.

ctiller:
looks like this does not compile

PR #30960, opened 2022-09-13
Remove `grpc_codegen`

ralphchung:

In https://github.com/grpc/grpc/pull/30570, build target `grpc_codegen` has been merged to `grpc_base`. We no longer need this.


Vignesh2208 (approved):

BUILD
```diff
@@ -2169,7 +2167,7 @@ grpc_cc_library(
     deps = [
         "gpr",
         "gpr_spinlock",
-        "grpc_codegen",
+        "grpc_public_hdrs",
```
ctiller, line 2170: It seems like here and in other places we want some concrete target, not the headers only target - or am I missing something?

Please check that we're only consuming interface not implementation, and either fix the dependency or include a note in the review for why this is safe here and in all places we're using this target.

ctiller (approved):

PR #30964, opened 2022-09-13
Print info when server starts for Python helloworld

XuanWang-Amos:
Print some info to indicate the server/client started successfully.

### DIFF
--- a/examples/python/helloworld/greeter_client.py
+++ b/examples/python/helloworld/greeter_client.py
@@ -26,6 +26,7 @@ def run():
     # NOTE(gRPC Python Team): .close() is possible on a channel and should be
     # used in circumstances in which the with statement does not fit the needs
     # of the code.
+    print("Will try to greet world ...")
     with grpc.insecure_channel('localhost:50051') as channel:
         stub = helloworld_pb2_grpc.GreeterStub(channel)
         response = stub.SayHello(helloworld_pb2.HelloRequest(name='you'))
--- a/examples/python/helloworld/greeter_server.py
+++ b/examples/python/helloworld/greeter_server.py
@@ -28,10 +28,12 @@ class Greeter(helloworld_pb2_grpc.GreeterServicer):
 
 
 def serve():
+    port = '50051'
     server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
     helloworld_pb2_grpc.add_GreeterServicer_to_server(Greeter(), server)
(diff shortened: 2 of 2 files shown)


examples/python/helloworld/greeter_client.py
```diff
@@ -26,6 +26,7 @@ def run():
     # NOTE(gRPC Python Team): .close() is possible on a channel and should be
     # used in circumstances in which the with statement does not fit the needs
     # of the code.
+    print("Will try to greet world ...")
```
gnossen, line 29: This is a little bit cryptic. What value do you think this log adds?

examples/python/helloworld/greeter_server.py
```diff
@@ -28,10 +28,12 @@ def SayHello(self, request, context):
 
 
 def serve():
+    port = '50051'
     server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
     helloworld_pb2_grpc.add_GreeterServicer_to_server(Greeter(), server)
-    server.add_insecure_port('[::]:50051')
+    server.add_insecure_port('[::]:' + port)
     server.start()
+    print("Server started, listening on " + port)
```
gnossen, line 36: If we're going to be reporting the port that we're listening on, we want the gRPC library to tell us the resolved port. In the case of port 50051, this will also be 50051. But If `port` is 0, the resolved port may be anything: 

`resolved_port = server.add_insecure_port('[::]:' + port)`

Then print `resolved_port`. You may also consider letting port be specified as a flag for the process.

gnossen (commented):
Thanks for the contribution. Our examples haven't seen any changes in quite a while.

examples/python/helloworld/greeter_client.py
XuanWang-Amos: I'm using the same wording in [Java quick start](https://grpc.io/docs/languages/java/quickstart/#run-the-example), this indicates the client is up and running and about to start the actual RPC call.

examples/python/helloworld/greeter_server.py
XuanWang-Amos: Make sense, I think fixed port is enough for quick start purpose, what do you think?

examples/python/helloworld/greeter_client.py
gnossen: Fair enough. Matching other implementations is a decent argument.

gnossen (approved):

PR #30966, opened 2022-09-13
Generate experiments during sanitization

drfloob:
This ensures the code always stays in sync with the configuration. Also adds gen_experiments to sanity checks.


ctiller:
This is a good thing to do, probably even necessary. It needs clang-format to run with it though, or it's going to be wrong (and writing a code generator to emit clang-format identical code is beyond the scope of what I'd be willing to do)

ctiller:
We probably need to get a clang-format docker image setup that can clang-format a set of passed in files.

ctiller (approved):

PR #30971, opened 2022-09-13
Add support to explicitly add priority for the handshakers using enums

anicr7:
Currently the order of the handshakers is controlled by a single bool(at_start). This doesn't allow for more complex use cases where the handshaker has to be done before tcp connect handshaker for example.

By explicitly adding enums that specify the priority, we allow for a cleaner abstraction for registering handshakers.


src/core/lib/transport/handshaker_factory.h
```diff
@@ -35,11 +37,38 @@ namespace grpc_core {
 
 class HandshakeManager;
 
+// Enum representing the priority of the client handshakers.
+// The order of the client handshakers is decided by the priority.
+// For example kPreTCPConnect handshakers are called before kTCPConnect and so
+// on.
+enum class HandshakerClientPriority : int {
+  kPreTCPConnect,
```
ctiller, line 45: Probably worth expanding the name and including one line on the kinds of things expected in each bucket

src/core/lib/transport/handshaker_factory.h
```diff
@@ -35,11 +37,38 @@ namespace grpc_core {
...
+  kHTTPConnect,
+  kSecurity,
+};
+
+// Enum representing the priority of the server handshakers.
+// The order of the server handshakers is decided by the priority.
+// For example kReadHeadSecurity handshakers are called before kSecurity and so
+// on.
+enum class HandshakerServerPriority : int {
+  kReadHeadSecurity,
```
ctiller, line 56: here too

src/core/lib/transport/handshaker_factory.h
```diff
@@ -35,11 +37,38 @@ namespace grpc_core {
...
+  kSecurity,
+};
+
+using HandshakerPriority =
```
ctiller, line 60: This is probably reaching for a too large a hammer for the job.

Consider folding HandshakerClientPriority/HandshakerServerPriority into one enum ... we can comment client/server only if necessary (heck, we can introduce a IsServer/IsClient if necessary too... but I don't think it's practically useful)

ctiller (changes requested):

src/core/lib/transport/handshaker_factory.h
anicr7: Done.

src/core/lib/transport/handshaker_factory.h
anicr7: Done, let me know if this good enough or if I should add more information.

src/core/lib/transport/handshaker_factory.h
anicr7: I initially had a single priority for both clients and servers but I thought it looked odd to have PreTCPConnectHandshaker for Server side handshakers. Adding a comment sounds good to me considering the current limited usage.

ctiller (approved):

src/core/lib/transport/handshaker_factory.h
```diff
@@ -35,11 +35,37 @@ namespace grpc_core {
...
+};
+
 class HandshakerFactory {
  public:
   virtual void AddHandshakers(const ChannelArgs& args,
                               grpc_pollset_set* interested_parties,
                               HandshakeManager* handshake_mgr) = 0;
+  // Return the priority associated with the handshaker.
+  // The priority can be either HandshakerClientPriority or
+  // HandshakerServerPriority depending on the type of handshaker.
```
markdroth, line 67: This comment looks out of date.  There is no `HandshakerServerPriority`.

src/core/lib/transport/handshaker_factory.h
```diff
@@ -35,11 +35,37 @@ namespace grpc_core {
 
 class HandshakeManager;
 
+// Enum representing the priority of the handshakers.
+// The order of the handshakers is decided by the priority.
+// For example kPreTCPConnect handshakers are called before kTCPConnect and so
+// on.
+enum class HandshakerPriority : int {
```
markdroth, line 42: Suggest moving this inside of the `HandshakerFactory` class.

src/core/lib/transport/handshaker_factory.h
```diff
@@ -35,11 +35,37 @@ namespace grpc_core {
...
+  kPreTCPConnectHandshakers,
+  // Handshakers responsible for the actual TCP connect establishment.
+  // Applicable mainly for Client handshakers.
+  kTCPConnectHandshakers,
+  // Handshakers responsible for the actual HTTP connect established. Applicable
+  // mainly for Client handshakers.
+  kHTTPConnectHandshakers,
+  // Handshakers that should be called before security handshakes but after
+  // connect establishment. Applicable mainly for Server handshakers currently.
+  kReadHeadSecurityHandshakers,
```
markdroth, line 54: This should be called `kReadAheadSecurityHandshakers` ("Ahead" instead of "Head").

markdroth (approved):
This looks good overall.  Comments are all cosmetic.

src/core/lib/transport/handshaker_factory.h
anicr7: Done, removed.

src/core/lib/transport/handshaker_factory.h
anicr7: Done.

src/core/lib/transport/handshaker_factory.h
anicr7: Done.

markdroth (approved):

PR #30978, opened 2022-09-14
EventEngine::WaitForPendingTasks

drfloob:
Adds a new public API method and multiple implementations of `EventEngine::WaitForPendingTasks`.

This requires a cherrypick, and will need an update on the gRFC.


src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -92,6 +95,18 @@ void PosixEventEngine::Run(EventEngine::Closure* closure) {
...
 }
 
+absl::Status PosixEventEngine::WaitForPendingTasks(Duration timeout) {
+  auto deadline = ToTimestamp(timer_manager_.Now(), timeout);
+  while (executor_.IsBusy()) {
+    if (timer_manager_.Now() > deadline) {
+      return absl::DeadlineExceededError(
+          "Timed out waiting for executor to become idle.");
+    }
+    absl::SleepFor(absl::Milliseconds(33));
```
markdroth, line 105: How was this constant chosen?  Maybe document?

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -108,6 +108,18 @@ void WindowsEventEngine::Run(EventEngine::Closure* closure) {
...
 }
 
+absl::Status WindowsEventEngine::WaitForPendingTasks(Duration timeout) {
+  auto deadline = ToTimestamp(timer_manager_.Now(), timeout);
+  while (executor_.IsBusy()) {
+    if (timer_manager_.Now() > deadline) {
+      return absl::DeadlineExceededError(
+          "Timed out waiting for executor to become idle.");
+    }
+    absl::SleepFor(absl::Milliseconds(33));
```
markdroth, line 118: Same here.

markdroth (commented):
We should probably have tests for this method.

src/core/lib/event_engine/posix_engine/posix_engine.cc
drfloob: Totally arbitrary, and I suspect 30 checks per second is overkill for this method. I considered using exponential backoff from say 20 milliseconds up to 1/4 the timeout (again arbitrary numbers). We don't have real workloads yet, so it's hard to tune. Do you have a recommendation?

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -92,6 +95,18 @@ void PosixEventEngine::Run(EventEngine::Closure* closure) {
   executor_.Run(closure);
 }
 
+absl::Status PosixEventEngine::WaitForPendingTasks(Duration timeout) {
```
ctiller, line 98: What actions would one take if the timeout expired?

ctiller, line 98: Is this function equivalent to (sans timeout):
```
executor_.PrepareFork();
executor_.PostFork();
```
in its observable behavior?

(if yes: should we formalize that pair and use it?)

drfloob: In benchmark and testing cases, I'd expect we assert a failure. If some production code is using it to attempt a safe engine shutdown, maybe logging to indicate a leak, similar to what we have on gRPC shutdown today.

drfloob: To the application, yes, given that there are no closures pending when `PrepareFork` returns.

Mechanically (and observable to system monitoring and whatnot), that combination would start and stop N threads, which is unnecessary.

drfloob:
@markdroth I've added some test suite tests for the new method, and plan to rely on the new backoff changes you're reviewing here: https://github.com/grpc/grpc/pull/30993

This PR otherwise depends on @ctiller fixing and relanding the thread_pool improvements.

src/core/lib/event_engine/posix_engine/posix_engine.cc
ctiller: Yeah... I'm trying to judge the trade off of adding another shutdown path versus refining an existing one. Yes one path would shutdown N threads: my assumption is we wouldn't care.

ctiller: It may be worth going for a `WaitForPendingTasksOrDie(timeout)` API then - it doesn't seem like it'd be safe to proceed if this fails.

drfloob: That could work. The one exception I can imagine at the moment is an engine that wants to offer a best effort flush of its queue on shutdown or fork, but that can be managed internally to the engine.

Some of the tests in the test suite will need to be death tests now, which are not terribly friendly on all platforms. Do you think we should make the change anyway?

drfloob: That's a fair argument. Thinking through caveats, it could get weird if a fork occurs when we're attempting to drain the queue using the same approaches. We'll need some special handling for those events overlapping if we want to reuse the code that starts & stops threads. As the PR exists now, I think it would work fine without that extra overhead. But the reuse is doable. WDYT?

src/core/lib/event_engine/thread_pool.cc
```diff
@@ -93,32 +94,41 @@ void ThreadPool::Add(absl::AnyInvocable<void()> callback) {
...
+    GPR_ASSERT(state_ != QueueState::kRunning);
   } else {
-    GPR_ASSERT(state_ == State::kRunning);
+    GPR_ASSERT(state_ == QueueState::kRunning);
   }
   state_ = state;
   cv_.SignalAll();
 }
 
+bool ThreadPool::Queue::IsBusy(int expected_idle_threads) {
```
ctiller, line 127: I don't think this is implementable once we have a work stealing implementation... Indeed I'm not convinced it is now.

The problem now:
I don't think this counts threads that are not waiting but are currently executing callbacks - for that we need to wait for thread termination.

The problem with work queues:
This gets much harder because now we can't lock the entire data structure at once, because doing so would imply having a thread pool wide mutex on the enqueue path - which we endeavor to avoid.

I suspect that the only reasonable way to get this done is to do something like the thread quiesce we do for fork, and in #30996 on the shutdown path too.

src/core/lib/event_engine/posix_engine/posix_engine.cc
ctiller: The best effort flush thing is a really dangerous pattern. It's never worked well for iomgr, I should not have added it, and we should not bring it forward. It's a relic pattern from when we didn't have sanitizers, but since it leaves the application in an undefined state it leaves the application vulnerable to data corruption or other problems post call.

Which is to say, either the application relies on callbacks being done (meaning this should be an OrDie function) or applications do not rely on being able to do this and so we should not export this function at all.

drfloob: Changed to `WaitForPendingTasksOrDie`, with one death test.

drfloob:
After discussing it, this is sufficiently controversial that I'll investigate other approaches for the problems that this feature solves.

PR #30985, opened 2022-09-14
Fix the build script

veblush:


drfloob (approved):
Thanks Esun! The failures look unrelated.

PR #30992, opened 2022-09-14
Revert "Revert "[chttp2] fix stream leak with queued flow control update and absence of writes (#30907)" (#30991)"

apolcyn:
Rolls forward https://github.com/grpc/grpc/pull/30907 after fixing the `grpc_iomgr_count_objects_for_testing` API.

Tested this works internally.


ctiller (approved):
Please use the 'import/copybara' check to do a test import

apolcyn:
import/copybara passed, and flakes here are pre-existing

PR #30993, opened 2022-09-14
Add Backoff::TimeoutUntilNextAttempt

drfloob:


drfloob:
Per discussion with @markdroth , this requires a cross-language agreement about backoff behavior: whether it's interpreted as an absolute frequency or as a wait between one event and another. Closing this PR, we'll circle back to an implementation when that's decided.

PR #30997, opened 2022-09-15
fixing php 8.2 deprecations

brettmc:
PHP 8.2 RC's are out, and grpc complains that `Creation of dynamic property Grpc\Call::$channel is deprecated`.
Adding a property for `channel` resolves this, and I also fixed a bunch of deprecations and minor issues in the tests themselves.

### DIFF
--- a/src/php/tests/unit_tests/TimevalTest.php
+++ b/src/php/tests/unit_tests/TimevalTest.php
@@ -18,6 +18,8 @@
  */
 class TimevalTest extends \PHPUnit\Framework\TestCase
 {
+    private $time;
+
     public function setUp(): void
     {
     }
@@ -62,9 +64,12 @@ class TimevalTest extends \PHPUnit\Framework\TestCase
         $this->assertSame('Grpc\Timeval', get_class($this->time));
     }
 
+    /**
+     * @todo in PHP 8.1+ implicit conversion from float to int is deprecated
+     */
     public function testConstructorWithFloat()
     {
-        $this->time = new Grpc\Timeval(123.456);
+        $this->time = new Grpc\Timeval((int)123.456);
         $this->assertNotNull($this->time);
         $this->assertSame('Grpc\Timeval', get_class($this->time));
         $timeFromInt = new Grpc\Timeval(123);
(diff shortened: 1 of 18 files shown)


src/php/tests/unit_tests/TimevalTest.php
```diff
@@ -62,9 +64,12 @@ public function testConstructorWithHex()
         $this->assertSame('Grpc\Timeval', get_class($this->time));
     }
 
+    /**
+     * @todo in PHP 8.1+ implicit conversion from float to int is deprecated
+     */
     public function testConstructorWithFloat()
     {
-        $this->time = new Grpc\Timeval(123.456);
+        $this->time = new Grpc\Timeval((int)123.456);
```
brettmc, line 72: I wasn't sure what to do with this test, since it no longer tests what it claims to. Perhaps revert the cast, and skip the test for 8.1+ ? (see https://php.watch/versions/8.1/deprecate-implicit-conversion-incompatible-float-string)

brettmc:
@stanley-cheung I think you're the right person to review this PR for me.

stanley-cheung:
Sorry for the delay. Wasn't aware of this until just now. Will look into this. Thanks for helping fix this.

stanley-cheung:
@brettmc Can I confirm that, once this PR is patched in, `src/php/bin/build_all_docker_images.sh` and `src/php/bin/run_all_docker_images.sh` run to completions with no errors?

brettmc:
@stanley-cheung - confirming that both `src/php/bin/build_all_docker_images.sh` and `src/php/bin/run_all_docker_images.sh` run to completion for me with this patch.

stanley-cheung (approved):
LGTM. Thanks for the contributions!

gsdaemon:
https://github.com/grpc/grpc/pull/30997/commits/2a6cf72caa3972d4c971105e3030cf2a209f04e9
this was not merged unfortunately and it still shows deprecation notices in php 8.2
`Notice (8192): Creation of dynamic property Grpc\Call::$channel is deprecated in grpc/src/lib/AbstractCall.php on line 60`

brettmc:
Hi @gsdaemon
This was released in v1.51.0: https://github.com/grpc/grpc/blob/v1.51.0/src/php/ext/grpc/call.c#L627 and is also in the most recently released version: https://github.com/grpc/grpc/blob/v1.52.1/src/php/ext/grpc/call.c#L627
Can you confirm the extension version you're using?
It looks like pecl by default is installing an earlier version (I get 1.47.4), so you could try to install an explicit version of 1.51+.
That said, it looks like 1.51+ has a bug that you don't want (I can't run my tests to confirm that the notice has gone away on any version of grpc that has the fix)

PR #31003, opened 2022-09-15
Clean up the TODO item in build target `grpc_base` -- part `slice`

ralphchung:

This PR fixes the TODO item in the `slice` part of build target `grpc_base`.


src/core/BUILD
```diff
@@ -1087,6 +1087,7 @@ grpc_cc_library(
         "absl/hash",
         "absl/strings",
     ],
+    visibility = ["@grpc:alt_grpc_base_legacy"],
```
ananda1066, line 1090: why is this visibility field needed?

ralphchung: We want to reserve compatibility. The `grpc_base` (the target we want to remove from) has this visibility. And I believe this is mostly used for internal.

ananda1066 (approved):

PR #31019, opened 2022-09-16
[ObjC] Run ios perf test from bazel

HannahShiSFB:
https://github.com/grpc/grpc-ios/issues/76
Migrate gRPC ObjC PerfTest


jtattermusch:
Note that currently the iOS perf test aren't being run by the CI at all (since AFAIK there is currently no kokoro job that would run run_tests.py tasks with the "ios-perf-*" prefix.
So adding the ios-perf tests to bazel isn't required to reach parity between the run_tests.py objc tests vs the objc bazel tests.

src/objective-c/tests/PerfTests/PerfTests.m
```diff
@@ -359,7 +359,8 @@ - (void)testMultipleChannels {
   NSMutableArray<NSString *> *addrs = [NSMutableArray arrayWithCapacity:kNumAddrs];
   NSMutableArray<RMTTestService *> *services = [NSMutableArray arrayWithCapacity:kNumAddrs];
   for (int i = 0; i < kNumAddrs; ++i) {
-    addrs[i] = [NSString stringWithFormat:@"127.0.0.%d", (i + 1)];
+    // http://readme.localtest.me/
+    addrs[i] = [NSString stringWithFormat:@"%d.localtest.me", (i + 1)];
```
dennycd, line 363: a bit more details on why we need to switch to localtest.me

HannahShiSFB: In order to connect to local server using multiple channels, prior to this PR, In `src/objective-c/tests/run_one_test.sh` script, it attaches 127.0.0.x IP addresses to localhost and remove them after the test, which requires root permission
`sudo ifconfig lo0 alias 127.0.0.$i up`
`sudo ifconfig lo0 -alias 127.0.0.$i`

This change utilize the fact that `xxx.localtest.me` always resolve to 127.0.0.1, so we use different domain names to get multiple channels without the need of root permission and custom script run form bazel.

For details of localtest.me, check https://readme.localtest.me/

dennycd (approved):
rest lgtm, thanks for the patch

src/objective-c/tests/BUILD
```diff
@@ -234,6 +263,23 @@ grpc_objc_ios_unit_test(
...
+    name = "PerfTestsPosix",
+    deps = [
+        "PerfTests-lib-Posix",
+    ],
+)
+
+grpc_objc_ios_unit_test(
+    name = "PerfTests",
+    env = {
+        "GRPC_CFSTREAM_RUN_LOOP": "1",
```
jtattermusch, line 276: AFAIK  GRPC_CFSTREAM_RUN_LOOP shouldn't be necessary anymore when running on the macos monterey images.

PR #31024, opened 2022-09-16
xDS resource validation: identify extensions by type_url instead of name

markdroth:
I found a couple of places where we were still identifying xDS extensions by the legacy name instead of the type_url of the extension config:
- TransportSocket (both UpstreamTlsContext in CDS and DownstreamTlsContext in LDS)
- CDS ClusterType (aggregate clusters)

Envoy is actively working on eliminating the last vestiges of the legacy names, so we should do the same thing.

### DIFF
--- a/src/core/ext/xds/xds_cluster.cc
+++ b/src/core/ext/xds/xds_cluster.cc
@@ -26,6 +26,7 @@
 #include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_join.h"
+#include "absl/strings/strip.h"
 #include "envoy/config/cluster/v3/circuit_breaker.upb.h"
 #include "envoy/config/cluster/v3/cluster.upb.h"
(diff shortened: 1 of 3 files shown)


yashykt:
Are we sure that TD is sending these new types? Let's trigger the xDS security job to make sure we are not breaking anything?

markdroth:
TD must already be sending these types, because even without this PR, we are already parsing those types from the value of the `Any` fields.

I'm not sure how to trigger the xDS interop jobs, but if you have a simple way to do so, please feel free to do it.

yashykt:
triggered job - https://fusion2.corp.google.com/invocations/bd994075-f28f-4a96-b580-a208fd2497de

yashykt (approved):
LGTM, but please wait for the job to be triggered job to be successful before mergin

markdroth:
xDS security interop job passed, so I'll go ahead and merge this.

PR #31025, opened 2022-09-16
xDS cluster resource type: use ValidationErrors and add unit test

markdroth:
Also adds some missing validation for outlier detection parameters, as per [gRFC A50](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md#validation).  I added this both to the xDS resource validation and to the LB policy config parsing code, with appropriate tests.


src/core/ext/xds/xds_common_types.cc
```diff
@@ -45,6 +46,25 @@
...
 namespace grpc_core {
 
+//
+// ParseDuration()
+//
+
+Duration ParseDuration(const google_protobuf_Duration* proto_duration,
+                       ValidationErrors* errors) {
+  int64_t seconds = google_protobuf_Duration_seconds(proto_duration);
+  if (seconds < 0 || seconds > 315576000000) {
```
yashykt, line 56: we are using these constants at multiple places. Let's define them in one place and use that name everywhere

yashykt (commented):
I seem to be missing something on how ScopedField works

yashykt (approved):
I don't have any objections to this structure but it does seem like in some places instead of a single return we have more extra lines

src/core/ext/xds/xds_common_types.cc
markdroth: We're using them in exactly two places, here and in the JSON parsing code.  I don't really see a convenient common dependency where we could put the constant, and I don't think the duplication is going to cause enough of a problem to warrant the overhead of creating a new module just for this constant.

PR #31027, opened 2022-09-16
Replace EventEngine::Promise with grpc_core::Notification

drfloob:


Vignesh2208:
Lgtm. Could you do a test import and check if this requires a cherrypick. Thanks!

Vignesh2208 (approved):

drfloob:
Master appears to be broken at the moment, so I couldn't do a cherrypick test, but I searched and did not find any uses of the Promise class outside of this codebase.

PR #31035, opened 2022-09-17
Add an end-to-end test that includes client-side fork

drfloob:
This continues where #29607 left off.

I've added a continuous integration job that exercises all tests labeled `"fork_test"`. Fork is currently broken on the `epoll1` poller, but appears to work fine with `poll` (~2000 runs so far).


drfloob:
I've confirmed there's a ~2% flake rate for the fork test, all failures are timeouts. https://source.cloud.google.com/results/invocations/b059a549-dd12-4a88-84e9-caa85d4aa8df/targets. The Failures manifest in various ways, but usually they get stuck in pre-fork.

cc @gnossen

PR #31036, opened 2022-09-17
[tls] Correctly handle unary constructor 

tamird:
@ctiller an alternative to #31030.


ctiller:
We've deleted the relevant code, so far sailing glorious hack.

PR #31041, opened 2022-09-17
Watcher wrapper exec

veblush:

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
@@ -471,6 +471,7 @@ WeightedTargetLb::WeightedChild::DelayedRemovalTimer::DelayedRemovalTimer(
       kChildRetentionInterval, [self = Ref()]() mutable {
         self->weighted_child_->weighted_target_policy_->work_serializer()->Run(
             [self = std::move(self)] {
+              ApplicationExecCtx app_exec_ctx;
               ExecCtx exec_ctx;
               self->OnTimerLocked();
             },


ctiller:
Dropped in ApplicationExecCtxs where they need to be too (for completeness)

src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -554,6 +554,8 @@ class ClientChannel::SubchannelWrapper : public SubchannelInterface {
       parent->chand_->work_serializer_->Run(
           [parent]() ABSL_EXCLUSIVE_LOCKS_REQUIRED(
               *parent_->chand_->work_serializer_) {
+            ApplicationExecCtx app_exec_ctx;
+            ExecCtx exec_ctx;
```
markdroth, line 558: I don't think this is the right fix.  We should never have to declare an `ExecCtx` here, because this lambda should never be run in a thread that doesn't already have an `ExecCtx`.  I don't think we want to require that every `WorkSerializer` closure has to declare its own `ExecCtx`, because (a) that shouldn't be necessary in the first place, and (b) it would result in a lot of nested `ExecCtx` instances.

I assume the core problem here is that this lambda is actually winding up being run in the EE timer callback function, where we call `WorkSerializer::Run()` to hop back into the `WorkSerializer`, and that happens to wind up running this closure from the `WorkSerializer` queue as well.  If so, I think the right fix is for the EE timer callback to declare the `ExecCtx` *before* calling `WorkSerializer::Run()` rather than inside the lambda passed to `WorkSerializer::Run()`, so that the `ExecCtx` is present for any closures that may happen to be invoked at that point.

Let me send a separate PR to fix this.

markdroth (changes requested):

src/core/ext/filters/client_channel/client_channel.cc
markdroth: Just sent #31047 to fix this.

ctiller:
Hey yup probably right... I missed this was the work serializer, thought I was adding it to the Run after.

veblush:
Closed in favor of https://github.com/grpc/grpc/pull/31047

PR #31046, opened 2022-09-19
Move objc TvTests to bazel (and make it "build-only")

jtattermusch:
Move objC TvTests to bazel. It needs to be done as "build-only" test, since bazel currently doesn't support running `tvos_unit_test` targets (so instead we introduce a test that "tests" that TvTests target was built successfully).

### DIFF
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -248,7 +248,8 @@ macos_unit_test(
     ],
 )
 
-# bazel run tvos_unit_test is not yet supported by xctestrunner
+# Note that bazel currently doesn't support running tvos_unit_test
+# See https://github.com/bazelbuild/rules_apple/blob/8d841342c238457896cd7596cc29b2d06c9a75f0/apple/testing/default_runner/tvos_test_runner.template.sh
 tvos_unit_test(
     name = "TvTests",
     minimum_os_version = "10.0",
@@ -261,6 +262,20 @@ tvos_unit_test(
         ":NSErrorUnitTests-lib",
         ":RxLibraryUnitTests-lib",
     ],
+)
+
+# TvTests is not runnable with bazel, so the best we can do is to test
+# that the testsuite builds correctly. This is achieved by having
+# a dummy sh_test depend on :TvTests.
+grpc_sh_test(
+    name = "tvtests_build_test",
+    srcs = [
+        "tvtests_build_test.sh"
+    ],
+    data = [
+        ":TvTests",
+    ],
+    uses_polling = False,
 )
 
 grpc_sh_test(
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -270,7 +270,7 @@ tvos_unit_test(
 grpc_sh_test(
     name = "tvtests_build_test",
     srcs = [
-        "tvtests_build_test.sh"
+        "tvtests_build_test.sh",
     ],
     data = [
         ":TvTests",
--- a/src/objective-c/tests/tvtests_build_test.sh
+++ b/src/objective-c/tests/tvtests_build_test.sh
@@ -1,4 +1,17 @@
 #!/bin/bash
+# Copyright 2022 The gRPC Authors
+#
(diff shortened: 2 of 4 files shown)


jtattermusch:
not sure what the issue with TvTests is, the logs aren't very telling:
https://source.cloud.google.com/results/invocations/584088f7-de0b-4179-970a-c7b33e338d0e/targets/%2F%2Fsrc%2Fobjective-c%2Ftests:TvTests/log

jtattermusch:
@dennycd  looks like the documentation says: https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-tvos.md#tvos_unit_test "Note: tvOS unit tests are not currently supported in the default test runner." - any idea how to work around this and still run tvos tests under bazel?

dennycd:
> @dennycd looks like the documentation says: https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-tvos.md#tvos_unit_test "Note: tvOS unit tests are not currently supported in the default test runner." - any idea how to work around this and still run tvos tests under bazel?

Let me check this locally. The bazel notes is referring to using test_runner context. But i think we are supplying our own test host instead (https://github.com/grpc/grpc/blob/343cb0ecb97c09daf18d0cf873fd239ddf4c36f7/src/objective-c/tests/BUILD#L226) 

Also see their test script which actually tested both scenarios 

- https://github.com/bazelbuild/rules_apple/blob/master/test/tvos_unit_test_test.sh

dennycd:
https://github.com/grpc/grpc-ios/issues/76

jtattermusch:
@dennycd  @HannahShiSFB  I would be grafeful if one of you contributed the finishing touches so that the TvTests pass (I'm not familiar with the details of iOS/TvOS testing so it's hard for me figure out what exactly is needed here).

HannahShiSFB:
TvOS test runner is not implemented in bazel

https://github.com/bazelbuild/rules_apple/blob/master/apple/testing/default_runner/tvos_test_runner.template.sh


WatchOS runner as well.

jtattermusch:
> TvOS test runner is not implemented in bazel
> 
> https://github.com/bazelbuild/rules_apple/blob/master/apple/testing/default_runner/tvos_test_runner.template.sh
> 
> WatchOS runner as well.

Alright, that's unfortunate.


@dennycd  What should be our next steps then? Currently the TvOS test job is terribly flaky on the CI, and the errors it produces are hard understand / debug by humans since the jobs produces huge amount not very comprehensible logs, so IMHO the TvOS test suite is basically useless the way it is.

- note that in bazel we could use `tvos_build_test` which doesn't really run the tests, but at least tests that they build correctly.
- is running TvOS tests continuously on the CI a requirement?

dennycd:
Thanks Jan. We may want to keep tvOS & watchOS test on cocoapod until we can find a viable alternative in bazel. For now those tests are relatively low value since the run the same set of test suite as our iOS/MacOS test, but w/ a different build context.  lgtm to reduce them to build test only or avoid running continuously on CI.

jtattermusch:
@dennycd  PTAL. I also updated the PR description.

src/objective-c/tests/BUILD
```diff
@@ -263,6 +264,20 @@ tvos_unit_test(
     ],
 )
 
+# TvTests is not runnable with bazel, so the best we can do is to test
+# that the testsuite builds correctly. This is achieved by having
+# a dummy sh_test depend on :TvTests.
+grpc_sh_test(
```
dennycd, line 270: reason why we don't just use [tvos_build_test](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-tvos.md#tvos_build_test) for this ?

dennycd (approved):
rest LGTM

src/objective-c/tests/BUILD
jtattermusch: That's what I tried first and I wasn't able to get it working. It can be done as a followup if you know how.

jtattermusch:
I will merge once the repo is open again.

PR #31047, opened 2022-09-19
EventEngine: fix callers of Run() and RunAfter() to create ExecCtx

markdroth:
See discussion in #31041.

Given the number of places we missed this, I think we may want to consider adding some sort of test to ensure that all callers of `Run()` or `RunAt()` will create their own `ExecCtx`.

CC @ctiller @veblush

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
@@ -219,7 +219,6 @@ class GrpcLb : public LoadBalancingPolicy {
     void SendClientLoadReportLocked();
 
     // EventEngine callbacks
-    void MaybeSendClientLoadReport();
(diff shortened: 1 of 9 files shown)


drfloob:
Interesting. I was hoping that any missing, _necessary_ exec_ctx instances would be ... less subtle than this. We had discussed trying to avoid a blanket policy to add them to every callback. I'm not sure they're needed in all of these cases, but we aren't paying a huge price to add them, so I'm ok with it.

drfloob (approved):

markdroth:
I don't think we can assume that the `ExecCtx` is not needed in any case that calls `EventEngine::Run()` or `EventEngine::RunAt()`.  There are too many subtle ways that callback ordering can cause bugs.

I don't think we can start removing `ExecCtx` instances until we remove all the code that actually uses the `ExecCtx`.

ctiller (approved):

PR #31051, opened 2022-09-19
Fix ruby windows ucrt build

apolcyn:
We were previously trying to define this macro through an env var, which actually only affects the `make` invocation that [extconf.rb](https://github.com/grpc/grpc/blob/27e5b4e1877840b58c92d5d853c20ab713d19187/src/ruby/ext/grpc/extconf.rb#L108) uses to build the grpc core library on non-windows platforms.

Note that all of the other `CPPFLAGS` set through env vars are targeting the grpc core library, rather than the ruby C-extension code, so it worked. But this new macro for windows ucrt *is* targeting the ruby C-extension code, so it looks like it needs to be set via $CFLAGS.

Should fix https://github.com/grpc/grpc/issues/30933

cc @johnnyshields


veblush (approved):

apolcyn:
artifact macos failure unrelated

PR #31065, opened 2022-09-20
Make objc kokoro jobs compatible with macos monterey image

jtattermusch:
enable switching them to monterey image, which should in turn solve the cfstream test flakiness issue.

### DIFF
--- a/tools/internal_ci/helper_scripts/prepare_build_macos_rc
+++ b/tools/internal_ci/helper_scripts/prepare_build_macos_rc
@@ -85,7 +85,7 @@ then
   echo "Setting default ruby version."
   rvm use 2.7.0 --default
   echo "Installing cocoapods."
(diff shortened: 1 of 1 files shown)


tools/internal_ci/helper_scripts/prepare_build_macos_rc
```diff
@@ -85,7 +85,7 @@ then
   echo "Setting default ruby version."
   rvm use 2.7.0 --default
   echo "Installing cocoapods."
-  time gem install cocoapods --version 1.3.1 --no-document
+  time gem install cocoapods --version 1.3.1 --no-document --user-install
```
jtattermusch, line 88: as a followup we should check whether this step is required for ruby deps at all.

dennycd (approved):

PR #31074, opened 2022-09-20
Posix event engine tcp client

Vignesh2208:
Implements the Posix event engine client Connect and CancelConnect methods
* Includes a test for connection timeouts and cancellation
* Tests successful connect and subsequent exchange of data using the event engine client test suite.


src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -46,7 +58,221 @@ namespace grpc_event_engine {
 namespace experimental {
 
 #ifdef GRPC_POSIX_SOCKET_TCP
+using grpc_event_engine::posix_engine::EventHandle;
```
drfloob, line 61: Please see https://github.com/grpc/grpc/pull/31413

test/core/event_engine/posix/BUILD
```diff
@@ -191,3 +194,26 @@ grpc_cc_test(
...
+grpc_cc_test(
+    name = "posix_event_engine_connect_test",
+    srcs = ["posix_event_engine_connect_test.cc"],
+    external_deps = ["gtest"],
+    language = "C++",
+    tags = [
+        "no_windows",
+    ],
+    uses_event_engine = True,
+    uses_polling = True,
```
drfloob, line 207: Does this test need to be replicated for every poller using the `@poller=X` test duplication mechanism?

test/core/event_engine/test_suite/BUILD
```diff
@@ -80,8 +80,11 @@ grpc_cc_test(
     tags = [
         "no_windows",
     ],
-    uses_polling = False,
+    uses_event_engine = True,
+    uses_polling = True,
```
drfloob, line 84: same question regarding poller duplication.

test/core/event_engine/test_suite/event_engine_test.cc
```diff
@@ -30,6 +30,6 @@ void SetEventEngineFactories(
     std::function<
         std::unique_ptr<grpc_event_engine::experimental::EventEngine>()>
         oracle_ee_factory) {
-  testing::AddGlobalTestEnvironment(
-      new EventEngineTestEnvironment(factory, oracle_ee_factory));
+  testing::AddGlobalTestEnvironment(new EventEngineTestEnvironment(
+      std::move(factory), std::move(oracle_ee_factory)));
```
drfloob, line 34: Let's make these arguments AnyInvocables, so the move is enforced.

test/core/event_engine/test_suite/posix_event_engine_test.cc
```diff
@@ -19,18 +19,26 @@
...
+#include "src/core/lib/experiments/experiments.h"
 #include "test/core/event_engine/test_suite/event_engine_test.h"
+#include "test/core/event_engine/test_suite/oracle_event_engine_posix.h"
 #include "test/core/util/test_config.h"
 
 int main(int argc, char** argv) {
   testing::InitGoogleTest(&argc, argv);
   grpc::testing::TestEnvironment env(&argc, argv);
+  if (!grpc_core::IsPosixEventEngineEnablePollingEnabled()) {
+    return 0;
```
drfloob, line 31: This will skip all timer tests as well if EventEngine polling is not enabled for the gRPC binary. I don't think we want that, the tests should be exercised regardless.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -0,0 +1,240 @@
...
+  WaitForSingleOwner(std::move(posix_ee));
+}
+
+}  // namespace posix_engine
+}  // namespace grpc_event_engine
+
+int main(int argc, char** argv) {
+  grpc::testing::TestEnvironment env(&argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  if (!grpc_core::IsPosixEventEngineEnablePollingEnabled()) {
```
drfloob, line 233: I don't think these _tests_ should be skipped. The experiment determines whether or not to use the EventEngine in production code ... the engine tests themselves should always be enabled. If some tests need to be skipped because the implementation is not ready to be tested, that's an important distinction to make.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -0,0 +1,240 @@
...
+  int client_socket;
+  int one = 1;
+  int flags;
+  std::vector<int> ret_sockets;
+  // Creating a new socket file descriptor.
+  if ((server_socket = socket(AF_INET6, SOCK_STREAM, 0)) <= 0) {
+    return absl::UnknownError(
+        absl::StrCat("Error creating socket: ", std::strerror(errno)));
+  }
+  // MacOS biulds fail if SO_REUSEADDR and SO_REUSEPORT are set in the same
```
drfloob, line 81: sp: builds

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -0,0 +1,240 @@
...
+  if (setsockopt(server_socket, SOL_SOCKET, SO_REUSEADDR, &opt, sizeof(opt))) {
+    return absl::UnknownError(
+        absl::StrCat("Error setsockopt(SO_REUSEADDR): ", std::strerror(errno)));
+  }
+  if (setsockopt(server_socket, SOL_SOCKET, SO_REUSEPORT, &opt, sizeof(opt))) {
+    return absl::UnknownError(
+        absl::StrCat("Error setsockopt(SO_REUSEPORT): ", std::strerror(errno)));
+  }
+
+  // Forcefully bind the new socket.
```
drfloob, line 92: Out of curiosity, why is this considered forecful?

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -0,0 +1,240 @@
...
+TEST(PosixEventEngineTest, IndefiniteConnectTimeoutOrRstTest) {
+  std::string target_addr = absl::StrCat(
+      "ipv6:[::1]:", std::to_string(grpc_pick_unused_port_or_die()));
+  auto resolved_addr = URIToResolvedAddress(target_addr);
+  std::shared_ptr<EventEngine> posix_ee = std::make_shared<PosixEventEngine>();
+  std::string resolved_addr_str =
+      SockaddrToString(&resolved_addr, true).value();
+  auto sockets = CreateConnectedSockets(resolved_addr);
+  grpc_core::Notification signal;
+  ASSERT_TRUE(sockets.ok());
```
drfloob, line 158: The returned status is never used, never shown. I believe any errors will show up as `Expected: true. Actual: false`. Consider instead doing the logging and asserts in the CreateConnectedSockets method itself.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -0,0 +1,240 @@
...
+  auto quota = grpc_core::ResourceQuota::Default();
+  args = args.Set(GRPC_ARG_RESOURCE_QUOTA, quota);
+  ChannelArgsEndpointConfig config(args);
+  auto memory_quota = absl::make_unique<grpc_core::MemoryQuota>("bar");
+  posix_ee->Connect(
+      [&signal, &resolved_addr_str](
+          absl::StatusOr<std::unique_ptr<EventEngine::Endpoint>> status) {
+        ASSERT_FALSE(status.ok());
+        absl::Status deadline_exceeded_expected_status =
+            absl::CancelledError(absl::StrCat(
```
drfloob, line 169: nit: Checking error strings can be error-prone and brittle, as they are not part of the contract. Consider if you added some data to that error string in the EventEngine implementation ... should this test be considered broken if the strings don't match exactly? How many tests should need to change if the error strings change? It's a subtle thing, but can be annoying when you test internal details rather than the API contract itself.

You could instead just assert the error _type_, which should be an invariant in the API contract.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -0,0 +1,240 @@
...
+                "Failed to connect to remote host: ", resolved_addr_str,
+                " with error: ",
+                absl::DeadlineExceededError("connect() timed out").ToString()));
+        absl::Status conn_reset_expected_status =
+            absl::CancelledError(absl::StrCat(
+                "Failed to connect to remote host: ", resolved_addr_str,
+                " with error: ",
+                absl::InternalError(absl::StrCat("getsockopt(SO_ERROR): ",
```
drfloob, line 177: InternalError usually implies an implementation problem, which ... I don't think this is. https://github.com/abseil/abseil-cpp/blob/85cc4947dbec515868e47a7e4e2b604b21e333cb/absl/status/status.h#L231-L236. That distinction may become important, so we should probably fix the implementation to ensure it returns something more appropriate, maybe FailedPrecondition.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -0,0 +1,240 @@
...
+  ChannelArgsEndpointConfig config(args);
+  auto memory_quota = absl::make_unique<grpc_core::MemoryQuota>("bar");
+  auto connection_handle = posix_ee->Connect(
+      [](absl::StatusOr<std::unique_ptr<EventEngine::Endpoint>> /*status*/) {
+        FAIL() << "The on_connect callback should not have run since the "
+                  "connection attempt was cancelled.";
+      },
+      URIToResolvedAddress(target_addr), config,
+      memory_quota->CreateMemoryAllocator("conn-2"), 3s);
+  ASSERT_TRUE(posix_ee->CancelConnect(connection_handle));
```
drfloob, line 220: What are the chances that this races with a failure before the timeout expires?

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -46,7 +58,221 @@ namespace grpc_event_engine {
...
+  EventHandle* fd;
+  absl::StatusOr<std::unique_ptr<EventEngine::Endpoint>> ep;
+
+  mu_.Lock();
+  GPR_ASSERT(fd_ != nullptr);
+  fd = std::exchange(fd_, nullptr);
+  bool connect_cancelled = connect_cancelled_;
+  mu_.Unlock();
+
+  if (engine_->Cancel(alarm_handle_)) {
```
drfloob, line 110: Is there a race here between OnWritable and OnTimeoutExpired? If OnWritable is started in one thread, but OnTimeoutExpires enters its critical section first, then I think AsyncConnect will be deleted while OnWritable is running.

include/grpc/event_engine/event_engine.h
```diff
@@ -17,6 +17,7 @@
 #include <grpc/support/port_platform.h>
 
 #include <functional>
+#include <memory>
```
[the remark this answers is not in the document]
Vignesh2208: I dont remember. I removed it now.

test/core/event_engine/posix/BUILD
Vignesh2208: Yes the event engine can instantiate both types of pollers. So I think its good to test connection cancellation/timeouts with both pollers

test/core/event_engine/posix/posix_event_engine_connect_test.cc
Vignesh2208: Per offline discussion I think we can keep this temporarily. The experiment flag is enabled in asan/tsan and debug builds. Its only disabled in opt builds. If the flag is not set, event engine doesn't create a poller and thus these tests would crash.

test/core/event_engine/test_suite/BUILD
Vignesh2208: Yes I wanted to test the posix event engine client and server tests against both types of pollers epoll and poll.

test/core/event_engine/test_suite/event_engine_test.cc
Vignesh2208: Done

test/core/event_engine/test_suite/posix_event_engine_test.cc
Vignesh2208: Same answer as before.

src/core/lib/event_engine/posix_engine/posix_engine.cc
Vignesh2208: I dont think so. If engine_->Cancel succeeds, then the OnTimeoutExpires callback shouldn't run. So it wont decrement the refcount. Instead we just increment the consumed_refs variable within the if block. If OnTimeoutExpires is running while OnWritable is also executing, the engine_->Cancel should have returned false so there shouldn't be an extra increment of the ref count. The rest of  OnWritable would also be executing under the locked mutex mu_

test/core/event_engine/posix/posix_event_engine_connect_test.cc
Vignesh2208: Done.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
Vignesh2208: Replaced it with FailedPreconditionError

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -46,7 +58,221 @@ namespace grpc_event_engine {
...
+  });
+  fd_->NotifyOnWrite(on_writable_);
+}
+
+AsyncConnect ::~AsyncConnect() { delete on_writable_; }
+
+void AsyncConnect::OnTimeoutExpired(absl::Status status) {
+  bool done = false;
+  {
+    absl::MutexLock lock(&mu_);
```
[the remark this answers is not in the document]
Vignesh2208: Replaced it in the pollers as well.

src/core/lib/event_engine/posix_engine/posix_engine.cc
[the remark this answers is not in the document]
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/posix_engine.h
```diff
@@ -38,10 +43,57 @@
...
+  std::shared_ptr<EventEngine> engine_;
+  ThreadPool* executor_;
+  EventEngine::TaskHandle alarm_handle_;
+  int refs_{2};
+  grpc_event_engine::posix_engine::EventHandle* fd_;
+  MemoryAllocator allocator_;
+  grpc_event_engine::posix_engine::PosixTcpOptions options_;
+  std::string resolved_addr_str_;
+  int64_t connection_handle_ = 0;
+  bool connect_cancelled_ = false;
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
Vignesh2208: Fixed.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
Vignesh2208: Fixed.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
Vignesh2208: Not apt wording. I changed the comment.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
Vignesh2208: God point. It can fail if the kernel sends a RST packet to the connection attempt. But I noticed that happens immediately. So I made a change to invoke CancelConnect only if the returned connection_handle is > 0 (which would be true only if the connection didn't fail immediately).

Vignesh2208:
Fixed your comments. Also found a minor bug in the poller so updated the poller api contract in 7133e494404d111f7c7d328df07c6e95a90e4ac8

src/core/lib/event_engine/posix_engine/posix_engine.cc
drfloob: I see, there are 2 `refs_` to start, so it won't be deleted in the cancel callback.

drfloob (approved):

PR #31077, opened 2022-09-21
Fix clang compiler errors for i386-windows-msvc

mattvera:
This seems to be a known issue with lambdas and clang when targeting i386-windows-msvc, see
https://github.com/llvm/llvm-project/issues/28673.

### DIFF
--- a/src/core/ext/filters/channel_idle/channel_idle_filter.cc
+++ b/src/core/ext/filters/channel_idle/channel_idle_filter.cc
@@ -141,23 +141,24 @@ void MaxAgeFilter::PostInit() {
     RefCountedPtr<grpc_channel_stack> channel_stack;
     MaxAgeFilter* filter;
     grpc_closure closure;
-  };
-  auto run_startup = [](void* p, grpc_error_handle) {
-    auto* startup = static_cast<StartupClosure*>(p);
-    // Trigger idle timer
-    startup->filter->IncreaseCallCount();
-    startup->filter->DecreaseCallCount();
-    grpc_transport_op* op = grpc_make_transport_op(nullptr);
-    op->start_connectivity_watch.reset(
-        new ConnectivityWatcher(startup->filter));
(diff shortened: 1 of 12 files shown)


markdroth:
You'll need to sign the CLA before we can look at this.

mattvera:
Thanks Mark. I've requested "Authorization to Contribute" from the CLA Managers for my organization ("Google LLC").

src/core/ext/filters/channel_idle/channel_idle_filter.cc
```diff
@@ -173,21 +174,24 @@ void MaxAgeFilter::PostInit() {
...
-                auto* channel_stack = static_cast<grpc_channel_stack*>(arg);
-                grpc_transport_op* op = grpc_make_transport_op(nullptr);
-                op->goaway_error = grpc_error_set_int(
-                    GRPC_ERROR_CREATE_FROM_STATIC_STRING("max_age"),
-                    GRPC_ERROR_INT_HTTP2_ERROR, GRPC_HTTP2_NO_ERROR);
-                grpc_channel_element* elem =
-                    grpc_channel_stack_element(channel_stack, 0);
-                elem->filter->start_transport_op(elem, op);
-                GRPC_CHANNEL_STACK_UNREF(channel_stack, "max_age send_goaway");
+              struct Closure {
```
mattvera, line 177: This is a poor choice of name (here and elsewhere). There's no closure here. I'll replace this with something else in a subsequent commit to this pull request. Open to suggestions.

ctiller:
I don't think this is a good trade-off in terms of complexity/maintainability for a platform we don't support.

markdroth:
Yeah, I agree with @ctiller here.  The changes in this PR make our code much less readable, and even if we accepted this change, there's basically no way we can prevent regressing here, since we don't have any infrastructure to test things on this platform.

I don't think we're willing to accept this patch.  You're welcome to maintain your own fork if you want, but I don't think we want to upstream these changes.

PR #31086, opened 2022-09-21
Adjust slice buffer build target

Vignesh2208:
This change enables some internal cleanup and avoids having to depend on whole of grpc to get definitions of functions defined in src/core/lib/slice/slice_buffer_api.cc


ctiller:
This adds slice_buffer_api.cc to two different build targets, one of which depends on the other: it's very likely this causes us some trouble down the road.

Let's get #30953 in instead so we can eliminate this file?

PR #31093, opened 2022-09-22
Observability: Experimental arg to disable client side tracing

yashykt:

### DIFF
--- a/include/grpc/impl/codegen/grpc_types.h
+++ b/include/grpc/impl/codegen/grpc_types.h
@@ -461,6 +461,13 @@ typedef struct {
  * channel arg. Int valued, milliseconds. Defaults to 10 minutes.*/
 #define GRPC_ARG_SERVER_CONFIG_CHANGE_DRAIN_GRACE_TIME_MS \
   "grpc.experimental.server_config_change_drain_grace_time_ms"
+/** EXPERIMENTAL. If non-zero, disables observability tracing and observability
+ * logging (not yet implemented) on the client channel. Note that this does not
+ * impact metrics/stats collection. This channel arg is intended as a way to
+ * avoid cyclic execution of observability logging and trace especially when the
+ * sampling rate of RPCs is very high which would generate a lot of data.
+ */
+#define GRPC_ARG_DISABLE_OBSERVABILITY "grpc.experimental.disable_observability"
 /** \} */
 
 /** Result of a grpc call. If the caller satisfies the prerequisites of a
--- a/include/grpc/impl/codegen/grpc_types.h
+++ b/include/grpc/impl/codegen/grpc_types.h
@@ -461,13 +461,6 @@ typedef struct {
  * channel arg. Int valued, milliseconds. Defaults to 10 minutes.*/
 #define GRPC_ARG_SERVER_CONFIG_CHANGE_DRAIN_GRACE_TIME_MS \
   "grpc.experimental.server_config_change_drain_grace_time_ms"
-/** EXPERIMENTAL. If non-zero, disables observability tracing and observability
- * logging (not yet implemented) on the client channel. Note that this does not
- * impact metrics/stats collection. This channel arg is intended as a way to
- * avoid cyclic execution of observability logging and trace especially when the
- * sampling rate of RPCs is very high which would generate a lot of data.
- */
-#define GRPC_ARG_DISABLE_OBSERVABILITY "grpc.experimental.disable_observability"
 /** \} */
 
 /** Result of a grpc call. If the caller satisfies the prerequisites of a
--- a/src/core/lib/channel/call_tracer.h
+++ b/src/core/lib/channel/call_tracer.h
@@ -33,18 +33,6 @@
 #include "src/core/lib/transport/metadata_batch.h"
 #include "src/core/lib/transport/transport.h"
 
-// TODO(yashykt): This might not be the right place for this channel arg, but we
-// don't have a better place for this right now.
-
-// EXPERIMENTAL. If zero, disables observability tracing and observability
-// logging (not yet implemented) on the client channel, defaults to true. Note
-// that this does not impact metrics/stats collection. This channel arg is
-// intended as a way to avoid cyclic execution of observability logging and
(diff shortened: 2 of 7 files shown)


src/cpp/ext/filters/census/client_filter.h
```diff
@@ -29,19 +29,28 @@
...
+   private:
+    OpenCensusCallTracer* tracer_ = nullptr;
+  };
+
+  grpc_error_handle Init(grpc_channel_element* elem,
+                         grpc_channel_element_args* args) override;
 
  private:
-  OpenCensusCallTracer* tracer_ = nullptr;
+  bool is_tracing_disabled_ = false;
```
markdroth, line 53: Double-negatives are confusing.  Let's make this:

```
bool tracing_enabled_ = true;
```

include/grpc/impl/codegen/grpc_types.h
```diff
@@ -461,6 +461,13 @@ typedef struct {
  * channel arg. Int valued, milliseconds. Defaults to 10 minutes.*/
 #define GRPC_ARG_SERVER_CONFIG_CHANGE_DRAIN_GRACE_TIME_MS \
   "grpc.experimental.server_config_change_drain_grace_time_ms"
+/** EXPERIMENTAL. If non-zero, disables observability tracing and observability
+ * logging (not yet implemented) on the client channel. Note that this does not
+ * impact metrics/stats collection. This channel arg is intended as a way to
+ * avoid cyclic execution of observability logging and trace especially when the
+ * sampling rate of RPCs is very high which would generate a lot of data.
+ */
+#define GRPC_ARG_DISABLE_OBSERVABILITY "grpc.experimental.disable_observability"
```
markdroth, line 470: Double-negatives are confusing.  Let's make this:

```
#define GRPC_ARG_OBSERVABILITY_ENABLE "grpc.experimental.observability_enable"
```

The default value can be true.

markdroth, line 470: I'd rather this not be defined in a public header.  Let's keep it internal and just have the o11y code use it from the internal header.

src/cpp/ext/filters/census/client_filter.cc
```diff
@@ -222,15 +238,26 @@ void OpenCensusCallTracer::OpenCensusCallAttemptTracer::RecordEnd(
   }
 }
 
+CensusContext OpenCensusCallTracer::OpenCensusCallAttemptTracer::
+    CreateCensusContextForCallAttempt(OpenCensusCallTracer* call_tracer) {
```
markdroth, line 242: Why make this a method on `OpenCensusCallAttemptTracer` when it's actually only accessing data on `OpenCensusCallTracer`?  Seems like it could just be a method on `OpenCensusCallTracer`, and the `OpenCensusCallAttemptTracer` could just change `CreateCensusContextForCallAttempt(parent_)` to `parent_->CreateCensusContextForCallAttempt()`.

markdroth (commented):
Overall, looks good -- just a few minor comments.

include/grpc/impl/codegen/grpc_types.h
yashykt: Done

yashykt: Done

src/cpp/ext/filters/census/client_filter.h
yashykt: Done

src/cpp/ext/filters/census/client_filter.cc
yashykt: Done

test/cpp/ext/filters/census/stats_plugin_end2end_test.cc
```diff
@@ -730,6 +731,55 @@ TEST_F(StatsPluginEnd2EndTest, TestAllSpansAreExported) {
...
+        ::opencensus::trace::Span::StartSpan("sampling", nullptr, options);
+    grpc::CensusContext app_census_context("root", &sampling_span,
+                                           ::opencensus::tags::TagMap{});
+    context.set_census_context(
+        reinterpret_cast<census_context*>(&app_census_context));
+    traces_recorder_->StartRecording();
+    grpc::Status status = stub_->Echo(&context, request, &response);
+    EXPECT_TRUE(status.ok());
+  }
+  absl::SleepFor(absl::Milliseconds(500 * grpc_test_slowdown_factor()));
```
veblush, line 761: Is there any way to do the same without sleep? This could be another source of making this test flaky.

yashykt: We've been using this 500ms sleep in this test for quite some time and I don't think this test has come up as flaky. 

That being said, I do want to remove the sleep from these tests. One way to do this would be to simply loop with a deadline till we get what we want, but if I'm doing this I'd like to do it for all the tests, and I would prefer not doing it in this PR

src/core/lib/channel/call_tracer.h
```diff
@@ -33,6 +33,18 @@
 #include "src/core/lib/transport/metadata_batch.h"
 #include "src/core/lib/transport/transport.h"
 
+// TODO(yashykt): This might not be the right place for this channel arg, but we
```
markdroth, line 36: I think this can live in src/cpp/ext/filters/census/open_census_call_tracer.h, since it's really an OpenCensus-specific channel arg.

markdroth (approved):
Just one comment about moving something to another file.  Otherwise, looks great!

src/core/lib/channel/call_tracer.h
yashykt: I think in the future when we have more of a common observability API, this would arg would eventually go there, but right now, I don't have strong opinions one way or the other. Moved to open_census_call_tracer.h

test/cpp/ext/filters/census/stats_plugin_end2end_test.cc
veblush: Assuming that this will be covered by later PR, I approve this. Thanks!

veblush (approved):

Issue #25082, opened 2021-01-02
grpcio fails to install on Apple Silicon


yoav-orca:
### What version of gRPC and what language are you using?
grpcio-1.34.0

### What operating system (Linux, Windows,...) and version?
macOS big Sur 11.0.1, running on M1 Apple Silicon

### What runtime / compiler are you using (e.g. python version or version of gcc)
```
Python 3.8.7 (default, Dec 30 2020, 02:09:32)
[Clang 12.0.0 (clang-1200.0.32.28)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
```

```
$ clang --version
Apple clang version 12.0.0 (clang-1200.0.32.27)
Target: arm64-apple-darwin20.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.

`pip install grpcio==1.34.0`

### What did you expect to see?
That grpcio will be successfully installed

### What did you see instead?
```
Collecting grpcio
  Using cached grpcio-1.34.0.tar.gz (21.0 MB)
    ERROR: Command errored out with exit status 1:
     command: /Users/yoav/Library/Caches/pypoetry/virtualenvs/orca-g6p4a6cZ-py3.8/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/jj/r3j0p91s2qgdpcpf5ptqr9qm0000gn/T/pip-install-n05vw0u0/grpcio_103d947f24954cde8b8a019b4c9f88fe/setup.py'"'"'; __file__='"'"'/private/var/folders/jj/r3j0p91s2qgdpcpf5ptqr9qm0000gn/T/pip-install-n05vw0u0/grpcio_103d947f24954cde8b8a019b4c9f88fe/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/jj/r3j0p91s2qgdpcpf5ptqr9qm0000gn/T/pip-pip-egg-info-z41yqiw6
[...]

PR #31096, opened 2022-09-22
Build with System OpenSSL on Mac OS arm64

amchii:
Boringssl currently does not support macOS arm64, so installation on these platform with default config will undoubtedly fail. That means the env var 'GRPC_PYTHON_BUILD_SYSTEM_OPENSSL' must be set to true, but `GRPC_PYTHON_BUILD_SYSTEM_ZLIB=1` mentioned in this issue https://github.com/grpc/grpc/issues/25082, I think it ain't necessary, because:

1. zlib support build on macOS arm64, see https://github.com/madler/zlib/issues/559
2. I successfully built gRPC after uninstall system zlib (and not set `GRPC_PYTHON_BUILD_SYSTEM_ZLIB=1`

I don't know when will grpc provide python wheel for macOS arm64, but at least I think installing grpcio from source by default should be not failed.

@gnossen

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -133,17 +133,20 @@ BUILD_OVERRIDE_BORING_SSL_ASM_PLATFORM = os.environ.get(
 # present, then it will still attempt to use Cython.
 BUILD_WITH_CYTHON = _env_bool_value('GRPC_PYTHON_BUILD_WITH_CYTHON', 'False')
 
+# Currently, boringssl does not support macOS arm64, so we try to use the system
+# installation of openssl to build gRPC locally by default.
+if "darwin" in sys.platform and "arm64" == platform.machine().lower():
+    sys.stderr.write("Boringssl currently does not support macOS arm64, so we'll try to use the system installation "
+                     "of 'openssl' to build by default, make sure you have 'openssl' installed in this case\n")
+    BUILD_WITH_SYSTEM_OPENSSL_DEFAULT_ENV = "True"
+else:
+    BUILD_WITH_SYSTEM_OPENSSL_DEFAULT_ENV = "False"
+
 # Export this variable to use the system installation of openssl. You need to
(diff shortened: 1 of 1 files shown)


gnossen (approved):
Thanks for the contribution!

setup.py
```diff
@@ -138,6 +138,12 @@ def _env_bool_value(env_name, default):
 # runtime, the shared library must be installed
 BUILD_WITH_SYSTEM_OPENSSL = _env_bool_value('GRPC_PYTHON_BUILD_SYSTEM_OPENSSL',
                                             'False')
+# Currently, boringssl does not support macOS arm64, so we MUST try to use the system
+# installation of openssl to build gRPC locally.
+if "darwin" in sys.platform and "arm64" == platform.machine().lower():
+    sys.stderr.write("Boringssl currently does not support macOS arm64, so we'll try to use the system "
+                     "installation of 'openssl' to build, make sure you have 'openssl' installed\n")
+    BUILD_WITH_SYSTEM_OPENSSL = True
```
gnossen, line 146: Actually, can we just change the default value here so that it is still possible to set this to `False` in the future?

amchii: Use system installation of openssl by default on macOS arm64: https://github.com/grpc/grpc/pull/31096/commits/69aee68bd68ad3a6c79c438f2d6dfd1542253478.

gnossen (approved):

gnossen:
One minor thing: Can you please run `tools/distrib/yapf_code.sh` to format your code?

amchii:
> One minor thing: Can you please run `tools/distrib/yapf_code.sh` to format your code?

Formatted: https://github.com/grpc/grpc/pull/31096/commits/200fb50e1c46ba9ee501eeba8a6007fcf7ec080d

Issue #30979, opened 2022-09-14
Signficant unnecessary but avoidable allocations when TCP trace enabled at ignored log-level


mgravell:
There are multiple code-paths that follow the pattern:

```
        if (grpc_tcp_trace.enabled()) {
          size_t i;
          for (i = 0; i < tcp->read_slices->count; i++) {
            char* dump = grpc_dump_slice(tcp->read_slices->slices[i],
                                         GPR_DUMP_HEX | GPR_DUMP_ASCII);
            gpr_log(GPR_INFO, ... blah ..., dump);
            gpr_free(dump);
          }
        }
```

where `grpc_dump_slice` builds a nul-terminated string of the fragment; this applies to most network read/write; useful for debugging.

However! If the log-level is low (meaning: less logging - for example ERROR only), but the trace is enabled: this performs a **LOT** of unnecessary CPU and memory overhead, including allocating/releasing the memory for the chunks, and performing hex+ascii dumps - only to have it all ignored by the log-level check.

Proposal: add a log-level test earlier, so that expensive work is not performed for no reason; for example:

```
if (grpc_tcp_trace.enabled() && gpr_should_log(GPR_LOG_SEVERITY_INFO)) // whatever the relevant log-level is
```

or, perhaps more cleanly (but with a new overload):

```
if (grpc_tcp_trace.enabled(GPR_LOG_SEVERITY_INFO)) // whatever the relevant log-level is
```

Candidate scenarios:

- https://github.com/grpc/grpc/blob/c6ebbb5482cf93623a73e54c31277c174c97222f/src/core/lib/iomgr/endpoint_cfstream.cc#L122 (DEBUG)
- https://github.com/grpc/grpc/blob/c6ebbb5482cf93623a73e54c31277c174c97222f/src/core/lib/iomgr/endpoint_cfstream.cc#L229 (DEBUG)
- https://github.com/grpc/grpc/blob/70dd34b4382632b4e668c3d968ca49bb3a67076c/src/core/lib/iomgr/tcp_windows.cc#L364 (INFO)
- https://github.com/grpc/grpc/blob/70dd34b4382632b4e668c3d968ca49bb3a67076c/src/core/lib/iomgr/tcp_windows.cc#L211 (INFO)
- https://github.com/grpc/grpc/blob/32bc9a86562353372968bdf45e13bfeed57b00e8/src/core/lib/security/transport/secure_endpoint.cc#L401 (INFO)
[...]

PR #31098, opened 2022-09-22
Core; fix avoidable work (allocs+CPU) when logs enabled but trace level means work will not be used

mgravell:
core; ref #30979

1. avoid calling `grpc_dump_slice` if the log level is too low (and the result will be ignored)
2. use `GRPC_TRACE_FLAG_ENABLED(x)` over `x.enabled()` in the touched code


mgravell:
See also @tomersh26's [remark here](https://github.com/grpc/grpc/issues/30979#issuecomment-1274562313) about possibly also pushing this to the Grpc.Core nuget package (now in either maintenance mode or non-maintenance mode)

veblush:
Yash, please take a look at this!

yashykt (approved):

yashykt:
Please run `tools/distrib/clang_format_code.sh`

markdroth (approved):

PR #31100, opened 2022-09-22
Revert "Revert "Remove `grpc_codegen`""

ralphchung:
Reverts grpc/grpc#31099

PR #31099 itself does not have an issue. The fix should be done for the importing to Google internally.


ralphchung:
Waiting for resolving merge conflict after a later import.

PR #31102, opened 2022-09-22
Rename RETURN_IF_ERROR to RETURN_IF_NOT_OK

Vignesh2208:


drfloob (approved):

src/core/lib/gprpp/status_helper.h
```diff
@@ -38,7 +38,7 @@ struct google_rpc_Status;
 struct upb_Arena;
 }
 
-#define RETURN_IF_ERROR(expr)           \
```
ctiller, line 41: Let's change this to `GRPC_RETURN_IF_NOT_OK` - we should in general keep things in our own namespace

PR #31107, opened 2022-09-22
Add authorization_provider library to cmake builds

jebr224:
This change adds authorization policy providers functionality to the cmake builds.


jebr224:
Is there an example "Validator: LABEL" I could look at, to understand what "mergeable" is looking for?

jebr224:
I looked over a couple of the test below. Is there a subset of artifacts that used to gate entry. I think some of these failures might not be directly related to my change. If that is not the case let me know what I need to focus on.

Spot check
---------------

Looks like there is a ruby issue on MacOS
```
Fetching gem metadata from https://rubygems.org/.........
public_suffix-5.0.0 requires ruby version >= 2.6, which is incompatible with the
current version, ruby 2.5.0p0
```
Windows shows a twine depends on packaging. 
```
  File "C:\Python310\lib\site-packages\twine\cli.py", line 18, in <module>
    from packaging import requirements
ModuleNotFoundError: No module named 'packaging'
```
C++ and objective C iOS have permissions issues
```
++ gem install cocoapods --version 1.7.2 --no-document
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.6.0 directory.
```

jebr224:
abandon for 
https://github.com/grpc/grpc/pull/31168

Issue #31012, opened 2022-09-16
Move `grpc_insecure_*_credentials_create` declarations to `grpc/grpc.h` 


mehrdada:
Right now the `grpc_insecure_credentials_create` and `grpc_insecure_server_credentials_create` declarations are in `grpc_security.h`.  However, they are critical in doing pretty any work in grpc without utilizing secure credentials.  That would mean the users will have to include both headers even if they are relying only on insecure gRPC.  It would make sense to move the insecure credentials to `grpc.h` (or alternatively abolish grpc_security.h as a separate include?)

drfloob:
> However, they are critical in doing pretty any work in grpc without utilizing secure credentials

None of the examples in [examples/cpp/](https://github.com/grpc/grpc/tree/868b7d82a78333f040e5fad91800c98f858fd6b1/examples/cpp) have to include that header. I'm not sure why you have the impression that it's always required, but it is not.

mehrdada:
I'm sorry, I thought `lang/core` tag makes it clear enough--not sure if this has anything to do with C++ examples.  This is related to C core API for wrapped languages and such. In particular, this example should compile fine without including grpc_security header:

```
#include <grpc/grpc.h>

// Should work fine without this, but does not:
// #include <grpc/grpc_security.h>

int main() {
  grpc_init();
  grpc_server* server = grpc_server_create(0, 0);
  grpc_completion_queue* cq = grpc_completion_queue_create_for_next(0);
  grpc_server_register_completion_queue(server, cq, 0);
  grpc_server_add_http2_port(server, "127.0.0.1:8080", grpc_insecure_server_credentials_create());
  grpc_server_shutdown_and_notify(server, cq, 0);
  grpc_completion_queue_next(cq, gpr_inf_future(GPR_CLOCK_REALTIME), 0);
  grpc_completion_queue_shutdown(cq);
  grpc_server_destroy(server);
  grpc_completion_queue_destroy(cq);
  grpc_shutdown();
}
```

but makes the C compiler warn about undeclared function `grpc_insecure_server_credentials_create`.

drfloob:
[...]

PR #31111, opened 2022-09-23
Move insecure credentials declarations from `grpc_security.h` to `grpc_insecure_credentials.h`

ralphchung:
This PR fixes the issue #31012.


markdroth (approved):

mehrdada (approved):
LGTM--thanks!

drfloob (approved):

ctiller:
I think grpc.h is the wrong place for this: we likely want to move them to grpc_insecure.h or the like so that we can visibility restrict it.

ctiller (changes requested):
I don't think this is the right approach.

markdroth:
I think Craig is right.  Let's move them to grpc_insecure_creds.h.

ralphchung:
Hi all @ctiller @markdroth @mehrdada @drfloob,

I have updated the PR according to the discussion.

ctiller:
I think this looks good, but I also think it's a public API change and so will require a gRFC (see https://github.com/grpc/proposal)

mehrdada (approved):

Issue #31064, opened 2022-09-20
xds-protos: Regenerate with protoc >= 3.19.0


serge30:
Hello,

I'm trying to use `xds-protos` packet with `protobuf==4.21.6` and get Python error on import anything from this package. As I can see from error message, it suggests to regenerate Python code with `protoc >= 3.19.0`

Can we have a new version of `xds-protos` package please?

--
Thanks,
Serhii

### What version of gRPC and what language are you using?

```
xds-protos==0.0.11
grpcio==1.48.1
protobuf==4.21.6
six==1.16.0
```

### What operating system (Linux, Windows,...) and version?

MacOS Monterey 12.6

### What runtime / compiler are you using (e.g. python version or version of gcc)

```
Python 3.8.14 (default, Sep  6 2022, 23:17:06)
[Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
```

### What did you do?

```
(venv) serge@macbook xds_test % pip install xds-protos
Collecting xds-protos
  Using cached xds_protos-0.0.11-py3-none-any.whl (1.8 MB)
Collecting grpcio
  Using cached grpcio-1.48.1-cp38-cp38-macosx_12_0_arm64.whl
Collecting protobuf
  Using cached protobuf-4.21.6-cp37-abi3-macosx_10_9_universal2.whl (484 kB)
Collecting six>=1.5.2
  Using cached six-1.16.0-py2.py3-none-any.whl (11 kB)
Installing collected packages: six, protobuf, grpcio, xds-protos
Successfully installed grpcio-1.48.1 protobuf-4.21.6 six-1.16.0 xds-protos-0.0.11
WARNING: You are using pip version 22.0.4; however, version 22.2.2 is available.
You should consider upgrading via the '/Users/serge/projects/xds_test/venv/bin/python3.8 -m pip install --upgrade pip' command.

(venv) serge@macbook xds_test % python3
Python 3.8.14 (default, Sep  6 2022, 23:17:06)
[Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

>>> import envoy.config.cluster.v3.cluster_pb2
```

### What did you expect to see?

No errors

### What did you see instead?

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #31113, opened 2022-09-23
Update xds-protos package to pull in protobuf 4.X

gnossen:
Fixes https://github.com/grpc/grpc/issues/31064

If someone runs `pip install grpcio-csds` right now, they'll get

```
grpcio-csds==1.49.1
grpcio==1.49.1
xds-protos=0.0.11
protobuf==4.21.6
```

This combination does not work together because `xds-protos` 0.0.11 packages 3.X protobuf generated code.

The obvious fix here is to push an `xds-protos==0.0.12` with 4.X generated protos and with `protobuf~=4.0.0` range. How will this work for people who for whatever reason need to stick with old protos. At the top level, they'll have

```
grpcio-csds
protobuf~=3.0.0
```

This will give them

- `protobuf==3.20.2` (since it's the latest compatible with `protobuf~=3.0.0`)
- `grpcio-csds==1.48.2` (since it's the latest compatible with `protobuf==3.20.2`)
- `grpcio==1.48.2` (since it's the only compatible with `grpcio-csds==1.48.2`)
- `xds-protos==0.0.11` (since it's the latest compatible with `protobuf==3.20.2`)

So this should solve our problem. No backports needed.

Going forward though, we really need to do an atomic release containing both the new csds and the new xds-protos. We should probably just release one per grpcio version.

This did not fail in CI because the `grpcio-csds` tests were not properly set up to run with the `setup.py` test runner. I have a separate PR to correct that issue.

### DIFF
--- a/tools/distrib/python/xds_protos/setup.py
+++ b/tools/distrib/python/xds_protos/setup.py
@@ -32,13 +32,13 @@ CLASSIFIERS = [
     'License :: OSI Approved :: Apache Software License',
 ]
 INSTALL_REQUIRES = [
-    'grpcio',
-    'protobuf',
+    'grpcio>=1.49.0',
+    'protobuf>=4.21.6,<5.0dev',
 ]
 SETUP_REQUIRES = INSTALL_REQUIRES + ['grpcio-tools']
 setuptools.setup(
     name='xds-protos',
-    version='0.0.11',
+    version='0.0.12',
     packages=PACKAGES,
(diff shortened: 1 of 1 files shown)


XuanWang-Amos (approved):
Change LGTM, just one typo in description, the xds-protos version should be 0.0.12 instead of 0.12.0.

gnossen:
Indeed. I thought I had fixed all of those versions. Fixed.

PR #31114, opened 2022-09-23
Update comment on GRPC_ARG_TSI_MAX_FRAME_SIZE

ashithasantosh:
This comment was added before frame size negotiation.

With frame size negotiation we can have a larger value for max frame size. Negotiation takes into consideration old binaries and non-zerocopy protectors like gRPC Go, and makes sure to agree on a maximum frame size value that is acceptable at both endpoints.

### DIFF
--- a/include/grpc/impl/codegen/grpc_types.h
+++ b/include/grpc/impl/codegen/grpc_types.h
@@ -281,11 +281,7 @@ typedef struct {
 #define GRPC_SSL_SESSION_CACHE_ARG "grpc.ssl_session_cache"
 /** If non-zero, it will determine the maximum frame size used by TSI's frame
  *  protector.
- *
- *  NOTE: Be aware that using a large "max_frame_size" is memory inefficient
- *        for non-zerocopy protectors. Also, increasing this value above 1MiB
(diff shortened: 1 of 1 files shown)


ctiller:
Does this imply that ALTS has a frame size negotiation, and is it dynamic or just at handshake time?

ashithasantosh:
Yes ALTS has frame size negotiation. It is only at handshake time.

ctiller:
And just to check, are there any other TSI supported transports which do not support frame size negotiation?

ashithasantosh:
It is supported by LOAS2 (flag controlled) and ALTS for zero copy frame protectors (which as I understand is the default)
https://source.corp.google.com/piper///depot/google3/security/loas/l2/internal/transport_security/loas2_transport_security_common.cc;rcl=462721822;l=459
https://github.com/grpc/grpc/blob/ba8af0157ba9b39b6b584a8b5de3bd38a2c43588/src/core/tsi/alts/handshaker/alts_tsi_handshaker.cc#L175
The other transports like S2A and SSL do not have zero copy protectors, so they do not implement frame size negotiation

ctiller (approved):

soheilhy (approved):

PR #31125, opened 2022-09-26
Fix ruby builds on kokoro macos monterey images

jtattermusch:
Prerequisite for migrating the ruby basictests to the kokoro macos monterey image.


apolcyn (approved):

jtattermusch:
the distrib tests python on windows failure seems unrelated.

PR #31128, opened 2022-09-27
Invoking Poller::Work in posix event engine constructor

Vignesh2208:
This PR creates a separate posix event engine constructor for tests and also instantiates the default poller in the event engine constructor when running in production. The Poller::Work method is invoked recursively and the poller is shutdown in the event engine destructor.

The PR also adds an unused MemoryAllocator argument to grpc_event_engine::posix_engine::CreatePosixEndpoint function.


src/core/lib/event_engine/posix_engine/event_poller_posix_default.cc
```diff
@@ -58,5 +62,11 @@ PosixEventPoller* GetDefaultPoller(Scheduler* scheduler) {
   return poller;
 }
 
+#else  // GRPC_POSIX_SOCKET_TCP
+
+PosixEventPoller* GetDefaultPoller(Scheduler* /*scheduler*/) { return nullptr; }
```
ctiller, line 67: Change name to MakeDefaultPoller, MakeEpoll1Poller, MakePollPoller, etc

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -15,23 +15,89 @@
...
+  // TODO(vigneshbabu): The timeout specified here is arbitrary. For instance,
+  // this can be improved by setting the timeout to the next expiring timer.
+  auto result = poller_->Work(24h, [this]() {
+    ++shutdown_ref_;
+    executor_.Run([this]() { PollerWorkInternal(); });
+  });
+  if (result == Poller::WorkResult::kDeadlineExceeded) {
+    // The event engine is not shutting down but the next asynchronous
+    // PollerWorkInternal did not get scheduled. Schedule it now.
+    ++shutdown_ref_;
```
ctiller, line 73: shutdown_ref_.fetch_add(1)

src/core/lib/event_engine/posix_engine/event_poller_posix_default.cc
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/posix_engine.cc
Vignesh2208: Removed the implementation. Now we dont need a shutdown_ref. Instead the poller state is managed by a PosixEnginePollerManager class which is held as a shared_ptr by the event engine. In the event engine destructor, we no simple trigger shutdown on this class.

Vignesh2208:
Friendly ping on this one :) Thanks!

I also had to modify //test/cpp/microbenchmakrs/bm_event_engine_run.cc --> In one of the benchmarks a closure was getting deleted at the end of the benchmark after the event engine destructor runs. But with the new model, the threadpool can persist even after event engine destruction and thus the closure may still exist in the threadpool after event engine destruction so its not safe to destroy the closure at the end of the benchmark. I updated the benchmark to never delete the closure and ignore the memory leak instead with absl::IngoreLeak. Do you think this is ok ?

Vignesh2208:
Also I'm pretty sure the PHP failure (memory leak) is due to the threadpool getting destroyed after the event engine. This is not really a memory leak but I'm not sure if we should configure the valgrind memory leak test here or should we just go back to the older model where we wait for the poller to completely exit in the event engine destructor ?

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -15,23 +15,136 @@
...
+void PosixEnginePollerManager::TriggerShutdown() {
+  // If the poller is external, dont try to shut it down. Otherwise
+  // set poller state to PollerState::kShuttingDown.
+  if (poller_state_.exchange(PollerState::kShuttingDown) ==
+      PollerState::kExternal) {
+    poller_ = nullptr;
+    return;
+  }
+
+  if (poller_ != nullptr) {
```
ctiller, line 85: I don't see any way that poller_ could be nullptr here?
Maybe we make it a `grpc_event_engine::posix_engine::PosixEventPoller* const poller_` in the declaration to enforce that it's initialized and avoid the need for the ifs.

ctiller (approved):

src/core/lib/event_engine/posix_engine/posix_engine.cc
Vignesh2208: Removed the if statement. I couldn't make it a const though because the value needs to change a few lines above.

PR #31129, opened 2022-09-27
Make TimerManager threads non-joinable

drfloob:
On gRPC shutdown, any unjoined TimerManager threads will cause TSAN to detect thread leaks. This PR resolves issues I saw on end2end test shutdown in another PR, where a single timer manager thread was always alive after the test ended.

The long-term solution is to integrate the new ThreadPool here, but this unblocks me for now.


ctiller (commented):
I'm good with this... if it blows up in our face I think the solution will be to just do the timer integration. That said, I'd appreciate a cherry-pick/TGP prior to hitting submit.

drfloob:
The train ran a few hours late today, but did finally arrive 9 hours later ... none of the 20 errors were related.

ctiller (approved):

PR #31132, opened 2022-09-27
[Ruby] Update required_ruby_version from Ruby 2.5 to Ruby 2.6

kaylareopelle:
When attempting to install grpc 1.49.1 on Ruby version 2.5.9, the following error is raised: 

"grpc-1.49.1-x86_64-linux requires ruby version >= 2.6, < 3.2.dev, which is incompatible with the current version, ruby 2.5.9p229"

This change updates the documentation to make the Ruby version requirement public.


kaylareopelle:
I believe this error is coming up because [the latest minor version of `googleapis-common-protos-types` requires Ruby 2.6 and above](https://rubygems.org/gems/googleapis-common-protos-types/versions/1.4.0). An alternative solution could be to update the dependency version requirements in the gemspec to 1.3.2.

LOC: https://github.com/grpc/grpc/blob/a81391d9316bf77fdea3ded66fd844692c70109a/grpc.gemspec#L34

```diff
- s.add_dependency 'googleapis-common-protos-types', '~> 1.0'
+ s.add_dependency 'googleapis-common-protos-types', '~> 1.0', '<= 1.3.2'
```

kaylareopelle:
This looks like it's been resolved in 1.49.1! Thanks!

kaylareopelle:
The `required_ruby_version` problem has returned. It also seems like Ruby 3.3 isn't supported either. Could you reconsider this PR?

kaylareopelle:
cc @yashykt

apolcyn (approved):

kaylareopelle:
Thanks for the approval, @apolcyn! What else do we need to do to bring this across the finish line?

apolcyn:
@kaylareopelle sorry for the delay. We also need to make this update to `templates/grpc.gemspec.template` (this file is normally generated from the prior).

kaylareopelle:
Thanks, @apolcyn! I appreciate your help! 

I've updated the PR to include the change in `templates/grpc.gemspec.template` as well.

chadlwilson:
> The `required_ruby_version` problem has returned. It also seems like Ruby 3.3 isn't supported either. Could you reconsider this PR?

@kaylareopelle I believe the Ruby 3.3 problem is likely only with the native gems (to avoid building from source) since they have an extra requirement (`< 3.3.DEV`) in the gemspec generated (I think) by `rake-compiler-dock` when building the gems. Since 3.3.0 final is released now, see #35399 for my attempt to move this forward if approved and tests pass.

Issue #30746, opened 2022-08-25
C# Protobuf msbuild task fails with glob pattern if "@" is present in the path


Blackclaws:
### What version of gRPC and what language are you using?
Grpc.AspNetCore 2.47.0 / Grpc.Tools 2.47.0

### What operating system (Linux, Windows,...) and version?
Linux

### What runtime / compiler are you using (e.g. .NET Core SDK version `dotnet --info`)
6.0.301

### What did you do?
We are distributing our proto files using scoped npm packages. These naturally have an "@" as part of their path.
Using the Protobuf msbuild task with those and a glob pattern fails with the compiler failing to add the generated files to the build.

```
    <Protobuf Include="..\..\..\node_modules\@arfinity\authentication-service-protobuf\**\*.proto">
      <GrpcServices>Client</GrpcServices>
      <Link>Protos\AuthenticationService\*.proto</Link>
      <Access>Internal</Access>
      <ProtoRoot>..\..\..\node_modules\@arfinity\authentication-service-protobuf</ProtoRoot>
      <AdditionalImportDirs>..\..\..\node_modules\@arfinity\protobuf-library</AdditionalImportDirs>
    </Protobuf>
```

The generation of the files works fine. Its just that they aren't included in the compilation for some reason. Removing the glob and directly referencing a single specific file works.

### What did you expect to see?
I expected all generated files are compiled into the build.

### What did you see instead?
The files aren't included in the build and the following error is logged as well:
```
0>/home/felix/.nuget/packages/grpc.tools/2.47.0/build/_protobuf/Google.Protobuf.Tools.targets(240,5): Warning  : The following files have no known outputs, and will be always recompiled as if out-of-date:
```

Note that there isn't any file listed here, this error just pops up.

### Anything else we should know about your project / environment?

The issue occurs anytime an @ is part of the path that is being globbed. Even if you don't explicitly make it part of the path using ** with directories containing @ signs triggers this bug.

tonydnewell:
@Blackclaws I've managed to reproduce this problem.  I shall look in detail later.

As a workaround are you able to create a link with a safe name for the directory that contains the '@'?
e.g. `ln -s safename @arfinity` and use that name in the csproj file?

Blackclaws:
Yes that workaround works and its what I'm currently using. I have to link it outside the node_modules folder of course, but for now its workable solution until the bug is fixed.

tonydnewell:
[...]

PR #31133, opened 2022-09-27
Fix msbuild failing when '@' is present in path

tonydnewell:
Fix for https://github.com/grpc/grpc/issues/30746 - "C# Protobuf msbuild task fails with glob pattern if "@" is present in the path"

There was ambiguity in references to `%(Source)` metadata on some items. These are now qualified to remove the ambiguity.  This is what was causing the problems when there was a '@' in the path to the proto files.

Also fix the printing of the warning for 'no known outputs' - it was using the wrong item list so no file names where output.

Testing on Windows and Linux on various Grpc projects.


jtattermusch (approved):

jtattermusch:
@tonydnewell  Some required tests haven't run since the branch of this PR is slightly old - can you please rebase on top of master to re-run the tests. I will then merge.

PR #31150, opened 2022-09-28
Reland tcp endpoints for posix event engine

Vignesh2208:
Revert https://github.com/grpc/grpc/pull/31134

When the posix_endpoint_test runs inside a docker, tx zero-copy fails and sees an ENOBUFS error. This is because of improperly configured ulimit hard memlock and RLIMIT_MEMLOCK values in the docker. The RPC gets stuck and  eventually times out . The fix is the diff of the latest commit with the base: https://github.com/grpc/grpc/pull/31150/commits/29c0642de6d9a2519124ab157f1a50d43ae91699. The fix prevents gRPC from using tx-zero copy if RLIMIT_MEMLOCK or ulimit hard memlock values are not set. It uses getrlimit to read the RLIMIT_MEMLOCK value and parses "/etc/security/limits.conf" file to extract ulimit hard memlock value.

* RLIMIT_MEMLOCK is a process wide limit which controls the max amount of virtual memory belonging to a process that may be locked into RAM.
* ulimit hard memlock is a per-socket limit which controls the max amount of virtual memory that may be locked into RAM.

Tx zero-copy respects both of these limits and raises an ENOBUFS error if the socket attempts to send more bytes than any of these limits.


src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1262 @@
...
+      // hard memlock ulimit is ignored for privileged user.
+      return UINT64_MAX;
+    }
+    // Parses /etc/security/limits.conf file for a line of the following format:
+    // * hard memlock <value>
+    // It extracts <value> and returns it. A value of UINT64_MAX represents
+    // unlimited or infinity. Hard memlock value should be set to allow zerocopy
+    // sendmsgs to succeed. It controls the maximum amount of memory that can be
+    // locked by a socket in RAM.
+    std::ifstream limits_file("/etc/security/limits.conf");
```
drfloob, line 139: Could the hard memlock setting be overriden by configurations in `/etc/security/limits.d/`? And are there any supported operating systems that would have this file in another location?

Vignesh2208: Hmm. Interesting. Looks like /etc/security/limits.d/ can have user/program specific overrides and /etc/security/limits.conf is the system default one. I updated the logic to parse every file in /etc/security/limits.d/ + /etc/security/limits.conf to see if any of them have hard memlock set. If thats true, then tx zero copy is allowed to proceed. However its still possible that the memlock may not be set for the program/user running the grpc process. In that case tx zero copy will still see ENOBUFS and we will print a log. I think that should be sufficient.

Vignesh2208: Only linux supports this feature and its specified in these locations /etc/security/limits.d/ or /etc/security/limits.conf

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -100,6 +109,89 @@ ssize_t TcpSend(int fd, const struct msghdr* msg, int* saved_errno,
 }
 
 #ifdef GRPC_LINUX_ERRQUEUE
+
+#define GRPC_LOG_EVERY_N_SEC(n, format, ...)                                  \
```
drfloob, line 113: I'd recommend putting this with general logging utils as well. And is time_precise required here?

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -1074,12 +1190,31 @@ PosixEndpointImpl::PosixEndpointImpl(EventHandle* handle,
...
+      gpr_log(
+          GPR_ERROR,
+          "Tx zero-copy will not be used by gRPC since RLIMIT_MEMLOCK value is "
+          "not set. Consider raising its value with setrlimit().");
+    } else if (GetUlimitHardMemLock() == 0) {
       zerocopy_enabled = false;
-      gpr_log(GPR_ERROR, "Failed to set zerocopy options on the socket.");
+      gpr_log(GPR_ERROR,
+              "Tx zero-copy will not be used by gRPC since hard memlock ulimit "
+              "value is not set. Use ultimit -l <value> to set its value.");
```
drfloob, line 1203: Did you mean `ulimit`?

drfloob (commented):
I did a superficial pass, mostly LGTM, but I'd recommend getting a review from someone more familiar with zero-copy workings.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
Vignesh2208: Moved it to src/core/lib/gprpp/time.h

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -0,0 +1,1262 @@
...
+      gpr_log(GPR_INFO, format, __VA_ARGS__);                                 \
+    }                                                                         \
+  } while (0)
+
+#define CAP_IS_SUPPORTED(cap) (prctl(PR_CAPBSET_READ, (cap), 0) > 0)
+
+// Ulimit hard memlock controls per socket limit for maximum locked memory in
+// RAM.
+uint64_t GetUlimitHardMemLock() {
+  static uint64_t kUlimitHardMemLock = []() -> uint64_t {
```
ctiller, line 144: this should be const

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -122,39 +123,55 @@ ssize_t TcpSend(int fd, const struct msghdr* msg, int* saved_errno,
 
 #define CAP_IS_SUPPORTED(cap) (prctl(PR_CAPBSET_READ, (cap), 0) > 0)
 
+uint64_t ParseUlimitMemLockFromFile(std::string file_name) {
+  std::ifstream limits_file(file_name);
```
ctiller, line 115: let's use grpc_load_file here to keep it consistent with everywhere else we read files in production code
(and avoid some bloat!)

Vignesh2208: Done. I created two separate files src/core/lib/gprpp/load_file{.h,.cc} to handle loading data from files. It returns a grpc_core::Slice type.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc

src/core/lib/gprpp/load_file.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include "absl/status/statusor.h"
+
+#include "src/core/lib/slice/slice.h"
+
+namespace grpc_core {
+
+// Loads the content of a file into a slice. add_null_terminator will add a NULL
+// terminator if true.
+absl::StatusOr<Slice> GrpcLoadFile(std::string filename,
```
ctiller, line 30: This can just be `LoadFile` - the grpc-ness can be inferred by the grpc_core namespace.

ctiller: Let's actually split this out and land it separately - it should probably have its own test, and we need to update the other 90-odd call sites.

I'd recommend:
1. load_file.{h,cc} PR to land this bit
2. this pr
3. pr to convert grpc_load_file call sites

(1) is likely to stick, (2) has some risk attached, and (3) is work we should do to clean up here

ctiller (changes requested):

src/core/lib/gprpp/load_file.h
Vignesh2208: I landed a LoadFile PR: https://github.com/grpc/grpc/pull/31176 now.

ctiller (approved):

PR #31152, opened 2022-09-28
[PosixEventEngine] Ensure threads are shut down before destroying other engine state

drfloob:
This fixes a race condition where closures attempt to deregister themselves after the engine's mutex and task_handle map have been destroyed. 0 failures in 120k runs (various asan, msan, tsan, opt, dbg builds).

The breakage was bisected to https://github.com/grpc/grpc/pull/30764


ctiller (commented):
Entirely too many semantics are attached to the ordering of fields in structs, but this looks like it'll fix it - -thanks!

ctiller (approved):

Vignesh2208 (approved):

PR #31161, opened 2022-09-29
Visibility change to orphanable for xds_client_core

allanrbo:
Change of visibility for Google internal project.


allanrbo:
@markdroth has the context on this one. I would assign this PR to him but I apparently do not have access to do so in Github.

markdroth (approved):

PR #31166, opened 2022-09-29
Precondition ChannelArgs with EventEngines

drfloob:
If an EventEngine is not explicitly provided to ChannelArgs, the default EventEngine will be set when ChannelArgs are preconditioned.

This relands the second half of #30556, along with some added engine source migrations. shared_ptr support was added in https://github.com/grpc/grpc/pull/31056


src/core/lib/event_engine/default_event_engine.cc
```diff
@@ -78,5 +81,18 @@ void ResetDefaultEventEngine() {
...
 
+namespace {
+grpc_core::ChannelArgs EnsureEventEngineInChannelArgs(
+    grpc_core::ChannelArgs args) {
+  if (args.ContainsObject<EventEngine>()) return args;
+  return args.SetObject<EventEngine>(GetDefaultEventEngine());
+}
+}  // namespace
+
+void RegisterEventEngine(grpc_core::CoreConfiguration::Builder* builder) {
```
markdroth, line 92: Suggest calling this `RegisterEventEngineChannelArgPreconditioning()`, to make it clear what it's actually registering.

markdroth (commented):
This looks good overall!

include/grpc/impl/codegen/grpc_types.h
```diff
@@ -457,6 +457,8 @@ typedef struct {
  * channel arg. Int valued, milliseconds. Defaults to 10 minutes.*/
 #define GRPC_ARG_SERVER_CONFIG_CHANGE_DRAIN_GRACE_TIME_MS \
   "grpc.experimental.server_config_change_drain_grace_time_ms"
+/** EventEngine shared_pointer */
```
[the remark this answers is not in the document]
drfloob: Done

src/core/lib/event_engine/default_event_engine.cc
drfloob: Done

src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
```diff
@@ -478,16 +476,17 @@ void WeightedTargetLb::UpdateStateLocked() {
...
+    : weighted_child_(std::move(weighted_child)) {
   timer_handle_ =
-      engine_->RunAfter(kChildRetentionInterval, [self = Ref()]() mutable {
-        ApplicationCallbackExecCtx app_exec_ctx;
-        ExecCtx exec_ctx;
-        self->weighted_child_->weighted_target_policy_->work_serializer()->Run(
-            [self = std::move(self)] { self->OnTimerLocked(); },
-            DEBUG_LOCATION);
-      });
+      weighted_child_->weighted_target_policy_->channel_args()
```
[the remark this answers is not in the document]
drfloob: Sounds good, let's chat about channel-wide mechanisms soon. For now, I've made the change to have the WeightedTargetLb own an engine ref (taken from ChannelArgs on lb policy construction).

src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
```diff
@@ -175,6 +175,10 @@ class WeightedTargetLb : public LoadBalancingPolicy {
       return picker_wrapper_;
     }
 
+    WeightedTargetLb* weighted_target_policy() {
```
markdroth, line 178: This method is not necessary.  `DelayedRemovalTimer` is nested inside of `WeightedChild` and can therefore access its private data members directly.

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_manager.cc
```diff
@@ -466,7 +462,9 @@ void XdsClusterManagerLb::ClusterChild::Orphan() {
   // the child.
   picker_wrapper_.reset();
   if (delayed_removal_timer_handle_.has_value()) {
-    engine_->Cancel(*delayed_removal_timer_handle_);
+    xds_cluster_manager_policy_->channel_args()
```
markdroth, line 465: This LB policy should use the same pattern as the grpclb and weighted_target LB policies: The LB policy should be given a `std::shared_ptr<EventEngine>` data member that will be set in its ctor using `GetObjectRef<>()`, and the nested classes can access that data member directly instead of fishing it out of channel args themselves.

src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc
drfloob: Fixed, thanks.

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_manager.cc
drfloob: Good call, done. I've also renamed the Subchannel member to keep things consistent.

ctiller:
I'm not sure this will be safely usable yet: we need to have some story with init.cc and the reset of event engine we do there prior to shutdown. If the event engine is not quiesced before iomgr shuts down, we have a lot of code that assumes it will be available that will not be callable from `EventEngine::Run` and friends.

drfloob:
Thanks for pointing that out, @ctiller. ResetDefaultEventEngine isn't necessary anymore now that gRPC keeps a weak_ptr instead of a raw pointer, so I've removed the function https://github.com/grpc/grpc/pull/31213. To clarify the story, though, this means that new EventEngines will be created as necessary, even after gRPC is shut down (for however long that concept exists). And gRPC internals will not keep those engines alive if the callers of `GetDefaultEventEngine` let go of their references.

drfloob:
I believe that means this PR is safe to land. Please chime in if you disagree.

ctiller:
This nor #31213 are safe to land right now. We definitely have code that assumes that it can access iomgr related code that can be triggered via an `EventEngine::Run` callback.

ctiller:
(see #31208 for a writeup on some of the stuff happening here)

markdroth:
Craig's #31265 should allow us to move forward with most of this, but I think we want to do it a little differently.

I've sent https://github.com/drfloob/grpc/pull/333 against this PR with what I think is the approach we want here.  (Note that most of the changes there are from merging master; the actual changes that I made are in [the last 5 commits](https://github.com/drfloob/grpc/pull/333/files/f71af7dceb34b35d8dbd0eee39906816b9fa1cef..5e0ac029aa2a5d694f0caaaa526eacc308277f44)).

The main idea here is that we want most of the modules within C-core to get the EE instance from a central location in the channel rather than storing their own refs to the EE instance.  For now, that central location is the `grpc_channel_stack` struct (which will eventually be merged into `grpc_core::Channel`).  Both `ChannelIdleFilter` and `ClientChannel` are now getting the EE instance from that central point.  And I've added a new method to the LB helper interface for the LB policies to use, so that they don't need to dig it out of channel args or store their own refs either.

Note that using `grpc_channel_stack` as the central point to store the EE instance makes sense, since we probably do want to hold our own ref to the EE instance in both the parent channel and in each subchannel stack.  However, it also means that we're storing another ref in the dynamic filter stack, which is not strictly necessary.  But we can work on ways to optimize out some redundancy later as needed (and I suspect some of that will naturally fall out of the promise conversion).

Also, we don't yet want to allow passing in the EE instance from the application via channel args, so I've removed the channel arg macro from the include tree and changed the string to start with `grpc.internal.`, so that it cannot be passed in from outside.  We are still using the infrastructure from this PR to pass the EE instance around via channel args *within* C-core; we're only disallowing passing it in from outside.  This will change as we get further along with eliminating the iomgr code.

Note that we are still using channel args to pass the EE instance down into the `Subchannel` object itself, which is still storing its own ref to the EE instance.  That is actually a bit sub-optimal, since this effectively duplicates the ref that will already be held in the subchannel stack.  However, the subchannel stack exists only when the subchannel is actually connected, and it needs to schedule the backoff timer when it is not connected, so I don't see any easy way to avoid this duplication for now.

Finally, Craig tells me that with #31265, there are deadlock issues if code within iomgr itself tries to instantiate an EE instance, so I've reverted the changes resolve_address_posix.{h,cc}.  I think it's fine for iomgr to use the default global EE instance, since iomgr is global anyway, not per-channel.

@ctiller and @drfloob, please review my changes and let me know what you think.  Thanks!

drfloob:
Thanks Mark! I like this approach, and I've merged it in.

> Also, we don't yet want to allow passing in the EE instance from the application via channel args

For transparency, the main reason is to simplify our debugging as we continue to land EventEngine usage in core. Having multiple EventEngines in play adds complexity, which we can defer to make the migration easier. There's nothing technically preventing us from exposing the API at this point. If there are sufficiently motivating use cases to open the API, we will do so.

test/core/channel/channel_stack_test.cc
```diff
@@ -32,10 +33,14 @@
...
 static grpc_error_handle channel_init_func(grpc_channel_element* elem,
                                            grpc_channel_element_args* args) {
-  EXPECT_EQ(args->channel_args->num_args, 1);
-  EXPECT_EQ(args->channel_args->args[0].type, GRPC_ARG_INTEGER);
-  EXPECT_STREQ(args->channel_args->args[0].key, "test_key");
-  EXPECT_EQ(args->channel_args->args[0].value.integer, 42);
+  int test_value = grpc_channel_args_find_integer(args->channel_args,
+                                                  "test_key", {-1, 0, INT_MAX});
+  EXPECT_EQ(test_value, 42);
+  // TODO(hork): replace the string with a macro. See TODO in channel_args.h
```
markdroth, line 39: If we need to use this in tests, I'd be fine with adding such a macro, as long as it's in a file in src instead of in include.

test/core/filters/filter_fuzzer.cc
```diff
@@ -247,7 +248,10 @@ RefCountedPtr<AuthorizationEngine> LoadAuthorizationEngine(
 template <typename FuzzerChannelArgs>
 ChannelArgs LoadChannelArgs(const FuzzerChannelArgs& fuzz_args,
                             GlobalObjects* globals) {
-  ChannelArgs args = ChannelArgs().SetObject(ResourceQuota::Default());
+  ChannelArgs args =
```
markdroth, line 251: Should we just precondition the channel args here?  If these args are going to be passed to a filter, the filter will assume that they've already been preconditioned.

BUILD
```diff
@@ -1759,6 +1759,7 @@ grpc_cc_library(
         "iomgr_timer",
         "ref_counted_ptr",
         "//src/core:arena",
+        "//src/core:channel_args_preconditioning",
```
markdroth, line 1762: This dependency should not be needed.  Preconditioning should be done inside of C-core, not at the C++ layer.

BUILD
```diff
@@ -1818,6 +1820,7 @@ grpc_cc_library(
...
```
markdroth, line 1823: Same here.

src/core/lib/channel/channel_args_preconditioning.h
```diff
@@ -53,6 +53,9 @@ class ChannelArgsPreconditioning {
   // Returns a new channel args object that is owned by the caller.
   ChannelArgs PreconditionChannelArgs(const grpc_channel_args* args) const;
 
+  // Take channel args and precondition them.
+  ChannelArgs PreconditionChannelArgs(const ChannelArgs& args) const;
```
markdroth, line 57: I don't think we should need this override.  The only places where we should be preconditioning channel args are at the C-core API boundary, where the caller passes in C-style args and we return `grpc_core::ChannelArgs`.

BUILD
drfloob: Done

BUILD
drfloob: Done

test/core/filters/filter_fuzzer.cc
drfloob: Done

test/core/channel/channel_stack_test.cc
drfloob: Added in channel_args.h as `GRPC_INTERNAL_ARG_EVENT_ENGINE`

src/core/ext/filters/client_channel/lb_policy/xds/xds_wrr_locality.cc
```diff
@@ -313,6 +314,11 @@ absl::string_view XdsWrrLocalityLb::Helper::GetAuthority() {
   return xds_wrr_locality_->channel_control_helper()->GetAuthority();
 }
 
+grpc_event_engine::experimental::EventEngine*
+XdsWrrLocalityLb::Helper::GetEventEngine() {
+  return this->xds_wrr_locality_->channel_control_helper()->GetEventEngine();
```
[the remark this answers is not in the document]
drfloob: Done

src/core/lib/channel/channel_args_preconditioning.h
drfloob: See my comment above regarding wherever we create ChannelArgs internally. There are exceptions where it's necessary, and I didn't think it would be right to backtrack to using the C-type just for this. 

This gave me an idea, however: We could instead ensure that all ChannelArgs are preconditioned in the constructor.

src/core/lib/channel/channel_stack.cc
```diff
@@ -117,8 +119,12 @@ grpc_error_handle grpc_channel_stack_init(
     }
   }
 
+  auto conditioned_channel_args = grpc_core::CoreConfiguration::Get()
```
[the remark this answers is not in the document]
markdroth: In principal, I think this preconditioning should be done where the channel args are created, not where they are used.  How many places do we create new channel args within C-core?

We should also consider why we are creating channel args within C-core in the first place.  In principal, there should be very few places where this happens -- it should generally be necessary only when constructing a control-plane channel within C-core, such as the xDS channel, the grpclb channel, or the ALTS channel, and those places should get preconditioning automatically via `grpc_channel_create()`.  If there are other places where we're doing this, I think we should look at each one and decide whether it's actually necessary.

For example, do we actually have any cases where we use `ChannelStackBuilder` without calling `SetChannelArgs()`?  If not, maybe we should just change the `ChannelStackBuilder` API to require passing in channel args to its ctor, so that this case goes away.

I don't think we want the preconditioning to happen whenever we instantiate a `ChannelArgs` object, because those objects get copied all the time within C-core.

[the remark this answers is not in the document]
drfloob: > How many places do we create new channel args within C-core?

To begin with, I count around 30 default-constructed ChannelArgs class members within src/core. I haven't gone through to see how many get overwritten, or could otherwise be accessed before being overwritten - LB policies and client channel resolvers make up about a third of the set. I'll take a deeper look here.

I did the experiment you suggested with ChannelStackBuilder, indeed we can add a ChannelArgs argument to the constructor, and all PR tests pass. So we may be safe with that change alone.

> I don't think we want the preconditioning to happen whenever we instantiate a ChannelArgs object, because those objects get copied all the time within C-core.

... and we'd be adding a dependency cycle between core config and channel args, as well as making making channel args fragments more expensive (as used in resolvers). It was a thought, but wouldn't pan out right now.

src/core/lib/channel/channel_args_preconditioning.h
drfloob: I removed this overload. I have a hunch we may still need it, but I was able to remove the need for it in this PR. Barring discovery that any internally-created ChannelArgs need preconditioning, I think it can stay gone.

src/core/lib/channel/channel_stack_builder.h
```diff
@@ -63,6 +64,7 @@ class ChannelStackBuilder {
   grpc_transport* transport() const { return transport_; }
 
   // Set channel args.
+  // args *must be* preconditioned already.
   ChannelStackBuilder& SetChannelArgs(const ChannelArgs& args);
```
markdroth, line 68: Do we still call this anywhere?  If not, we can remove it.

src/core/lib/channel/channel_args_preconditioning.h
markdroth: Looks like you removed the definition of the overload in the .cc file but forgot to remove the declaration here in the .h file.

markdroth (approved):
This looks good to me, modulo a couple of minor comments below.

Please make sure @ctiller reviews as well before merging.

Thanks!

src/core/lib/channel/channel_stack_builder.h
drfloob: It was used in one place, but I was able to refactor that usage away without much extra spaghettification.

drfloob:
> Looks like you removed the definition of the overload in the .cc file but forgot to remove the declaration here in the .h file.
Good catch, removed.

src/core/lib/channel/channel_args_preconditioning.h
drfloob: Removed, thanks.

src/core/lib/surface/channel.h
```diff
@@ -158,7 +156,7 @@ class Channel : public RefCounted<Channel>,
   }
 
   grpc_event_engine::experimental::EventEngine* event_engine() const {
-    return event_engine_.get();
+    return channel_stack_->EventEngine();
```
ctiller, line 159: (no change required)

Probably ok for now, we might want to cache the pointer on this object at some point though (thinking about number of pointer chases on common per-call paths)

CC @markdroth

ctiller (approved):
looks good... left one comment that needs no action (but wanted to note)

src/core/lib/surface/channel.h
markdroth: I was thinking about that, but wouln't any per-call paths get this pointer from the promise context instead?  I would expect that only control-plane code would actually wind up using this method.

PR #31187, opened 2022-09-30
Make event engine client test more robust

Vignesh2208:
In preparation for landing clients, the test is modified to use a shared_ptr event engine.


test/core/event_engine/test_suite/client_test.cc
```diff
@@ -195,7 +209,10 @@ TEST_F(EventEngineClientTest, MultipleIPv6ConnectionsToOneOracleListenerTest) {
     // Create a test EventEngine client endpoint and connect to a one of the
     // addresses bound to the oracle listener. Verify that the connection
     // succeeds.
-    ChannelArgsEndpointConfig config;
+    grpc_core::ChannelArgs args;
```
drfloob, line 212: consider renaming these variables so it's clear you didn't intend to use the same-named variables in the outer scope.

test/core/event_engine/test_suite/client_test.cc
```diff
@@ -147,6 +155,9 @@ TEST_F(EventEngineClientTest, ConnectExchangeBidiDataTransferTest) {
                                     client_endpoint.get())
                     .ok());
   }
+  client_endpoint.reset(nullptr);
```
drfloob, line 158: the default argument is effectively a `nullptr`, you can just say `client_endpoint.reset();`. https://en.cppreference.com/w/cpp/memory/unique_ptr/reset

test/core/event_engine/test_suite/client_test.cc
Vignesh2208: renamed it to client_config and client_args

test/core/event_engine/test_suite/client_test.cc
Vignesh2208: Fixed.

test/core/event_engine/test_suite/client_test.cc
```diff
@@ -135,7 +144,8 @@ TEST_F(EventEngineClientTest, ConnectExchangeBidiDataTransferTest) {
   EXPECT_TRUE(client_endpoint != nullptr);
```
drfloob, line 144: I'm not sure how this slipped in previously, but let's clean this up now. Comparing unique_ptr directly with a nullptr is wonky ... it works, but relies on some hidden casting to bool, and it appears to be deprecated in c++20. See https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_cmp. Also, doing direct comparisons like this in googletest, you want to use the family of `EXPECT_NE(a, b)` tests rather than `EXPECT_TRUE(a != b)` because errors like "expected: nullptr, actual: 0x12345" can be more useful than "expected: true, actual: false".

Consider instead `EXPECT_NE(client_endpoint.get(), nullptr)`.

test/core/event_engine/test_suite/client_test.cc
```diff
@@ -195,7 +211,10 @@ TEST_F(EventEngineClientTest, MultipleIPv6ConnectionsToOneOracleListenerTest) {
     // Create a test EventEngine client endpoint and connect to a one of the
     // addresses bound to the oracle listener. Verify that the connection
     // succeeds.
-    ChannelArgsEndpointConfig config;
+    grpc_core::ChannelArgs client_args;
+    auto quota = grpc_core::ResourceQuota::Default();
```
drfloob, line 215: Please rename this as well, it hides a variable in the outer scope.

Vignesh2208: Fixed.

test/core/event_engine/test_suite/client_test.cc
Vignesh2208: Fixed.

drfloob (approved):

PR #31188, opened 2022-09-30
Event Engine related minor cleanup

Vignesh2208:
Fixes some includes, build issues and bug in poller fallback logic in preparation for landing clients.


src/core/lib/event_engine/posix_engine/event_poller_posix_default.cc
```diff
@@ -37,17 +37,21 @@ bool PollStrategyMatches(absl::string_view strategy, absl::string_view want) {
 }  // namespace
 
 PosixEventPoller* GetDefaultPoller(Scheduler* scheduler) {
-  grpc_core::UniquePtr<char> poll_strategy =
-      GPR_GLOBAL_CONFIG_GET(grpc_poll_strategy);
+  static const char* poll_strategy =
```
drfloob, line 40: There's some chat of having scoped configs, for tests at least. When that happens, I think we'd want to query the poller every time, but ... maybe YAGNI at this point. Something to consider.

Vignesh2208: Surprisingly it causes tsan issues. The poller is expected to be created each time the event engine constructor runs and this can coincide with GPR_GLOBAL_CONFIG_GET(grpc_poll_strategy) invoked elsewhere

drfloob: That sounds like a bug in global config. @ctiller is this on your radar?

Vignesh2208: GPR_GLOBAL_CONFIG_GET is not thread safe. but I dont want to add a mutex as well. I think since our test infrastructure only sets the variable once prior to execution of the whole test, perhaps its ok to read it into a static global variable once

PR #31191, opened 2022-09-30
xDS interop GCE framework: pin grpcio-tools to use protobuf 3.x

sergiitk:
https://source.cloud.google.com/results/invocations/ae6a6af2-f1d7-4982-9b4c-8edfa75d46d4


gnossen (approved):
Can you please also add a TODO to unpin? This is good signal for us currently when this breaks.

PR #31213, opened 2022-10-03
[merged separately] Remove ResetDefaultEventEngine

drfloob:
Now that it is a weak_ptr, there's no need to explicitly reset it. When the tracked shared_ptr is deleted, the weak_ptr will fail to lock, and a new default EventEngine will be created from within `GetDefaultEventEngine`


ctiller (approved):
I think this all got merged with https://github.com/grpc/grpc/pull/31265

drfloob:
Thanks Craig, closing this.

PR #31225, opened 2022-10-03
Removed no_error_on_hotpath

veblush:
This is not possible with `absl::Status` which doesn't have a way to know whether it's created in a certain period.


ctiller:
This is an unfortunate regression.

veblush:
> This is an unfortunate regression.

Yes, it is. We might be able to do this by having our own creator function but not sure how valuable it would be at the cost of having another function.

yashykt (approved):

Issue #23390, opened 2020-07-06
C# - Server fails to release sockets


mgravell:
(cross-reference: https://github.com/protobuf-net/protobuf-net.Grpc/issues/109)

It seems that `Server` is very reluctant to release sockets, even when `KillAsync` or `ShutdownAsync` is called and awaited - subsequent calls to `Start()` on the same port are likely to fail. Is this known/intentional?

Tested (with failure) on `2.29.0` and `2.30.0`

Example:

``` c#
[Fact]
public async Task ServerFailsToReleasePort()
{
    var port = 10050;

    var server = new Server()
    {
        Ports = { new ServerPort("localhost", port, ServerCredentials.Insecure) }
    };
    // it seems that the port is already in use at this point

    // try all kinds of ways to release the port
    await server.ShutdownAsync();
    // await server.KillAsync(); // kill also doesn't release, but can't Shutdown *and* Kill
    await server.ShutdownTask;
            

    var server2 = new Server()
    {
        Ports = { new ServerPort("localhost", port, ServerCredentials.Insecure) }
    };

    // server2.Services.Add(...) // note: repros even without any services added
    server2.Start(); // <--- crash, socket is still in use.
}
```

mgravell:
(adding @TFTomSun)

TFTomSun:
@jtattermusch on windows

rburns1293:
```C#
class Program
  {
    const int Port = 30051;

    public async static Task Main(string[] args)
    {
        try {
          // Start and shutdown servers repeatedly.
          for (int i = 0; i < 10; i++) {
            Server server = new Server
            {
              Ports = { new ServerPort("localhost", Port, ServerCredentials.Insecure) }
            };
            server.Start();
            await server.ShutdownAsync();
        }
        Console.WriteLine("No exceptions.");
        } catch (Exception e) {
            Console.WriteLine(e);
        }
    }
  }
```

I tried starting up and shutting down servers repeatedly but couldn't recreate the issue on ```2.29.0``` or ```2.30.0```.  This is for ```netcoreapp3.1``` and ```.net48``` on windows 10.  @mgravell I know it's been a while but do you still have a copy of your repo to verify you still see the issue? Apart from something wrong in my test, I wonder if a windows update might have resolved this.

tonydnewell:
This is the same as https://github.com/grpc/grpc/issues/30167

The problem is that Server.Start() is not called before Server.ShutdownAsync().
[...]

PR #31232, opened 2022-10-04
Handle server Shutdown when no Start called

tonydnewell:
When ports are added to a Server they are bound and listening before Start is called. If Start is never called then the Shutdown does not release the listening ports. See https://github.com/grpc/grpc/issues/23390

This fix:
- logs a warning in Shutdown if Start has not been called
- calls Start in Shutdown if any ports have been added so that the core library cleans up correctly


jtattermusch (approved):
LGTM.

jtattermusch:
@veblush can you merge into the 1.46.x branch? I don't have permissions for that.

jtattermusch:

PR #31124, opened 2022-09-26
Make legacy artifact -> packages -> distribtest jobs a noop

jtattermusch:
See https://github.com/grpc/grpc/issues/26549.

The legacy artifact -> packages -> distribtest flow has been replaced by  per-language distribtest jobs.
(see internal cl/476587480 for the changes to the release process)

PR #31237, opened 2022-10-04
Use 64bit linker on windows even when creating 32bit targets

veblush:
To address occasional errors on Windows

```
LINK : the 32-bit linker (C:\PROGRA~2\MIB055~1\2017\COMMUN~1\VC\Tools\MSVC\1414~1.264\bin\Hostx86\x86\link.exe) failed to do memory mapped file I/O on `grpc.lib' and is going to restart linking with a 64-bit linker for better throughput
```

and from [this](https://developercommunity.visualstudio.com/t/memory-error-for-linker-in-vs-155-x64/160714), setting `PreferredToolArchitecture=x64` would be the right solution.

### DIFF
--- a/tools/run_tests/helper_scripts/build_cxx.bat
+++ b/tools/run_tests/helper_scripts/build_cxx.bat
@@ -45,7 +45,7 @@ If "%GRPC_CMAKE_GENERATOR%" == "Ninja" (
 ) else (
   @rem Use one of the Visual Studio generators.
 
-  cmake -G "%GRPC_CMAKE_GENERATOR%" -A "%GRPC_CMAKE_ARCHITECTURE%" -DgRPC_BUILD_TESTS=ON -DgRPC_BUILD_MSVC_MP_COUNT=%GRPC_RUN_TESTS_JOBS% %* ../.. || goto :error
+  cmake -G "%GRPC_CMAKE_GENERATOR%" -A "%GRPC_CMAKE_ARCHITECTURE%" -DCMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE=x64 -DgRPC_BUILD_TESTS=ON -DgRPC_BUILD_MSVC_MP_COUNT=%GRPC_RUN_TESTS_JOBS% %* ../.. || goto :error
 
   @rem GRPC_RUN_TESTS_CXX_LANGUAGE_SUFFIX will be set to either "c" or "cxx"
   cmake --build . --target buildtests_%GRPC_RUN_TESTS_CXX_LANGUAGE_SUFFIX% --config %MSBUILD_CONFIG% || goto :error
--- a/tools/run_tests/run_tests.py
+++ b/tools/run_tests/run_tests.py
@@ -278,8 +278,8 @@ class CLanguage(object):
             self._cmake_architecture_windows = 'x64' if self.args.arch == 'x64' else 'Win32'
             # when builing with Ninja, the VS common tools need to be activated first
             self._activate_vs_tools_windows = activate_vs_tools
-            # always uses x64 host toolkits to secure more memory for the build
-            self._vs_tools_architecture_windows = 'x64' if self.args.arch == 'x64-x64' else 'x64-x86'
+            # "x64_x86" means create 32bit binaries, but use 64bit toolkit to secure more memory for the build
+            self._vs_tools_architecture_windows = 'x64' if self.args.arch == 'x64' else 'x64_x86'
 
         else:
             if self.platform == 'linux':
--- a/tools/run_tests/run_tests.py
+++ b/tools/run_tests/run_tests.py
@@ -278,7 +278,8 @@ class CLanguage(object):
             self._cmake_architecture_windows = 'x64' if self.args.arch == 'x64' else 'Win32'
             # when builing with Ninja, the VS common tools need to be activated first
             self._activate_vs_tools_windows = activate_vs_tools
-            self._vs_tools_architecture_windows = 'x64' if self.args.arch == 'x64' else 'x86'
+            # always uses x64 host toolkits to secure more memory for the build
+            self._vs_tools_architecture_windows = 'x64' if self.args.arch == 'x64-x64' else 'x64-x86'
 
         else:
             if self.platform == 'linux':


veblush:
Hmm cmake doesn't use this `PreferredToolArchitecture`

```
LINK Pass 1: command "C:\PROGRA~2\MIB055~1\2017\COMMUN~1\VC\Tools\MSVC\1414~1.264\bin\Hostx86\x86\link.exe /nologo @CMakeFiles/transport_stream_receiver_test.rsp /out:transport_stream_receiver_test.exe /implib:transport_stream_receiver_test.lib /pdb:transport_stream_receiver_test.pdb /version:0.0 /machine:X86 /debug /INCREMENTAL /subsystem:console /MANIFEST /MANIFESTFILE:CMakeFiles\transport_stream_receiver_test.dir/intermediate.manifest CMakeFiles\transport_stream_receiver_test.dir/manifest.res" failed (exit code 1) with the following output:

LINK : the 32-bit linker (C:\PROGRA~2\MIB055~1\2017\COMMUN~1\VC\Tools\MSVC\1414~1.264\bin\Hostx86\x86\link.exe) failed to do memory mapped file I/O on `grpc.lib' and is going to restart linking with a 64-bit linker for better throughput
```

veblush:
`_vs_tools_architecture_windows` turned out to be the right thing to configure.

tools/run_tests/run_tests.py
```diff
@@ -278,7 +278,8 @@ def configure(self, config, args):
             self._cmake_architecture_windows = 'x64' if self.args.arch == 'x64' else 'Win32'
             # when builing with Ninja, the VS common tools need to be activated first
             self._activate_vs_tools_windows = activate_vs_tools
-            self._vs_tools_architecture_windows = 'x64' if self.args.arch == 'x64' else 'x86'
+            # always use x64 toolset to overcome the memory limit of 32bit linker
+            self._vs_tools_architecture_windows = 'x64'
```
jtattermusch, line 282: wouldn't this make 32bit build on windows silently become 64bit? https://github.com/grpc/grpc/blob/0afcc375890eef6de23c8e0a29b6953b3f07b8e0/tools/run_tests/helper_scripts/build_cxx.bat#L31

Let's doublecheck that before proceeding with this PR.

veblush: Sure! My expectation is that it only controls the tool and generated artifacts should follow `_cmake_architecture_windows` but let's double check.

veblush:
[grpc_build_artifacts_multiplatform](https://fusion2.corp.google.com/invocations/9be02189-a8e6-4573-b738-12cd42bcda90/targets): running

jtattermusch:
> grpc_build_artifacts_multiplatform

FYI the grpc_build_artifacts_multiplatform has basically became a noop after cl/476587480 and https://github.com/grpc/grpc/pull/31124. For artifact build / distribtests, you should look at per-language and per-platform distribtests jobs (e.g. "Distribution Tests Python Linux" on this PR).

tools/run_tests/run_tests.py
jtattermusch: I don't think your assumptions are correct. There are actually 2 ways of building (see the if then else in the script)

1. build with Ninja, which doesn't set architecture through cmake, but instead relies on the default architecture for the vs tools you selected (and I think this is the case that will get broken by your change)
https://github.com/grpc/grpc/blob/0afcc375890eef6de23c8e0a29b6953b3f07b8e0/tools/run_tests/helper_scripts/build_cxx.bat#L41
2. build with cmake's visual studio generator, which lets you pass `-A` architecture flag to cmake
https://github.com/grpc/grpc/blob/0afcc375890eef6de23c8e0a29b6953b3f07b8e0/tools/run_tests/helper_scripts/build_cxx.bat#L48

veblush: Yeah, `vcvarsall` didn't turn out to be effective. But this change worked. I verified this PR on the personal kokoro image and that's the output.

```
T:\src\github\grpc>file workspace_c++_windows_dbg_native_default_default_buildonly\cmake\build\useful_test.exe
file workspace_c++_windows_dbg_native_default_default_buildonly\cmake\build\useful_test.exe
workspace_c++_windows_dbg_native_default_default_buildonly\cmake\build\useful_test.exe: PE32 executable (console) Intel 80386, for MS Windows

T:\src\github\grpc>file workspace_c++_windows_dbg_native_x64_cmake_ninja_vs2017_buildonly\cmake\build\useful_test.exe
file workspace_c++_windows_dbg_native_x64_cmake_ninja_vs2017_buildonly\cmake\build\useful_test.exe
workspace_c++_windows_dbg_native_x64_cmake_ninja_vs2017_buildonly\cmake\build\useful_test.exe: PE32+ executable (console) x86-64, for MS Windows
```

tools/run_tests/helper_scripts/build_cxx.bat
```diff
@@ -45,7 +45,7 @@ If "%GRPC_CMAKE_GENERATOR%" == "Ninja" (
 ) else (
   @rem Use one of the Visual Studio generators.
 
-  cmake -G "%GRPC_CMAKE_GENERATOR%" -A "%GRPC_CMAKE_ARCHITECTURE%" -DgRPC_BUILD_TESTS=ON -DgRPC_BUILD_MSVC_MP_COUNT=%GRPC_RUN_TESTS_JOBS% %* ../.. || goto :error
+  cmake -G "%GRPC_CMAKE_GENERATOR%" -A "%GRPC_CMAKE_ARCHITECTURE%" -DCMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE=x64 -DgRPC_BUILD_TESTS=ON -DgRPC_BUILD_MSVC_MP_COUNT=%GRPC_RUN_TESTS_JOBS% %* ../.. || goto :error
```
jtattermusch, line 48: This might work, but it's still unclear to me whether the build failure we're seeing only happens when building using visual studio generator (which this change could fix) and it doesn't happen in ninja builds (which are unaffected by this change).

In fact, I think a saw ninja builds that suffer from the same problem and therefore should be also fixed:
https://source.cloud.google.com/results/invocations/0d427a67-8e2b-4374-b04a-43b21e43b553/targets;collapsed=/github%2Fgrpc%2Ftoplevel_run_tests_invocations%2Frun_tests_c_windows_dbg_native/tests

jtattermusch: also note that CMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE only exists starting from cmake 3.8 
https://cmake.org/cmake/help/latest/variable/CMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE.html, but currently the minimum cmake version in the CMakeLists.txt is lower (I think 3.5)

tools/run_tests/helper_scripts/build_cxx.bat
```diff
@@ -38,14 +38,13 @@ If "%GRPC_CMAKE_GENERATOR%" == "Ninja" (
 
   @rem Select MSVC compiler (cl.exe) explicitly to make sure we don't end up gcc from mingw or cygwin
   @rem (both are on path in kokoro win workers)
-  cmake -G "%GRPC_CMAKE_GENERATOR%" -DCMAKE_C_COMPILER="cl.exe" -DCMAKE_CXX_COMPILER="cl.exe" -DgRPC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE="%MSBUILD_CONFIG%" %* ../.. || goto :error
+  cmake -G "%GRPC_CMAKE_GENERATOR%" -DCMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE=x64 -DCMAKE_C_COMPILER="cl.exe" -DCMAKE_CXX_COMPILER="cl.exe" -DgRPC_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE="%MSBUILD_CONFIG%" %* ../.. || goto :error
```
jtattermusch, line 41: I don't think just using -DCMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE=x64 for ninja works. The option is only supported by the visual studio generator (i.e. `-G Visual Studio XYZ` not for `-G Ninja` like here. If this gets accepted, I think it will simply get ignored by the ninja build. Note that with Ninja generator, cmake doesn't really pick the toolset for you. It just realies on cl.exe compiler to be on PATH,  which is done by the GRPC_BUILD_ACTIVATE_VS_TOOLS block above.

jtattermusch: I think I found what you need:
https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170#developer_command_file_locations

You'll need to modify the  GRPC_BUILD_ACTIVATE_VS_TOOLS block to pass the right arguments to vcvarsall.bat or use a different .bat file (vcvarsamd64_x86.bat in this case?).

jtattermusch: Passing `x64_x86` instead of just `x86` to vcvarsall.bat (on line 31) should be enough here?

veblush: Jan, how about having this change for msvc build first since it's been failing? We can follow up on ninja after that. Since I'm not familar with Ninja build setup, I'm not sure I can make a right change soon enough.

veblush:
Hmm it seems that it still fails with this change `DCMAKE_VS_PLATFORM_TOOLSET_HOST_ARCHITECTURE=x64`

```cmd.exe /C "cd . && "C:\Program Files\CMake\bin\cmake.exe" -E vs_link_exe --intdir=CMakeFiles\transport_stream_receiver_test.dir --manifests  -- C:\PROGRA~2\MIB055~1\2017\COMMUN~1\VC\Tools\MSVC\1414~1.264\bin\Hostx86\x86\link.exe /nologo @CMakeFiles/transport_stream_receiver_test.rsp  /out:transport_stream_receiver_test.exe /implib:transport_stream_receiver_test.lib /pdb:transport_stream_receiver_test.pdb /version:0.0  /machine:X86 /debug /INCREMENTAL /subsystem:console  && cd ."

LINK Pass 1: command "C:\PROGRA~2\MIB055~1\2017\COMMUN~1\VC\Tools\MSVC\1414~1.264\bin\Hostx86\x86\link.exe /nologo @CMakeFiles/transport_stream_receiver_test.rsp /out:transport_stream_receiver_test.exe /implib:transport_stream_receiver_test.lib /pdb:transport_stream_receiver_test.pdb /version:0.0 /machine:X86 /debug /INCREMENTAL /subsystem:console /MANIFEST /MANIFESTFILE:CMakeFiles\transport_stream_receiver_test.dir/intermediate.manifest CMakeFiles\transport_stream_receiver_test.dir/manifest.res" failed (exit code 1) with the following output:
LINK : the 32-bit linker (C:\PROGRA~2\MIB055~1\2017\COMMUN~1\VC\Tools\MSVC\1414~1.264\bin\Hostx86\x86\link.exe) failed to do memory mapped file I/O on `grpc.lib' and is going to restart linking with a 64-bit linker for better throughput

LINK : restarting link with 64-bit linker `C:\tools\msys64\usr\bin\link.exe'
```

 [log](https://source.cloud.google.com/results/invocations/6c08a5f7-7932-4cf3-b5e4-10e302612464/targets/github%2Fgrpc%2Ftoplevel_run_tests_invocations%2Frun_tests_c%2B%2B_windows_dbg_native_default_default_buildonly/tests)

veblush:
It still used the x86 tools set even with the option.

```
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.14.26428/bin/Hostx86/x86/cl.exe

-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.14.26428/bin/Hostx86/x86/cl.exe -- works

-- Detecting C compiler ABI info

-- Detecting C compiler ABI info - done

-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.14.26428/bin/Hostx86/x86/cl.exe

-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.14.26428/bin/Hostx86/x86/cl.exe -- works
```

veblush:
Oh I just realized that `windows_dbg_native_default_default_buildonly` is using ninja. So all my changes for non ninja build weren't relevant at all.

tools/run_tests/helper_scripts/build_cxx.bat
veblush: As my comment I left below, it turned out that all builds use Ninja so I need to fix for Ninja actually.

jtattermusch:
I pushed a fix to this PR - let's see what the tests say.

jtattermusch (approved):
LGTM assuming the tests pass and no more edits are needed.

veblush:
Test passed! Thanks, Jan!

PR #31244, opened 2022-10-05
weighted_target LB: use uint64_t for aggregate weights to avoid overflow

markdroth:


test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -193,7 +193,7 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType> {
   // Default values for locality fields.
   static const char kDefaultLocalityRegion[];
   static const char kDefaultLocalityZone[];
-  static const int kDefaultLocalityWeight = 3;
+  static const uint32_t kDefaultLocalityWeight = 3;
```
yashykt, line 196: Is this 32 bit? Also, why this change? If we are making this change, shouldn't we be consistent and change priority too?

yashykt (approved):

test/cpp/end2end/xds/xds_end2end_test_lib.h
markdroth: This matches the definition of the field in the xDS protos.

The reason for making this change for weight is that we realized that there's a risk of integer overflow when we aggregate the weights in the weighted_target policy (which is what this PR fixes).  We don't ever aggregate priorities, so there's no need to change the type we're using there.

markdroth:
Now that #31272 has been merged, I've had to change the test to keep the total weight to uint32 max (which I guess should have been merged as part of #31272, not this PR).

I don't currently have a good way to test that we support totals that exceed uint32 max, but I will add that in the future when I write a unit test for the weighted_target policy.

PR #31248, opened 2022-10-05
Removed legacy macros completely

veblush:
Internal repo is now free from legacy macros so it's time to wipe them out.


yashykt (approved):
Please make sure that copybara/import is green before merging or alternatively, cherry-pick it

PR #31250, opened 2022-10-05
Update stream leak test to use transport counter instead of iomgr object counter

apolcyn:
This is to allow `test/core/transport/chttp2/stream_leak_with_queued_flow_control_update_test.cc` to stay around after the iomgr object count API is removed (in https://github.com/grpc/grpc/pull/31239).

I've confirmed this still repros the original issue (described in https://github.com/grpc/grpc/pull/30907) when the fix is removed. 

cc @ctiller


ctiller (approved):

apolcyn:
ruby macos flake is a git download failure

PR #31251, opened 2022-10-05
outlier detection: remove env var protection

markdroth:
Enable outlier detection functionality by default, as described in [gRFC A50](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md).


yashykt (approved):
Release notes yes probably?

Issue #24960, opened 2020-12-10
Adding a test with openssl or variants without TLS v1.3 support


veblush:
From the recent regression fixed by #24955, it's necessary to have a test with openssl or variants without TLS v1.3 support. It will be added to Linux primarily because it's quite easy to establish those environment with docker. Libressl could be considered to be added too.

jiangtaoli2016:
Another related issue https://github.com/grpc/grpc/pull/24959

matthewstevenson88:
As @jiangtaoli2016 mentioned, it would be useful to also test OpenSSL 1.0 and BoringSSL 1.0, since SSL transport security has a bunch of #ifdef's that single out those older versions.

veblush:
I'm trying to understand what's needed specifically to prevent the previous bug. I like to know **what test** could've run on **what environment**. Is there any existing unit test to run for this purpose? Also if we managed to run some test for this, if client and server are using the same OpenSSL 1.0 which doesn't support TLS1.3, would it run into the error? Does it need an environment where the client is built with openssl 1.0 but the server is built with openssl 1.1?

matthewstevenson88:
Thanks for looking into this Esun! The specific failure that was fixed by #24995 would have definitely been caught by an interop test where e.g. the client is using LibreSSL and the server is using a newer OpenSSL version. Here, client/server should be interchangeable, it would have been caught if the client was using the new OpenSSL version and the server was using LibreSSL, I think.

If the client and server were both running OpenSSL 1.0 (or one was running OpenSSL 1.0 and the other was running a newer OpenSSL version), then we would **not** have seen this exact issue, I think. I suggested testing the older OpenSSL versions because I'm worried about preventing a similar bug in the future.

To guard against future bugs, the ideal scenario would be to have the following tests:

- Run [ssl_transport_security_test.cc](https://github.com/grpc/grpc/blob/master/test/core/tsi/ssl_transport_security_test.cc), [ssl_session_cache_test.cc](https://github.com/grpc/grpc/blob/master/test/core/tsi/ssl_session_cache_test.cc), and the C-core e2e test with `h2_ssl_*` or `h2_tls` [fixtures](https://github.com/grpc/grpc/tree/master/test/core/end2end/fixtures) against new/old OpenSSL, new/old BoringSSL, LibreSSL.
[...]

PR #31256, opened 2022-10-06
[OpenSSL] Support for OpenSSL 3

jrandolf:
Update from gtcooke94:
This PR adds support to build and test with OpenSSL3. There were some hiccups with tests as the tests with openssl haven't been exercised in a few months, so they needed some work to fix.

Right now I expect all test files to pass except the following:
- h2_ssl_cert_test
- ssl_transport_security_utils_test

I confirmed locally that these tests fail with OpenSSL 1.1.1 as well, thus we are at least not introducing regressions.


ZhenLian:
Thanks for contributing to gRPC @jrandolf !
I wasn't aware that OpenSSL 3.x has came out. Let me discuss with a few folks within our team to discuss if we should start to support OpenSSL 3.x, and what the priority is. Thank you!
CC @erm-g

chrisse74:
When discussing OpenSSL 3.0 please be aware, that RHEL 9 will switch to OpenSSL 3.0 and will no longer support building against OpenSSL 1.1.1, see https://www.redhat.com/en/blog/experience-bringing-openssl-30-rhel-and-fedora

jre21:
@ZhenLian @erm-g  What were the results of the prioritization discussion?  Upstream support for OpenSSL 1.1.1 ends in six months; I certainly hope you have a plan to ensure your users aren't forced to continue running a security-critical library past the point when vulnerability fixes stop being readily available.

erm-g:
> @ZhenLian @erm-g What were the results of the prioritization discussion? 
Hi @jre21, we're having few internal discussions about the upgrade. The issue is prioritized and we're working on it

erm-g:
b/272519091

AntonRechkov:
Folks, OpenSSL 1.1.1 is EOL in September - https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/
Do you have timeline for switch to OpenSSL 3.0?

Also recommendation from security to use 3.0 and not 3.1 since 3.0 is LTS

erm-g:
> Folks, OpenSSL 1.1.1 is EOL in September - https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/ Do you have timeline for switch to OpenSSL 3.0?
> 
> Also recommendation from security to use 3.0 and not 3.1 since 3.0 is LTS

Hi @AntonRechkov , we're targeting end of July. However, I wouldn't use the term 'switch' in this case since we're still planning to have BoringSSL as default, keep supporting build options for OpenSSL < 3, but start supporting build options for OpenSSL 3.0.

erm-g:
Targeting 1.58 release

chchwy:
My company is currently migrating all projects from OpenSSL 1.1.1 to OpenSSL 3 due to security requirements. I found that gRPC has become one of the few open source projects that still doesn't support OpenSSL 3. I would really appreciate it if this pull request could be merged as soon as possible.

gtcooke94 (changes requested):
The new compiler option needs to be added here - as-is, the `run_tests.py` script can't accept the new option as an argument
https://github.com/grpc/grpc/blob/76203ba5891c8a4a67d8331b125cf7450a677b13/tools/run_tests/run_tests.py#L1702

gtcooke94:
I'm trying to run this locally, going to try to get someone from the team that knows the CI infra better than I do to help out

After adding the compiler option, I'm now getting a failure because it looks like a `tools/dockerfile/test/.cxx_debian11_openssl305_x64.current_version` file is needed that points at an image on `us-docker.pkg.dev` - I'm not clear who generates these, looking into it

src/core/lib/security/credentials/jwt/jwt_verifier.cc
```diff
@@ -535,6 +540,7 @@ static EVP_PKEY* pkey_from_jwk(const Json& json, const char* kty) {
     gpr_log(GPR_ERROR, "Unsupported key type %s.", kty);
     goto end;
   }
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
```
gtcooke94, line 543: This doesn't have a matching `#endif` directive

I believe the right thing to do is change the `#if` immediately below this to an `#elif`

jtattermusch:
> I'm trying to run this locally, going to try to get someone from the team that knows the CI infra better than I do to help out
> 
> After adding the compiler option, I'm now getting a failure because it looks like a `tools/dockerfile/test/.cxx_debian11_openssl305_x64.current_version` file is needed that points at an image on `us-docker.pkg.dev` - I'm not clear who generates these, looking into it

https://github.com/grpc/grpc/blob/master/tools/dockerfile/push_testing_images.sh

gtcooke94:
With #33944 merged these tests will run in CI now - can you merge master and uncomment this line: https://github.com/grpc/grpc/blob/master/tools/run_tests/run_tests_matrix.py#L360

gtcooke94:
Sorry for the messy commit history, it'll all get rebased when we merge though. There were some sneaky problems building tests specifically with OpenSSL (all versions). Been trying to play whack-a-mole with them

gtcooke94 (commented):
I've written a chunk of this now, so commenting to remove my "request changes" review and not giving explicit approval

tools/run_tests/run_tests_matrix.py
```diff
@@ -357,7 +357,7 @@ def _create_portability_test_jobs(
         "gcc7",
         # 'gcc10.2_openssl102', // TODO(b/283304471): Enable this later
         "gcc12",
-        # "gcc12_openssl309", // TODO: Enable this later
+        "gcc12_openssl309",
```
matthewstevenson88, line 360: Have we confirmed that this portability test runs the unit tests for all of the affected files?

Additionally, have we confirmed that this portability test runs at least one of the SSL/TLS e2e tests?

matthewstevenson88 (commented):
Thanks for putting this together! Just a few small questions. I'm also going to seek out another reviewer with OpenSSL-expertise, because I'm worried about introducing gotchas here (especially in the JWT verifier).

tools/run_tests/run_tests_matrix.py
gtcooke94: We have at least confirmed that it compiles tests - see this comment: https://github.com/grpc/grpc/pull/33944#issuecomment-1661040670

I'm not 100% confident on saying this portability test infra is running all of them, @veblush can you comment?

test/core/tsi/ssl_transport_security_test.cc
```diff
@@ -1244,13 +1244,15 @@ TEST(SslTransportSecurityTest, MainTest) {
       // BoringSSL and OpenSSL have different behaviors on mismatched ALPN.
       ssl_tsi_test_do_handshake_alpn_client_no_server();
       ssl_tsi_test_do_handshake_alpn_client_server_mismatch();
-#endif
-      ssl_tsi_test_do_handshake_alpn_server_no_client();
-      ssl_tsi_test_do_handshake_alpn_client_server_ok();
+      // These tests fail with openssl3 and openssl111 currently but not
```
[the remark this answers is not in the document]
gtcooke94: I haven't investigated why they fail, I just verified locally that they also fail when building with OpenSSL 1.1.1, so them failing with OpenSSL3 is not unexpected. They've just not been tested against OpenSSL recently, so I guess what I'm saying is at the least this isn't a regression from the current state.

I'm also looking into why they fail

src/core/lib/security/credentials/jwt/json_token.cc
```diff
@@ -158,7 +162,11 @@ void grpc_auth_json_key_destruct(grpc_auth_json_key* json_key) {
     json_key->client_email = nullptr;
   }
   if (json_key->private_key != nullptr) {
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
```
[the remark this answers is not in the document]
gtcooke94: In making these changes I was trying to minimize changes that would affect the standard building with BoringSSL, so most everything is guarded with #if OPENSSL_VERSION_NUMBER

I don't think it would be bad to update this if all the APIs are available, but I think those changes could be put off to a different PR, what do you think?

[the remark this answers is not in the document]
matthewstevenson88: Good call, agreed. Let's defer these changes to a follow-up PR.

test/core/tsi/ssl_transport_security_test.cc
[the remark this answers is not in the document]
matthewstevenson88: OK gotcha. If it's a build error that also exists for OpenSSL 1.1.1, then that's less concerning. It would be good to get to the bottom of the failure, if possible. Please feel free to send me any logs if I can be helpful.

tools/run_tests/run_tests_matrix.py
matthewstevenson88: If it's not running the tests, could we use the strategy in https://github.com/grpc/grpc/issues/24960#issuecomment-807158529 to manually run each of the tests to confirm we're not introducing regressions for OpenSSL 3?

gtcooke94: I've been doing this locally for the tests that I'm changing. When you say "each of the tests" do you mean all of the test targets?

src/core/lib/security/credentials/jwt/json_token.cc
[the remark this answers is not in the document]
matthewstevenson88: Can we file a bug to track this follow-up?

Please feel free to resolve once this is done.

src/core/tsi/alts/crypt/aes_gcm.cc
```diff
@@ -35,7 +35,12 @@ constexpr size_t kKdfCounterLen = 6;
 constexpr size_t kKdfCounterOffset = 2;
 constexpr size_t kRekeyAeadKeyLen = kAes128GcmKeyLength;
 
-// Struct for additional data required if rekeying is enabled.
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+const char kEvpMacAlgorithm[] = "HMAC";
+char kEvpDigest[] = "SHA-256";
```
matthewstevenson88, line 40: nit: Can we mark this as "const"?

src/core/tsi/alts/crypt/aes_gcm.cc
```diff
@@ -196,7 +201,7 @@ static grpc_status_code aes_gcm_derive_aead_key(uint8_t* dst,
     return GRPC_STATUS_INTERNAL;
   }
   HMAC_CTX_cleanup(&hmac);
-#else
+#elif OPENSSL_VERSION_NUMBER < 0x30000000L
```
matthewstevenson88, line 204: Should this be `#if` rather than `#elif`?

matthewstevenson88 (commented):
Generally LGTM, with a couple of minor open comments. I'll wait until we hear back from Stefan and co. before approving.

src/core/tsi/alts/crypt/aes_gcm.cc
gtcooke94: It's part of an #if/#elif/#else that's structured
```
#if OpenSSL < 1.1
// 1.02 impl
#elif OpenSSL < 3
// 1.1 impl
#else
// 3 impl
#endif
```

src/core/tsi/alts/crypt/aes_gcm.cc
gtcooke94: It didn't want to compile as a const
```
/usr/local/google/home/gregorycooke/repos/grpc/src/core/tsi/alts/crypt/aes_gcm.cc:223:58: error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
  223 |   params[0] = OSSL_PARAM_construct_utf8_string("digest", kEvpDigest, 0);
      |                                                          ^~~~~~~~~~
      |                                                          |
      |                                                          const char*
In file included from /usr/local/google/home/gregorycooke/openssl3-install/include/openssl/evp.h:32,
                 from /usr/local/google/home/gregorycooke/repos/grpc/src/core/tsi/alts/crypt/aes_gcm.cc:26:
/usr/local/google/home/gregorycooke/openssl3-install/include/openssl/params.h:88:68: note:   initializing argument 2 of ‘OSSL_PARAM OSSL_PARAM_construct_utf8_string(const char*, char*, size_t)’
   88 | OSSL_PARAM OSSL_PARAM_construct_utf8_string(const char *key, char *buf,
      |                                                              ~~~~~~^~~
```

src/core/tsi/alts/crypt/aes_gcm.cc
matthewstevenson88: Ah sorry I didn't expand the file enough.

src/core/lib/security/credentials/jwt/jwt_verifier.cc
```diff
@@ -563,9 +575,38 @@ static EVP_PKEY* pkey_from_jwk(const Json& json, const char* kty) {
   tmp_e = nullptr;
   result = EVP_PKEY_new();
   EVP_PKEY_set1_RSA(result, rsa);  // uprefs rsa.
+#else
+
+  if (!OSSL_PARAM_BLD_push_BN(bld, "n", tmp_n) ||
+      !OSSL_PARAM_BLD_push_BN(bld, "e", tmp_e) ||
+      (params = OSSL_PARAM_BLD_to_param(bld)) == NULL) {
+    gpr_log(GPR_ERROR, "Could not create OSSL_PARAM");
+    OSSL_PARAM_BLD_free(bld);
```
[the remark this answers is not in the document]
gtcooke94: Good catch! Moved to 607, I think it's cleaner that way

test/core/end2end/h2_ssl_cert_test.cc
```diff
@@ -260,12 +260,13 @@ TEST_P(H2SslCertTest, SimpleRequestBody) {
 
 #ifndef OPENSSL_IS_BORINGSSL
 #if GPR_LINUX
-TEST_P(H2SslCertTest, SimpleRequestBodyUseEngine) {
-  test_server1_key_id.clear();
-  test_server1_key_id.append("engine:libengine_passthrough:");
-  test_server1_key_id.append(test_server1_key);
-  simple_request_body(fixture_.get(), GetParam().result);
-}
+// TODO(gtcooke94) currently failing on OpenSSL3.0 and 1.1.1
```
matthewstevenson88, line 263: Please delete the commented-out code and only leave the TODO.

PR #31263, opened 2022-10-06
Update rules_python and local_config_python

kchygoe:
This PR updates rules_python to v0.13.0 and use `system_python` function from `@upb` instead of old `python_configure` function.  This makes the python treatment in bazel consistent with upb and protobuf repos.


gnossen:
I'm fairly certain this breaks our build. We need the `python_configure` function for `Python.h` to build our C extension, which `rules_python` does not provide. We do want to replace `python_configure` with a toolchain, but this is going to take much more work, including upstream changes in `rules_python`.

gnossen:
Right:

```
ERROR: /root/.cache/bazel/_bazel_root/7bb51ef4c4ce2ac55dce6a30e338cda8/external/com_github_grpc_grpc/src/python/grpcio/grpc/_cython/BUILD.bazel:25:12: no such package '@local_config_python//': The repository '@local_config_python' could not be resolved: Repository '@local_config_python' is not defined and referenced by '@com_github_grpc_grpc//src/python/grpcio/grpc/_cython:cygrpc.so'
```

kchygoe:
@gnossen Thank you for the reply! I checked  `system_python(name = "local_config_python")` which is from `@upb` in `WORKSPACE` working for me, but let me check the details. I'll make an update soon.

PR #31271, opened 2022-10-06
WIP: GRPC_ERROR_BUILDER

veblush:
TBD

### DIFF
--- a/src/core/ext/filters/channel_idle/channel_idle_filter.cc
+++ b/src/core/ext/filters/channel_idle/channel_idle_filter.cc
@@ -186,9 +186,10 @@ void MaxAgeFilter::PostInit() {
               auto fn = [](void* arg, grpc_error_handle) {
                 auto* channel_stack = static_cast<grpc_channel_stack*>(arg);
                 grpc_transport_op* op = grpc_make_transport_op(nullptr);
-                op->goaway_error = grpc_error_set_int(
-                    GRPC_ERROR_CREATE("max_age"),
-                    StatusIntProperty::kHttp2Error, GRPC_HTTP2_NO_ERROR);
+                op->goaway_error = GRPC_ERROR_BUILDER(kUnknown, "max_age")
+                                       .Set(StatusIntProperty::kHttp2Error,
+                                            GRPC_HTTP2_NO_ERROR)
+                                       .build();
                 grpc_channel_element* elem =
                     grpc_channel_stack_element(channel_stack, 0);
                 elem->filter->start_transport_op(elem, op);
@@ -277,9 +278,10 @@ void ChannelIdleFilter::StartIdleTimer() {
 
 void ChannelIdleFilter::CloseChannel() {
   auto* op = grpc_make_transport_op(nullptr);
-  op->disconnect_with_error = grpc_error_set_int(
-      GRPC_ERROR_CREATE("enter idle"),
-      StatusIntProperty::ChannelConnectivityState, GRPC_CHANNEL_IDLE);
+  op->disconnect_with_error =
+      GRPC_ERROR_BUILDER(kUnknown, "enter idle")
+          .Set(StatusIntProperty::ChannelConnectivityState, GRPC_CHANNEL_IDLE)
+          .build();
   // Pass the transport op down to the channel stack.
   auto* elem = grpc_channel_stack_element(channel_stack_, 0);
   elem->filter->start_transport_op(elem, op);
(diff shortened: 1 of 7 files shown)


src/core/ext/filters/channel_idle/channel_idle_filter.cc
```diff
@@ -277,9 +278,10 @@ void ChannelIdleFilter::StartIdleTimer() {
 
 void ChannelIdleFilter::CloseChannel() {
   auto* op = grpc_make_transport_op(nullptr);
-  op->disconnect_with_error = grpc_error_set_int(
-      GRPC_ERROR_CREATE("enter idle"),
-      StatusIntProperty::ChannelConnectivityState, GRPC_CHANNEL_IDLE);
+  op->disconnect_with_error =
+      GRPC_ERROR_BUILDER(kUnknown, "enter idle")
```
ctiller, line 282: Suggest `ErrorBuilder()`

src/core/ext/filters/channel_idle/channel_idle_filter.cc
```diff
@@ -277,9 +278,10 @@ void ChannelIdleFilter::StartIdleTimer() {
...
+          .Set(StatusIntProperty::ChannelConnectivityState, GRPC_CHANNEL_IDLE)
```
ctiller, line 283: Since the list of properties is known, I'd suggest:
`.WithChannelConnectivityState(GRPC_CHANNEL_IDLE)`

src/core/ext/filters/channel_idle/channel_idle_filter.cc
```diff
@@ -277,9 +278,10 @@ void ChannelIdleFilter::StartIdleTimer() {
...
+          .build();
```
ctiller, line 284: `.Build()`

ctiller (commented):
This probably needs a design... I'd actually like us to move to a place where we centralize error creation in one place instead of inlining a bunch of things everywhere, so that we'd have:

```
op->goaway_error = MaxAgeExpiryError();
```

and the actual definition of that in an `errors.cc` file elsewhere.

src/core/ext/filters/channel_idle/channel_idle_filter.cc
veblush: If you suggest to not use a macro, it's not feasible without `absl::SourceLocation` which can fetch this information without relying on macro.

src/core/ext/filters/channel_idle/channel_idle_filter.cc
veblush: Is this to reduce the code length?

src/core/ext/filters/channel_idle/channel_idle_filter.cc
veblush: Yep. Also we can consider to have an implicit converter to absl::Status to be able to omit Build() function call.

veblush:
@ctiller Yep, I'm thinking of having a design review and this is more like a scratch pad to do some experiment. I expected you might check this out when creating this and you did ;)

I'd like to be cautious of having all error creation logic in a single place because it may become a giant ball with a big nose. It can have some helper functions used often like `wsa_os_error` but this might be too much to like xDS related error functions. Also I'd like to move towards having a simple error instead of having many attributes. (e.g. Status(NOT_FOUND, "abc file is not found") instead Status(NOT FOUND).Set(filename, "abc"))

src/core/ext/filters/channel_idle/channel_idle_filter.cc
ctiller: That's spelled `grpc_core::SourceLocation` and we have it.

src/core/ext/filters/channel_idle/channel_idle_filter.cc
ctiller: also to improve readability... there's a lot to read here and most of it's noise that doesn't help illustrate the intent of the code

src/core/ext/filters/channel_idle/channel_idle_filter.cc
veblush: Oh that's amazing. But this doesn't seem to work on all platforms gRPC supports. Two things to do;
 1. Extending the check to use __builtin_LINE. (e.g. gcc 6 has it but it doesn't has `__has_builtin` function so we can allow it if gcc >= 6 even when there is no `__has_builtin`)
 2. Having a consensus on whether this source_location information can be omitted in certain environment such as msvc 2017.

In case of grpc_error, I guess error location isn't that critical so I'm fine with optional absence of it.

PR #31272, opened 2022-10-06
xDS: NACK EDS if sum of locality weights in a priority exceeds uint32 max

markdroth:


src/core/ext/xds/xds_endpoint.cc
```diff
@@ -366,6 +367,19 @@ absl::StatusOr<XdsEndpointResource> EdsResourceParse(
       const auto& priority = eds_resource.priorities[i];
       if (priority.localities.empty()) {
         errors.AddError(absl::StrCat("priority ", i, " empty"));
+      } else {
+        // Check that the sum of the locality weights in this priority
+        // does not exceed the max value for a uint32.
+        uint64_t total_weight = 0;
+        for (const auto& p : priority.localities) {
+          total_weight += p.second.lb_weight;
+          if (total_weight > std::numeric_limits<uint32_t>::max()) {
```
yashykt, line 376: Is UINT32_MAX an acceptable value?

yashykt (approved):

src/core/ext/xds/xds_endpoint.cc
markdroth: I'm not sure exactly what you're asking here.

If you're asking why we use `std::numeric_limits<uint32_t>::max()` instead of `UINT32_MAX`, the former is the idiomatic C++ way of spelling the latter, which is a C-style approach.

If you're asking why this is `>` instead of `>=`, it's intentional.  The uint32 max value is allowed here; it's only a problem if it exceeds that value.

If you're asking something else, then please clarify.  Thanks!

PR #31273, opened 2022-10-07
Add grpc-java 1.44.2 to client_matrix.py

sergiitk:


sergiitk:
Adhoc run: https://source.cloud.google.com/results/invocations/49aa383c-eb37-4226-a289-6a2d420b8296

ejona86 (approved):
This is fine by itself, but we should probably do a batched update for the rest.

Issue #30015, opened 2022-06-15
How about adding a config for round_robin to refresh dns ip.


wqh17101:
### Is your feature request related to a problem? Please describe.
I am using grpc python client with config "round_robin" as a long-connection channel ,it will discover the ips of a node when the channel creating and build the list of subchannels with these ips. However i find that it will only trigger dns resolver once at the time of starting,it means that when i expand the server by adding a new node, i will never find the new one.
For example:
```
my server has 3 nodes -- ipA,ipB,ipB.
so i build the channel with config "round_robin".And it will build three subchannels to these ips.
Then i add a new node to my server ipD.So there is four nodes for my server.
But the channel i built before will not find the new node.
```
### Describe the solution you'd like
Add a config for client to trigger dns resolver and  refresh the ip list by some time period.

### Describe alternatives you've considered
Write a round_robin class by myself,but i do not know how to inheritance your round_robin and configure it  when the channel built.

### Additional context
Add any other context about the feature request here.

wqh17101:
Also, if i set  `grpc.max_connection_age_ms=1000` on server, the client channel will fail on all subchannels after a success response and will not update ip to continue to  work. It is not robust.

wqh17101:
@nguyenbaviet I write one by myself, but i hope grpc-python can offer this ability officially. 
My implementation is that writing a dns relover by socket lib and start  a thread to call dns resolver by loop, when it found ip_set changed, rebuild the channel.

wqh17101:
@lidizheng What do you think of this problem

PR #31275, opened 2022-10-07
Add a channel manager for users to resolve dns through polling

wqh17101:
To resolve the problem in #30015 

example:
```
class MyChannelPool(ChannelPool):
    def init_channel(self, host, port):
        return grpc.insecure_channel(f'{host}:{port}')

pool = MyChannelPool("localhost", "50052", 3)
cm = ChannelManager(channel_pool=pool)
channel = cm.channel_pool[0]
```
everyone can define their own channel init function,and manage it by ChannelManager for resolving dns through polling and refreshing the channel when the ip_set changed.

### DIFF
--- a/src/python/grpcio/grpc/_channel.py
+++ b/src/python/grpcio/grpc/_channel.py
@@ -1595,17 +1595,27 @@ class ThreadingMixIn(threading.Thread):
     """
 
     def __init__(self):
-        super(ThreadingMixIn, self).__init__()
+        super().__init__()
         self._event = Event()
-        self._worker_thread: threading.Thread = threading.Thread()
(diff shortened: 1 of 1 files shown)


wqh17101:
@lidizheng Anyone can help me ?

gnossen:
Thanks for the contribution! I think this is a great addition to live outside of the gRPC library itself. Long term, we have plans to supply a more flexible [name resolution plugin system similar to the one already offered by Go and Java](https://grpc.io/docs/guides/custom-name-resolution/), but that will require some lower level integrations with the C++ layer rather than something that purely wraps our existing public API.

PR #31293, opened 2022-10-07
xDS: change remaining registries to be non-global

markdroth:


markdroth:
Moving this back to a draft -- looks like I need to first restructure some of the tests that assume the HTTP filter registry is global.

ctiller (approved):

PR #31294, opened 2022-10-07
Add a BUILD file in src/core

yashykt:
This needs to be cherry-picked


src/core/BUILD
```diff
@@ -0,0 +1,43 @@
...
+
+package(
+    default_visibility = ["//visibility:public"],
+    features = [
+        "layering_check",
+        "-parse_headers",
+    ],
+)
+
+grpc_cc_library(
```
ctiller, line 37: Can we do `event_engine_common` before getting too much further... it's got a dependency on `include/` that I think is going to inform the rest of the decisions we might make on this.

yashykt: Thanks for the suggestion!

veblush (approved):

src/core/BUILD
```diff
@@ -0,0 +1,65 @@
...
+
+load(
+    "//bazel:grpc_build_system.bzl",
+    "grpc_cc_library",
+)
+
+licenses(["reciprocal"])
+
+package(
+    default_visibility = ["//visibility:public"],
```
ctiller, line 23: I think this is inadvisable.

src/core/BUILD
```diff
@@ -0,0 +1,65 @@
...
+    features = [
+        "layering_check",
+    ],
+)
+
+exports_files(glob(["**"]))
```
ctiller, line 29: Can you say what this is for?
What visibility should it have?
Is there a chance to remove it?

tools/distrib/fix_build_deps.py
```diff
@@ -498,7 +507,10 @@ def best(self, scorer):
             choices = new_choices
 
         best = None
-        final_scorer = lambda x: (total_avoidness(x), scorer(x), total_score(x))
+
+        def final_scorer(x):
```
ctiller, line 511: Why can't this be a lambda?

tools/distrib/fix_build_deps.py
```diff
@@ -296,15 +296,22 @@ def grpc_cc_library(name,
         proto_hdr = '%s%s' % ((parsing_path + '/' if parsing_path else ''),
                               proto.replace('.proto', '.pb.h'))
         skip_headers[name].add(proto_hdr)
+
+    # Convert the source or header target to a relative path.
+    def get_filename(name, parsing_path):
```
ctiller, line 301: functions should be at global scope, not inlined

ctiller (commented):
Some questions about visibility and plans for such before this is submitted.

tools/distrib/fix_build_deps.py
yashykt: blame yapf :P

tools/distrib/fix_build_deps.py
yashykt: Done

src/core/BUILD
yashykt: Thanks!

src/core/BUILD
yashykt: Good point, Thanks!

tools/distrib/fix_build_deps.py
yashykt: It was actually my vscode formatter. Reverted.

ctiller (approved):

Issue #28597, opened 2022-01-18
C# protoc plugin: deprecated option in proto definitions should add Obsolete attribute to generated c# code


mrt181:
### Is your feature request related to a problem? Please describe.
Adding the `Obsolete` attribute to an rpc method that is overriden warns that the base method is non-obselete:
`Obsolete member 'SomeService.SomeMethod(SomeRequest, ServerCallContext)' overrides non-obsolete member 'SomeService.SomeServiceBase.SomeMethod(SomeRequest, ServerCallContext)' [SomeService]csharp(CS0809)`

The proto definitions look like this: 
```proto
service SomeService {
  rpc SomeMethod (SomeRequest) returns (SomeReply) {
    option deprecated = true;
  }
```

This is ignored in the generated code.
```cs
      [global::System.CodeDom.Compiler.GeneratedCode("grpc_csharp_plugin", null)]
      public virtual global::System.Threading.Tasks.Task<global::GrpcSomeService.SomeReply> SomeMethod(global::GrpcSomeService.SomeRequest request, grpc::ServerCallContext context)
```

It also does not work on the enum level:
```proto
enum SomeEnum {
  option deprecated = true;
  A = 0 [deprecated = true];
}
```
No attribute is added: 
```cs
#region Enums
  public enum SomeEnum {
    [pbr::OriginalName("A")] A = 0,
}
```

It works on the message level:
```proto
message SomeRequest {
  option deprecated = true;
  SomeEnum somefield = 1;
}
```
Which adds the `Obsolete` attribute:
```cs
  [global::System.ObsoleteAttribute]
  public sealed partial class SomeRequest : pb::IMessage<SomeRequest>
```

### Describe the solution you'd like
Add the Obsolete attribute on the generated code when the deprecated option is set in the proto definition.

### Describe alternatives you've considered
ignore warnings and use additional documentation and communication.

jtattermusch:
Since the issue description mentions deprecated option for Enum values, see also https://github.com/protocolbuffers/protobuf/issues/10513

tonydnewell:
For info: I've looked at the different places that the `deprecated` option can be used in `proto` files and compared the code generated by C# and Java. In all cases Java generates code with the correct `@Deprecated` attribute but the C# generator only currently supports a few cases:

- file level - **not supported**
- message - supported - `[global::System.ObsoleteAttribute]` added to the class
- field - supported - `[global::System.ObsoleteAttribute]` added to the class that wraps access to the field
[...]

PR #31299, opened 2022-10-09
C# protoc plugin: Annotate rpc methods which are deprecated with the [Obsolete] attribute.

chuchu:
This is my fix for #28597 which is currently assigned to @jtattermusch.

Annotate rpc methods which are deprecated with the obsolete attribute.
Suppress usage warning for this usage in the generated code itself.

For the pragma I did not create an extra method because in one occurrence the line break is in the beginning and in the other at the end. So one single method cannot fit both.

I was not able to execute the test cases. Nearly all tests are failing with and without my change. I do not know why.
I could also not find tests for this class so I did not wrote any new ones.

### DIFF
--- a/src/compiler/csharp_generator.cc
+++ b/src/compiler/csharp_generator.cc
@@ -110,6 +110,13 @@ bool GenerateDocCommentBody(grpc::protobuf::io::Printer* printer,
   return GenerateDocCommentBodyImpl(printer, location);
 }
 
+void GenerateObsoleteAttributeIfObsolete(grpc::protobuf::io::Printer* printer,
+                                           const MethodDescriptor* method) {
+  if(method->options().deprecated()) {
+    printer->Print("[global::System.Obsolete]\n");
+  }
+}
+
 void GenerateDocCommentServerMethod(grpc::protobuf::io::Printer* printer,
                                     const MethodDescriptor* method) {
   if (GenerateDocCommentBody(printer, method)) {
@@ -449,8 +456,11 @@ void GenerateServerClass(Printer* out, const ServiceDescriptor* service) {
   out->Indent();
   for (int i = 0; i < service->method_count(); i++) {
     const MethodDescriptor* method = service->method(i);
+
     GenerateDocCommentServerMethod(out, method);
     GenerateGeneratedCodeAttribute(out);
+    GenerateObsoleteAttributeIfObsolete(out, method);
+
     out->Print(
         "public virtual $returntype$ "
         "$methodname$($request$$response_stream_maybe$, "
@@ -529,6 +539,7 @@ void GenerateClientStub(Printer* out, const ServiceDescriptor* service) {
       // unary calls have an extra synchronous stub method
       GenerateDocCommentClientMethod(out, method, true, false);
       GenerateGeneratedCodeAttribute(out);
+      GenerateObsoleteAttributeIfObsolete(out, method);
       out->Print(
           "public virtual $response$ $methodname$($request$ request, "
           "grpc::Metadata "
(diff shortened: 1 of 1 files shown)


jtattermusch:
Can you please provide an example of before-and-after generated C# code?

src/compiler/csharp_generator.cc
```diff
@@ -110,6 +110,13 @@ bool GenerateDocCommentBody(grpc::protobuf::io::Printer* printer,
   return GenerateDocCommentBodyImpl(printer, location);
 }
 
+void GenerateObsoleteAttributeIfObsolete(grpc::protobuf::io::Printer* printer,
+                                         const MethodDescriptor* method) {
+  if (method->options().deprecated()) {
+    printer->Print("[global::System.Obsolete]\n");
```
jtattermusch, line 116: I remember there was a problem with the .NET Obsolete attribute that it will break existing users' builds unless it is silenced, so adding the Obsolete attribute to the generated code is as a matter of fact a breaking change (and we try to avoid breaking API changes as much as possible)
So while adding [Obsolete] atrribute for "deprecated" elements may seem to make sense, I not clear to me whether this is actually the right thing to do.

chuchu: Hi Jan,

thanks for you comment! When a method, which is annotated with the obsolete attribute is referenced, a compiler warning is generated. That is why I have disabled the warning in the generated code. We should should not generate warnings without an actual usage. When the project is configured to treat warnings as errors the build fails. So you are right. This is a breaking change. On the other hand, when the server removes the deprecated method eventually the build is broken as well.

What do you think about an top level option in the proto file to generate the C# obsolete attribute? Or some other option which could be used on the command line. This could give a bit more control.

BTW are there any tests which verify that a given proto produces a given C# file?

Here is an example. Given this proto snippet:

```protobuf
rpc SayHelloOld (HelloRequest) returns (HelloReply) {
  option deprecated = true;
}
```
Will produce this C# snippets:

```csharp
[global::System.Obsolete]
public virtual global::System.Threading.Tasks.Task<global::Helloworld.HelloReply> SayHelloOld(global::Helloworld.HelloRequest request, grpc::ServerCallContext context)
{
  throw new grpc::RpcException(new grpc::Status(grpc::StatusCode.Unimplemented, ""));
}
```
```csharp
public static grpc::ServerServiceDefinition BindService(GreeterBase serviceImpl)
{
  return grpc::ServerServiceDefinition.CreateBuilder()
    #pragma warning disable CS0612
    .AddMethod(__Method_SayHelloOld, serviceImpl.SayHelloOld)
    #pragma warning restore CS0612
    .AddMethod(__Method_SayHelloNew, serviceImpl.SayHelloNew).Build();
}
```
The complete example is here.
[sample.tar.gz](https://github.com/grpc/grpc/files/9789132/sample.tar.gz)

Best regards,
Chris

chuchu:
What is the problem with the "Interop ALTS Tests (internal CI)" test?
It doesn't look related to my changes. But you never know. :-)

src/compiler/csharp_generator.cc
chuchu: Hi Jan.

Thanks for your reply! 

It looks like the deprecation of rpc methods is in some grey area. The proto3 language guide contains only an option to deprecate field options. It recommends that languages should create compiler warnings when such a field is used. For the deprecation of messages I can only find some older [issues](https://github.com/protocolbuffers/protobuf/issues/1734) discussing it.

I tested it a bit with different languages and the most languages do not support it. Java and Golang do. C++, C#, Ruby and Python do not.

The documentation and implementation of the deprecation option is inadequate. This leads to confusion because Java devs use it and are wondering why C# devs ignore it.

I like your approach of versioning APIs using namespaces. Maybe because of the existence of the deprecation option it is done in a different way.

Whatever approach we choose it should be consistent. The deprecation option is not and it will be discussed again and again. Either we implement it from end to end or we get rid of it. But that would be not backward compatible.

chuchu:
Topic is stale. Closing the request.

PR #31304, opened 2022-10-10
fix broken objc test jobs

jtattermusch:
There's currently a CBF for objc jobs:
```
++ LANG=en_US.UTF-8
++ gem install cocoapods --version 1.7.2 --no-document --user-install
WARNING:  You don't have /Users/kbuilder/.gem/ruby/2.6.0/bin in your PATH,
	  gem executables will not run.
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.6.0/wrappers directory.
```


jtattermusch:
Confirmed with an adhoc run that this makes the objC tests green again.

dennycd (approved):
LGTM

PR #31312, opened 2022-10-11
Remove `include/grpcpp/impl/codegen/channel_interface.h`

ralphchung:
This PR is a part of [b/196639718](http://b/196639718). It tries to move the content of `include/grpcpp/impl/codegen/channel_interface.h` to `include/impl/channel_interface.h`. We do not remove anything to achieve backward compatibility. The file `include/grpcpp/impl/codegen/channel_interface.h` should be removed in the future.


ctiller:
I think it might make sense to move this one into include/grpcpp/channel_interface.h, and not in the impl directory, since it's exposed in the top level interface and pretty necessary for mocking/fakes... wdyt @markdroth

markdroth:
Why do we need the `impl` directory in the first place?  Do we have established criteria for determining what would go in `impl` vs in the top-level directory?

ctiller (approved):

PR #31324, opened 2022-10-11
Update visibility for `grpc_public_hdrs`

ralphchung:
In #31100, we removed `grpc_codegen` and replace it with `grpc_public_hdrs`. However, `grpc_public_hdrs` does not get updated to get the same visibility as `grpc_codegen`. This PR does the job.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -654,6 +654,7 @@ grpc_cc_library(
         "avoid_dep",
         "nofixdeps",
     ],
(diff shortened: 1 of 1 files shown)


ctiller (approved):

ctiller (changes requested):
chatting about this before approving

ralphchung:
Not needed anymore.

PR #31329, opened 2022-10-12
Remove `include/grpcpp/impl/codegen/config_protobuf.h`

ralphchung:
This PR is a part of [b/196639718](http://b/196639718). It tries to move the content of `include/grpcpp/impl/codegen/config_protobuf.h` to `include/config_protobuf.h`. We do not remove anything to achieve backward compatibility. The file `include/grpcpp/impl/codegen/config_protobuf.h` should be removed in the future.


ctiller (approved):
I think looks good... this one will definitely need a cherry-pick

ralphchung:
We may not need this change anymore since https://github.com/grpc/grpc/pull/31688 will remove `config_protobuf.h`.

PR #31333, opened 2022-10-12
Remove `include/grpcpp/impl/codegen/async_generic_service.h` completely

ralphchung:
Follow up for #30956. This PR removes `include/grpcpp/impl/codegen/async_generic_service.h` completely.

This PR requires cherry picking.


ralphchung:

PR #31337, opened 2022-10-12
Make .pc files require re2 and libcares

### LINKED ISSUE #31336: Shouldn't the pkg-config file list re2 and cares as dependencies?
The pkg-config files generated by the build script (grpc.pc, grpc++.pc) don't list re2 and libcares as dependencies, even though they are. The pkg-config files only add `-lre2` and `-lcares` to LDFLAGS through `Libs`. On systems where libre2 and libcares happen to be in the default library search path, this happens to work, but on systems where they're anywhere else, it breaks.

1.49.1

macOS 13

Apple clang version 14.0.0

### What did you do?

Ran `pkg-config --libs grpc`.

I expected to see `-L/opt/homebrew/Cellar/re2/20220601/lib` and `-L/opt/homebrew/Cellar/c-ares/1.18.1_1/lib` among the flags produced, since those flags are required to link against grpc's dependencies libcares and libre2 on my systems, as shown by `pkg-config --libs re2 libcares`.

### What did you see instead?

I saw an output which only contains `-lre2` and `-lcares` but not the necessary `-L` flags. This means linking with `cc myprogram.o $(pkg-config --libs grpc)` doesn't work.

mortie:
This fixes #31336.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23125,8 +23125,8 @@ generate_pkgconfig(
   "gRPC"
   "high performance general RPC framework"
   "${gRPC_CORE_VERSION}"
-  "gpr openssl absl_any_invocable absl_base absl_bind_front absl_cleanup absl_cord absl_core_headers absl_flat_hash_map absl_flat_hash_set absl_function_ref absl_hash absl_inlined_vector absl_memory absl_optional absl_random_random absl_span absl_status absl_statusor absl_str_format absl_strings absl_synchronization absl_time absl_type_traits absl_utility absl_variant"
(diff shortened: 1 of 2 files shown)


veblush:
Thanks for making this chance but you need to make a corresponding change on https://github.com/grpc/grpc/blob/master/templates/CMakeLists.txt.template.

mortie:
Done. I tried to make the change to the .template file such that it will generate the same CMakeLists.txt, but the generate_projects.sh script didn't work for me, so I wasn't able to verify it.

veblush:
Closed in favor of #31660.

PR #31338, opened 2022-10-12
xds: proto for control of fake control plane during interop testing

larry-safran:
Service definition for controlling a test control plane as specified in go/grpc-xds-control-plane-tests


larry-safran:
@ejona86 As mentioned, this will need to be added to grpc-proto, but at least can be reviewed.

larry-safran:
Abandoned project

PR #31339, opened 2022-10-12
GcpObservability: Add experimental public target

yashykt:

### DIFF
--- a/include/grpcpp/ext/gcp_observability.h
+++ b/include/grpcpp/ext/gcp_observability.h
@@ -14,12 +14,8 @@
 // limitations under the License.
 //
 
-#ifndef GRPC_INTERNAL_CPP_EXT_GCP_OBSERVABILITY_GCP_OBSERVABILITY_H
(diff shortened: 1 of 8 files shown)


include/grpcpp/ext/gcp_observability.h
```diff
@@ -14,12 +14,8 @@
 // limitations under the License.
 //
 
-#ifndef GRPC_INTERNAL_CPP_EXT_GCP_OBSERVABILITY_GCP_OBSERVABILITY_H
-#define GRPC_INTERNAL_CPP_EXT_GCP_OBSERVABILITY_GCP_OBSERVABILITY_H
-
-// TODO(yashykt): This file would have been in the top include/grpcpp directory
-// instead of in src/, but I'm not yet sure about the naming, so keeping it here
-// till we decide.
+#ifndef GRPCPP_EXT_GCP_OBSERVABILITY_H
```
yashykt, line 17: not sure whether we should have a gcp directory? @veblush @markdroth thoughts?

veblush: I think both would work but I prefer this one because we don't know yet how many extensions will come under gcp.

veblush (approved):

PR #31341, opened 2022-10-13
OpenCensus: Move measures, views and CensusContext to include file

yashykt:
Proposal in https://github.com/grpc/proposal/pull/334

Note that these are moved under the `grpc::experimental` namespace in the include file, but the original header files in `src` maintain `using` declarations in the `grpc` namespace for all of them so that, so that existing users of these APIs don't get broken.


src/cpp/ext/filters/census/grpc_plugin.cc
```diff
@@ -75,93 +75,103 @@ ::opencensus::trace::Span GetSpanFromServerContext(
...
+  return reinterpret_cast<const grpc::experimental::CensusContext*>(
+             context->census_context())
       ->Span();
 }
 
 // These measure definitions should be kept in sync across opencensus
 // implementations--see
 // https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcMeasureConstants.java.
-::opencensus::tags::TagKey ClientMethodTagKey() {
+::opencensus::tags::TagKey experimental::ClientMethodTagKey() {
```
markdroth, line 86: Instead of adding an `experimental::` prefix to each of these functions, let's just wrap them in a `namespace experimental { ... }` block.

markdroth (approved):

PR #31351, opened 2022-10-13
xDS Listener: use ValidationErrors and add unit test

markdroth:
Additional changes:
- Use `absl::variant<>` to differentiate between API listeners and TCP listeners.
- Use `absl::variant<>` to differentiate between an RDS reference and an inline RouteConfig.
- Use common xDS extension code for transport socket extensions in both Listener and Cluster resources.
- Improve xDS HTTP filter registration mechanism.
- Use `ValidationErrors` for HTTP filter config validation.


test/core/xds/xds_listener_resource_type_test.cc
```diff
@@ -0,0 +1,1962 @@
...
+}
+
+//
+// HttpConnectionManager tests
+//
+
+// These tests cover common behavior for both API listeners and TCP
+// listeners, so we run them in both contexts.
+class HttpConnectionManagerTest : public XdsListenerTest,
+                                  public ::testing::WithParamInterface<bool> {
```
yashykt, line 188: nit: It might be better to have a class/struct instead of a raw bool

yashykt (approved):

test/core/xds/xds_listener_resource_type_test.cc
markdroth: Done.

PR #31353, opened 2022-10-13
ruby: remove tests that try to assert a certain behavior after forking incorrectly

apolcyn:
This is a probable fix for b/242771103

Some tests were added a while back to assert a certain behavior (a certain exception) after forking incorrectly - i.e. after forking when grpc was already initialized in the parent process.

Such forking behavior is essentially undefined, though, and may be prone to bugs with locks being in a bad state, etc. So let's remove these tests.


veblush (approved):

apolcyn:
cc @ctiller please merge when you have a chance

PR #31358, opened 2022-10-14
Cleanup grpc_sample_fuzzers job

jtattermusch:
AFAICT the job doesn't publish any "github checks", so it has no need to pull the "grpc_checks_private_key".


ctiller:
Hi, we should finish off https://github.com/grpc/grpc/pull/30777 first, and there's absolutely no need for this to run on master (that's what ossfuzz/clusterfuzz are for)

jtattermusch:
PTAL

ctiller (approved):

PR #31361, opened 2022-10-14
core tests: absl::make_unique -> std::make_unique

markdroth:
Internally, see cl/480931577.


ctiller (approved):
looks like it needs an iwyu run (to clear out some absl headers i'll wager)

Issue #18111, opened 2019-02-20
[Flake] Interop ALTS java->go server time out


yang-g:
https://source.cloud.google.com/results/invocations/45c5d7af-84de-4ac3-8f27-ed0acb9b4c6d/targets/grpc%2Fcore%2Fpull_request%2Flinux%2Fgrpc_interop_alts/log
This looks very similar to #15434

```
2019-02-20 12:16:43,025 Running test ping_pong
Exception in thread "main" java.lang.AssertionError: Timed out waiting for response
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertNotNull(Assert.java:712)
	at io.grpc.testing.integration.AbstractInteropTest.pingPong(AbstractInteropTest.java:725)
	at io.grpc.testing.integration.TestServiceClient.runTest(TestServiceClient.java:269)
	at io.grpc.testing.integration.TestServiceClient.run(TestServiceClient.java:212)
	at io.grpc.testing.integration.TestServiceClient.main(TestServiceClient.java:70)
Shutting down
2019-02-20 12:16:43,025 FAILED: cloud_to_cloud:java:go_server:ping_pong:alts [ret=1, pid=2259, time=7.2sec]
```

gnossen:
Spotted again in the wild: https://source.cloud.google.com/results/invocations/c2359a46-d3d2-49a1-a5e8-9bdd5a24767b/log

jtattermusch:
based on go/grpc-flaky dashboard, I don't think we are seeing this anymore

dapengzhang0:
How flaky is the test? The timeout for this test is 5 seconds, so there is a very small chance that timeout could happen.

gnossen:
Two instances since May 14th that we've noticed. I don't have a number to put on it at the moment.

gnossen:
Another instance: https://source.cloud.google.com/results/invocations/6dc06a00-8d77-469c-979b-da6ace70fdd8/log

apolcyn:
another instance: https://source.cloud.google.com/results/invocations/29f20b8e-1031-490a-bc0d-edb3a12be523/targets/grpc%2Fcore%2Fpull_request%2Flinux%2Fgrpc_interop_alts/log

dapengzhang0:
@temawi would be the candidate to triage the issue.

drfloob:
[...]

PR #31368, opened 2022-10-15
Add min/max tls version to tls creds options core and C++ APIs.

rockspore:
Plumb the APIs to core and C++.


rockspore:
@ZhenLian How may I assign this to you for a first look? Thanks.

rockspore:
Interop ALTS Tests #18111
Portability Tests Windows #31401
Neither of these seems related to this PR.

test/core/end2end/fixtures/h2_tls_common.h
```diff
@@ -238,8 +238,8 @@ inline grpc_channel_credentials* create_tls_channel_credentials(
 inline grpc_server_credentials* create_tls_server_credentials(
     fullstack_secure_fixture_data* ffd) {
   grpc_tls_credentials_options* options = grpc_tls_credentials_options_create();
-  options->set_min_tls_version(ffd->tls_version);
-  options->set_max_tls_version(ffd->tls_version);
```
ZhenLian, line 242: The current h2 ssl/tls tests contain a lot of tests with similar code. Taking `h2_ssl_tls12.cc` and `h2_ssl_tls13.cc` for example, most of the code is same, except maybe just one line setting the TLS version.
Would you mind updating them so that they can be better structured?
You can do this in a separate PR, or combining that in this PR. Thank you!

include/grpcpp/security/tls_credentials_options.h
```diff
@@ -104,6 +104,17 @@ class TlsCredentialsOptions {
...
+  // handshake. If not set, the underlying SSL library will pick the version
+  // automatically.
+  // @param tls_version: The minimum TLS version.
+  void set_min_tls_version(grpc_tls_version tls_version);
+  // Sets the maximum TLS version that will be negotiated during the TLS
+  // handshake. If not set, the underlying SSL library will pick the version
+  // automatically.
+  // @param tls_version: The maximum TLS version.
+  void set_max_tls_version(grpc_tls_version tls_version);
+
```
ZhenLian, line 117: CPP need some unity tests and functional tests as well...

ZhenLian: For unity tests, you can probably put them under `grpc/test/cpp/server/credentials_test.cc` and `grpc/test/cpp/client/credentials_test.cc`. For e2e tests, see my comment below...

ZhenLian (changes requested):
Thanks Luwei for the change. This is looking pretty good!
My comments are mostly based on the tests. Feel free to let me know if you have any questions. Thank you!

test/cpp/end2end/tls_key_export_test.cc
```diff
@@ -163,6 +163,8 @@ class TlsKeyLoggingEnd2EndTest : public ::testing::TestWithParam<TestScenario> {
           GRPC_SSL_REQUEST_AND_REQUIRE_CLIENT_CERTIFICATE_AND_VERIFY);
       server_creds_options.watch_identity_key_cert_pairs();
       server_creds_options.watch_root_certs();
+      server_creds_options.set_min_tls_version(grpc_tls_version::TLS1_2);
```
ZhenLian, line 166: This can be a good example of e2e test, but it is testing some other functionality(TLS key exporting).
Could you please write another test similar to this one but testing the TLS version setting behavior? You can also move any shared code into a common place, similar to `h2_tls_common.h`

test/core/end2end/fixtures/h2_tls_common.h
rockspore: Of course not!

But I think those two files you named test the old SSL APIs while my current PR focuses on plumbing the new TLS APIs. Therefore, it's probably better to re-structure those in a separate PR, which I will work on after this one. Let me know if I am missing anything.

include/grpcpp/security/tls_credentials_options.h
rockspore: I added the unit tests there. PTAL. Thanks.

test/cpp/end2end/tls_key_export_test.cc
rockspore: I looked at several tests in this directory, it seems all tests construct their fixture classes independently. I followed the `grpc_auth_...` test and created a new `tls_creds_options_test.cc`. More tests could be added but for the purpose of this PR I put the tls version tests only. Let me know what you think.

test/core/end2end/fixtures/h2_tls_common.h
rockspore: By taking a closer look at all the fixtures, it seems to me that they fall into different categories, and code is largely duplicated within each of them. For example, the `h2_sockpair...` is one of them. My guess is they are intentionally broken into different test binaries so that they can be executed in parallel. This makes me wonder if a merge of tls12 and tls13 is needed. Let me know what you think. Thanks.

test/core/end2end/fixtures/h2_tls_common.h
```diff
@@ -238,8 +238,8 @@ inline grpc_channel_credentials* create_tls_channel_credentials(
...
+  grpc_tls_credentials_options_set_min_tls_version(options, ffd->tls_version);
+  grpc_tls_credentials_options_set_max_tls_version(options, ffd->tls_version);
```
ZhenLian, line 242: client set up needs to be changed into using C APIs as well(line ~220)

ZhenLian: Our original test sets the TLS version as a prerequisite: it assumes every tests should have the same TLS version settings and it should work properly. That's why it's inside the common function.
But now since we are testing the version setting behavior, maybe we could turn that into separate test cases and test some more scenarios? A couple of cases I could think of:

1. client[min 1.2, max 1.3] server [min 1.2, max1.3] -> supposed to work
2. client[min 1.3, max 1.3] server [min 1.3, max1.3] -> supposed to work
3. client[min 1.2, max 1.2] server [min 1.3, max1.3] -> what will happen in this case? I don't remember off hand but you can probably do a quick experiment here
4. client[min 1.3, max 1.3] server [min 1.2, max1.2] -> not sure as well
5. client/server[min 1.3, max 1.2] -> what will happen in this case? initialization failure or runtime failure?

To test these, you probably don't need fancy creds reloading behaviors so you can just go with StaticDataProvider.

ZhenLian (dismissed):
Thanks Luwei for the change! I posted some more ideas for the C core e2e tests.
I will take a look at the C++ e2e tests once we are all good with the core tests.
Let me know if you have any questions. Thanks!

test/core/end2end/fixtures/h2_tls_common.h
ZhenLian: I saw you opened another PR...was that intended to solve this particular comment?

test/core/end2end/fixtures/h2_tls_common.h
rockspore: I will address the rest later, but for tls version mismatches between server/client, I tested locally and it would cause handshake error. This could be tested but the rpc wouldn't be returned immediately and instead it would time out which result in a ~10 seconds unit test duration. Is this acceptable? Or is there other ways to terminate it sooner?

test/core/end2end/fixtures/h2_tls_common.h
rockspore: Yeah that PR was to reduce redundant code by refactoring common functions into h2_tls_common.h. As I pointed out, these tests are all for the old SSL APIs though, so I didn't include the change into this PR.

ZhenLian: I see, sure I will close this thread as you plan to do that in a separate PR

test/core/end2end/fixtures/h2_tls_common.h
ZhenLian: Hmmm I see. What exactly is the error message? It times out probably because it is re-trying the handshake. That's probably why core tests usually don't include failure cases. But we shall probably add a few these kind of tests. This was also discussed in another thread I had with @markdroth 
What I am think right now is to create a test without retry(probably some code in `h2_full_no_retry.cc` can be a reference), which hopefully will make things a bit easier. @markdroth Hi Mark, can I know what's the typical way for writing these e2e tests against failure behaviors? Do you have any other suggestions?

markdroth: I am assuming that we're using the default `pick_first` LB policy and that there is only one address returned by the resolver, since I think that should be the case in all of the core e2e tests.

If there is a handshake failure, then the connection attempt will fail, and `pick_first` will put the channel in `TRANSIENT_FAILURE`.  Any RPC sent on the channel will then fail by default.  However, if the RPC is using `wait_for_ready`, then it will remain queued until either (a) a subsequent connection attempt succeeds, or (b) the RPC's deadline expires.  The individual test can control whether it is using `wait_for_ready`, and it can expect the RPC to fail as appropriate.

Retries should not be an issue here.  If the test is not explicitly configuring retries, then there should not be any retries in this case.

ZhenLian: Thanks Mark for the reply!
If I understand correctly, in our case, the RPC should fail directly, as I didn't find any configurations for `wait_for_ready` in our TLS tests. That seems an unexpected behavior.
@rockspore let's do some investigations offline and see what is the cause of this particular problem. Feel free to let me know if you have any concerns. Thanks!

rockspore: Thanks Zhen and Mark for the details. I realize I was referring to the cpp end2end test I added when I mentioned the timeout. Sorry for misleading.

As for the core end2end test, it's actually not clear to me how to construct a failed test due to handshake errors by adding a fixture, since each fixture will be used to run all the test scenarios, which contain successful and failed RPCs. Looking at the definition of the `grpc_end2end_test_config`, I can't find any existing field which can inform the test that the fixture can cause issues when establishing the channel. Hypothetically, even if we added a new field there, all the test functions invoked in `end2end_tests.cc` would have to be modified to check it, which would be a sizable effort.

Let me know if there is some different approach that I overlooked here. Otherwise, I wonder if we could skip tls version mismatch tests in core end2end and only test it in the cpp ones. @ZhenLian

rockspore: Discussed with Zhen offline. I will need to add new fixtures as well as new tests with some specific `_test_options` in `generate_tests.bzl`. Will come back once I get that done.

rockspore: I added fixtures and test for the bad tls versions. For some reason, the corresponding test couldn't end itself till the 10 seconds timeout specified by at [test_config.cc](https://github.com/grpc/grpc/blob/7c5e3267981b5cdef015438040ad6d5e62ff4044/test/core/util/test_config.cc#L163). I have confirmed that `grpc_shutdown()` in the main function in `h2_tls_wrong_versions.cc` actually returned. To reproduce this, one can run:

```
bazel test //test/core/end2end:h2_tls_wrong_versions_test@status_unavailable --test_output=streamed
```

when this branch is checked out.

@ZhenLian PTAL when you get a chance. Thanks.

rockspore:
As @ZhenLian will not have enough bandwidth to further review my PRs in a while, @markdroth Could you PTAL when you get a chance (or please assign a different reviewer if you think they are better suited)? Thank you!

The APIs in this PR should be fairly simple. As mentioned in the comment above, such a test would invoke low-level TSI errors and result in test timeout before grpc could successfully shut down. Therefore I am currently unsure whether it is still worth adding such e2e tests in core to test wrong TLS versions. Please let me know what you think.

test/core/end2end/fixtures/h2_tls_wrong_versions.cc
```diff
@@ -0,0 +1,170 @@
...
+}
+
+static grpc_end2end_test_fixture chttp2_create_fixture_simple_fullstack_3(
+    const grpc_channel_args*, const grpc_channel_args*) {
+  grpc_end2end_test_fixture f;
+  int port = grpc_pick_unused_port_or_die();
+  fullstack_secure_fixture_data* ffd = new fullstack_secure_fixture_data();
+  memset(&f, 0, sizeof(f));
+  ffd->localaddr = grpc_core::JoinHostPort("localhost", port);
+  SetChannelTlsVersion(ffd, SecurityPrimitives::TlsVersion::V_13,
```
erm-g, line 80: Would it make sense to add minVersion <= maxVersion check on gRPC level? For example in src/core/lib/security/credentials/tls/grpc_tls_credentials_options.cc
`If max lvl already set, ASSERT   min_tls_version <= options->get max_tls_version
options->set_min_tls_version(min_tls_version);
`
My concern is if underlying behavior of OpenSSL changes we might introduce unpredictable behavior if we don't implement such chekc

rockspore: This is a good point but I don't think it should be the setter's job to check the versions. IIUC, tls_credentials_options is not immutable once created and can be consumed multiple times if needed. Consider the following use case:
```
options->set_min_tls_version(1.2);
options->set_max_tls_version(1.2);

// some work using the options without losing the ownership of the pointer.

// set tls versions differently for some other use.
options->set_min_tls_version(1.3); // now this would error if we check it here.
options->set_max_tls_version(1.3);
```
This admittedly is not a standard use pattern, since passing the options to the tls_credentials normally transfers the ownership of the options, but rejecting it seems weird to me.

In fact, tls_credentials_options has no functions like verify or validate. It would be better for the consumer to perform such a check, i.e., a check can be done when the options are passed into tls_credentials, either channel or server. In this case, I think it's better to be added in a separate PR.

erm-g: The snippet you described can be implemented like
`
options->set_min_tls_version(1.2);
options->set_max_tls_version(1.2);

// some work using the options without losing the ownership of the pointer.

// set tls versions differently for some other use.
options->set_min_tls_version(1.3); // Error here says 'Currently your min version is > your max version that can lead to unstable behavior of underlying SSL lib
options->set_max_tls_version(1.3);

//User changes it to

options->set_max_tls_version(1.3);
options->set_min_tls_version(1.3); 
`
But I agree that it's a better idea to perform such check at channel or server creation level.

rockspore: Thanks. In this case, I have removed those 2 fixtures where min/max versions are reversed, given that a subsequent task would be to add checks and require different tests.

For some reason, the timeout issue has disappeared when I test locally today. The test always finishes in 1.1 seconds. Are you able to reproduce it?

erm-g: I had an offline conversation with Mark to discuss the best practices for this repo. He suggested to do the check (and inform user in case of failed check) as early as possible. So I'm suggesting smth like 
`options->set_min_tls_version(1.2);
options->set_max_tls_version(1.2);

// some work using the options without losing the ownership of the pointer.

// set tls versions differently for some other use.
options->set_min_tls_version(1.3); // Error here says 'Currently your min version is > your max version that can lead to unstable behavior of underlying SSL lib
options->set_max_tls_version(1.3);

//User changes it to

options->set_max_tls_version(1.3);
options->set_min_tls_version(1.3);` 
@markdroth

rockspore: After some discussion with @erm-g, we all agree that a check is needed as early as possible. Instead of doing this at the setter, I found there already exists a [sanity check](https://github.com/grpc/grpc/blob/f76948b12d9e853cea19bbafaf098533f6907d93/src/core/lib/security/credentials/tls/tls_credentials.cc#L43) invoked right before the options are used to created the credentials. So I added a check there and some unit tests.

rockspore: There is some memory safety issue I need to handle. Will get back here when the PR is ready for a look again.

rockspore: All relevant tests should have passed. PTAL when you have time, @erm-g

erm-g (commented):
LGTM
@markdroth - we're done with internal reviews, could you take a look (or ask a member of your team to do it)?

rockspore:
@erm-g Could you approve the PR if it looks good to you? You can do so by going to the "Files changed" section and the option is under the "Review Changes" green button. Thank you!

erm-g (approved):
LGTM
@markdroth - we're done with internal reviews, could you take a look (or ask a member of your team to do it)?

rockspore:
Going to create a new one.

PR #31371, opened 2022-10-16
append TLS key entries into SSL key log file

fklassen:
This corrects an issue where a developer is unable to decrypt live TLS captures in Wireshark. Instead the user must take multiple steps to be able to view decrypted TLS application data, including saving a capture on disk and reloading the capture.

By appending rather than overwriting the TLS key log file entries, it is possible to configure Wireshark TLS (Pre)-Master-Secret log filename before the capture is started (pointing to and empty or existing file). Then, when a packet capture is started, decrypted data is available immediately. Further, if the capture is restarted, no keys are lost and no Wireshark reconfiguration is required.

Appending rather than overwriting TLS key files follows the technique used by openssl binary, Chrome and others. All executables are able to append to the same keylog file simultaneously, and Wireshark is able to decrypt all apps simultaneously.


Vignesh2208:
Thanks. triggered running the CI checks.

Vignesh2208 (approved):

drfloob (approved):

PR #31376, opened 2022-10-17
[C#] Move insecure compose validation into Channel

JamesNK:
Related to https://github.com/grpc/grpc-dotnet/pull/1802

I tested what happens when validation was removed from creating channel credentials with Grpc.Core. The good news is that "invalid" calls (i.e. insecure + call credentials) don't succeed. The bad is the failure error isn't helpful. A channel would throw `ArgumentNullException` because it would attempt to compose null native credentials.

This PR makes the error message easy to understand. It's the same message that `ChannelCredentials.Compose` threw previously.

@jtattermusch

### DIFF
--- a/src/csharp/Grpc.Core.Tests/ChannelTest.cs
+++ b/src/csharp/Grpc.Core.Tests/ChannelTest.cs
@@ -123,5 +123,16 @@ namespace Grpc.Core.Tests
             // check that Channel.ShutdownAsync has run
             Assert.AreEqual(ChannelState.Shutdown, channel.State);
         }
+
+        [Test]
+        public void CompositeCredentialsWithInsecureThrow()
+        {
+            var compositeCredentials = ChannelCredentials.Create(
+                ChannelCredentials.Insecure,
+                CallCredentials.FromInterceptor((context, metadata) => Task.CompletedTask));
+            var ex = Assert.Throws(typeof(InvalidOperationException), () => new Channel("localhost", compositeCredentials));
+            Assert.AreEqual("CallCredentials can't be composed with InsecureCredentials. " +
+                "CallCredentials must be used with secure channel credentials like SslCredentials.", ex.Message);
+        }
     }
 }
--- a/src/csharp/Grpc.Core.Tests/ChannelTest.cs
+++ b/src/csharp/Grpc.Core.Tests/ChannelTest.cs
@@ -129,7 +129,7 @@ namespace Grpc.Core.Tests
         {
             var compositeCredentials = ChannelCredentials.Create(
                 ChannelCredentials.Insecure,
-                CallCredentials.FromInterceptor((context, metadata) => Task.CompletedTask));
+                CallCredentials.FromInterceptor((context, metadata) => TaskUtils.CompletedTask));
             var ex = Assert.Throws(typeof(InvalidOperationException), () => new Channel("localhost", compositeCredentials));
             Assert.AreEqual("CallCredentials can't be composed with InsecureCredentials. " +
                 "CallCredentials must be used with secure channel credentials like SslCredentials.", ex.Message);
--- a/src/csharp/Grpc.Core/Channel.cs
+++ b/src/csharp/Grpc.Core/Channel.cs
@@ -69,20 +69,28 @@ namespace Grpc.Core
             EnsureUserAgentChannelOption(this.options);
             this.environment = GrpcEnvironment.AddRef();
 
-            this.completionQueue = this.environment.PickCompletionQueue();
-            using (var nativeChannelArgs = ChannelOptions.CreateChannelArgs(this.options.Values))
+            try
             {
-                var nativeCredentials = credentials.ToNativeCredentials();
-                if (nativeCredentials != null)
+                this.completionQueue = this.environment.PickCompletionQueue();
+                using (var nativeChannelArgs = ChannelOptions.CreateChannelArgs(this.options.Values))
                 {
-                    this.handle = ChannelSafeHandle.CreateSecure(nativeCredentials, target, nativeChannelArgs);
+                    var nativeCredentials = credentials.ToNativeCredentials();
+                    if (nativeCredentials != null)
+                    {
+                        this.handle = ChannelSafeHandle.CreateSecure(nativeCredentials, target, nativeChannelArgs);
+                    }
+                    else
+                    {
+                        this.handle = ChannelSafeHandle.CreateInsecure(target, nativeChannelArgs);
+                    }
                 }
-                else
-                {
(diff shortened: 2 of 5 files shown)


src/csharp/Grpc.Core.Tests/ChannelTest.cs
```diff
@@ -123,5 +123,16 @@ public async Task ChannelBaseShutdownAsyncInvokesShutdownAsync()
             // check that Channel.ShutdownAsync has run
             Assert.AreEqual(ChannelState.Shutdown, channel.State);
         }
+
+        [Test]
+        public void CompositeCredentialsWithInsecureThrow()
+        {
+            var compositeCredentials = ChannelCredentials.Create(
+                ChannelCredentials.Insecure,
+                CallCredentials.FromInterceptor((context, metadata) => Task.CompletedTask));
```
jtattermusch, line 132: `/var/local/git/grpc/src/csharp/Grpc.Core.Tests/ChannelTest.cs(132,77): error CS0117: 'Task' does not contain a definition for 'CompletedTask' [/var/local/git/grpc/src/csharp/Grpc.Core.Tests/Grpc.Core.Tests.csproj]`. This is due to the .NET45 target.

You can use https://github.com/grpc/grpc/blob/fc302e37f2c1ec1db112bafd26427d65301ad9dd/src/csharp/Grpc.Core/Utils/TaskUtils.cs#L32 instead.

jtattermusch (approved):
LGTM after doublechecking the tests are green (I'll go through the results once they finish).

JamesNK:
My environment is messy - I have tests failing on the base branch - but I see the new test fails if the debugger isn't attached. I don't know what that means.

```
2) Failed : Grpc.Core.Tests.ChannelTest.CompositeCredentialsWithInsecureThrow
  Expected: <System.InvalidOperationException>
  But was:  <System.AppDomainUnloadedException: Attempted to access an unloaded AppDomain.
   at Grpc.Core.Internal.CallCredentialsSafeHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalDispose()
   at Grpc.Core.Internal.DefaultChannelCredentialsConfigurator.CreateNativeCompositeCredentials(ChannelCredentials channelCredentials, CallCredentials callCredentials)
   at Grpc.Core.Internal.DefaultChannelCredentialsConfigurator.<>c__DisplayClass9_0.<SetCompositeCredentials>b__0()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at Grpc.Core.Internal.DefaultChannelCredentialsConfigurator.GetOrCreateNativeCredentials(ChannelCredentials key, Func`1 nativeCredentialsFactory)
   at Grpc.Core.Internal.DefaultChannelCredentialsConfigurator.SetCompositeCredentials(Object state, ChannelCredentials channelCredentials, CallCredentials callCredentials)
   at Grpc.Core.ChannelCredentials.CompositeChannelCredentials.InternalPopulateConfiguration(ChannelCredentialsConfiguratorBase configurator, Object state)
   at Grpc.Core.Internal.ChannelCredentialsExtensions.ToNativeCredentials(ChannelCredentials credentials)
   at Grpc.Core.Channel..ctor(String target, ChannelCredentials credentials, IEnumerable`1 options)
   at Grpc.Core.Channel..ctor(String target, ChannelCredentials credentials)
   at Grpc.Core.Tests.ChannelTest.<>c__DisplayClass11_0.<CompositeCredentialsWithInsecureThrow>b__1()
   at NUnit.Framework.Assert.Throws(IResolveConstraint expression, TestDelegate code, String message, Object[] args)>
   at Grpc.Core.Tests.ChannelTest.CompositeCredentialsWithInsecureThrow()
```

Changing the error to throw outside the using statement appears to fix the test on my machine. Hopefully, that fixes the hangs.

jtattermusch:
I look into the hanging tests and I found there was a bug in the Channel constructor:
- when channel is being created and credentials creation throws, we end up with no Channel instance (so we cannot do channel.Shutdown()), but the GrpcEnvironment.AddRef() has already been created. (and when the environment is not shutdown properly it can lead to hang, which is what made the ChannelTest got stuck IMHO).
- I added a fix for that.

@JamesNK please review my changes to this PR.

@tonydnewell  FYI

JamesNK (commented):
LGTM apart from the way the exception is rethrown.

src/csharp/Grpc.Core/Channel.cs
```diff
@@ -69,20 +69,28 @@ public Channel(string target, ChannelCredentials credentials, IEnumerable<Channe
...
+                    else
+                    {
+                        this.handle = ChannelSafeHandle.CreateInsecure(target, nativeChannelArgs);
+                    }
                 }
+                GrpcEnvironment.RegisterChannel(this);
+            }
+            catch (Exception e)
+            {
+                Task.Run(async () => await GrpcEnvironment.ReleaseAsync().ConfigureAwait(false));
```
JamesNK, line 91: `this.environment` from `GrpcEnvironment.AddRef()` isn't passed to `GrpcEnvironment.ReleaseAsync()`? Is it just reference counting?

src/csharp/Grpc.Core/Channel.cs
```diff
@@ -69,20 +69,28 @@ public Channel(string target, ChannelCredentials credentials, IEnumerable<Channe
...
+                throw e;
```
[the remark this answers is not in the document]
jtattermusch: of course, my bad. thanks for the fix!

src/csharp/Grpc.Core/Channel.cs
jtattermusch: yes, basically. GrpcEnvironment is a singleton with global state of grpc native (and e.g. owns the threadpool).

src/csharp/Grpc.Core/Channel.cs
```diff
@@ -69,20 +69,28 @@ public Channel(string target, ChannelCredentials credentials, IEnumerable<Channe
...
+                    else
+                    {
+                        this.handle = ChannelSafeHandle.CreateInsecure(target, nativeChannelArgs);
+                    }
                 }
+                GrpcEnvironment.RegisterChannel(this);
+            }
+            catch (Exception)
+            {
+                Task.Run(async () => await GrpcEnvironment.ReleaseAsync().ConfigureAwait(false));
```
JamesNK, line 91: nit: Discard task to ensure analyzers don't complain about unawaited task.
```suggestion
                // Constructor can't be async.
                _ = Task.Run(async () => await GrpcEnvironment.ReleaseAsync().ConfigureAwait(false));
```

apolcyn (approved):

jtattermusch:
test failures are unrelated.

jtattermusch:

PR #31382, opened 2022-10-17
C# generator - Add GeneratedCodeAttribute attribute to generated service class

mbalous:
`GeneratedCodeAttribute` is much easier to work with than some arbitrary region or xml-doc which are the measures used at the moment.

This improvement solves real world use case, if you check your codebase using Roslyn and try to filter out tool-generated code. It's quite clumsy to do it at the moment ( relying on regions or xml-doc ).

Using attribute `GeneratedCodeAttribute` is the usual way to mark such code in .NET world.

### DIFF
--- a/src/compiler/csharp_generator.cc
+++ b/src/compiler/csharp_generator.cc
@@ -23,6 +23,8 @@
 #include <sstream>
 #include <vector>
 
+#include <grpcpp/grpcpp.h>
+
 #include "src/compiler/config.h"
 #include "src/compiler/csharp_generator_helpers.h"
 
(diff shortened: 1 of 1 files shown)


src/compiler/csharp_generator.cc
```diff
@@ -23,6 +23,8 @@
 #include <sstream>
 #include <vector>
 
+#include <grpcpp/grpcpp.h>
```
jtattermusch, line 26: this is definitely unnecessary.

jtattermusch:
I can reopen if I hear back from the contributor.

PR #31392, opened 2022-10-18
[EventEngine] Refactor TimerManager to leverage a shared ThreadPool

drfloob:
This is a step towards a work-stealing thread pool implementation: unifying thread pools so that work stealing affects timer callbacks as well.

A subsequent step will expose the timeout logic so that the timer wakeup and check can be triggered externally (by pollers, in the common case).

### DIFF
--- a/src/core/lib/event_engine/posix_engine/timer_manager.h
+++ b/src/core/lib/event_engine/posix_engine/timer_manager.h
@@ -50,7 +50,7 @@ class TimerManager final : public grpc_event_engine::experimental::Forkable {
  public:
   explicit TimerManager(
       std::shared_ptr<grpc_event_engine::experimental::ThreadPool> thread_pool);
-  ~TimerManager();
+  ~TimerManager() override;
 
   grpc_core::Timestamp Now() { return host_.Now(); }
 


src/core/lib/event_engine/posix_engine/timer_manager.cc
```diff
@@ -155,54 +72,32 @@ void TimerManager::MainLoop() {
...
 bool TimerManager::IsTimerManagerThread() { return g_timer_thread; }
 
-TimerManager::TimerManager() : host_(this) {
+TimerManager::TimerManager(
+    std::shared_ptr<grpc_event_engine::experimental::ThreadPool> thread_pool)
+    : host_(this), thread_pool_(std::move(thread_pool)) {
   timer_list_ = std::make_unique<TimerList>(&host_);
-  grpc_core::MutexLock lock(&mu_);
-  StartThread();
+  main_thread_ = grpc_core::Thread(
```
ctiller, line 92: I think this should be `thread_pool->Run([this] { MainLoop(); });`

drfloob: It is intentionally not on the thread pool. There is now just one persistent TimerManager main loop. In the current incarnation, we do not want it executing closures, timer or otherwise, when work stealing is landed (presuming that happens before the timer manager's polling is integrated with the I/O polling thread, which I will do as soon as Vignesh's work on the poller https://github.com/grpc/grpc/pull/31128 is merged and stable).

Alternative solutions could exist here that let a TimerManager with multiple MainLoops run on the thread pool, with a fair amount of added complexity. This PR is an intermediate step that allows work stealing to influence timer callbacks. I don't necessarily think it's worth managing a pool within a pool to allow callbacks to execute in the same thread, especially when that code is going away very quickly.

src/core/lib/event_engine/posix_engine/timer_manager.h
```diff
@@ -44,30 +44,23 @@ namespace posix_engine {
 // all times, and thus effectively preventing the thundering herd problem.
 // TODO(ctiller): consider unifying this thread pool and the one in
 // thread_pool.{h,cc}.
-class TimerManager final : public grpc_event_engine::experimental::Forkable {
+class TimerManager final {
```
Vignesh2208, line 47: Could you elaborate on why this is no longer a Forkable ?

drfloob: This class no longer needs to do anything special on fork, which is to say that it is trivially forkable.

ctiller: How is that true if it starts a thread?

src/core/lib/event_engine/posix_engine/timer_manager.cc
ctiller: I wasn't suggesting a pool within a pool, just that we have one MainLoop running inside the thread pool, instead of firing up a thread. My intent there is to ensure we've resolved any memory management hazards here (of the form 'who owns who') before moving onto the next step.

No need to block on this though, useful discussion.

src/core/lib/event_engine/posix_engine/timer_manager.h
drfloob: Ah right, the fact that it's non-joinable doesn't help the forking problem. Thanks both for the reminder.

src/core/lib/event_engine/posix_engine/timer_manager.cc
drfloob: > No need to block on this though, useful discussion.

Agreed! Thanks. Updating this with our chat: the fact that the MainLoop is in a non-thread-pool thread makes it harder to have those shared_ptr cycles and shutdown issues. We'll have to consider them again when the MainLoop thread merges into the poller thread(s) on the thread pool.

src/core/lib/event_engine/posix_engine/timer_manager.h
drfloob: Fixed

ctiller (approved):

Vignesh2208 (commented):
lgtm

PR #31400, opened 2022-10-19
Introducing USDT DTrace Support

ydinkin-gcv:
Recent Linux versions have introduced the concept of eBPF USDT probes, which allows devs to instrument their libraries with tracepoints to be consumed by various eBPF tools.

This PR aims to:
1. Add the infrastructure for future USDT integration into gRPC, by integrating DTrace (the tool defining probe interfaces) into its build systems.
2. Introduce basic USDTs for gRPC's transport layer, allowing tracing of gRPC streams.

For now, DTrace support has been added as totally opt-in, only for cpp, with no runtime overhead if disabled and with minimal overhead when enabled but unused. In the future, we aim to continue improving USDT integration into gRPC to extend its usefulness and eventually turn it on by default. This will probably require additional testing infrastructure that's not part of this PR.

- In order to build grpc w/ USDT, add:
  - bazel: `--define GRPC_DTRACE=true`
  - CMake: `-DgRPC_DTRACE=ON`
  - make support was not implemented
- `dtrace` is invoked during build, and currently needs to be provided by the env
  - `apt install systemtap-sdt-dev` works for Ubuntu 20.04
- Here's [a quick bpftrace script](https://gist.github.com/ydinkin-gcv/6c5dab3a5336c21eeea797e9117d2be6) that traces gRPC streams, mainly for reference
  - Run it using `sudo ./bpftrace -p $(pidof <target process>) <path/to/grpc_transport_trace.bt>`
  - Note that. old bpftrace versions USDT support is flaky (did not work for me with `v0.9.4`, had to upgrade to `v0.16.0`)

I'd be happy to provide any additional context and motivation for the adding said probes :)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -22,6 +22,10 @@ load(
     "grpc_upb_proto_reflection_library",
     "python_config_settings",
 )
+load(
+    "//bazel:dtrace.bzl",
+    "dtrace_generate_provider",
+)
 load("@bazel_skylib//lib:selects.bzl", "selects")
 
 licenses(["reciprocal"])
@@ -146,6 +150,16 @@ config_setting(
 config_setting(
     name = "use_strict_warning",
     values = {"define": "use_strict_warning=true"},
+)
+
+config_setting(
+    name="grpc_dtrace",
+    values = {"define": "grpc_dtrace=true"},
+)
+
+dtrace_generate_provider(
+    name="grpc_dtrace_provider",
+    path="//src/core:ext/dtrace/dtrace_provider.d",
 )
 
 python_config_settings()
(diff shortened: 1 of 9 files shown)


veblush:
We may want to have a design review on this.

ctiller:
We definitely want a design review here: I'd like to see a doc describing context and motivation, and also what hooks are needed. @jtattermusch needs to be involved in testing strategy, and there probably needs to be a discussion of how thoroughly we want this tested enabled vs disabled. I'd like to involve security teams here too, since I don't think we're in a place to judge implications of this.

ctiller:
(is this something we'll want cross language support on? If so we will want to include @ejona86 and @dfawley and get a gRFC in place for)

markdroth:
I agree that we need to discuss whether this is something C-core-specific or something that we want cross-language.

I think we definitely want a gRFC for this, even if it's only going to be C-core specific.

ydinkin-gcv:
Hi,
Thank you all for your inputs :)
I'll put a gRFC together to have a more in-depth discussion.

PR #31364, opened 2022-10-14
Revert "upgrade rules_apple and apple_support bazel deps (#30896)"

veblush:
Revert https://github.com/grpc/grpc/pull/30844 on 1.50.x branch for 1.50.1 release

PR #31407, opened 2022-10-20
Partially revert "upgrade rules_apple and apple_support bazel deps"

jtattermusch:
This reverts commit 19fda811fb1f8e4260fb46dcc2ec451f4e41dad6 of https://github.com/grpc/grpc/pull/30896, but also upgrades rules_apple from 0.31.3 to 0.32.0.

It was already reverted on the 1.50.x branch: https://github.com/grpc/grpc/pull/31364

Since ObjC bazel tests are broken with python3.9 (which is what's available on kokoro monterey) with the original version of rules_apple 0.31.3, I'm also upgrading to 0.32.0 (which should still support bazel 4.x, but has the python3.9 fix: see https://github.com/bazelbuild/rules_apple/pull/1168/commits/51c4f83544496755da2eadc3ffcacd04d5b0efac  and https://github.com/google/xctestrunner/compare/64a9be0b6fa833b4b2371729c5c8cdd2c6f7775b..5a1b0c158f2debace60722ddbcb5035e3387810e)

I checked manually that `OVERRIDE_BAZEL_VERSION=4.2.2 tools/bazel build :grpc++` still passes with this change.

### DIFF
--- a/bazel/grpc_deps.bzl
+++ b/bazel/grpc_deps.bzl
@@ -386,10 +386,10 @@ def grpc_deps():
     if "build_bazel_rules_apple" not in native.existing_rules():
         http_archive(
             name = "build_bazel_rules_apple",
-            sha256 = "0052d452af7742c8f3a4e0929763388a66403de363775db7e90adecb2ba4944b",
+            sha256 = "77e8bf6fda706f420a55874ae6ee4df0c9d95da6c7838228b26910fc82eea5a2",
             urls = [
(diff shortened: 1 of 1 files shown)


jtattermusch:
[Objc Bazel Test](https://source.cloud.google.com/results/invocations/1f7561b1-6178-48a6-ae48-bceb64574c9c) is green!

gnossen (approved):

PR #31411, opened 2022-10-20
fix gprc_cli segfault when using binary in and out

MicahRam:
Fixes segfault when using `grpc_cli call --binary_input --binary_output`


eugeneo (approved):

drfloob:
Thanks for the fix, @MicahRam

drfloob (approved):

drfloob:
Rerunning UBSAN to ensure it was an infra flake.

PR #31413, opened 2022-10-20
Add sanity check for fully-qualified using statements

drfloob:
This adds a sanity check to ensure that `using` statements which expose a name in a new namespace will be fully-qualified. This only applies to `grpc.*` and `google.*` top level namespaces at the moment.

Using fully-qualified namespaces isn't always required, but per readability recommendations, it's a good practice.


ctiller:
Most of these are introducing names into the top level namespace, which never requires the `::` prefix - and I've not found a recommendation anywhere for introducing it. Since adding it makes code strictly less readable by adding line noise, can we have a reason cited for doing so?

drfloob:
The guidelines are internal to Google, but this code applies nonetheless. There's some explicit disagreement with your position on readability and helpfulness to the reader.

markdroth:
Can you point me at the readability guidelines that recommend this?  I'm not aware of any, and I don't quite see the benefit, especially when the `using` statement is already at the top level.

drfloob:
> Can you point me at the readability guidelines that recommend this? 

Please see my comment in https://github.com/grpc/grpc/pull/31413#issuecomment-1286099507. Responded out of band.

markdroth:
I think that this falls under the guidance that no one should ever declare their own `grpc` namespace, at any level.

drfloob:
> I think that this falls under the guidance that no one should ever declare their own grpc namespace, at any level.

That's fair. We can drop this change in that case, and assert the same should be true for the `google` and `grpc_*` namespaces in use, at any level.

PR #31418, opened 2022-10-20
xDS RouteConfig: use ValidationErrors and add unit test

markdroth:
Specific changes:
- Use `absl::variant<>` for `HashPolicy` types.
- Pull validation of resulting LB policy configs out of `ClusterSpecifierPlugin` registry and into `RouteConfig` validation.
- Don't ignore missing `Any` fields in HTTP filter configs if `is_optional` is true, since `is_optional` should apply only to unsupported filter types, not to malformed resource protos.
- Use `ExtractXdsExtension()` for `HttpConnectionManager` config itself.
- Minor API improvements to `StatusCodeSet` to make it a bit more useful in tests.
- Changed `StringMatcher` to return the underlying error message when a regex fails to compile.
- Fix a bug whereby a `ClusterSpecifierPlugin` would not be returned if it was not used in a vhost, even if it was used in another vhost.


yashykt:
can we move bug fixes to a separate PR? The only reason I ask that is it helps when eyeballing which PRs need to be potentially backported

markdroth:
I split the bug fix into #31583, which I plan to merge before this PR.

src/core/ext/xds/xds_route_config.cc
```diff
@@ -149,58 +149,53 @@ XdsRouteConfigResource::Route::RouteAction::HashPolicy::operator=(
...
-      if (other.regex != nullptr) return false;
-    } else {
-      if (other.regex == nullptr) return false;
-      return header_name == other.header_name &&
-             regex->pattern() == other.regex->pattern() &&
-             regex_substitution == other.regex_substitution;
-    }
+bool XdsRouteConfigResource::Route::RouteAction::HashPolicy::Header::operator==(
+    const Header& other) const {
+  if (header_name != other.header_name) return false;
```
yashykt, line 169: Let's maintain the same order as the declaration of the members - header_name, regex, regex_substitution even when comparing

yashykt (approved):

PR #31428, opened 2022-10-21
XdsClient: add tests for resource types where AllResourcesRequiredInSotW() returns true

markdroth:
Also fixes some bugs discovered while adding the test:
- We were not setting the state in the cache entry to `DOES_NOT_EXIST`, so a subsequently started watcher was not getting the proper notification.
- If the server sent an empty update because all of the subscribed resources have been removed, we failed to increment the resource version.


markdroth:
For context, this refers to the Listener and Cluster resource types, as described in https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#grouping-resources-into-responses.

yashykt:
Please check if these fixes are also things that need to be backported

src/core/ext/xds/xds_client.cc
```diff
@@ -1066,15 +1066,17 @@ void XdsClient::ChannelState::AdsCallState::OnRecvMessage(
...
                     resource_state.watchers);
               }
             }
           }
         }
       }
-      // If we had valid resources, update the version.
-      if (result.have_valid_resources) {
+      // If we had valid resources or the update was empty, update the version.
+      if (result.have_valid_resources || result.errors.empty()) {
```
yashykt, line 1079: is an empty update allowed?

yashykt (approved):

src/core/ext/xds/xds_client.cc
markdroth: Yes.  For the LDS and CDS resources in SotW, this will happen if every resource that the client is subscribed to is deleted on the server.

markdroth:
> Please check if these fixes are also things that need to be backported

I suspect these fixes don't require back-ports in and of themselves, but the changes in #31446 almost certainly will, so we can do these at the same time.

PR #25398, opened 2021-02-10
Fix races in resolve_address and resolve_address_posix tests

apolcyn:
This should fix internal test flakinesss bugs: b/179683282, b/171497698, and b/179083103

The polling loops in each of these tests currently exits when `args->done_atm` has been set. And this is set by `grpc_resolve_address` callbacks <i>before</i> acquiring `args->mu`. So there's a race where `pollset_work` can wake up due to timeout, exit, and then destroy `args->mu`, right before we try to acquire `args->mu` and kick the pollset.

PR #31443, opened 2022-10-24
c-ares DNS resolver: fix logical race between resolution timeout/cancellation and fd readability

apolcyn:
This PR adds a test originally meant to hit the bug described in https://github.com/grpc/grpc/pull/31426. This test caught a more general, long-standing bug which is fixed here:

If we:

1) Have more than one DNS query outstanding on the same file descriptor (for example, an `A` query and a `AAAA` query)
2) Cancel the overall resolution (for example due to the overall [timeout](https://github.com/grpc/grpc/blob/99aa924a5c7c813a3c0d647df34d10db1fec9cff/src/core/ext/filters/client_channel/resolver/dns/c_ares/grpc_ares_wrapper.cc#L298) kicking in)
3) Concurrently with 2), receive data from the DNS server for **one** of those queries (for example, just the `A` query)

... then, we can hit a bug where we run the `on_readable` callback for the `A` query and [process the data as if the resolution hasn't been cancelled](https://github.com/grpc/grpc/blob/99aa924a5c7c813a3c0d647df34d10db1fec9cff/src/core/ext/filters/client_channel/resolver/dns/c_ares/grpc_ares_wrapper.cc#L367). Even though we've already [shut down](https://github.com/grpc/grpc/blob/99aa924a5c7c813a3c0d647df34d10db1fec9cff/src/core/ext/filters/client_channel/resolver/dns/c_ares/grpc_ares_wrapper.cc#L249) the file descriptor with the still-pending `AAAA` query, the c-ares library thinks we still want to wait for this `AAAA` query to finish. But it never will finish, because we'll stop polling I/O at this point without ever having told c-ares to cancel the query.

The fix is to explicitly check for overall cancellation in the I/O callbacks and tell c-ares to cancel everything in that case.

---------

While we're here:

- fix the same race in `resolver_component_test.cc` that was fixed for other tests in https://github.com/grpc/grpc/pull/25398
- fix a trace log

### DIFF
--- a/test/cpp/naming/resolver_component_test.cc
+++ b/test/cpp/naming/resolver_component_test.cc
@@ -662,10 +662,10 @@ TEST(ResolverComponentTest, TestResolvesRelevantRecordsWithConcurrentFdStress) {
   socket_stress_thread.join();
 }
 
-TEST(ResolverComponentTest, TestResolvesRelevantRecordsWith1MsTimeout) {
+TEST(ResolverComponentTest, TestDoesntCrashOrHangWith1MsTimeout) {
   // Queries in this test could either complete successfully or time out
-  // show cancellation. This test doesn't care - we just care that the
-  // query completes and doesn't crash, leak, etc.
+  // and show cancellation. This test doesn't care - we just care that the
+  // query completes and doesn't crash, hang, leak, etc.
   RunResolvesRelevantRecordsTest(
       ResultHandler::Create,
       grpc_core::ChannelArgs().Set(GRPC_ARG_DNS_ARES_QUERY_TIMEOUT_MS, 1));
--- a/test/cpp/naming/resolver_component_test.cc
+++ b/test/cpp/naming/resolver_component_test.cc
@@ -192,9 +192,9 @@ gpr_timespec TestDeadline(void) {
 
 struct ArgsStruct {
   gpr_event ev;
-  gpr_atm done_atm;
   gpr_mu* mu;
-  grpc_pollset* pollset;
+  bool done;              // guarded by mu
+  grpc_pollset* pollset;  // guarded by mu
   grpc_pollset_set* pollset_set;
   std::shared_ptr<grpc_core::WorkSerializer> lock;
   grpc_channel_args* channel_args;
@@ -211,7 +211,7 @@ void ArgsInit(ArgsStruct* args) {
   args->pollset_set = grpc_pollset_set_create();
   grpc_pollset_set_add_pollset(args->pollset_set, args->pollset);
   args->lock = std::make_shared<grpc_core::WorkSerializer>();
-  gpr_atm_rel_store(&args->done_atm, 0);
+  args->done = false;
   args->channel_args = nullptr;
 }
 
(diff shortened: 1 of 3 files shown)


test/cpp/naming/resolver_component_test.cc
```diff
@@ -192,9 +192,9 @@ gpr_timespec TestDeadline(void) {
 
 struct ArgsStruct {
   gpr_event ev;
-  gpr_atm done_atm;
   gpr_mu* mu;
```
markdroth, line 195: Can we convert this to use `grpc_core::Mutex` instead, so that we can use lock annotations and allow the compiler to enforce the lock guard?

We can do the same thing in the tests that were fixed in #25398.

test/cpp/naming/resolver_component_test.cc
```diff
@@ -656,12 +655,22 @@ TEST(ResolverComponentTest, TestResolvesRelevantRecordsWithConcurrentFdStress) {
...
-  RunResolvesRelevantRecordsTest(ResultHandler::Create);
+  RunResolvesRelevantRecordsTest(ResultHandler::Create,
+                                 grpc_core::ChannelArgs());
   // Shutdown and join stress thread
   gpr_event_set(&done_ev, reinterpret_cast<void*>(1));
   socket_stress_thread.join();
 }
 
+TEST(ResolverComponentTest, TestDoesntCrashOrHangWith1MsTimeout) {
+  // Queries in this test could either complete successfully or time out
```
markdroth, line 666: I guess there's no way to trigger exactly the sequence of events that we care about here without having a much more flexible fake DNS server, right?

test/cpp/naming/resolver_component_test.cc
apolcyn: I'd like to do that, but we can't because this `gpr_mu` actually belongs to the pollset, so we'd need to migrate pollset to `grpc_core::Mutex` first.

test/cpp/naming/resolver_component_test.cc
apolcyn: Yeah, I think we'd actually need some kind of mocked endpoint to do that. While it would be nice to have a test like that, it's not obvious to me how we could structure something like that, given the way that the c-ares library interacts directly with file descriptors.

While the current test doesn't take a deterministic code path, it at least won't flake in a way that yields a false positive. And it's been useful so far (since it quickly surfaced this bug while I was trying to repro the separate issue of https://github.com/grpc/grpc/pull/31426, I think it's giving some previously-missing but important coverage).

test/cpp/naming/resolver_component_test.cc
markdroth: Ah, okay.  This is probably something that we'll be able to address as part of migrating this code to EventEngine.

test/cpp/naming/resolver_component_test.cc
markdroth: Maybe what we need is a fake c-ares implementation.  We could implement an interface that wraps around c-ares and have the DNS resolver code use c-ares via that interface.  We could then also implement a fake version of that interface, so that we can have it respond to the DNS resolver code however we want.

But in any case, I think this is a project for another day.

markdroth (approved):

test/cpp/naming/resolver_component_test.cc
apolcyn: > Maybe what we need is a fake c-ares implementation. We could implement an interface that wraps around c-ares and have the DNS resolver code use c-ares via that interface.

Perhaps. But IMO this code pretty tightly coupled to the behavior of the c-ares library, and I think most of the meat of this code is about properly handling the behavior of the c-ares library. So I'd be weary of mocking at the c-ares API layer for the risk that we'd just end up testing our mock.

While I'm not really sure how to do it at the moment, I'd still be a lot more excited about something that mocked out only the "untrusted" parts of the code, e.g. timing and the bytes received off the wire.

PR #28411, opened 2021-12-23
Make Static Data Provider Capable of Handling Cert Updates

ZhenLian:
This intends to replace https://github.com/grpc/grpc/pull/26663.
It does the following things:
1. expose two new APIs for `StaticDataCertificateProvider` that can update the root certs as well as the key-cert pair list being used
2. rename `StaticCertificateProvider` to `DataWatcherCertificateProvider` since it is not "static" anymore

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/28411)

PR #31445, opened 2022-10-24
Modify StaticDataCertificateProvider to InMemory... to handle cert updates

rockspore:
This intends to replace #28411. I tried to address all the comments, particularly the structural concern raised in that PR.

Also had to made updates to the C++ APIs and e2e tests since the original cert provider has been in use since that PR.


rockspore:
@ZhenLian No rush to look at this. I'll focus on the other PR for now.

rockspore:
This PR is based on #28411 with all the comments received there. Namely:
* I changed the class name to `InMemory...` since now it allows the application to set root certs and key/cert pairs in memory.
* I took the suggestion of using a `SingleCertificateCache` and had the same logic reused in both the `InMemoryCertificateProvider` and the `FileWatcher...` one.

@markdroth As you gave the review of the aforementioned PR, would you mind taking a look at this? Thank you!

rockspore:
Now that I understand more, this feels like a breaking change or at least would introduce a huge migration effort. Will close this for now.

PR #31446, opened 2022-10-24
XdsClient: fix behavior for does-not-exist timer and stream termination

markdroth:
This fixes XdsClient behavior to match what is described in [gRFC A57](https://github.com/grpc/proposal/pull/335):
- Don't run the resource does-not-exist timer unless the channel is connected.
- Don't report error on stream termination unless we didn't receive any responses on the stream.


src/core/ext/xds/xds_client.cc
```diff
@@ -178,10 +181,28 @@ class XdsClient::ChannelState::AdsCallState
...
+    }
+
     void MaybeStartTimer(RefCountedPtr<AdsCallState> ads_calld)
         ABSL_EXCLUSIVE_LOCKS_REQUIRED(&XdsClient::mu_) {
-      if (!timer_start_needed_) return;
-      timer_start_needed_ = false;
+      // Don't start timer if we've already either seen the resource or
+      // marked it as non-existing.
+      if (resource_seen_) return;
+      // Don't start timer if we haven't yet sent the resource.
```
yashykt, line 200: s/resource/request?

yashykt (approved):

src/core/ext/xds/xds_client.cc
markdroth: Done.

src/core/ext/xds/xds_transport.h
```diff
@@ -70,14 +69,45 @@ class XdsTransportFactory : public InternallyRefCounted<XdsTransportFactory> {
...
   };
 
+  // An interface for reporting connectivity state events for a transport.
+  //
+  // A transport can be in one of three states:
+  // - CONNECTING: transport is attempting to establish a connection
+  // - READY: transport is connected and ready to start streams
+  // - TRANSIENT_FAILURE: connection attempt has failed
+  //
+  // A transport is assumed to start in state CONNECTING.  Expected
```
yashykt, line 79: I think this comment should state that the transitions are written keeping pick first in mind

src/core/ext/xds/xds_client.cc
```diff
@@ -1108,9 +1208,14 @@ void XdsClient::ChannelState::AdsCallState::OnStatusReceived(
...
       // Try to restart the call.
       parent_->OnCallFinishedLocked();
-      // Send error to all watchers for the channel.
-      chand()->SetChannelStatusLocked(absl::UnavailableError(absl::StrFormat(
-          "xDS call failed; status: %s", status.ToString().c_str())));
+      // If we didn't receive a response on the stream, report the
+      // stream failure as a connectivity failure, which will report the
+      // error to all watchers of resources on this channel.
+      if (!seen_response_) {
+        chand()->SetChannelStatusLocked(absl::UnavailableError(
```
yashykt, line 1215: if we didn't receive a response and the channel also sees a connectivity state change, we would report multiple errors. (probably not an issue though)

yashykt (approved):

src/core/ext/xds/xds_transport.h
markdroth: I don't think that would really be appropriate, since this transport interface is supposed to abstract away gRPC-specific details like pick_first.  (It's probably already a bit more detailed than it needs to be, but I was trying to make it clear how to implement the interface for other systems internally.)

src/core/ext/xds/xds_client.cc
markdroth: Yeah, that's true.  That's not really new with this PR, though, and it doesn't cause any problems.

PR #31447, opened 2022-10-24
Introduce internally_modified tag and add it for event_engine_base_hdrs target

Vignesh2208:


ctiller (approved):

ctiller (changes requested):
holding, design discussion ongoing

PR #31450, opened 2022-10-25
Fix hanging UnaryCall with corrupt status

tonydnewell:
This doesn't address why the status could not be decoded, but it does wrap reading the status and metadata so that exceptions are not thrown so that the call to HandleUnaryResponse happens and resources are cleaned up correctly.
See https://github.com/grpc/grpc/issues/29854 for more discussion.

### DIFF
--- a/src/csharp/Grpc.Core/Internal/AsyncCall.cs
+++ b/src/csharp/Grpc.Core/Internal/AsyncCall.cs
@@ -116,7 +116,8 @@ namespace Grpc.Core.Internal
                             {
                                 using (profiler.NewScope("AsyncCall.UnaryCall.HandleBatch"))
                                 {
-                                    HandleUnaryResponse(success, ctx.GetReceivedStatusOnClient(), ctx.GetReceivedMessageReader(), ctx.GetReceivedInitialMetadata());
+                                    HandleUnaryResponse(success, ctx.TryGetReceivedStatusOnClient(),
+                                        ctx.GetReceivedMessageReader(), ctx.TryGetReceivedInitialMetadata());
                                 }
                             }
                             catch (Exception e)
--- a/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
+++ b/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
@@ -86,6 +86,23 @@ namespace Grpc.Core.Internal
             return MetadataArraySafeHandle.ReadMetadataFromPtrUnsafe(metadataArrayPtr);
         }
 
+        // Gets data of recv_initial_metadata completion without throwing an
+        // exception if the data is corrupt. This is to stop UnaryCall from hanging
+        // if there is some internal problem.
+        // See https://github.com/grpc/grpc/issues/29854
+        public Metadata TryGetReceivedInitialMetadata()
+        {
+            try
+            {
+                return GetReceivedInitialMetadata();
+            }
+            catch (Exception e)
+            {
+                Logger.Error(e, "Failed to read initial metadata");
+                return Metadata.Empty;
+            }
+        }
+
         // Gets data of recv_status_on_client completion.
         public ClientSideStatus GetReceivedStatusOnClient()
         {
@@ -99,6 +116,24 @@ namespace Grpc.Core.Internal
             var metadata = MetadataArraySafeHandle.ReadMetadataFromPtrUnsafe(metadataArrayPtr);
 
             return new ClientSideStatus(status, metadata);
+        }
+
+        // Gets data of recv_status_on_client completion without throwing an
+        // exception if the data is corrupt. This is to stop UnaryCall from hanging
+        // if there is some internal problem.
+        // See https://github.com/grpc/grpc/issues/29854
+        public ClientSideStatus TryGetReceivedStatusOnClient()
+        {
+            try
+            {
+                return GetReceivedStatusOnClient();
+            }
+            catch (Exception e)
+            {
+                Logger.Error(e, "Failed to read client status");
+                Status status = new Status(StatusCode.Internal, "Failed to read client status", e);
+                return new ClientSideStatus(status, Metadata.Empty);
+            }
         }
 
         public IBufferReader GetReceivedMessageReader()


tonydnewell:
This does not address why the status is corrupt when reading it. I cannot reproduce that. It looks like the memory is being overwritten while the UTF-8 bytes are being decoded.

This fix should stop any hanging in a unary call by wrapping the reading of the status (and metadata) and not throwing an exception. A new error status is produced instead containing the original exception and also a message is logged.

It was not possible to produce a unit test. The only way I could test this was to alter the code for `BatchContextSafeHandle.GetReceivedStatusOnClient()` and throw an exception:
```C#
    // TODO REMOVE - FOR TESTING ONLY
    if (details.Equals("Stream removed"))
    {
        // Fake an exception
        throw new System.ArgumentException("forced exception for Stream removed");
    }
```
and kill the server during a unary call to get the "Stream removed" error.

tonydnewell:
@jtattermusch

zhenomansh:
Guys! Any chance to have it in the release/service pack? :)

src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
```diff
@@ -86,6 +86,23 @@ public Metadata GetReceivedInitialMetadata()
             return MetadataArraySafeHandle.ReadMetadataFromPtrUnsafe(metadataArrayPtr);
         }
 
+        // Gets data of recv_initial_metadata completion without throwing an
+        // exception if the data is corrupt. This is to stop UnaryCall from hanging
+        // if there is some internal problem.
+        // See https://github.com/grpc/grpc/issues/29854
+        public Metadata TryGetReceivedInitialMetadata()
```
jtattermusch, line 93: I went through the thread of #29854 and didn't see any mention of failing to parse the InitalMetadata. The only issue that's reported is an error being thrown when GetReceivedStatusOnClient tries to convert the statusDetails to string. So it seems that proactively trying to modify GetReceivedInitialMetadata is unnecessary?

The location of the problem:
https://github.com/grpc/grpc/blob/90b0ea387e988b8615882e152a5152ed0801ec06/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs#L94

src/csharp/Grpc.Core/Internal/AsyncCall.cs
```diff
@@ -116,7 +116,8 @@ public TResponse UnaryCall(TRequest msg)
                             {
                                 using (profiler.NewScope("AsyncCall.UnaryCall.HandleBatch"))
                                 {
-                                    HandleUnaryResponse(success, ctx.GetReceivedStatusOnClient(), ctx.GetReceivedMessageReader(), ctx.GetReceivedInitialMetadata());
+                                    HandleUnaryResponse(success, ctx.TryGetReceivedStatusOnClient(),
```
jtattermusch, line 119: Changing the way the data gets read from context here would make the behavior for unary calls and for streaming calls inconsistent and I don't think we want that. I think instead, we should make a small modification to the original GetReceivedStatusOnClient() method and revert the changes here.

src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
```diff
@@ -101,6 +118,24 @@ public ClientSideStatus GetReceivedStatusOnClient()
             return new ClientSideStatus(status, metadata);
         }
 
+        // Gets data of recv_status_on_client completion without throwing an
+        // exception if the data is corrupt. This is to stop UnaryCall from hanging
+        // if there is some internal problem.
+        // See https://github.com/grpc/grpc/issues/29854
+        public ClientSideStatus TryGetReceivedStatusOnClient()
```
jtattermusch, line 125: Instead of introducing a new method, we should modify the original GetReceivedStatusOnClient().

AFAIK The only problem experienced in #29854 was the inability to convert the detailsPtr to details string (via 
MarshalUtils.PtrToStringUTF8)

I think we should simply have a  try - catch around `MarshalUtils.PtrToStringUTF8(detailsPtr, (int)detailsLength.ToUInt32());` (and possibly also around `string debugErrorString = Marshal.PtrToStringAnsi(...)`) and log an error if the conversion fails (which should never happen under normal circumstances).

https://github.com/grpc/grpc/blob/90b0ea387e988b8615882e152a5152ed0801ec06/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs#L94

jtattermusch: One way you could do this is to wrap the MarshalUtils.PtrToStringUTF8 into a new private method (e.g. `PtrToStringUTF8WithSafeFallback` and call it instead of MarshalUtils.PtrToStringUTF8 here.
https://github.com/grpc/grpc/blob/90b0ea387e988b8615882e152a5152ed0801ec06/src/csharp/Grpc.Core/Internal/MarshalUtils.cs#L37

jtattermusch (commented):
I left comments to make it clear what are the changes we want.
If we do what I proposed (a slight modification to GetReceivedStatusOnClient() only), I'll be happy accept the patch.

zhenomansh:
[jtattermusch](https://github.com/jtattermusch),

I believe there are couple of approaches how to prevent the string conversion method from throwing an exception, however, from my perspective, the major big is that the catch of the exception inside the UnaryCall method prevents setting of the unaryResponseTcs result and straight after unaryResponseTcs.Task.GetAwaiter().GetResult() is called. It is a simple bug, causing critical thread hanging. This bit is what we really need. It has no difference for us if the exception be thrown from the GetReceivedStatusOnClient method, or be set by unaryResponseTcs.SetException method. It is a few lines of code which we wait for 7 months already. Could you fix it first, and then continue discussing possible ways of handling/logging the original exception? Pleeease :) :) :)

jtattermusch:
@tonydnewell  can you please update this PR to address my comments soon so we end up with a workaround that we can merge. I plan to do a patch release of Grpc.Core once the workaround is merged.

jtattermusch:
> @tonydnewell can you please update this PR to address my comments soon so we end up with a workaround that we can merge. I plan to do a patch release of Grpc.Core once the workaround is merged.

I went ahead and implemented the solution: https://github.com/grpc/grpc/pull/31853

src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
tonydnewell: I was being thorough and handling all cases that could cause a hang if an exception was thrown

src/csharp/Grpc.Core/Internal/AsyncCall.cs
tonydnewell: The change is to make sure no exceptions are thrown in `UnaryCall`  to avoid a hang (by wrapping _all_ methods that are being called before calling `HandleUnaryResponse`) and make sure that the consequences of any exception are cleaned up.

In the original `HandleUnaryResponse` there was already a `TryDeserialize` - I am just applying the same pattern to the other methods `TryGetReceivedStatusOnClient` and `TryGetReceivedInitialMetadata`.

jtattermusch:
Superseded by https://github.com/grpc/grpc/pull/31853.

PR #31453, opened 2022-10-25
Fix ODR violations due to multiple definitions of socket_mutator

Vignesh2208:


ctiller (approved):

drfloob (changes requested):
(adding caution because we're moving fast on fixes) FYI, this breaks the build, please do not force merge this.

drfloob (approved):

PR #31457, opened 2022-10-25
Second attempt: xDS Listener: use ValidationErrors and add unit test

markdroth:


drfloob:
WDYT about doing a test import of this PR before we merge?

markdroth:
> WDYT about doing a test import of this PR before we merge?

I already did.  That's how I know it actually solves the problem. :)

drfloob (approved):

PR #31463, opened 2022-10-26
Visibility change to ref_counted_ptr for xds_client_core

allanrbo:
Hi @markdroth . I believe we need this for the Google internal project we are doing. I was running the IWYU tool and realized that ref_counted_ptr are currently only available accidentally through transitive dependencies.


ctiller:
We might be better off just granting whatever has xds client visibility ref_counted_ptr visibility internally

allanrbo:
That option sounds good to me as well. Will wait and hear @markdroth 's opinion too.

markdroth:
Either option will work, but Craig's suggestion is probably simpler, so let's just go with that.

PR #31436, opened 2022-10-24
Revert "Revert "Add back C# distribtests for Grpc.Tools""

jtattermusch:
Reverts grpc/grpc#31412.

Second attempt to add https://github.com/grpc/grpc/pull/31399.
The original PR was reverted because the packages from testcodegen.proto clashed with some internal .proto files. Changing the package should help.

PR #31464, opened 2022-10-26
Revert "Fix msbuild failing when '@' is present in path"

jtattermusch:
Reverts grpc/grpc#31133

I just found that the changes in the PR very likely break the (non-yet-added) C# distribtests from 
https://github.com/grpc/grpc/pull/31436.
On the PR I got this failure 
https://source.cloud.google.com/results/invocations/08298adc-8586-494c-8488-836f3b505cef/targets;collapsed=/github%2Fgrpc%2Fdistribtests/tests;passed=true and the problem goes away when I reverted the changes from #31133 locally.

The best way forward is to revert #31133, then merge the PR that adds the distribtest and only reintroduce #31133 when the distribtests are happy.


jtattermusch:
@tonydnewell

apolcyn (approved):

PR #31471, opened 2022-10-26
c2p resolver: don't fall back to dns if a bootstrap is present but federation is enabled

apolcyn:
Along with https://github.com/grpc/grpc/pull/29922, this change will allow us to start integration tests for this use case


markdroth:
I want to write this in a slightly different way, so I'm going to merge these changes into #29922.

PR #31482, opened 2022-10-27
Reland: "[EventEngine] Refactor TimerManager to leverage a shared ThreadPool (#31392)"

drfloob:
Relands #31392 with a fix for a 5% flake caused by a Windows EventEngine deadlock on shutdown.


ctiller (approved):
Exactly how I typed the fix too!

drfloob:
> Exactly how I typed the fix too!

Excellent pair-debugging session yesterday, by the way, thanks for the help!

PR #31503, opened 2022-10-31
Make grpc_microbenchmarks_diff run under docker for better maintainability

jtattermusch:
- allows running the microbenchmarks_diff test under docker, so that it doesn't really on a bunch of stuff installed on the kokoro worker.
- this is a prerequisite to switching the grpc_microbenchmarks_diff job from a custom "grpc-perf" image (which we want to deprecate) to a standard "grpc-ubuntu20" image (which is used by most of out other jobs), thus requiring much less maintenance.
- running the microbenchmarks under docker should have no effect on the measured performance AFAICT.

### DIFF
--- a/tools/internal_ci/linux/grpc_microbenchmark_diff.sh
+++ b/tools/internal_ci/linux/grpc_microbenchmark_diff.sh
@@ -16,15 +16,11 @@
 # This script is invoked by Kokoro and runs a diff on the microbenchmarks
 set -ex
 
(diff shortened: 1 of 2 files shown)


jtattermusch:
Microbenchmarks diff seems to be passing (after injecting a C/C++ change to make sure the benchmarks are triggered): https://source.cloud.google.com/results/invocations/bd751f60-0c08-4c3f-8c39-fe96a5572a7b/log

veblush (approved):

PR #31504, opened 2022-10-31
Run grpc_android job under docker for better maintainability.

jtattermusch:
- otherwise android SDK and NDK needs to be installed on the kokoro workers, which requires non-trivial maintenance and makes it hard to upgrade VM images.


jtattermusch:
CC @ralphchung

tools/dockerfile/test/android_ndk/Dockerfile
```diff
@@ -93,15 +93,27 @@ RUN apt-get update && apt-get -y install openjdk-8-jdk && apt-get clean
...
-# Install Android NDK and cmake using sdkmanager
-RUN mkdir -p ~/.android && touch ~/.android/repositories.cfg
+ENV ANDROID_HOME /opt/android-sdk
 RUN yes | $ANDROID_SDK_PATH/tools/bin/sdkmanager --licenses  # accept all licenses
-RUN $ANDROID_SDK_PATH/tools/bin/sdkmanager ndk-bundle 'cmake;3.6.4111459'
-ENV ANDROID_NDK_PATH $ANDROID_SDK_PATH/ndk-bundle
-ENV ANDROID_SDK_CMAKE $ANDROID_SDK_PATH/cmake/3.6.4111459/bin/cmake
+
+# Install Android NDK
+RUN curl -L -o android-ndk.zip http://dl.google.com/android/repository/android-ndk-r16b-linux-x86_64.zip && \
```
ralphchung, line 105: I am not sure which NDK version should be used. From my recent experiments, NDK 16 can make my code work, but some later versions cannot.

jtattermusch: AFAICT, this version is what was used in the original kokoro ubuntu16.04 workers (which is where the grpc_android job is currently running on). So starting with the same version makes sense (and we can adjust in the future if needed).

veblush (approved):

jtattermusch:
The "Android" test is green, so this should be good to merge: https://source.cloud.google.com/results/invocations/83e1f048-4962-4f66-91cc-ead2c9ace402

PR #31510, opened 2022-10-31
Fix FilterFuzzer's handling of EventEngines

drfloob:


drfloob:
The filter fuzzer currently hangs forever if it evaluates the `max_age` filter. I bisected that problem to https://github.com/grpc/grpc/pull/31265, which will need to be fixed before this fix can land.

PR #31511, opened 2022-10-31
[EventEngine][work stealing] Integrate WorkQueue into the ThreadPool

drfloob:
A step towards work stealing: replace the std::queue inside of the ThreadPool with the highly-tuned WorkQueue.

Unexpected performance improvement: 2x faster execution of `ThreadPoolTest.CanStartLotsOfClosures` in `opt` builds: ~900ms now, compared to consistently > 2s in the previous implementation. I improved the test to ensure all callbacks were actually being executed.

Performance improvement for EventEngine::Run benchmarks:

image

The few performance decreases are in "deep" tests, where a closure will schedule itself 100/1000 times in a chain with no fan out, where greater regressions correspond to deeper chains.


drfloob:
Superseded by #32869. This PR drifted out of sync, and some critical changes landed in the ThreadPool. Starting anew.

PR #30300, opened 2022-07-14
Update Clang 14

veblush:

PR #31518, opened 2022-11-01
Update Clang 15

veblush:
Update all docker images using clang from 13 to 15.

A few changes:
- clang_format slightly changes how to format code.
- clang_tidy introduced new following checks which had to be disabled because they needed significant code changes;
   - `bugprone-assignment-in-if-condition`
   - `bugprone-easily-swappable-parameters`
   - `bugprone-unchecked-optional-access`
- Following checks of clang_tidy became more strict so some changes are made to accommodate them;
  - `google-readability-casting`: It started having a few false positives so I disabled it. We may want to revisit this check.
  - `performance-inefficient-vector-operation`: It started catching more so I ran autofix to change suggested fixes.
  - `readability-simplify-boolean-expr`: It introduced more checks (DeMorgan) but its suggestion might not be ideal so I added a new option to exclude DeMorgan case.

This PR deprecates https://github.com/grpc/grpc/pull/30300

### DIFF
--- a/tools/dockerfile/grpc_iwyu/iwyu.sh
+++ b/tools/dockerfile/grpc_iwyu/iwyu.sh
@@ -24,12 +24,12 @@ CPU_COUNT=`nproc`
 
 rm -rf iwyu || true
 git clone https://github.com/include-what-you-use/include-what-you-use.git iwyu
-# latest commit on the clang 14 branch
+# latest commit on the clang 15 branch
 cd ${IWYU_ROOT}/iwyu
-git checkout abd5d2bd6320867d3605227a7f798a4e08fef896
+git checkout f20eadd2060103543dac46732cb4dd7dbed970e2
 mkdir -p ${IWYU_ROOT}/iwyu_build
 cd ${IWYU_ROOT}/iwyu_build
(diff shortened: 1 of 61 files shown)


tools/dockerfile/grpc_iwyu/iwyu.sh
```diff
@@ -24,12 +24,12 @@ CPU_COUNT=`nproc`
...
 git clone https://github.com/include-what-you-use/include-what-you-use.git iwyu
-# latest commit on the clang 13 branch
+# latest commit on the clang 15 branch
 cd ${IWYU_ROOT}/iwyu
-git checkout fbd921d6640bf1b18fe5a8a895636215367eb6b9
+git checkout f20eadd2060103543dac46732cb4dd7dbed970e2
 mkdir -p ${IWYU_ROOT}/iwyu_build
 cd ${IWYU_ROOT}/iwyu_build
-cmake -G "Unix Makefiles" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DLLVM_ROOT_DIR=/usr/lib/llvm-13 ${IWYU_ROOT}/iwyu 
+cmake -G "Unix Makefiles" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DLLVM_ROOT_DIR=/usr/lib/llvm-14 ${IWYU_ROOT}/iwyu
```
ctiller, line 32: why is this llvm-14?

veblush: Oops. It's a mistake. Fixed!

markdroth (approved):

PR #31525, opened 2022-11-02
Cmake add separate export for plugin targets

clementperon:
Add a separate CMake export for the plugin targets to separate binaries
and libraries into their own CMake target export files. Skip the cross
compiled binary plugin targets during cross compile because they are not
usable and not always available.

The Yocto build system doesn't install cross compiled binaries into the
target sysroot. This makes the CMake gRPC config useless as it checks
the existent of binaries and fails without the binaries.

This is a strip down version of https://github.com/grpc/grpc/pull/22498 and related to https://github.com/grpc/grpc/pull/26148 and https://github.com/grpc/grpc/issues/26857.

@veblush @jtattermusch @herbrechtsmeier

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11092,7 +11092,7 @@ target_link_libraries(grpc_cpp_plugin
 
 
 if(gRPC_INSTALL)
-  install(TARGETS grpc_cpp_plugin EXPORT gRPCTargets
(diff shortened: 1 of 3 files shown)


jtattermusch (approved):
Seems fair. LGTM as long as all the tests pass.
I'd also like to get an opinion from @veblush before we merge.

veblush (approved):
LGTM.

PR #31527, opened 2022-11-02
Fix msbuild failing when '@' is present in path (2nd attempt)

tonydnewell:
Fix for https://github.com/grpc/grpc/issues/30746 - "C# Protobuf msbuild task fails with glob pattern if "@" is present in the path"

After investigating there is a feature/bug with MSBuild: https://github.com/dotnet/msbuild/issues/8104
where setting the `Source` metadata from the `Identity` metadata needs unescaping if the original items come from
a wildcard match and the paths contain an '@' character. See https://github.com/dotnet/msbuild/issues/8104#issuecomment-1298557627 for details.

Also as a precaution make sure no duplicate files in the list of files passed to the CSC compiler.

Tested on Linux and Windows with MSBuild versions 15, 16 and 17


tonydnewell:
@jtattermusch new fix for https://github.com/grpc/grpc/issues/30746 which should also fix the distribution tests.
I'm adding tests to Grpc.Tools.Tests separately.

src/csharp/Grpc.Tools/build/_protobuf/Google.Protobuf.Tools.targets
```diff
@@ -190,17 +194,24 @@
     <ItemGroup>
       <_Protobuf_CodeCompile Include="@(Protobuf_ExpectedOutputs->Distinct())"
          Condition=" '%(Source)' != '' and '@(Protobuf_Compile->WithMetadataValue('CompileOutputs', 'true'))' != '' " />
-      <Compile Include="@(_Protobuf_CodeCompile)" />
+      <!-- Adding Distinct() is just a precaution as the CSC compiler complains if there are 
+           duplicate source files
+      -->
+      <Compile Include="@(_Protobuf_CodeCompile->Distinct())" />
```
jtattermusch, line 200: qq: if there are duplicates, wouldn't it be fair to report that as a problem? Did this actually happen to you during testing?

jtattermusch (approved):
LGTM after there's some clarification on my question.

src/csharp/Grpc.Tools/build/_protobuf/Google.Protobuf.Tools.targets
tonydnewell: The Distinct() is just a precaution. No duplicates seen in testing with the new Unescape() fix (duplicates were seen with the old qualified Source fix).

PR #31535, opened 2022-11-02
Update grpc_deps() to add omit parameters.

laramiel:
This updates grpc dependencies to be more aligned with best practices according to https://github.com/bazelbuild/bazel/issues/1952

This should help us (https://github.com/google/tensorstore) to build gRPC with a smaller set of repositories.

### DIFF
--- a/bazel/grpc_deps.bzl
+++ b/bazel/grpc_deps.bzl
@@ -90,7 +90,7 @@ def grpc_deps(
       bind_externals = bind_externals,
       configure_python = configure_python,
       omit_io_bazel_rules_python = omit_io_bazel_rules_python,
-      omit_cython = omit_cython):
+      omit_cython = omit_cython)
 
 
 def grpc_deps_impl(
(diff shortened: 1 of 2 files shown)


jtattermusch:
@gnossen  can you please review?

jtattermusch:
CC @veblush

gnossen:
Thank you for the contribution! This is something we've been thinking about for a while. I have a couple of high level observations/questions.

How would backward compatibility work with this approach? The set of flags seems to be tied closely to our set of dependencies, which should (ideally) be an implementation detail from the perspective of our users. For example, we've tried to shelter our users from listing all of their transitive dependencies by introducing the `grpc_deps` function. Our set of dependencies is definitely going to change in the future. How do we deal with that using these flags?

There are two cases: removing a dependency and adding a dependency.

If we remove a dependency, we can leave its corresponding flag and make it a no-op. This is perfectly backward compatible and probably what most people removing a dependency would do on their first try. It _does_ however mean that our set of flags can only ever grow.

If we add a dependency, we'll have to add a corresponding flag. I have no confidence that folks adding dependencies will know about this flag system and also remember to add a new flag for their new dependency. So i think we probably want some automation to ensure that the set of dependencies matches up with the set of flags. I don't think we need to block this PR on that, but it's something we'll need to add in the near future.

PR #31540, opened 2022-11-03
A basic set of EventEngine listener tests

Vignesh2208:


test/core/event_engine/test_suite/server_test.cc
```diff
@@ -12,12 +12,263 @@
...
+  auto quota = grpc_core::ResourceQuota::Default();
+  args = args.Set(GRPC_ARG_RESOURCE_QUOTA, quota);
+  ChannelArgsEndpointConfig config(args);
+  auto status = test_ee->CreateListener(
+      std::move(accept_cb), [](absl::Status /*status*/) {}, config,
+      std::make_unique<grpc_core::MemoryQuota>("foo"));
+  EXPECT_TRUE(status.ok());
+
+  std::unique_ptr<Listener> listener = std::move(*status);
+  EXPECT_TRUE(listener->Bind(URIToResolvedAddress(target_addr)).ok());
```
drfloob, line 98: These two lines should probably be asserts instead of expects, since the rest of the test would be testing an invalid state otherwise. The same is probably true for the endpoint checks below.

test/core/event_engine/test_suite/server_test.cc
```diff
@@ -12,12 +12,263 @@
...
+  EXPECT_TRUE(listener->Start().ok());
+
+  oracle_ee->Connect(
+      [&client_endpoint,
+       &client_signal](absl::StatusOr<std::unique_ptr<Endpoint>> status) {
+        if (!status.ok()) {
+          gpr_log(GPR_ERROR, "Connect failed: %s",
```
drfloob, line 105: Can this be an assert instead?

test/core/event_engine/test_suite/server_test.cc
```diff
@@ -12,12 +12,263 @@
...
+
+constexpr int kNumExchangedMessages = 100;
+
+}  // namespace
+
+// Create a connection using the oracle EventEngine to a listener created
+// by the Test EventEngine and exchange bi-di data over the connection.
+// For each data transfer, verify that data written at one end of the stream
+// equals data read at the other end of the stream.
+TEST_F(EventEngineServerTest, ServerConnectExchangeBidiDataTransferTest) {
```
drfloob, line 68: This test might flake on infrastructure network issues, which we have seen. We can watch it for a bit and see the flake rate. It might be nice to improve the test so that infrastructure flakes do not show up as test flakes/failures (e.g., add retries on identified poor network conditions), but we can consider that later. Just noting that this may flake through no fault of the code.

test/core/event_engine/test_suite/server_test.cc
```diff
@@ -12,12 +12,263 @@
...
+    EXPECT_TRUE(SendValidatePayload(GetNextSendMessage(), client_endpoint.get(),
+                                    server_endpoint.get())
+                    .ok());
+
+    // Send from server to client and verify data read at the client.
+    EXPECT_TRUE(SendValidatePayload(GetNextSendMessage(), server_endpoint.get(),
+                                    client_endpoint.get())
+                    .ok());
+  }
+  client_endpoint.reset();
```
drfloob, line 134: Are these being reset to ensure they don't hold EventEngine refs?

test/core/event_engine/test_suite/server_test.cc
Vignesh2208: Replaced all expects with asserts in both client_test.cc and server_test.cc

test/core/event_engine/test_suite/server_test.cc
Vignesh2208: Yes. replaced in both client_test.cc and server_test.cc

test/core/event_engine/test_suite/server_test.cc
Vignesh2208: Yeah. Otherwise the subsequent WaitForSingleOwner call will not return.

test/core/event_engine/test_suite/server_test.cc
```diff
@@ -12,12 +12,263 @@
...
+          grpc_core::MemoryAllocator /*memory_allocator*/) {
+        server_endpoint = std::move(ep);
+        server_signal.Notify();
+      };
+
+  grpc_core::ChannelArgs args;
+  auto quota = grpc_core::ResourceQuota::Default();
+  args = args.Set(GRPC_ARG_RESOURCE_QUOTA, quota);
+  ChannelArgsEndpointConfig config(args);
+  auto status = test_ee->CreateListener(
```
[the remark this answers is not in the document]
Vignesh2208: Fixed in both client_test.cc and server_test.cc

test/core/event_engine/test_suite/server_test.cc
drfloob: Ok, thanks for clarifying.

test/core/event_engine/test_suite/client_test.cc
```diff
@@ -114,46 +114,40 @@ TEST_F(EventEngineClientTest, ConnectExchangeBidiDataTransferTest) {
...
       std::move(accept_cb),
       [](absl::Status status) { GPR_ASSERT(status.ok()); }, config,
       std::make_unique<grpc_core::MemoryQuota>("foo"));
-  EXPECT_TRUE(status.ok());
+  ASSERT_TRUE(listener.ok());
 
-  std::unique_ptr<Listener> listener = std::move(*status);
-  EXPECT_TRUE(listener->Bind(URIToResolvedAddress(target_addr)).ok());
-  EXPECT_TRUE(listener->Start().ok());
+  ASSERT_TRUE((*listener)->Bind(URIToResolvedAddress(target_addr)).ok());
```
drfloob, line 123: This can be `ASSERT_TRUE(listener->Bind(...).ok())`, no need for `(*listener)->`. Part of the reason it's recommend to name StatusOr variables as the value types themselves is because the StatusOr class overrides `operator->`, allowing the variable to act as the type.

This pattern happens again below.

test/core/event_engine/test_suite/client_test.cc
```diff
@@ -114,46 +114,40 @@ TEST_F(EventEngineClientTest, ConnectExchangeBidiDataTransferTest) {
...
+  ASSERT_TRUE((*listener)->Start().ok());
 
   test_ee->Connect(
       [&client_endpoint,
        &client_signal](absl::StatusOr<std::unique_ptr<Endpoint>> status) {
```
drfloob, line 128: nit: As recommend elsewhere, StatusOr variable names should match underlying type. Would suggest naming this parameter `endpoint` or some such.

This pattern happens again below

drfloob (commented):
Looks pretty good! Just a few style-guide and syntax things remaining.

test/core/event_engine/test_suite/client_test.cc
Vignesh2208: Somehow it doesn't compile when I do that. I de-referenced the StatusOr now immediately when its created.

test/core/event_engine/test_suite/client_test.cc
Vignesh2208: Fixed.

drfloob (approved):

PR #31544, opened 2022-11-03
[test] Round robin lb policy test

eugeneo:


test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+  }
+  ExpectNoStateChange();
+
+  // All channels ready
+  SetSubchannelConnectivityState(1, GRPC_CHANNEL_READY);
+  SetSubchannelConnectivityState(2, GRPC_CHANNEL_READY);
+  ExpectState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+
+  std::unordered_set<std::string> reportedUris;
```
markdroth, line 153: This should be called `reported_uris`, as per the style guide.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+  // Picker should return each address once, we do not care about the order.
+  for (size_t i = 0; i < 20; ++i) {
+    absl::optional<std::string> address;
+    GetPickedAddress(picker.get(), address);
+    EXPECT_TRUE(address.has_value());
+    reportedUris.insert(*address);
+  }
+  EXPECT_EQ(reportedUris.size(), 3);
```
markdroth, line 161: This can be:

```
EXPECT_THAT(reportedUris, ::testing::UnorderedElementsAre(
    UriForSubchannel(0), UriForSubchannel(1), UriForSubchannel(2)));
```

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+  SetSubchannelConnectivityState(2, GRPC_CHANNEL_READY);
+  ExpectState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+
+  std::unordered_set<std::string> reportedUris;
+  // Picker should return each address once, we do not care about the order.
+  for (size_t i = 0; i < 20; ++i) {
+    absl::optional<std::string> address;
+    GetPickedAddress(picker.get(), address);
+    EXPECT_TRUE(address.has_value());
```
markdroth, line 158: This should be `ASSERT` instead of `EXPECT`, since if it does not have a value, the next line will be undefined behavior.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+
+  // All channels ready
+  SetSubchannelConnectivityState(1, GRPC_CHANNEL_READY);
+  SetSubchannelConnectivityState(2, GRPC_CHANNEL_READY);
+  ExpectState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+
+  std::unordered_set<std::string> reportedUris;
+  // Picker should return each address once, we do not care about the order.
+  for (size_t i = 0; i < 20; ++i) {
```
markdroth, line 155: Here we are doing 20 picks, but we are adding the address of each pick to `reportedUris` just once, so we're not actually checking that we're getting each address back the same number of times or that they're in the right order.  Instead, I suggest something like this:

```
void ExpectRoundRobinPicks(
    LoadBalancingPolicy::SubchannelPicker* picker,
    const std::vector<std::string>& addresses,
    SourceLocation location = SourceLocation()) {
  auto find_address_index = [&](const std::string& address) -> absl::optional<size_t> {
    for (size_t i = 0; i < addresses.size(); ++i) {
      if (address == addresses[i]) return i;
    }
    return absl::nullopt;
  };
  // Make 20 passes across all addresses.
  absl::optional<size_t> last_index;
  for (size_t i = 0; i < 20 * addresses.size(); ++i) {
    auto address = GetPickedAddress(picker.get());
    ASSERT_TRUE(address.has_value())
        << location.file() << ":" << location.line();
    auto address_index = find_address_index(*address);
    ASSERT_TRUE(address_index.has_value())
        << location.file() << ":" << location.line();
    if (!last_address.has_value()) {
      last_address = address_index;
    } else {
      EXPECT_EQ(*address_index, (*last_index + 1) % addresses.size())
          << location.file() << ":" << location.line();
      last_index = address_index;
    }
  }
}
```

With that helper function, individual tests can do something like this:

```
// First subchannel reports READY.  Picker always chooses that subchannel.
SetSubchannelConnectivityState(0, GRPC_CHANNEL_READY);
ExpectRoundRobinPicks(picker.get(), {address0});
// Second subchannel reports READY.  Picker should round-robin across the two READY subchannels.
SetSubchannelConnectivityState(1, GRPC_CHANNEL_READY);
ExpectRoundRobinPicks(picker.get(), {address0, address1});
// Last subchannel reports READY.  Picker should round-robin across all three subchannels.
SetSubchannelConnectivityState(2, GRPC_CHANNEL_READY);
ExpectRoundRobinPicks(picker.get(), {address0, address1, address2});
```

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+  SetSubchannelConnectivityState(0, GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  // Picker should return the same subchannel repeatedly.
+  for (size_t i = 0; i < 3; ++i) {
+    ExpectPickComplete(picker.get(), UriForSubchannel(0));
+  }
+  ExpectNoStateChange();
+
+  // All channels ready
+  SetSubchannelConnectivityState(1, GRPC_CHANNEL_READY);
```
markdroth, line 148: After the second subchannel is ready, the picker should return the two ready subchannels.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+  }
+  EXPECT_EQ(reportedUris.size(), 3);
+  for (size_t i = 0; i < 3; ++i) {
+    EXPECT_NE(reportedUris.find(UriForSubchannel(i)), reportedUris.end())
+        << "Subchannel" << i;
+  }
+  ExpectNoStateChange();
+
+  // Shutdown channel is not considered
+  SetSubchannelConnectivityState(1, GRPC_CHANNEL_TRANSIENT_FAILURE);
```
markdroth, line 169: As mentioned above, a subchannel will not transition from READY directly to TRANSIENT_FAILURE.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+  ExpectState(GRPC_CHANNEL_TRANSIENT_FAILURE);
+  ExpectNoStateChange();
+}
+
+TEST_F(RoundRobinTest, NoChannels) {
+  auto policy = MakeLbPolicy("round_robin");
+  LoadBalancingPolicy::UpdateArgs update_args;
+  update_args.addresses.emplace();
+  absl::Status status = ApplyUpdate(std::move(update_args), policy.get());
+  EXPECT_TRUE(absl::IsUnavailable(status));
```
markdroth, line 237: In this case, the LB policy will also report state TRANSIENT_FAILURE, so we should check that behavior as well.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -451,7 +465,31 @@ class LoadBalancingPolicyTest : public ::testing::Test {
...
+                          absl::string_view address_uri,
+                          SourceLocation location = SourceLocation()) {
+    absl::optional<std::string> picked_uri;
+    GetPickedAddress(picker, picked_uri, location);
+    EXPECT_EQ(*picked_uri, address_uri)
+        << location.file() << ":" << location.line();
+  }
+
+  // Requests a pick on picker and expects a Fail result.
+  void ExpectPickFail(LoadBalancingPolicy::SubchannelPicker* picker,
```
markdroth, line 483: This method should check the status with which the LB policy fails the pick, since that status will wind up being used as the failure status of the data plane RPC, and we want to make sure that such statuses are properly helpful for our users.

I suggest adding a required parameter of type `std::function<void(absl::Status)>` that we will pass the failure status to.  The caller can pass in a lambda that checks the status code and value.

markdroth (commented):

markdroth:
Also, it looks like the sanity test is failing.  I think if you run `./tools/buildgen/generate_projects.sh`, that should fix most of it.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -433,11 +446,12 @@ class LoadBalancingPolicyTest : public ::testing::Test {
         << location.file() << ":" << location.line();
   }
 
-  // Requests a pick on picker and expects a Complete result whose
-  // subchannel has the specified address.
-  void ExpectPickComplete(LoadBalancingPolicy::SubchannelPicker* picker,
-                          absl::string_view address_uri,
-                          SourceLocation location = SourceLocation()) {
+  // Requests a pick on picker and return the address from a Complete result.
+  // Has to be void and use output argument so the assertions still work
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: Done.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -433,11 +446,12 @@ class LoadBalancingPolicyTest : public ::testing::Test {
...
+  void GetPickedAddress(LoadBalancingPolicy::SubchannelPicker* picker,
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+#include "test/core/util/test_config.h"
+
+namespace grpc_core {
+namespace testing {
+namespace {
+
+class RoundRobinTest : public LoadBalancingPolicyTest {
+ public:
+  OrphanablePtr<LoadBalancingPolicy> RoundRobinPolicy(
+      size_t channelsCount = 3) {
```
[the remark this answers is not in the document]
eugeneo: Done. I will do my best, but this will take some adjusting...

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: I added a check to make sure each channel is called the same number of times. I specifically did not check the pick order as my understanding it is not a part of a public contract (i.e. I see the randomization of the first pick was added later).

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: Added a check.

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: Changed.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
eugeneo: I inlined this function as there had only been a single call site.

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: Done.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,251 @@
...
+#include "test/core/client_channel/lb_policy/lb_policy_test_lib.h"
+#include "test/core/util/test_config.h"
+
+namespace grpc_core {
+namespace testing {
+namespace {
+
+class RoundRobinTest : public LoadBalancingPolicyTest {
+ public:
+  OrphanablePtr<LoadBalancingPolicy> RoundRobinPolicy(
```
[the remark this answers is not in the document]
eugeneo: Did the refactor and added a test case

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,373 @@
...
+      "ipv4:127.0.0.1:441",
+      "ipv4:127.0.0.1:442",
+      "ipv4:127.0.0.1:443",
+  };
+  auto policy = MakeLbPolicy("round_robin");
+
+  // One channel that is ready. Gets picked all the time.
+  auto status = ApplyUpdate(BuildUpdateArgs({uris[0]}), policy.get());
+  ASSERT_TRUE(status.ok()) << status;
+  ExpectState(GRPC_CHANNEL_CONNECTING);
```
markdroth, line 324: This should follow the same structure as above:

```
ExpectConnectingUpdate();
auto* subchannel = FindSubchannel(kAddresses[0]);
subchannels.push_back(subchannel);  // Save for later.
auto picker = WaitForConnected();
ExpectRoundRobinPicks(picker.get(), {kAddresses[0]});
```

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,373 @@
...
+  auto picked = ExpectPickAddress(picker.get());
+  EXPECT_EQ(*picked, uris[0]);
+  ExpectNoStateChange();
+
+  // Second channel added, connecting. Only the first channel gets picked still
+  status = ApplyUpdate(BuildUpdateArgs({uris[0], uris[1]}), policy.get());
+  ASSERT_TRUE(status.ok()) << status;
+  ExpectState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  EXPECT_EQ(*ExpectPickAddress(picker.get()), uris[0]);
```
markdroth, line 339: This will crash if `ExpectPickAddress()` returns nullopt.

I think we should just use `ExpectRoundRobinPicks()` here.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,373 @@
...
+
+TEST_F(RoundRobinTest, AddressListChange) {
+  std::array<std::string, 3> uris = {
+      "ipv4:127.0.0.1:441",
+      "ipv4:127.0.0.1:442",
+      "ipv4:127.0.0.1:443",
+  };
+  auto policy = MakeLbPolicy("round_robin");
+
+  // One channel that is ready. Gets picked all the time.
```
markdroth, line 321: Please change "channel" to "subchannel".

Same thing throughout.  There is nothing in this test that actually uses a channel, everything here is subchannels.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,373 @@
...
+          << "Subchannel " << uri << location.file() << ":" << location.line();
+    }
+  }
+};
+
+TEST_F(RoundRobinTest, SingleChannel) {
+  std::string uri = "ipv4:127.0.0.1:441";
+  auto policy = MakeLbPolicy("round_robin");
+  auto status = ApplyUpdate(BuildUpdateArgs({uri}), policy.get());
+
```
markdroth, line 88: Nit: Please don't leave blank lines within a function.  I generally find that anywhere I really want to break up the flow, I'm better off using a one-line comment than just having a blank line.

Same thing throughout.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,373 @@
...
+  ASSERT_TRUE(status.ok()) << status;
+  ExpectState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  EXPECT_EQ(*ExpectPickAddress(picker.get()), uris[0]);
+  EXPECT_EQ(*ExpectPickAddress(picker.get()), uris[0]);
+  ExpectNoStateChange();
+
+  // Second channel ready. Both channels are now picked.
+  FindSubchannel(uris[1], {})
+      ->SetConnectivityState(GRPC_CHANNEL_READY, absl::OkStatus());
```
markdroth, line 345: This new subchannel is not going to go from IDLE (the state it reports when it is first created) directly to READY.  We should either (a) use `CreateSubchannel()` to add the subchannel in the subchannel pool and set its state before we send the update containing the new address to the LB policy, or (b) go through the normal sequence of having the LB policy request a connection on the subchannel and then transitioning first to CONNECTING before then moving to READY.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -0,0 +1,373 @@
...
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ExpectPickAddresses(picker.get(), {uris[0], uris[1]});
+  ExpectNoStateChange();
+
+  // First channel removed, third added and made ready. First channel should
+  // not show up.
+  status = ApplyUpdate(BuildUpdateArgs({uris[1], uris[2]}), policy.get());
+  EXPECT_TRUE(status.ok()) << status;
+  FindSubchannel(uris[2], {})->SetConnectivityState(GRPC_CHANNEL_READY, {});
```
markdroth, line 354: Same here: The third subchannel will not go directly from IDLE to READY.

markdroth (commented):
Sorry for the huge number of comments here!  The main reason for that is that this review forced me to really step back and think about what cases we're likely to want to test as we improve our LB policy unit test coverage, and therefore what set of helper functions we really want to support in the test framework as the right set of "atoms" to use to construct the individual tests.  I've sent you #31632 with what I think is the right set of helper functions, and I think they now provide everything we need to write these tests the way we want to.  So the comments here are mainly trying to explain the mapping between what you have now and the way I'd like to see these kinds of tests written going forward.

Overall, my goal for the individual tests is that they be concise and easy to understand.  They should not depend on magic constants or behavior that is not obvious from reading the individual test, but they should avoid unnecessary verbosity by being written in terms of sensible "atoms" that define the common set of checks that we care about (e.g., always checking that the picker queues when the state is CONNECTING).  I tried to design those atoms so that each step of interacting with the LB policy is just a very small number of lines in the test code (ideally as little as one line, where possible).

I think you should be able to revert your changes to lb_policy_test_lib.h in favor of the ones in #31632.  Then you should be able to change the round_robin tests to use those updated helper functions.  Hopefully, the tests will come out shorter and easier to understand.

Thanks for doing this!  In addition to being a good way for you to get some familiarity with the LB policy API, it's also been really helpful in terms of forcing me to think through how we want these tests to look.  So I apologize for all the churn here, but I think the result will be worth it.

Please let me know if you have any questions about any of this.  Thanks!

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -45,185 +45,212 @@ namespace {
...
   ExpectNoStateChange();
 
-  // Shutdown channel is not considered
-  SetSubchannelConnectivityState(1, GRPC_CHANNEL_TRANSIENT_FAILURE);
+  second_subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE, absl::OkStatus());
   ExpectReresolutionRequest();
-  picker = ExpectState(GRPC_CHANNEL_READY);
+  ExpectState(GRPC_CHANNEL_READY);
   ExpectNoStateChange();
+  second_subchannel->SetConnectivityState(GRPC_CHANNEL_TRANSIENT_FAILURE,
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: I think I fixed this... At this point I am unable to trace this comment to original code.

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: Can't find this code, assume fixed.

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: done

test/core/client_channel/lb_policy/round_robin_test.cc
eugeneo: done

eugeneo (commented):

markdroth (commented):
This is definitely moving in the right direction!

My main comment here is the long one about writing the tests such that they're not too specific about no-op state changes and making sure that we check the picker behavior after every single state change.

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -44,318 +44,236 @@ namespace {
...
-  ExpectState(GRPC_CHANNEL_CONNECTING);
-  for (size_t i = 0; i < uris.size() - i; ++i) {
-    auto subchannel = FindSubchannel(uris[i]);
+  ExpectConnectingUpdate();
+  for (auto address : {kFirstAddress, kSecondAddress}) {
+    auto subchannel = FindSubchannel(address);
     ASSERT_NE(subchannel, nullptr);
+    subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+    WaitForConnected();
     subchannel->SetConnectivityState(GRPC_CHANNEL_TRANSIENT_FAILURE,
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -622,11 +657,19 @@ class LoadBalancingPolicyTest : public ::testing::Test {
...
+                                  ChannelArgs args = ChannelArgs(),
+                                  SourceLocation location = SourceLocation()) {
+    for (auto& subchannel_key_state : subchannel_pool_) {
+      SubchannelState& state = subchannel_key_state.second;
+      if (state.address() == address &&
+          subchannel_key_state.first.args() == args) {
+        return &state;
+      }
+    }
+    EXPECT_NE(nullptr, nullptr)
```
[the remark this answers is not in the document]
eugeneo: Removed

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -157,14 +157,17 @@ class LoadBalancingPolicyTest : public ::testing::Test {
...
     // will be reported to all associated SubchannelInterface objects.
     void SetConnectivityState(grpc_connectivity_state state,
-                              const absl::Status& status = absl::OkStatus()) {
+                              const absl::Status& status = absl::OkStatus(),
+                              SourceLocation location = SourceLocation()) {
       if (state == GRPC_CHANNEL_TRANSIENT_FAILURE) {
         EXPECT_FALSE(status.ok())
-            << "bug in test: TRANSIENT_FAILURE must have non-OK status";
+            << "bug in test: TRANSIENT_FAILURE must have "
+            << "non-OK status" << location.file() << ":" << location.line();
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -44,318 +44,236 @@ namespace {
...
-  auto picker = ExpectState(GRPC_CHANNEL_CONNECTING);
-  ExpectPickQueued(picker.get());
-  // LB policy should have requested a connection on this subchannel.
-  EXPECT_TRUE(FindSubchannel(uris[0])->ConnectionRequested());
-
-  for (const std::string& uri : uris) {
-    auto subchannel = FindSubchannel(uri);
+  ExpectConnectingUpdate();
+  std::array<SubchannelState*, 3> subchannels = {
+      FindSubchannel(kFirstAddress),
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -44,318 +44,236 @@ namespace {
...
-  picker = ExpectState(GRPC_CHANNEL_READY);
-  // Picker should return the same subchannel repeatedly.
-  for (size_t i = 0; i < 3; ++i) {
-    ExpectPickComplete(picker.get(), uris[0]);
+  for (auto subchannel : subchannels) {
+    subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+  }
+  for (size_t i = 0; i < subchannels.size(); i++) {
+    subchannels[i]->SetConnectivityState(GRPC_CHANNEL_READY);
+    // For the first subchannel, we use WaitForConnected() to drain any queued
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/round_robin_test.cc
```diff
@@ -44,318 +44,236 @@ namespace {
...
-  ExpectNoStateChange();
-
-  second_subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE, absl::OkStatus());
+  subchannels[1]->SetConnectivityState(GRPC_CHANNEL_IDLE);
   ExpectReresolutionRequest();
-  ExpectState(GRPC_CHANNEL_READY);
+  auto picker = WaitForConnected();
+  ExpectRoundRobinPicks(picker.get(), {kFirstAddress, kThirdAddress});
+  ExpectNoStateChange();
+  subchannels[1]->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
```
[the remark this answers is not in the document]
eugeneo: Done

eugeneo:
Rebased, modernized. Ready for review.

PR #31549, opened 2022-11-03
Remove remaining XDS dependencies when `grpc_no_xds=true` is specified

jbms:
Previously, gRPC still pulled in XDS dependencies even when `grpc_no_xds=true` was specified.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -2602,7 +2602,6 @@ grpc_cc_library(
         "//src/core:ext/filters/client_channel/health/health_check_client.cc",
         "//src/core:ext/filters/client_channel/http_proxy.cc",
         "//src/core:ext/filters/client_channel/lb_policy/child_policy_handler.cc",
-        "//src/core:ext/filters/client_channel/lb_policy/oob_backend_metric.cc",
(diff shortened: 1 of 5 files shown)


markdroth:
I don't think this is correct.  The ORCA proto, although it lives in the xDS repo, is not really part of xDS; it's part of the backend metric reporting mechanism described in [gRFC A51](https://github.com/grpc/proposal/blob/master/A51-custom-backend-metrics.md).  It is intended to work even when gRPC is built without xDS functionality.

jbms:
What I'm really trying to avoid is that those ORCA deps depend on `protoc-gen-validate` which brings in a dependency on go.  That greatly increases the footprint of gRPC when added as a dependency of another project.

markdroth:
That's a reasonable point.  @htuch, how would you feel about removing the protoc-gen-validate dependency for the ORCA protos?

jbms:
Looks like I was mistaken --- it only depends on validate.proto from that repo, not the go_binary itself.

That makes the dependency much less problematic.

PR #31555, opened 2022-11-04
[ObjC] run cf stream ios test with bazel

HannahShiSFB:
Existing cf stream xcodebuild test is still needed for running with sanitizers.

Sanitizer support for ios was added in https://github.com/bazelbuild/rules_apple/pull/1590 but is not released yet.

Will wait for a new release and update build_bazel_rules_apple inhttps://github.com/grpc/grpc/blob/v1.50.0/bazel/grpc_deps.bzl#L376

@dennycd

### DIFF
--- a/src/objective-c/tests/CFStreamTests/CFStreamClientTests.mm
+++ b/src/objective-c/tests/CFStreamTests/CFStreamClientTests.mm
@@ -1,0 +1,208 @@
+/*
+ *
+ * Copyright 2018 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


dennycd:
@HannahShiSFB  can u rebase and check Bazel RBE TSAN C/C++  test again, thanks

src/objective-c/tests/CFStreamTests/CFStreamClientTests.mm
```diff
@@ -0,0 +1,208 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#import <XCTest/XCTest.h>
```
dennycd, line 19: ideally we avoids forking a copy of existing test copy from https://github.com/grpc/grpc/tree/master/test/core/iomgr/ios/CFStreamTests but instead point to the same shared test source.  If not, consider mark this as a copied ?

HannahShiSFB: It’s not possible to reference files in other directories in bazel.
This copy is temporary until bazel apple rules 2.0 release which added sanitizer support. 
https://github.com/bazelbuild/rules_apple/releases

sampajano (approved):
thanks :)

PR #31560, opened 2022-11-04
XdsClient: implement fuzzer

markdroth:


ctiller (approved):
This is a good baseline...let's keep talking on how to get better semantic checks in place.

PR #31564, opened 2022-11-05
EventEngine::RunAfter: handshaker

yijiem:
Migrate handshaker from using iomgr timer to EventEngine::RunAfter.


src/core/lib/transport/handshaker.cc
```diff
@@ -201,10 +197,15 @@ void HandshakeManager::DoHandshake(grpc_endpoint* endpoint,
...
-    GRPC_CLOSURE_INIT(&on_timeout_, &HandshakeManager::OnTimeoutFn, this,
-                      grpc_schedule_on_exec_ctx);
-    grpc_timer_init(&deadline_timer_, deadline, &on_timeout_);
+    const Duration time_to_deadline = deadline - Timestamp::Now();
+    deadline_timer_handle_ =
+        event_engine_->RunAfter(time_to_deadline, [self = Ref()]() mutable {
+          ApplicationCallbackExecCtx callback_exec_ctx;
+          ExecCtx exec_ctx;
+          self->Shutdown(GRPC_ERROR_CREATE("Handshake timed out"));
+          // HandshakeManager deletion might require an active ExecCtx.
```
drfloob, line 206: Interesting! Are there tests that fail if `self` isn't released before the end of the callback?

src/cpp/common/version_cc.cc
```diff
@@ -19,7 +19,7 @@
 #include <string>
 
 #include <grpcpp/grpcpp.h>
-#include <grpcpp/support/config.h>
```
drfloob, line 22: This is an autogenerated file, the template would need to be changed instead. But it is unrelated to this PR regardless, so should be rolled back.

src/core/lib/transport/handshaker.cc
```diff
@@ -58,16 +58,20 @@ std::string HandshakerArgsString(HandshakerArgs* args) {
 
 }  // namespace
 
-HandshakeManager::HandshakeManager() {}
+HandshakeManager::HandshakeManager()
+    : RefCounted(GRPC_TRACE_FLAG_ENABLED(grpc_handshaker_trace)
+                     ? "HandshakeManager"
+                     : nullptr),
+      event_engine_(GetDefaultEventEngine()) {}
```
drfloob, line 65: Using `GetDefaultEventEngine()` is a great start to prove the migration will work! I did a quick investigation, and I think ultimately (before we merge the PR) the right thing to do here may be to pass an `std::shared_ptr<EventEngine>` argument in the Handshakemanager constructor. The call sites that construct a HandshakeManager will likely have an instance available in their channel_args.

drfloob (commented):
This is a great start, Yijie! I left a few comments, please let me know if you have any questions.

src/cpp/common/version_cc.cc
yijiem: Restored.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -27,7 +25,6 @@
 #include <utility>
 #include <vector>
 
-#include "absl/meta/type_traits.h"
```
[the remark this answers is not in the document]
yijiem: yeah, I reverted the change and ran `tools/distrib/iwyu.sh` again and now I got a different set of results...

src/core/lib/transport/handshaker.cc
```diff
@@ -16,32 +16,32 @@
...
 #include <grpc/impl/codegen/grpc_types.h>
 #include <grpc/slice_buffer.h>
 #include <grpc/support/alloc.h>
 #include <grpc/support/log.h>
+#include <string>
+#include <utility>
 
+#include "absl/status/status.h"
+#include "absl/strings/str_format.h"
+#include "handshaker.h"
```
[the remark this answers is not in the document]
yijiem: ah, my bad, not sure how it went in there... good catch!

src/core/lib/transport/handshaker.cc
yijiem: This is mostly based on the code here: https://github.com/grpc/grpc/blob/05ab5c08a468a1555b07fd9a70cb9edc5bb849d7/src/core/ext/filters/client_channel/subchannel.cc#L928

The closure holds a ref to the HandshakeManager and when it is destructed it triggers a series of destruction steps on the event_engine thread which may need an ExecCtx to be live on the thread. Thus, self.reset() triggers this process before the previous ExecCtx is destructed. Otherwise we may see this crash in one of the tests: https://paste.googleplex.com/4627665280565248

src/core/lib/transport/handshaker.cc
yijiem: Thanks for the hint! This really seems to work!

src/core/lib/transport/handshaker.cc
yijiem: Wrong paste link sorry... This should be the one: https://paste.googleplex.com/5265162796793856

drfloob: Got it, thank you!

drfloob (approved):
This looks good, thanks Yijie! Please let me know when you have the cherrypick ready.

src/core/lib/transport/handshaker.h
```diff
@@ -97,16 +100,18 @@ class Handshaker : public RefCounted<Handshaker> {
 
 class HandshakeManager : public RefCounted<HandshakeManager> {
  public:
-  HandshakeManager();
+  explicit HandshakeManager(
+      std::shared_ptr<grpc_event_engine::experimental::EventEngine>
```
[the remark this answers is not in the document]
drfloob: That could work, the args should still be the same when cancellation needs to occur later. And it would mean not having to do a cherrypick. FYI Yijie, if `args_.args` are reset as on line 117, then it won't contain the EventEngine, only preconditioned ChannelArgs are guaranteed to have them.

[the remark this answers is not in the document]
markdroth: We can still save the shared_ptr as a data member of `HandshakeManager`.  That way, we don't have to worry about channel args being manipulated by one of the handshakers and potentially pointing to a different EE instance later when we want to cancel the timer.

All I'm really arguing here is that instead of passing the EE instance to the ctor, we can just set it from channel args in `DoHandshake()`.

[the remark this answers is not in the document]
yijiem: sounds good, though the requirement of having a valid EventEngine object in ChannelArgs is not quite obvious this way vs passing from the ctor. It could make e.g. future test writer's job a bit harder if they see a crash. Also not sure if we want to be defensive here to check if event_engine_ is valid or not before use.

src/core/lib/transport/handshaker.cc
```diff
@@ -58,16 +60,20 @@ std::string HandshakerArgsString(HandshakerArgs* args) {
 
 }  // namespace
 
-HandshakeManager::HandshakeManager() {}
+HandshakeManager::HandshakeManager(std::shared_ptr<EventEngine> event_engine)
+    : RefCounted(GRPC_TRACE_FLAG_ENABLED(grpc_handshaker_trace)
+                     ? "HandshakeManager"
+                     : nullptr),
+      event_engine_(event_engine) {}
```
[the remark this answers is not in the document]
yijiem: obsolete, but thanks for catching it! I will remember it next time :)

src/core/lib/transport/handshaker.cc
```diff
@@ -193,6 +190,7 @@ void HandshakeManager::DoHandshake(grpc_endpoint* endpoint,
       grpc_slice_buffer_swap(args_.read_buffer,
                              &(acceptor->pending_data->data.raw.slice_buffer));
     }
+    event_engine_ = args_.args.GetObjectRef<EventEngine>();
```
markdroth, line 193: Suggest moving this down to line 203, right before we actually use it, just so all of the EE-related code is in one place.

test/core/security/ssl_server_fuzzer.cc
```diff
@@ -97,9 +102,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     auto handshake_mgr =
         grpc_core::MakeRefCounted<grpc_core::HandshakeManager>();
     sc->add_handshakers(grpc_core::ChannelArgs(), nullptr, handshake_mgr.get());
-    handshake_mgr->DoHandshake(mock_endpoint, grpc_core::ChannelArgs(),
-                               deadline, nullptr /* acceptor */,
-                               on_handshake_done, &state);
+    grpc_core::ChannelArgs channel_args;
```
markdroth, line 105: Suggest combining this line and the next one.

test/core/security/ssl_server_fuzzer.cc
```diff
@@ -97,9 +102,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     auto handshake_mgr =
         grpc_core::MakeRefCounted<grpc_core::HandshakeManager>();
     sc->add_handshakers(grpc_core::ChannelArgs(), nullptr, handshake_mgr.get());
```
markdroth, line 104: Suggest moving this line right before the call to `DoHandshake()` and using the same channel args for both this call and that one, since that's more consistent with how these APIs will be used in the production code.

src/core/lib/transport/handshaker.h
```diff
@@ -122,37 +125,39 @@ class HandshakeManager : public RefCounted<HandshakeManager> {
...
+      handshakers_ ABSL_GUARDED_BY(mu_);
   // The index of the handshaker to invoke next and closure to invoke it.
-  size_t index_ = 0;
-  grpc_closure call_next_handshaker_;
+  size_t index_ ABSL_GUARDED_BY(mu_) = 0;
+  grpc_closure call_next_handshaker_ ABSL_GUARDED_BY(mu_);
   // The acceptor to call the handshakers with.
-  grpc_tcp_server_acceptor* acceptor_;
+  grpc_tcp_server_acceptor* acceptor_ ABSL_GUARDED_BY(mu_);
   // Deadline timer across all handshakers.
```
markdroth, line 152: Suggest moving these 3 lines down to line 159, so it's right next to the EE data member.

markdroth (approved):
This looks great!

A few minor remaining suggestions for organization, but otherwise this looks ready to merge!

PR #31565, opened 2022-11-05
C++ize TracedBuffer Interface and eliminate lock contention during getsockopt operation

Vignesh2208:
The iomgr TracedBuffer interface is modified to use C++ constructs so that it essentially looks similar to the traced buffer list used by posix event engine.

TracedBuffer::AddNewentry function is also modified to avoid holding a mutex when invoking getsockopt syscall

### DIFF
--- a/src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
+++ b/src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
@@ -28,6 +28,7 @@
 #include <grpc/support/log.h>
 #include <grpc/support/time.h>
 
+#include "src/core/lib/gprpp/sync.h"
 #include "src/core/lib/iomgr/port.h"
 
 #ifdef GRPC_LINUX_ERRQUEUE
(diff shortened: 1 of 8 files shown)


src/core/lib/iomgr/buffer_list.h
```diff
@@ -100,54 +100,63 @@ struct Timestamps {
...
+    uint32_t seq_no_; /* The sequence number for the last byte in the buffer */
+    void* arg_;       /* The arg to pass to timestamps_callback */
+    Timestamps ts_;   /* The timestamps corresponding to this buffer */
+  };
+  Mutex mu_;
+  // TracedBuffers are ordered by sequence number and would need to be processed
+  // in a FIFO order starting with the smallest sequence number. To enable this,
+  // they are stored in a std::list which allows easy appends and forward
+  // iteration operations.
+  std::list<TracedBuffer> buffer_list_;
```
soheilhy, line 147: List is doubly-linked list and is gonna have more overheads compared to intrusive. 

If we use intrusive list, we can avoid the expensive copy of tcp info (see my comment above).

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
```diff
@@ -198,23 +199,27 @@ void ExtractOptStatsFromCmsg(ConnectionMetrics* metrics,
...
   new_elem.ts_.scheduled_time.time = gpr_inf_past(GPR_CLOCK_REALTIME);
   new_elem.ts_.sent_time.time = gpr_inf_past(GPR_CLOCK_REALTIME);
   new_elem.ts_.acked_time.time = gpr_inf_past(GPR_CLOCK_REALTIME);
-
   if (GetSocketTcpInfo(&new_elem.ts_.info, fd) == 0) {
     ExtractOptStatsFromTcpInfo(&new_elem.ts_.sendmsg_time.metrics,
                                &new_elem.ts_.info);
   }
+  grpc_core::MutexLock lock(&mu_);
+  buffer_list_.push_back(new_elem);
```
soheilhy, line 213: `new_elem.ts_.info` is a a very large structure and is not trivially copy-able.  Can we make return `next_` to `TraceBuffer`, allocate it on the heap and make TracedBufferList use an intrusive list approach to avoid this copy?

soheilhy (commented):
Thank you for the quick turnaround!

src/core/lib/iomgr/buffer_list.h
Vignesh2208: Thanks. Updated it to use a singly linked list containing a head and tail ptr for quick append and iteration.

src/core/lib/event_engine/posix_engine/traced_buffer_list.cc
Vignesh2208: fixed.

soheilhy (approved):
Looks great. Thanks!

Vignesh2208:
Thanks for the review!

PR #31567, opened 2022-11-07
Fix PHP distribtest on monterey

jtattermusch:
This is a prerequisite to switching php distribtests job to kokoro monterey workers.
- make sure the PHP distribtests work fine on MacOS monterey workers as well (see b/257261061#comment4)
- do some cleanup to improve the readability of the test scripts (e.g. make it clear what's the purpose of find and get rid of a complicated piped command).

With the setup as-is, the distribtests should be passing on both macos mojave and macos monterey.


jtattermusch:
Passing adhoc run on macos monterey: 
https://source.cloud.google.com/results/invocations/783f906d-004a-47a3-8952-5e601fe6910f

stanley-cheung (approved):
Thanks for taking care of this!

PR #31568, opened 2022-11-07
Switch PHP distribtests to php7.4 (with significant speedup)

jtattermusch:
Based on https://github.com/grpc/grpc/pull/31567.
Leads to significant speedup on the kokoro monterey workers, as they come with php7.4 pre-installed.
The simplification of the prepare script is also nice.


stanley-cheung (approved):
LGTM. Thanks

jtattermusch:
FTR The duration of the grpc_php_distribtests job dropped from ~30mins to ~12min thanks to this change and https://github.com/grpc/grpc/pull/31567

PR #31571, opened 2022-11-07
ObservabilityLogging: Add interface for logging and config parsing implementation for GCP observability

yashykt:


src/cpp/ext/filters/logging/logging_sink.h
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+   private:
+    uint32_t max_metadata_bytes_;
+    uint32_t max_message_bytes_;
+  };
+
+  virtual ~LoggingSink() = default;
+
+  virtual Config FindMatch(bool is_client, absl::string_view path) = 0;
```
veblush, line 55: How about having `optional<Config>` to distinguish these two cases (logging or not logging)? And then it'd be clearer. (e.g. we don't need to have `ShouldLog()` function anymore)

test/cpp/ext/gcp/observability_logging_sink_test.cc
```diff
@@ -0,0 +1,306 @@
...
+}
+
+TEST(GcpObservabilityLoggingSinkTest, LoggingConfigClientWildCardEntries) {
+  const char* json_str = R"json({
+      "cloud_logging": {
+        "client_rpc_events": [
+          {
+            "methods": ["*"],
+            "max_metadata_bytes": 4096,
+            "max_message_bytes": 4096
```
veblush, line 54: nit: how about having different values for max_metadata_bytes and max_message_bytes like 1024 and 4096 respectively so that LoggingSink::Config(1024, 4096) looks easy to understand what's for what.

src/cpp/ext/filters/logging/logging_sink.h
yashykt: There are two fields - metadata and message which can different logging configurations, and as per go/totw/123#when-not-to-use-stdoptional, we shouldn't be using optional if we already have a distinguished value for such cases.

test/cpp/ext/gcp/observability_logging_sink_test.cc
yashykt: done

src/cpp/ext/filters/logging/logging_sink.h
veblush: Fair point.

veblush (approved):

PR #31574, opened 2022-11-07
Align print format with actual argument types

riazantsev:
Specify the correct print types to align with actual arguments.

Signed-off-by: Volodymyr Riazantsev <riazantsevv@google.com>


eugeneo:
Another fix was merged: https://github.com/grpc/grpc/pull/31576

PR #31580, opened 2022-11-08
Use correct namespace for checking if the isDefaultRootsPemSet method exists.

frankdejonge:
The `method_exists` function requires a fully qualified class name to be sent to check if a method exists. The current class was missing the namespace, which means the function always returns `false`. In our application this caused the credentials to be loaded many times over, which ate up some CPU. This bug fix ensures that this is only run once per request.

I'm figuring out how to sign the CLA, please bear with me :)

### DIFF
--- a/src/php/lib/Grpc/BaseStub.php
+++ b/src/php/lib/Grpc/BaseStub.php
@@ -43,7 +43,7 @@ class BaseStub
      */
     public function __construct($hostname, $opts, $channel = null)
     {
-        if (!method_exists('ChannelCredentials', 'isDefaultRootsPemSet') ||
+        if (!method_exists('Grpc\ChannelCredentials', 'isDefaultRootsPemSet') ||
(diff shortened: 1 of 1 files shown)


barrycommins:
Hi, is there any possibility of getting this reviewed?

`method_exists('ChannelCredentials', 'isDefaultRootsPemSet')` always returns `false` because `ChannelCredentials` is in the `Grpc` namespace and not the default namespace.

dunglas (approved):

frankdejonge:
@stanley-cheung sorry for the ping, I'm trying to fix a bug in the client library but I have no clue how to push this forward. Do you have any advice?

alto-php:
Thanks for the fix, any chance you can add a test for it as well?

frankdejonge:
Unless you want to go for static analysis, there is not really test case for this 😅

alto-php (approved):

PR #31586, opened 2022-11-08
explain that keys must be lower cased

ijrsvt:
* Upper case labels can cause the metadata to be rejected by certain servers


XuanWang-Amos:
Thanks for adding the change, we're indeed doesn't support upper case as metadata key: https://github.com/grpc/grpc/blob/master/src/core/lib/surface/validate_metadata.cc#L65-L71

XuanWang-Amos (approved):

PR #31588, opened 2022-11-09
Add a check to flush stale TracedBuffer entries to prevent them growing too long.

Vignesh2208:
By default, any traced buffer entry more than 10secs old is considered stale. 10sec value is hard-coded for now.

### DIFF
--- a/src/core/lib/iomgr/buffer_list.cc
+++ b/src/core/lib/iomgr/buffer_list.cc
@@ -20,8 +20,8 @@
 
 #include "src/core/lib/iomgr/buffer_list.h"
 
-#include "grpc/support/time.h"
 #include <grpc/support/log.h>
+#include <grpc/support/time.h>
 
 #include "src/core/lib/gprpp/global_config.h"
 #include "src/core/lib/gprpp/sync.h"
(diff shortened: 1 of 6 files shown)


src/core/lib/iomgr/buffer_list.cc
```diff
@@ -30,6 +32,10 @@
 #include <string.h>
 #include <time.h>
 
+GPR_GLOBAL_CONFIG_DEFINE_INT32(
```
ctiller, line 35: @soheilhy any experience on why folks would change this?

yousukseung: I think it's fine to hard-code.

I believe this CL was designed after cl/331007927, then the default value (10s) has been working fine and I'm not aware of any case that overrides. At that time this value was chosen arbitrarily and I needed a control knob.

soheilhy (changes requested):

src/core/lib/iomgr/buffer_list.cc
Vignesh2208: Removed the flag now and hard-coded it.

PR #31592, opened 2022-11-09
Promise-based Endpoint

ralphchung:


include/grpc/event_engine/promise_endpoint.h
```diff
@@ -0,0 +1,77 @@
...
+
+class PromiseEndpoint {
+ public:
+  PromiseEndpoint(
+      std::unique_ptr<grpc_event_engine::experimental::EventEngine::Endpoint>
+          endpoint,
+      grpc_event_engine::experimental::SliceBuffer already_received);
+  ~PromiseEndpoint();
+
+  grpc_core::Promise<absl::Status> Write(
```
ctiller, line 43: We'll want this to be a concrete type eventually, not the type-erased Promise<> wrapper.

We can do this one of two ways:
Define a c++98 functor with a polling operator() - the existing pipe.h does this, or use the example in map_pipe.h and leverage c++14 auto return types (I think my preferred)

One factoring technique that may be useful is to provide PollXyz() private methods that the lambdas can call out to:

```
auto Read(size_t length) {
  StartRead(length);
  return [this] { PollReadComplete(); }
}
```

include/grpc/event_engine/promise_endpoint.h
```diff
@@ -0,0 +1,77 @@
...
+      grpc_event_engine::experimental::SliceBuffer data);
+  grpc_core::Promise<
+      absl::StatusOr<grpc_event_engine::experimental::SliceBuffer>>
```
ctiller, line 46: These types should be the `grpc_core::{Slice,SliceBuffer}` types (which share the same underlying type)

ctiller (commented):
I've only reviewed interface, but hopefully this helps

src/core/lib/transport/promise_endpoint.cc
```diff
@@ -0,0 +1,61 @@
...
+#include <grpc/support/log.h>
+
+namespace grpc {
+
+namespace internal {
+
+PromiseEndpoint::PromiseEndpoint(
+    std::unique_ptr<grpc_event_engine::experimental::EventEngine::Endpoint>
+        endpoint,
+    grpc_core::SliceBuffer already_received)
```
Vignesh2208, line 32: Whats the purpose of this already_received SliceBuffer parameter ?

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+
+class PromiseEndpoint {
+ public:
+  PromiseEndpoint(
+      std::unique_ptr<grpc_event_engine::experimental::EventEngine::Endpoint>
+          endpoint,
+      grpc_core::SliceBuffer already_received);
+  ~PromiseEndpoint();
+
+  auto Write(grpc_core::SliceBuffer data) {
```
Vignesh2208, line 58: grpc_core::SliceBuffer may evolve over time to be different from grpc_event_engine::experimental::SliceBuffer. I think the input arguments to Read(..) and Write(..) must be of type grpc_event_engine::experimental::SliceBuffer

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+    grpc_core::MutexLock lock(&write_mutex_);
+
+    /// Previous write result has not been polled.
+    GPR_ASSERT(!write_result_.has_value());
+
+    /// TODO: Is there a better way to convert?
+    grpc_slice_buffer_swap(data.c_slice_buffer(),
```
Vignesh2208, line 65: grpc_event_engine::experimental::SliceBuffer class already has a Swap method defined. You can use that instead of swapping the grpc_slice_buffer c types

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+  auto Read(size_t num_bytes) {
+    grpc_core::MutexLock lock(&read_mutex_);
+
+    /// Previous read result has not been polled.
+    GPR_ASSERT(!read_result_.has_value());
+
+    /// Should not have pending reads
+    GPR_ASSERT(pending_read_buffer_.Count() == 0u);
+
+    if (read_buffer_.Count() < num_bytes) {
```
Vignesh2208, line 99: Count() method returns number of slices rather than number of bytes in the buffer. Perhaps you meant to use Length() method ?

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+          [this, num_bytes, &read_callback](absl::Status status) {
+            if (!status.ok()) {
+              pending_read_buffer_.Clear();
+              temporary_read_buffer_.Clear();
+
+              grpc_core::MutexLock lock(&read_mutex_);
+              read_result_ = status;
+              read_waker_.Wakeup();
+            } else {
+              while (temporary_read_buffer_.Count() > 0u) {
```
Vignesh2208, line 110: Again perhaps Count() should be replaced by Length() ?

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+                pending_read_buffer_.Append(temporary_read_buffer_.TakeFirst());
+              }
+
+              if (read_buffer_.Count() + pending_read_buffer_.Count() <
```
Vignesh2208, line 114: Again perhaps Count() should be replaced by Length() ?

src/core/lib/transport/promise_endpoint.h
Vignesh2208, line 110: Instead of this loop, you can use the MoveFirstNBytesIntoSliceBuffer method defined in grpc_event_engine::experimental::SliceBuffer

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+                }
+
+                grpc_core::MutexLock lock(&read_mutex_);
+                read_result_ = status;
+                read_waker_.Wakeup();
+              }
+            }
+          };
+      temporary_read_buffer_.Clear();  /// may be redundant
+      endpoint_->Read(read_callback, &temporary_read_buffer_,
```
Vignesh2208, line 133: read_callback appears to be a local variable defined inside the if (read_buffer_.Count() < num_bytes) { ...} block and it also gets called recursively inside the read_callback. I think this might cause heap use-after free issues

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+        grpc_slice_buffer_move_first(read_buffer_.c_slice_buffer(), num_bytes,
+                                     ret.c_slice_buffer());
+
+        read_result_.reset();
+        return ret;
+      }
+    };
+  }
+
+  auto ReadSlice(size_t num_bytes) {
```
Vignesh2208, line 161: Is it possible to re-use some code from Read() implementation inside this function ?

Vignesh2208 (changes requested):
Made a first pass. Thanks for working on this!

src/core/lib/transport/promise_endpoint.cc
ralphchung: Please refer to the design doc ([internal link](https://docs.google.com/document/d/1vRy0yse-d1heLQRmLPo_0figsTPXJAnNN84tBCAne_s/edit?resourcekey=0-JvUPdq0LaZq8gMkgT9Pzlw#heading=h.qgvc5vr55ytg)).

src/core/lib/transport/promise_endpoint.h
ralphchung: Craig and I have discussed this. Please refer to the comments above.

src/core/lib/transport/promise_endpoint.h
ralphchung: Same as above.

src/core/lib/transport/promise_endpoint.h
ralphchung: I have thought about that. But I think there is some optimization we can do in this special case, so I chose not to.

src/core/lib/transport/promise_endpoint.h
ralphchung: This could be answer to why I cannot run other test cases locally (I did not push it since it fails.). My original thought on this is that `endpoint_->Read()` takes `read_callback` by value, so it should be ok to do this?

src/core/lib/transport/promise_endpoint.h
ralphchung: I think the reason for this is that I cannot use `grpc_event_engine::experimental::SliceBuffer`.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+                pending_read_buffer_.Append(temporary_read_buffer_.TakeFirst());
+              }
+
+              if (read_buffer_.Count() + pending_read_buffer_.Count() <
+                  num_bytes) {
+                /// A further read is needed.
+                endpoint_->Read(read_callback, &temporary_read_buffer_,
+                                nullptr /* uses default arguments */);
+              } else {
+                while (pending_read_buffer_.Count() > 0u) {
```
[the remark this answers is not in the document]
ralphchung: I think the reason for this is that I cannot use `grpc_event_engine::experimental::SliceBuffer`.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+        read_waker_ = grpc_core::Activity::current()->MakeNonOwningWaker();
+        return grpc_core::Pending();
+      } else if (!read_result_->ok()) {
+        const absl::Status ret = *read_result_;
+        read_result_.reset();
+
+        return ret;
+      } else {
+        grpc_core::SliceBuffer ret;
+        grpc_slice_buffer_move_first(read_buffer_.c_slice_buffer(), num_bytes,
```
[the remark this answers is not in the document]
ralphchung: I think the reason for this is that I cannot use `grpc_event_engine::experimental::SliceBuffer`.

src/core/lib/transport/promise_endpoint.h
ctiller: see #31831 for an approach I'm exploring (will want to get `SliceBufferCast` done next)

src/core/lib/transport/promise_endpoint.h
ctiller: You could use `grpc_event_engine::experimental::SliceBuffer` internally and convert at the end if that helps (and I think we'll be able to make that final conversion 'free' soon)

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+    grpc_core::MutexLock lock(&read_mutex_);
+
+    /// Previous read result has not been polled.
+    GPR_ASSERT(!read_result_.has_value());
+
+    /// Should not have pending reads
+    GPR_ASSERT(pending_read_buffer_.Count() == 0u);
+
+    if (read_buffer_.Length() < num_bytes) {
+      const std::function<void(absl::Status)> read_callback =
```
ctiller, line 100: const auto (no need for type erasure here)

also... consider moving `num_bytes`, `read_callback` into members on the `PromiseBasedEndpoint`: storing more than one pointer on a callback will force a memory allocation when we call `Endpoint::Read` and convert this callback into an `AnyInvocable`.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+    GPR_ASSERT(!read_result_.has_value());
+
+    if (read_buffer_.Length() < num_bytes) {
+      /// TODO: handle potential integer overflow
+      /// TODO: evaluate the lifespan of `read_args`
+      const struct grpc_event_engine::experimental::EventEngine::Endpoint::
+          ReadArgs read_args = {static_cast<int64_t>(num_bytes)};
+
+      const std::function<void(absl::Status)> read_callback =
+          [this, num_bytes, read_args, &read_callback](absl::Status status) {
```
ctiller, line 174: const auto (no need for type erasure here)

also... consider moving `num_bytes`, `read_callback` into members on the `PromiseBasedEndpoint`: storing more than one pointer on a callback will force a memory allocation when we call `Endpoint::Read` and convert this callback into an `AnyInvocable`.

src/core/lib/transport/promise_endpoint.h
ctiller: Worth considering whether the aggregation code can be moved into `SliceBuffer` however (although I'll not insist that happens)

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+        grpc_slice_buffer_move_first(read_buffer_.c_slice_buffer(), num_bytes,
+                                     ret.c_slice_buffer());
+
+        read_result_.reset();
+        return ret;
+      }
+    };
+  }
+
+  auto ReadSlice(size_t num_bytes) {
```
ctiller, line 161: Add comment: // Returns a promise that resolves to `grpc_core::Slice`

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+        return grpc_core::Pending();
+      } else {
+        const auto ret = *write_result_;
+        write_result_.reset();
+        return ret;
+      }
+    };
+  }
+
+  auto Read(size_t num_bytes) {
```
ctiller, line 90: Add comment: // Returns a promise that resolves to `grpc_core::SliceBuffer`

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+        grpc_core::MutableSlice ret =
+            grpc_core::MutableSlice::CreateUninitialized(num_bytes);
+        read_buffer_.MoveFirstNBytesIntoBuffer(num_bytes, ret.data());
+
+        return grpc_core::Slice(std::move(ret));
+      }
+    };
+  }
+
+  auto ReadByte() {
```
ctiller, line 235: Comment: // Returns a promise that resolves to a uint8_t

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,315 @@
...
+    grpc_core::MutexLock lock(&read_mutex_);
+
+    /// Previous read result has not been polled.
+    GPR_ASSERT(!read_result_.has_value());
+
+    if (read_buffer_.Length() == 0u) {
+      const std::function<void(absl::Status)> read_callback =
```
ctiller, line 242: `const auto read_callback = `

PR #26570, opened 2021-06-28
Increase perAttemptRecvTimeout to avoid flakiness.

markdroth:
This fixes a flake that was occurring under MSAN because the original attempt hit the perAttemptRecvTimeout and therefore got cancelled before the request ever actually went out on the wire, and therefore the first attempt seen by the server was actually the second attempt, so the assertion that the request did not have the `grpc-previous-rpc-attempts` header was failing.  Here's an example of the failure:

https://source.cloud.google.com/results/invocations/7b741380-acb1-418b-a226-8f1f18d20024/targets/%2F%2Ftest%2Fcore%2Fend2end:h2_ssl_proxy_test@retry_per_attempt_recv_timeout@poller%3Dpoll/log

Increasing the perAttemptRecvTimeout seems to give the first request enough time to go out on the wire, even under MSAN.  The test now passes when run 1000x on RBE under MSAN.

PR #31605, opened 2022-11-10
Try fix @retry_per_attempt_recv_timeout flake on MacOS

jtattermusch:
See b/258191715.

Basically, it seems that the mac flake is a side effect of another fix: https://github.com/grpc/grpc/pull/26570:

- perAttemptRecvTimeout was increased by the PR 1sec to 2sec
- that means that the second and third call only start being processed 3 seconds after the start of the call.
- since the call on the client side has a timeout of 5 seconds, the test only has 2 seconds to finish the second and third call and that's sometimes not enough on MacOS (mac tests are more prone to timing issue since where there can be other tests running in parallel and the machines are less beefy than what we have on linux/windows).

The proposed fix is to also increase the total timeout of the call (proportionally to the increased perAttemptRecvTimeout) , so that there's more time left to finish second and third call.

### DIFF
--- a/test/core/end2end/tests/retry_per_attempt_recv_timeout.cc
+++ b/test/core/end2end/tests/retry_per_attempt_recv_timeout.cc
@@ -151,7 +151,7 @@ static void test_retry_per_attempt_recv_timeout(
 
   grpc_core::CqVerifier cqv(f.cq);
 
-  gpr_timespec deadline = five_seconds_from_now();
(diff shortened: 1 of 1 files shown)


jtattermusch:
To test the fix, I ran 3 batches of 2000 test runs with an ephemeral kokoro job and I wasn't able to reproduce anymore (and I was able to reproduce without this fix).

markdroth (approved):
Thanks for fixing this!

PR #31607, opened 2022-11-10
Revived google-readability-casting

veblush:
Revived `google-readability-casting` which was disabled temporarily by https://github.com/grpc/grpc/pull/31518. It's still acting up on some edge cases but it's easy to tame using brace initialization.

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/subchannel_list.h
+++ b/src/core/ext/filters/client_channel/lb_policy/subchannel_list.h
@@ -324,8 +324,8 @@ void SubchannelData<SubchannelListType,
   pending_watcher_ =
       new Watcher(this, subchannel_list()->WeakRef(DEBUG_LOCATION, "Watcher"));
   subchannel_->WatchConnectivityState(
-      std::unique_ptr<SubchannelInterface::ConnectivityStateWatcherInterface>(
-          pending_watcher_));
+      std::unique_ptr<SubchannelInterface::ConnectivityStateWatcherInterface>{
+          pending_watcher_});
 }
 
 template <typename SubchannelListType, typename SubchannelDataType>
--- a/src/core/ext/filters/client_channel/lb_policy/subchannel_list.h
+++ b/src/core/ext/filters/client_channel/lb_policy/subchannel_list.h
@@ -324,8 +324,9 @@ void SubchannelData<SubchannelListType,
   pending_watcher_ =
       new Watcher(this, subchannel_list()->WeakRef(DEBUG_LOCATION, "Watcher"));
   subchannel_->WatchConnectivityState(
-      std::unique_ptr<SubchannelInterface::ConnectivityStateWatcherInterface>{
-          pending_watcher_});
+      // NOLINTNEXTLINE(google-readability-casting)
+      std::unique_ptr<SubchannelInterface::ConnectivityStateWatcherInterface>(
+          pending_watcher_));
 }
 
 template <typename SubchannelListType, typename SubchannelDataType>
--- a/src/core/ext/transport/chttp2/transport/flow_control.cc
+++ b/src/core/ext/transport/chttp2/transport/flow_control.cc
@@ -255,8 +255,7 @@ TransportFlowControl::TargetInitialWindowSizeBasedOnMemoryPressureAndBdp()
   //                                                                pressure
   const double kAnythingGoesPressure = 0.2;
   const double kAdjustedToBdpPressure = 0.5;
-  const double kAnythingGoesWindow =
-      std::max(static_cast<double>(1 << 24), bdp);
+  const double kAnythingGoesWindow = std::max(double{1 << 24}, bdp);
   if (memory_pressure < kAnythingGoesPressure) {
     return kAnythingGoesWindow;
   } else if (memory_pressure < kAdjustedToBdpPressure) {
--- a/src/core/ext/transport/chttp2/transport/flow_control.cc
+++ b/src/core/ext/transport/chttp2/transport/flow_control.cc
@@ -122,7 +122,7 @@ uint32_t TransportFlowControl::MaybeSendUpdate(bool writing_anyway) {
       announced_window_ != target_announced_window) {
     const uint32_t announce =
         static_cast<uint32_t>(Clamp(target_announced_window - announced_window_,
-                                    int64_t(0), kMaxWindowUpdateSize));
+                                    int64_t{0}, kMaxWindowUpdateSize));
     announced_window_ += announce;
     return announce;
   }
@@ -255,7 +255,8 @@ TransportFlowControl::TargetInitialWindowSizeBasedOnMemoryPressureAndBdp()
   //                                                                pressure
   const double kAnythingGoesPressure = 0.2;
   const double kAdjustedToBdpPressure = 0.5;
-  const double kAnythingGoesWindow = std::max(double(1 << 24), bdp);
+  const double kAnythingGoesWindow =
+      std::max(static_cast<double>(1 << 24), bdp);
   if (memory_pressure < kAnythingGoesPressure) {
     return kAnythingGoesWindow;
   } else if (memory_pressure < kAdjustedToBdpPressure) {
@@ -347,23 +348,24 @@ FlowControlAction TransportFlowControl::PeriodicUpdate() {
       }
       // Though initial window 'could' drop to 0, we keep the floor at
       // kMinInitialWindowSize
-      UpdateSetting(
-          GRPC_CHTTP2_SETTINGS_INITIAL_WINDOW_SIZE,
-          &target_initial_window_size_,
-          static_cast<int32_t>(Clamp(target, double(kMinInitialWindowSize),
-                                     double(kMaxInitialWindowSize))),
-          &action, &FlowControlAction::set_send_initial_window_update);
+      UpdateSetting(GRPC_CHTTP2_SETTINGS_INITIAL_WINDOW_SIZE,
+                    &target_initial_window_size_,
+                    static_cast<int32_t>(Clamp(
+                        target, static_cast<double>(kMinInitialWindowSize),
+                        static_cast<double>(kMaxInitialWindowSize))),
+                    &action,
+                    &FlowControlAction::set_send_initial_window_update);
       // get bandwidth estimate and update max_frame accordingly.
       double bw_dbl = bdp_estimator_.EstimateBandwidth();
       // we target the max of BDP or bandwidth in microseconds.
       UpdateSetting(
           GRPC_CHTTP2_SETTINGS_MAX_FRAME_SIZE, &target_frame_size_,
-          static_cast<int32_t>(Clamp(
-              std::max(
-                  static_cast<int32_t>(Clamp(bw_dbl, 0.0, double(INT_MAX))) /
-                      1000,
-                  static_cast<int32_t>(target_initial_window_size_)),
-              16384, 16777215)),
+          static_cast<int32_t>(
+              Clamp(std::max(static_cast<int32_t>(Clamp(
+                                 bw_dbl, 0.0, static_cast<double>(INT_MAX))) /
+                                 1000,
+                             static_cast<int32_t>(target_initial_window_size_)),
+                    16384, 16777215)),
           &action, &FlowControlAction::set_send_max_frame_size_update);
     }
 
@@ -409,7 +411,7 @@ int64_t StreamFlowControl::DesiredAnnounceSize() const {
       return std::min(min_progress_size_, kMaxWindowDelta);
     }
   }();
-  return Clamp(desired_window_delta - announced_window_delta_, int64_t(0),
+  return Clamp(desired_window_delta - announced_window_delta_, int64_t{0},
                kMaxWindowUpdateSize);
 }
 
@@ -423,9 +425,8 @@ FlowControlAction StreamFlowControl::UpdateAction(FlowControlAction action) {
     // Currently set at half the initial window size or 8kb (whichever is
     // greater). 8kb means we don't send rapidly unnecessarily when the initial
     // window size is small.
-    const int64_t hurry_up_size =
-        std::max(static_cast<int64_t>(tfc_->sent_init_window()) / 2,
-                 static_cast<int64_t>(8192));
+    const int64_t hurry_up_size = std::max(
+        static_cast<int64_t>(tfc_->sent_init_window()) / 2, int64_t{8192});
     if (desired_announce_size > hurry_up_size) {
       urgency = FlowControlAction::Urgency::UPDATE_IMMEDIATELY;
     }
--- a/src/core/ext/xds/xds_listener.cc
+++ b/src/core/ext/xds/xds_listener.cc
@@ -628,8 +628,8 @@ absl::optional<XdsListenerResource::FilterChainMap::CidrRange> CidrRangeParse(
         google_protobuf_UInt32Value_value(prefix_len_proto),
         (reinterpret_cast<const grpc_sockaddr*>(cidr_range.address.addr))
                     ->sa_family == GRPC_AF_INET
-            ? uint32_t(32)
-            : uint32_t(128));
+            ? static_cast<uint32_t>(32)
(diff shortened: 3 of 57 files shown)


ctiller:
Stating for the record that this seems like a readability regression to me: what's the upside?

src/core/ext/filters/client_channel/lb_policy/subchannel_list.h
```diff
@@ -324,8 +324,8 @@ void SubchannelData<SubchannelListType,
   pending_watcher_ =
       new Watcher(this, subchannel_list()->WeakRef(DEBUG_LOCATION, "Watcher"));
   subchannel_->WatchConnectivityState(
-      std::unique_ptr<SubchannelInterface::ConnectivityStateWatcherInterface>(
-          pending_watcher_));
+      std::unique_ptr<SubchannelInterface::ConnectivityStateWatcherInterface>{
```
markdroth, line 327: Why is this change necessary?  `std::unique_ptr<SubchannelInterface::ConnectivityStateWatcherInterface>` is a class type, so parens should be fine here.  This is not a cast, it's constructing a new object.

src/core/ext/xds/xds_listener.cc
```diff
@@ -628,8 +628,8 @@ absl::optional<XdsListenerResource::FilterChainMap::CidrRange> CidrRangeParse(
         google_protobuf_UInt32Value_value(prefix_len_proto),
         (reinterpret_cast<const grpc_sockaddr*>(cidr_range.address.addr))
                     ->sa_family == GRPC_AF_INET
-            ? uint32_t(32)
-            : uint32_t(128));
+            ? static_cast<uint32_t>(32)
```
markdroth, line 631: I think these two lines can use braces instead of `static_cast<>`.

src/core/lib/address_utils/sockaddr_utils.cc
```diff
@@ -371,7 +371,7 @@ void grpc_sockaddr_mask_bits(grpc_resolved_address* address,
     } else if (mask_bits >= 32) {
       return;
     }
-    uint32_t mask_ip_addr = (~(uint32_t(0))) << (32 - mask_bits);
+    uint32_t mask_ip_addr = (~(static_cast<uint32_t>(0))) << (32 - mask_bits);
```
markdroth, line 374: All of the ocurrances in this file can use braces instead of `static_cast<>`.

src/core/lib/gpr/string.cc
```diff
@@ -89,7 +89,7 @@ static dump_out dump_out_create(void) {
 
 static void dump_out_append(dump_out* out, char c) {
   if (out->length == out->capacity) {
-    out->capacity = std::max(size_t(8), 2 * out->capacity);
+    out->capacity = std::max(static_cast<size_t>(8), 2 * out->capacity);
```
markdroth, line 92: The occurrances in this file can use braces instead of `static_cast<>`.

src/core/lib/http/parser.cc
```diff
@@ -337,7 +337,8 @@ static grpc_error_handle addbyte_body(grpc_http_parser* parser, uint8_t byte) {
   }
 
   if (*body_length == parser->body_capacity) {
-    parser->body_capacity = std::max(size_t(8), parser->body_capacity * 3 / 2);
+    parser->body_capacity =
+        std::max(static_cast<size_t>(8), parser->body_capacity * 3 / 2);
```
markdroth, line 341: This should use braces.

src/core/lib/iomgr/ev_poll_posix.cc
```diff
@@ -1193,7 +1193,7 @@ static void pollset_set_add_pollset(grpc_pollset_set* pollset_set,
   gpr_mu_lock(&pollset_set->mu);
   if (pollset_set->pollset_count == pollset_set->pollset_capacity) {
     pollset_set->pollset_capacity =
-        std::max(size_t(8), 2 * pollset_set->pollset_capacity);
+        std::max(static_cast<size_t>(8), 2 * pollset_set->pollset_capacity);
```
markdroth, line 1196: All the changes in this file should use braces.

src/core/lib/resource_quota/memory_quota.cc
```diff
@@ -371,7 +371,8 @@ void BasicMemoryQuota::Start() {
           std::tuple<const char*, RefCountedPtr<ReclaimerQueue::Handle>> arg) {
         auto reclaimer = std::move(std::get<1>(arg));
         if (GRPC_TRACE_FLAG_ENABLED(grpc_resource_quota_trace)) {
-          double free = std::max(intptr_t(0), self->free_bytes_.load());
+          double free =
+              std::max(static_cast<intptr_t>(0), self->free_bytes_.load());
```
markdroth, line 375: The changes in this file should use braces.

I'll stop noting these individually, but please fix throughout the PR.  Anywhere that the argument is a constant value, it's probably better to use braces than call it a cast.

test/core/transport/binder/binder_transport_test.cc
```diff
@@ -192,8 +192,8 @@ class MetadataEncoder {
   template <typename Which>
   void Encode(Which, const typename Which::ValueType& value) {
     metadata_.emplace_back(
-        std::string(Which::key()),
-        std::string(grpc_core::Slice(Which::Encode(value)).as_string_view()));
+        std::string{Which::key()},
```
markdroth, line 195: `std::string` is a class, so this should not require braces.  It's constructing a new object, not doing a cast.

markdroth (commented):
@ctiller The rationale for this check is in https://google.github.io/styleguide/cppguide.html#Casting.  I agree that the recent changes have made it a little more cumbersome, though.

I do think we need a check to avoid C-style casts, because those do actually cause bugs.  If there was another flag we could use here that would continue to do that check without any of the new requirements that they seem to have added here in the new version of clang, I'd be open to that.  But if we have to choose between enabling this the way it is or leaving it off, I'd rather enable it.

src/core/ext/filters/client_channel/lb_policy/subchannel_list.h
veblush: This is to address this weird false positive of clang-tidy. Clang-tidy keeps saying it should change so I did it to workaround it.

test/core/transport/binder/binder_transport_test.cc
veblush: Like the above case, this is again to address this weird false positive of clang-tidy. There were two cases in total.

veblush:
@markdroth To address your comments; I replaced `static_cast<(\w+)>\((\d+)\)` with `$1{$2}` throughout all gRPC code.

src/core/ext/filters/client_channel/lb_policy/subchannel_list.h
markdroth: This seems obviously wrong.

Can we instead leave it as-is and use a NOLINT comment to have clang-tidy leave it alone?

test/core/transport/binder/binder_transport_test.cc
markdroth: I'd prefer to use a NOLINT comment instead here.  I think using parens is correct here.

test/core/channel/channelz_test.cc
```diff
@@ -449,16 +449,16 @@ TEST_F(ChannelzRegistryBasedTest, GetTopChannelsMoreGaps) {
   Json array = (*parsed_json->mutable_object())["channel"];
   ValidateJsonArraySize(array, 2);
   std::vector<intptr_t> uuids = GetUuidListFromArray(array.array_value());
-  EXPECT_EQ(static_cast<intptr_t>(3), uuids[0]);
-  EXPECT_EQ(static_cast<intptr_t>(5), uuids[1]);
+  EXPECT_EQ(intptr_t{3}, uuids[0]);
```
markdroth, line 452: In all of the cases where we're specifying a type inside of a gtest `EXPECT_*` macro, it's generally better to just use `U` or `UL` suffixes instead of spelling out the specific type.  So for example, in this case, you can change `intptr_t{3}` to `3UL`.  Or if a test currently says `unsigned{1}`, you can change that to `1U`.

Please fix throughout.

ctiller: Is 3UL guaranteed to be intptr_t on all platforms? (I genuinely don't know)

test/core/transport/binder/binder_transport_test.cc
veblush: I don't have a strong opinion here but shouldn't `std::string{}` be a proper construction call?

test/core/channel/channelz_test.cc
markdroth: Good question.  I know that I've always used a `UL` suffix in `EXPECT_*` macros for things like `size_t`, and it's always worked fine, so I assume it will for `intptr_t` too.  But if it causes the test to fail on any platform we care about, we can always switch back to the current syntax for any `intptr_t` values.

test/core/transport/binder/binder_transport_test.cc
markdroth: I believe that calls the initializer_list ctor, not the single-argument ctor.  But there's no reason we should need to do that here.  The braces are ugly and confusing, and they really don't provide any benefit in this case.

test/core/channel/channelz_test.cc
veblush: `EXPECT_EQ` doesn't require the type of two given value to have the exact same so I think it should be fine with just plain 3 here.

markdroth: Historically, I have found these suffixes to frequently be needed to get `EXPECT_*` macros to compile properly on all platforms.  But if it works without it, that's fine with me.

test/core/transport/binder/binder_transport_test.cc
veblush: This could be actually both a regular ctor and initializer_list ctor. Yay another C++ thing! But I'm adding NOLINT comment here.

src/core/ext/transport/chttp2/transport/flow_control.cc
```diff
@@ -255,7 +255,8 @@ TransportFlowControl::TargetInitialWindowSizeBasedOnMemoryPressureAndBdp()
   //                                                                pressure
   const double kAnythingGoesPressure = 0.2;
   const double kAdjustedToBdpPressure = 0.5;
-  const double kAnythingGoesWindow = std::max(double(1 << 24), bdp);
+  const double kAnythingGoesWindow =
+      std::max(static_cast<double>(1 << 24), bdp);
```
markdroth, line 259: This one should probably also use braces instead of `static_cast<>`.

test/core/iomgr/buffer_list_test.cc
```diff
@@ -47,12 +47,12 @@ TEST(BufferListTest, Testshutdownflusheslist) {
...
   gpr_atm verifier_called[NUM_ELEM];
   for (auto i = 0; i < NUM_ELEM; i++) {
-    gpr_atm_rel_store(&verifier_called[i], static_cast<gpr_atm>(0));
+    gpr_atm_rel_store(&verifier_called[i], gpr_atm{0});
     tb_list.AddNewEntry(i, 0, static_cast<void*>(&verifier_called[i]));
   }
   tb_list.Shutdown(nullptr, absl::OkStatus());
   for (auto i = 0; i < NUM_ELEM; i++) {
-    ASSERT_EQ(gpr_atm_acq_load(&verifier_called[i]), static_cast<gpr_atm>(1));
+    ASSERT_EQ(gpr_atm_acq_load(&verifier_called[i]), gpr_atm{1});
```
markdroth, line 55: This can presumably just say `1`, with no explicit type specified.

test/core/iomgr/buffer_list_test.cc
```diff
@@ -82,10 +82,10 @@ TEST(BufferListTest, Testverifiercalledonack) {
       TestVerifierCalledOnAckVerifier);
   grpc_core::TracedBufferList tb_list;
   gpr_atm verifier_called;
-  gpr_atm_rel_store(&verifier_called, static_cast<gpr_atm>(0));
+  gpr_atm_rel_store(&verifier_called, gpr_atm{0});
   tb_list.AddNewEntry(213, 0, &verifier_called);
   tb_list.ProcessTimestamp(&serr, nullptr, &tss);
-  ASSERT_EQ(gpr_atm_acq_load(&verifier_called), static_cast<gpr_atm>(1));
+  ASSERT_EQ(gpr_atm_acq_load(&verifier_called), gpr_atm{1});
```
markdroth, line 88: Same here.

test/core/iomgr/buffer_list_test.cc
```diff
@@ -102,10 +102,10 @@ TEST(BufferListTest, Testrepeatedshutdown) {
...
```
markdroth, line 108: And here.

test/core/transport/chttp2/context_list_test.cc
```diff
@@ -97,13 +97,13 @@ TEST_F(ContextListTest, ExecuteFlushesList) {
...
     s[i]->byte_counter = kByteOffset;
-    gpr_atm_rel_store(&verifier_called[i], static_cast<gpr_atm>(0));
+    gpr_atm_rel_store(&verifier_called[i], gpr_atm{0});
     ContextList::Append(&list, s[i]);
   }
   Timestamps ts;
   ContextList::Execute(list, &ts, absl::OkStatus());
   for (auto i = 0; i < kNumElems; i++) {
-    EXPECT_EQ(gpr_atm_acq_load(&verifier_called[i]), static_cast<gpr_atm>(1));
+    EXPECT_EQ(gpr_atm_acq_load(&verifier_called[i]), gpr_atm{1});
```
markdroth, line 106: And here.

test/core/transport/chttp2/context_list_test.cc
```diff
@@ -151,12 +151,12 @@ TEST_F(ContextListTest, NonEmptyListEmptyTimestamp) {
...
     s[i]->context = &verifier_called[i];
     s[i]->byte_counter = kByteOffset;
-    gpr_atm_rel_store(&verifier_called[i], static_cast<gpr_atm>(0));
+    gpr_atm_rel_store(&verifier_called[i], gpr_atm{0});
     ContextList::Append(&list, s[i]);
   }
   ContextList::Execute(list, nullptr, absl::OkStatus());
   for (auto i = 0; i < kNumElems; i++) {
-    EXPECT_EQ(gpr_atm_acq_load(&verifier_called[i]), static_cast<gpr_atm>(1));
+    EXPECT_EQ(gpr_atm_acq_load(&verifier_called[i]), gpr_atm{1});
```
markdroth, line 159: And here.

test/cpp/common/channel_arguments_test.cc
```diff
@@ -99,7 +99,7 @@ class ChannelArgumentsTest : public ::testing::Test {
   void VerifyDefaultChannelArgs() {
     grpc_channel_args args;
     SetChannelArgs(channel_args_, &args);
-    EXPECT_EQ(static_cast<size_t>(1), args.num_args);
+    EXPECT_EQ(size_t{1}, args.num_args);
```
markdroth, line 102: And here.

test/cpp/util/byte_buffer_test.cc
```diff
@@ -70,7 +70,7 @@ TEST_F(ByteBufferTest, Clear) {
   Slice s(kContent1);
   ByteBuffer buffer(&s, 1);
   buffer.Clear();
-  EXPECT_EQ(static_cast<size_t>(0), buffer.Length());
+  EXPECT_EQ(size_t{0}, buffer.Length());
```
markdroth, line 73: And here.

veblush:
@markdroth I addressed your last comments (thanks!) and I had to rebase this PR onto master to address one clang-tidy error introduced by a recent commit.

markdroth:
Please do not rebase once a review has started!  That makes it basically impossible for the reviewer to see what has changed since their last review.

Instead, please just merge master into your branch and push the merge commit.

markdroth (approved):

PR #31610, opened 2022-11-10
Added googletest clang-tidy checks

veblush:
Added two following clang-tidy checks
-  `google-upgrade-googletest-case`
-  `google-readability-avoid-underscore-in-googletest-name`

### DIFF
--- a/test/cpp/end2end/hybrid_end2end_test.cc
+++ b/test/cpp/end2end/hybrid_end2end_test.cc
@@ -540,7 +540,7 @@ TEST_F(HybridEnd2endTest, AsyncRequestStreamResponseStream) {
 }
 
 // Add a second service with one sync method.
-TEST_F(HybridEnd2endTest, asyncRequestStreamResponseStreamSyncDupService) {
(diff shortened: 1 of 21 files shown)


test/cpp/end2end/hybrid_end2end_test.cc
```diff
@@ -540,7 +540,7 @@ TEST_F(HybridEnd2endTest, AsyncRequestStreamResponseStream) {
 }
 
 // Add a second service with one sync method.
-TEST_F(HybridEnd2endTest, AsyncRequestStreamResponseStream_SyncDupService) {
+TEST_F(HybridEnd2endTest, asyncRequestStreamResponseStreamSyncDupService) {
```
drfloob, line 543: I'm curious about the different uses of CamelCase and lowerCamelCase in test names, I see both in this PR. Is TEST_F expected to use lowerCamelCase, while TEST uses CamelCase? If so, I don't see this in the clang tidy documentation.

veblush: Oh this is my mistake. It should start with a capital letter. Thanks for catching this. I just changed all other cases.

PR #31611, opened 2022-11-10
Add query delay flag to Prometheus query script.

wanlin31:
This PR adds the flag allowing user to configure the delay before query
the Prometheus.

### DIFF
--- a/tools/run_tests/performance/prometheus.py
+++ b/tools/run_tests/performance/prometheus.py
@@ -274,7 +274,8 @@ def main() -> None:
         '--delay_seconds',
         default=0,
         type=int,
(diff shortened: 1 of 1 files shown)


tools/run_tests/performance/prometheus.py
```diff
@@ -269,6 +270,12 @@ def main() -> None:
         default=False,
         help='Suppress informative output',
     )
+    argp.add_argument(
+        '--delay_seconds',
+        default=0,
+        type=int,
+        help='Configure delay in seconds to perform Prometheus queries, default is 0s',
```
paulosjca, line 277: Nit: I would say "default is 0" or "default is no delay" to be clear.

paulosjca (approved):

PR #31612, opened 2022-11-10
LB policy API: make pickers ref-counted

markdroth:
This eliminates some unnecessary picker wrapping in various parent LB policies.  It also paves the way for an upcoming WRR policy implementation in which the picker will need an internal ref for a timer.


yousukseung (approved):
Thanks for the refactor.

For my education, why don't we replace this with shared_ptr?

markdroth:
> For my education, why don't we replace this with shared_ptr?

Internally, see go/grpc-c-core-ref-counted-types-update, slide 8.

PR #31614, opened 2022-11-10
Add global structure of allocators + experiment

ananda1066:
Add global array of allocators, divided into small and big, and implement experiment to return bytes of some big allocator whenever Take is called. The next step is to remove compact reclamation and instead go through all of the big allocators asynchronously under high memory pressure to return free bytes. 

The thresholds and how many bytes to return / when to return them will likely need some tuning, but this experiment doesn't seem to cause any significant contention differences.


src/core/lib/resource_quota/memory_quota.h
```diff
@@ -319,6 +332,21 @@ class BasicMemoryQuota final
   friend class ReclamationSweep;
   class WaitForSweepPromise;
 
+  class AllocatorBucket {
+   public:
+    struct Shard {
+      absl::flat_hash_map<uint64_t, GrpcMemoryAllocatorImpl*> allocator_map;
```
ctiller, line 338: add `ABSL_GUARDED_BY(shard_mu)`

src/core/lib/resource_quota/memory_quota.h
```diff
@@ -327,9 +355,14 @@ class BasicMemoryQuota final
   std::atomic<intptr_t> free_bytes_{kInitialSize};
   // The total number of bytes in this quota.
   std::atomic<size_t> quota_size_{kInitialSize};
+  // Assign a unique ID to each allocator.
+  std::atomic<uint64_t> curr_allocator_id_{0};
```
ctiller, line 359: why assign an id, versus `reinterpret_cast<uintptr_t>(allocator)`?

doing so would let you eliminate the id storage and passing around everywhere

src/core/lib/resource_quota/memory_quota.cc
```diff
@@ -450,6 +501,82 @@ void BasicMemoryQuota::Return(size_t amount) {
...
+  AllocatorBucket::Shard& old_shard = allocators_[0].SelectShard(allocator);
+
+  {
+    absl::MutexLock l(&old_shard.shard_mu);
+    if (old_shard.allocator_map.erase(id) == 0) return;
+  }
+
+  AllocatorBucket::Shard& new_shard = allocators_[1].SelectShard(allocator);
+
+  new_shard.shard_mu.Lock();
```
ctiller, line 575: Could do a `MutexLock` here

src/core/lib/resource_quota/memory_quota.cc
```diff
@@ -450,6 +501,82 @@ void BasicMemoryQuota::Return(size_t amount) {
...
+  AllocatorBucket::Shard& old_shard = allocators_[1].SelectShard(allocator);
+
+  {
+    absl::MutexLock l(&old_shard.shard_mu);
+    if (old_shard.allocator_map.erase(id) == 0) return;
+  }
+
+  AllocatorBucket::Shard& new_shard = allocators_[0].SelectShard(allocator);
+
+  new_shard.shard_mu.Lock();
```
ctiller, line 554: Could do a `MutexLock` here

src/core/lib/resource_quota/memory_quota.cc
```diff
@@ -450,6 +501,82 @@ void BasicMemoryQuota::Return(size_t amount) {
...
+  }
+
+  AllocatorBucket::Shard& big_shard = allocators_[1].SelectShard(allocator);
+
+  big_shard.shard_mu.Lock();
+  big_shard.allocator_map.erase(id);
+  big_shard.shard_mu.Unlock();
+}
+
+void BasicMemoryQuota::MoveAllocatorBigToSmall(
```
ctiller, line 538: Suggest renaming to `MaybeMoveAllocatorBigToSmall` -- since there's a conditionality here that would be useful to call out at call sites

src/core/lib/resource_quota/memory_quota.cc
```diff
@@ -450,6 +501,82 @@ void BasicMemoryQuota::Return(size_t amount) {
...
+  {
+    absl::MutexLock l(&small_shard.shard_mu);
+    if (small_shard.allocator_map.erase(id) == 1) {
+      return;
+    }
+  }
+
+  AllocatorBucket::Shard& big_shard = allocators_[1].SelectShard(allocator);
+
+  big_shard.shard_mu.Lock();
```
ctiller, line 533: Use `MutexLock` here

src/core/lib/resource_quota/memory_quota.h
```diff
@@ -327,9 +355,14 @@ class BasicMemoryQuota final
...
 
   // Reclaimer queues.
   ReclaimerQueue reclaimers_[kNumReclamationPasses];
+  // List of all allocators sorted into 2 buckets, small (<100 KB free bytes)
+  // and large (>500 KB free bytes).
+  std::array<AllocatorBucket, 2> allocators_;
```
ctiller, line 365: I don't think we're getting benefit from an array here, let's just go with `AllocatorBucket big_allocator_; AllocatorBucket small_allocator_;`?

src/core/lib/resource_quota/memory_quota.h
```diff
@@ -372,10 +405,47 @@ class GrpcMemoryAllocatorImpl final : public EventEngineMemoryAllocatorImpl {
       // Try to immediately return some free'ed memory back to the total quota.
       MaybeDonateBack();
     }
+    if (is_big_.load(std::memory_order_relaxed)) {
```
ctiller, line 408: I haven't proved it yet, but this pattern of having an atomic guarding the state of a second thing that has a lock worries me: it seems likely that we end up with cases where is_big is true and we're in the small list, or the opposite - especially in the situation where we've got multiple concurrent allocations & deallocations happening.

Instead, I'd propose eliminating `is_big_` entirely, and adding a method `ResolveBigSmallList(uint64_t old_free_bytes, uint64_t new_free_bytes)` (I'm sure you'll choose a better name!)

The logic for this new method:
```
void ResolveBigSmallList(uint64_t old_free_bytes, uint64_t new_free_bytes) {
  while (true) {
    if (new_free_bytes < kSmallAllocatorThreshold) {
      if (old_free_bytes < kSmallAllocatorThreshold) return; // no change in status
      MaybeMoveAllocatorBigToSmall();
    } else if (new_free_bytes > kBigAllocatorThreshold) {
      if (old_free_bytes > kBigAllocatorThreshold) return; // no change in status
      MaybeMoveAllocatorSmallToBig();
    } else {
      // no change necessary: in the in-between space we don't care which list we're on
      return;
    }
    // We made some modifications: loop and make sure we ultimately land in the right place
    old_free_bytes = new_free_bytes;
    new_free_bytes = free_bytes_.load(std::memory_order_relaxed);
  }
}
```
now whatever fluctuations happen we can assuredly get back to a consistent state, and we don't need to have that tracking variable.

ctiller (changes requested):
This is a really good start... I think we can eliminate some of the extra state introduced though. Left some comments, but ping if I wasn't clear on things.

src/core/lib/resource_quota/memory_quota.h
ananda1066: changed to hash_set of allocator ptrs.

src/core/lib/resource_quota/memory_quota.h
ananda1066: done

src/core/lib/resource_quota/memory_quota.h
ananda1066: Done

src/core/lib/resource_quota/memory_quota.h
```diff
@@ -394,8 +452,13 @@ class GrpcMemoryAllocatorImpl final : public EventEngineMemoryAllocatorImpl {
   // Name of this allocator
   absl::string_view name() const { return name_; }
 
+  size_t getFreeBytes() const {
```
ctiller, line 455: This should have an upper case first letter... `GetFreeBytes`

ctiller (approved):
One trivial change then this lgtm

PR #31619, opened 2022-11-10
HTTP2: Improve "GOAWAY received" status message

yashykt:
Right now, RPCs that fail due to GOAWAYs just get "GOAWAY received" in the error status. Users generally need to either look in the logs or print the debug error string, to view more details.

After this change, users would see "GOAWAY received; Error code: xx; Debug Text: xxxxxx"


src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -973,9 +973,15 @@ void grpc_chttp2_add_incoming_goaway(grpc_chttp2_transport* t,
   }
```
ctiller, line 973: Can we delete this empty if whilst here?

yashykt: done. Thanks!

veblush (approved):

PR #31626, opened 2022-11-11
Added url validation for aws metadata endpoints in aws external account

aeitzman:
This PR adds validation for all the AWS metadata endpoints used in aws_external_account_credentials to match the new changes in the [AIP](https://github.com/aip-dev/google.aip.dev/pull/959).


src/core/lib/security/credentials/external/aws_external_account_credentials.cc
```diff
@@ -73,6 +77,18 @@ std::string UrlEncode(const absl::string_view& s) {
   return result;
 }
 
+bool ValidateAwsUrl(const std::string& urlString) {
+  absl::StatusOr<grpc_core::URI> url = grpc_core::URI::Parse(urlString);
+  if (!url.ok()) return false;
+  absl::string_view host;
+  absl::string_view port;
+  grpc_core::SplitHostPort(url->authority(), &host, &port);
+  if (host == awsEc2MetadataIpv4Address || host == awsEc2MetadataIpv6Address) {
```
markdroth, line 86: This can just say:

```
return host == awsEc2MetadataIpv4Address || host == awsEc2MetadataIpv6Address;
```

src/core/lib/security/credentials/external/aws_external_account_credentials.cc
```diff
@@ -73,6 +77,18 @@ std::string UrlEncode(const absl::string_view& s) {
   return result;
 }
 
+bool ValidateAwsUrl(const std::string& urlString) {
+  absl::StatusOr<grpc_core::URI> url = grpc_core::URI::Parse(urlString);
```
markdroth, line 81: You'll need to remove all of the `grpc_core::` prefixes in this code, since this function is already inside of that namespace.

src/core/lib/security/credentials/external/aws_external_account_credentials.cc
```diff
@@ -298,12 +333,6 @@ void AwsExternalAccountCredentials::OnRetrieveRegionInternal(
 
 void AwsExternalAccountCredentials::RetrieveRoleName() {
   absl::StatusOr<URI> uri = URI::Parse(url_);
-  if (!uri.ok()) {
```
markdroth, line 301: I realize that you're now parsing the URI earlier, so this should always succeed, but we don't really have any structural guarantee that a future code change wouldn't modify the value of `url_` between the previous parsing and this one, so it seems unsafe to rely on that.

I think we should either (a) store the URI in parsed form so that we don't have to re-parse it here, or (b) check the result of parsing here.

src/core/lib/security/credentials/external/aws_external_account_credentials.cc
```diff
@@ -363,12 +392,6 @@ void AwsExternalAccountCredentials::RetrieveSigningKeys() {
   }
   std::string url_with_role_name = absl::StrCat(url_, "/", role_name_);
   absl::StatusOr<URI> uri = URI::Parse(url_with_role_name);
-  if (!uri.ok()) {
```
markdroth, line 366: Same here.

markdroth (changes requested):

markdroth (approved):

markdroth:
The test failure here looks like an unrelated problem, so I'll go ahead and merge this.

PR #28206, opened 2021-11-24
Revert "use WorkSerializer for subchannel connectivity state notifications"

ctiller:
Reverts grpc/grpc#28111

This is causing ASAN failures internally

PR #28111, opened 2021-11-16
use WorkSerializer for subchannel connectivity state notifications

markdroth:
This uses the new APIs added in #27902 to replace the hacky custom queue we have now.

PR #31634, opened 2022-11-12
Second attempt: subchannel: use WorkSerializer for connectivity state notifications

markdroth:
This is a second attempt at #28111, which was reverted in #28206.  However, the original PR is from almost a year ago (I simply haven't had time to retry it until now), and the code has changed a lot since then, so an automated roll-forward didn't work; instead, I manually redid the changes on the current code-base.

The internal problem that triggered the rollback does not reproduce with this PR, so I assume that some other change over the last year fixed it.


src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -777,12 +766,16 @@ void Subchannel::Orphan() {
...
+  {
+    MutexLock lock(&mu_);
+    GPR_ASSERT(!shutdown_);
+    shutdown_ = true;
+    connector_.reset();
+    connected_subchannel_.reset();
+    health_watcher_map_.ShutdownLocked();
+  }
+  // Drain any connectivity state notifications after releasing the mutex.
+  work_serializer_.DrainQueue();
```
yashykt, line 778: On line 784 and line 790 we lock the mutex but we don't drain the work serializer. What is the chance that `get_or_add` will in the future do a WorkSerializer schedule and cause things to get stuck, similar for line 790

src/core/ext/filters/client_channel/subchannel.cc
```diff
@@ -717,19 +690,27 @@ channelz::SubchannelNode* Subchannel::channelz_node() {
...
+          DEBUG_LOCATION);
+      watcher_list_.AddWatcherLocked(std::move(watcher));
+    } else {
+      health_watcher_map_.AddWatcherLocked(
+          WeakRef(DEBUG_LOCATION, "health_watcher"), *health_check_service_name,
+          std::move(watcher));
+    }
   }
+  // Drain any connectivity state notifications after releasing the mutex.
+  work_serializer_.DrainQueue();
```
yashykt, line 713: Also consider what happens if in `CancelConnectivityStateWatch` on line 716, we later add some WorkSerializer schedules in one of the `RemoveWatcherLocked` functions

yashykt (approved):
Added some comments on some things to consider but overall looks ok. After submitting, please make sure to keep an eye on things

yashykt:
Also, as a sanity check, please make sure that the test from b/207498927 continues to pass with this change

PR #31654, opened 2022-11-15
[BinderTransport] Fix BinderTransport server race conditions

LittleCVR:

### DIFF
--- a/src/core/ext/transport/binder/transport/binder_transport.cc
+++ b/src/core/ext/transport/binder/transport/binder_transport.cc
@@ -610,6 +610,8 @@ static void perform_transport_op_locked(void* transport_op,
   if (op->set_accept_stream) {
     gbt->accept_stream_user_data = op->set_accept_stream_user_data;
     gbt->accept_stream_fn = op->set_accept_stream_fn;
+    gpr_log(GPR_DEBUG, "accept_stream_fn_called_count_ = %d",
+            gbt->accept_stream_fn_called_count_);
     while (gbt->accept_stream_fn_called_count_ > 0) {
       --gbt->accept_stream_fn_called_count_;
       gbt->combiner->Run(
--- a/src/core/ext/transport/binder/transport/binder_transport.cc
+++ b/src/core/ext/transport/binder/transport/binder_transport.cc
@@ -370,6 +370,19 @@ class MetadataEncoder {
 }  // namespace
 }  // namespace grpc_binder
 
+static void accept_stream_locked(void* gt, grpc_error_handle /*error*/) {
+  grpc_binder_transport* gbt = static_cast<grpc_binder_transport*>(gt);
+  if (gbt->accept_stream_fn) {
+    gpr_log(GPR_INFO, "Accepting a stream");
+    // must pass in a non-null value.
+    (*gbt->accept_stream_fn)(gbt->accept_stream_user_data, &gbt->base, gbt);
+  } else {
+    ++gbt->accept_stream_fn_called_count_;
+    gpr_log(GPR_INFO, "accept_stream_fn not set, current count = %d",
+            gbt->accept_stream_fn_called_count_);
+  }
+}
+
 static void perform_stream_op_locked(void* stream_op,
                                      grpc_error_handle /*error*/) {
   grpc_transport_stream_op_batch* op =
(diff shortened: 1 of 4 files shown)


src/core/ext/transport/binder/transport/binder_transport.cc
```diff
@@ -370,6 +370,19 @@ class MetadataEncoder {
 }  // namespace
 }  // namespace grpc_binder
 
+static void accept_stream_locked(void* gt, grpc_error_handle /*error*/) {
+  grpc_binder_transport* gbt = static_cast<grpc_binder_transport*>(gt);
+  if (gbt->accept_stream_fn) {
+    gpr_log(GPR_INFO, "Accepting a stream");
```
sifmelcara, line 376: Nit: Since this is codepath is executed for every RPC call, consider `GPR_DEBUG` instead?

src/core/ext/transport/binder/transport/binder_transport.cc
```diff
@@ -595,8 +608,14 @@ static void perform_transport_op_locked(void* transport_op,
     gbt->state_tracker.RemoveWatcher(op->stop_connectivity_watch);
   }
   if (op->set_accept_stream) {
-    gbt->accept_stream_fn = op->set_accept_stream_fn;
     gbt->accept_stream_user_data = op->set_accept_stream_user_data;
+    gbt->accept_stream_fn = op->set_accept_stream_fn;
```
sifmelcara, line 612: I think we will also want to add a `GPR_INFO`/`GPR_DEBUG` log here (log would be helpful even when `gbt->accept_stream_fn_called_count_` is `0`)

src/core/ext/transport/binder/transport/binder_transport.h
```diff
@@ -74,6 +74,7 @@ struct grpc_binder_transport {
   void (*accept_stream_fn)(void* user_data, grpc_transport* transport,
                            const void* server_data) = nullptr;
   void* accept_stream_user_data = nullptr;
+  int accept_stream_fn_called_count_{0};
```
sifmelcara, line 77: Nit: Add a comment for this variable

src/core/ext/transport/binder/wire_format/wire_reader_impl.cc
```diff
@@ -212,6 +214,7 @@ absl::Status WireReaderImpl::ProcessTransaction(transaction_code_t code,
       int64_t num_bytes = -1;
       GRPC_RETURN_IF_ERROR(parcel->ReadInt64(&num_bytes));
       gpr_log(GPR_DEBUG, "received acknowledge bytes = %" PRId64, num_bytes);
+      wire_writer_set_notification_.WaitForNotification();
```
sifmelcara, line 217: 1. Add comment to explain why we need this and in what potential scenario it will deadlock
2. Use `WaitForNotificationWithTimeout` instead. Log an error and return proper `absl::Status` when timeout happens

src/core/ext/transport/binder/wire_format/wire_reader_impl.cc
```diff
@@ -281,6 +284,7 @@ absl::Status WireReaderImpl::ProcessStreamingTransaction(
     }
   }
   if (need_to_send_ack) {
+    wire_writer_set_notification_.WaitForNotification();
```
sifmelcara, line 287: Same here

src/core/ext/transport/binder/wire_format/wire_reader_impl.h
```diff
@@ -130,6 +130,7 @@ class WireReaderImpl : public WireReader {
 
   // Used to send ACK.
   std::shared_ptr<WireWriter> wire_writer_;
+  grpc_core::Notification wire_writer_set_notification_;
```
sifmelcara, line 133: We can probably call this `wire_writer_ready_` or something like that, WDYT?

sifmelcara:
Also, note that we need to add a tag to the commit message title, I think both `[BinderTransport] ...` and `BinderTransport: ...` would work

src/core/ext/transport/binder/wire_format/wire_reader_impl.cc
sifmelcara: (Also should add a TODO to emphasize that this is a temporary workaround)

src/core/ext/transport/binder/transport/binder_transport.cc
LittleCVR: This doesn't really expose debugging details like pointer address so I would give it `GPR_INFO`. Besides, `GPR_INFO` won't be printed by default anyway.

sifmelcara (approved):
Looks good! Before merging, we probably should do a quick manual test to make sure this patch won't accidentally break other things that are using binder transport

LittleCVR:
I have tested manually with several real use cases, all seem good.

PR #31659, opened 2022-11-15
Dont install ruby macos artifact deps on v1.49.x

apolcyn:
These dependencies are causing installation failures and preventing a successful package build on the 1.49.x branch.

Since we don't actually use macos for the ruby artifact build anymore (we haven't since 1.45.x), let's stop installing them on this branch (master already doesn't use them).

cc @jtattermusch


apolcyn:
closing in favor of https://github.com/grpc/grpc/pull/31664

PR #31660, opened 2022-11-15
Clean-up pkgconfig

veblush:
Fixing https://github.com/grpc/grpc/issues/31336, superseding https://github.com/grpc/grpc/pull/31337

This change does

- Introducing `Requires.private` to pkg-config files of gRPC so that we can differentiate `public dependency` and `private dependency`. (e.g. abseil is public but upb is not)
- Changing gRPC depend on `re2`, `libcares`, and `zlib` packages instead of just linking against them so that pkg-config could find the right targets. (to address #31336)
  - To do this, re2 should be updated to get pkg-config support. (https://github.com/grpc/grpc/pull/31665)
- Removing `run_distrib_test_cmake_module_install_pkgconfig.sh` since it cannot pass with builtin pkg-config packages such as re2 now because a pre-installed re2 which will be used for application could be different from one used to build gRPC (module one), which is no-go.


veblush:
This isn't ready yet for review.

veblush:
Filed https://github.com/google/re2/issues/399 to address this missing re2 pkg-config installation in cmake install.

jtattermusch (approved):
LGTM, we should revisit our C++ package support story at some point,
The complexity of e.g our pkgconfig support is getting out of hand.

PR #31661, opened 2022-11-15
[EventEngine] Modify iomgr to allow creation of event engine clients and client side endpoints

Vignesh2208:
A shim layer for event engine endpoints.

The posix event engine clients and client side endpoints are also created when the event_engine_client experiment is enabled.


include/grpc/event_engine/slice_buffer.h
```diff
@@ -51,7 +51,13 @@ namespace experimental {
 /// an experimental API.
 class SliceBuffer {
  public:
-  explicit SliceBuffer() { grpc_slice_buffer_init(&slice_buffer_); }
+  SliceBuffer() { grpc_slice_buffer_init(&slice_buffer_); }
+  // Transfers slices into this new SliceBuffer, leaving the parameter empty.
+  // Does not take ownership of the slice_buffer argument.
+  explicit SliceBuffer(grpc_slice_buffer* slice_buffer) {
```
ctiller, line 57: let's add a named constructor for this...

`static SliceBuffer TakeCSliceBuffer(grpc_slice_buffer& slice_buffer)`

Vignesh2208: Done.

ctiller: can we make this private then? To force users to spell out what they want?

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -86,6 +100,41 @@ void do_tcp_client_global_init(void) {
   g_connection_shards = new std::vector<struct ConnectionShard>(num_shards);
 }
 
+int64_t event_engine_tcp_client_connect(
+    grpc_closure* on_connect, grpc_endpoint** endpoint,
+    const grpc_event_engine::experimental::EndpointConfig& config,
+    const grpc_resolved_address* addr, grpc_core::Timestamp deadline) {
+  auto resource_quota = reinterpret_cast<grpc_core::ResourceQuota*>(
+      config.GetVoidPointer(GRPC_ARG_RESOURCE_QUOTA));
```
ctiller, line 108: `config.GetObject<grpc_core::ResourceQuota>()`

ctiller (changes requested):

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: There's no GetObject method in EndpointConfig object. Its only present in the channel_args type.

src/core/lib/iomgr/event_engine_shims/resolved_address.cc
```diff
@@ -0,0 +1,41 @@
...
+#include "src/core/lib/iomgr/event_engine_shims/resolved_address.h"
+
+#include <grpc/event_engine/event_engine.h>
+
+#include "src/core/lib/iomgr/resolved_address.h"
+
+namespace grpc_event_engine {
+namespace experimental {
+
+EventEngine::ResolvedAddress CreateResolvedAddress(
```
drfloob, line 25: I think we _may_ need this later, after the shims are gone. Can we move these conversion methods to `src/core/lib/event_engine/resolved_address_internal.h`, and keep the implementations with the rest of `src/core/lib/event_engine/resolved_address.cc`?

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -86,6 +100,41 @@ void do_tcp_client_global_init(void) {
   g_connection_shards = new std::vector<struct ConnectionShard>(num_shards);
 }
 
+int64_t event_engine_tcp_client_connect(
```
drfloob, line 103: This appears to be implementation-agnostic. Can it be moved out of the posix-specific code paths? Otherwise we'll _duplicate_ it for every engine implementation we want to try with shims.

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -86,6 +100,41 @@ void do_tcp_client_global_init(void) {
...
+       endpoint](absl::StatusOr<std::unique_ptr<EventEngine::Endpoint>> ep) {
+        grpc_core::ExecCtx exec_ctx;
+        absl::Status conn_status = ep.ok() ? absl::OkStatus() : ep.status();
+        if (ep.ok()) {
+          *endpoint = grpc_event_engine_endpoint_create(std::move(*ep));
+        } else {
+          *endpoint = nullptr;
+        }
+        grpc_core::ExecCtx::Run(DEBUG_LOCATION, on_connect,
+                                absl_status_to_grpc_error(conn_status));
```
drfloob, line 121: Is this conversion necessary? `typedef absl::Status grpc_error_handle;` is the latest state of things.

test/core/iomgr/endpoint_pair_test.cc
```diff
@@ -18,18 +18,108 @@
...
+#ifdef GRPC_POSIX_SOCKET_TCP
+
+using namespace std::chrono_literals;
+
+namespace {
+
+using ::grpc_event_engine::experimental::ChannelArgsEndpointConfig;
+using ::grpc_event_engine::experimental::EventEngine;
+
+EventEngine::ResolvedAddress URIToResolvedAddress(std::string address_str) {
```
drfloob, line 54: This should probably live next to `ResolvedAddressToURI` in `src/core/lib/event_engine/tcp_socket_utils.h`, but return `absl::StatusOr<ResolvedAddress>`

test/core/iomgr/endpoint_tests.cc
```diff
@@ -134,8 +135,12 @@ static void read_and_write_test_read_handler(void* data,
   struct read_and_write_test_state* state =
       static_cast<struct read_and_write_test_state*>(data);
 
-  state->bytes_read += count_slices(
-      state->incoming.slices, state->incoming.count, &state->current_read_data);
+  if (error.ok()) {
```
drfloob, line 138: Is this extra error check necessary? Seems weird to check `error.ok()` 3 times in a 15 line function.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
```diff
@@ -0,0 +1,341 @@
...
+#include "src/core/lib/event_engine/posix.h"
+#include "src/core/lib/event_engine/tcp_socket_utils.h"
+#include "src/core/lib/gprpp/sync.h"
+#include "src/core/lib/iomgr/closure.h"
+#include "src/core/lib/iomgr/endpoint.h"
+#include "src/core/lib/iomgr/error.h"
+#include "src/core/lib/iomgr/port.h"
+#include "src/core/lib/transport/error_utils.h"
+
+extern grpc_core::TraceFlag grpc_tcp_trace;
```
drfloob, line 40: I'd suggest adding liberal `grpc_event_engine_trace`s as well, from `src/core/lib/event_engine/trace.h`

src/core/lib/iomgr/event_engine_shims/endpoint.cc
```diff
@@ -0,0 +1,341 @@
...
+      delete this;
+    }
+  }
+
+  // Returns a managed grpc_endpoint object. It retains ownership of the
+  // object.
+  grpc_endpoint* GetGrpcEndpoint() { return &eeep_->base; }
+
+  // Read using the underlying EventEngine endpoint object.
+  inline void Read(absl::AnyInvocable<void(absl::Status)> on_read,
```
drfloob, line 92: Is `inline` terribly meaningful here? This is a cc file, this code should only show up in a single translation unit right?

src/core/lib/iomgr/event_engine_shims/endpoint.cc
```diff
@@ -0,0 +1,341 @@
...
+  eeep->wrapper->Ref();
+  EventEngine::Endpoint::ReadArgs read_args = {min_progress_size};
+  SliceBuffer* read_buffer = new (&eeep->read_buffer) SliceBuffer(slices);
+  eeep->wrapper->Read(
+      [eeep, cb, slices](absl::Status status) {
+        auto* read_buffer = reinterpret_cast<SliceBuffer*>(&eeep->read_buffer);
+        grpc_slice_buffer_move_into(read_buffer->c_slice_buffer(), slices);
+        read_buffer->~SliceBuffer();
+        {
+          grpc_core::ExecCtx exec_ctx;
```
drfloob, line 200: Do we need an ApplicationCallbackExecCtx?

include/grpc/event_engine/slice_buffer.h
Vignesh2208: Done.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -1203,6 +1203,8 @@ PosixEndpointImpl::PosixEndpointImpl(EventHandle* handle,
                                      const PosixTcpOptions& options)
     : sock_(PosixSocketWrapper(handle->WrappedFd())),
       on_done_(on_done),
+      peer_address_(),
```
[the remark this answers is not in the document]
Vignesh2208: It was causing one msan failure when the event_engine clients are enabled.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
Vignesh2208: When run on all the tests it didn't seem like this was needed. I think we can add it back if we encounter some bug.

test/core/iomgr/endpoint_tests.cc
Vignesh2208: The test flakes without this checks. Sometimes the connection can get closed while the read callback is waiting for bytes to be read. If the socket returns -1 with no bytes read, then the status will be set to not ok. We should not attempt to call count_slices and read the slice buffer in this case. Previously this didn't happen because the test assumed that an ExecCtx::Flush will execute all pending callbacks and no read callback will be in wait while the socket gets closed.

test/core/iomgr/endpoint_pair_test.cc
Vignesh2208: Moved it to src/core/lib/event_engine/tcp_socket_utils.h. Also had to create a new target called parse_address in BUILD. This contains the functions necessary to implement the URIToResolvedAddress Method.

src/core/lib/iomgr/event_engine_shims/resolved_address.cc
Vignesh2208: Moved it now.

src/core/lib/iomgr/event_engine_shims/endpoint.h
```diff
@@ -0,0 +1,35 @@
...
+namespace grpc_event_engine {
+namespace experimental {
+
+grpc_endpoint* grpc_event_engine_endpoint_create(
+    std::unique_ptr<EventEngine::Endpoint> ee_endpoint);
+
+}  // namespace experimental
+}  // namespace grpc_event_engine
+
+/// Creates an internal grpc_endpoint struct from an EventEngine Endpoint.
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

src/core/BUILD
```diff
@@ -2024,7 +2024,10 @@ grpc_cc_library(
         "//:windows": ["windows_event_engine"],
         "//:windows_msvc": ["windows_event_engine"],
         "//:windows_other": ["windows_event_engine"],
-        "//conditions:default": ["posix_event_engine"],
+        "//conditions:default": [
+            "posix_event_engine",
+            "posix_event_engine_endpoint",
```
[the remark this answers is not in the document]
Vignesh2208: Removed.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
Vignesh2208: Removed.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
Vignesh2208: Added an event_engine trace now to event_engine_shims/endpoint.cc and event_engine_shims/endpoint.h

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: Moved it to event_engine_shims/tcp_client{.h,.cc}

src/core/lib/iomgr/tcp_client_posix.cc
Vignesh2208: Yes the absl_status_to_grpc_error function also sets the RPC status through an int property. A lot of tests fail without the RPC status.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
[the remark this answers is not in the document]
drfloob: What was causing what kind of an MSAN failure? Again, IIUC these lines are redundant, the default constructors will be used for those members in class definition order.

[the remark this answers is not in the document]
Vignesh2208: Removed.

test/core/event_engine/posix/posix_event_engine_connect_test.cc
```diff
@@ -161,7 +162,7 @@ TEST(PosixEventEngineTest, IndefiniteConnectTimeoutOrRstTest) {
         EXPECT_EQ(status.status().code(), absl::StatusCode::kUnknown);
         signal.Notify();
       },
-      URIToResolvedAddress(target_addr), config,
+      *URIToResolvedAddress(target_addr), config,
```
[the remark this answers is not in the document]
Vignesh2208: Fixed it now. I also de-duplicated the usage of URIToResolvedAddress wherever possible.

test/core/iomgr/endpoint_pair_test.cc
```diff
@@ -18,18 +18,98 @@
...
 #include "src/core/lib/gpr/useful.h"
 #include "src/core/lib/gprpp/crash.h"
+#include "src/core/lib/gprpp/notification.h"
+#include "src/core/lib/iomgr/event_engine_shims/endpoint.h"
+#include "src/core/lib/resource_quota/memory_quota.h"
 #include "test/core/iomgr/endpoint_tests.h"
+#include "test/core/util/port.h"
 #include "test/core/util/test_config.h"
 
+#ifdef GRPC_POSIX_SOCKET_TCP
```
drfloob, line 45: This is generic code, and should work work the WindowsEventEngine as well.

test/core/iomgr/endpoint_pair_test.cc
```diff
@@ -44,7 +124,16 @@ static grpc_endpoint_test_fixture create_fixture_endpoint_pair(
   a[0].type = GRPC_ARG_INTEGER;
   a[0].value.integer = static_cast<int>(slice_size);
   grpc_channel_args args = {GPR_ARRAY_SIZE(a), a};
+#ifdef GRPC_POSIX_SOCKET_TCP
```
drfloob, line 127: Same comment regarding non-posix-specific functionality.

test/core/iomgr/endpoint_tests.cc
drfloob: Here's a version that checks the error once:

```
static void read_and_write_test_read_handler_read_done(
    read_and_write_test_state* state) {
  gpr_log(GPR_DEBUG, "Read handler done");
  gpr_mu_lock(g_mu);
  GRPC_LOG_IF_ERROR("pollset_kick", grpc_pollset_kick(g_pollset, nullptr));
  gpr_mu_unlock(g_mu);
}

static void read_and_write_test_read_handler(void* data,
                                             grpc_error_handle error) {
  struct read_and_write_test_state* state =
      static_cast<struct read_and_write_test_state*>(data);
  if (!error.ok()) {
    state->read_done = 1;
    read_and_write_test_read_handler_read_done();
    return;
  }
  state->bytes_read += count_slices(
      state->incoming.slices, state->incoming.count, &state->current_read_data);
  if (state->bytes_read == state->target_bytes) {
    state->read_done = 2;
    read_and_write_test_read_handler_read_done();
    return;
  }
  // We perform many reads one after another. If grpc_endpoint_read and the
  // read_handler are both run inline, we might end up growing the stack
  // beyond the limit. Schedule the read on ExecCtx to avoid this.
  grpc_core::ExecCtx::Run(DEBUG_LOCATION, &state->read_scheduler,
                          absl::OkStatus());
}
```

Without thread annotations, it isn't super clear to me what needs to be guarded by g_mu, so the state->read_done value may need to move into the helper method (but I don't think so).

drfloob (commented):
I think this is probably my last pass on this PR, LGTM after the remaining comments are addressed.

test/core/iomgr/endpoint_pair_test.cc
Vignesh2208: Removed now.

test/core/iomgr/endpoint_pair_test.cc
Vignesh2208: Removed now.

test/core/iomgr/endpoint_tests.cc
Vignesh2208: I altered it to this format. I think read_done needs to be set under a mutex so its done inside the handler now.

test/core/iomgr/endpoint_pair_test.cc
Vignesh2208: The experiment checking is now done inside src/core/lib/event_engine/shim{.h,.cc} files. The ifdef statements are only added there.

drfloob (approved):

src/core/lib/iomgr/event_engine_shims/endpoint.cc
ctiller: would recommend adding it back... we're probably not testing thoroughly enough, and that callback could trigger *anything*

src/core/lib/iomgr/event_engine_shims/tcp_client.cc
```diff
@@ -0,0 +1,80 @@
...
+    grpc_closure* on_connect, grpc_endpoint** endpoint,
+    const grpc_event_engine::experimental::EndpointConfig& config,
+    const grpc_resolved_address* addr, grpc_core::Timestamp deadline) {
+  auto resource_quota = reinterpret_cast<grpc_core::ResourceQuota*>(
+      config.GetVoidPointer(GRPC_ARG_RESOURCE_QUOTA));
+  auto addr_uri = grpc_sockaddr_to_uri(addr);
+  EventEngine::ConnectionHandle handle = GetDefaultEventEngine()->Connect(
+      [on_connect,
+       endpoint](absl::StatusOr<std::unique_ptr<EventEngine::Endpoint>> ep) {
+        grpc_core::ExecCtx exec_ctx;
```
ctiller, line 48: ApplicationCallbackContext here too

src/core/lib/iomgr/event_engine_shims/endpoint.cc
```diff
@@ -0,0 +1,381 @@
...
+        auto* write_buffer =
+            reinterpret_cast<SliceBuffer*>(&eeep->write_buffer);
+        write_buffer->~SliceBuffer();
+        if (GRPC_TRACE_FLAG_ENABLED(grpc_tcp_trace)) {
+          gpr_log(GPR_INFO, "TCP: %p WRITE (peer=%s) error=%s", eeep->wrapper,
+                  std::string(eeep->wrapper->PeerAddress()).c_str(),
+                  status.ToString().c_str());
+        }
+        {
+          grpc_core::ExecCtx exec_ctx;
```
ctiller, line 269: ApplicationExecCtx

src/core/lib/iomgr/event_engine_shims/endpoint.cc
```diff
@@ -0,0 +1,381 @@
...
+      for (i = 0; i < slices->count; i++) {
+        char* dump =
+            grpc_dump_slice(slices->slices[i], GPR_DUMP_HEX | GPR_DUMP_ASCII);
+        gpr_log(GPR_DEBUG, "WRITE DATA: %s", dump);
+        gpr_free(dump);
+      }
+    }
+  }
+  SliceBuffer* write_buffer = new (&eeep->write_buffer)
+      SliceBuffer(SliceBuffer::TakeCSliceBuffer(*slices));
```
ctiller, line 257: `// TODO(someone): SliceBufferCast<>`

ctiller (approved):
Some nits, but we should fix the ApplicationExecCtx's now rather than wait for an outage.

src/core/lib/iomgr/event_engine_shims/tcp_client.cc
Vignesh2208: Added it now.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
Vignesh2208: Added it now.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
Vignesh2208: Added a TODO for me

Vignesh2208:
> Some nits, but we should fix the ApplicationExecCtx's now rather than wait for an outage.

Thanks for the review Craig! Sounds good. I added it now.

HannahShiSFB:
Hi AJ, will you provide a CF-shim like posix or IOS needs to do it? Thank you!
@drfloob 

cc: @sampajano

drfloob:
@HannahShiSFB These same shims are intended to work for all EventEngine implementations. You should only need a small configuration change on iOS. When you are ready, either myself or @sampajano can help get you configured if you need.

HannahShiSFB:
In my last attempt, the shim only works with posix engine, it crashed because it expects an event engine with fd, which is not the case for cfstream.
Let me try that again and figure the configurations needed for ios.

drfloob:
> In my last attempt, the shim only works with posix engine, it crashed because it expects an event engine with fd, which is not the case for cfstream.

Ah, I see. You are correct, the shims made some assumptions that aren't true anymore with iOS. Here's a PR that should fix the shims for you: https://github.com/grpc/grpc/pull/32363

Issue #28642, opened 2022-01-20
Allow tuning thread parameters for callback API server and client


xim:
I'd like to be able to tune thread paramemters in GRPC. Specifically the thread count and stack size when using the callback API. 

### Is your feature request related to a problem? Please describe.
When running on an embedded system, or for whatever the use case, I feel users should be allowed to choose these parameters, if for no other reason than for verifying the performance tradeoffs.

### Describe the solution you'd like
It doesn't need to be fancy. E.g:

```diff
--- a/src/cpp/common/completion_queue_cc.cc
+++ b/src/cpp/common/completion_queue_cc.cc
@@ -52,8 +52,15 @@ struct CallbackAlternativeCQ {
     refs++;
     if (refs == 1) {
       cq = new CompletionQueue;
-      int num_nexting_threads =
-          grpc_core::Clamp(gpr_cpu_num_cores() / 2, 2u, 16u);
+
+      int num_nexting_threads = grpc_core::Clamp(gpr_cpu_num_cores() / 2, 2u, 16u);
+      if (auto n_threads = std::getenv("GRPC_CALLBACK_NEXTING_THREAD_COUNT"))
+          num_nexting_threads = std::stoul(n_threads);
+
+      grpc_core::Thread::Options options;
+      if (auto stack_size = std::getenv("GRPC_CALLBACK_NEXTING_STACK_KB"))
+          options.set_stack_size(std::stoul(stack_size) * 1024);
+
       nexting_threads = new std::vector<grpc_core::Thread>;
       for (int i = 0; i < num_nexting_threads; i++) {
         nexting_threads->emplace_back(
@@ -93,7 +100,7 @@ struct CallbackAlternativeCQ {
                 functor->functor_run(functor, ev.success);
               }
             },
-            cq);
+            cq, nullptr, options);
       }
       for (auto& th : *nexting_threads) {
         th.Start();
```

### Describe alternatives you've considered
None of the APIs in grpc make these parameters readily available, it might be possible to implement this in a way that lets these parameters be set for all implementation models.

### Additional context
grpc seems to a decent match for some high-power embedded products, but these services might want to control resource usage, especially memory footprint.

xim:
I guess I'd like to add to this request that the rest of the global threads should also be made configurable wrt stack size.

drfloob:
[...]

PR #31662, opened 2022-11-15
Add ability to control callback service threadpool with resource quota

nick-cerebras:
Right now there is a hard coded upper bound of 16 threads in the thread pool for Callback Services. #28642 requested a mechanism to control the number of threads. (It also requested controlling per thread stack size which wasn't addressed here).

AFAICT, the existing `SetResourceQuota` just controls the thread manager for synchronous servers. This exposes an additional `SetCallbackResourceQuota` to the server builder to provide plumbing to tune a CallbackService. For mixed servers with synchronous calls and callbacks two ResourceQuotas could be used to determine how to share the underlying hardware.

### DIFF
--- a/include/grpcpp/completion_queue.h
+++ b/include/grpcpp/completion_queue.h
@@ -413,7 +413,8 @@ class CompletionQueue : private grpc::internal::GrpcLibrary {
     return true;
   }
 
-  static CompletionQueue* CallbackAlternativeCQ();
+  static CompletionQueue* CallbackAlternativeCQ(
+      grpc_resource_quota* server_rq = nullptr);
   static void ReleaseCallbackAlternativeCQ(CompletionQueue* cq);
 
   grpc_completion_queue* cq_;  // owned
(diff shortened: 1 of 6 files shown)


drfloob:
Hi @nick-cerebras . Sorry for the delay, this has been a point of discussion for a little while among the team. Coming soon, we'll have a much more powerful means of controlling all thread resources in the gRPC-core library, called the [EventEngine](https://github.com/grpc/grpc/blob/90524b0e1173d2e7482db6b72f14cd13e907970a/include/grpc/event_engine/event_engine.h). We still have a public `SetResourceQuota` API to support, and it's not decided yet if we're going to want to expand that API surface for callbacks, given that it'll be a bit redundant to the knobs that the EventEngine provides, and that custom EventEngines (which any application can provide) may not heed the thread limits set in the ResourceQuotas.

While an API like this may be introduced, this likely is not the implementation we would go with given what's coming later this year. Thank you for your efforts on this regardless.

Issue #31644, opened 2022-11-14
python: interop server image not build correctly


ejona86:
As seen at https://source.cloud.google.com/results/invocations/d9401d96-b2cb-4587-a171-81965b544af1 (grpc_interop_tocloud):

```
2022-11-14 18:27:08,029 invalid command name 'py39/bin/python'

2022-11-14 18:27:08,029 FAILED: interop_server_pythonasyncio [ret=1, pid=75143, time=362.5sec]
Server "pythonasyncio" has exited prematurely.
```

gnossen:
@XuanWang-Amos This seems like it should be relatively simple and is probably a good way to get acquainted with the interop test framework. Would you mind taking a look?

XuanWang-Amos:
Able to reproduce using this command:
`tools/run_tests/run_interop_tests.py -l python -s pythonasyncio --use_docker --http2_interop --internal_ci -t -j 8`

Looks like some issue with our docker set up.

PR #31663, opened 2022-11-15
Fix Python AIO interop test

XuanWang-Amos:
Fix https://github.com/grpc/grpc/issues/31644.

We changed the wrong line in this commit( https://github.com/grpc/grpc/commit/179a7c7459d2fd9ae24cae6ea906bf2e8bac8300#) causing interop test fail with error `invalid command name 'py39/bin/python'`. This PR fixes that.


gnossen (approved):
Yikes. Good find!

PR #25528, opened 2021-02-22
Update config selector for LDS update with no route config

lidizheng:
There is a corner case: when LDS is updated without any route config, the filter config update is ignored.

I thought enabling RDS in LDS may affect the processing of LDS, but as discussed offline, our xDS client will religiously follow the in-effective LDS to handle route updates. And processing route update has nothing to do with processing the filter config update from LDS, except refreshing the config selector. This means LDS should have a new code path to handle filter config update in itself, as title.

Related: https://github.com/grpc/grpc/pull/24354

PR #31668, opened 2022-11-16
xds resolver: fix edge cases in interactions between LDS and RDS

markdroth:
This fixes a couple of bugs accidentally introduced in #31457.  It adds tests for those bugs and for the edge cases discussed in #25528.

Once this gets merged, it will need to be back-ported to the 1.51 branch before the release.  CC @gnossen


test/cpp/end2end/xds/xds_routing_end2end_test.cc
```diff
@@ -127,6 +130,215 @@ TEST_P(LdsDeletionTest, ListenerDeletionIgnored) {
...
 }
 
+using LdsRdsInteractionTest = XdsEnd2endTest;
+
+INSTANTIATE_TEST_SUITE_P(
+    XdsTest, LdsRdsInteractionTest,
+    ::testing::Values(XdsTestType().set_enable_rds_testing()),
+    &XdsTestType::Name);
+
+TEST_P(LdsRdsInteractionTest, SwitchFromRdsToInlineRouteConfig) {
```
yashykt, line 140: which of these tests were failing before this?

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -883,26 +883,44 @@ void XdsResolver::OnListenerUpdate(XdsListenerResource listener) {
...
+            route_config_watcher_ = nullptr;
+          }
+          // Start watch for the new RDS resource name.
+          route_config_name_ = std::move(*rds_name);
           auto watcher = MakeRefCounted<RouteConfigWatcher>(Ref());
           route_config_watcher_ = watcher.get();
           XdsRouteConfigResourceType::StartWatch(
               xds_client_.get(), route_config_name_, std::move(watcher));
+        } else {
+          // RDS resource name has not changed, so no watch needs to be
```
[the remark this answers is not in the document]
yashykt: good point, thanks!

test/cpp/end2end/xds/xds_routing_end2end_test.cc
```diff
@@ -96,11 +97,13 @@ TEST_P(LdsDeletionTest, ListenerDeletionIgnored) {
...
   // Unset LDS resource and wait for client to ACK the update.
   balancer_->ads_service()->UnsetResource(kLdsTypeUrl, kServerName);
-  const auto deadline = absl::Now() + absl::Seconds(30);
+  const auto deadline =
+      absl::Now() + (absl::Seconds(30) * grpc_test_slowdown_factor());
   while (true) {
     ASSERT_LT(absl::Now(), deadline) << "timed out waiting for LDS ACK";
     response_state = balancer_->ads_service()->lds_response_state();
     if (response_state.has_value()) break;
+    absl::SleepFor(absl::Seconds(1) * grpc_test_slowdown_factor());
```
yashykt, line 106: uh oh! Why do we need a sleep?

test/cpp/end2end/xds/xds_routing_end2end_test.cc
```diff
@@ -127,6 +130,240 @@ TEST_P(LdsDeletionTest, ListenerDeletionIgnored) {
...
+  SetListenerAndRouteConfiguration(balancer_.get(), std::move(listener),
+                                   default_route_config_);
+  // Wait for the LDS update to be ACKed.
+  const auto deadline =
+      absl::Now() + (absl::Seconds(30) * grpc_test_slowdown_factor());
+  while (true) {
+    ASSERT_LT(absl::Now(), deadline) << "timed out waiting for LDS ACK";
+    response_state = balancer_->ads_service()->lds_response_state();
+    if (response_state.has_value()) break;
+    absl::SleepFor(absl::Seconds(1) * grpc_test_slowdown_factor());
```
yashykt, line 306: here too, why do we need the sleep?

test/cpp/end2end/xds/xds_routing_end2end_test.cc
```diff
@@ -127,6 +130,240 @@ TEST_P(LdsDeletionTest, ListenerDeletionIgnored) {
...
+  // that as an ACK.  When we have time, fix the behavior of the fake
+  // xDS server, and then change this test to ensure that there is no RDS
+  // ACK within the 30-second timeout period.
+  const auto deadline =
+      absl::Now() + (absl::Seconds(30) * grpc_test_slowdown_factor());
+  while (true) {
+    ASSERT_LT(absl::Now(), deadline) << "timed out waiting for RDS ACK";
+    response_state = balancer_->ads_service()->rds_response_state();
+    if (response_state.has_value()) break;
+    absl::SleepFor(absl::Seconds(1) * grpc_test_slowdown_factor());
```
yashykt, line 195: here too, why do we need the sleep?

test/cpp/end2end/xds/xds_routing_end2end_test.cc
yashykt: If we need it anyway, can you add a comment as to why?

yashykt (approved):
The sleep bothers me but otherwise looks ok

test/cpp/end2end/xds/xds_routing_end2end_test.cc
markdroth: The sleep just avoids wasting CPU time.  Without it, this is just a busy loop, constantly checking the same thing over and over again.  With the sleep, we pause to actually give the system time to change state before we check again.

Note that the sleep does not affect the overall 30-second deadline for getting the ACK.

test/cpp/end2end/xds/xds_routing_end2end_test.cc
markdroth: Same answer as above.  It doesn't change the behavior, just avoids wasting CPU time.

test/cpp/end2end/xds/xds_routing_end2end_test.cc
markdroth: Same answer as above.  It doesn't change the behavior, just avoids wasting CPU time.

test/cpp/end2end/xds/xds_routing_end2end_test.cc
yashykt: something like using a notification might be better

PR #31677, opened 2022-11-16
Add type annotations to _channel.py and _server.py

XuanWang-Amos:
Add type annotation to the following files:
* `src/python/grpcio/grpc/_channel.py`.
* `src/python/grpcio/grpc/_server.py`.


src/python/grpcio/grpc/_channel.py
```diff
@@ -1270,11 +1402,14 @@ def channel_spin():
...
-def _channel_managed_call_management(state):
+def _channel_managed_call_management(state: _ChannelCallState):
 
     # pylint: disable=too-many-arguments
-    def create(flags, method, host, deadline, metadata, credentials,
-               operationses, event_handler, context):
+    def create(flags: int, method: str, host: str, deadline: float,
+               metadata: Optional[MetadataType],
+               credentials: Optional[cygrpc.CallCredentials],
+               operationses: Iterable[Iterable[cygrpc.Operation]],
```
gnossen, line 1411: Can this be `Sequence[Sequence[cygrpc.Operation]]`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1312,8 +1447,16 @@ def create(flags, method, host, deadline, metadata, credentials,
...
 
 class _ChannelConnectivityState(object):
-
-    def __init__(self, channel):
+    lock: threading.RLock
+    channel: grpc.Channel
+    polling: bool
+    connectivity: grpc.ChannelConnectivity
+    try_to_connect: bool
+    callbacks_and_connectivities: Sequence[Sequence[Union[
```
gnossen, line 1455: I think the outer container should be `List` here. The inner container is more difficult. It looks like it's currently a list of fixed size 2 where the zeroth element is of type `Callable[[grpc.ChannelConnectivity], None]` and first element is of type `Optional[grpc.ChannelConnectivity]`. Can we just change the sites that mutate this member to use tuples so that this type can become `List[Tuple[Callable[[grpc.ChannelConnectivity], None], Optional[grpc.ChannelConnectivity]]]`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1322,15 +1465,16 @@ def __init__(self, channel):
...
+    def reset_postfork_child(self) -> None:
         self.polling = False
         self.connectivity = None
         self.try_to_connect = False
         self.callbacks_and_connectivities = []
         self.delivering = False
 
 
-def _deliveries(state):
+def _deliveries(
```
gnossen, line 1476: Why `Optional`? Should this be `List[Callable[[grpc.ChannelConnectivity], None]]`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1340,7 +1484,9 @@ def _deliveries(state):
     return callbacks_needing_update
 
 
-def _deliver(state, initial_connectivity, initial_callbacks):
+def _deliver(state: _ChannelConnectivityState,
+             initial_connectivity: grpc.ChannelConnectivity,
+             initial_callbacks: Sequence[Callable]) -> None:
```
gnossen, line 1489: `Sequence[Callable[[grpc.ChannelConnectivity], None]]`

src/python/grpcio/grpc/_channel.py
```diff
@@ -1360,7 +1506,8 @@ def _deliver(state, initial_connectivity, initial_callbacks):
                 return
 
 
-def _spawn_delivery(state, callbacks):
+def _spawn_delivery(state: _ChannelConnectivityState,
+                    callbacks: Sequence[Callable]) -> None:
```
gnossen, line 1510: `Sequence[Callable[[grpc.ChannelConnectivity], None]]`

src/python/grpcio/grpc/_channel.py
```diff
@@ -1410,7 +1558,8 @@ def _poll_connectivity(state, channel, initial_try_to_connect):
                         _spawn_delivery(state, callbacks)
 
 
-def _subscribe(state, callback, try_to_connect):
+def _subscribe(state: _ChannelConnectivityState, callback: Callable,
```
gnossen, line 1561: `Callable[[grpc.ChannelConnectivity], None]`

src/python/grpcio/grpc/_channel.py
```diff
@@ -1430,7 +1579,7 @@ def _subscribe(state, callback, try_to_connect):
             state.callbacks_and_connectivities.append([callback, None])
 
 
-def _unsubscribe(state, callback):
+def _unsubscribe(state: _ChannelConnectivityState, callback: Callable) -> None:
```
gnossen, line 1582: `Callable[[grpc.ChannelConnectivity], None]`

src/python/grpcio/grpc/_channel.py
```diff
@@ -1439,15 +1588,19 @@ def _unsubscribe(state, callback):
                 break
 
 
-def _augment_options(base_options, compression):
+def _augment_options(
+        base_options: Optional[ChannelArgumentType],
```
gnossen, line 1592: I'm realizing now that this name is misleading. `ChannelArgument` is singular, but this is actually of type `Sequence[Tuple[str, Any]]`. Can we make `ChannelArgumentType` just be `Tuple[str, Any]`? This way, we will see `Sequence[ChannelArgumentType]` in `_channel.py`.

src/python/grpcio/grpc/_channel.py
```diff
@@ -1439,15 +1588,19 @@ def _unsubscribe(state, callback):
...
+        compression: Optional[grpc.Compression]) -> Sequence[Tuple[str, Any]]:
```
gnossen, line 1593: `Sequence[Tuple[str, Any]]` is (currently) the same as `ChannelArgumentType`.

src/python/grpcio/grpc/_channel.py
gnossen, line 1592: Why `Optional`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1439,15 +1588,19 @@ def _unsubscribe(state, callback):
...
     compression_option = _compression.create_channel_option(compression)
     return tuple(base_options) + compression_option + ((
         cygrpc.ChannelArgKey.primary_user_agent_string,
         _USER_AGENT,
     ),)
 
 
-def _separate_channel_options(options):
+def _separate_channel_options(
+    options: Optional[ChannelArgumentType]
```
gnossen, line 1602: Why `Optional`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1461,8 +1614,14 @@ def _separate_channel_options(options):
...
 class Channel(grpc.Channel):
     """A cygrpc.Channel-backed implementation of grpc.Channel."""
-
-    def __init__(self, target, options, credentials, compression):
+    _single_threaded_unary_stream: bool
+    _channel: cygrpc.Channel
+    _call_state: _ChannelCallState
+    _connectivity_state: _ChannelConnectivityState
+
+    def __init__(self, target: str, options: Optional[ChannelArgumentType],
```
gnossen, line 1622: Why `Optional`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1484,30 +1643,37 @@ def __init__(self, target, options, credentials, compression):
         if cygrpc.g_gevent_activated:
             cygrpc.gevent_increment_channel_count()
 
-    def _process_python_options(self, python_options):
+    def _process_python_options(
+            self, python_options: Optional[ChannelArgumentType]) -> None:
```
gnossen, line 1647: Why `Optional`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1484,30 +1643,37 @@ def __init__(self, target, options, credentials, compression):
...
         """Sets channel attributes according to python-only channel options."""
         for pair in python_options:
             if pair[0] == grpc.experimental.ChannelOptions.SingleThreadedUnaryStream:
                 self._single_threaded_unary_stream = True
 
-    def subscribe(self, callback, try_to_connect=None):
+    def subscribe(self,
+                  callback: Callable,
```
gnossen, line 1654: `Callable[[grpc.ChannelConnectivity], None]`

src/python/grpcio/grpc/_channel.py
```diff
@@ -1484,30 +1643,37 @@ def __init__(self, target, options, credentials, compression):
...
+                  try_to_connect: Optional[bool] = None) -> None:
         _subscribe(self._connectivity_state, callback, try_to_connect)
 
-    def unsubscribe(self, callback):
+    def unsubscribe(self, callback: Callable) -> None:
```
gnossen, line 1658: `Callable[[grpc.ChannelConnectivity], None]`

gnossen (commented):
Not a full review yet. I've gotten through all files except for `_server.py`. I'll do the rest tomorrow.

In general, this is looking very good. Don't pay attention to the volume of comments -- there's just a _lot_ of code that we're researching and annotating here.

src/python/grpcio/grpc/_channel.py
```diff
@@ -81,18 +90,33 @@
...
+    condition: threading.Condition
+    due: Set[cygrpc.OperationType]
+    initial_metadata: Optional[MetadataType]
+    response: Any
+    trailing_metadata: Optional[MetadataType]
+    code: Optional[grpc.StatusCode]
+    details: Optional[str]
+    debug_error_string: Optional[str]
+    cancelled: bool
+    callbacks: Sequence[Callable]
```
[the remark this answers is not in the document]
XuanWang-Amos: I can't find mach information about type of `Callback`, looks like this is a place where domain knowledge comes in handy :)

src/python/grpcio/grpc/_channel.py
```diff
@@ -188,9 +217,12 @@ def handle_event(event):
 
 
 #pylint: disable=too-many-statements
-def _consume_request_iterator(request_iterator, state, call, request_serializer,
-                              event_handler):
-    """Consume a request iterator supplied by the user."""
+def _consume_request_iterator(request_iterator: Iterator, state: _RPCState,
+                              call: Union[cygrpc.IntegratedCall,
```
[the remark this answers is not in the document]
XuanWang-Amos: Make sense, but I prefer leave it to another PR since it changes cython layer, I'll keep a note on this.

src/python/grpcio/grpc/_channel.py
```diff
@@ -188,9 +217,12 @@ def handle_event(event):
...
+                                          cygrpc.SegregatedCall],
+                              request_serializer: SerializingFunction,
+                              event_handler: Optional[Callable]) -> None:
```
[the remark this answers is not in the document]
XuanWang-Amos: Will change it, should we do the same for `event_handler` in `_channel_managed_call_management.create()`?

src/python/grpcio/grpc/_channel.py
```diff
@@ -289,71 +321,77 @@ class _InactiveRpcError(grpc.RpcError, grpc.Call, grpc.Future):
...
+    ) -> Optional[types.TracebackType]:
         """See grpc.Future.traceback."""
         try:
             raise self
         except grpc.RpcError:
             return sys.exc_info()[2]
 
-    def add_done_callback(self, fn, timeout=None):  # pylint: disable=unused-argument
+    def add_done_callback(self,
+                          fn: DoneCallbackType,
```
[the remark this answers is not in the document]
XuanWang-Amos: Changed, looks like it's true for all `add_done_callback()` function since they all inherited from `grpc.Future`, will change them all.

src/python/grpcio/grpc/_channel.py
```diff
@@ -371,28 +409,35 @@ class _Rendezvous(grpc.RpcError, grpc.RpcContext):
       _deadline: A float representing the deadline of the RPC in seconds. Or
         possibly None, to represent an RPC with no deadline at all.
     """
-
-    def __init__(self, state, call, response_deserializer, deadline):
+    _state: _RPCState
+    _call: Union[cygrpc.SegregatedCall, cygrpc.IntegratedCall]
```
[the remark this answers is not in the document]
XuanWang-Amos: Noted.

src/python/grpcio/grpc/_channel.py
```diff
@@ -407,7 +452,7 @@ def cancel(self):
             else:
                 return False
 
-    def add_callback(self, callback):
+    def add_callback(self, callback: Callable) -> bool:
```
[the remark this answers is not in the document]
XuanWang-Amos: Sorry missed this, changed.

src/python/grpcio/grpc/_channel.py
```diff
@@ -550,15 +598,15 @@ def traceback(self, timeout=None):
                 except grpc.RpcError:
                     return sys.exc_info()[2]
 
-    def add_done_callback(self, fn):
+    def add_done_callback(self, fn: DoneCallbackType) -> None:
```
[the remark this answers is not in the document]
XuanWang-Amos: Sure, changed `fn` annotation to `Callable[[grpc.Future], None]` for all `add_done_callback` functions.

src/python/grpcio/grpc/_channel.py
```diff
@@ -826,7 +877,10 @@ def _response_ready():
                     raise self
 
 
-def _start_unary_request(request, timeout, request_serializer):
+def _start_unary_request(
+    request: Any, timeout: Optional[float],
+    request_serializer: SerializingFunction
+) -> Tuple[Optional[float], Optional[bytes], Optional[_InactiveRpcError]]:
```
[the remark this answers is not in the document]
XuanWang-Amos: Yeah, `grpc.RpcError` should be enough, `_InactiveRpcError` doesn't adds much value here.

src/python/grpcio/grpc/_channel.py
```diff
@@ -838,18 +892,23 @@ def _start_unary_request(request, timeout, request_serializer):
         return deadline, serialized_request, None
 
 
-def _end_unary_response_blocking(state, call, with_call, deadline):
+def _end_unary_response_blocking(
+    state: _RPCState, call: grpc.Call, with_call: bool,
+    deadline: Optional[float]
+) -> Union[ResponseType, Tuple[ResponseType, grpc.Call]]:
```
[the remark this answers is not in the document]
XuanWang-Amos: Yeah it's not a good practice to return entirely difference response in one function, I'll keep an note on this one too.

src/python/grpcio/grpc/_channel.py
```diff
@@ -838,18 +892,23 @@ def _start_unary_request(request, timeout, request_serializer):
...
             return state.response, rendezvous
         else:
             return state.response
     else:
-        raise _InactiveRpcError(state)
+        raise _InactiveRpcError(state)  # pytype: disable=not-instantiable
 
 
-def _stream_unary_invocation_operationses(metadata, initial_metadata_flags):
+def _stream_unary_invocation_operationses(
```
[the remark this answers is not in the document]
XuanWang-Amos: Oh I just noticed that, glad to change all of them back to normal form.

src/python/grpcio/grpc/_channel.py
```diff
@@ -861,16 +920,17 @@ def _stream_unary_invocation_operationses(metadata, initial_metadata_flags):
     )
 
 
-def _stream_unary_invocation_operationses_and_tags(metadata,
-                                                   initial_metadata_flags):
+def _stream_unary_invocation_operationses_and_tags(
+    metadata: Optional[MetadataType], initial_metadata_flags: int
+) -> Sequence[Tuple[Sequence[cygrpc.Operation], None]]:
```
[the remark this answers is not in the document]
XuanWang-Amos: Sounds good to me, changed it to `Optional[UserTag]`, and also changed the the first part to `Sequence[Sequence[cygrpc.Operation]]` to be aligned with the return value of `_stream_unary_invocation_operations`.

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: `Iterable` might not be a `Sequence`, but looks like we're only use `for ... in ` for this Iterable so in this case should be fine to change it to Sequence of Sequence. Will also change the doc string too.

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: Yes, fixed it.

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: it should be `List`, changed to `List[Callable[[grpc.ChannelConnectivity], None]]`

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: Changed to List[Callable[[grpc.ChannelConnectivity], None]]

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: Sure that's sounds more clear, will change `ChannelArgumentType` to `Tuple[str, Any]`

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: You're right, fixed :)

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: On second thought, we don't need `Optional` here, will remove it. (this comes from the 2nd return value of `_separate_channel_options`, it can be an empty list but not `None`)

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: Took a looks gain, `options` is required here. (This is the same `options` passed from channel creation (`Channel.__init__`), the doc string didn't mention it can be `None` or `Optional`).

PS: Glad we're adding types so we don't need manually search for keywords like `optional` or `None`.

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: As mentioned above, will remove `Optional` here.

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: Same as `core_options`, it can be an empty list but not `None`, will remove `Optional` here.

src/python/grpcio/grpc/_channel.py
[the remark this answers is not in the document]
gnossen: Please file a Github issue, assign yourself, and add a TODO comment here.

src/python/grpcio/grpc/_channel.py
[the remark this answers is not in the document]
gnossen: Yes.

src/python/grpcio/grpc/_channel.py
gnossen: Please create a github issue and add a TODO.

src/python/grpcio/grpc/_channel.py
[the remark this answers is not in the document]
XuanWang-Amos: Sure, just created an issue: https://github.com/grpc/grpc/issues/31703

src/python/grpcio/grpc/_channel.py
XuanWang-Amos: Sure, issue: https://github.com/grpc/grpc/issues/31704

src/python/grpcio/grpc/_channel.py
```diff
@@ -20,13 +20,22 @@
...
-import grpc
-from grpc import _common
-from grpc import _compression
-from grpc import _grpcio_metadata
+import types
+from typing import (Any, Callable, Iterable, Iterator, Optional, Sequence, Set,
+                    Tuple, Union)
+
+import grpc  # pytype: disable=pyi-error
+from grpc import _common  # pytype: disable=pyi-error
```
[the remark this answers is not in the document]
XuanWang-Amos: I think this is related to pytype version, pytype will have this pyi-error even for files which we already added types like `_common.py` ([failed test](https://source.cloud.google.com/results/invocations/a122e7d1-2b1b-43e9-bf89-e7bc97cfa081/targets/github%2Fgrpc%2Frun_tests%2Fsanity_linux_dbg_native%2Ftools%2Fdistrib%2Fcheck_pytype.sh/tests)), I tried test it locally with 2022.6.14 version and it passed.

Since I already have an action item to update pytype version to match the version used in g3, I created an issue to track it: https://github.com/grpc/grpc/issues/31706.

gnossen:
High level thought: Could we use a tool like [`typeguard`](https://typeguard.readthedocs.io/en/latest/) to give us even more confidence in our types? The Cython type system is not strong enough to convince me that we're always getting out of it what we think we're getting out of it.

src/python/grpcio/grpc/_server.py
```diff
@@ -769,10 +869,28 @@ class _ServerStage(enum.Enum):
...
+    rpc_states: Set[_RPCState]
+    due: Set[str]
+    server_deallocated: bool
 
     # pylint: disable=too-many-arguments
-    def __init__(self, completion_queue, server, generic_handlers,
-                 interceptor_pipeline, thread_pool, maximum_concurrent_rpcs):
+    def __init__(self, completion_queue: cygrpc.CompletionQueue,
+                 server: cygrpc.Server,
+                 generic_handlers: Optional[Sequence[grpc.GenericRpcHandler]],
```
gnossen, line 890: I don't think this can be `None` at this point.

src/python/grpcio/grpc/_server.py
```diff
@@ -769,10 +869,28 @@ class _ServerStage(enum.Enum):
 
 
 class _ServerState(object):
+    lock: threading.RLock
+    completion_queue: cygrpc.CompletionQueue
+    server: cygrpc.Server
+    generic_handlers: Optional[List[grpc.GenericRpcHandler]]
```
gnossen, line 875: I don't think this needs to be `Optional`.

src/python/grpcio/grpc/_server.py
```diff
@@ -769,10 +869,28 @@ class _ServerStage(enum.Enum):
...
+    interceptor_pipeline: Optional[_interceptor._ServicePipeline]
+    thread_pool: futures.ThreadPoolExecutor
+    stage: _ServerStage
+    termination_event: threading.Event
+    shutdown_events: Sequence
```
gnossen, line 880: `List[threading.Event]`

src/python/grpcio/grpc/_server.py
```diff
@@ -793,30 +911,33 @@ def __init__(self, completion_queue, server, generic_handlers,
         self.server_deallocated = False
 
 
-def _add_generic_handlers(state, generic_handlers):
+def _add_generic_handlers(
+        state: _ServerState,
+        generic_handlers: Iterable[grpc.GenericRpcHandler]) -> None:
```
gnossen, line 916: `Sequence`? The callers all offer `Sequence` and that would give us more flexibility when changing this function in the future.

src/python/grpcio/grpc/_server.py
```diff
@@ -945,49 +1069,60 @@ def _validate_generic_rpc_handlers(generic_rpc_handlers):
                 'not have "service" method!'.format(generic_rpc_handler))
 
 
-def _augment_options(base_options, compression):
+def _augment_options(
+    base_options: Optional[Sequence[ChannelArgumentType]],
```
gnossen, line 1073: `base_options` cannot be `None` [at this layer](https://github.com/grpc/grpc/blob/c20455e8363836aae4750d4325704496a5a3c278/src/python/grpcio/grpc/__init__.py#L2070).

src/python/grpcio/grpc/_server.py
```diff
@@ -945,49 +1069,60 @@ def _validate_generic_rpc_handlers(generic_rpc_handlers):
...
 class _Server(grpc.Server):
+    _state: _ServerState
 
     # pylint: disable=too-many-arguments
-    def __init__(self, thread_pool, generic_handlers, interceptors, options,
-                 maximum_concurrent_rpcs, compression, xds):
+    def __init__(self, thread_pool: futures.ThreadPoolExecutor,
+                 generic_handlers: Optional[Sequence[grpc.GenericRpcHandler]],
+                 interceptors: Optional[Sequence[grpc.ServerInterceptor]],
+                 options: Optional[Sequence[ChannelArgumentType]],
```
gnossen, line 1087: No need for the `Optional`

src/python/grpcio/grpc/_server.py
```diff
@@ -945,49 +1069,60 @@ def _validate_generic_rpc_handlers(generic_rpc_handlers):
                 'not have "service" method!'.format(generic_rpc_handler))
 
 
-def _augment_options(base_options, compression):
+def _augment_options(
+    base_options: Optional[Sequence[ChannelArgumentType]],
+    compression: Optional[grpc.Compression]
+) -> Optional[Sequence[ChannelArgumentType]]:
```
gnossen, line 1075: No need for the `Optional`

gnossen (commented):
And that should cover the rest of the files in the PR.

src/python/grpcio/grpc/_channel.py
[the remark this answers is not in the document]
XuanWang-Amos: Created new issue: https://github.com/grpc/grpc/issues/31725

src/python/grpcio/grpc/_typing.py
```diff
@@ -15,15 +15,16 @@
...
 RequestType = TypeVar('RequestType')
 ResponseType = TypeVar('ResponseType')
 SerializingFunction = Callable[[Any], bytes]
 DeserializingFunction = Callable[[bytes], Any]
 MetadataType = Sequence[Tuple[str, Union[str, bytes]]]
-ChannelArgumentType = Sequence[Tuple[str, Any]]
-EOFType = type(EOF)
+ChannelArgumentType = Tuple[str, Any]
 DoneCallbackType = Callable[[Any], None]
+CallbackType = Callable[[], None]
```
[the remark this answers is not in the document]
XuanWang-Amos: Thanks for the suggestion, I still prefer use a name for it like we do for `DoneCallbackType`, I'll change it to `NullaryCallbackType`.

src/python/grpcio/grpc/_server.py
```diff
@@ -87,6 +97,20 @@ class _HandlerCallDetails(
 
 
 class _RPCState(object):
+    condition: threading.Condition
+    due = Set[str]
+    request: Any
+    clinet: str
```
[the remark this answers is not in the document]
XuanWang-Amos: Changed to `client`, and the type is `str`.

src/python/grpcio/grpc/_server.py
```diff
@@ -87,6 +97,20 @@ class _HandlerCallDetails(
...
+    initial_metadata_allowed: bool
+    compression_algorithm: Optional[grpc.Compression]
+    disable_next_compression: bool
+    trailing_metadata: Optional[MetadataType]
+    code: Optional[grpc.StatusCode]
+    details: Optional[bytes]
+    statused: bool
+    rpc_errors: Sequence[Exception]
```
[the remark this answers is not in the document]
XuanWang-Amos: Changed to `List`.

src/python/grpcio/grpc/_server.py
```diff
@@ -87,6 +97,20 @@ class _HandlerCallDetails(
...
+    callbacks: Optional[Sequence[CallbackType]]
```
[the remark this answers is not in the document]
XuanWang-Amos: Yeah `List` is better, not sure why I used `Sequence`...

src/python/grpcio/grpc/_server.py
```diff
@@ -213,7 +253,10 @@ def receive_message(receive_message_event):
     return receive_message
 
 
-def _send_initial_metadata(state):
+def _send_initial_metadata(
+    state: _RPCState
+) -> Callable[[cygrpc.BaseEvent], Tuple[Optional[_RPCState],
+                                        Sequence[Callable]]]:
```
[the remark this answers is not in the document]
XuanWang-Amos: Changed whole return value to `ServerCallbackTag`.

src/python/grpcio/grpc/_server.py
```diff
@@ -233,60 +279,64 @@ def send_message(unused_send_message_event):
...
+    def is_active(self) -> bool:
         with self._state.condition:
             return _is_rpc_state_active(self._state)
 
-    def time_remaining(self):
+    def time_remaining(self) -> float:
         return max(self._rpc_event.call_details.deadline - time.time(), 0)
 
-    def cancel(self):
+    def cancel(self) -> bool:
```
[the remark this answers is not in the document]
XuanWang-Amos: Oh that's right, fixed.

src/python/grpcio/grpc/_server.py
```diff
@@ -233,60 +279,64 @@ def send_message(unused_send_message_event):
...
-    def invocation_metadata(self):
+    def invocation_metadata(self) -> Optional[MetadataType]:
         return self._rpc_event.invocation_metadata
 
-    def peer(self):
+    def peer(self) -> str:
         return _common.decode(self._rpc_event.call.peer())
 
-    def peer_identities(self):
+    def peer_identities(self) -> Optional[Iterable[bytes]]:
```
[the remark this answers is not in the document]
XuanWang-Amos: Make sense, let's save user the additional effort here. I'll keep a note and update documentation later.

src/python/grpcio/grpc/_server.py
```diff
@@ -233,60 +279,64 @@ def send_message(unused_send_message_event):
...
         return cygrpc.peer_identities(self._rpc_event.call)
 
-    def peer_identity_key(self):
+    def peer_identity_key(self) -> Optional[str]:
         id_key = cygrpc.peer_identity_key(self._rpc_event.call)
         return id_key if id_key is None else _common.decode(id_key)
 
-    def auth_context(self):
+    def auth_context(self) -> Mapping[str, Iterable[bytes]]:
```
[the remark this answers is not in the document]
XuanWang-Amos: Changed :)

src/python/grpcio/grpc/_server.py
```diff
@@ -945,49 +1069,60 @@ def _validate_generic_rpc_handlers(generic_rpc_handlers):
...
+    def wait_for_termination(self, timeout: Optional[float] = None) -> bool:
         # NOTE(https://bugs.python.org/issue35935)
         # Remove this workaround once threading.Event.wait() is working with
         # CTRL+C across platforms.
         return _common.wait(self._state.termination_event.wait,
                             self._state.termination_event.is_set,
                             timeout=timeout)
 
-    def stop(self, grace):
+    def stop(self, grace: Optional[float]) -> threading.Event:
```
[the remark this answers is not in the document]
XuanWang-Amos: Oops, missed last function, will add annotations.

src/python/grpcio/grpc/_server.py
XuanWang-Amos: You're right, `Sequence` should be enough, will do the same for `interceptors` and `options`.

src/python/grpcio/grpc/_server.py
XuanWang-Amos: Is it possible to add other `Event` to this list? (Looks like we won't based on `_server.py`, but better to confirm).

src/python/grpcio/grpc/_server.py
XuanWang-Amos: [add_generic_rpc_handlers](https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/__init__.py#L1407) says it's expecting an iterable in public documentation, since not all `Iterable` are `Sequence`, change it to sequence might break some integration?

src/python/grpcio/grpc/_server.py
XuanWang-Amos: Fixed.

src/python/grpcio/grpc/_server.py
```diff
@@ -174,7 +207,10 @@ def _abort(state, call, code, details):
         state.due.add(token)
 
 
-def _receive_close_on_server(state):
+def _receive_close_on_server(
+    state: _RPCState
+) -> Callable[[cygrpc.BaseEvent], Tuple[Optional[_RPCState],
```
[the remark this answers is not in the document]
gnossen: I think it's important to have "tag" somewhere in the name. These aren't the only type of callback used within the server, but they _are_ the only type of _tag_ used by the server.

src/python/grpcio/grpc/_server.py
gnossen: Yes. Only `_server.py` is allowed to access this API, so checking that `_server.py` is only using `threading.Event` is sufficient.

src/python/grpcio/grpc/_server.py
gnossen: That's true. We _could_ convert to a sequence by doing `list(user_input)` somewhere near the top, but this is fine for now.

src/python/grpcio/grpc/_channel.py
gnossen: This API is entirely internal, so we should be able to control what the call sites provide.

src/python/grpcio/grpc/_server.py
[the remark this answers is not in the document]
gnossen: Looks like this is still `Sequence`.

gnossen:
A couple of meta comments about the review process:

- Please move a PR out of a "draft" state before setting an "assignee" or asking for review.
- Please make sure that the commit resolving a comment has been pushed before marking the comment as resolved. This means that the reviewer can use the resolution status of a comment as a signal for whether something is ready for another round of review.

src/python/grpcio/grpc/_server.py
XuanWang-Amos: Changed to `List[grpc.GenericRpcHandler]`.

src/python/grpcio/grpc/_server.py
```diff
@@ -552,8 +618,11 @@ def _status(rpc_event, state, serialized_response):
             state.due.add(_SEND_STATUS_FROM_SERVER_TOKEN)
 
 
-def _unary_response_in_pool(rpc_event, state, behavior, argument_thunk,
-                            request_deserializer, response_serializer):
+def _unary_response_in_pool(
+        rpc_event: cygrpc.BaseEvent, state: _RPCState, behavior: Callable,
+        argument_thunk: Callable,
```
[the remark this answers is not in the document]
XuanWang-Amos: Unfortunately pytype will complain `TypeVar 'RequestType' appears only once in the function signature`, I'll put `Callable[[], Any]` here which is not ideal but slightly better than just `Callable`.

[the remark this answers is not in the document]
gnossen: Oh, that's disappointing. Thanks for trying.

src/python/grpcio/grpc/_channel.py
```diff
@@ -883,18 +946,31 @@ def _determine_deadline(user_deadline):
 
 
 class _UnaryUnaryMultiCallable(grpc.UnaryUnaryMultiCallable):
+    _channel: cygrpc.Channel
+    _managed_call: cygrpc.IntegratedCall
```
gnossen, line 950: This type is not right. Same for the constructor.

gnossen (commented):
Getting very close. Can you please file a tracking issue against yourself for the runtime type verification work?

XuanWang-Amos:
> Getting very close. Can you please file a tracking issue against yourself for the runtime type verification work?

Sure, just created an issue: https://github.com/grpc/grpc/issues/31748

gnossen (approved):
Thanks for sticking through on this. This PR is a _huge_ step forward for the library.

PR #31684, opened 2022-11-17
xDS: fix crash on wrong listener type (both client and server side)

markdroth:
This will also need to be back-ported to 1.51.x.  @gnossen

### DIFF
--- a/test/cpp/end2end/xds/xds_routing_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_routing_end2end_test.cc
@@ -50,6 +50,35 @@ TEST_P(LdsTest, NacksInvalidListener) {
       ::testing::HasSubstr("Listener has neither address nor ApiListener"));
 }
 
+// Tests that we go into TRANSIENT_FAILURE if the Listener is not an API
+// listener.
+TEST_P(LdsTest, NotAnApiListener) {
+  Listener listener = default_server_listener_;
+  listener.set_name(kServerName);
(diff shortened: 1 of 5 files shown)


test/cpp/end2end/xds/xds_routing_end2end_test.cc
```diff
@@ -50,6 +50,35 @@ TEST_P(LdsTest, NacksInvalidListener) {
...
+  const auto deadline =
+      absl::Now() + (absl::Seconds(30) * grpc_test_slowdown_factor());
+  while (true) {
+    ASSERT_LT(absl::Now(), deadline) << "timed out waiting for LDS ACK";
+    auto response_state = balancer_->ads_service()->lds_response_state();
+    if (response_state.has_value()) {
+      EXPECT_EQ(response_state->state, AdsServiceImpl::ResponseState::ACKED);
+      break;
+    }
+    absl::SleepFor(absl::Seconds(1) * grpc_test_slowdown_factor());
```
yashykt, line 78: as in the other PR, I would want a notification mechanism instead, but I don't want to block the PR, so a TODO is fine if you agree

yashykt (approved):

test/cpp/end2end/xds/xds_routing_end2end_test.cc
markdroth: Added a TODO.

PR #31690, opened 2022-11-17
doc/PROTOCOL-HTTP2.md: Clarify Percent-Byte-Unencoded comment

evanj:
The comment about Percent-Byte-Unencoded used to read: "space and VCHAR, except %". Attempt to clarify that the unencoded characters are SPACE (0x20) through TILDE (0x7E) inclusive, excluding % (0x25). This more directly represents how it is implemented.

### DIFF
--- a/doc/PROTOCOL-HTTP2.md
+++ b/doc/PROTOCOL-HTTP2.md
@@ -111,7 +111,7 @@ For requests, **EOS** (end-of-stream) is indicated by the presence of the END_ST
 * **Status** → "grpc-status" 1\*DIGIT ; 0-9
 * **Status-Message** → "grpc-message" Percent-Encoded
 * **Percent-Encoded** → 1\*(Percent-Byte-Unencoded / Percent-Byte-Encoded)
-* **Percent-Byte-Unencoded** → 1\*( %x20-%x24 / %x26-%x7E ) ; space and VCHAR, except %
+* **Percent-Byte-Unencoded** → 1\*( %x20-%x24 / %x26-%x7E ) ; SPACE through ~ inclusive, except %
 * **Percent-Byte-Encoded** → "%" 2HEXDIGIT ; 0-9 A-F
 
 **Response-Headers** & **Trailers-Only** are each delivered in a single HTTP2 HEADERS frame block. Most responses are expected to have both headers and trailers but **Trailers-Only** is permitted for calls that produce an immediate error. Status must be sent in **Trailers** even if the status code is OK.


yijiem:
This seems reasonable to me, directed to Eric @ejona86 for review.

ejona86:
I've been on the fence for the change. But I think I've now convinced myself that the current version is better than this change.

I think "how it is implemented" is not very relevant. That's not the purpose of the doc. It specifies "what is correct," not "how to make it correct." And that is useful to both implementors and users debugging. As an implementor, I think `%x20-%x24 / %x26-%x7E`  is the important part. Yes, you can choose to implement that with two ranges, as a range+exclusion, or a look up table, but that's up to you. But sometimes you question whether you understand things, or you aren't an implementor, or it is 3 AM, there's the "space and VCHAR, except %". "Of course % would be excluded, because percent encoding." FWIW, personally I find `' ' <= char && char <= '~' && char != '%'` less readable than `0x20 <= char && char <= 0x7E && char != '%'`, but to each his own.

As an example of how I think it is useful, RFC 7230 (HTTP/1) has:

```
     qdtext         = HTAB / SP /%x21 / %x23-5B / %x5D-7E / obs-text
```

Which is really `\t` and x20-xFF, except for `"`, `\`, and `DEL`. But ABNF doesn't let you say that. If there was such a comment (there isn't), it'd have nothing to do with however you may want to implement it. It's about comprehension, and seeing `\` and `"` are excluded make a lot of sense for a quoted string; it isn't some random ranges but "the obvious characters without the obvious characters."

ejona86:
(I don't mind at all if you have another alternative suggestion for it. In thinking about it myself it just seemed there weren't that many other options available.)

ejona86:
VCHAR is in lots of RFCs, but it is referenced via:

> Production rules are using <a href="https://tools.ietf.org/html/rfc5234">ABNF syntax</a>.

Which includes:

>         VCHAR          =  %x21-7E
>                                ; visible (printing) characters

But the mention in this doc isn't clear that stuff like VCHAR is there. So maybe we should just expand that short text to mention it includes things like VCHAR and DIGIT.

I will say I noticed that HEXDIGIT as used in this doc is incorrect, and it should be HEXDIG, but that's not the sort of thing that causes problem given the comment of `0-9 A-F` (although maybe "upper case only" is unclear; honestly, I wonder if many implementations are supporting lower-case as well).

Reopening since it does sound like there's some changes to be done.

yashykt:
what's the pending AI here? and who's plate is it on?

PR #31717, opened 2022-11-20
EventEngine::RunAfter: PollingResolver

yijiem:


src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -100,29 +101,22 @@ void PollingResolver::ShutdownLocked() {
...
+            "[polling resolver %p] re-resolution timer fired: shutdown_=%d",
+            this, shutdown_);
   }
   have_next_resolution_timer_ = false;
-  if (error.ok() && !shutdown_) {
+  if (!shutdown_) {
     StartResolvingLocked();
   }
-  Unref(DEBUG_LOCATION, "retry-timer");
+  Unref(DEBUG_LOCATION, "retry_timer");
```
drfloob, line 119: This `Unref` can be removed if the timer callback owns a Ref local variable, the unreffing will be handled automatically on closure destruction. See the RunAfter usage in `src/core/ext/filters/client_channel/lb_policy/weighted_target/weighted_target.cc`.

drfloob:
Great start! Please let me know if you have any questions. And don't forget to add the label `release notes: no` to get the Mergeable test passing.

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
yijiem: Done... I'm such a newbie...

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -89,7 +90,7 @@ void PollingResolver::RequestReresolutionLocked() {
 
 void PollingResolver::ResetBackoffLocked() {
   if (have_next_resolution_timer_) {
-    grpc_timer_cancel(&next_resolution_timer_);
+    event_engine_->Cancel(next_resolution_timer_handle_);
```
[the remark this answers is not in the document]
yijiem: Great catch, added logging for cancellation.

test/core/client_channel/resolvers/dns_resolver_test.cc
```diff
@@ -41,45 +43,51 @@
 #include "src/core/lib/uri/uri_parser.h"
 #include "test/core/util/test_config.h"
 
-static std::shared_ptr<grpc_core::WorkSerializer>* g_work_serializer;
+namespace grpc {
```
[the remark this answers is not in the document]
yijiem: Reverted, I thought using
```
namespace grpc {
namespace testing {
namespace {
```
was our convention for tests, a quick codesearch tells otherwise:
https://source.corp.google.com/search?q=file:third_party%2Fgrpc%20file:.*_test%5C.cc%20-%22namespace%20testing%22&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental and it's perfectly fine to use static for file local functions and variables.

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -200,9 +200,18 @@ void PollingResolver::GetResultStatus(absl::Status status) {
...
-    GRPC_CLOSURE_INIT(&on_next_resolution_, OnNextResolution, this, nullptr);
-    grpc_timer_init(&next_resolution_timer_, next_try, &on_next_resolution_);
+    next_resolution_timer_handle_ =
+        event_engine_->RunAfter(timeout, [self = Ref()]() mutable {
+          ApplicationCallbackExecCtx callback_exec_ctx;
+          ExecCtx exec_ctx;
+          auto* self_ptr = static_cast<PollingResolver*>(self.get());
+          self_ptr->work_serializer_->Run(
+              [self = std::move(self)]() {
+                auto* self_ptr = static_cast<PollingResolver*>(self.get());
```
drfloob, line 210: This cast in this innermost lambda is unnecessary, RefCountedPtr overrides `operator->`.

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -200,9 +200,18 @@ void PollingResolver::GetResultStatus(absl::Status status) {
         gpr_log(GPR_INFO, "[polling resolver %p] retrying immediately", this);
       }
     }
-    Ref(DEBUG_LOCATION, "next_resolution_timer").release();
-    GRPC_CLOSURE_INIT(&on_next_resolution_, OnNextResolution, this, nullptr);
-    grpc_timer_init(&next_resolution_timer_, next_try, &on_next_resolution_);
+    next_resolution_timer_handle_ =
```
drfloob, line 203: Now that this code is sizable and duplicated, it might be worth wrapping is inside a method like 'ScheduleNextResolutionTimer(timeout)`, to be used here and on line 246.

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -58,7 +58,8 @@ PollingResolver::PollingResolver(ResolverArgs args,
       tracer_(tracer),
       interested_parties_(args.pollset_set),
       min_time_between_resolutions_(min_time_between_resolutions),
-      backoff_(backoff_options) {
+      backoff_(backoff_options),
+      event_engine_(channel_args.GetObjectRef<EventEngine>()) {
```
drfloob, line 62: I should have noticed this in my last pass through: we don't need to store another shared_ptr to the engine since the shared_ptr in channel_args keeps the engine alive, a raw pointer will suffice. This can be `channel_args_.GetObject<EventEngine>()`.

drfloob (commented):
A few more minor things to address, but it looks pretty good!

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
ctiller: (please comment that sharing!)

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -99,30 +104,41 @@ void PollingResolver::ShutdownLocked() {
...
-      [self, error]() { self->OnNextResolutionLocked(error); }, DEBUG_LOCATION);
+void PollingResolver::ScheduleNextResolutionTimer(const Duration& timeout) {
+  next_resolution_timer_handle_ =
+      event_engine_->RunAfter(timeout, [self = Ref()]() mutable {
+        ApplicationCallbackExecCtx callback_exec_ctx;
+        ExecCtx exec_ctx;
+        auto* self_ptr = static_cast<PollingResolver*>(self.get());
+        self_ptr->work_serializer_->Run(
+            [self = std::move(self)]() {
+              auto* self_ptr = static_cast<PollingResolver*>(self.get());
```
drfloob, line 125: This `self_ptr` is unnecessary here, this lambda can simply be:

```
[self = std::move(self)]() mutable { self->OnNextResolutionLocked(); }
```

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -88,8 +89,12 @@ void PollingResolver::RequestReresolutionLocked() {
 }
 
 void PollingResolver::ResetBackoffLocked() {
-  if (have_next_resolution_timer_) {
-    grpc_timer_cancel(&next_resolution_timer_);
+  if (next_resolution_timer_handle_.has_value()) {
```
markdroth, line 92: Suggest refactoring this block into a helper method called `MaybeCancelNextResolutionTimer()`, so that we don't need to repeat the code both here and on line 107.

src/core/ext/filters/client_channel/resolver/polling_resolver.h
```diff
@@ -116,6 +110,12 @@ class PollingResolver : public Resolver {
     kReresolutionRequestedWhileCallbackWasPending,
   };
   ResultStatusState result_status_state_ = ResultStatusState::kNone;
+  /// next resolution timer
+  absl::optional<grpc_event_engine::experimental::EventEngine::TaskHandle>
+      next_resolution_timer_handle_{absl::nullopt};
+  // a raw pointer will suffice since the shared_ptr in channel_args_ will keep
+  // the EventEngine alive
+  grpc_event_engine::experimental::EventEngine* event_engine_;
```
markdroth, line 118: This pointer probably isn't necessary.  We can probably just grab it out of `channel_args_` each time we need it.  The code here is not performance-sensitive, so it's probably better to avoid wasting memory than to optimize this access.

src/core/ext/filters/client_channel/resolver/polling_resolver.h
```diff
@@ -116,6 +110,12 @@ class PollingResolver : public Resolver {
     kReresolutionRequestedWhileCallbackWasPending,
   };
   ResultStatusState result_status_state_ = ResultStatusState::kNone;
+  /// next resolution timer
+  absl::optional<grpc_event_engine::experimental::EventEngine::TaskHandle>
+      next_resolution_timer_handle_{absl::nullopt};
```
markdroth, line 115: No need to explicitly initialize to `absl::nullopt`; that's the default for `absl::optional<>`.

markdroth (commented):
A few minor drive-by comments...

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
drfloob: Disregard my suggestion, `self` here is a `Resolver*`, not a `PollingResolver*`.

markdroth: This problem comes up a lot.  The fact that `Ref()` on a subclass of `Resolver` returns a pointer to the base class type instead of a pointer to the subclass is a problem, but I haven't had time to figure out a good solution.

Fortunately, there's another bug in `RefCountedPtr<>` that offsets the above bug to some degree, which is that it allows down-casting without an explicit cast, which it really shouldn't.  But the fact that it does allow this means that you can do something like this:

```
RefCountedPtr<PollingResolver> self = Ref();
channel_args_.GetObject<EventEngine>()->RunAfter(
    timeout, [self = std::move(self)]() mutable { ... });
```

Now, inside the lambda, `self` is the subclass type, not the base class type.

It would be good to find a way to fix both of those bugs.  The second bug should be fairly easy to fix, although I don't want to fix it until we also fix the first one. :)

src/core/lib/gprpp/ref_counted_ptr.h
```diff
@@ -49,6 +49,8 @@ class RefCountedPtr {
   template <typename Y>
   // NOLINTNEXTLINE(google-explicit-constructor)
   RefCountedPtr(RefCountedPtr<Y>&& other) noexcept {
+    static_assert(std::has_virtual_destructor<T>::value,
```
markdroth, line 52: This could potentially break things, because [`RefCounted<>` supports objects with non-virtual dtors](https://github.com/grpc/grpc/blob/024a0f738898c5440dc4fa0102f72e6712cbb748/src/core/lib/gprpp/ref_counted.h#L259).

What's the benefit that we would get from this?  It's not clear to me that it actually buys us anything, and given the problem I just mentioned, I think it would be better to just remove it.

drfloob (approved):

yijiem (commented):
My bad, just noticed that the response has not actually been sent... didn't know how to use this <start review>...<end review> flow... Sorry for having merged the commit without actually responding to all comments...

PR #31724, opened 2022-11-21
Add concurrency into python example in route_guide

ngokchaoho:
- replace await single tasks sequentially with await task_group

In the asyncio example for route_guide for python, await is used everywhere and hence effectively removing concurrency. For example, if we await `guide_get_one_feature` sequentially, we are still doing task one by one instead of doing concurrently despite using asyncio.

Therefore, I think it is better to show example adding multiple tasks at the same time without dependency. 

Maybe @lidizheng can have a look at this.

### DIFF
--- a/examples/python/route_guide/asyncio_route_guide_client.py
+++ b/examples/python/route_guide/asyncio_route_guide_client.py
@@ -57,7 +57,7 @@ async def guide_get_feature(stub: route_guide_pb2_grpc.RouteGuideStub) -> None:
             stub, route_guide_pb2.Point(latitude=0, longitude=0)
         )
     )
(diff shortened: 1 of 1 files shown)


gnossen:
I think this change is reasonable. With beginners' examples though, we always have to strike a balance between clarity and robustness. Can you please add a comment explaining what's going on here. Not all readers of the example are going to be familiar with `asyncio`

ngokchaoho:
Thank you. Let me write some comments.

ngokchaoho:
Hi @gnossen please have a look if this comment is enough.

gnossen (approved):

PR #31728, opened 2022-11-21
Remove `include/grpc/impl/codegen/connectivity_state.h`

ralphchung:
This PR is a part of [b/196639718](http://b/196639718). It tries to move the content of `include/grpc/impl/codegen/connectivity_state.h` to `include/grpcpp/impl/connectivity_state.h`. We do not remove anything to achieve backward compatibility. The file `include/grpc/impl/codegen/connectivity_state.h` should be removed in the future.

This does not require any internal change, which has been verified through a test import CL.


ananda1066 (approved):

markdroth:
Why is the new file in the `impl` subdirectory?

More generally, what is the `impl` subdirectory for at this point?  Do we have established criteria for determining what should go in the `impl` subdirectory and what should go in the top-level directory?

ctiller:
[haven't yet looked at the PR]

I think the separation is the same as it's always been: `impl/` is for header files whose filenames do not form a part of the api, but need to be in the public tree for some reason.

The common case we're seeing is some data structure wants to be referred to without any of the associated functions, and so we'll have an `impl/` header file with the data structure, and include it from some public header file that includes other pieces.

markdroth:
I think the connectivity state enum is actually part of the public API -- I believe it's even exposed as part of the C++ API.  So wouldn't that imply that it should not be in `impl`?

PR #31730, opened 2022-11-21
Introduce the xds_override_host policy

eugeneo:
1. Introduce and setup a policy that needs to be
   enabled with a GRPC_EXPERIMENTAL_XDS_ENABLE_HOST_OVERRIDE
   environment variable.
2. Make the policy delegate to a child policy.
3. Add basic testing coverage.


eugeneo:
I consider this ready for the review as the code is behind a variable.

markdroth (commented):

BUILD
```diff
@@ -740,6 +740,7 @@ grpc_cc_library(
         "//src/core:grpc_lb_policy_ring_hash",
         "//src/core:grpc_lb_policy_round_robin",
         "//src/core:grpc_lb_policy_weighted_target",
+        "//src/core:grpc_lb_policy_xds_override_host",
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/BUILD
```diff
@@ -4236,6 +4236,39 @@ grpc_cc_library(
     ],
 )
 
+grpc_cc_library(
+    name = "grpc_lb_policy_xds_override_host",
+    srcs = [
+        "ext/filters/client_channel/lb_policy/xds_override_host/xds_override_host.cc",
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -0,0 +1,129 @@
...
+                        policy_.get()),
+            absl::OkStatus());
+  for (absl::string_view address : addresses) {
+    auto subchannel = FindSubchannel({address});
+    ASSERT_TRUE(subchannel);
+    subchannel->ConnectionRequested();
+    subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+    subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+  }
+  WaitForConnected();
```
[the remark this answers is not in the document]
eugeneo: Still struggling with this... PTAL

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -0,0 +1,129 @@
...
+  ExpectQueueEmpty();
+  // 2. Now we switch to a round-robin
+  EXPECT_EQ(ApplyUpdate(BuildUpdate(addresses,
+                                    MakeXdsOverrideHostConfig("round_robin")),
+                        policy_.get()),
+            absl::OkStatus());
+  for (absl::string_view address : addresses) {
+    auto subchannel = FindSubchannel({address});
+    ASSERT_TRUE(subchannel);
+    subchannel->ConnectionRequested();
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -0,0 +1,129 @@
...
+    subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+    subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+  }
+  auto picker = WaitForConnected();
+  // Pick first policy will always pick first!
+  EXPECT_EQ(ExpectPickComplete(picker.get()), "ipv4:127.0.0.1:441");
+  EXPECT_EQ(ExpectPickComplete(picker.get()), "ipv4:127.0.0.1:441");
+  ExpectQueueEmpty();
+  // 2. Now we switch to a round-robin
```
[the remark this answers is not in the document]
eugeneo: Done, though there's significant overlap between two.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -0,0 +1,129 @@
...
+  // 1. We use pick_first as a child
+  EXPECT_EQ(ApplyUpdate(BuildUpdate(addresses, MakeXdsOverrideHostConfig()),
+                        policy_.get()),
+            absl::OkStatus());
+  ExpectConnectingUpdate();
+  for (absl::string_view address : addresses) {
+    auto subchannel = FindSubchannel(
+        {address}, ChannelArgs().Set(GRPC_ARG_INHIBIT_HEALTH_CHECKING, true));
+    ASSERT_TRUE(subchannel);
+    subchannel->ConnectionRequested();
```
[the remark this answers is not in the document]
eugeneo: I am trying to check that child policy is actually different after the update...

eugeneo (commented):
Thank you for your review. Please take another look.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -43,49 +42,77 @@ class XdsOverrideHostTest : public LoadBalancingPolicyTest {
...
+    ASSERT_TRUE(subchannel->ConnectionRequested());
     subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
     subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+    EXPECT_NE(ExpectState(GRPC_CHANNEL_READY), nullptr);
   }
-  WaitForConnected();
-  WaitForConnected();
-  picker = WaitForConnected();
+  auto picker = ExpectState(GRPC_CHANNEL_READY);
+  EXPECT_NE(picker, nullptr);
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/lb_policy/xds_override_host/xds_override_host.cc
```diff
@@ -0,0 +1,404 @@
...
+    void AddTraceEvent(TraceSeverity severity,
+                       absl::string_view message) override;
+
+   private:
+    RefCountedPtr<XdsOverrideHostLb> xds_override_host_policy_;
+  };
+
+  ~XdsOverrideHostLb() override;
+
+  static std::string MakeKeyForAddress(const ServerAddress& address);
```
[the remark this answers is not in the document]
eugeneo: Deleted in this commit, will discuss in the next one.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -37,6 +34,28 @@ class XdsOverrideHostTest : public LoadBalancingPolicyTest {
...
   }
 
+  void SetUp() override {
+    grpc_core::SetEnv("GRPC_EXPERIMENTAL_XDS_ENABLE_HOST_OVERRIDE", "TRUE");
+    grpc_init();
+    policy_ = MakeLbPolicy("xds_override_host_experimental");
+  }
+
+  void TearDown() override {
+    // Note: Can't safely trigger this from inside the FakeHelper dtor,
```
markdroth, line 44: Instead of duplicating this code from the parent class, just call the method from the parent class directly:

```
LoadBalancingPolicyTest::TearDown();
```

That way, if we add anything else to the parent class `TearDown()` in the future, we won't need to duplicate it here.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -37,6 +34,28 @@ class XdsOverrideHostTest : public LoadBalancingPolicyTest {
          Json::Object{{"childPolicy", Json::Array{{child_policy_config}}}}}}});
   }
 
+  void SetUp() override {
+    grpc_core::SetEnv("GRPC_EXPERIMENTAL_XDS_ENABLE_HOST_OVERRIDE", "TRUE");
+    grpc_init();
```
markdroth, line 39: I think you can move `grpc_init()` and `grpc_shutdown()` back down to `main()`.  Instead, just add `grpc_core::CoreConfiguration::Reset()` here (in addition to the one in `TearDown()`), just to make sure that the env var takes effect.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -82,12 +101,15 @@ TEST_F(XdsOverrideHostTest, SwapChildPolicy) {
   auto subchannel =
       FindSubchannel({kAddresses[0]},
                      ChannelArgs().Set(GRPC_ARG_INHIBIT_HEALTH_CHECKING, true));
-  ASSERT_TRUE(subchannel);
+  ASSERT_NE(subchannel, nullptr);
   ASSERT_TRUE(subchannel->ConnectionRequested());
+  subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+  subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+  ASSERT_NE(WaitForConnected(), nullptr);
```
markdroth, line 108: Please check the behavior of the picker instead of just asserting that it's non-null.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -97,14 +119,14 @@ TEST_F(XdsOverrideHostTest, SwapChildPolicy) {
             absl::OkStatus());
   for (absl::string_view address : kAddresses) {
     auto subchannel = FindSubchannel({address});
-    ASSERT_TRUE(subchannel);
+    ASSERT_NE(subchannel, nullptr);
     ASSERT_TRUE(subchannel->ConnectionRequested());
```
markdroth, line 123: This should not be passing.  The first subchannel should already have been reporting READY before round_robin starts up, so it should never request a connection on the first subchannel, only on the second one.

If this is actually passing as currently written, then I think there's a bug somewhere.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -97,14 +119,14 @@ TEST_F(XdsOverrideHostTest, SwapChildPolicy) {
...
     subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
     subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
     EXPECT_NE(ExpectState(GRPC_CHANNEL_READY), nullptr);
```
markdroth, line 126: Please check the behavior of the picker instead of just asserting that it's non-null.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -97,14 +119,14 @@ TEST_F(XdsOverrideHostTest, SwapChildPolicy) {
...
   }
   auto picker = ExpectState(GRPC_CHANNEL_READY);
-  EXPECT_NE(picker, nullptr);
+  ASSERT_NE(picker, nullptr);
   ExpectPickComplete(picker.get());
```
markdroth, line 130: This call doesn't look to be necessary.  We're checking the behavior of the picker in the loop below.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -122,7 +144,7 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
       absl::InvalidArgumentError("Missing policy config"));
 }
 
-TEST_F(XdsOverrideHostTest, ConfigRequiresChildPolicy) {
+TEST_F(XdsOverrideHostTest, ValidateChildPolicyConfig) {
```
markdroth, line 147: Suggest moving the config tests to their own file, since they don't require any of the other infrastructure being set up in this test.  For an example of an LB policy config test, see https://github.com/grpc/grpc/blob/master/test/core/client_channel/outlier_detection_lb_config_parser_test.cc.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -131,6 +153,59 @@ TEST_F(XdsOverrideHostTest, ConfigRequiresChildPolicy) {
...
+
+          }}});
+  EXPECT_EQ(result.status(),
+            absl::InvalidArgumentError(
+                "errors validating xds_override_host LB policy config: "
+                "[field:childPolicy error:No known policies in list: ]"));
+}
+
+TEST_F(XdsOverrideHostTest, XdsOverrideHostLbPolicyDisabled) {
+  policy_.reset();
```
markdroth, line 202: I suggest moving this to a separate test class, so that you don't need all of this boilerplate:

```
using XdsOverrideHostDisabledTest = LoadBalancingPolicyTest;

TEST_F(XdsOverrideHostDisabledTest, PolicyDoesNotExist) {
  EXPECT_EQ(MakeLbPolicy("xds_override_host_experimental"), nullptr);
}
```

(Note that this depends on my suggestion above to move `grpc_init()` and `grpc_shutdown()` back to `main()`.)

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -0,0 +1,220 @@
...
+    subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+    EXPECT_NE(ExpectState(GRPC_CHANNEL_READY), nullptr);
+  }
+  auto picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
+  ExpectPickComplete(picker.get());
+  std::unordered_set<std::string> picked;
+  for (size_t i = 0; i < kAddresses.size(); i++) {
+    auto pick = ExpectPickComplete(picker.get());
+    EXPECT_TRUE(pick.has_value());
```
markdroth, line 134: This should be an `ASSERT` instead of an `EXPECT`, since otherwise the next line will cause a crash.

markdroth (commented):
This is getting really close!

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -361,41 +411,42 @@ class XdsOverrideHostLbFactory : public LoadBalancingPolicyFactory {
...
-    }
-    if (!errors.ok()) {
-      return errors.status(
-          "errors validating xds_override_host LB policy config");
-    }
-    return MakeRefCounted<XdsOverrideHostLbConfig>(std::move(child_policy));
+    return LoadRefCountedFromJson<XdsOverrideHostLbConfig>(
+        json, JsonArgs(),
+        "errors validating xds_override_host LB policy config");
+    // ValidationErrors errors;
```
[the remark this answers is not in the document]
eugeneo: Done (Oops!)

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -342,6 +358,40 @@ void XdsOverrideHostLb::Helper::AddTraceEvent(TraceSeverity severity,
...
+    return;
+  }
+  auto& child_config = it->second;
+
+  // Note that if type is not an array than framework will provide a more
+  // precise error message
+  if (child_config.type() == Json::Type::ARRAY &&
+      child_config.array_value().size() > 1) {
+    errors->AddError("exactly one child config should be specified");
+    return;
```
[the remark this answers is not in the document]

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -342,6 +358,40 @@ void XdsOverrideHostLb::Helper::AddTraceEvent(TraceSeverity severity,
...
+  if (it == json.object_value().end()) {
+    errors->AddError("field not present");
+    return;
+  }
+  auto& child_config = it->second;
+
+  // Note that if type is not an array than framework will provide a more
+  // precise error message
+  if (child_config.type() == Json::Type::ARRAY &&
+      child_config.array_value().size() > 1) {
```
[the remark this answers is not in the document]
eugeneo: Yeah, I feel like this check should be for all policies (if it even makes sense).

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -342,6 +358,40 @@ void XdsOverrideHostLb::Helper::AddTraceEvent(TraceSeverity severity,
...
+void XdsOverrideHostLbConfig::JsonPostLoad(const Json& json, const JsonArgs&,
+                                           ValidationErrors* errors) {
+  ValidationErrors::ScopedField field(errors, ".childPolicy");
+  auto it = json.object_value().find("childPolicy");
+  if (it == json.object_value().end()) {
+    errors->AddError("field not present");
+    return;
+  }
+  auto& child_config = it->second;
+
```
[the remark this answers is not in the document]

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -342,6 +358,40 @@ void XdsOverrideHostLb::Helper::AddTraceEvent(TraceSeverity severity,
...
+      .Finish();
+}
+
+void XdsOverrideHostLbConfig::JsonPostLoad(const Json& json, const JsonArgs&,
+                                           ValidationErrors* errors) {
+  ValidationErrors::ScopedField field(errors, ".childPolicy");
+  auto it = json.object_value().find("childPolicy");
+  if (it == json.object_value().end()) {
+    errors->AddError("field not present");
+    return;
```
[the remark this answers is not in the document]

eugeneo (commented):

test/core/client_channel/xds_override_host_lb_config_parser_test.cc
```diff
@@ -0,0 +1,166 @@
...
+  EXPECT_EQ(service_config.status(),
+            absl::InvalidArgumentError(
+                "errors validating service config: [field:loadBalancingConfig "
+                "error:errors validating xds_override_host LB policy config: "
+                "[field:childPolicy error:No known policies in list: ]]"));
+}
+
+class XdsOverrideHostDisabledConfigParsingTest : public ::testing::Test {
+ public:
+  static void SetUpTestSuite() { grpc_init(); }
```
markdroth, line 128: If you take my suggestion above about moving `grpc_init()` and `grpc_shutdown()` to main, then I don't think we need these `SetUpTestSuite()` and `TearDownTestSuite()` methods.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: For this PR, let's just change the code not to check this, since it shouldn't be happening.  The first subchannel should not be asked to connect, and we should not be having it report CONNECTING again, because it was already in READY state.

I think that after we switch from pick_first to round_robin, we should initially get a new picker that repeatedly returns only the first subchannel, because that subchannel was already connected from when we were using pick_first.  Then we can check that round_robin requests a connection on the second subchannel, then have that subchannel report CONNECTING followed by READY, and then we can expect a picker that round-robins between the two.

In other words, I think what we want here is this:

```
  // 2. Now we switch to a round-robin
  EXPECT_EQ(ApplyUpdate(BuildUpdate(kAddresses,
                                    MakeXdsOverrideHostConfig("round_robin")),
                        policy_.get()),
            absl::OkStatus());
  // First subchannel should already be connected, so round_robin
  // should initially return a picker that repeatedly returns the same subchannel.
  ASSERT_NE(picker, nullptr);
  EXPECT_EQ(kAddresses[0], ExpectPickComplete(picker.get());
  EXPECT_EQ(kAddresses[0], ExpectPickComplete(picker.get());
  // The round_robin policy should have requested a connection on the second subchannel.
  auto* subchannel = FindSubchannel({kAddresses[1]});
  // The subchannel reports CONNECTING.  We'll get a new READY picker, but it will
  // continue to hand out only the first subchannel.
  ASSERT_NE(picker, nullptr);
  EXPECT_EQ(kAddresses[0], ExpectPickComplete(picker.get());
  EXPECT_EQ(kAddresses[0], ExpectPickComplete(picker.get());  
  // Now the subchannel reports READY.  We get a new picker that returns both subchannels.
  ASSERT_NE(picker, nullptr);
  std::set<std::string> picked;
  for (size_t i = 0; i < kAddresses.size(); i++) {
    auto pick = ExpectPickComplete(picker.get());
    ASSERT_TRUE(pick.has_value());
    picked.insert(*pick);
  }
  EXPECT_THAT(picked, ::testing::UnorderedElementsAreArray(kAddresses));
```

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -123,15 +112,16 @@ TEST_F(XdsOverrideHostTest, SwapChildPolicy) {
     ASSERT_TRUE(subchannel->ConnectionRequested());
     subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
     subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
-    EXPECT_NE(ExpectState(GRPC_CHANNEL_READY), nullptr);
+    picker = ExpectState(GRPC_CHANNEL_READY);
+    EXPECT_NE(picker, nullptr);
+    ExpectPickComplete(picker.get());
```
markdroth, line 117: We should check that the picker is handing out the expected address.

markdroth (commented):
This is very close!

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -0,0 +1,420 @@
...
+#include "src/core/lib/transport/connectivity_state.h"
+
+namespace grpc_core {
+
+TraceFlag grpc_lb_xds_override_host_trace(false, "xds_override_host_lb");
+
+namespace {
+
+// TODO (eostroukhov): Remove once this policy is no longer experimental
+bool XdsOverrideHostLbEnabled() {
```
markdroth, line 65: I just realized that we don't actually need to guard the creation of this policy with the env var.  I think we really need the env var guard only for the xDS config, because if that config is not specified, then this policy is a no-op anyway.

So I think you can remove this env var guard and the "disabled" tests, which makes things a little simpler.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: This would basically be testing `round_robin`. I believe it is inefficient to duplicate those tests.
Note that there's a check all addresses are picked a few lines below.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -123,15 +112,16 @@ TEST_F(XdsOverrideHostTest, SwapChildPolicy) {
     ASSERT_TRUE(subchannel->ConnectionRequested());
     subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
     subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
-    EXPECT_NE(ExpectState(GRPC_CHANNEL_READY), nullptr);
+    picker = ExpectState(GRPC_CHANNEL_READY);
+    EXPECT_NE(picker, nullptr);
```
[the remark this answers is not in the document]
eugeneo: To manage the scope, I will create a new PR with the test case for swapping the child policy. It looks like it is not quite behaving the way it is supposed to.

These comments will be addressed there.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -144,68 +134,16 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
-
-          }}});
-  EXPECT_EQ(result.status(),
-            absl::InvalidArgumentError(
-                "errors validating xds_override_host LB policy config: "
-                "[field:childPolicy error:No known policies in list: ]"));
-}
+class XdsOverrideHostDisabledTest : public LoadBalancingPolicyTest {
+ protected:
+  static void SetUpTestSuite() {
```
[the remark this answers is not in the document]
eugeneo: No more variables 😀

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -105,7 +91,10 @@ TEST_F(XdsOverrideHostTest, SwapChildPolicy) {
   ASSERT_TRUE(subchannel->ConnectionRequested());
   subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
   subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
-  ASSERT_NE(WaitForConnected(), nullptr);
+  auto picker = WaitForConnected();
+  ASSERT_NE(picker, nullptr);
+  ExpectPickComplete(picker.get());
+
```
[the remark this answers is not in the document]
eugeneo: I am suggesting to bifurcate this test case to a separate PR.

test/core/client_channel/xds_override_host_lb_config_parser_test.cc
eugeneo: Yep. Removed.

test/core/client_channel/xds_override_host_lb_config_parser_test.cc
```diff
@@ -0,0 +1,166 @@
...
+#include "test/core/util/test_config.h"
+
+namespace grpc_core {
+namespace testing {
+namespace {
+
+class XdsOverrideHostConfigParsingTest : public ::testing::Test {
+ public:
+  static void SetUpTestSuite() {
+    grpc_core::SetEnv("GRPC_EXPERIMENTAL_XDS_ENABLE_HOST_OVERRIDE", "TRUE");
```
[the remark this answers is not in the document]
eugeneo: Not needed.

test/core/client_channel/xds_override_host_lb_config_parser_test.cc
```diff
@@ -0,0 +1,166 @@
...
+    grpc_init();
```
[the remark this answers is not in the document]
eugeneo: Done

eugeneo (commented):
Thank you for your review. I addressed the comments. Please take another look.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -355,14 +345,17 @@ void XdsOverrideHostLb::Helper::AddTraceEvent(TraceSeverity severity,
                                                                      message);
 }
 
+static const auto kJsonLoader =
```
markdroth, line 348: Please put this inside of the `XdsOverrideHostLbConfig::JsonLoader()` method.  It does not need to be accessed outside of that method, so there's no reason to put it outside.

Let's stick with the same approach that we use everywhere else.  For example:

https://github.com/grpc/grpc/blob/5933b52e1117e0aca1cd3635ef0370577bbff865/src/core/ext/filters/client_channel/lb_policy/outlier_detection/outlier_detection.cc#L1055

test/core/client_channel/xds_override_host_lb_config_parser_test.cc
```diff
@@ -33,19 +33,7 @@ namespace grpc_core {
...
-    grpc_init();
-  }
-
-  static void TearDownTestSuite() {
-    grpc_shutdown_blocking();
-    grpc_core::UnsetEnv("GRPC_EXPERIMENTAL_XDS_ENABLE_HOST_OVERRIDE");
-    grpc_core::CoreConfiguration::Reset();
-  }
-};
+class XdsOverrideHostConfigParsingTest : public ::testing::Test {};
```
markdroth, line 36: Since we no longer need to do anything in this test suite, we can remove it completely and change all of the `TEST_F()` invocations below to be `TEST()` instead.

markdroth (commented):
This looks great!  Just a couple of minor concerns left to address.

eugeneo (commented):
Addressed the comments, PTAL.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
markdroth: I don't know of any reason why local static variables would be frowned upon.

markdroth (approved):
This looks great!

eugeneo:
Test failure does not seem to be caused by this change. Rerun was successful: https://fusion2.corp.google.com/invocations/0d30ae96-b38c-4466-89ee-76c28d8cf2c6/targets

PR #31735, opened 2022-11-22
[EventEngine] WindowsEndpoint

drfloob:
Initial implementation, tested basic communication.

This also modifies the EventEngine TCP socket utils which were forked from iomgr for the PosixEventEngine, and makes them once again usable on other platforms.

Future work:

* I expect that once the client is roughed out, the needs for two-stage shutdown and ref counting will be more obvious (and testable).


src/core/lib/event_engine/windows/resolved_address.cc
```diff
@@ -0,0 +1,97 @@
...
+  GRPC_RETURN_IF_ERROR(path.status());
+  absl::StatusOr<grpc_core::URI> uri =
+      grpc_core::URI::Create(*scheme, /*authority=*/"", std::move(path.value()),
+                             /*query_parameter_pairs=*/{}, /*fragment=*/"");
+  if (!uri.ok()) return uri.status();
+  return uri->ToString();
+}
+
+// TODO(hork): implement normalization
+absl::StatusOr<std::string> ResolvedAddressToString(
```
ctiller, line 59: I count two implementations of this in the code base now, making this a third.

Given this is relatively complex and bug-prone code to write & maintain, I'd like us to have a plan to get down to one. I'd strongly prefer that we never exceed two -- can this implementation be merged with another to achieve that?

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+absl::Status WSAErrorToStatusWithMessage(
+    int wsa_error, absl::string_view custom_message,
+    const grpc_core::DebugLocation& location) {
+  // See
+  // https://learn.microsoft.com/en-us/windows/win32/api/Winsock2/nf-winsock2-wsasend
+  // https://learn.microsoft.com/en-us/windows/win32/api/Winsock2/nf-winsock2-wsarecv
+  char* wsa_message = gpr_format_message(wsa_error);
+  std::string message;
+  if (!custom_message.empty()) {
+    std::string message = absl::StrCat(wsa_message, ": ", custom_message);
```
ctiller, line 61: remove the `std::string` here - we don't want to declare a new variable and then immediately throw it away

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+    std::string message = wsa_message;
+  }
+  gpr_free(wsa_message);
+  switch (wsa_error) {
+    case 0:
+      return absl::OkStatus();
+    case WSAECONNRESET:
+    case WSAECONNABORTED:
+    case WSA_OPERATION_ABORTED:
+      return StatusCreate(absl::StatusCode::kAborted, message, location, {});
```
ctiller, line 72: agree we should be careful with status codes here - this could constitute an api change

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+    : peer_address_(peer_address),
+      socket_(std::move(socket)),
+      allocator_(std::move(allocator)),
+      handle_read_event_(this),
+      handle_write_event_(this),
+      executor_(executor) {
+  sockaddr addr;
+  int addr_len = sizeof(addr);
+  if (getsockname(socket_->socket(), &addr, &addr_len) < 0) {
+    GPR_ASSERT(false &&
```
ctiller, line 110: I'd prefer gpr_log, abort() -- this isn't an assert

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+  socket_->MaybeShutdown(absl::OkStatus());
+}
+
+void WindowsEndpoint::Read(absl::AnyInvocable<void(absl::Status)> on_read,
+                           SliceBuffer* buffer, const ReadArgs* /* args */) {
+  // TODO(hork): last_read_buffer from iomgr: Is it only garbage, or optimized?
+  GRPC_EVENT_ENGINE_TRACE("WindowsEndpoint::%p reading", this);
+  // Prepare the WSABUF struct
+  WSABUF wsa_buffers[kMaxWSABUFCount];
+  // TODO(hork): use read hint instead of the default?
```
ctiller, line 128: ideally we should not return from a read until the read hint is achieved (it's labelled `min_progress_size` in iomgr and that seems a better name)

if we do so, the calling code will assuredly call read again and not process the incoming memory until that limit is achieved anyway

whether that's important now or not I have no opinion, but ultimately we'll want to post *at least* min_progress_size bytes for an async read to avoid spurious wakeups from the kernel (we could post more, that would be fine)

src/core/lib/event_engine/windows/windows_endpoint.h
```diff
@@ -0,0 +1,90 @@
...
+  const EventEngine::ResolvedAddress& GetPeerAddress() const override;
+  const EventEngine::ResolvedAddress& GetLocalAddress() const override;
+
+ private:
+  // Base class for the Read- and Write-specific event handler callbacks
+  class BaseEventClosure : public EventEngine::Closure {
+   public:
+    explicit BaseEventClosure(WindowsEndpoint* endpoint);
+    virtual void Run() = 0;
+    void SetCallback(absl::AnyInvocable<void(absl::Status)> cb) {
```
ctiller, line 47: We always call `SetCallback` then `SetSliceBuffer` - suggest combining these into a single `Init`/`Prime`/`Prepare` call

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+                            SliceBuffer* data, const WriteArgs* /* args */) {
+  if (grpc_event_engine_trace.enabled()) {
+    for (int i = 0; i < data->Count(); i++) {
+      auto str = data->RefSlice(i).as_string_view();
+      gpr_log(GPR_INFO, "WindowsEndpoint::%p WRITE (peer=%s): %.*s", this,
+              peer_address_string_.c_str(), str.length(), str.data());
+    }
+  }
+  GPR_ASSERT(data->Count() <= UINT_MAX);
+  WSABUF local_buffers[kMaxWSABUFCount];
```
ctiller, line 187: suggest `absl::InlinedVector<WSABUF, kMaxWSABUFCount>` here and eschewing the complex hand allocation/deallocation

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+  LPWSABUF allocated = nullptr;
+  MemoryAllocator::Reservation memory_reservation;
+  if (data->Count() > kMaxWSABUFCount) {
+    size_t bytes_to_allocate = sizeof(WSABUF) * data->Count();
+    memory_reservation = allocator_.MakeReservation(bytes_to_allocate);
+    allocated = (WSABUF*)gpr_malloc(bytes_to_allocate);
+    buffers = allocated;
+  }
+  for (int i = 0; i < data->Count(); i++) {
+    auto slice = data->RefSlice(i);
```
ctiller, line 198: meta: we need a way of getting this data out without referencing the contained slice - doing so is incredibly inefficient

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+  }
+  // First, let's try a synchronous, non-blocking write.
+  DWORD bytes_sent;
+  int status = WSASend(socket_->socket(), buffers, (DWORD)data->Count(),
+                       &bytes_sent, 0, nullptr, nullptr);
+  size_t async_buffers_offset;
+  if (status == 0) {
+    if (bytes_sent == data->Length()) {
+      // Write completed, exiting early
+      executor_->Run(
```
ctiller, line 211: Interesting question I don't know the answer to: we do a synchronous write for efficiency here, but then need an extra allocation to call the callback.... does this render the synchronous path moot?

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,333 @@
...
+  cb_ = &AbortOnEvent;
+  absl::Status status;
+  if (write_info->wsa_error() != 0) {
+    status = WSAErrorToStatusWithMessage(write_info->wsa_error(),
+                                         "Error in WSASend", DEBUG_LOCATION);
+  } else {
+    GPR_ASSERT(write_info->bytes_transferred() == buffer_->Length());
+  }
+  endpoint_->executor_->Run(
+      [cb = std::move(cb), status = std::move(status)]() mutable {
```
ctiller, line 325: The read callback is made inline without going through the executor again... is there a reason this one also cannot?

test/core/event_engine/windows/create_sockpair.h
```diff
@@ -22,6 +22,11 @@
 namespace grpc_event_engine {
 namespace experimental {
 
+sockaddr_in GetLoopbackAddress();
```
ctiller, line 25: Pedant: prefer `GetSomeIpv4LoopbackAddress` -- rfc5735 has loopback as being all of `127.0.0.0/8` for ipv4

ctiller (changes requested):

src/core/lib/event_engine/windows/windows_endpoint.cc
drfloob: Agreed. `min_progress_size` is currently unused in the Windows iomgr implementation, but we can improve upon that here. I've added the override.

drfloob (commented):
Thanks for your review. I responded with a couple of questions, fixed most suggestions, and have one or two left to do.

src/core/lib/event_engine/windows/windows_endpoint.cc
drfloob: Done

src/core/lib/event_engine/windows/resolved_address.cc
drfloob: This is done, but because of it, I think we'll need a cherrypick.

src/core/lib/event_engine/windows/windows_endpoint.cc
ctiller: https://github.com/grpc/grpc/pull/31757 starts this work

src/core/lib/event_engine/windows/windows_endpoint.cc
```diff
@@ -0,0 +1,273 @@
...
+              peer_address_string_.c_str(), str.length(), str.data());
+    }
+  }
+  GPR_ASSERT(data->Count() <= UINT_MAX);
+  absl::InlinedVector<WSABUF, kMaxWSABUFCount> buffers(data->Count());
+  for (int i = 0; i < data->Count(); i++) {
+    auto slice = data->RefSlice(i);
+    GPR_ASSERT(slice.size() <= ULONG_MAX);
+    buffers[i].len = slice.size();
+    buffers[i].buf = (char*)slice.begin();
```
ctiller, line 149: nit: `reinterpret_cast` -- avoid c-style casts

ctiller (approved):

src/core/lib/event_engine/windows/resolved_address.cc
drfloob: Correction: this does not require a cherrypick, I did a dry run and found no uses (which makes sense)

PR #31742, opened 2022-11-22
Improve Windows network error logging in IOCP test

drfloob:

### DIFF
--- a/test/core/event_engine/windows/iocp_test.cc
+++ b/test/core/event_engine/windows/iocp_test.cc
@@ -44,6 +44,12 @@ using ::grpc_event_engine::experimental::Poller;
 using ::grpc_event_engine::experimental::SelfDeletingClosure;
 using ::grpc_event_engine::experimental::ThreadPool;
 using ::grpc_event_engine::experimental::WinSocket;
+
+void LogErrorMessage(int messageid, absl::string_view context) {
+  char* utf8_message = gpr_format_message(messageid);
+  gpr_log(GPR_ERROR, "Error in %s: %s", context, utf8_message);
+  gpr_free(utf8_message);
+}
 }  // namespace
 
 class IOCPTest : public testing::Test {};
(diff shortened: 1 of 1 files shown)


test/core/event_engine/windows/iocp_test.cc
```diff
@@ -75,7 +81,10 @@ TEST_F(IOCPTest, ClientReceivesNotificationOfServerSend) {
     // Expecting error 997, WSA_IO_PENDING
     EXPECT_EQ(status, -1);
     int last_error = WSAGetLastError();
-    ASSERT_EQ(last_error, WSA_IO_PENDING);
+    EXPECT_EQ(last_error, WSA_IO_PENDING);
```
ralphchung, line 84: I guess `gpr_log()` is not necessary in tests. We may use the tool provided by gtest by using something like the following to make it cleaner. And same below.
```c++
ASSERT_EQ(last_error, WSA_IO_PENDING) <<
    absl::Strformat("Error in %s: %s\n", "WSARecv",
                    gpr_format_message(last_error) /* may need to handle with `std::unique_ptr` with custom deleter */);
```

drfloob: Changing the public interface in `<grpc/support/log_windows.h>` is far out of scope, so this may be a bit awkward. We could wrap the method with something like:

```
std::unique_ptr<char*> GetWrappedFormattedMessage(int last_error, absl::string_view context) { 
  /* mechanics to get formatted message and wrap with w/ custom deleter */
}

ASSERT_EQ(last_error, WSA_IO_PENDING) << GetWrappedFormattedMessage(last_error, "WSARecv").get();
```

But I'm not sure it's worth the added complexity. We could also maybe tolerate the memory leak in these tests and ignore the need to free the `char*`, but I'd rather not risk it.

In future work, I'll be introducing canonical absl::Status conversions for Windows errors, which will make this cleaner. I'd suggest leaving it as is for now.

Japan50079:
AJ Heller ***@***.***>于2022年11月23日 周三03:02写道：

test/core/event_engine/windows/iocp_test.cc
ralphchung: Using `std::string` seems to be a acceptable workaround in my opinion, but I would leave this call back to you.

ralphchung (changes requested):
Probably add some TODO items to remind us to clean this up in the future.

Japan50079:
Cheng-Yu Chung ***@***.***>于2022年11月23日 周三03:37写道：

drfloob:
> Probably add some TODO items to remind us to clean this up in the future.
Done

ralphchung (approved):

PR #31743, opened 2022-11-22
Remove `include/grpcpp/impl/codegen/interceptor.h` completely

ralphchung:
Follow up for #31245. This PR removes `include/grpcpp/impl/codegen/interceptor.h` completely.


Japan50079:
Cheng-Yu Chung ***@***.***>于2022年11月23日 周三03:35写道：

ralphchung:

PR #31753, opened 2022-11-23
Revert "[event_engine] Move combiner executor usage to event engine"

ralphchung:
Reverts grpc/grpc#31713 since it breaks importing.


ralphchung:
Closed this PR since the issue in the importing was fixed forward.

drfloob:
We'll push ahead with the revert since this change introduces an init ref leak in `end2end:channelz_service_test`, apparently only on Windows.

yijiem (approved):

PR #31769, opened 2022-11-29
[Python Type Hints] Add type annotation to Python public APIs and ancillary packages

XuanWang-Amos:
### Description

Add type annotation to Python public APIs and all ancillary packages.

### Changes included
* Add type annotation to `__init__.py` and all ancillary packages, including:
  * grpcio-admin
  * grpcio-channelz
  * grpcio-csds
  * grpcio-health-checking
  * grpcio-reflection
  * grpcio-status
* Enabled run time type check using `Typeguard`.
* Fixed some internal type issues found by `Typeguard`.

### Testing
- Passed existing `pytype` checks.
- Added `Typeguard` for run time type checks.
  - It's added as a subset of unit tests and can be trigger using command like this: `//src/python/grpcio_tests/tests/unit:_server_test.typeguard`

### TODO
- Mark typing as experimental.


src/python/grpcio/grpc/__init__.py
```diff
@@ -1506,13 +1559,15 @@ def wait_for_termination(self, timeout=None):
 #################################  Functions    ################################
 
 
-def unary_unary_rpc_method_handler(behavior,
-                                   request_deserializer=None,
-                                   response_serializer=None):
+def unary_unary_rpc_method_handler(
+    behavior: Callable[[Any, Any], Any],
```
XuanWang-Amos, line 1563: Not entirely sure here, but based on [example](https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests/qps/benchmark_server.py#L39), all `behavior` should take two arguments, one for request and one for context.

src/python/grpcio/grpc/__init__.py
```diff
@@ -1705,13 +1777,14 @@ def composite_channel_credentials(channel_credentials, *call_credentials):
...
-                           require_client_auth=False):
+def ssl_server_credentials(
+        private_key_certificate_chain_pairs: Sequence[Sequence],
+        root_certificates: Optional[bytes] = None,
+        require_client_auth: Optional[bool] = False) -> ServerCredentials:
     """Creates a ServerCredentials for use with an SSL-enabled Server.
 
     Args:
-      private_key_certificate_chain_pairs: A list of pairs of the form
+      private_key_certificate_chain_pairs: A sequence of pairs of the form
```
XuanWang-Amos, line 1787: Changed it to `Sequence` since some users are using [sequence](https://github.com/gaia-pipeline/pythonsdk/blob/c4d25740f038ed6e9cffcf68d5dde9ddc2a4f804/gaiasdk/sdk.py#L142).

src/python/grpcio/grpc/__init__.py
```diff
@@ -1398,19 +1449,20 @@ class Server(abc.ABC):
...
-    def add_generic_rpc_handlers(self, generic_rpc_handlers):
+    def add_generic_rpc_handlers(
+            self, generic_rpc_handlers: Sequence[GenericRpcHandler]) -> None:
         """Registers GenericRpcHandlers with this Server.
 
         This method is only safe to call before the server is started.
 
         Args:
-          generic_rpc_handlers: An iterable of GenericRpcHandlers that will be
+          generic_rpc_handlers: A sequence of GenericRpcHandlers that will be
```
XuanWang-Amos, line 1459: We're using [`Sequence` in aio](https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/src/python/grpcio/grpc/aio/_server.py#L67), since all iterables are sequence, changed here to `Sequence` too.

gnossen:
High level thoughts:
- What about [grpc-testing](https://github.com/grpc/grpc/tree/master/src/python/grpcio_testing)? It's fine to put this off for another PR, but I want to make sure it's on your radar.
- Since these are public APIs, we need to somehow indicate to people in a highly visible way that they are experimental for the time being. I would recommend adding some text saying this to the README file for each package (example: [1](https://github.com/grpc/grpc/blob/master/src/python/grpcio/README.rst), [2](https://github.com/grpc/grpc/blob/7e201fbe42718d48226d1d795abe6796943601db/src/python/grpcio_admin/README.rst)). This will show up on [the Pypi page](https://pypi.org/project/grpcio/). And we should also add this to the release notes.

src/python/grpcio/grpc/__init__.py
```diff
@@ -834,13 +866,14 @@ def __call__(self,
...
-    def with_call(self,
-                  request_iterator,
-                  timeout=None,
-                  metadata=None,
-                  credentials=None,
-                  wait_for_ready=None,
-                  compression=None):
+    def with_call(
+            self,
+            request_iterator: Iterable,
```
gnossen, line 871: Same comment as above.

src/python/grpcio/grpc/__init__.py
```diff
@@ -801,12 +833,12 @@ class StreamUnaryMultiCallable(abc.ABC):
 
     @abc.abstractmethod
     def __call__(self,
-                 request_iterator,
-                 timeout=None,
-                 metadata=None,
-                 credentials=None,
-                 wait_for_ready=None,
-                 compression=None):
+                 request_iterator: Iterable,
```
gnossen, line 836: Why just `Iterable` and not `RequestIterable`?

src/python/grpcio/grpc/__init__.py
```diff
@@ -869,12 +902,12 @@ def with_call(self,
 
     @abc.abstractmethod
     def future(self,
-               request_iterator,
-               timeout=None,
-               metadata=None,
-               credentials=None,
-               wait_for_ready=None,
-               compression=None):
+               request_iterator: Iterable,
```
gnossen, line 905: Same comment as above.

src/python/grpcio/grpc/__init__.py
```diff
@@ -905,12 +938,12 @@ class StreamStreamMultiCallable(abc.ABC):
 
     @abc.abstractmethod
     def __call__(self,
-                 request_iterator,
-                 timeout=None,
-                 metadata=None,
-                 credentials=None,
-                 wait_for_ready=None,
-                 compression=None):
+                 request_iterator: Iterable,
```
gnossen, line 941: Likewise, why not `RequestIterableType`?

src/python/grpcio/grpc/__init__.py
```diff
@@ -946,7 +979,9 @@ class Channel(abc.ABC):
     """
 
     @abc.abstractmethod
-    def subscribe(self, callback, try_to_connect=False):
+    def subscribe(self,
+                  callback: Callable[[ChannelConnectivity], None],
+                  try_to_connect: Optional[bool] = False) -> None:
```
gnossen, line 984: While the implemenation _does_ allow a `None` value for `try_to_connect`, I see no value in surfacing that here. Let's just make this `bool`.

gnossen (commented):
Not a full review yet, but I wanted to get the first batch of comments out.

src/python/grpcio/grpc/__init__.py
gnossen: You're right. The docstrings are imprecise. We should probably add "and a ServiceContext object" to all 4 of these docstrings.

This reminds me. Can we also look into making sure that [our generated documentation](https://grpc.github.io/grpc/python/grpc.html) displays these new types?

src/python/grpcio_channelz/grpc_channelz/v1/_servicer.py
```diff
@@ -24,7 +24,10 @@ class ChannelzServicer(_channelz_pb2_grpc.ChannelzServicer):
     """Servicer handling RPCs for service statuses."""
 
     @staticmethod
-    def GetTopChannels(request, context):
+    def GetTopChannels(
```
gnossen, line 27: This is a meta note for all ancillary packages. Please manually test to make sure that introducing a typing error causes a CI failure.

src/python/grpcio_csds/grpc_csds/__init__.py
```diff
@@ -24,21 +27,21 @@ class ClientStatusDiscoveryServiceServicer(
     """CSDS Servicer works for both the sync API and asyncio API."""
 
     @staticmethod
-    def FetchClientStatus(request, unused_context):
+    def FetchClientStatus(request: Any, unused_context: Any) -> Any:
```
gnossen, line 30: I see you used `Any` here because of the current build issue with `xds_protos`. For the moment, I think it would be more helpful to just leave the types off of the RPC handlers. This isn't actually public API anyway.

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -56,18 +61,19 @@ def next(self):
     def __next__(self):
         return self._next()
 
-    def add(self, response):
+    def add(self, response: Any) -> None:
```
gnossen, line 64: `Any` -> `health_pb2.HealthCheckResponse`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -56,18 +61,19 @@ def next(self):
...
-    def close(self):
+    def close(self) -> None:
         with self._condition:
             self._open = False
             self._condition.notify()
 
 
-def _watcher_to_send_response_callback_adapter(watcher):
+def _watcher_to_send_response_callback_adapter(
+        watcher: _Watcher) -> Callable[[Any], None]:
```
gnossen, line 76: `Any` -> `Optional[health_pb2.HealthCheckResponse]`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -80,18 +86,24 @@ def send_response_callback(response):
...
     """Servicer handling RPCs for service statuses."""
+    _lock: threading.RLock
+    _server_status: Mapping[str, Any]
+    _send_response_callbacks: Mapping[str, Callable[[Any], None]]
+    _gracefully_shutting_down: bool
 
     def __init__(self,
-                 experimental_non_blocking=True,
-                 experimental_thread_pool=None):
+                 experimental_non_blocking: Optional[bool] = True,
```
gnossen, line 95: I don't think this is `Optional`.

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -80,18 +86,24 @@ def send_response_callback(response):
 
 class HealthServicer(_health_pb2_grpc.HealthServicer):
     """Servicer handling RPCs for service statuses."""
+    _lock: threading.RLock
+    _server_status: Mapping[str, Any]
```
gnossen, line 90: `Any` -> `int`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -80,18 +86,24 @@ def send_response_callback(response):
...
+    _send_response_callbacks: Mapping[str, Callable[[Any], None]]
```
gnossen, line 91: I think this is `Mapping[str, Set[Callable[[health_pb2.HealthCheckResponse], None]]]`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -80,18 +86,24 @@ def send_response_callback(response):
...
+                     futures.ThreadPoolExecutor] = None):
         self._lock = threading.RLock()
         self._server_status = {"": _health_pb2.HealthCheckResponse.SERVING}
         self._send_response_callbacks = {}
         self.Watch.__func__.experimental_non_blocking = experimental_non_blocking
         self.Watch.__func__.experimental_thread_pool = experimental_thread_pool
         self._gracefully_shutting_down = False
 
-    def _on_close_callback(self, send_response_callback, service):
+    def _on_close_callback(self, send_response_callback: Callable[[Any], None],
```
gnossen, line 105: `Any` -> `health_pb2.HealthCheckResponse`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -101,7 +113,8 @@ def callback():
 
         return callback
 
-    def Check(self, request, context):
+    def Check(self, request: _health_pb2.HealthCheckRequest,
+              context: Any) -> _health_pb2.HealthCheckResponse:
```
gnossen, line 117: `Any` -> `ServicerContext`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -111,7 +124,12 @@ def Check(self, request, context):
                 return _health_pb2.HealthCheckResponse(status=status)
 
     # pylint: disable=arguments-differ
-    def Watch(self, request, context, send_response_callback=None):
+    def Watch(
+        self,
+        request: _health_pb2.HealthCheckRequest,
+        context: Any,
```
gnossen, line 130: `Any` -> `ServicerContext`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -111,7 +124,12 @@ def Check(self, request, context):
...
+        send_response_callback: Optional[Callable[[Any], None]] = None
```
gnossen, line 131: `Any` -> `health_pb2.HealthCheckResponse`

src/python/grpcio_health_checking/grpc_health/v1/health.py
```diff
@@ -134,7 +152,8 @@ def Watch(self, request, context, send_response_callback=None):
                 self._on_close_callback(send_response_callback, service))
         return blocking_watcher
 
-    def set(self, service, status):
+    def set(self, service: str,
+            status: _health_pb2.HealthCheckResponse.ServingStatus) -> None:
```
gnossen, line 156: I think this should just be `int`. I also think this should have failed. Can you please make sure that type checking is working properly here? You may need to start generating type stubs for this package by modifying the corresponding `setup.py`.

src/python/grpcio_status/grpc_status/rpc_status.py
```diff
@@ -62,7 +63,7 @@ def from_call(call):
     return None
 
 
-def to_status(status):
+def to_status(status: status_pb2.Status) -> _Status:
```
gnossen, line 66: `_Status` -> `grpc.Status`. `_Status` is an internal type. We need to use the public interface, not the internal implementation.

gnossen (commented):
With this batch of comments, the whole PR should be covered.

src/python/grpcio/grpc/__init__.py
```diff
@@ -392,7 +412,7 @@ def code(self):
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def details(self):
+    def details(self) -> Optional[str]:
```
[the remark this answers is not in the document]
XuanWang-Amos: For return it make sense, `decode` returns `str` is still valid for `Optional[str]`, but for the input, `Anystr` shouldn't allow `None`.

I tested using the following code, looks like Pytype is not able to detect this issue:
```
def decode(b: AnyStr) -> Optional[str]:
    if isinstance(b, bytes):
        return b.decode('utf-8', 'replace')
    return b

class _RPCState(object):
    details: Optional[str]

    def __init__(self, details: Optional[str]):
        self.details = decode(details)

state = _RPCState(None) # No complain here
decode(None) # Pytype complains
```

src/python/grpcio/grpc/__init__.py
```diff
@@ -426,7 +446,10 @@ class UnaryUnaryClientInterceptor(abc.ABC):
     """Affords intercepting unary-unary invocations."""
 
     @abc.abstractmethod
-    def intercept_unary_unary(self, continuation, client_call_details, request):
+    def intercept_unary_unary(self, continuation: Callable[
+        [ClientCallDetails, RequestType], Any],
```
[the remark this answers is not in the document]
XuanWang-Amos: Make sense, I'll add those `TODO`s.

src/python/grpcio_health_checking/grpc_health/v1/health.py
XuanWang-Amos: `_health_pb2.HealthCheckResponse.ServingStatus` was treated as `Any` thus Pytype didn't complain here.

src/python/grpcio/grpc/_server.py
```diff
@@ -361,9 +363,9 @@ def abort(self, code: grpc.StatusCode, details: str) -> None:
             self._state.aborted = True
             raise Exception()
 
-    def abort_with_status(self, status: grpc.Status) -> None:
+    def abort_with_status(self, status: grpc.Status) -> NoReturn:
         self._state.trailing_metadata = status.trailing_metadata
-        self.abort(status.code, status.details)
+        self.abort(status.code, status.details)  # pytype: disable=bad-return-type
```
XuanWang-Amos, line 368: Pytype is complaining here return should be `None`, but since `abort` always throws error, we don't expect any return here.

src/python/grpcio_health_checking/grpc_health/v1/health.py
gnossen: Ah, because of the lack of stubs. We need to make sure that the corresponding `_pb2.pyi` files get generated for each of the ancillary packages (and make sure that they're actually getting picked up too).

gnossen:
We'll want to add [a `py.typed` file](https://peps.python.org/pep-0561/#packaging-type-information) to each package to indicate that it supports type checking.

src/python/grpcio/grpc/__init__.py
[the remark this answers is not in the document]
gnossen: I did some digging into this and, unfortunately, I think the answer is to add yet _another_ type checker: `mypy`. See [this explanation here](https://google.github.io/pytype/typing_faq.html#:~:text=before%20being%20used.-,Why%20is%20pytype%20not%20more%20like%20%24other_language%3F,-We%20often%20get) and [these slides](https://docs.google.com/presentation/d/1GYqLeLkknjYaYX2JrMzxX8LGw_rlO-6kTk-VNPVG9gY/edit#slide=id.p) on the specific differences between `pytype` and `mypy`. I've pulled `mypy` down and can confirm that it detects this issue.

src/python/grpcio/grpc/__init__.py
```diff
@@ -616,7 +646,8 @@ def __call__(self, metadata, error):
 class AuthMetadataPlugin(abc.ABC):
     """A specification for custom authentication."""
 
-    def __call__(self, context, callback):
+    def __call__(self, contex: AuthMetadataContext,
```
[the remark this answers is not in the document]
gnossen: Actually, now that I think about it, the name of this parameter doesn't matter since it's positional and this is just the interface, not a concrete implementation.

src/python/grpcio/grpc/__init__.py
```diff
@@ -1160,7 +1205,8 @@ def send_initial_metadata(self, initial_metadata):
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def set_trailing_metadata(self, trailing_metadata):
+    def set_trailing_metadata(
+            self, trailing_metadata: Optional[MetadataType]) -> None:
```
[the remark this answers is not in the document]
gnossen: Your example is the opposite of the situation above. `Optional[int]` should be on the parent, not the child. The following, _does_ fail with `pytype` for me:

```python
import abc

from typing import Optional

class A(metaclass=abc.ABCMeta):
    @abc.abstractmethod
    def foo(self, bar: Optional[int]) -> None:
        raise NotImplementedError()

class B(A):
    def foo(self, bar: int) -> None:
        return

b = B()
print(b.foo(123))
```

```
File "/usr/local/google/home/rbellevi/dev/tmp/pytype-testing/subclass2.py", line 11, in B: Overriding method signature mismatch [signature-mismatch]
  Base signature: 'def A.foo(self, bar: Optional[int]) -> None'.
  Subclass signature: 'def B.foo(self, bar: int) -> None'.
  Type mismatch for parameter 'bar'.
```

We need to get to the bottom of why the pytype check running in CI did not catch this.

bazel/internal_python_rules.bzl
```diff
@@ -38,6 +41,7 @@ def internal_py_grpc_test(name, **kwargs):
         tests = [
             name + ".native",
             name + ".gevent",
+            name + ".typecheck",
```
gnossen, line 44: I think `typecheck` is too generic a suffix here. Let's go with `typeguard` instead.

src/python/grpcio_tests/tests/reflection/_reflection_client_test.py
```diff
@@ -15,6 +15,10 @@
 
 import unittest
 
+from typeguard.importhook import install_import_hook
```
gnossen, line 18: This is leaking a detail of the typeguard test into a non typeguard test file that needs to be able to run even when typeguard is not installed. Can we just include all ancillary packages in `bazel/_run_time_type_check_main.py`?

Same comment for every file besides `bazel/_run_time_type_check_main.py` that calls `install_import_hook`.

src/python/grpcio_tests/tests/unit/_abort_test.py
```diff
@@ -140,6 +140,8 @@ def test_abort_with_status(self):
         self.assertEqual(rpc_error.details(), _ABORT_DETAILS)
         self.assertEqual(rpc_error.trailing_metadata(), _ABORT_METADATA)
 
+    @unittest.skipIf(test_common.running_under_run_time_type_check(),
+                     "This test case used unsupportted types")
```
gnossen, line 144: Nit: `s/unsupportted/unsupportedg` here and throughout.

tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
```diff
@@ -19,6 +19,8 @@ RESULTSTORE_RESULTS_FLAG="--bazelrc=tools/remote_build/include/test_locally_with
 TEST_TARGETS="//src/python/... //tools/distrib/python/grpcio_tools/... //examples/python/..."
 BAZEL_FLAGS="--test_output=errors"
 
+python3 -m pip install typeguard
```
gnossen, line 22: [We need to include this dependency via Bazel](https://github.com/grpc/grpc/blob/843cf42d6a76291b98be267152a3c5d99de8c965/requirements.bazel.txt), not out of band.

tools/run_tests/helper_scripts/build_python.sh
```diff
@@ -165,7 +165,7 @@ pip_install_dir_and_deps() {
 pip_install -U gevent
 
 pip_install --upgrade cython
-pip_install --upgrade six protobuf>=4.21.3
+pip_install --upgrade six protobuf>=4.21.3 typeguard
```
gnossen, line 168: Likewise, since `typeguard` is only a dependency of a Bazel-run test, we should not be installing it at the system level here.

gnossen (commented):
This runtime type checking work is great! I'm feeling a lot more confidence that there will be no issues with these types now.

We'll want to add [a `py.typed` file](https://peps.python.org/pep-0561/#packaging-type-information) to each package to indicate that it supports type checking.

PR #31778, opened 2022-11-30
[cleanups] Remove `include/grpc/impl/codegen/port_platform.h`

ralphchung:
This PR is a part of [b/196639718](http://b/196639718). It tries to move the content of `include/grpc/impl/codegen/port_platform.h` to `include/grpc/support/port_platform.h`. We do not remove anything to achieve backward compatibility. The file `include/grpc/impl/codegen/port_platform.h` should be removed in the future.


ralphchung:
Corresponding internal change is blocked by removing `class CoreCodegen`.

drfloob (approved):

PR #31785, opened 2022-12-01
Fixing some bugs in posix event engine

Vignesh2208:
The bugs fixed here were unearthed when testing event engine clients with iomgr shim layer in https://github.com/grpc/grpc/pull/31661

The bugs fixed here fall under the following main categories:

* error annotations: The absl Status codes retuned in posix_engine/ev_epoll1_linux.cc, posix_engine/ev_poll_posix.cc and posix_engine/posix_endpoint.cc are modified to include some extra annotations and GRPC status codes. This conforms with the tcp_annotate_error function defined in src/core/lib/iomgr/tcp_posix.cc

* data corruption in posix_endpoint.cc: A new MutableData method is added to event_engine/SliceBuffer.h - which allows peeking into individual slices. This fixes a bug associated with the previous usage here a copy of the slice may be returned in some cases leading to use-after free bugs when the copy goes out of scope.

* changes to error messages: (posix_engine.cc, posix_endpoint.cc): The absl::Status error messages are changed to those expected by some tests.

* issues in tcp_socket_utils: The fixes here mostly correspond to bugs in converting a UDS socket address to the correct URI. There are also issues related to incorrectly populated length fields of some returned EventEngine::ResolvedAddress types.


include/grpc/event_engine/slice_buffer.h
```diff
@@ -118,12 +125,26 @@ class SliceBuffer {
   /// associated slice.
   Slice RefSlice(size_t index);
 
+  /// Returns pointer to the buffer contained in the slice at the specified
+  /// index. The returned buffer is mutable.
+  std::tuple<uint8_t*, size_t> MutableData(size_t index) {
+    return std::make_tuple<uint8_t*, size_t>(
+        GPR_SLICE_START_PTR(slice_buffer_.slices[index]),
```
ralphchung, line 132: Please use GPRC_SLICE_* since GPR_SLICE_* are being deprecated. https://github.com/grpc/grpc/pull/31774

include/grpc/event_engine/slice_buffer.h
```diff
@@ -118,12 +125,26 @@ class SliceBuffer {
...
+
   /// The total number of bytes held by the SliceBuffer
   size_t Length() { return slice_buffer_.length; }
 
   /// Return a pointer to the back raw grpc_slice_buffer
   grpc_slice_buffer* c_slice_buffer() { return &slice_buffer_; }
 
+  // Returns a SliceBuffer that transfers slices into this new SliceBuffer,
+  // leaving the input parameter empty.
+  static SliceBuffer TakeCSliceBuffer(grpc_slice_buffer& slice_buffer) {
```
ralphchung, line 144: It is coding style thing. Since we are modifying input `slice_buffer`, should we use `grpc_slice_buffer* slice_buffer` instead?

include/grpc/event_engine/slice_buffer.h
```diff
@@ -118,12 +125,26 @@ class SliceBuffer {
...
+    return SliceBuffer(&slice_buffer);
```
ralphchung, line 145: Should use move ctor by `return SliceBuffer(std::move(slice_buffer));`

include/grpc/event_engine/slice_buffer.h
```diff
@@ -118,12 +125,26 @@ class SliceBuffer {
   /// associated slice.
   Slice RefSlice(size_t index);
 
+  /// Returns pointer to the buffer contained in the slice at the specified
+  /// index. The returned buffer is mutable.
+  std::tuple<uint8_t*, size_t> MutableData(size_t index) {
```
drfloob, line 130: Craig and I discussed alternatives in the context of my WindowsEventEngine work as well, we should discuss it next week. This is a public API change, and needs to be carefully thought through. I'm not sure adding raw slice data accessor methods to the SliceBuffer is the right choice here, if anything they belong on the Slice type, but RefSlice might not get us the Slice we need (I implemented `PeekSlice` for this reason, but it's not ideal either).

Is my hypothesis correct that the bugs you're seeing are with uses of inlined slices in the Write API?

include/grpc/event_engine/slice_buffer.h
drfloob, line 144: This appears to be unused.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
```diff
@@ -86,7 +86,7 @@ void LockfreeEvent::DestroyEvent() {
     // with post-deletion (see the note in the constructor) we want the bit
     // pattern to prevent error retention in a deleted object
   } while (!state_.compare_exchange_strong(curr, kShutdownBit,
-                                           std::memory_order_relaxed,
+                                           std::memory_order_acq_rel,
```
drfloob, line 89: Why was this necessary? I imagine there may be performance implications.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -1132,7 +1165,12 @@ void PosixEndpointImpl::Write(
     current_zerocopy_send_ = zerocopy_send_record;
     handle_->NotifyOnWrite(on_write_);
   } else {
-    on_writable(status);
+    // TODO(vigneshbabu): Consider eventually running this callback inline to
+    // avoid a thread hop. At the time of submission, it causes deadlocks which
+    // should be reolved after ExecCtx removal.
+    engine_->Run([on_writable = std::move(on_writable), status]() mutable {
+      on_writable(status);
```
drfloob, line 1172: Unless you're 100% sure it isn't necessary, EE callbacks should begin with local defines of ExecCtx and ApplicationExecCtx variables. They aren't guaranteed to exist otherwise.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -1167,11 +1207,22 @@ PosixEndpointImpl::PosixEndpointImpl(EventHandle* handle,
...
-      *sock.PeerAddressString());
+      peer_addr_string.ok() ? *peer_addr_string : "");
   self_reservation_ = memory_owner_.MakeReservation(sizeof(PosixEndpointImpl));
-  local_address_ = *sock.LocalAddress();
-  peer_address_ = *sock.PeerAddress();
+  auto local_address = sock.LocalAddress();
+  if (local_address.ok()) {
+    local_address_ = *local_address;
+  } else {
+    local_address_ = EventEngine::ResolvedAddress{};
```
drfloob, line 1218: `local_address_` should be default-constructed already.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -1167,11 +1207,22 @@ PosixEndpointImpl::PosixEndpointImpl(EventHandle* handle,
...
+  }
+  auto peer_address = sock.PeerAddress();
+  if (peer_address.ok()) {
+    peer_address_ = *peer_address;
+  } else {
+    peer_address_ = EventEngine::ResolvedAddress{};
```
drfloob, line 1224: Same comment as above.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -488,6 +488,8 @@ class PosixEndpointImpl : public grpc_core::RefCounted<PosixEndpointImpl> {
     return local_address_;
   }
 
+  int Fd() { return handle_->WrappedFd(); }
```
drfloob, line 491: This method does not appear to be used. Internal-only? Also, for any posix-specific behavior, consider if it will need to be supported for other posix-compliant engines. If so, we should figure out where that common interface should live.

drfloob (changes requested):

include/grpc/event_engine/slice_buffer.h
```diff
@@ -118,12 +125,26 @@ class SliceBuffer {
...
+    return std::make_tuple<uint8_t*, size_t>(
+        GPR_SLICE_START_PTR(slice_buffer_.slices[index]),
+        GPR_SLICE_LENGTH(slice_buffer_.slices[index]));
+  }
+
   /// The total number of bytes held by the SliceBuffer
   size_t Length() { return slice_buffer_.length; }
```
ralphchung, line 137: Probably unrelated -- this could be a const member function.

include/grpc/event_engine/slice_buffer.h
Vignesh2208: Yes the bugs were exactly due to the use of out-of-scope inlined slices in the Write API.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
Vignesh2208: Its safer to use acq_rel because there is a load of state_ and store back into state_. Using relaxed for both these load and store operations may not be safe because other threads may also load and store from the state_ variable. We should typically only use relaxed only when we know there wont be any contention on the atomic variable while the operation is being performed. I think under the hood, acquire and release trigger some cache synchronizations which are not done with relaxed semantics.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
Vignesh2208: I think rather than taking a dependency on ExecCtx here, we can enforce inside core that the on_write callback should contain an ExecCtx and an ApplicationExecCtx within the callback body. This would be similar to how were are migrating the RunAfter timers.

include/grpc/event_engine/slice_buffer.h
Vignesh2208: what alternatives are you thinking ?

include/grpc/event_engine/slice_buffer.h
Vignesh2208: Good point. Fixed.

include/grpc/event_engine/slice_buffer.h
```diff
@@ -51,7 +52,13 @@ namespace experimental {
 /// an experimental API.
 class SliceBuffer {
  public:
-  explicit SliceBuffer() { grpc_slice_buffer_init(&slice_buffer_); }
+  SliceBuffer() { grpc_slice_buffer_init(&slice_buffer_); }
+  // Transfers slices into this new SliceBuffer, leaving the parameter empty.
+  // Does not take ownership of the slice_buffer argument.
+  explicit SliceBuffer(grpc_slice_buffer* slice_buffer) {
```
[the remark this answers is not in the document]
Vignesh2208: This constructor wasn't used in this PR. Removed.

include/grpc/event_engine/slice_buffer.h
Vignesh2208: This wasn't used in this PR. Removed.

include/grpc/event_engine/slice_buffer.h
Vignesh2208: This constructor wasn't used in this PR. Removed. But I dont think the slice_buffer needs to be moved. Its already a reference so no gains to be had in moving it.

include/grpc/event_engine/slice_buffer.h
Vignesh2208: Removed.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc

src/core/lib/event_engine/posix_engine/posix_endpoint.cc

src/core/lib/event_engine/posix_engine/posix_endpoint.h
Vignesh2208: It will be needed in the iomgr shim layer. I've removed it for now in this PR though.

include/grpc/event_engine/slice_buffer.h
Vignesh2208: Fixed.

include/grpc/event_engine/slice_buffer.h
ralphchung: What I heard about this class `SliceBuffer` is that the interface for this class should be stable. I am not sure if adding new API needs more discussion.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
ralphchung: Has `std::memory_order_relaxed` caused any problems? I think your reasoning is somehow different from what I understand. First, since `state_` is an atomic variable, there should not have race condition on the variable itself, which is not a valid reason to replace `memory_order_relaxed` with `memory_order_acq_rel`. Second, I think `std::memory_order_relaxed` is about we don't care about the order of operations to the variable `state_`.

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -74,6 +76,47 @@ absl::StatusOr<std::string> GetScheme(
...
+  const sockaddr_un* unix_addr = reinterpret_cast<const sockaddr_un*>(addr);
+  bool abstract =
+      resolved_addr->size() < sizeof(unix_addr->sun_family) + 1 ||
+      (unix_addr->sun_path[0] == '\0' && unix_addr->sun_path[1] != '\0');
+  std::string scheme;
+  std::string path;
+  if (abstract) {
+    scheme = "unix-abstract";
+    if (resolved_addr->size() < sizeof(unix_addr->sun_family) + 1) {
+      path = "";
```
ralphchung, line 97: Is this necessary?

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -305,9 +348,17 @@ absl::StatusOr<std::string> ResolvedAddressToURI(
   if (resolved_address.size() == 0) {
     return absl::InvalidArgumentError("Empty address");
   }
-  auto scheme = GetScheme(resolved_address);
+  EventEngine::ResolvedAddress resolved_addr = resolved_address;
```
ralphchung, line 351: The name looks confusing with `resolved_address`.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
Vignesh2208: memory_order_relaxed may cause the compiler to re-arrange other load/store operations across the atomic. For e.g any store that happens after the atomic in the code may be moved by the compiler before the atomic operation. This can cause un-intended contention on those variables.

src/core/lib/event_engine/tcp_socket_utils.cc
Vignesh2208: Changed it name simply to addr

src/core/lib/event_engine/tcp_socket_utils.cc
Vignesh2208: No removed it now.

include/grpc/event_engine/slice_buffer.h
ralphchung: Any updates for this?

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -70,6 +72,45 @@ absl::StatusOr<std::string> GetScheme(
           "Unknown scheme: %d", resolved_address.address()->sa_family));
   }
 }
+
+#ifdef GRPC_HAVE_UNIX_SOCKET
+absl::StatusOr<std::string> ResolvedAddrToUriUnixIfPossible(
+    EventEngine::ResolvedAddress* resolved_addr) {
```
ralphchung, line 78: Is checking `nullptr` necessary?

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -70,6 +72,45 @@ absl::StatusOr<std::string> GetScheme(
...
+  }
+  const sockaddr_un* unix_addr = reinterpret_cast<const sockaddr_un*>(addr);
+  bool abstract =
+      resolved_addr->size() < sizeof(unix_addr->sun_family) + 1 ||
+      (unix_addr->sun_path[0] == '\0' && unix_addr->sun_path[1] != '\0');
+  std::string scheme;
+  std::string path;
+  if (abstract) {
+    scheme = "unix-abstract";
+    if (resolved_addr->size() >= sizeof(unix_addr->sun_family) + 1) {
```
ralphchung, line 92: So it is possible for `path` to be an empty string?

include/grpc/event_engine/slice_buffer.h
drfloob: https://github.com/grpc/grpc/pull/31831

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
drfloob: I think that could work. We can stipulate that any non-EventEngine-internal uses of the following within core will need to instantiate their own ExecCtx and ApplicationCallbackExecCtx within the callback:

* `::Run`
* `::RunAfter`
* `::Connect`'s OnConnectCallback
* `::CreateListener`'s AcceptCallback and the on_shutdown callback
* `::DNSResolver::Lookup*` callbacks
* `::Endpoint::Read` and `::Endpoint::Write` callbacks.

src/core/lib/event_engine/tcp_socket_utils.cc
Vignesh2208: Not really. This function is static to this file and not exposed outside.

src/core/lib/event_engine/tcp_socket_utils.cc
Vignesh2208: Yes. The path can be empty. The purpose of this function is to merely return the path no matter what the path maybe.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
Vignesh2208: Added a simple benchmark now to test/core/event_engine/posix/lock_free_event_test.cc. There is no significant difference with memory_order_acq_rel or memory_order_relaxed.

Benchmark performance with std::memory_order_acq_rel:
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
BM_LockFreeEvent/threads:1        11.2 ns         11.2 ns     62053789
BM_LockFreeEvent/threads:2        5.61 ns         11.2 ns     62163488
BM_LockFreeEvent/threads:4        2.80 ns         11.2 ns     61730840
BM_LockFreeEvent/threads:8        1.72 ns         13.8 ns     61951176
BM_LockFreeEvent/threads:16      0.921 ns         14.7 ns     44932656
BM_LockFreeEvent/threads:32      0.539 ns         17.2 ns     40743104
BM_LockFreeEvent/threads:64      0.310 ns         18.9 ns     35077312

Benchmark performance with std::memory_order_relaxed
----------------------------------------------------------------------
Benchmark                            Time             CPU   Iterations
----------------------------------------------------------------------
BM_LockFreeEvent/threads:1        11.2 ns         11.2 ns     61997253
BM_LockFreeEvent/threads:2        5.60 ns         11.2 ns     62566640
BM_LockFreeEvent/threads:4        3.27 ns         13.1 ns     40000000
BM_LockFreeEvent/threads:8        1.53 ns         12.2 ns     61195560
BM_LockFreeEvent/threads:16      0.998 ns         15.9 ns     48845520
BM_LockFreeEvent/threads:32      0.557 ns         17.8 ns     38054176
BM_LockFreeEvent/threads:64      0.320 ns         18.5 ns     38627200

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -572,23 +599,27 @@ void PosixEndpointImpl::Read(absl::AnyInvocable<void(absl::Status)> on_read,
...
   } else {
     min_progress_size_ = 1;
   }
   Ref().release();
   if (is_first_read_) {
+    UpdateRcvLowat();
     // Endpoint read called for the very first time. Register read callback
     // with the polling engine.
     is_first_read_ = false;
+    read_mu_.Unlock();
```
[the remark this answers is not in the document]
Vignesh2208: Updated to use absl::ReleasableMutexLock

include/grpc/event_engine/slice_buffer.h
ralphchung: If we still need to wait for the PR, at least left some TODO comments.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
ralphchung: I am ok with this. Please have @drfloob take a look as well.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -566,29 +594,33 @@ void PosixEndpointImpl::HandleRead(absl::Status status) {
 void PosixEndpointImpl::Read(absl::AnyInvocable<void(absl::Status)> on_read,
                              SliceBuffer* buffer,
                              const EventEngine::Endpoint::ReadArgs* args) {
-  read_mu_.Lock();
+  absl::ReleasableMutexLock lock(&read_mu_);
```
ralphchung, line 597: I was proposing using RAII to drop the need to explicitly call `lock.Release()`. Reading the logic, I don't know if there is a better way to achieve that.

Vignesh2208: Yes. ReleasableMutexLock will ensure the lock is released if not already done so when the function returns. Its safer.

ralphchung:
I have reviewed this PR, but I will let @drfloob to make the final call.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
drfloob: LGTM

drfloob:
This LGTM so far, but slice_cast should land soon, and this will need a re-review afterwards.

Vignesh2208:
> This LGTM so far, but slice_cast should land soon, and this will need a re-review afterwards.

Updated it to use SliceCast now. Could you take a look :) Thanks!

drfloob:
Update: this should use Craig's mutable SliceCast https://github.com/grpc/grpc/pull/31959. It's probably worth waiting a day for it, rather than using const_cast.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
```diff
@@ -141,8 +142,10 @@ void LockfreeEvent::NotifyOn(PosixEngineClosure* closure) {
         // contains a pointer to the shutdown-error). If the fd is shutdown,
         // schedule the closure with the shutdown error
         if ((curr & kShutdownBit) > 0) {
+          ABSL_ANNOTATE_IGNORE_READS_BEGIN();
```
drfloob, line 145: Why do we want to ignore races on the following line?

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
```diff
@@ -125,7 +125,7 @@ absl::Status PrepareTcpClientSocket(PosixSocketWrapper sock,
   return absl::OkStatus();
 }
 
-#endif  // GRPC_POSIX_SOCKET_UTILS_COMMON
+#endif /* GRPC_POSIX_SOCKET_UTILS_COMMON */
```
drfloob, line 128: Why change these? This undoes some recent work in converting c-style comments to c++-style comments. https://github.com/grpc/grpc/pull/31923

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -224,10 +278,15 @@ absl::optional<int> ResolvedAddressIsWildcard(
...
-  if (ResolvedAddressIsV4Mapped(resolved_addr, &addr_normalized)) {
-    return ResolvedAddressToString(addr_normalized);
+  if (!ResolvedAddressIsV4Mapped(resolved_addr, &addr_normalized)) {
+    addr_normalized = resolved_addr;
   }
-  return ResolvedAddressToString(resolved_addr);
+  auto scheme = GetScheme(addr_normalized);
+  GRPC_RETURN_IF_ERROR(scheme.status());
+  if (*scheme == "unix") {
+    return ResolvedAddrToUriUnixIfPossible(&addr_normalized, true);
```
drfloob, line 287: The `ResolvedAddressToURI` function relies on the family of `To*String` functions, and they serve slightly different purposes. For a `To*String` function to now rely on and return a URI adds some confusing inconsistency. I'd recommend having `ResolvedAddressToNormalizedString` handle Unix sockaddrs internally. And if a client needs a URI, they should call a function that explicitly returns a URI (not a `To*String` function).

src/core/lib/event_engine/posix_engine/lockfree_event.cc
Vignesh2208: Removed for now. But might need to add it back later

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
Vignesh2208: Fixed.

src/core/lib/event_engine/tcp_socket_utils.cc
Vignesh2208: There was some logic which was re-usable. Created a ResolvedAddrToUnixPathIfPossible internal function now to be used by To*String functions.

src/core/lib/event_engine/posix_engine/lockfree_event.cc
drfloob: If we do consider reintroducing this, I think we should have air-tight reasons for intentionally ignoring races. As in, strong confidence that no race could be problematic, or justification like "this is intentionally a racey read because <X>."

tools/distrib/fix_build_deps.py
```diff
@@ -50,6 +50,8 @@
     # TODO(ctiller) remove this
     'absl/base/internal/endian.h':
         'absl/base',
+    'absl/base/dynamic_annotations.h':
```
drfloob, line 53: this is unnecessary now.

test/core/event_engine/posix/lock_free_event_test.cc
```diff
@@ -151,17 +153,61 @@ TEST(LockFreeEventTest, MultiThreadedTest) {
...
+  void Run(
+      grpc_event_engine::experimental::EventEngine::Closure* closure) override {
+    closure->Run();
+  }
+
+  void Run(absl::AnyInvocable<void()> cb) override { cb(); }
+};
+
+// A benchmark which repeatedly registers a NotifyOn callback and invokes the
+// callback with SetReady. This benchmakr is intended to measure the cost of
```
drfloob, line 172: s/benchmakr/benchmark/

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -224,10 +312,15 @@ absl::optional<int> ResolvedAddressIsWildcard(
...
     const EventEngine::ResolvedAddress& resolved_addr) {
   EventEngine::ResolvedAddress addr_normalized;
-  if (ResolvedAddressIsV4Mapped(resolved_addr, &addr_normalized)) {
-    return ResolvedAddressToString(addr_normalized);
+  if (!ResolvedAddressIsV4Mapped(resolved_addr, &addr_normalized)) {
+    addr_normalized = resolved_addr;
+  }
+  auto scheme = GetScheme(addr_normalized);
+  GRPC_RETURN_IF_ERROR(scheme.status());
+  if (*scheme == "unix") {
```
drfloob, line 320: Now that `ResolvedAddressToString` will return a unix path for unix sockets, I think this scheme checking is redundant.

Vignesh2208: Fixed.

test/core/event_engine/posix/lock_free_event_test.cc
Vignesh2208: Fixed.

tools/distrib/fix_build_deps.py
Vignesh2208: Removed.

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -111,36 +111,13 @@ absl::StatusOr<std::string> ResolvedAddrToUnixPathIfPossible(
...
-    scheme = "unix";
-    size_t maxlen = sizeof(unix_addr->sun_path);
-    if (strnlen(unix_addr->sun_path, maxlen) == maxlen) {
-      return absl::InvalidArgumentError("UDS path is not null-terminated");
-    }
-    path = unix_addr->sun_path;
-  }
+  auto path = ResolvedAddrToUnixPathIfPossible(resolved_addr);
+  GRPC_RETURN_IF_ERROR(path.status());
+  std::string scheme = path->rfind('\0', 0) == 0 ? "unix-abstract" : "unix";
```
drfloob, line 116: this can be simplified to `path->at(0) == '\0'`. See https://godbolt.org/z/9G5jGWTEz. Regarding bounds-checking, I think other bits of this function would fail if path were an empty string.

Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
```diff
@@ -609,7 +620,7 @@ bool PosixSocketWrapper::IsIpv6LoopbackAvailable() {
       sockaddr_in6 addr;
       memset(&addr, 0, sizeof(addr));
       addr.sin6_family = AF_INET6;
-      addr.sin6_addr.s6_addr[15] = 1;  // [::1]:0
+      addr.sin6_addr.s6_addr[15] = 1; /* [::1]:0 */
```
drfloob, line 623: you missed this instance of using c-style comments in your cleanup.

drfloob (approved):
Looks good! Just one final cleanup I commented on. Please address that before merging.

Vignesh2208:
> Looks good! Just one final cleanup I commented on. Please address that before merging.

Thanks. Fixed it.

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
Vignesh2208: Fixed.

PR #31788, opened 2022-12-01
xDS stateful session affinity: implement C-core filter

markdroth:
This is the initial code for the StatefulSession filter.  In a subsequent PR, I will add the xDS HTTP filter registration, which will include code to process the xDS config for this filter.

There are no tests here yet, so I have absolutely no guarantee that this code actually works properly.  I will chat with @ctiller next week to see if we are at a point where we have the necessary infrastructure to write a unit test for a promise-based filter.  But if not, we will still be able to flush out problems when we add the end-to-end tests.


src/core/ext/filters/stateful_session_affinity/stateful_session_affinity_filter.cc
```diff
@@ -0,0 +1,231 @@
...
+#include "src/core/lib/transport/metadata_batch.h"
+#include "src/core/lib/transport/transport.h"
+
+namespace grpc_core {
+
+TraceFlag grpc_stateful_session_affinity_filter_trace(
+    false, "stateful_session_affinity_filter");
+
+UniqueTypeName XdsHostOverrideTypeName() {
+  static UniqueTypeName::Factory kFactory("xds_host_override");
```
eugeneo, line 65: I would like to make sure we use consistent naming so we don't unnecessarily confuse the users. LB policy is "xds_override_host" (different word order). Which one would you prefer? I can rename the LB policy.

src/core/ext/filters/stateful_session_affinity/stateful_session_affinity_filter.cc
```diff
@@ -0,0 +1,231 @@
...
+  // For now, just choose the first value.
+  absl::string_view value = values.front();
+  // Base64-decode it.
+  std::string decoded_value;
+  if (!absl::Base64Unescape(value, &decoded_value)) return absl::nullopt;
+  // Copy it into the arena, since it will need to persist until the LB pick.
+  char* arena_value =
+      static_cast<char*>(GetContext<Arena>()->Alloc(decoded_value.size()));
+  memcpy(arena_value, decoded_value.c_str(), decoded_value.size());
+  return absl::string_view(arena_value, decoded_value.size());
```
eugeneo, line 223: I meant to ask about this before - but wouldn't it be safer to return an `std::string`? Compilers are good at optimizing so there should not be too many unnecessary copies.

src/core/ext/filters/stateful_session_affinity/stateful_session_affinity_service_config_parser.cc
```diff
@@ -0,0 +1,84 @@
...
+
+#include "src/core/lib/channel/channel_args.h"
+
+namespace grpc_core {
+
+const JsonLoaderInterface*
+StatefulSessionAffinityMethodParsedConfig::CookieConfig::JsonLoader(
+    const JsonArgs&) {
+  static const auto* loader = JsonObjectLoader<CookieConfig>()
+                                  .OptionalField("name", &CookieConfig::name)
```
eugeneo, line 33: If I understand `CookieConfig::JsonPostload` right, this field is not optional. Are there any reasons to declare it Optional here?

markdroth:
@ctiller As per our discussion, I've added a metadata annotation to indicate if it's Trailers-Only.  None of this is currently tested, so I'm not 100% sure it works, but we'll verify that in a future PR when we add an e2e test for this feature.

eugeneo (approved):
No comments for me, learned some more 😀

PR #31791, opened 2022-12-02
Add type stub generation for bazel  rules

kimmywork:
Added generating pyi stub option (`--pyi_out=...`) in `py_proto_library`.

### DIFF
--- a/bazel/python_rules.bzl
+++ b/bazel/python_rules.bzl
@@ -28,6 +28,7 @@ load(
 )
 
 _GENERATED_PROTO_FORMAT = "{}_pb2.py"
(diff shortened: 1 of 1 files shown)


gnossen:
@kimmywork Thanks for the contribution! This seems like a good change. Can you please add a test to [this directory](https://github.com/grpc/grpc/tree/master/test/distrib/bazel/python) to make sure that there isn't a regression in this functionality?

I'd imagine the test would do the following:
- depend on a `py_proto_library`
- do _some_ sort of type checking such that the test would fail if you reverted your implementation changes

PR #31794, opened 2022-12-02
Fix use-after-free when FakeHelper is freed before the test tear down

eugeneo:
Some tests may have FakeHelper freed before the TearDown, such as when the policy fails to be created. This causes an use-after-free error when check for no pending events is performed.

This PR fixes this issue by moving the check to the FakeHelper dtor.

### DIFF
--- a/test/core/client_channel/lb_policy/lb_policy_test_lib.h
+++ b/test/core/client_channel/lb_policy/lb_policy_test_lib.h
@@ -218,6 +218,16 @@ class LoadBalancingPolicyTest : public ::testing::Test {
                std::shared_ptr<WorkSerializer> work_serializer)
         : test_(test), work_serializer_(std::move(work_serializer)) {}
 
+    ~FakeHelper() override {
+      // Note: Can't safely trigger this from inside the FakeHelper dtor,
+      // because if there is a picker in the queue that is holding a ref
+      // to the LB policy, that will prevent the LB policy from being
+      // destroyed, and therefore the FakeHelper will not be destroyed.
+      // (This will cause an ASAN failure, but it will not display the
+      // queued events, so the failure will be harder to diagnose.)
+      ExpectQueueEmpty();
+    }
+
     // Called at test tear-down time to ensure that we have not left any
     // unexpected events in the queue.
     void ExpectQueueEmpty(SourceLocation location = SourceLocation()) {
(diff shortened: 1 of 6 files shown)


eugeneo:
Specifically, this addresses a memory leak in a test that checks that the policy could not have been created because it was behind the env var.

markdroth:
Can you show me a reproduction and point me at the leak?  I'd like to better understand what problem you're trying to solve.

In general, this doesn't seem like a good approach to me.  If we really wanted the state to outlive the helper, we could just store the state directly in the test object; we would not need to put it in a separate `HelperState` object.  And the reason that I didn't do that to begin with is that I was hedging against the possibility that in the future, we might have test cases that need to create more than one LB policy, in which case each one would need its own helper instance.

eugeneo:
I added a test case that caused the memory leak in my setup - it is when the helper is not even created. I am not suggesting adding it.

I also realized the best way to address this would be to call `ExpectQueueEmpty` from the `~FakeHelper`. Please take a look.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -218,6 +218,16 @@ class LoadBalancingPolicyTest : public ::testing::Test {
                std::shared_ptr<WorkSerializer> work_serializer)
         : test_(test), work_serializer_(std::move(work_serializer)) {}
 
+    ~FakeHelper() override {
+      // Note: Can't safely trigger this from inside the FakeHelper dtor,
```
markdroth, line 222: This comment explains why this approach won't work. :)

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -375,16 +385,6 @@ class LoadBalancingPolicyTest : public ::testing::Test {
...
       : work_serializer_(std::make_shared<WorkSerializer>()) {}
 
-  void TearDown() override {
-    // Note: Can't safely trigger this from inside the FakeHelper dtor,
-    // because if there is a picker in the queue that is holding a ref
-    // to the LB policy, that will prevent the LB policy from being
-    // destroyed, and therefore the FakeHelper will not be destroyed.
-    // (This will cause an ASAN failure, but it will not display the
-    // queued events, so the failure will be harder to diagnose.)
-    helper_->ExpectQueueEmpty();
```
markdroth, line 385: I think the right solution here is just to check whether `helper_` is null here before calling a method on it.

test/core/client_channel/lb_policy/non_existing_policy_test.cc
```diff
@@ -0,0 +1,63 @@
...
+#include "test/core/util/test_config.h"
+
+namespace grpc_core {
+namespace testing {
+namespace {
+
+class NonExistingPolicyTest : public LoadBalancingPolicyTest {};
+
+TEST_F(NonExistingPolicyTest, Basic) {
+  auto policy = MakeLbPolicy("non-existing-policy");
```
markdroth, line 48: To fix this test case, how about changing `MakeLbPolicy()` such that if it returns null, it resets `helper_` to null as well?

eugeneo:
This pull request is no longer relevant.

PR #31797, opened 2022-12-03
Log xDS node ID with verbosity INFO

mohanli-ml:
Log xDS node ID so that we can better debug DirectPath issues.


markdroth:
This is the wrong place to do this logging.  This will log the node ID only if the client is using the internal bootstrap config generated by the C2P resolver; if the client is instead using an external bootstrap config for PSM, then this log message will actually log the wrong node ID, because the bootstrap config generated by the C2P resolver will never be used.

I think the right place to do this is in the XdsClient, probably here:

https://github.com/grpc/grpc/blob/1e13612d4a12e9f7ad41a6c02dc43e3f4428565a/src/core/ext/xds/xds_client.cc#L1479

mohanli-ml:
Hi Mark, why the C2P resolver gets used in the PSM case? If not, then this c2p node ID will not be logged. For PSM since the node ID can be found in the bootstrap file, it is not very interested to log it.

mohanli-ml:
> This is the wrong place to do this logging. This will log the node ID only if the client is using the internal bootstrap config generated by the C2P resolver; if the client is instead using an external bootstrap config for PSM, then this log message will actually log the wrong node ID, because the bootstrap config generated by the C2P resolver will never be used.
> 
> I think the right place to do this is in the XdsClient, probably here:
> 
> https://github.com/grpc/grpc/blob/1e13612d4a12e9f7ad41a6c02dc43e3f4428565a/src/core/ext/xds/xds_client.cc#L1479

Hi Mark, why the C2P resolver gets used in the PSM case? If not, then this c2p node ID will not be logged. For PSM since the node ID can be found in the bootstrap file, it is not very interested to log it.

markdroth:
The C2P resolver will always be used for DirectPath.  However, if the same client is using both DirectPath and PSM, then they will use the external bootstrap config generated for PSM rather than the bootstrap config generated internally by the C2P resolver.  The bootstrap config in the C2P resolver gets used only if there is no external bootstrap config provided.

src/core/ext/xds/xds_client.cc
```diff
@@ -1477,6 +1477,10 @@ XdsClient::XdsClient(
     gpr_log(GPR_INFO, "[xds_client %p] creating xds client", this);
   }
   GPR_ASSERT(bootstrap_ != nullptr);
+  if (bootstrap_->node()) {
```
markdroth, line 1480: Nit: Please use `!= nullptr` here rather than treating a pointer as a bool.

src/core/ext/xds/xds_client.cc
```diff
@@ -1477,6 +1477,10 @@ XdsClient::XdsClient(
...
+    gpr_log(GPR_INFO, "[xds_client %p] creating xds client with node ID: %s",
```
markdroth, line 1481: I think this should just say `xDS node ID: %s`.

mohanli-ml:
> makes it impossible for the reviewer to see what has changed since their last review

Sorry for this, will keep in mind.

src/core/ext/xds/xds_client.cc
mohanli-ml: Done

src/core/ext/xds/xds_client.cc
mohanli-ml: Done.

markdroth (approved):

PR #31804, opened 2022-12-05
Make tcp client posix test more robust

Vignesh2208:
Previous approach to test connection cancellation is not very robust. Even if the backlog specified to listen is 1, the kernel continues to accept a certain number of SYN packets before dropping them. With this approach, we create client sockets in a loop and try to make each socket connect to the listening socket until the connection attempt times out. This loop attempts to identify the number of new connection attempts that will be allowed by the kernel before any subsequent connection attempts become pending indefinitely.

This bug was unearthed during testing undertaken with event engine endpoints in https://github.com/grpc/grpc/pull/31661 and it caused some flakiness.


test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,79 @@ void test_fails(void) {
...
-
+  int one = 1;
+  bool tried_ipv4 = false;
   /* create a phony server */
-  svr_fd = socket(AF_INET, SOCK_STREAM, 0);
-  ASSERT_GE(svr_fd, 0);
-  ASSERT_EQ(bind(svr_fd, (struct sockaddr*)addr, (socklen_t)resolved_addr.len),
-            0);
+  svr_fd = socket(AF_INET6, SOCK_STREAM, 0);
+  while (svr_fd < 0 ||
```
ralphchung, line 227: Using `while` looks a bit weird to me. It is actually not a loop.

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,79 @@ void test_fails(void) {
...
 
+  std::vector<int> client_sockets;
+  // Create and connect client sockets until the connection attempt times out.
+  // Even if the backlog specified to listen is 1, the kernel continues to
+  // accept a certain number of SYN packets before dropping them. This loop
+  // attempts to identify the number of new connection attempts that will
+  // be allowed by the kernel before any subsequent connection attempts
+  // become pending indefinitely.
+  while (true) {
+    int client_socket = socket(tried_ipv4 ? AF_INET : AF_INET6, SOCK_STREAM, 0);
```
ralphchung, line 252: Do you need to check if `client_socket < 0`?

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,79 @@ void test_fails(void) {
...
-            0);
+  svr_fd = socket(AF_INET6, SOCK_STREAM, 0);
+  while (svr_fd < 0 ||
+         bind(svr_fd, reinterpret_cast<sockaddr*>(resolved_addr.addr),
+              resolved_addr.len) != 0) {
+    if (tried_ipv4) {
+      gpr_log(GPR_ERROR,
+              "Skipping test. Failed to create a phony server bound to ipv6 or "
+              "ipv4 address");
+      return;
```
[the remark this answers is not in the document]
Vignesh2208: No we dont want to fail the test. If this happens, it will likely be due to an issue with the testing infrastructure.

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Added the check.

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,79 @@ void test_fails(void) {
...
+  std::vector<int> client_sockets;
+  // Create and connect client sockets until the connection attempt times out.
+  // Even if the backlog specified to listen is 1, the kernel continues to
+  // accept a certain number of SYN packets before dropping them. This loop
+  // attempts to identify the number of new connection attempts that will
+  // be allowed by the kernel before any subsequent connection attempts
+  // become pending indefinitely.
+  while (true) {
+    int client_socket = socket(tried_ipv4 ? AF_INET : AF_INET6, SOCK_STREAM, 0);
+    setsockopt(client_socket, SOL_SOCKET, SO_REUSEADDR, &one, sizeof(one));
```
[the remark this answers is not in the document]
Vignesh2208: Done.

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: I was trying to reuse the code but I guess it affected readability. Unrolled the loop now.

test/core/iomgr/tcp_client_posix_test.cc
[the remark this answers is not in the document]
ralphchung: Then how can we tell whether that happens even if it should be unlikely?

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,82 @@ void test_fails(void) {
...
+  // Create and connect client sockets until the connection attempt times out.
+  // Even if the backlog specified to listen is 1, the kernel continues to
+  // accept a certain number of SYN packets before dropping them. This loop
+  // attempts to identify the number of new connection attempts that will
+  // be allowed by the kernel before any subsequent connection attempts
+  // become pending indefinitely.
+  while (create_more_client_connections) {
+    const int kOne = 1;
+    int client_socket = socket(tried_ipv4 ? AF_INET : AF_INET6, SOCK_STREAM, 0);
+    ASSERT_GT(client_socket, 0);
```
ralphchung, line 256: Though unlikely, I think the proper check is `ASSERT_GE(client_socket, 0)`. 0 is a valid fd.

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,82 @@ void test_fails(void) {
...
+    setsockopt(client_socket, SOL_SOCKET, SO_REUSEADDR, &kOne, sizeof(kOne));
+    // Make fd non-blocking.
+    int flags = fcntl(client_socket, F_GETFL, 0);
+    EXPECT_EQ(fcntl(client_socket, F_SETFL, flags | O_NONBLOCK), 0);
```
ralphchung, line 260: It looks like it should be `ASSERT_EQ` instead.

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,82 @@ void test_fails(void) {
...
+                resolved_addr.len) == -1) {
+      if (errno == EINPROGRESS) {
+        struct pollfd pfd;
+        pfd.fd = client_socket;
+        pfd.events = POLLOUT;
+        pfd.revents = 0;
+        int ret = poll(&pfd, 1, 1000);
+        if (ret == -1) {
+          gpr_log(GPR_ERROR, "poll() failed during connect; errno=%d", errno);
+          abort();
```
drfloob, line 272: Prefer using FAIL over abort in googletests. `FAIL() << "poll() failed during connect; errno=" << errno;`

test/core/iomgr/tcp_client_posix_test.cc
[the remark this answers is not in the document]
Vignesh2208: A failure here is not due to any bug in the gRPC code. We dont want to block PR submissions on that

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Fixed.

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Fixed.

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Fixed

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,80 @@ void test_fails(void) {
...
-  resolved_addr.len = static_cast<socklen_t>(sizeof(struct sockaddr_in));
-  addr->sin_family = AF_INET;
-
+  bool tried_ipv4 = false;
   /* create a phony server */
-  svr_fd = socket(AF_INET, SOCK_STREAM, 0);
-  ASSERT_GE(svr_fd, 0);
-  ASSERT_EQ(bind(svr_fd, (struct sockaddr*)addr, (socklen_t)resolved_addr.len),
-            0);
+  svr_fd = socket(AF_INET6, SOCK_STREAM, 0);
```
ralphchung, line 225: Just for the readability, I think having helper functions might help the entire flow more neat and clean.

```c++
/// returns socket file descriptor or error status
absl::StatusOr<int> TryIpv6(absl::string_view address); /// or any better function name
absl::StatusOr<int> TryIpv4(absl::string_view address);

auto ret = TryIpv6(absl::StrCat("ipv6:[::1]:", std::to_string(grpc_pick_unused_port_or_die())));
if (!ret.ok()) {
  ret = TryIpv4(absl::StrCat("ipv4:127.0.0.1:", std::to_string(grpc_pick_unused_port_or_die()));
}
if (!ret.ok()) {
  gpr_log(GPR_ERROR, ...);
  return;
}
const int server_fd = *ret;
...
```

ralphchung (approved):
Left another comment for readability. I am fine with most of the code. Whether taking the comment is your call.

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Cool. Updated it to use std::functions to make it a bit more readable.

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,84 @@ void test_fails(void) {
...
-      reinterpret_cast<struct sockaddr_in*>(resolved_addr.addr);
   int svr_fd;
   grpc_closure done;
   grpc_core::ExecCtx exec_ctx;
-
-  memset(&resolved_addr, 0, sizeof(resolved_addr));
-  resolved_addr.len = static_cast<socklen_t>(sizeof(struct sockaddr_in));
-  addr->sin_family = AF_INET;
-
+  bool tried_ipv4 = false;
```
ralphchung, line 222: Should not need this anymore.

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,84 @@ void test_fails(void) {
...
+  ASSERT_TRUE(grpc_parse_uri(target_ipv6_addr_uri, &resolved_addr));
+  auto try_bind = [&](int sock) {
+    if (sock < 0 || bind(sock, reinterpret_cast<sockaddr*>(resolved_addr.addr),
+                         resolved_addr.len) != 0) {
+      return false;
+    }
+    return true;
```
ralphchung, line 229: Could be a one-liner `return (sock >= 0 && bind(...) == 0);`

test/core/iomgr/tcp_client_posix_test.cc
```diff
@@ -207,27 +211,84 @@ void test_fails(void) {
...
-  svr_fd = socket(AF_INET, SOCK_STREAM, 0);
-  ASSERT_GE(svr_fd, 0);
-  ASSERT_EQ(bind(svr_fd, (struct sockaddr*)addr, (socklen_t)resolved_addr.len),
-            0);
+  svr_fd = socket(AF_INET6, SOCK_STREAM, 0);
+  // Try ipv6
+  if (!try_bind(svr_fd)) {
+    // Failed to bind ipv6. Try ipv4
+    ASSERT_TRUE(grpc_parse_uri(target_ipv4_addr_uri, &resolved_addr));
+    svr_fd = socket(AF_INET, SOCK_STREAM, 0);
```
ralphchung, line 237: Remember to close `svr_fd` if it is valid.

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Fixed.

test/core/iomgr/tcp_client_posix_test.cc
Vignesh2208: Needed in line 263

ralphchung (approved):

PR #31807, opened 2022-12-05
Add global structure of allocators + experiment

ananda1066:
Roll-forward of https://github.com/grpc/grpc/pull/31759, after test errors fixed in separate PR.


src/core/lib/resource_quota/memory_quota.cc
```diff
@@ -424,6 +433,21 @@ void BasicMemoryQuota::Take(size_t amount) {
   if (prior >= 0 && prior < static_cast<intptr_t>(amount)) {
     if (reclaimer_activity_ != nullptr) reclaimer_activity_->ForceWakeup();
   }
+
+  if (IsFreeLargeAllocatorEnabled()) {
+    GrpcMemoryAllocatorImpl* chosen_allocator = nullptr;
+    int shard_idx = rand() % big_allocators_.shards.size();
```
ctiller, line 439: We should avoid using `rand()`... it's terrible on several axis (concurrency, and randomness).
It's probably ok to put an `absl::Bitgen` as a member variable and use `absl::Uniform` here.

ctiller (approved):

src/core/lib/resource_quota/memory_quota.cc
ananda1066: Discussed offline - using atomic per allocator to decide shard.

PR #31812, opened 2022-12-05
XdsClient: inject version string at construction time

allanrbo:
@markdroth 

Here's a suggestion for a refactoring to let the user_agent_name and user_agent_version be ctor injectable. This allows our internal use case to not rely on grpc_version_string(), and thereby a step closer to removing the very broad dep "//net/grpc".

Thoughts for alternatives:
 - I could also put these two strings in a struct called something like AgentStrings. Would simplify the ctors a little.
 - Or I could put them in XdsBootstrap::Node perhaps? Though XdsBootstrap seems like it corresponds to a YAML/JSON file or something, so decided that would probably be wrong.

### DIFF
--- a/src/core/ext/xds/xds_api.cc
+++ b/src/core/ext/xds/xds_api.cc
@@ -45,7 +45,6 @@
 #include "upb/upb.h"
 #include "upb/upb.hpp"
 
-#include <grpc/grpc.h>
 #include <grpc/impl/codegen/gpr_types.h>
 #include <grpc/status.h>
 #include <grpc/support/log.h>
@@ -65,8 +64,8 @@ XdsApi::XdsApi(XdsClient* client, TraceFlag* tracer,
       tracer_(tracer),
       node_(node),
       symtab_(symtab),
-      user_agent_name_(user_agent_name),
-      user_agent_version_(user_agent_version) {}
+      user_agent_name_(std::move(user_agent_name)),
+      user_agent_version_(std::move(user_agent_version)) {}
 
 namespace {
 
--- a/src/core/ext/xds/xds_api.cc
+++ b/src/core/ext/xds/xds_api.cc
@@ -58,36 +58,15 @@
 
 namespace grpc_core {
 
-// If gRPC is built with -DGRPC_XDS_USER_AGENT_NAME_SUFFIX="...", that string
-// will be appended to the user agent name reported to the xDS server.
-#ifdef GRPC_XDS_USER_AGENT_NAME_SUFFIX
-#define GRPC_XDS_USER_AGENT_NAME_SUFFIX_STRING \
-  " " GRPC_XDS_USER_AGENT_NAME_SUFFIX
-#else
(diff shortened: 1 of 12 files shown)


allanrbo:
I see the CI wants me to add a label, release notes: no, but the cogwheel icon next to labels is missing for me, so I guess I don't have access to do so.

src/core/ext/xds/xds_api.cc
```diff
@@ -58,36 +58,15 @@
...
       tracer_(tracer),
       node_(node),
       symtab_(symtab),
-      user_agent_name_(absl::StrCat("gRPC C-core ", GPR_PLATFORM_STRING,
-                                    GRPC_XDS_USER_AGENT_NAME_SUFFIX_STRING)),
-      user_agent_version_(
-          absl::StrCat("C-core ", grpc_version_string(),
-                       GRPC_XDS_USER_AGENT_NAME_SUFFIX_STRING,
-                       GRPC_XDS_USER_AGENT_VERSION_SUFFIX_STRING)) {}
+      user_agent_name_(user_agent_name),
```
markdroth, line 68: Both of these lines should use `std::move()` to avoid unnecessary string copies.

src/core/ext/xds/xds_client.cc
```diff
@@ -1471,7 +1472,8 @@ XdsClient::XdsClient(
       transport_factory_(std::move(transport_factory)),
       request_timeout_(resource_request_timeout),
       xds_federation_enabled_(XdsFederationEnabled()),
-      api_(this, &grpc_xds_client_trace, bootstrap_->node(), &symtab_),
+      api_(this, &grpc_xds_client_trace, bootstrap_->node(), &symtab_,
+           user_agent_name, user_agent_version),
```
markdroth, line 1476: Please use `std::move()` for these two parameters to avoid unnecessary string copies.

markdroth (commented):
This approach looks good.

I'm pretty sure there are some places in tests where we instantiate `XdsClient` that will need to be updated.

src/core/ext/xds/xds_api.cc
allanrbo: Done

src/core/ext/xds/xds_client.cc
allanrbo: Done

markdroth:
I've triggered the tests.

markdroth (approved):
I'll trigger the tests again.

allanrbo:
Error was:
```
test/cpp/server/load_reporter/load_reporter_test.cc:163
Value of: static_cast<double>(lb_feedback.server_utilization())
Expected: is approximately 0.62543432939541344 (absolute error <= 1.0000000000000001e-05)
  Actual: 0.615824 (of type double), which is -0.00961082 from 0.625434
```
Looks like flakiness. Is there a way to retrigger just this test?

markdroth:
I'm pretty sure that failure is an unrelated flake.  I think this is ready to merge.

PR #31813, opened 2022-12-06
[cleanup] Remove an empty loop

LeiW000:
The loop is empty. So remove it.

### DIFF
--- a/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
+++ b/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
@@ -2389,8 +2389,6 @@ static void read_action_locked(void* tp, grpc_error_handle error) {
       error = GRPC_ERROR_CREATE_REFERENCING("Failed parsing HTTP/2", errors,
                                             GPR_ARRAY_SIZE(errors));
     }
-    for (i = 0; i < GPR_ARRAY_SIZE(errors); i++) {
-    }
 
     if (t->initial_window_update != 0) {
       if (t->initial_window_update > 0) {


drfloob:
Thanks for offering to do a cleanup. Please sign the CLA, and we can take a look.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -2379,18 +2379,16 @@ static void read_action_locked(void* tp, grpc_error_handle error) {
   }
   std::swap(err, error);
   if (t->closed_with_error.ok()) {
-    size_t i = 0;
     grpc_error_handle errors[3] = {error, absl::OkStatus(), absl::OkStatus()};
-    for (; i < t->read_buffer.count && errors[1] == absl::OkStatus(); i++) {
+
+    for (size_t i = 0; i < t->read_buffer.count && errors[1] == absl::OkStatus(); i++) {
```
krishnacx, line 2384: do..while loop can be better here. Since the first iteration condition would be true anyways.

LeiW000: Good idea.

LeiW000:
@drfloob , I'm still working on signing the CLA. I have a question. Is the CLA I need to sign specific to Grpc project?

thisisnotapril:
@LeiW000 you appear to have signed the CLA already so all good there! https://github.com/grpc/grpc/pull/31813#issuecomment-1338560165

drfloob:
@LeiW000 it looks like this is failing sanity checks, you'll need to run clang-format to get the checks to pass. Let me know if you need help with that.

LeiW000:
@drfloob , @thisisnotapril , thanks for your help. I have run the clang-format script and updated the PR.

@krishnacx, I have addressed your comment with the latest PR.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
drfloob: It's possible to have an empty buffer. See test errors. Please revert this bit, and just remove the unused loop in this PR.

LeiW000:
I saw some failures on the sanity tests. But I don't think it has anything to do with the change. If I'm wrong, please let me know.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -2381,16 +2381,17 @@ static void read_action_locked(void* tp, grpc_error_handle error) {
   if (t->closed_with_error.ok()) {
     size_t i = 0;
     grpc_error_handle errors[3] = {error, absl::OkStatus(), absl::OkStatus()};
-    for (; i < t->read_buffer.count && errors[1] == absl::OkStatus(); i++) {
+
+    do {
```
drfloob, line 2385: Commenting again for visibility: I think the test failures are due to this change, `read_buffer.count` may be 0. I recommend reverting this piece, and we can land the cleanup of the empty loop.

eugeneo (approved):

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
drfloob: Looks good, thanks for the contribution. Merged.

PR #31816, opened 2022-12-06
An EventEngine subclass that may be implemented by posix based event engines

Vignesh2208:
The posix EE subclass interface is defined in include/grpc/event_engine/posix.h.

This is necessary to handle some file descriptor based operations (e.g RPC-switch) which dont need to be supported in other event engines.

Creating this PR to start some discussion.

### DIFF
--- a/src/core/lib/event_engine/posix.h
+++ b/src/core/lib/event_engine/posix.h
@@ -94,6 +94,9 @@ class PosixListenerWithFdSupport : public EventEngine::Listener {
   /// connection.
   virtual absl::Status HandleExternalConnection(int listener_fd, int fd,
                                                 SliceBuffer* pending_data) = 0;
+
+  /// Shutdown/stop listening on all bind Fds.
+  virtual void ShutdownListeningFds() = 0;
 };
 
 /// Defines an interface that posix event engines may implement to


include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+    /// address.
+    using OnPosixBindNewFdCallback =
+        absl::AnyInvocable<void(absl::StatusOr<int> listener_fd)>;
+    /// Bind an address/port to this Listener.
+    ///
+    /// It is expected that multiple addresses/ports can be bound to this
+    /// Listener before Listener::Start has been called. Returns either the
+    /// bound port or an appropriate error status. The on_bind_new_fd callback
+    /// is invoked once for each newly bound listener fd that may be created by
+    /// this Bind operation.
```
krishnacx, line 98: Is description about function parameters no required ?

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+    /// \a listener_fd - The listening socket fd that accepted the new client
+    /// connection.
+    /// \a endpoint - The EventEngine endpoint to handle data exchange over the
+    /// new client connection.
+    /// \a memory_allocation - The callback may use the provided memory
+    /// allocator to handle memory allocation operations.
+    using PosixAcceptCallback = absl::AnyInvocable<void(
+        int listener_fd, std::unique_ptr<Endpoint> endpoint,
+        MemoryAllocator memory_allocator)>;
+    /// Called when a posix listener bind operation completes. A single bind
```
krishnacx, line 82: Add empty line before

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#ifndef GRPC_EVENT_ENGINE_POSIX_H
+#define GRPC_EVENT_ENGINE_POSIX_H
+
+#include <grpc/support/port_platform.h>
```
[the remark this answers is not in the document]
Vignesh2208: I think it should be the first header. Otherwise sanity checks would fail.

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+namespace experimental {
+
+/// This defines an EventEngine interface that all posix specific event engines
+/// must implement.
+class PosixEventEngine : public EventEngine {
+  class PosixEventEngineEndpoint : public EventEngine::Endpoint {
+   public:
+    /// Returns the file descriptor associated with the posix endpoint.
+    virtual int GetWrappedFd() = 0;
+    /// Shutdown the endpoint. After this function call its illegal to invoke
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+    /// not closed while the endpoint is shutdown. The pointer will be updated
+    /// to hold the endpoint's assocaited file descriptor.
+    /// \a on_release_cb - a callback to be invoked when the endpoint is
+    /// shutdown and the file descriptor is released. It should only be invoked
+    /// if release_fd is not nullptr.
+    virtual void Shutdown(
+        int* release_fd,
+        absl::AnyInvocable<void(absl::Status)> on_release_cb) = 0;
+  };
+  /// Creates a PosixEventEngineEndpoint from an fd which is already assumed to
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+  class PosixEventEngineEndpoint : public EventEngine::Endpoint {
+   public:
+    /// Returns the file descriptor associated with the posix endpoint.
+    virtual int GetWrappedFd() = 0;
+    /// Shutdown the endpoint. After this function call its illegal to invoke
+    /// any other methods on the endpoint.
+    /// \a release_fd - a pointer to hold the file descriptor associated with
+    /// the posix endpoint. If it is not nullptr, then the file descriptor is
+    /// not closed while the endpoint is shutdown. The pointer will be updated
+    /// to hold the endpoint's assocaited file descriptor.
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+    /// \a on_release_cb - a callback to be invoked when the endpoint is
+    /// shutdown and the file descriptor is released. It should only be invoked
+    /// if release_fd is not nullptr.
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,130 @@
...
+    virtual void Shutdown(
+        int* release_fd,
+        absl::AnyInvocable<void(absl::Status)> on_release_cb) = 0;
+  };
+  /// Creates a PosixEventEngineEndpoint from an fd which is already assumed to
+  /// be connected to a remote peer.
+  /// \a fd - The connected socket file descriptor.
+  /// \a peer_addr - The address of the peer to which the provided fd has been
+  /// connected.
+  /// \a config - Additional configuration to apply to the endpoint.
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

include/grpc/event_engine/posix.h
Vignesh2208: Added the description now.

include/grpc/event_engine/posix.h
Vignesh2208: Fixed.

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,158 @@
...
+    /// any other methods on the endpoint.
+    /// \a release_fd - a pointer to hold the file descriptor associated with
+    /// the posix endpoint. If it is not nullptr, then the file descriptor is
+    /// not closed while the endpoint is shutdown. The pointer will be updated
+    /// to hold the endpoint's associated file descriptor.
+    /// \a on_release_cb - a callback to be invoked when the endpoint is
+    /// shutdown and the file descriptor is released. It should be invoked only
+    /// if release_fd is not nullptr.
+    virtual void Shutdown(
+        int* release_fd,
```
markdroth, line 54: Why not just pass the fd to `on_release_fd()`, so that the caller doesn't have access to the fd until it's actually safe to use?

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,158 @@
...
+    using OnPosixBindNewFdCallback =
+        absl::AnyInvocable<void(absl::StatusOr<int> listener_fd)>;
+
+    /// Bind an address/port to this Listener.
+    ///
+    /// It is expected that multiple addresses/ports can be bound to this
+    /// Listener before Listener::Start has been called. Returns either the
+    /// bound port or an appropriate error status.
+    /// \a addr - The address to listen for incoming connections.
+    /// \a on_bind_new_fd The callback is invoked once for each newly bound
```
markdroth, line 108: In what case will there be more than one fd here?  We are binding to a specific address, so wouldn't that be just one fd?

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,158 @@
...
+    /// listener fd that may be created by this Bind operation.
+    virtual absl::StatusOr<int> Bind(
```
markdroth, line 110: What is the use-case for this?  Do we actually have cases where we need to use the EE impl to create a listen socket and then take the fd for that socket?  If so, why do we need the EE impl for this -- can't the caller just do the listen directly?

include/grpc/event_engine/posix.h
Vignesh2208: When binding to a Wildcard address, we try to bind (create a socket) to all available interfaces. In that case more than one socket would be created - one per interface.

include/grpc/event_engine/posix.h
Vignesh2208: I changed the on_release_fd callback now to take an absl::StatusOr<int> argument (representing the released fd).

include/grpc/event_engine/posix.h
```diff
@@ -0,0 +1,158 @@
...
+#include <grpc/event_engine/port.h>
+#include <grpc/event_engine/slice_buffer.h>
+
+namespace grpc_event_engine {
+namespace experimental {
+
+#ifdef GRPC_EVENT_ENGINE_POSIX
+
+/// This defines an EventEngine interface that all posix specific event engines
+/// must implement.
```
drfloob, line 37: I would say `may implement`. It's not an explicit requirement for all engines, but is required to support certain functionality. All of our posix-friendly engines will likely implement it, but this is a public API.

Vignesh2208: fixed

ctiller:
Any chance we can make this private before landing? -- I'm looking at the internal use cases we have for it, and I don't think we can get rid of these abstractions for a while (year, maybe two depending on who can work on it) -- but I think we could get rid of them eventually and if so that's desirable.

drfloob:
> Any chance we can make this private before landing? 

@ctiller How do you propose we support public EventEngine implementations participating in posix-specific behavior? We have public APIs to support, `grpc_channel_create_from_fd` and whatnot.

ctiller:
I think we should add `grpc_channel_create_from_endpoint`, and make `grpc_channel_create_from_fd` a thin wrapper around that.

New EventEngine implementations will not get to participate in `grpc_channel_create_from_fd` but they will not need to either (as `grpc_channel_create_from_endpoint` is equivalent).

drfloob:
After some discussion, I agree with the approach, and the suggestion to move the headers under `src/` for now.

Vignesh2208:
> After some discussion, I agree with the approach, and the suggestion to move the headers under `src/` for now.

Sounds good. I moved it to src/core/lib/event_engine/posix.h

src/core/lib/event_engine/posix.h
```diff
@@ -0,0 +1,159 @@
...
+namespace grpc_event_engine {
+namespace experimental {
+
+#ifdef GRPC_EVENT_ENGINE_POSIX
+
+/// This defines an EventEngine interface that posix specific event engines
+/// may implement to support additional functionality.
+class PosixEngine : public EventEngine {
+ public:
+  class PosixEventEngineEndpoint : public EventEngine::Endpoint {
```
drfloob, line 40: The names are going to be confusing, for instance:

```
class PosixEventEngine : public PosixEngine {
  class Endpoint : public PosixEngine::PosixEventEngineEndpoint {
    ...
  }
}
```

Using these with an interface like `void DoAThing(PosixEngine::PosixEventEngineEndpoint* endpoint) { ...  }`, you're probably going to have to look it up to determine if it the argument is taking the interface or the implementation (`PosixEventEngine::Endpoint` is the implementation).

I'd like to suggest an attribute-based alternative for these interfaces, not extending the base types, including strictly the added functionality. For instance:

```
// an Endpoint interface
class EndpointFdSupport {
  virtual int GetWrappedFd() = 0;
  /* ... */
};

class ListenerFdSupport {
  virtual absl::StatusOr<int> BindWithFd(
      const ResolvedAddress& addr,
      OnPosixBindNewFdCallback on_bind_new_fd) = 0;  
  /* ... */
};

class EventEngineFdSupport {
  virtual std::unique_ptr<PosixEventEngineEndpoint> CreatePosixEndpointFromFd(
      int fd, const EndpointConfig& config,
      MemoryAllocator memory_allocator) = 0;  
  /* ... */
}
```

What we'll have then is something like:

```
class PosixEventEngine : public EventEngine, public EventEngineFdSupport {
  class Endpoint : public EndpointFdSupport {};
};
```

And usage would look like `void DoAThing(EndpointFdSupport* endpoint) { ... }`

drfloob: Maybe `EndpointWithFdSupport` describes the object a bit better.

@ctiller does this mesh with your ideas here as well?

src/core/lib/event_engine/posix.h
```diff
@@ -0,0 +1,156 @@
...
+#include <grpc/event_engine/memory_allocator.h>
+#include <grpc/event_engine/port.h>
+#include <grpc/event_engine/slice_buffer.h>
+
+namespace grpc_event_engine {
+namespace experimental {
+
+/// This defines an interface that posix specific event engines endpoints
+/// may implement to support additional file descriptor related functionality.
+class PosixEndpointFdSupport {
```
drfloob, line 36: nomenclature nit: Suggest `PosixEndpointWithFdSupport`, throughout the PR. Reasoning: `void DoSomething(Type* t);` if I have a function that takes thing, and I ask "what kind of thing is it?" - "It is a Posix Endpint Fd support" vs "It is a Posix Endpoint with Fd support". The latter is more natural.

drfloob (commented):
This LGTM, but I'd like Craig's opinion before we solidify it.

src/core/lib/event_engine/posix.h
Vignesh2208: Fixed. Had to make an update though: I changed PosixEndpointWithFdSupport, PosixListenerWithFdSupport and PosixEventEngineWithFdSupport to inherit from EventEngine::Endpoint, EventEngine::Listener and EventEngine respectively.

The reason is, for e.g if we reinterpret_cast and EventEngine::Endpoint* type to PosixEndpointWithFdSupport*, the reinterpret cast doesnt appear to work correctly and was leading to run time issues.

ctiller (approved):

drfloob (approved):

PR #31819, opened 2022-12-06
xds_override_host LB: add basic support for overriding host

eugeneo:
This PR does not include the logic to take the subchannel state into account. That will come in the next PRs.


src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -295,14 +342,52 @@ OrphanablePtr<LoadBalancingPolicy> XdsOverrideHostLb::CreateChildPolicyLocked(
...
 //
 
 RefCountedPtr<SubchannelInterface> XdsOverrideHostLb::Helper::CreateSubchannel(
     ServerAddress address, const ChannelArgs& args) {
-  return xds_override_host_policy_->channel_control_helper()->CreateSubchannel(
-      address, args);
+  auto subchannel =
+      xds_override_host_policy_->channel_control_helper()->CreateSubchannel(
+          address, args);
+  xds_override_host_policy_->RegisterSubchannel(address, subchannel);
```
markdroth, line 389: This handles adding a subchannel to the map, but it doesn't look like we're handling *removing* a subchannel from the map anywhere.  If the child policy unrefs a subchannel, we should not continue to hold a ref to it in the map, because that could potentially hold open a connection that is not actually needed.

I don't think we can do this if we hold a ref to the subchannel in our map, because then we have no way to know when the child policy has unreffed the subchannel, since its ref-count will never go to zero.  Instead, I think we need a structure here similar to that of the `outlier_detection` policy:

- The `AddressMapEntry` struct should become a ref-counted class.  (This is analogous to the `SubchannelState` class in the `outlier_detection` policy.)
- When a subchannel is created, we wrap it in our own subchannel object, which will hold a ref to the relevant `AddressMapEntry`.
- The `AddressMapEntry::subchannel` field should should be a raw pointer instead of a `RefCountedPtr<>` -- it should not hold ownership of the subchannel, since that will prevent the subchannel from being destroyed by the child policy.  Instead, when the subchannel wrapper is unreffed by the child policy, its dtor can check whether `AddressMapEntry::subchannel` is set to point to `this`, and if so, it will reset it to null.

I think this approach also points to the right way to track the subchannel connectivity state: the subchannel wrapper can intercept the watcher started by the child policy, and it can update the state in the `AddressMapEntry` before passing it on to the child policy.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -421,6 +421,11 @@ class LoadBalancingPolicyTest : public ::testing::Test {
     return address;
   }
 
+  static std::unique_ptr<LoadBalancingPolicy::MetadataInterface> MakeMetadata(
```
markdroth, line 424: Suggest moving this down to line 562, so that it's next to the picker code.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -556,9 +561,10 @@ class LoadBalancingPolicyTest : public ::testing::Test {
 
   // Does a pick and returns the result.
   LoadBalancingPolicy::PickResult DoPick(
-      LoadBalancingPolicy::SubchannelPicker* picker) {
+      LoadBalancingPolicy::SubchannelPicker* picker,
+      const std::map<std::string, std::string>& metadata_map = {}) {
     ExecCtx exec_ctx;
-    FakeMetadata metadata({});
+    FakeMetadata metadata{metadata_map};
```
markdroth, line 567: Please use parens instead of braces.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -74,6 +80,76 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  const std::array<absl::string_view, 2> kAddresses = {"ipv4:127.0.0.1:441",
+                                                       "ipv4:127.0.0.1:442"};
+  EXPECT_EQ(policy_->name(), "xds_override_host_experimental");
+  // 1. We use pick_first as a child
+  EXPECT_EQ(ApplyUpdate(BuildUpdate(kAddresses,
+                                    MakeXdsOverrideHostConfig("round_robin")),
+                        policy_.get()),
+            absl::OkStatus());
+  ExpectConnectingUpdate();
+  EXPECT_NE(ExpectState(GRPC_CHANNEL_CONNECTING), nullptr);
```
markdroth, line 93: Reminders:
- Do not explicitly check that we get a certain number of repeat updates for the same connectivity state, because that's ultimately an implementation detail.  After `ExpectConnectingUpdate()`, we should never call `ExpectState(GRPC_CHANNEL_CONNECTING)`, unless there was some other state in between.  In general, the thing after `ExpectConnectingUpdate()` should be either `WaitForConnected()` or `WaitForConnectionFailed()`.
- Whenever we get a connectivity state update, we need to check the picker's behavior instead of just checking that the picker is non-null.  This is ultimately the thing that actually defines how the LB policy works, so if we're not checking it, then we have no guarantee that the LB policy is actually doing the right thing.

Please fix throughout.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -74,6 +80,76 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
       absl::InvalidArgumentError("Missing policy config"));
 }
 
+TEST_F(XdsOverrideHostTest, OverrideHost) {
+  const std::array<absl::string_view, 2> kAddresses = {"ipv4:127.0.0.1:441",
+                                                       "ipv4:127.0.0.1:442"};
+  EXPECT_EQ(policy_->name(), "xds_override_host_experimental");
+  // 1. We use pick_first as a child
```
markdroth, line 87: This comment says pick_first, but the code below actually configures round_robin.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -74,6 +80,76 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  pick_arg[std::string(kOverrideHostHeaderName)] = std::string("127.0.0.1:441");
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[0]);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[0]);
+}
+
+TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
+  const std::array<absl::string_view, 2> kAddresses = {"ipv4:127.0.0.1:441",
+                                                       "ipv4:127.0.0.1:442"};
+  EXPECT_EQ(policy_->name(), "xds_override_host_experimental");
+  // 1. We use pick_first as a child
```
markdroth, line 122: This comment says pick_first, but the code below is using round_robin.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
eugeneo: Done.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -160,23 +180,37 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
...
-            xds_override_host_lb, this);
+            xds_override_host_lb.get(), this);
   }
 }
 
 LoadBalancingPolicy::PickResult XdsOverrideHostLb::Picker::Pick(
     LoadBalancingPolicy::PickArgs args) {
+  std::string buffer;
+  auto override_host =
+      args.initial_metadata->Lookup(kOverrideHostHeaderName, &buffer);
```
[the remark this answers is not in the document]
eugeneo: Done.

eugeneo (commented):
Done. Please take another look.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: I think it's important to keep in mind that the whole goal of the LB policy is to return the right picks.  All of the subchannel connectivity management that we're testing here **exists only for the purpose of ensuring that the policy returns the right picks**.  Given that, I think it's very important to always verify exact picker behavior, because that's ultimately the thing we actually care about.  If that's wrong, then the test is useless, because it's not verifying that the LB policy is behaving correctly.

As long as we're relying on the picking behavior of simple underlying policies like pick_first and round_robin, I don't think it's going to add any particular maintenance burden to the tests to check the pick results, because the picking behavior of those two policies is not going to change.  And if we're already going to assert that the picker returns complete picks, I don't think we gain anything by not actually checking that the resulting address is the one we expect.

Please change all occurrances of `EXPECT_TRUE(ExpectPickComplete(picker.get()).has_value())` to instead do `EXPECT_EQ(ExpectPickComplete(picker.get()), "ipv4:127.0.0.1:441")` (or whatever address is appropriate).

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -100,33 +95,34 @@ TEST_F(XdsOverrideHostTest, OverrideHost) {
...
+  ASSERT_NE(picker, nullptr);
+  EXPECT_TRUE(ExpectPickComplete(picker.get()).has_value());
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
   // Make sure child policy works
   EXPECT_NE(ExpectPickComplete(picker.get()), ExpectPickComplete(picker.get()));
   // Check that the host is overridden
-  std::map<std::string, std::string> pick_arg{
-      {std::string(kOverrideHostHeaderName), "127.0.0.1:442"}};
+  std::map<UniqueTypeName, std::string> pick_arg{
```
markdroth, line 109: Please call this `call_attributes`.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -100,33 +95,34 @@ TEST_F(XdsOverrideHostTest, OverrideHost) {
...
+      {XdsHostOverrideTypeName(), "127.0.0.1:442"}};
   EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
   EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
-  pick_arg[std::string(kOverrideHostHeaderName)] = std::string("127.0.0.1:441");
+  pick_arg[XdsHostOverrideTypeName()] = std::string("127.0.0.1:441");
   EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[0]);
   EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[0]);
 }
 
 TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
```
markdroth, line 118: s/Channel/Subchannel/

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+  EXPECT_TRUE(ExpectPickComplete(picker.get()).has_value());
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
   // Make sure child policy works
   EXPECT_NE(ExpectPickComplete(picker.get()), ExpectPickComplete(picker.get()));
   // Check that the host is overridden
-  std::map<std::string, std::string> pick_arg{
-      {std::string(kOverrideHostHeaderName), "no such host"}};
-
+  std::map<UniqueTypeName, std::string> pick_arg{
```
markdroth, line 145: Please call this `call_attributes`.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+      {XdsHostOverrideTypeName(), "no such host"}};
   std::unordered_set<absl::optional<std::string>> picks{
       ExpectPickComplete(picker.get(), pick_arg),
       ExpectPickComplete(picker.get(), pick_arg)};
-
   ASSERT_THAT(picks, UnorderedElementsAre(kAddresses[0], kAddresses[1]));
```
markdroth, line 150: I think this can say `UnorderedElementsAreArray(kAddresses)`.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+TEST_F(XdsOverrideHostTest, DISABLED_SubchannelsComeAndGo) {
+  const std::array<absl::string_view, 3> kAddresses = {
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
+  EXPECT_EQ(ApplyUpdate(BuildUpdate(kAddresses,
+                                    MakeXdsOverrideHostConfig("round_robin")),
+                        policy_.get()),
+            absl::OkStatus());
+  ExpectConnectingUpdate();
+  RefCountedPtr<LoadBalancingPolicy::SubchannelPicker> picker;
+  // Ready up both subchannels
```
markdroth, line 162: s/both/all/

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+  for (auto address : kAddresses) {
+    auto subchannel = FindSubchannel({address});
+    ASSERT_NE(subchannel, nullptr);
+    ASSERT_TRUE(subchannel->ConnectionRequested());
+    subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+    subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+    picker = WaitForConnected();
```
markdroth, line 169: I don't think it makes sense to check this inside the loop.  Let's check that we get connected just once, after we exit the loop.  After that, each subsequent update should be another READY update with a new picker.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+  }
+  picker = WaitForConnected();
+  ASSERT_NE(picker, nullptr);
+  EXPECT_TRUE(ExpectPickComplete(picker.get()).has_value());
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
+  // Make sure child policy works
+  EXPECT_NE(ExpectPickComplete(picker.get()), ExpectPickComplete(picker.get()));
+  // Check that the host is overridden
+  std::map<UniqueTypeName, std::string> pick_arg{
```
markdroth, line 181: Please call this `call_attributes`.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+      {XdsHostOverrideTypeName(), "127.0.0.1:442"}};
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  // Some other address is gone
+  EXPECT_EQ(
+      ApplyUpdate(BuildUpdate({"ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442"},
+                              MakeXdsOverrideHostConfig("round_robin")),
+                  policy_.get()),
+      absl::OkStatus());
+  WaitForConnected();
```
markdroth, line 191: I don't think we should accept CONNECTING updates here.  We should expect only READY updates.

Also, please always check the behavior of the resulting picker.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+  ASSERT_NE(picker, nullptr);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  // "Our" address is gone
+  EXPECT_EQ(
+      ApplyUpdate(BuildUpdate({"ipv4:127.0.0.1:441", "ipv4:127.0.0.1:443"},
+                              MakeXdsOverrideHostConfig("round_robin")),
+                  policy_.get()),
+      absl::OkStatus());
+  WaitForConnected();
```
markdroth, line 202: Same comment as on line 191.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+      ExpectPickComplete(picker.get(), pick_arg)};
+  EXPECT_THAT(picks,
+              UnorderedElementsAre("ipv4:127.0.0.1:441", "ipv4:127.0.0.1:443"));
+  // And now it is back
+  EXPECT_EQ(
+      ApplyUpdate(BuildUpdate({"ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"},
+                              MakeXdsOverrideHostConfig("round_robin")),
+                  policy_.get()),
+      absl::OkStatus());
+  WaitForConnected();
```
markdroth, line 216: Same comment as on line 191.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -135,21 +131,96 @@ TEST_F(XdsOverrideHostTest, OverrideHostChannelNotFound) {
...
+  std::map<UniqueTypeName, std::string> pick_arg{
+      {XdsHostOverrideTypeName(), "no such host"}};
   std::unordered_set<absl::optional<std::string>> picks{
       ExpectPickComplete(picker.get(), pick_arg),
       ExpectPickComplete(picker.get(), pick_arg)};
-
   ASSERT_THAT(picks, UnorderedElementsAre(kAddresses[0], kAddresses[1]));
 }
+
+TEST_F(XdsOverrideHostTest, DISABLED_SubchannelsComeAndGo) {
```
markdroth, line 153: Why is this test disabled?

markdroth (commented):

Most comments are fairly minor.  The only really significant one is the one about the mutex structure.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Thanks for pointing it out. I disabled it when I was chasing an ASAN issue.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done! Thanks.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

eugeneo (commented):
Thank you for the reviews. Comments were not minor 😉

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -295,14 +342,52 @@ OrphanablePtr<LoadBalancingPolicy> XdsOverrideHostLb::CreateChildPolicyLocked(
...
+  if (subchannel_record == subchannel_by_address_map_.end()) {
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_xds_override_host_trace)) {
+      gpr_log(
+          GPR_INFO,
+          "[xds_override_host_lb %p] Subchannel for address %s was not found",
+          this, key.c_str());
+    }
+    return nullptr;
+  }
+  if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_xds_override_host_trace)) {
```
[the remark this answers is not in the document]
eugeneo: Removed the trace.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -68,11 +68,109 @@ namespace {
...
+        maybe_entry = it->second;
+      }
+    }
+    return MakeRefCounted<SubchannelWrapper>(std::move(subchannel),
+                                             maybe_entry);
+  }
+
+ private:
+  absl::Mutex mu_;
+  std::map<std::string, RefCountedPtr<SubchannelEntry>, std::less<>>
```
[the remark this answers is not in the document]
markdroth: Interesting!  I didn't realize this was needed.

I guess the distinction is that the default comparator for `std::map<>` is `std::less<K>`, where `K` is the key type (i.e., in this case, `std::less<std::string>`), and I guess `std::less<>` and `std::less<std::string>` don't behave exactly the same way.

TIL.

markdroth (commented):
This is getting closer!

I have only two substantive structural comments.  One is about the map lookups vs. ref-counting for the map values.  The other is about the test structure still needing work.  I think all of the other comments are fairly minor.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -95,27 +95,26 @@ TEST_F(XdsOverrideHostTest, OverrideHost) {
...
     subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
   }
-  auto picker = WaitForConnected();
-  ASSERT_NE(picker, nullptr);
-  EXPECT_TRUE(ExpectPickComplete(picker.get()).has_value());
-  picker = ExpectState(GRPC_CHANNEL_READY);
-  ASSERT_NE(picker, nullptr);
-  EXPECT_TRUE(ExpectPickComplete(picker.get()).has_value());
-  picker = ExpectState(GRPC_CHANNEL_READY);
+  WaitForConnected();
```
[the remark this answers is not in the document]
eugeneo: Rewrote, using the new utilities. PTAL.

eugeneo (commented):
Thank you for the review. I believe I addressed all the comments. Please take another look.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -41,36 +41,77 @@ class XdsOverrideHostTest : public LoadBalancingPolicyTest {
          Json::Object{{"childPolicy", Json::Array{{child_policy_config}}}}}}});
   }
 
+  absl::optional<RefCountedPtr<LoadBalancingPolicy::SubchannelPicker>>
+  InitRoundRobin(absl::Span<const absl::string_view> addresses) {
```
markdroth, line 45: Suggest calling this something like `ExpectStartupWithRoundRobin()`.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -41,36 +41,77 @@ class XdsOverrideHostTest : public LoadBalancingPolicyTest {
...
 
 TEST_F(XdsOverrideHostTest, DelegatesToChild) {
-  ASSERT_NE(policy_, nullptr);
-  const std::array<absl::string_view, 2> kAddresses = {"ipv4:127.0.0.1:441",
-                                                       "ipv4:127.0.0.1:442"};
-  EXPECT_EQ(policy_->name(), "xds_override_host_experimental");
+  // Send address list to LB policy.
+  const std::array<absl::string_view, 3> kAddresses = {
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
   EXPECT_EQ(ApplyUpdate(BuildUpdate(kAddresses, MakeXdsOverrideHostConfig()),
```
markdroth: Might as well replace most of the body of this test with calling `InitRoundRobin()`, since they seem to be identical.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -81,30 +122,13 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
-  WaitForConnected();
-  ExpectState(GRPC_CHANNEL_READY);
-  auto picker = ExpectState(GRPC_CHANNEL_READY);
+  // Send address list to LB policy.
+  const std::array<absl::string_view, 3> kAddresses = {
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
+  auto maybe_picker = InitRoundRobin(kAddresses);
+  ASSERT_TRUE(maybe_picker.has_value());
+  auto picker = *maybe_picker;
   ASSERT_NE(picker, nullptr);
```
markdroth, line 131: This assertion should probably move into `WaitForRoundRobinListChange()`, because we're already calling the picker in that function, so if it's null, we'd crash there.

I think we should add the following in lb_policy_test.h at line 575:

```
EXPECT_NE(update.picker, nullptr);
if (update.picker == nullptr) return false;
```

Once we do that, we no longer need the assertion here, because `WaitForRoundRobinListChange()` will guarantee that if it returns non-nullopt, the picker will never be null.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -81,30 +122,13 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
-  }
-  WaitForConnected();
-  ExpectState(GRPC_CHANNEL_READY);
-  auto picker = ExpectState(GRPC_CHANNEL_READY);
+  // Send address list to LB policy.
+  const std::array<absl::string_view, 3> kAddresses = {
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
+  auto maybe_picker = InitRoundRobin(kAddresses);
+  ASSERT_TRUE(maybe_picker.has_value());
+  auto picker = *maybe_picker;
```
markdroth, line 130: Instead of moving this into a different variable, you can just access it directly via the `absl::optional<>`.  Suggest writing it like this:

```
auto picker = InitRoundRobin(kAddresses);
ASSERT_TRUE(picker.has_value());
```

Then just change the code below to use `picker->get()` instead of `picker.get()`.

Same thing in all tests.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -81,30 +122,13 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
 }
 
 TEST_F(XdsOverrideHostTest, OverrideHost) {
```
markdroth, line 124: I love how concise these two tests are now!  Very readable!

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -116,58 +140,30 @@ TEST_F(XdsOverrideHostTest, OverrideHost) {
...
   ASSERT_NE(picker, nullptr);
-  // Make sure child policy works
-  EXPECT_NE(ExpectPickComplete(picker.get()), ExpectPickComplete(picker.get()));
-  // Check that the host is overridden
   std::map<UniqueTypeName, std::string> call_attributes{
       {XdsHostOverrideTypeName(), "no such host"}};
-  std::unordered_set<absl::optional<std::string>> picks{
-      ExpectPickComplete(picker.get(), call_attributes),
-      ExpectPickComplete(picker.get(), call_attributes)};
+  std::unordered_set<std::string> picks;
```
markdroth, line 152: Instead of doing this by hand, I suggest adding a `call_attributes` parameter to `ExpectRoundRobinPicks()` and then just use that method here.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -181,8 +233,23 @@ LoadBalancingPolicy::PickResult XdsOverrideHostLb::Picker::Pick(
     return PickResult::Fail(absl::InternalError(
```
markdroth, line 233: Still need to move this check (lines 232-235) down to line 245, so that we're not checking that the child picker is non-null unless we're actually going to use it.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+
+TEST_F(XdsOverrideHostTest, SubchannelsComeAndGo) {
+  const std::array<absl::string_view, 3> kAddresses = {
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
+  auto maybe_picker = InitRoundRobin(kAddresses);
+  ASSERT_TRUE(maybe_picker.has_value());
+  auto picker = *maybe_picker;
+  ASSERT_NE(picker, nullptr);
+  // Make sure child policy works
+  EXPECT_NE(ExpectPickComplete(picker.get()), ExpectPickComplete(picker.get()));
```
markdroth, line 169: Is this expectation actually passing?  It seems like it shouldn't be, because we're using round_robin here, so there should not be two successive picks for the same address.

In any case, I don't think this check is actually needed, since `InitRoundRobin()` is already checking the initial picker behavior for us.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  // Send address list to LB policy.
+  const std::array<absl::string_view, 3> kAddresses = {
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
+  auto maybe_picker = InitRoundRobin(kAddresses);
+  ASSERT_TRUE(maybe_picker.has_value());
+  auto picker = *maybe_picker;
+  ASSERT_NE(picker, nullptr);
+  // Check that the host is overridden
+  std::map<UniqueTypeName, std::string> call_attributes{
+      {XdsHostOverrideTypeName(), "127.0.0.1:442"}};
```
markdroth, line 134: Suggest using `kAddresses[1]` here instead of duplicating the string.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  EXPECT_EQ(ExpectPickComplete(picker.get(), call_attributes), kAddresses[1]);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), call_attributes), kAddresses[1]);
+  call_attributes[XdsHostOverrideTypeName()] = std::string("127.0.0.1:441");
```
markdroth, line 137: Similarly, suggest using `kAddresses[0]` here.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
+  auto maybe_picker = InitRoundRobin(kAddresses);
+  ASSERT_TRUE(maybe_picker.has_value());
+  auto picker = *maybe_picker;
+  ASSERT_NE(picker, nullptr);
+  // Make sure child policy works
+  EXPECT_NE(ExpectPickComplete(picker.get()), ExpectPickComplete(picker.get()));
+  // Check that the host is overridden
+  std::map<UniqueTypeName, std::string> call_attributes{
+      {XdsHostOverrideTypeName(), "127.0.0.1:442"}};
```
markdroth, line 172: Suggest using `kAddresses[1]` here.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  EXPECT_EQ(ExpectPickComplete(picker.get(), call_attributes), kAddresses[1]);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), call_attributes), kAddresses[1]);
+  // Some other address is gone
+  EXPECT_EQ(
+      ApplyUpdate(BuildUpdate({"ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442"},
+                              MakeXdsOverrideHostConfig("round_robin")),
+                  policy_.get()),
+      absl::OkStatus());
+  ExpectState(GRPC_CHANNEL_READY);
```
markdroth, line 181: This is still (a) checking for individual state changes instead of checking only when we actually expect a behavior change and (b) getting a state change and not checking the resulting picker behavior.

I think we should instead use `WaitForRoundRobinListChange()` here to wait for the LB policy to switch to the new address list.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  ASSERT_NE(picker, nullptr);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), call_attributes), kAddresses[1]);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), call_attributes), kAddresses[1]);
+  // "Our" address is gone
+  EXPECT_EQ(
+      ApplyUpdate(BuildUpdate({"ipv4:127.0.0.1:441", "ipv4:127.0.0.1:443"},
+                              MakeXdsOverrideHostConfig("round_robin")),
+                  policy_.get()),
+      absl::OkStatus());
+  ExpectState(GRPC_CHANNEL_READY);
```
markdroth, line 192: Same here: Please use `WaitForRoundRobinListChange()` to wait for the policy to switch to the new address list.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ExpectQueueEmpty();
+  ASSERT_NE(picker, nullptr);
+  std::array<absl::optional<std::string>, 2> picks = {
```
markdroth, line 196: Similar comment as on line 152 above: This should just call `ExpectRoundRobinPicks()` with a `call_attributes` parameter.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+      ExpectPickComplete(picker.get(), call_attributes)};
+  EXPECT_THAT(picks, UnorderedElementsAreArray(
+                         {"ipv4:127.0.0.1:441", "ipv4:127.0.0.1:443"}));
+  // And now it is back
+  EXPECT_EQ(
+      ApplyUpdate(BuildUpdate({"ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"},
+                              MakeXdsOverrideHostConfig("round_robin")),
+                  policy_.get()),
+      absl::OkStatus());
+  ExpectState(GRPC_CHANNEL_READY);
```
markdroth, line 207: Same as above: Please use `WaitForRoundRobinListChange()` to wait for the policy to switch to the new address list.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -86,6 +121,97 @@ TEST_F(XdsOverrideHostTest, NoConfigReportsError) {
...
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
+  picks = {ExpectPickComplete(picker.get(), call_attributes),
```
markdroth, line 210: Same as above: Please call `ExpectRoundRobinPicks()` with a `call_attributes` parameter.

markdroth (commented):
This is looking really good!  All of my remaining comments are fairly minor; I have no more structural concerns.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Not sure what you mean. Have not found a function named `InitRoundRobin()`

eugeneo (commented):
Comments addressed. Please take another look.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -658,11 +659,16 @@ class LoadBalancingPolicyTest : public ::testing::Test {
   // Gets num_picks complete picks from picker and returns the resulting
   // list of addresses, or nullopt if a non-complete pick was returned.
   absl::optional<std::vector<std::string>> GetCompletePicks(
-      LoadBalancingPolicy::SubchannelPicker* picker, size_t num_picks,
-      SourceLocation location = SourceLocation()) {
+      LoadBalancingPolicy::SubchannelPicker* picker,
+      const std::map<UniqueTypeName, std::string> call_attributes = {},
+      size_t num_picks = 3, SourceLocation location = SourceLocation()) {
```
markdroth, line 664: I think `num_picks` should not have a default value here, because there isn't really a default number of picks that will be appropriate in most circumstances.  The caller should be required to pass this parameter.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: Why do you need `ExpectState(GRPC_CHANNEL_READY)` here?  I don't think that should be necessary.  What error do you get if you remove that line?

I don't think tests should have to directly call `ExpectState()` directly in the vast majority of cases.  And if there is some edge case in which a test does need to do that, it cannot ignore the returned picker -- tests must *always* verify the behavior of the picker.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: I was referring to the function that we renamed to `ExpectStartupWithRoundRobin()`.  What I'm saying is that the body of the `DelegatesToChild` test is basically duplicating the functionality of `ExpectStartupWithRoundRobin()`, so we can basically replace most of the body of the test with a call to that function.

markdroth (commented):
This is very close!  Just a few small issues left.

Also, looks like you need to merge master into this branch to fix some conflicts.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
markdroth: Instead of dropping the default for `call_attributes`, just reorder the parameters so that `num_picks` is before `call_attributes`.

I've done this in https://github.com/eugeneo/grpc/pull/1.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: It looks like there were two problems here.

1. There was a bug in RR causing it to generate spurious picker updates whenever it swapped out the subchannel list.  I've fixed this in #31939.
2. The purpose of calling `WaitForRoundRobinListChange()` here is to ensure that the LB policy has actually swapped over to the new address list.  You can't check that by using the call attributes here, because then you're forcing the picker to return a specific address instead of checking that it's returning the new list of addresses that we're looking for.  Instead, you need to first call `WaitForRoundRobinListChange()` to check for the new list without the call attributes, and then use the call attributes to make sure they're working as expected.  I've made this change in https://github.com/eugeneo/grpc/pull/1.

eugeneo (commented):
@markdroth thanks for the pull request! I merged it into this one. Please take another look.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -559,31 +559,31 @@ class LoadBalancingPolicyTest : public ::testing::Test {
...
+  RefCountedPtr<LoadBalancingPolicy::SubchannelPicker>
   WaitForRoundRobinListChange(
       absl::Span<const absl::string_view> old_addresses,
       absl::Span<const absl::string_view> new_addresses,
       const std::map<UniqueTypeName, std::string> call_attributes = {},
       size_t num_iterations = 3, SourceLocation location = SourceLocation()) {
-    absl::optional<RefCountedPtr<LoadBalancingPolicy::SubchannelPicker>> retval;
+    gpr_log(GPR_INFO, "Waiting for expected RR addresses...");
+    RefCountedPtr<LoadBalancingPolicy::SubchannelPicker> retval;
     WaitForStateUpdate(
```
markdroth, line 570:
```suggestion
    size_t num_picks =
        std::max(new_addresses.size(), old_addresses.size()) * num_iterations;
    WaitForStateUpdate(
```

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -559,31 +559,31 @@ class LoadBalancingPolicyTest : public ::testing::Test {
...
         [&](FakeHelper::StateUpdate update) {
           EXPECT_EQ(update.state, GRPC_CHANNEL_READY)
               << location.file() << ":" << location.line();
           if (update.state != GRPC_CHANNEL_READY) return false;
           // Get enough picks to round-robin num_iterations times across all
           // expected addresses.
-          auto picks =
-              GetCompletePicks(update.picker.get(), call_attributes,
-                               new_addresses.size() * num_iterations, location);
+          auto picks = GetCompletePicks(update.picker.get(), num_iterations,
```
markdroth, line 577:
```suggestion
          auto picks = GetCompletePicks(update.picker.get(), num_picks,
```

markdroth (commented):
Looks like you accidentally changed the number of picks in `WaitForRoundRobinListChange()` when you merged in the changes.

eugeneo:
@markdroth thanks. I accepted the suggested changes.

markdroth (approved):
This looks great!

PR #31820, opened 2022-12-06
xDS cluster: use absl::variant for cluster type

markdroth:
There are no functional changes here, this is just a code cleanup.


eugeneo (approved):
So much to learn...

(No comments, lgtm!)

PR #31793, opened 2022-12-02
xDS stateful session affinity: support xDS HTTP filter config

markdroth:
Built on #31788.

PR #31823, opened 2022-12-06
Reland: xDS stateful session affinity: support xDS HTTP filter config

markdroth:
Initial attempt was #31793, reverted in #31821.


drfloob (approved):
LGTM. Note for posterity: pure revert-of-revert, needs internal changes.

PR #31826, opened 2022-12-06
PollingResolver: Honor cancel action in Timer callback

yijiem:
This is learned from AJ's PR #30045. When `EventEngine::Cancel()` failed, it might mean that the timer callback has already been scheduled. Therefore, we reset the timer handle and check that condition in the timer callback which is similar to the original code.

### DIFF
--- a/src/core/ext/filters/client_channel/resolver/polling_resolver.cc
+++ b/src/core/ext/filters/client_channel/resolver/polling_resolver.cc
@@ -121,10 +121,11 @@ void PollingResolver::OnNextResolutionLocked() {
             "[polling resolver %p] re-resolution timer fired: shutdown_=%d",
             this, shutdown_);
   }
-  next_resolution_timer_handle_.reset();
-  if (!shutdown_) {
+  // If we haven't been cancelled nor shutdown, then start resolving.
+  if (next_resolution_timer_handle_.has_value() && !shutdown_) {
     StartResolvingLocked();
   }
+  next_resolution_timer_handle_.reset();
 }
 
 void PollingResolver::MaybeCancelNextResolutionTimer() {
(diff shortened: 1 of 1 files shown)


src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -121,10 +121,11 @@ void PollingResolver::OnNextResolutionLocked() {
             "[polling resolver %p] re-resolution timer fired: shutdown_=%d",
             this, shutdown_);
   }
-  next_resolution_timer_handle_.reset();
-  if (!shutdown_) {
+  // If we haven't been cancelled nor shutdown, then start resolving.
```
markdroth, line 124: I don't think this PR is needed.

There are two places where the timer may be cancelled from:
- `ResetBackoffLocked()`: In this case, we want to abort the backoff delay and immediately re-resolve, so we *do* want to invoke `StartResolvingLocked()`.
- `ShutdownLocked()`: In this case, we do *not* want to invoke `StartResolvingLocked()`, but the existing code is already not doing that, because `shutdown_` is true.

yijiem: In the original `void PollingResolver::OnNextResolutionLocked(grpc_error_handle error)` function before the timer migration in #31717, this check was written like:
```
if (error.ok() && !shutdown_) {
  StartResolvingLocked();
}
```
See https://github.com/grpc/grpc/pull/31717/files#diff-beb9a20cca271b8a75724fca2a6587a6d0c3e2f79d856169b358419a057c6fcdL122.

So the logic was if the timer has not been cancelled and we are not shutting down, invoke `StartResolvingLocked()`. Maybe that is not intended in the first place?

yijiem: Done. Thanks for the great review Mark! Very clear! And it leaves a great record for future reference as well!

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
```diff
@@ -121,10 +124,11 @@ void PollingResolver::OnNextResolutionLocked() {
...
+  if (next_resolution_timer_handle_.has_value() && !shutdown_) {
     StartResolvingLocked();
   }
+  next_resolution_timer_handle_.reset();
```
markdroth, line 131: Suggest moving this up to line 129, since there's no need to reset it if it's already nullopt.

markdroth (approved):

src/core/ext/filters/client_channel/resolver/polling_resolver.cc
yijiem: Done.

PR #31814, opened 2022-12-06
Sends out-of-band close for streams from client side when destroying transport

LittleCVR:
Due to its design, simply destroying a Binder object doesn't really tell the other end to close theirs, thus, streams won't be closed, either. We have to explicitly send a close message to each stream in order to close them.

PR #31833, opened 2022-12-07
Revert "Sends out-of-band close for streams from client side when destroying transport"

ananda1066:
Reverts grpc/grpc#31814. This change causes TSAN/UBSAN flakes for `binder_server_test`, e.g. https://source.cloud.google.com/results/invocations/6d01cf10-dc4f-4406-bc1b-8ddcfed42567/targets. Reproduced in 10 runs and verified that commit prior does not cause this flakiness.


ctiller:
(lgtm clearly)

PR #31836, opened 2022-12-08
Use EventEngine::Run in LoadBalancingPolicy::QueuePicker

drfloob:

### DIFF
--- a/src/core/lib/load_balancing/lb_policy.cc
+++ b/src/core/lib/load_balancing/lb_policy.cc
@@ -18,11 +18,10 @@
 
 #include <grpc/support/port_platform.h>
 
-#include "lb_policy.h"
+#include "src/core/lib/load_balancing/lb_policy.h"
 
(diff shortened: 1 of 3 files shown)


src/core/lib/load_balancing/lb_policy.cc
```diff
@@ -18,12 +18,11 @@
 
 #include <grpc/support/port_platform.h>
 
-#include "src/core/lib/load_balancing/lb_policy.h"
+#include "lb_policy.h"
```
markdroth, line 21: Looks like iwyu did something weird here.  I think we still need the full path to this include, but it seems to have changed the path here and then added the right path as a separate include on line 25.

markdroth (approved):
Other than the iwyu weirdness, this looks great!

src/core/lib/load_balancing/lb_policy.cc
drfloob: I haven't seen that one before. Fixed.

PR #31838, opened 2022-12-08
EventEngine::RunAfter migration SubchannelStreamClient

yijiem:
Note that the `StartRetryTimerLocked` code path is not covered by test. And I'm having trouble adding a test case to cover it. Seems like we need the health service (default_health_check_service) to shutdown and end the server-side streaming RPC (and send the trailing metadata) but keep the subchannel (and subchannel_stream_client) alive so that it could retry a Health.Watch RPC.


src/core/ext/filters/client_channel/subchannel_stream_client.cc
```diff
@@ -94,8 +95,10 @@ void SubchannelStreamClient::Orphan() {
     MutexLock lock(&mu_);
     event_handler_.reset();
     call_state_.reset();
-    if (retry_timer_callback_pending_) {
-      grpc_timer_cancel(&retry_timer_);
+    if (retry_timer_handle_.has_value()) {
+      connected_subchannel_->args().GetObject<EventEngine>()->Cancel(
```
drfloob, line 99: I've seen differing opinions here, so it's your call, but I might prefer to see the engine pointer cached locally on SubchannelStreamClient construction to save the lookup cost.

src/core/ext/filters/client_channel/subchannel_stream_client.cc
```diff
@@ -124,40 +127,39 @@ void SubchannelStreamClient::StartRetryTimerLocked() {
   if (event_handler_ != nullptr) {
     event_handler_->OnRetryTimerStartLocked(this);
   }
-  Timestamp next_try = retry_backoff_.NextAttemptTime();
+  const Timestamp next_try = retry_backoff_.NextAttemptTime();
+  const Duration timeout = next_try - Timestamp::Now();
```
drfloob, line 131: nit: this could be one line, such as `  const Duration timeout = retry_backoff_.NextAttemptTime() - Timestamp::Now();`.

drfloob (commented):
LGTM with a few suggestions. I think we should merge https://github.com/grpc/grpc/pull/31850 before this PR, though.

src/core/ext/filters/client_channel/subchannel_stream_client.cc
yijiem: good point, done.

src/core/ext/filters/client_channel/subchannel_stream_client.cc
yijiem: done.

src/core/ext/filters/client_channel/subchannel_stream_client.cc
```diff
@@ -124,40 +127,37 @@ void SubchannelStreamClient::StartRetryTimerLocked() {
...
     } else {
       gpr_log(GPR_INFO, "%s %p: ... retrying immediately.", tracer_, this);
     }
   }
-  // Ref for callback, tracked manually.
-  Ref(DEBUG_LOCATION, "health_retry_timer").release();
-  retry_timer_callback_pending_ = true;
-  grpc_timer_init(&retry_timer_, next_try, &retry_timer_callback_);
+  retry_timer_handle_ = event_engine_->RunAfter(
+      timeout, [self = Ref(DEBUG_LOCATION, "health_retry_timer")] {
```
markdroth, line 142: I think we need to release this ref before the `ExecCtx` goes out of scope.

markdroth (approved):
Just one minor nit, otherwise looks great!

src/core/ext/filters/client_channel/subchannel_stream_client.cc
yijiem: Great callout! This makes `Unref` called before the `ExecCtx` goes out of scope which matches with the original code when the timer callback gets called.

PR #31841, opened 2022-12-08
fake_handshaker_server: Allow specifying a different peer identity.

pefoley2:
This is helpful when writing integration tests with systems that expect a specific value for the peer.


veblush:
@pefoley2 This looks good to me. Would you elaborate how you use this?

veblush (approved):

PR #31848, opened 2022-12-09
WindowsEventEngine Client implementation

drfloob:
Doing this work inspired a few changes I'd like to make in later PRs.

* A bazel macro for EventEngine test suite definitions, which will create the right bazel libraries/tests/binaries for all of the specialized tests and tools. The two factory definitions are a bit redundant, and should only need to be defined once.
* ResolvedAddress needs a mutable alternative, and/or a redesign as a simpler struct. Some copying could be eliminated.
* The echo client is added here to inspire reviewer confidence, but will probably be landed in a separate PR. The test suite's `ConnectToNonExistentListenerTest` passes, but since there is no Listener implementation yet, the rest do not. The WindowsEventEngine-specialized echo client works.


src/core/lib/event_engine/windows/windows_engine.h
```diff
@@ -104,16 +99,52 @@ class WindowsEventEngine : public EventEngine,
...
+    EventEngine::OnConnectCallback on_connected_user_callback
+        ABSL_GUARDED_BY(mu);
+    EventEngine::Closure* on_connected ABSL_GUARDED_BY(mu);
+    std::unique_ptr<WinSocket> socket ABSL_GUARDED_BY(mu);
+    EventEngine::ResolvedAddress address ABSL_GUARDED_BY(mu);
+    MemoryAllocator allocator ABSL_GUARDED_BY(mu);
+  };
+
+  // A poll worker which schedules itself unless kicked
+  class IOCPWorker : public EventEngine::Closure {
```
Vignesh2208, line 119: Maybe this could be renamed to IOCPWorkClosure ?

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -26,33 +26,66 @@
...
+// ---- IOCPWorker ----
+
+WindowsEventEngine::IOCPWorker::IOCPWorker(Executor* executor, IOCP* iocp)
+    : executor_(executor), iocp_(iocp) {
+  executor_->Run(this);
+}
+
+void WindowsEventEngine::IOCPWorker::Run() {
+  auto result = iocp_->Work(std::chrono::seconds(60), [this] {
+    workers_++;
```
Vignesh2208, line 56: Please use atomic fetch_add and fetch_sub on this.

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -140,15 +178,171 @@ bool WindowsEventEngine::IsWorkerThread() {
...
+         status = GRPC_WSA_ERROR(
+             WSAGetLastError(),
+             "WSAIoctl(SIO_GET_EXTENSION_FUNCTION_POINTER)")]() mutable {
+      on_connect(status);
+    });
+    return invalid_connection_handle;
+  }
+  // bind the local address
+  auto local_address = ResolvedAddressMakeWild6(0);
+  istatus = bind(sock, local_address.address(), local_address.size());
```
Vignesh2208, line 264: why is there a bind to local address ? Could you add a comment

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -140,15 +178,171 @@ bool WindowsEventEngine::IsWorkerThread() {
...
+      Run([on_connect = std::move(on_connect), status]() mutable {
+        on_connect(status);
+      });
+      watched_socket->MaybeShutdown(status);
+      return invalid_connection_handle;
+    }
+  }
+  GPR_ASSERT(watched_socket != nullptr);
+  auto connection_state = std::make_shared<ConnectionState>();
+  grpc_core::MutexLock lock(&connection_state->mu);
```
Vignesh2208, line 293: Is there a need to grab this mutex as soon as connection_state is created ? Can it be moved to line 301 maybe ?

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -140,15 +178,171 @@ bool WindowsEventEngine::IsWorkerThread() {
...
+  if (!known_connection_handles_.contains(handle)) {
+    GRPC_EVENT_ENGINE_TRACE(
+        "Unknown connection handle: %s",
+        HandleToString<EventEngine::ConnectionHandle>(handle).c_str());
+    return false;
+  }
+  auto* connection_state = reinterpret_cast<ConnectionState*>(handle.keys[0]);
+  connection_state->socket->MaybeShutdown(
+      absl::CancelledError("CancelConnect"));
+  known_connection_handles_.erase(handle);
```
Vignesh2208, line 341: I think if CancelConnect succeeds, the timer started in line 309 also needs to be cancelled. Otherwise the timer callback will execute after deadline expires and it may not delete connection_state->on_connected.

test/core/event_engine/test_suite/tools/echo_client.cc
```diff
@@ -0,0 +1,143 @@
...
+  SliceBuffer buf;
+  grpc_core::Notification read_done;
+  endpoint->Read(
+      [&](absl::Status status) {
+        if (!status.ok()) {
+          gpr_log(GPR_ERROR, "Error reading from endpoint: %s",
+                  status.ToString().c_str());
+          exit(1);
+        }
+        Slice received = buf.TakeFirst();
```
Vignesh2208, line 89: Its not always guaranteed that the first slice read will contain the whole message appended in line 67. The ReceiveAndEchoMessage function should perhaps take an expected number of bytes to read as an argument and recursively call endpoint->Read until the expected number of bytes have been read

Vignesh2208 (changes requested):
Left few comments. Thanks for working on this!

src/core/lib/event_engine/windows/windows_engine.cc
drfloob: Is there a style guide requirement here? Or are you recommending a different memory order? Atomic int's overload of `operator++` does an atomic fetch_add using default `std::memory_order_seq_cst`, and I prefer the less verbose syntax when possible. https://en.cppreference.com/w/cpp/atomic/atomic/operator_arith

src/core/lib/event_engine/windows/windows_engine.cc
drfloob: `ConnectionState`'s thread guard annotations would be unhappy with that change. We could move some things around, create a constructor for the type instead (which would only partially construct the object), or reduce the set of members protected by the lock, but would it help anything? This lock only affects this one connection, which is going to take just as long whether we lock before or after setting those member variables.

src/core/lib/event_engine/windows/windows_engine.cc
drfloob: https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-bind ... this would be the equivalent to adding a link to `man 2 bind` around the posix implementation's use of bind, it's a bit overkill.

> The bind function is required before the use of the [send](https://learn.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-send) or [WSASend](https://learn.microsoft.com/en-us/windows/desktop/api/winsock2/nf-winsock2-wsasend) functions which do not perform an implicit bind and are allowed only on connected sockets, which means the socket must have already been bound for it to be connected.

test/core/event_engine/test_suite/tools/echo_client.cc
drfloob: Understood. I'll do a fix before we release it in a separate PR, I just pulled the first slice here for testing purposes.

src/core/lib/event_engine/windows/windows_engine.h
drfloob: Done

src/core/lib/event_engine/windows/windows_engine.cc
drfloob: Good catch, fixed.

src/core/lib/event_engine/windows/windows_engine.cc
ctiller: I have a strong preference to not using `operator++` and the like with `std::atomic`: the syntax sugar papers over things that really need to be thought about (memory ordering!) and disguises the usage of atomics through the code base. We should instead be working to call out that something special is happening here.

src/core/lib/event_engine/utils.h
```diff
@@ -25,12 +25,16 @@
 namespace grpc_event_engine {
 namespace experimental {
 
-std::string HandleToString(EventEngine::TaskHandle handle);
+// Returns a string representation of the EventEngine::*Handle types
+template <typename Handle>
```
ctiller, line 29: Nit:

This would generate less bloat and faster build times if rewritten to:

```
std::string HandleToStringInner(uintptr_t a, uintptr_t b);
template <typename Handle>
std::string HandleToString(const Handle& handle) { return HandleToStringInner(a, b); }
```

(note: now the absl::StrCat isn't duplicated per type, and can be moved into the implementation file so that we're not including complex headers from our headers).

src/core/lib/event_engine/windows/windows_engine.h
```diff
@@ -104,16 +99,58 @@ class WindowsEventEngine : public EventEngine,
   bool Cancel(TaskHandle handle) override;
 
  private:
-  struct Closure;
+  // State of an active connection.
+  // Managed by a shared_ptr, owned exclusively by the timeout callback and the
+  // OnConnectCompleted callback herein.
+  struct ConnectionState {
+    // everything is guarded by mu;
+    grpc_core::Mutex mu;
```
ctiller, line 107: `ABSL_ACQUIRE_BEFORE(WindowsEventEngine::connection_mu_)`

(I'm not sure I've typed that correctly but - we should probably add some variant of ACQUIRE_BEFORE/ACQUIRE_AFTER to these two mutexes since their acquisition order is relied upon in the implementation)

src/core/lib/event_engine/windows/windows_engine.h
```diff
@@ -104,16 +99,58 @@ class WindowsEventEngine : public EventEngine,
...
+    EventEngine::ConnectionHandle connection_handle ABSL_GUARDED_BY(mu);
+    EventEngine::TaskHandle timer_handle ABSL_GUARDED_BY(mu);
+    EventEngine::OnConnectCallback on_connected_user_callback
+        ABSL_GUARDED_BY(mu);
+    EventEngine::Closure* on_connected ABSL_GUARDED_BY(mu);
```
ctiller, line 112: Should we maybe make this a `unique_ptr` and remove some manual memory management?

ctiller (changes requested):

Vignesh2208 (approved):

src/core/lib/event_engine/utils.h
drfloob: Done, thanks.

src/core/lib/event_engine/windows/windows_engine.h
drfloob: Fixed. I identified and fixed a lock re-acquisition issue at the same time, the code has changed a bit.

src/core/lib/event_engine/windows/windows_engine.h
drfloob: In this current design, the closure keeps the state object alive, and executing (or deleting) the closure triggers the shared_ptr<ConnectionState> ref decrement. If we changed this to a unique_ptr, we'd have more complex memory management of the state object itself, having to do something like manually release a held ref at the end of the closure.

src/core/lib/event_engine/windows/windows_engine.cc
ctiller: Please update to fetch_add and fetch_sub for clarity (or if possible use `grpc_core::RefCount`)

ctiller (changes requested):
Looks ok, I'd like us to take a consistent stance on std::atomic allowed operations however, and I don't think ++, -- should be in that set.

drfloob:
> Looks ok, I'd like us to take a consistent stance on std::atomic allowed operations however, and I don't think ++, -- should be in that set.

Removed, given the 2v1 stance. I didn't find any other uses of atomic `operator++` in src/core. I don't think this is easily lintable, so I'd recommend creating a gRPC style guide if something like this is going to be enforceable.

ctiller (approved):

PR #31850, opened 2022-12-09
Add a test case to verify SubchannelStreamClient retry when Health.Watch ends

yijiem:


yijiem:
From local code coverage run, this change increases code coverage for the `subchannel_stream_client.cc` file from 80.8% lines and 92% functions to 90% lines and 100% functions.

markdroth (approved):

PR #28036, opened 2021-11-15
[BinderTransport] Avoid JNI types in public API

sifmelcara:
In this commit we replaces `jobject` type in public API with `void*`.

By doing this, users can compile non-Android code that
references/invokes these APIs. This is useful for running buildable
tests with JDK toolchain.

This commit also cleans up some argument names.

internal bug number 205927046

PR #31852, opened 2022-12-09
[BinderTransport] Make `CreateBinderChannel` parameter names consistent

rock1246:
* Rename `jni_env` to `jni_env_void`
* Rename `application` to `context`

### DIFF
--- a/include/grpcpp/create_channel_binder.h
+++ b/include/grpcpp/create_channel_binder.h
@@ -36,7 +36,7 @@ namespace experimental {
 /// name and class name will be used identify the specific application component
 /// to connect to.
 ///
-/// \param jni_env Pointer to a JNIEnv structure
+/// \param jni_env_void Pointer to a JNIEnv structure
 /// \param context The context that we will use to invoke \a bindService See
 /// https://developer.android.com/reference/android/content/Context#bindService(android.content.Intent,%20android.content.ServiceConnection,%20int)
 /// for detail.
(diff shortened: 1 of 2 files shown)


sifmelcara:
We will also want to update parameter names in https://github.com/grpc/grpc/blob/master/src/core/ext/transport/binder/client/channel_create.cc

If you have time, can you also look into why checks like https://clang.llvm.org/extra/clang-tidy/checks/readability/inconsistent-declaration-parameter-name.html didn't get triggered in this case? Maybe it is because clang-tidy didn't process things in `include/` folder, maybe it is because the check is not turned on at all

sifmelcara:
Just in case you don't know, you can run https://github.com/grpc/grpc/blob/master/tools/distrib/sanitize.sh to sanitize the code (this is similar to the concept of pre-commit hook that you are familiar with)

LittleCVR (approved):

rock1246:
> We will also want to update parameter names in https://github.com/grpc/grpc/blob/master/src/core/ext/transport/binder/client/channel_create.cc
> 

IIUC, I already changed the file.

> If you have time, can you also look into why checks like https://clang.llvm.org/extra/clang-tidy/checks/readability/inconsistent-declaration-parameter-name.html didn't get triggered in this case? Maybe it is because clang-tidy didn't process things in `include/` folder, maybe it is because the check is not turned on at all

Thanks for the info! I didn't know that.

sifmelcara:
(For reference, here is a relevant old PR: #28036 )

sifmelcara:
> > We will also want to update parameter names in https://github.com/grpc/grpc/blob/master/src/core/ext/transport/binder/client/channel_create.cc
> 
> IIUC, I already changed the file.

You are right, for some reason I didn't see the change

sifmelcara (approved):

PR #31853, opened 2022-12-09
Work around UnaryCall hang with corrupt status detail string #29854

jtattermusch:
An improved and more targeted version of https://github.com/grpc/grpc/pull/31450.

Fixes the unary call hang #29854 by eating the exception when attempting to parse a corrupt status details string (and returning an empty status string instead).

### DIFF
--- a/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
+++ b/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
@@ -137,7 +137,7 @@ namespace Grpc.Core.Internal
             return true;
         }
 
-        // This method decodes the status details string "safely" and returns and empty
+        // This method decodes the status details string "safely" and returns an empty
         // string in case it seems corrupt.
         // This is to stop UnaryCall from hanging if there is an issue with decoding
         // the UTF8 status details string (likely due to an unexplained corruption).
--- a/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
+++ b/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
@@ -17,6 +17,7 @@
 #endregion
 
 using System;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
 using Grpc.Core;
@@ -91,7 +92,7 @@ namespace Grpc.Core.Internal
         {
             UIntPtr detailsLength;
             IntPtr detailsPtr = Native.grpcsharp_batch_context_recv_status_on_client_details(this, out detailsLength);
-            string details = MarshalUtils.PtrToStringUTF8(detailsPtr, (int)detailsLength.ToUInt32());
+            string details = StatusDetailsPtrToStringUTF8(detailsPtr, (int)detailsLength.ToUInt32());
             string debugErrorString = Marshal.PtrToStringAnsi(Native.grpcsharp_batch_context_recv_status_on_client_error_string(this));
             var status = new Status(Native.grpcsharp_batch_context_recv_status_on_client_status(this), details, debugErrorString != null ? new CoreErrorDetailException(debugErrorString) : null);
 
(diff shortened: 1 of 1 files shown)


jtattermusch:
@tonydnewell  and @apolcyn, please review.

jtattermusch:
@zhenomansh please confirm that this change is enough to work around the problem you're seeing.

zhenomansh:
> @zhenomansh please confirm that this change is enough to work around the problem you're seeing.

@jtattermusch, according to all collected logs from production the fix should work.

I know I am annoying :), but I still believe it is better to set the exception in the catch area in the UnaryCall method just in case of another unpredictable failure like this string formatting from any of methods inside try.

                            try
                            {
                                using (profiler.NewScope("AsyncCall.UnaryCall.HandleBatch"))
                                {
                                    HandleUnaryResponse(success, ctx.GetReceivedStatusOnClient(), ctx.GetReceivedMessageReader(), ctx.GetReceivedInitialMetadata());
                                }
                            }
                            catch (Exception e)
                            {
                                Logger.Error(e, "Exception occurred while invoking completion delegate.");
                                **unaryResponseTcs.SetException(e);**
                            }

jtattermusch:
> > @zhenomansh please confirm that this change is enough to work around the problem you're seeing.
> 
> @jtattermusch, according to all collected logs from production the fix should work.
> 
> I know I am annoying :), but I still believe it is better to set the exception in the catch area in the UnaryCall method just in case of another unpredictable failure like this string formatting from any of methods inside try.
> 
> ```
>                         try
>                         {
>                             using (profiler.NewScope("AsyncCall.UnaryCall.HandleBatch"))
>                             {
>                                 HandleUnaryResponse(success, ctx.GetReceivedStatusOnClient(), ctx.GetReceivedMessageReader(), ctx.GetReceivedInitialMetadata());
>                             }
>                         }
>                         catch (Exception e)
>                         {
>                             Logger.Error(e, "Exception occurred while invoking completion delegate.");
>                             **unaryResponseTcs.SetException(e);**
>                         }
> ```

I don't agree. If HandleUnaryResponse() doesn't run (which in your case won't since ctx.GetReceivedStatusOnClient() will throw even before HandleUnaryResponse will run), it will leak resources since HandleUnaryResponse handles most of the cleanup after the call (including the underlying native objects).  So the right approach is to make sure ctx.GetReceivedStatusOnClient() doesn't throw (which is what my workaround is doing) and then HandleUnaryResponse can "finish" the call and handle all the necessary cleanup.

tonydnewell (approved):

tonydnewell:
While this fix addresses a workaround for the precise problem reported it is not as thorough or robust as my workaround that addressed any possibility of a hang in `UnaryCall` - see https://github.com/grpc/grpc/pull/31450#discussion_r1045201501

jtattermusch:
> While this fix addresses a workaround for the precise problem reported it is not as thorough or robust as my workaround that addressed any possibility of a hang in `UnaryCall` - see [#31450 (comment)](https://github.com/grpc/grpc/pull/31450#discussion_r1045201501)

We discussed this offline and agreed that this is the best way forward - since we don't know the exact root cause of https://github.com/grpc/grpc/issues/29854 and this PR is basically a workaround, we don't want to risk introducing unintended consequences (performance regressions, losing consistency in implementation between sync vs unary calls, etc.) more that it's necessary, so the workaround should be as minimal as possible. I believe this PR satisfies that property.

jtattermusch:
The test failures are C# unrelated and are due to tests being run on an old branch.

src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs
```diff
@@ -136,6 +137,27 @@ protected override bool ReleaseHandle()
             return true;
         }
 
+        // This method decodes the status details string "safely" and returns and empty
```
krishnacx, line 140: *returns an

apolcyn (approved):

jtattermusch:
@veblush  can you please merge this into the 1.46.x branch?

jtattermusch:

PR #31856, opened 2022-12-09
Use enums for experiments

ananda1066:


src/core/lib/experiments/experiments.h
```diff
@@ -25,27 +25,61 @@
 
 namespace grpc_core {
 
-inline bool IsTcpFrameSizeTuningEnabled() { return IsExperimentEnabled(0); }
-inline bool IsTcpReadChunksEnabled() { return IsExperimentEnabled(1); }
-inline bool IsTcpRcvLowatEnabled() { return IsExperimentEnabled(2); }
-inline bool IsPeerStateBasedFramingEnabled() { return IsExperimentEnabled(3); }
-inline bool IsFlowControlFixesEnabled() { return IsExperimentEnabled(4); }
+enum ExperimentNames {
```
ctiller, line 28: enum class

src/core/lib/experiments/experiments.h
```diff
@@ -25,27 +25,61 @@
...
+  TcpFrameSizeTuning = 0,
```
ctiller, line 29: prefix with k

ctiller, line 29: drop the `= X` -- first one is always implicitly 0, and dropping it will avoid merge conflicts

ctiller: (we could add a static assert for the first/last if we wanted to prove that)

PR #31865, opened 2022-12-10
EventEngine::RunAfter migration ConnectionAttemptInjector

yijiem:

### DIFF
--- a/test/cpp/end2end/connection_attempt_injector.cc
+++ b/test/cpp/end2end/connection_attempt_injector.cc
@@ -20,7 +20,9 @@
 #include "absl/utility/utility.h"
 
 #include "src/core/lib/address_utils/sockaddr_utils.h"
+#include "src/core/lib/event_engine/default_event_engine.h"
 #include "src/core/lib/gprpp/sync.h"
+#include "src/core/lib/iomgr/exec_ctx.h"
 
 // defined in tcp_client.cc
 extern grpc_tcp_client_vtable* grpc_tcp_client_impl;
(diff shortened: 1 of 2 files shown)


test/cpp/end2end/connection_attempt_injector.cc
```diff
@@ -191,17 +193,19 @@ ConnectionAttemptInjector::InjectedDelay::InjectedDelay(
     grpc_pollset_set* interested_parties, const EndpointConfig& config,
     const grpc_resolved_address* addr, grpc_core::Timestamp deadline)
     : attempt_(closure, ep, interested_parties, config, addr, deadline) {
-  GRPC_CLOSURE_INIT(&timer_callback_, TimerCallback, this, nullptr);
   grpc_core::Timestamp now = grpc_core::Timestamp::Now();
   duration = std::min(duration, deadline - now);
-  grpc_timer_init(&timer_, now + duration, &timer_callback_);
+  grpc_event_engine::experimental::GetDefaultEventEngine()->RunAfter(
+      duration, [this] {
+        grpc_core::ApplicationCallbackExecCtx callback_exec_ctx;
```
drfloob, line 200: Just a comment, no fix required: This is an interesting case in that I think these contexts are necessary, but they may have been _technically_ required in the previous implementation as well - `grpc_timer_init` may use `ExecCtx::Run` directly in a few situations, which will crash if one is not on the stack.

drfloob (approved):
LGTM, thanks Yijie.

PR #31868, opened 2022-12-10
fix generate_projects.sh fails in M1 Mac

HannahShiSFB:
use `buildozer-darwin-arm64` package with ARM Mac

add `if __name__ == "__main__":` test which is required on Mac after Python 3.8 because `spawn` is used in `multiprocessing`

### DIFF
--- a/tools/distrib/buildozer.sh
+++ b/tools/distrib/buildozer.sh
@@ -33,6 +33,7 @@ function download_buildozer() {
     platform="$(uname -sm)"
     case "${platform}" in
         "Linux x86_64")     download_link="https://github.com/bazelbuild/buildtools/releases/download/${BUILDOZER_VERSION}/buildozer-linux-amd64";;
+        "Linux aarch64")    download_link="https://github.com/bazelbuild/buildtools/releases/download/${BUILDOZER_VERSION}/buildozer-linux-arm64";;
         "Darwin x86_64")    download_link="https://github.com/bazelbuild/buildtools/releases/download/${BUILDOZER_VERSION}/buildozer-darwin-amd64";;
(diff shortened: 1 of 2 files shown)


tools/distrib/buildozer.sh
```diff
@@ -30,11 +30,12 @@ function error_handling() {
 }
 
 function download_buildozer() {
-    platform="$(uname -s)"
+    platform="$(uname -sm)"
     case "${platform}" in
-        Linux*)     download_link="https://github.com/bazelbuild/buildtools/releases/download/${BUILDOZER_VERSION}/buildozer-linux-amd64";;
-        Darwin*)    download_link="https://github.com/bazelbuild/buildtools/releases/download/${BUILDOZER_VERSION}/buildozer-darwin-amd64";;
-        *)          error_handling "Unsupported platform: ${platform}";;
+        "Linux x86_64")     download_link="https://github.com/bazelbuild/buildtools/releases/download/${BUILDOZER_VERSION}/buildozer-linux-amd64";;
```
drfloob, line 35: Please do the same for `Linux arm64`, since there's a `buildozer-linux-arm64` release as well.

drfloob (commented):
Thanks Hannah. CC'ing @sampajano for visibility.

I don't think many of us have the opportunity to test on M1, so if there are other python scripts (those in tools/distrib/sanitize.sh especially) that require changes for M1, a cleanup would be appreciated.

sampajano (approved):
@drfloob Thanks for cc'ing! Lgtm besides your comment :)

drfloob:
Were you waiting on me, @sampajano? If so, LGTM.

drfloob (approved):

PR #31869, opened 2022-12-10
Upgrade boringssl

### LINKED ISSUE #31334: Undefined behavior in BoringSSL-GRPC

gRPC-Core 1.49, Objective-C

iOS 15

Apple clang version 14.0.0 (clang-1400.0.29.102)

### What did you do?

Tried using GRPC with Clang's Undefined Behavior Sanitizer enabled.

No undefined behavior.

### What did you see instead?

Undefined behavior in BoringSSL-GRPC: "Applying zero offset to null pointer" in CRYPTO_poly1305_update().

This bug is already fixed upstream, see https://bugs.chromium.org/p/boringssl/issues/detail?id=523. However, the `BoringSSL-GRPC.podspec` references an old commit from 2021. Please update to a newer version of BoringSSL.

HannahShiSFB:
Update Boringssl (which is more than a year old) for 
https://boringssl-review.googlesource.com/c/boringssl/+/52125
to fix  #31334

Update to revision https://github.com/google/boringssl/commit/2a3f97f7970676b7f6855f5d7504e03976494d75 because next commit  https://github.com/google/boringssl/commit/6cda656e4a2e1545e23b30dfb7dc05fe423e3f6c added a `time.h` file which hides the `time.h` in apple SDK. Will investigate later.

Also updated the readme file replacing `main-with-bazel` (which is not updated) with `master-with-bazel` (updated by 
BoringSSL Robot every few hours)

Can not do the last step "- Run `bazel/update_mirror.sh` to update GCS mirror." due to lack of permission. @sampajano


sampajano:
@HannahShiSFB Thanks for checking! I was able to run "bazel/update_mirror.sh" successfully. thanks!

jtattermusch:
A few concerns:
- we've have an internal script that updates boringssl submodule now - I'd have to check that the changes you've made using the open source instructions are aligned with the changes generated by the internal script.
- looks like we can't really update all the way to the newest boringssl version currently - that is a problem that needs to be solved.

Good catch with the `main-with-bazel` vs `master-with-bazel` branches - looks like we might need to update the internal script as well. Also, as a side effect of this, we haven't upgraded boringssl for recent releases since  our automatic was looking at the  "main-with-bazel" branch, which isn't being updated anymore).

HannahShiSFB:
in favor of #32357

PR #31871, opened 2022-12-12
[cleanups] Remove `include/grpc/impl/codegen/slice.h`

ralphchung:
This PR is a part of [b/196639718](http://b/196639718). It tries to move the content of `include/grpc/impl/codegen/slice.h` to `include/grpc/impl/slice_type.h`. We do not remove anything to achieve backward compatibility. The file `include/grpc/impl/codegen/slice.h` should be removed in the future.

This does not require any internal change, which has been verified through a test import CL.


Vignesh2208:
Can you move it to slice_def.h instead of slice_type.h ?

PR #31877, opened 2022-12-13
[EventEngine] RunAfter migration: Chttp2Connector

yijiem:

### DIFF
--- a/src/core/ext/transport/chttp2/client/chttp2_connector.cc
+++ b/src/core/ext/transport/chttp2/client/chttp2_connector.cc
@@ -178,7 +178,7 @@ void Chttp2Connector::OnHandshakeDone(void* arg, grpc_error_handle error) {
                                           args->read_buffer,
                                           &self->on_receive_settings_, nullptr);
       RefCountedPtr<Chttp2Connector> cc = self->Ref();
-      self->event_engine_ = args->args.GetObject<EventEngine>();
+      self->event_engine_ = self->args_.channel_args.GetObject<EventEngine>();
       self->timer_handle_ = self->event_engine_->RunAfter(
           self->args_.deadline - Timestamp::Now(), [self = std::move(cc)] {
             ApplicationCallbackExecCtx callback_exec_ctx;
--- a/src/core/ext/transport/chttp2/client/chttp2_connector.cc
+++ b/src/core/ext/transport/chttp2/client/chttp2_connector.cc
@@ -211,8 +211,15 @@ void Chttp2Connector::OnReceiveSettings(void* arg, grpc_error_handle error) {
         self->result_->Reset();
       }
       self->MaybeNotify(error);
-      self->args_.channel_args.GetObject<EventEngine>()->Cancel(
-          self->timer_handle_);
+      if (self->timer_handle_.has_value()) {
+        if (self->args_.channel_args.GetObject<EventEngine>()->Cancel(
+                *self->timer_handle_)) {
+          // If we have cancelled the timer successfully, call Notify() again
+          // since the timer callback will not be called now.
+          self->MaybeNotify(absl::OkStatus());
+        }
+        self->timer_handle_.reset();
+      }
     } else {
       // OnTimeout() was already invoked. Call Notify() again so that notify_
       // can be invoked.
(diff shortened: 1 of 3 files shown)


src/core/ext/transport/chttp2/client/chttp2_connector.cc
```diff
@@ -171,13 +174,16 @@ void Chttp2Connector::OnHandshakeDone(void* arg, grpc_error_handle error) {
...
-      GRPC_CLOSURE_INIT(&self->on_timeout_, OnTimeout, self,
-                        grpc_schedule_on_exec_ctx);
-      grpc_timer_init(&self->timer_, self->args_.deadline, &self->on_timeout_);
+      self->timer_handle_ =
+          self->args_.channel_args.GetObject<EventEngine>()->RunAfter(
+              self->args_.deadline - Timestamp::Now(), [self = self->Ref()] {
+                ApplicationCallbackExecCtx callback_exec_ctx;
+                ExecCtx exec_ctx;
+                static_cast<Chttp2Connector*>(self.get())->OnTimeout();
+              });
```
yijiem, line 186: An alternative is to write something like:

```
// self->Ref() returns a RefCountedPtr<SubchannelConnector>, an
// implicit downcast will be used here:
RefCountedPtr<Chttp2Connector> cc = self->Ref();
self->timer_handle_ =
    self->args_.channel_args.GetObject<EventEngine>()->RunAfter(
        self->args_.deadline - Timestamp::Now(), [self = std::move(cc)] {
          ApplicationCallbackExecCtx callback_exec_ctx;
          ExecCtx exec_ctx;
          self->OnTimeout();
        });
```

I'm not sure which one is more readable or better in other ways. So went with the one without creating an extra `cc` variable.

src/core/ext/transport/chttp2/client/chttp2_connector.cc
```diff
@@ -205,7 +211,15 @@ void Chttp2Connector::OnReceiveSettings(void* arg, grpc_error_handle error) {
         self->result_->Reset();
       }
       self->MaybeNotify(error);
-      grpc_timer_cancel(&self->timer_);
+      if (self->timer_handle_.has_value()) {
+        if (self->args_.channel_args.GetObject<EventEngine>()->Cancel(
```
ctiller, line 215: We probably don't want to be digging this out of channel args repeatedly

src/core/ext/transport/chttp2/client/chttp2_connector.cc
ctiller: I like this form

yijiem: Done

src/core/ext/transport/chttp2/client/chttp2_connector.cc
yijiem: Added an `event_engine_` member, PTAL.

src/core/ext/transport/chttp2/client/chttp2_connector.cc
```diff
@@ -171,13 +174,17 @@ void Chttp2Connector::OnHandshakeDone(void* arg, grpc_error_handle error) {
...
                         grpc_schedule_on_exec_ctx);
-      self->Ref().release();  // Ref held by OnTimeout()
       grpc_chttp2_transport_start_reading(self->result_->transport,
                                           args->read_buffer,
                                           &self->on_receive_settings_, nullptr);
-      GRPC_CLOSURE_INIT(&self->on_timeout_, OnTimeout, self,
-                        grpc_schedule_on_exec_ctx);
-      grpc_timer_init(&self->timer_, self->args_.deadline, &self->on_timeout_);
+      RefCountedPtr<Chttp2Connector> cc = self->Ref();
+      self->event_engine_ = self->args_.channel_args.GetObject<EventEngine>();
```
drfloob, line 181: Does it make more sense to store this on `Connect` when the rest of the members are initialized as well?

yijiem: Done, agree it is clearer this way.

drfloob (approved):

PR #31879, opened 2022-12-13
Fix return type of abort()

Dreamsorcerer:
This says the method always raises an exception, so it should not return `None`.

This leads to typing errors where mypy is unaware that code after an abort() call will not be run.


XuanWang-Amos:
Hi, thanks for the fix, we'll close the PR and create another to fix it.

PR #31890, opened 2022-12-14
Move posix event engine to new interface defined in #31816

Vignesh2208:
The additional missing methods are implemented.

This new interface defines additional file descriptor specific methods to event engine endpoint and listener for file descriptor specific operations.


src/core/lib/event_engine/posix_engine/posix_engine_listener.h
```diff
@@ -189,33 +204,71 @@ class PosixEngineListener
...
   std::shared_ptr<PosixEngineListenerImpl> impl_;
+  // Set to true when the listener had been explicitly shutdown.
+  std::atomic<bool> shutdown_{false};
 };
 
 #else  // GRPC_POSIX_SOCKET_TCP
 
-class PosixEngineListener
-    : public grpc_event_engine::experimental::EventEngine::Listener {
+class PosixEngineListener : public PosixListenerWithFdSupport {
```
ralphchung, line 239: Why do we need this supporting nothing?

Vignesh2208: This is in-case the target is built on non-posix systems. An attempt to use the object should fail.

src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
```diff
@@ -183,13 +188,47 @@ void PosixEngineListenerImpl::AsyncConnectionAcceptor::NotifyOnAccept(
...
+  }
+  if (fd < 0) {
+    return absl::UnknownError(
+        absl::StrCat("HandleExternalConnection: Invalid peer socket: ", fd));
+  }
+  PosixSocketWrapper sock(fd);
+  (void)sock.SetSocketNoSigpipeIfPossible();
+  std::string peer_name = *sock.PeerAddressString();
+  auto endpoint = CreatePosixEndpoint(
+      /*handle=*/poller_->CreateHandle(fd, peer_name,
```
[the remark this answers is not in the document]
Vignesh2208: No poller_ is never nullptr.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -1168,20 +1168,29 @@ void PosixEndpointImpl::Write(
...
   handle_->ShutdownHandle(why);
   Unref();
 }
 
 PosixEndpointImpl ::~PosixEndpointImpl() {
-  handle_->OrphanHandle(on_done_, nullptr, "");
+  int release_fd = -1;
+  handle_->OrphanHandle(on_done_,
+                        on_release_fd_ == nullptr ? nullptr : &release_fd, "");
+  if (on_release_fd_ != nullptr) {
```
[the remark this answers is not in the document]
Vignesh2208: If on_release_fd_ is provided, it should be called regardless of whether release_fd is -1 or not. So only checking for on_release_fd_ should be sufficient.

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
ralphchung: Is compile-time error preferred or runtime error?

Vignesh2208: It already would generate a compile time error. The windows build/checks will fail without this. I need to do some build cleanup once event engine is landed to make sure this doesn't get pulled in windows builds. We should be able to remove this at that point.

Vignesh2208: By default all targets declared in src/core/BUILD and in BUILD will be built on all platforms by the kokoro checks.

src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
```diff
@@ -183,13 +188,47 @@ void PosixEngineListenerImpl::AsyncConnectionAcceptor::NotifyOnAccept(
...
+    return absl::UnknownError(absl::StrCat(
+        "HandleExternalConnection: Invalid listener socket: ", listener_fd));
+  }
+  if (fd < 0) {
+    return absl::UnknownError(
+        absl::StrCat("HandleExternalConnection: Invalid peer socket: ", fd));
+  }
+  PosixSocketWrapper sock(fd);
+  (void)sock.SetSocketNoSigpipeIfPossible();
+  std::string peer_name = *sock.PeerAddressString();
```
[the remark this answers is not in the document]
Vignesh2208: Good point. Added a check now.

src/core/lib/event_engine/posix_engine/posix_endpoint.h
```diff
@@ -622,18 +629,30 @@ class PosixEndpoint
     return impl_->GetLocalAddress();
   }
 
+  int GetWrappedFd() override { return impl_->GetWrappedFd(); }
+
+  void Shutdown(absl::AnyInvocable<void(absl::StatusOr<int> release_fd)>
+                    on_release_fd) override {
+    if (!shutdown_.exchange(true, std::memory_order_acq_rel)) {
+      impl_->MaybeShutdown(absl::InternalError("Endpoint closing"),
```
drfloob, line 637: nit: InternalError is reserved for serious errors. Aborted may be a better choice here, though OkStatus would be best since this is a normal occurrence - though I believe we lose the message with OkStatus, so it may be off the table.

Please read the status code definitions here: https://github.com/abseil/abseil-cpp/blob/master/absl/status/status.h#L67

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -558,12 +561,56 @@ EventEngine::ConnectionHandle PosixEventEngine::Connect(
...
+  GPR_DEBUG_ASSERT(fd > 0);
+  PosixEventPoller* poller = poller_manager_->Poller();
+  GPR_DEBUG_ASSERT(poller != nullptr);
+  EventHandle* handle =
+      poller->CreateHandle(fd, "tcp-client", poller->CanTrackErrors());
+  return CreatePosixEndpoint(handle, nullptr, shared_from_this(),
+                             std::move(memory_allocator),
+                             TcpOptionsFromEndpointConfig(config));
+#else   // GRPC_POSIX_SOCKET_TCP
+  GPR_ASSERT(false &&
```
drfloob, line 578: Inconvenient timing: https://github.com/grpc/grpc/pull/31757 just landed, which removes all occurrences of `GPR_ASSERT(false && ...)`, replacing them with `Crash(...)`.

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -558,12 +561,56 @@ EventEngine::ConnectionHandle PosixEventEngine::Connect(
...
+          bool /*is_external*/, MemoryAllocator allocator,
+          SliceBuffer* /*pending_data*/) mutable {
+        on_accept_cb(std::move(ep), std::move(allocator));
+      };
+  return std::make_unique<PosixEngineListener>(
+      std::move(posix_on_accept), std::move(on_shutdown), config,
+      std::move(memory_allocator_factory), poller_manager_->Poller(),
+      shared_from_this());
+#else   // GRPC_POSIX_SOCKET_TCP
+  GPR_ASSERT(false &&
```
drfloob, line 603: Should now be `Crash("EventEngine::CreateListener is not supported on this platform");`

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
```diff
@@ -148,6 +164,7 @@ class PosixEngineListenerImpl
     }
 
    private:
+    PosixListenerWithFdSupport::OnPosixBindNewFdCallback on_append_ = nullptr;
```
drfloob, line 167: AnyInvocable's default constructor leaves it in an empty state already.

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
```diff
@@ -119,9 +126,18 @@ class PosixEngineListenerImpl
...
+
+    void UpdateOnAppendCallback(
+        PosixListenerWithFdSupport::OnPosixBindNewFdCallback on_append) {
+      on_append_ = std::move(on_append);
+    }
+
     void Append(ListenerSocket socket) override {
       acceptors_.push_back(new AsyncConnectionAcceptor(
           listener_->engine_, listener_->shared_from_this(), socket));
+      if (on_append_ != nullptr) {
```
drfloob, line 138: nit, optional: `if(on_append_)` also works, since AnyInvocable overrides `operator bool`, and the nullptr comparison is equally synthetic since `on_append_` is not a pointer. I'm not sure if we want to standardize on either syntax.

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
```diff
@@ -189,33 +204,71 @@ class PosixEngineListener
...
-    : public grpc_event_engine::experimental::EventEngine::Listener {
+class PosixEngineListener : public PosixListenerWithFdSupport {
  public:
   PosixEngineListener() = default;
   ~PosixEngineListener() override = default;
-  absl::StatusOr<int> Bind(const grpc_event_engine::experimental::EventEngine::
-                               ResolvedAddress& /*addr*/) override {
+  absl::StatusOr<int> Bind(
+      const EventEngine::ResolvedAddress& /*addr*/) override {
+    GPR_ASSERT(false &&
```
drfloob, line 245: Should now be `Crash(...)`

src/core/lib/event_engine/posix_engine/posix_endpoint.h
Vignesh2208: changed it to absl::FailedPreconditionError

src/core/lib/event_engine/posix_engine/posix_engine.cc
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/posix_engine.cc
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
Vignesh2208: Fixed.

drfloob (approved):

PR #31894, opened 2022-12-15
[cleanups] Clean up `GrpcLibraryInitializer`

ralphchung:
After the removal of `g_core_codegen_interface`, `GrpcLibraryInitializer` no longer needs to exist. It is time to remove it.


src/cpp/client/credentials_cc.cc
```diff
@@ -21,12 +21,11 @@
 
 namespace grpc {
 
-static grpc::internal::GrpcLibraryInitializer g_gli_initializer;
```
ctiller, line 24: maybe remove constructors and destructors entirely from this cc file?

ralphchung: Forgot to catch this. Updated.

ctiller (approved):
looks fantastic!

PR #31904, opened 2022-12-15
WRR: implement WRR LB policy

markdroth:


src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -0,0 +1,942 @@
...
+    gpr_log(GPR_INFO,
+            "[WRR %p] subchannel %s: qps=%f, cpu_utilization=%f: setting "
+            "weight=%f weight_=%f now=%s last_update_time_=%s "
+            "non_empty_since_=%s",
+            wrr_.get(), key_.c_str(), qps, cpu_utilization, weight, weight_,
+            now.ToString().c_str(), last_update_time_.ToString().c_str(),
+            non_empty_since_.ToString().c_str());
+  }
+  if (weight > 0 && non_empty_since_ == Timestamp::InfFuture()) {
+    non_empty_since_ = now;
```
yousukseung, line 403: We should reset this if the weight is zero as well.

markdroth: If the weight is 0, we return on line 388.

I don't think we want to differentiate between receiveing a backend metric report with all values 0 and not receiving any backend metric report.  I think we want to treat both cases the same by applying the expiration time.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -0,0 +1,955 @@
...
+  if (GRPC_TRACE_FLAG_ENABLED(grpc_lb_wrr_trace)) {
+    gpr_log(GPR_INFO,
+            "[WRR %p] subchannel %s: qps=%f, cpu_utilization=%f: setting "
+            "weight=%f weight_=%f now=%s last_update_time_=%s "
+            "non_empty_since_=%s",
+            wrr_.get(), key_.c_str(), qps, cpu_utilization, weight, weight_,
+            now.ToString().c_str(), last_update_time_.ToString().c_str(),
+            non_empty_since_.ToString().c_str());
+  }
+  if (non_empty_since_ == Timestamp::InfFuture()) non_empty_since_ = now;
```
yousukseung, line 403: I think my previous comment is lost.

Design and Stubby reset it to infy if weight is 0. We reviewed the snippet together (the atomic-only version) in the design (pasted below). I see no reason to diverge vs. Stubby only here. I think it makes sense to hold off trusting the weight until it reports valid/non-empty weights consecutively, which is the definition in the design.

MaybeUpdateWeight(qps, cpu_utilization) {
   uint64 now = epoch_ms();
   uint64 new_time = (qps > 0 && cpu_utilization > 0) ? now : max_uint64;
   uint64 non_empty_since = 
      subchannel.last_time.compare_exchange(max_uint64, new_time);
   if (now - non_empty_since < blackout_period) return;
   subchannel.weight.store(ComputeWeight(qps, cpu_utilization));
}

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -0,0 +1,955 @@
...
+    }
+    subchannel()->RequestConnection();
+  } else if (new_state == GRPC_CHANNEL_READY) {
+    // If we transition back to READY state, restart the blackout period.
+    // Note that we cannot guarantee that we will never receive
+    // lingering callbacks for backend metric reports from the previous
+    // connection after the new connection has been established, but they
+    // should be masked by new backend metric reports from the new
+    // connection by the time the blackout period ends.
+    weight_->ResetNonEmptySince();
```
yousukseung, line 863: Why don't we do this when the channel moves out of READY and also simply invalidate all weight information at all? As you mentioned since there can always be an extreme timing issue w.r.t. synchronization, it looks to me that resetting values at the earliest is safer and easier to maintain. Also all weight states (weight and two timestamps) become invalid will never be used as soon as a subchannel becomes non-READY so we don't need to keep them, correct?

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
yousukseung: I meant to reset the blackout timer before returning.

This is a non-trivial deviation from what the blackout period does and I think we should do what's proven as discussed. And resetting makes sense to me. An explicit zero weight can happen e.g. when the QPS counter is reset/restarted, CPU% going out of range etc. These seem to be as good signals or better to hold off accepting load reports from that subchannel until it stabilizes, as we do to a subchannel comes back online to prevent churns.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -0,0 +1,955 @@
...
+  }
+  return PickResult::Complete(subchannel_info.subchannel,
+                              std::move(subchannel_call_tracker));
+}
+
+size_t WeightedRoundRobin::Picker::PickIndex() {
+  // Grab the mutex and do WRR pick if we have a scheduler.
+  {
+    MutexLock lock(&mu_);
+    if (scheduler_.has_value()) {
```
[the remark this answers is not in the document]
yousukseung: I am not clear what you mean by "before writing the atomic", but StaticStrideScheduler::Pick() loops and the loop count is technically unbounded with a non-negligible chance to loop for a single-digit number of times. AFAIK this is a known trade-off that StaticStrideScheduler::Pick() design takes but assuming it's concurrent. I think we should use StaticStrideScheduler in a way that it is designed and optimized for now, esp. given there's an upcoming PR to make it concurrent.

And I recall the LB pick() call was mentioned as one of the most important choke point in Stubby in the past. Any delay here adds up to the RPC latency directly that services care the most, so I think it's well worth it to optimize where possible.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -0,0 +1,955 @@
...
+
+    RefCountedPtr<WeightedRoundRobin> wrr_;
+    const bool use_per_rpc_utilization_;
+    const Duration weight_update_period_;
+    const Duration weight_expiration_period_;
+    const Duration blackout_period_;
+    std::vector<SubchannelInfo> subchannels_;
+
+    Mutex mu_;
+    uint32_t scheduler_state_ ABSL_GUARDED_BY(&mu_);
```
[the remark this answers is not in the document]
yousukseung: By persisent I meant for the lifetime of the WeightRoundRobinPolicy as in Stubby. I misread that the counter is per-scheduler in the previous comment but I think it does not change the point I wanted to make that keeping next_seq_ stable the whole time always benefits. This section also mentions this is an important feature: go/static-stride-scheduler#concurrency.

For example, since we use a map to dedup addresses the order is stable as long as the subchannel set is the same, and keeping next_seq_ persisent across pickers clearly helps improving the stability. This includes resolver updates with the same server address or a flapping subchannel (not perfect). In all I don't see a reason not to do what Stubby does here.

class WRR { 
   atomic<uint32_t> next_seq_ { rand() };
   uint32_t NextSeq() { return next_seq_ .fetch_add(1); }
   void UpdateWeights(..) {
       StaticStrideScheduler::Make(.., [this](){ return NextSeq(); });
   }
}

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -0,0 +1,955 @@
...
+    absl::optional<StaticStrideScheduler> scheduler_ ABSL_GUARDED_BY(&mu_);
+    absl::optional<grpc_event_engine::experimental::EventEngine::TaskHandle>
+        timer_handle_ ABSL_GUARDED_BY(&mu_);
+
+    // Used when falling back to RR.
+    size_t last_picked_index_;
```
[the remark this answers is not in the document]
yousukseung: Please disregard, I misread it was per scheduler and not per picker.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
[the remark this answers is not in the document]
markdroth: I think we already discussed this extensively in the design.  Whenever we update the picker, the list of subchannels can change (e.g., because we got a new address list from the channel, or because a subchannel transitioned into or out of READY state), so the indexes of the subchannels change.  When the indexes change, there is no benefit to retaining the scheduler state, because it may no longer point to the same place in the rotation that it had previously pointed to.

Note that this is exactly consistent with how RR works in gRPC: whenever we generate a new picker, we [start from a random index in the list](https://github.com/grpc/grpc/blob/7fd808f6f9d12e3e63dc299984567aae591cf2eb/src/core/ext/filters/client_channel/lb_policy/round_robin/round_robin.cc#L212).  I think it makes sense to do the same thing in WRR until and unless it proves to be a problem.  Let's not add complexity that we don't know if we're actually going to need.

(To be fair, there are some cases today where we update the picker without the subchannel indexes actually changing.  This can happen whenever a subchannel changes from one non-READY state to another, such as going from IDLE to CONNECTING.  In these cases, we do wind up needlessly losing the scheduler state, and we could optimize that by avoiding a picker update in those cases.  But again, I don't think we should add the extra complexity here until and unless we know we're going to need it.)

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
markdroth: It looks like you're bouncing back and forth between comment threads here, and it's fairly confusing.  I think I addressed this in the [thread below](https://github.com/grpc/grpc/pull/31904#discussion_r1071639216).

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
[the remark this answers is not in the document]
markdroth: We also discussed the fact that the scheduler treats a weight of 0 as "average weight", not as "do not pick this address", which means that we have to change the indexes whenever a subchannel transitions into or out of state READY.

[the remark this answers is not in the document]
markdroth: As per our conversation, I've changed this to store the scheduler state in the LB policy itself so that it can be retained across pickers.

I'm still not sure this will actually yield the same benefit for gRPC that it does for Stubby, because gRPC will not order the addresses the same way that Stubby does, especially when backend tasks move around.  But it doesn't add too much complexity, and it does pave the way for #31973.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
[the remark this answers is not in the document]
markdroth: I wound up changing this as part of the change we discussed to share the scheduler state between pickers.  But we're still not going to get any additional parallelism here until #31973 lands.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
markdroth: As we discussed, I think the current structure is required, because the orca protos don't currently differentiate between 0 and unset.

If you think you can figure out a way to represent this logic with one fewer timestamp, let me know.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -549,11 +554,13 @@ void WeightedRoundRobin::Picker::BuildSchedulerAndStartTimerLocked() {
     gpr_log(GPR_INFO, "[WRR %p picker %p] new weights: %s", wrr_.get(), this,
             absl::StrJoin(weights, " ").c_str());
   }
-  scheduler_ = StaticStrideScheduler::Make(
-      weights,
-      // This requires holding mu_, but we don't have a way to plumb the
-      // annotation through absl::AnyInvocable<>.
-      [this]() ABSL_NO_THREAD_SAFETY_ANALYSIS { return scheduler_state_++; });
+  auto scheduler = StaticStrideScheduler::Make(
+      weights, [this]() { return wrr_->scheduler_state_.fetch_add(1); });
```
yousukseung, line 558: We should be good with the relaxed ordering here, also the case per Stubby.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -549,11 +554,13 @@ void WeightedRoundRobin::Picker::BuildSchedulerAndStartTimerLocked() {
...
+  if (scheduler.has_value()) {
+    scheduler_ = std::make_shared<StaticStrideScheduler>(std::move(*scheduler));
+  } else {
+    scheduler_.reset();
+  }
```
yousukseung, line 563: With #31973, since we share the same mu_, it appears the current implementation will block all new pick() calls every time we rebuild the map. If so, this seems to somewhat negate gains from this change and #31973 periodically. Can we have a separate mutex only for scheduler_ e.g. with using a wrapper? As I shared I think pick() is one of the most critical path regarding RPC latency and close optimization is well worth it.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -523,17 +524,21 @@ WeightedRoundRobin::PickResult WeightedRoundRobin::Picker::Pick(
...
   }
+  // If we have a scheduler, use it to do a WRR pick.
+  if (scheduler != nullptr) return scheduler->Pick();
   // We don't have a scheduler (i.e., either all of the weights are 0 or
   // there is only one subchannel), so fall back to RR.
-  last_picked_index_ = (last_picked_index_ + 1) % subchannels_.size();
-  return last_picked_index_;
+  size_t index = last_picked_index_.fetch_add(1);
+  if (index > 0 && index % subchannels_.size() == 0) {
+    last_picked_index_.fetch_sub(subchannels_.size());
```
yousukseung, line 539: Do we need this? I think we can let it just grow and wrap and always take the modulo.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
markdroth: Good point.  I've split this up so that we use separate mutexes for the timer and for the scheduler, to minimize the critial period for the latter mutex.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
yousukseung: I think it's not critical but still not sure if we need this. When it wraps it will jump to an index which will be approximately uniformly random, which is the same as when we create a new picker. Given we are not concerned with the impact of a new picker that much I think it's fine to not have this. And when the subchannel list is very small I guess the overhead becomes a bit unnecessary.

yousukseung (approved):
I have no more comments other than the RR index last comment.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
markdroth: We create a new picker only when something is changing (either the list of addresses changes or a subchannel connectivity state changes), in which case there's already going to be churn.  In the case where there is no external churn, I don't think we want this kind of discontinuity.

The existing RR policy in gRPC already does this, so I think it's good to be consistent here -- and this code path will be executed only when we don't have any weights anyway.  But I'll follow up on whether or not this is worth doing performance-wise.

markdroth: After discussing this with the other languages, it looks like this isn't really necessary, so I've removed it.

PR #31914, opened 2022-12-16
xds_override_host LB: account for subchannel connectivity state

eugeneo:


src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -492,6 +553,25 @@ XdsOverrideHostLb::SubchannelWrapper::~SubchannelWrapper() {
   }
 }
 
+void XdsOverrideHostLb::SubchannelWrapper::WatchConnectivityState(
+    std::unique_ptr<ConnectivityStateWatcherInterface> watcher) {
+  auto watcher_id = watcher.get();
+  auto wrapper =
+      std::make_unique<ConnectivityStateWatcher>(std::move(watcher), Ref());
```
markdroth, line 560: It looks like this PR is adding support for tracking the subchannel's connectivity state but doesn't yet include the EDS health state.  Are you planning to add that in this PR or do that in a separate PR?

Regardless of whether that's part of this PR or not, it's worth noting that when we do add support for endpoints in EDS health state DRAINING, we will need to watch the subchannel's connectivity state even after the child policy has stopped monitoring it.  Given that, instead of wrapping each watcher from the child policy, I think we should start our own watcher on the wrapped subchannel as soon as it is created, and when the child policy starts a watch, we can just propagate notifications from our watcher to the child policy's watcher.  We should not need to start or stop our watcher when the child policy starts or stops its watcher.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -229,15 +266,49 @@ XdsOverrideHostLb::Picker::Picker(
...
+  MutexLock lock(&subchannel_map_mu_);
+  auto it = subchannel_map_.find(address);
+  if (it == subchannel_map_.end()) {
+    return absl::nullopt;
+  }
+  auto subchannel = it->second.GetSubchannel();
+  if (subchannel == nullptr) {
+    return absl::nullopt;
+  }
+  auto connectivity_state = subchannel->connectivity_state();
```
markdroth, line 292: This requires synchronization.  The connectivity state notifications that set this value will run in the control plane `WorkSerializer`, but this code that reads them will be run in the picker, which does not run in the `WorkSerializer`, so there can be a read and a write in different threads at the same moment.

The easiest thing would probably be to change the type of `SubchannelWrapper::connectivity_state_` to `std::atomic<grpc_connectivity_state>`.

src/core/BUILD
```diff
@@ -4460,6 +4460,7 @@ grpc_cc_library(
     language = "c++",
     deps = [
         "channel_args",
+        "grpc_lb_subchannel_list",
```
markdroth, line 4463: This dependency should not be needed.  Only leaf policies need to use the subchannel_list library.

test/core/client_channel/lb_policy/lb_policy_test_lib.h
```diff
@@ -702,6 +709,7 @@ class LoadBalancingPolicyTest : public ::testing::Test {
       absl::Span<const absl::string_view> addresses,
       const std::map<UniqueTypeName, std::string> call_attributes = {},
       size_t num_iterations = 3, SourceLocation location = SourceLocation()) {
+    ASSERT_NE(picker, nullptr);
```
markdroth, line 712: I don't think we need this check, since `GetCompletePicks()` handles this case for us.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -166,6 +166,80 @@ TEST_F(XdsOverrideHostTest, SubchannelsComeAndGo) {
...
+      "ipv4:127.0.0.1:441", "ipv4:127.0.0.1:442", "ipv4:127.0.0.1:443"};
+  auto picker = ExpectStartupWithRoundRobin(kAddresses);
+  ASSERT_NE(picker, nullptr);
+  // Check that the host is overridden
+  std::map<UniqueTypeName, std::string> pick_arg{
+      {XdsHostOverrideTypeName(), "127.0.0.1:442"}};
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  auto subchannel = FindSubchannel(kAddresses[1]);
+  ASSERT_NE(subchannel, nullptr);
+  subchannel->SetConnectivityState(GRPC_CHANNEL_TRANSIENT_FAILURE,
```
markdroth, line 181: A subchannel will not go from READY directly to TRANSIENT_FAILURE; instead, it will go into IDLE, then RR will ask it to reconnect, it will go into CONNECTING, and then if that connection attempt fails, it will go into TRANSIENT_FAILURE.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -166,6 +166,80 @@ TEST_F(XdsOverrideHostTest, SubchannelsComeAndGo) {
...
+  // Check that the host is overridden
+  std::map<UniqueTypeName, std::string> pick_arg{
+      {XdsHostOverrideTypeName(), "127.0.0.1:442"}};
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  auto subchannel = FindSubchannel(kAddresses[1]);
+  ASSERT_NE(subchannel, nullptr);
+  subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
+  ExpectReresolutionRequest();
+  subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+  ExpectState(GRPC_CHANNEL_READY);
```
markdroth, line 203: Whenever you call `ExpectState()`, you need to check the resulting picker behavior.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -166,6 +166,80 @@ TEST_F(XdsOverrideHostTest, SubchannelsComeAndGo) {
...
+  std::map<UniqueTypeName, std::string> pick_arg{
+      {XdsHostOverrideTypeName(), "127.0.0.1:442"}};
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  auto subchannel = FindSubchannel(kAddresses[1]);
+  ASSERT_NE(subchannel, nullptr);
+  subchannel->SetConnectivityState(GRPC_CHANNEL_TRANSIENT_FAILURE,
+                                   absl::ResourceExhaustedError("Hmmmm"));
+  ExpectReresolutionRequest();
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ExpectRoundRobinPicks(picker.get(), {kAddresses[0], kAddresses[2]});
```
markdroth, line 185: We need to pass the call attributes in here, or else we're not actually testing that the host-override gets ignored if the subchannel is in state TRANSIENT_FAILURE.  As currently written, this is testing that round_robin does not return picks for a subchannel in that state; it's not actually testing anything about the xds_override_host policy.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -166,6 +166,80 @@ TEST_F(XdsOverrideHostTest, SubchannelsComeAndGo) {
...
+  auto subchannel = FindSubchannel(kAddresses[1]);
+  ASSERT_NE(subchannel, nullptr);
+  subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
+  ExpectReresolutionRequest();
+  subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+  ExpectState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
+  EXPECT_EQ(ExpectPickComplete(picker.get()), kAddresses[0]);
+  ExpectPickQueued(picker.get(), pick_arg);
```
markdroth, line 207: Same comment as on line 185: We need to pass the call attributes to the picker here, or we're not testing the right thing.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -166,6 +166,80 @@ TEST_F(XdsOverrideHostTest, SubchannelsComeAndGo) {
...
+  auto subchannel = FindSubchannel(kAddresses[1]);
+  ASSERT_NE(subchannel, nullptr);
+  subchannel->SetConnectivityState(GRPC_CHANNEL_TRANSIENT_FAILURE,
+                                   absl::ResourceExhaustedError("Hmmmm"));
+  ExpectReresolutionRequest();
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ExpectRoundRobinPicks(picker.get(), {kAddresses[0], kAddresses[2]});
+}
+
+TEST_F(XdsOverrideHostTest, SubchannelConnectingIsQueued) {
```
markdroth, line 188: I see tests here for TRANSIENT_FAILURE and for CONNECTING.  We should also have a test for IDLE.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -166,6 +166,80 @@ TEST_F(XdsOverrideHostTest, SubchannelsComeAndGo) {
...
+  ExpectReresolutionRequest();
+  subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+  ExpectState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
+  EXPECT_EQ(ExpectPickComplete(picker.get()), kAddresses[0]);
+  ExpectPickQueued(picker.get(), pick_arg);
+}
+
+TEST_F(XdsOverrideHostTest, UsingRingHashAsChild) {
```
markdroth, line 210: Looking at this code has made me realize that we don't actually get any benefit from testing with ring_hash.  The main difference between round_robin and ring_hash is that ring_hash won't immediately trigger a reconnection request when the subchannel transitions from READY to IDLE, and we do want a test to verify that the xds_override_host policy will trigger the reconnection even if the child policy doesn't.  However, I don't think we need the ring_hash policy to test that case, because we will also need to do that in the case where the EDS health state is DRAINING, where the child policy does not even know about the subchannel in the first place.  So once we add support for EDS health states, we can test what we need even with round_robin.

I do think that we should ultimately have some tests with ring_hash as the child policy, because that is a case that we want to make sure will work, but I think we should do that in the e2e tests, not here in this unit test.  So I suggest removing this test.

Note that removing this test will also unblock this PR.  I still need to find a better approach to solve the `ExecCtx` problem here, but I probably won't get to that until after the holidays.  However, I would prefer not to merge this PR with the `ExecCtx` declared in the test code as it currently is (I suggested that as a quick hack to allow you to make progress, but I would prefer not to merge the code that way), so if we keep this test, that would delay this PR.

markdroth (commented):

Most of my comments are minor, but there are two significant issues, one about synchronization and the other about safe use of the `WorkSerializer`.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -229,15 +266,49 @@ XdsOverrideHostLb::Picker::Picker(
...
+  if (connectivity_state == GRPC_CHANNEL_READY) {
+    return PickResult::Complete(subchannel->wrapped_subchannel());
+  } else if (connectivity_state == GRPC_CHANNEL_CONNECTING) {
+    return PickResult::Queue();
+  } else if (connectivity_state == GRPC_CHANNEL_IDLE) {
+    work_serializer()->Run([&]() { subchannel->RequestConnection(); },
```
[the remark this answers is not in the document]

eugeneo (commented):
Thank you for the review. Addressed the comments and implemented support for subchannels in idle state.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -199,46 +206,35 @@ TEST_F(XdsOverrideHostTest, SubchannelConnectingIsQueued) {
   ASSERT_NE(subchannel, nullptr);
   subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
   ExpectReresolutionRequest();
```
markdroth, line 208: We should also do this here:

```
```

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -199,46 +206,35 @@ TEST_F(XdsOverrideHostTest, SubchannelConnectingIsQueued) {
...
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ExpectPickQueued(picker.get(), pick_arg);
```
markdroth, line 210: Now that you've added this, I think this test is actually covering both the IDLE and CONNECTING cases, so we don't need a separate test for IDLE below.

markdroth (commented):
This is really close!  All remaining issues are fairly minor.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Gone now!

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/lb_policy_test_lib.h
eugeneo: Gone.

eugeneo (commented):
Thank you for the review. I addressed the comments. Please take another look.

markdroth (approved):
This looks great!

PR #31916, opened 2022-12-16
CMake's better handling c++14 requirement

veblush:
As Cmake got a better way to specify that C++14 or above is required with [target_compile_features](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#id5), gRPC needs to use this instead of using `CMAKE_CXX_STANDARD` which should be done by the application. This is also aligned with [how Cloud C++ does](https://github.com/googleapis/google-cloud-cpp/blob/5a32dd780d87a2af5b5df6f014c37ba600f8a8fb/cmake/FindgRPC.cmake#L260).

In order for gRPC to get this change, the minium version of cmake should be bumped to 3.8, which is acceptable as OSS C++ policy needs 3.10 as the oldest version of cmake to support.

### DIFF
--- a/examples/android/helloworld/app/CMakeLists.txt
+++ b/examples/android/helloworld/app/CMakeLists.txt
@@ -1,10 +1,5 @@
 cmake_minimum_required(VERSION 3.8)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
-
-# https://github.com/abseil/abseil-cpp/issues/626
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DABSL_NO_XRAY_ATTRIBUTES=1")
-
 set(helloworld_PROTOBUF_PROTOC_EXECUTABLE "/usr/local/bin/protoc" CACHE STRING "Protoc binary on host")
 set(helloworld_GRPC_CPP_PLUGIN_EXECUTABLE "/usr/local/bin/grpc_cpp_plugin" CACHE STRING "gRPC CPP plugin binary on host")
 
--- a/examples/android/helloworld/app/CMakeLists.txt
+++ b/examples/android/helloworld/app/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4.1)
+cmake_minimum_required(VERSION 3.8)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
 
--- a/examples/cpp/cmake/common.cmake
+++ b/examples/cpp/cmake/common.cmake
@@ -19,10 +19,6 @@
 
 cmake_minimum_required(VERSION 3.8)
 
-if (NOT DEFINED CMAKE_CXX_STANDARD)
-  set (CMAKE_CXX_STANDARD 14)
-endif()
-
 if(MSVC)
   add_definitions(-D_WIN32_WINNT=0x600)
 endif()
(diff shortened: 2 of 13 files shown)


veblush:
CC @coryan

examples/android/helloworld/app/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4.1)
+cmake_minimum_required(VERSION 3.8)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
coryan, line 3: Drive-by: looks like yo could remove this?

examples/cpp/cmake/common.cmake
```diff
@@ -17,7 +17,7 @@
 # See cmake_externalproject/CMakeLists.txt for all-in-one cmake build
 # that automatically builds all the dependencies before building route_guide.
 
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.8)
 
 if (NOT DEFINED CMAKE_CXX_STANDARD)
   set (CMAKE_CXX_STANDARD 14)
```
coryan, line 23: Seems like you could remove this too?

src/android/test/interop/app/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4.1)
+cmake_minimum_required(VERSION 3.8)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
```
coryan, line 3: Maybe you can remove this?

examples/android/helloworld/app/CMakeLists.txt
veblush: Is this because gRPC cmakefile has this already?

coryan: Yes, any target that depends on gRPC++ will inherit the requirement to use C++ 14 or higher.

examples/cpp/cmake/common.cmake
veblush: Done.

src/android/test/interop/app/CMakeLists.txt
veblush: Done.

veblush:
@drfloob AJ, would you review this as Jan is OOO?

tools/run_tests/run_tests.py
```diff
@@ -477,7 +477,7 @@ def _compiler_options(self, use_docker, compiler):
             _check_compiler(compiler, ['default', 'cmake'])
 
         if compiler == 'default' or compiler == 'cmake':
-            return ('debian11', [])
+            return ('debian11', ["-DCMAKE_CXX_STANDARD=14"])
```
jtattermusch, line 480: It feels odd that to our "default" build via run_tests.py we need to pass extra options.
Would that mean that  if one builds with cmake by hand, without passing any extra configuration, they would get a differently configured build that if they run run_tests.py? If so, that feels wrong.

jtattermusch (commented):
Overall looks fine, but I had one concern - see comment.

tools/run_tests/run_tests.py
veblush: This is required because of Apple clang which uses C++98 by default (yay). Also this is only for tests and gRPC library targets don't specify it so that C++ version can be picked by applications.

drfloob (approved):

PR #31917, opened 2022-12-17
Update all Debian:9 to Debian:10

veblush:
Debian:9 (stretch) reached EOL on  June 30, 2022 ([ref](https://wiki.debian.org/LTS)) so all docker images based on Debian:9 switch to Debian:10. 
This is prerequisite to https://github.com/grpc/grpc/pull/31916.


tools/dockerfile/test/android_ndk/Dockerfile
```diff
@@ -88,8 +88,12 @@ RUN apt-get update && apt-get install -y cmake && apt-get clean
 RUN mkdir /var/local/jenkins
 
 
-# Java required by Android SDK
-RUN apt-get update && apt-get -y install openjdk-8-jdk && apt-get clean
+# Java required by Android SDK (using Eclipse Temurin Package)
+RUN apt install -y wget apt-transport-https && \
```
jtattermusch, line 92: never heard of "temurin-8-jdk", why can't you simply install openjdk?
Debian 10 has "openjdk-11-jdk" and I'd prefer to avoid anything non-standard, since it makes things more confusing and harder to maintain.

jtattermusch (approved):
LGTM with a comment.

tools/dockerfile/test/android_ndk/Dockerfile
veblush: That's a sort of next official java 8 sdk (https://blog.adoptopenjdk.net/2021/03/transition-to-eclipse-an-update/). I wanted to keep everything else other than Debian upgrade here but I can try Jdk 11 here to see whether it works with Android SDK.

veblush: Oh it turned out that Android SDK this image uses does require Java 8 SDK. So this should be like this until we upgrade SDK.

veblush:
I'm going to get this merged as Abseil upgrade needs this change. I'll revisit Android SDK issue later.

PR #31928, opened 2022-12-18
[EventEngine] Modify iomgr to allow creation of posix event engine listeners and server side endpoints

Vignesh2208:


src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -332,7 +332,7 @@ PosixEventEngine::PosixEventEngine(PosixEventPoller* poller)
     : connection_shards_(std::max(2 * gpr_cpu_num_cores(), 1u)),
       executor_(std::make_shared<ThreadPool>()),
       timer_manager_(executor_) {
-  if (UseEventEngineClient()) {
+  if (UseEventEngineClient() || UseEventEngineServer()) {
```
ctiller, line 335: This would read better if we had a `bool NeedPosixEngine() { return UseClient() || UseServer(); }`
so you can write `!NeedPosixEngine()` below too

Vignesh2208: Good point. Fixed.

src/core/lib/iomgr/event_engine_shims/endpoint.h
```diff
@@ -28,6 +28,15 @@ namespace experimental {
 grpc_endpoint* grpc_event_engine_endpoint_create(
     std::unique_ptr<EventEngine::Endpoint> ee_endpoint);
 
+/// Returns true if the passed endpoint is an event engine shim endpoint.
+bool grpc_is_event_engine_endpoint(grpc_endpoint* ep);
+
+/// Destroys the passed in event engine shim endpoint and schedules the
+/// asynchronous execution of the on_release_fd callback. The int pointer fd is
+/// set to the enderlying endpoint's file descriptor.
```
krishnacx, line 36: enderlying -> underlying

krishnacx (commented):
Minor typos

src/core/lib/experiments/experiments.h
```diff
@@ -40,14 +40,15 @@ inline bool IsEventEngineClientEnabled() { return IsExperimentEnabled(7); }
 inline bool IsMonitoringExperimentEnabled() { return IsExperimentEnabled(8); }
 inline bool IsPromiseBasedClientCallEnabled() { return IsExperimentEnabled(9); }
 inline bool IsFreeLargeAllocatorEnabled() { return IsExperimentEnabled(10); }
+inline bool IsEventEngineServerEnabled() { return IsExperimentEnabled(11); }
```
drfloob, line 43: nit: suggest `IsEventEngineListenerEnabled`, for consistency. `s/Server/Listener/g` when referring to EventEngine objects.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
```diff
@@ -128,6 +128,12 @@ class EventEngineEndpointWrapper {
   void ShutdownUnref() {
     if (shutdown_ref_.fetch_sub(1, std::memory_order_acq_rel) ==
         kShutdownBit + 1) {
+#ifdef GRPC_POSIX_SOCKET_TCP
```
drfloob, line 131: This is going to be problematic on Posix-compatible platforms that aren't using an engine that implements the `PosixEndpointWithFdSupport` interface (we support this use case). We probably can't do RTTI since it might not be enabled everywhere, so we may want to invent some other means of detecting implementations at runtime. Let's brainstorm this a bit.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
```diff
@@ -171,6 +189,9 @@ class EventEngineEndpointWrapper {
   std::unique_ptr<grpc_event_engine_endpoint> eeep_;
   std::atomic<int64_t> refs_{1};
   std::atomic<int64_t> shutdown_ref_{1};
+#ifdef GRPC_POSIX_SOCKET_TCP
+  absl::AnyInvocable<void(absl::StatusOr<int>)> on_release_fd_ = nullptr;
```
drfloob, line 193: `= nullptr` is unnecessary with AnyInvocable.

src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -105,6 +136,100 @@ static grpc_error_handle tcp_server_create(grpc_closure* shutdown_complete,
   s->memory_quota = s->options.resource_quota->memory_quota();
   s->pre_allocated_fd = -1;
   gpr_atm_no_barrier_store(&s->next_pollset_to_assign, 0);
+  s->n_bind_ports = 0;
+  new (&s->listen_fd_to_index_map)
+      absl::flat_hash_map<int, std::tuple<int, int>>();
+  if (grpc_event_engine::experimental::UseEventEngineServer()) {
+    PosixEventEngineWithFdSupport::PosixAcceptCallback accept_cb =
```
drfloob, line 143: I'd suggest moving these 100 lines of code to their own EventEngine instantiation function.

src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -105,6 +136,100 @@ static grpc_error_handle tcp_server_create(grpc_closure* shutdown_complete,
...
+                          grpc_event_engine::experimental::
+                              grpc_event_engine_endpoint_create(std::move(ep)),
+                          read_notifier_pollset, acceptor);
+        };
+    auto on_shutdown_complete_cb =
+        grpc_event_engine::experimental::GrpcClosureToStatusCallback(
+            shutdown_complete);
+    auto listener =
+        reinterpret_cast<PosixEventEngineWithFdSupport*>(
+            grpc_event_engine::experimental::GetDefaultEventEngine().get())
```
drfloob, line 214: I do wonder if we'll see lifetime issues here, listeners outliving the default EventEngine instance, but we _shouldn't_ see it if this contract can be maintained in gRPC core.

https://github.com/grpc/grpc/blob/1c4b138322bbe5f6091092013b318aece2d3a55a/include/grpc/event_engine/event_engine.h#L360-L367

src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -530,6 +709,12 @@ static void tcp_server_start(grpc_tcp_server* s,
   GPR_ASSERT(s->on_accept_cb);
   GPR_ASSERT(s->active_ports == 0);
   s->pollsets = pollsets;
+  if (grpc_event_engine::experimental::UseEventEngineServer()) {
+    GPR_ASSERT(!s->shutdown_listeners);
+    (void)s->ee_listener->Start();
```
drfloob, line 714: We should not bury this error, it may be meaningful when other posix-compatible EventEngines are used.

src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -611,6 +799,17 @@ class ExternalConnectionHandler : public grpc_core::TcpServerFdHandler {
...
   // TODO(yangg) resolve duplicate code with on_read
   void Handle(int listener_fd, int fd, grpc_byte_buffer* buf) override {
+    if (grpc_event_engine::experimental::UseEventEngineServer()) {
+      grpc_event_engine::experimental::SliceBuffer pending_data;
+      if (buf != nullptr) {
+        pending_data =
+            grpc_event_engine::experimental::SliceBuffer::TakeCSliceBuffer(
+                buf->data.raw.slice_buffer);
+      }
+      (void)s_->ee_listener->HandleExternalConnection(listener_fd, fd,
```
drfloob, line 809: Why are we burying this potential error?

drfloob (commented):

src/core/lib/iomgr/tcp_server_posix.cc
Vignesh2208: I dont think this would be an issue with the default event engine at least. The listeners take a shared ptr to the event engine, so the event engine is alive as long as the listener is alive.

src/core/lib/event_engine/shim.cc
```diff
@@ -32,5 +32,15 @@ bool UseEventEngineClient() {
 #endif
 }
 
+bool UseEventEngineServer() {
+// TODO(hork, eryu): Adjust the ifdefs accordingly when event engine's become
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

src/core/lib/iomgr/event_engine_shims/endpoint.h
Vignesh2208: Fixed.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -463,11 +463,14 @@ void PosixEndpointImpl::PerformReclamation() {
...
+    Ref().release();
     memory_owner_.PostReclaimer(
         grpc_core::ReclamationPass::kBenign,
         [this](absl::optional<grpc_core::ReclamationSweep> sweep) {
-          if (!sweep.has_value()) return;
-          PerformReclamation();
+          if (sweep.has_value()) {
+            PerformReclamation();
+          }
+          Unref();
```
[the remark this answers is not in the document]
Vignesh2208: Good point. Fixed.

src/core/lib/experiments/experiments.yaml
```diff
@@ -119,4 +119,11 @@
   expiry: 2023/04/01
   owner: alishananda@google.com
   test_tags: [resource_quota_test]
+- name: event_engine_server
```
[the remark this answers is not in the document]
Vignesh2208: Fixed.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
Vignesh2208: Removed.

src/core/lib/iomgr/tcp_server_posix.cc
Vignesh2208: Moved it to a new function now.

src/core/lib/iomgr/tcp_server_posix.cc
Vignesh2208: Logged this error now.

src/core/lib/iomgr/tcp_server_posix.cc
Vignesh2208: Logged this error now.

src/core/lib/iomgr/event_engine_shims/endpoint.cc
Vignesh2208: Per offline discussion, I'll create a separate PR to support determining whether the EE supports a specific interface. It might be cleaner in-case we have to rollback either PR.

drfloob (approved):

PR #31930, opened 2022-12-18
Run multi arch ios unit test for M1 mac

HannahShiSFB:
1. pass `--ios_multi_cpus=x86_64,sim_arm64` to bazel to build multi-arch binary
2. add `ios_sim_arm64` config for `third_party/cares`
3. update `bazel-out/darwin-fastbuild/bin` to `bazel-bin` which points to `bazel-out/darwin_arm64-fastbuild/bin` on M1 macs.

tested with `SCHEME=InteropTestsLocalSSL src/objective-c/tests/run_one_test_bazel.sh`

---
Example failure:
```
2022-12-18 12:29:08.229 xctest[86573:242977] The bundle “InteropTestsLocalCleartext” couldn’t be loaded because it doesn’t contain a version for the current architecture. Try installing a universal version of the bundle.
```

@sampajano

### DIFF
--- a/test/distrib/bazel/python/namespaced/upper/example/BUILD
+++ b/test/distrib/bazel/python/namespaced/upper/example/BUILD
@@ -112,16 +112,16 @@ py_grpc_library(
 # in the following directory layout for python, which needs to properly be added to the imports.
 #
 # No Import or Strip (Can't compile if there are any proto dependencies)
-# bazel-out/darwin-fastbuild/bin/namespaced/upper/example/namespaced_example_pb2.py
+# bazel-bin/namespaced/upper/example/namespaced_example_pb2.py
 #
 # No import Prefix (Can't compile if there are any proto dependencies)
-# bazel-out/darwin-fastbuild/bin/namespaced/upper/example/_virtual_imports/namespaced_example_proto/namespaced_example_pb2.py
+# bazel-bin/namespaced/upper/example/_virtual_imports/namespaced_example_proto/namespaced_example_pb2.py
 #
 # No strip prefix (Can't compile if there are any proto dependencies)
-# bazel-out/darwin-fastbuild/bin/namespaced/upper/example/_virtual_imports/namespaced_example_proto/upper/example/namespaced/upper/example/namespaced_example_pb2.py
+# bazel-bin/namespaced/upper/example/_virtual_imports/namespaced_example_proto/upper/example/namespaced/upper/example/namespaced_example_pb2.py
 #
 # Both Import and Strip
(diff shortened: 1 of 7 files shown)


third_party/cares/cares.BUILD
```diff
@@ -106,6 +111,7 @@ copy_file(
         ":ios_armv7": "@com_github_grpc_grpc//third_party/cares:config_darwin/ares_config.h",
         ":ios_armv7s": "@com_github_grpc_grpc//third_party/cares:config_darwin/ares_config.h",
         ":ios_arm64": "@com_github_grpc_grpc//third_party/cares:config_darwin/ares_config.h",
```
sampajano, line 113: just curious - what's the difference between ios_sim_arm64 and ios_arm64? can the latter be used?

sampajano (approved):
Thanks so much for adding M1 support! 😃

test/distrib/bazel/python/namespaced/upper/example/BUILD
```diff
@@ -112,16 +112,16 @@ py_grpc_library(
 # in the following directory layout for python, which needs to properly be added to the imports.
 #
 # No Import or Strip (Can't compile if there are any proto dependencies)
-# bazel-out/darwin-fastbuild/bin/namespaced/upper/example/namespaced_example_pb2.py
```
sampajano, line 115: I'm not familiar with bazel/python. Do we need another reviewer or this should be ok?

sampajano:
> Thanks so much for adding M1 support! 😃

There seems to be some test failure tho.

third_party/cares/cares.BUILD
HannahShiSFB: I believe ios_sim_arm64 is for simulator on M1 and ios_arm64 is for device.
I tried running with ios_arm64 but failed with `incompatible platform (have 'iOS', need 'iOS-sim')`

sampajano: Ah thanks for explaining :)

sampajano:
> > Thanks so much for adding M1 support! 😃
> 
> There seems to be some test failure tho.

@HannahShiSFB is the Bazel Distribtest failure related to this PR? If not can we usually merge irrespective of the error?

HannahShiSFB:
Bazel Distribtest failure is not related, fix in #31999

sampajano:
> Bazel Distribtest failure is not related, fix in #31999. It's good to merge

Aha ok thanks for clarifying! Merging now. Thanks! 😃

Issue #27066, opened 2021-08-19
C# RpcException has very verbose exception.Message property


timmydo:
### Is your feature request related to a problem? Please describe.
I'm always frustrated when I catch a generic exception which is actually a RpcException and print ex.Message. It fills up my console with 100 lines of stack trace

### Describe the solution you'd like
I'd like to just see the error message without the stack trace.

### Describe alternatives you've considered
Catching RpcException specifically and printing the status code is possible but annoying to have to do.

### Additional context
I think the underlying issue is that Status.ToString() is used in the RpcException ctor for base.message--Status.ToString() prints stack traces. Maybe Status could have a .Message which doesn't print the stack.

Thanks!!!

jtattermusch:
@tonydnewell  can you please take a look?

jtattermusch:
possibly related: https://github.com/grpc/grpc/issues/25740

tonydnewell:
@jtattermusch would changing the format of the status string be a breaking change? Would anyone be parsing it?

jtattermusch:
> @jtattermusch would changing the format of the status string be a breaking change? Would anyone be parsing it?

I think changing the contents of status.ToString() would be fine. It is not a breaking change.

tonydnewell:
I've made a change in PR https://github.com/grpc/grpc/pull/31932 to remove the stack trace when adding the `DebugException` information to `Status.ToString()`

However I've not been able to reproduce the original reported problem with "100 lines of stack trace" since the only `DebugException` that I can see added is that in https://github.com/grpc/grpc/blob/af855eb64eea02d2f7b68d49c3d4d7a263649104/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs#L97 which doesn't contain a stack trace since it wasn't thrown.

jtattermusch:
Fixed by https://github.com/grpc/grpc-dotnet/pull/1992, the fix will be in the next release of Grpc.Core.Api (the upcoming 2.52.x). The fix will work both for grpc-dotnet and Grpc.Core, but for Grpc.Core, you need to explicitly use Grpc.Core.Api 2.52+ (once it's released).

jtattermusch:
We can potentially backport the change into 2.46.x as well, but for now there seems to be no reason to do so since Grpc.Core 2.46.x  will work just fine with more recent versions of Grpc.Core.Api.

Issue #25740, opened 2021-03-17
[C#] suggest populating `.InnerException` on `RpcException`


mgravell:
Currently, underlying exceptions behind `RpcException` are exposed in a non-idiomatic way, via the optional `.Status.DebugException`. This is very unusual, and means:

- no standard logging etc tooling will know to look for it
- most developers won't find it

Proposal: automatically populate `.InnerException` (the optional, expected API), when possible. This should be as simple as:

``` c#
- public RpcException(Status status, Metadata trailers, string message) : base(message)
+ public RpcException(Status status, Metadata trailers, string message) : base(message, status.DebugException)
```

(I'm happy to submit a PR)

mgravell:
@jtattermusch this come up again here: https://github.com/protobuf-net/protobuf-net.Grpc/issues/196 - would a PR be considered here?

jtattermusch:
I think this is something we can consider.

My main concern is that we must make it clear that RpcException.InnerException is only a mechanism to help with debugging, and the authoritative info about the RPC results is always the `Status` (along with the StatusCode). The InnerException can never traverse from the remote peer (as gRPC protocol has no way of serializing the exception objects and it also wouldn't be safe to do so). So InnerException is only useful when there is a "local" problem (e.g. cannot create a connection, failed to serialize/deserialize locally etc.).

I'd also like @JamesNK  to asses the impact of having RpcException.InnerException carry debug info sometimes from the grpc-dotnet perspective (I know grpc-dotnet makes use of `Status.DebugException` currently)

We'd need to implement this in a backward-compatible way, but that seems doable.

jtattermusch:
Fixed by https://github.com/grpc/grpc-dotnet/pull/1992, the fix will be in the next release of Grpc.Core.Api (the upcoming 2.52.x). The fix will work both for grpc-dotnet and Grpc.Core, but for Grpc.Core, you need to explicitly use Grpc.Core.Api 2.52+ (once it's released).

jtattermusch:
We can potentially backport the change into 2.46.x as well, but for now there seems to be no reason to do so since Grpc.Core 2.46.x  will work just fine with more recent versions of Grpc.Core.Api.

PR #31932, opened 2022-12-19
Remove stack trace from Status.ToString

tonydnewell:
Fix for https://github.com/grpc/grpc/issues/27066 - remove the stack trace from `Status.ToString()` when getting the info from the `DebugException`

However I've not been able to reproduce the original reported problem with "100 lines of stack trace" since the only `DebugException` that I can see added is that in
https://github.com/grpc/grpc/blob/af855eb64eea02d2f7b68d49c3d4d7a263649104/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs#L97

which doesn't contain a stack trace since it wasn't thrown.

### DIFF
--- a/src/csharp/Grpc.Core.Api/Status.cs
+++ b/src/csharp/Grpc.Core.Api/Status.cs
@@ -87,7 +87,8 @@ namespace Grpc.Core
         {
             if (DebugException != null)
             {
-                return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\", DebugException=\"{DebugException}\")";
+                return $"Status(StatusCode=\"{StatusCode}\", Detail=\"{Detail}\"," +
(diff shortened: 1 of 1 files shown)


tonydnewell:
@jtattermusch for review

jtattermusch:
I think the proposed fix is very reasonable, thanks! 
We should probably make the change against grpc-dotnet repository though, since that's where Grpc.Core.Api now lives.
I created https://github.com/grpc/grpc-dotnet/pull/1992

tonydnewell:
Updated to use FullName for exception

jtattermusch:
Superseded by https://github.com/grpc/grpc-dotnet/pull/1992 (but thanks for coming up with the initial fix, the PR is a variation of it, with some additions on top).
For now there's no need to backport to v1.46.x branch (see comment https://github.com/grpc/grpc/issues/25740#issuecomment-1359986756).

PR #31936, opened 2022-12-20
testing: bump timeout of grpc/abseil-head build job from 4 to 5 hours

apolcyn:
This test has recently timed out 4 times in a row


jtattermusch:
You're right. Looks like the job has been gradually getting slower and recently reached the 240min threshold
image

I think this is a result of more C/C++ tests being added over time.

jtattermusch (approved):

PR #16660, opened 2018-09-20
Rename the BoringSSL framework to openssl_grpc to avoid conflict with OpenSSL

muxi:
We renamed all the BoringSSL symbols but did not rename the framework or library's name. If a user write `#include <openssl/ssl.h>` they are still referring to the header of the internal version of BoringSSL.

cc: @srini100

PR #16358, opened 2018-08-15
Make symbols of BoringSSL private

muxi:
The change is to add prefix to BoringSSL symbols in gRPC Objective-C to avoid conflicting with OpenSSL for apps that depend on both SSL libraries (see detailed discussions in #6111). This PR makes gRPC to prefix its own BoringSSL symbols with `GRPC_SHADOW_` to avoid conflict with OpenSSL.

PR #31938, opened 2022-12-20
Update third_party/boringssl-with-bazel

jtattermusch:

Alternative to https://github.com/grpc/grpc/pull/31869

TODO: figure out the breakages caused by the update (python, objc, BinderTransport etc).


jtattermusch:
Python issue:
https://source.cloud.google.com/results/invocations/2f6a7e62-5bd6-4e44-a76d-0b8f6fad9cbd/targets/github%2Fgrpc%2Ftoplevel_run_tests_invocations%2Frun_tests_python_macos_opt_native/tests
```
Processing /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_python_macos_opt_native
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_python_macos_opt_native/py38/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_python_macos_opt_native/setup.py'"'"'; __file__='"'"'/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_python_macos_opt_native/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/1d/z50_0t6s6fsb1yj3_svwklk80000gn/T/pip-pip-egg-info-giy4ed6x
       cwd: /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_python_macos_opt_native/
  Complete output (5 lines):
    File "<string>", line 1, in <module>
    File "/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_python_macos_opt_native/setup.py", line 383, in <module>
      asm_files = grpc_core_dependencies.ASM_SOURCE_FILES[asm_key]
  KeyError: 'crypto_mac_x86_64'
  ----------------------------------------
```

ObjC issue:
https://source.cloud.google.com/results/invocations/af9e249e-6647-4c62-a0a5-d03d8d971fd8/targets/github%2Fgrpc%2Frun_tests%2Fobjc_macos_opt_native%2Fios-cpp-test-cronet/tests
```
** TEST FAILED **


The following build commands failed:
	CompileC /Users/kbuilder/Library/Developer/Xcode/DerivedData/Tests-cjuivpsqnidywdbibpzldwalcfuu/Build/Intermediates.noindex/Pods.build/Cronet-iphonesimulator/BoringSSL-GRPC.build/Objects-normal/x86_64/x509_vfy.o /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/BoringSSL-GRPC/src/crypto/x509/x509_vfy.c normal x86_64 c com.apple.compilers.llvm.clang.1_0.compiler (in target 'BoringSSL-GRPC' from project 'Pods')
(1 failure)
tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/BoringSSL-GRPC/src/crypto/x509/x509_vfy.c -o /Users/kbuilder/Library/Developer/Xcode/DerivedData/Tests-cjuivpsqnidywdbibpzldwalcfuu/Build/Intermediates.noindex/Pods.build/Cronet-iphonesimulator/BoringSSL-GRPC.build/Objects-normal/x86_64/x509_vfy.o
/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/BoringSSL-GRPC/src/crypto/x509/x509_vfy.c:1884:47: error: conflicting types for 'time'
  int64_t compare_time = (cmp_time == NULL) ? time(NULL) : *cmp_time;
                                              ^
In module 'Darwin' imported from /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/BoringSSL-GRPC/src/crypto/x509/x509_vfy.c:57:
/Applications/Xcode_14.1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator16.1.sdk/usr/include/time.h:118:8: note: previous declaration is here
time_t time(time_t *);
       ^
/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/BoringSSL-GRPC/src/crypto/x509/x509_vfy.c:1904:5: error: conflicting types for 'time'
    time(&t);
    ^
In module 'Darwin' imported from /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/BoringSSL-GRPC/src/crypto/x509/x509_vfy.c:57:
/Applications/Xcode_14.1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator16.1.sdk/usr/include/time.h:118:8: note: previous declaration is here
time_t time(time_t *);
       ^
2 errors generated.
```

davidben:
> KeyError: 'crypto_mac_x86_64'

Oh, this is probably from https://boringssl-review.googlesource.com/c/boringssl/+/51307. We cleaned that up because it was confusing for Arm Macs to use crypto_ios_aarch64 when the assembly was actually targeting all Apple platforms. (Conversely, it was confusing for the iOS simulator to use crypto_mac_x86_64.)

It's probably not so useful when it's a whole year of changes (too many to scroll through), but we try to call out potentially interesting changes with "Update-Note" lines in the CL description. Many of our update scripts dump something like `git log --grep=Update-Note old..new`.

I think you need to update the `asm_key` computation to use "crypto_apple_x86_64" instead of "crypto_mac_x86_64". (Separately, you might want to enable the assembly for Arm Mac.)

> error: conflicting types for 'time'

Hmm, not sure what's going on here. https://github.com/grpc/grpc/pull/31869#issue-1489094856 mentioned something about matching header file names, but that's an odd error to get in this context. Unless perhaps it's doing an implicit prototype (without warning??) and then some later step notices the mismatch.

If it is the matching header names, that's a bit odd. Is it possible BoringSSL's `.../include/openssl/` is getting picked up in the include path somehow? We only need/want `.../include/` in there.

davidben:
> If it is the matching header names, that's a bit odd. Is it possible BoringSSL's .../include/openssl/ is getting picked up in the include path somehow? We only need/want .../include/ in there.

Looking at the flags in the build log, a few things stand out to me:
```
 -I/Users/kbuilder/Library/Developer/Xcode/DerivedData/Tests-cjuivpsqnidywdbibpzldwalcfuu/Build/Products/Cronet-iphonesimulator/BoringSSL-GRPC/include
 -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Private
 -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Private/BoringSSL-GRPC
 -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Public
 -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Public/BoringSSL-GRPC
```
(I sorted them to get all the `-I` values. It's possible they're not in that order.)

It seems there's some weird thing going on with the BoringSSL-GRPC "pod"? That's this file, right?
https://github.com/grpc/grpc/blob/master/src/objective-c/BoringSSL-GRPC.podspec

I don't quite understand what's going on here, but it looks like the podspec has to play a lot of strange games with headers as part of some Apple-specific modules system? Is it possible one of these is resulting in BoringSSL's headers getting added directly to the include path?
https://github.com/grpc/grpc/blob/master/src/objective-c/BoringSSL-GRPC.podspec#L94

Perhaps someone familiar with podspecs can help fix this?

PS: After https://boringssl-review.googlesource.com/c/boringssl/+/50849, you don't need to wrap the arm_arch.h include here anymore:
https://github.com/grpc/grpc/blob/master/src/objective-c/BoringSSL-GRPC.podspec#L164

jtattermusch:
> > If it is the matching header names, that's a bit odd. Is it possible BoringSSL's .../include/openssl/ is getting picked up in the include path somehow? We only need/want .../include/ in there.
> 
> Looking at the flags in the build log, a few things stand out to me:
> 
> ```
>  -I/Users/kbuilder/Library/Developer/Xcode/DerivedData/Tests-cjuivpsqnidywdbibpzldwalcfuu/Build/Products/Cronet-iphonesimulator/BoringSSL-GRPC/include
>  -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Private
>  -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Private/BoringSSL-GRPC
>  -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Public
>  -I/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/test/cpp/ios/Pods/Headers/Public/BoringSSL-GRPC
> ```
> 
> (I sorted them to get all the `-I` values. It's possible they're not in that order.)
> 
> It seems there's some weird thing going on with the BoringSSL-GRPC "pod"? That's this file, right? https://github.com/grpc/grpc/blob/master/src/objective-c/BoringSSL-GRPC.podspec
> 
> I don't quite understand what's going on here, but it looks like the podspec has to play a lot of strange games with headers as part of some Apple-specific modules system? Is it possible one of these is resulting in BoringSSL's headers getting added directly to the include path? https://github.com/grpc/grpc/blob/master/src/objective-c/BoringSSL-GRPC.podspec#L94
> 
> Perhaps someone familiar with podspecs can help fix this?
> 
> PS: After https://boringssl-review.googlesource.com/c/boringssl/+/50849, you don't need to wrap the arm_arch.h include here anymore: https://github.com/grpc/grpc/blob/master/src/objective-c/BoringSSL-GRPC.podspec#L164

@sampajano can you help here?

sampajano:
> @sampajano can you help here?

@jtattermusch Thanks for looping me in.

I took a quick look at the issue but i don't really have any previous familiarity or expertise with Podspec files and the header conflict.

@HannahShiSFB - Since you've been working with the boring SSL upgrade and has seen the header conflict, i wonder if you have any extra insight on what's happening here? Please lemme know if you need me to take a look together with you to dig into it. Thanks!

davidben:
Is there someone on gRPC who does understand those files? You all evidently set them up at some point. :-)

Right now, all indications are that gRPC have somehow misconfigured the build and it only happened to be noticed now. If this is some fundamental constraint in cocoapods, we can try to figure something out, but someone needs to understand it before it can be declared a fundamental constraint.

HannahShiSFB:
My understanding is to make boringssl be able to coexist with openssl (#16358, #16660, etc.), follow things were added in BoringSSL-GRPC.podspec

1. flatten header directory [`src/include/openssl`](https://github.com/grpc/grpc/blob/v1.51.1/src/objective-c/BoringSSL-GRPC.podspec#L98) then put them under ['openssl_grpc'](https://github.com/grpc/grpc/blob/v1.51.1/src/objective-c/BoringSSL-GRPC.podspec#L114) 
2. replace all [`#include <openssl/xxx>` to `#include<openssl_grpc/xxx`](https://github.com/grpc/grpc/blob/v1.51.1/src/objective-c/BoringSSL-GRPC.podspec#L721)
3. include [`boringssl_prefix_symbols.h`](https://github.com/grpc/grpc/blob/v1.51.1/src/objective-c/BoringSSL-GRPC.podspec#L725) which renames all openssl exported symbols.

I tried disabled all the above 3 but the error remains, the build only succeeded with removing "openssl/time.h"

test code in #31989

davidben:
(1-3) are odd but I can't think of a reason why they'd cause problems. I don't understand modulemaps (BTW, the special cases around `arm_arch.h` should no longer be necessary after the upgrade), so I don't know if there's anything inherently wrong here.

> I tried disabled all the above 4 but the error remains, the build only succeeded with removing "openssl/time.h" file and the includes.

Hmm, if all of those four still preserve the error, it must be something else that Cocoapods is doing.

In principle, `openssl/time.h` should not be causing a problem, because `<time.h>` is not `<openssl/time.h>` and shouldn't pick up ours. My guess is, for some reason, the some directory with our public headers is getting added directly to the include path, and shouldn't be. Or perhaps the modulemap is doing something weird, because if `<time.h>` were _entirely_ not being included, there should be more errors.

Probably the next thing to look into is how modulemaps and Cocoapods works.

davidben:
Another thing to look at is what are these "hmap" files that are being used here, like BoringSSL-GRPC-project-headers.hmap. Those might be contributing to the messed up includes?

I've been playing around with -fmodules locally but so far have been unable to reproduce that error. I'm also very confused about which is the second `time` definition that's conflicting. If we're getting the old C implicit declaration, we should have gotten a ton more errors.

HannahShiSFB:
> for some reason, the some directory with our public headers is getting added directly to the include path, and shouldn't be

This is most likely what happened, I tried to remove some of the `openssl/` directory from the includes, it still builds. I don't know how to preserve the directory structure here.
https://github.com/CocoaPods/CocoaPods/issues/8873 seems related (but on the sources IIUC), and the recommendations of adding subspecs didn't work for me.

> if <time.h> were entirely not being included, there should be more errors

There were more errors if you comment out that in `x509_vfy.c`

davidben:
What happens if you add something like this line:
https://github.com/protocolbuffers/protobuf/blob/main/Protobuf-C++.podspec#L35

I don't understand what's going on with headermaps, but between that and https://openradar.appspot.com/FB8115994, it sounds like they cause a lot of problems.

davidben:
https://stackoverflow.com/a/7334072 suggests `HEADERMAP_INCLUDES_FLAT_ENTRIES_FOR_TARGET_BEING_BUILT=NO` might also work and be less of a heavyweight operation? https://stackoverflow.com/a/63631814 also has the names of some other options.

But it may be simpler to just disable it given how problematic this feature is. BoringSSL's small, and a leaf node in the dependency tree, so I doubt this is a particularly useful build optimization.

HannahShiSFB:
Yeah, add `'USE_HEADERMAP' => 'NO'` works

patch to the podspec:
```
diff --git a/src/objective-c/BoringSSL-GRPC.podspec b/src/objective-c/BoringSSL-GRPC.podspec
index 529dacc37b..8c3e5e1c6e 100644
--- a/src/objective-c/BoringSSL-GRPC.podspec
+++ b/src/objective-c/BoringSSL-GRPC.podspec
@@ -151,6 +151,11 @@ Pod::Spec.new do |s|
     ss.dependency "#{s.name}/Interface", version
   end

+  s.pod_target_xcconfig = {
+    # Do not let src/include/openssl/time.h override system API
+    'USE_HEADERMAP' => 'NO',
+  }
+
   s.prepare_command = <<-END_OF_COMMAND
     set -e
     # Add a module map and an umbrella header
```

jtattermusch:
> Yeah, add `'USE_HEADERMAP' => 'NO'` works
> 
> patch to the podspec:
> 
> ```
> diff --git a/src/objective-c/BoringSSL-GRPC.podspec b/src/objective-c/BoringSSL-GRPC.podspec
> index 529dacc37b..8c3e5e1c6e 100644
> --- a/src/objective-c/BoringSSL-GRPC.podspec
> +++ b/src/objective-c/BoringSSL-GRPC.podspec
> @@ -151,6 +151,11 @@ Pod::Spec.new do |s|
>      ss.dependency "#{s.name}/Interface", version
>    end
> 
> +  s.pod_target_xcconfig = {
> +    # Do not let src/include/openssl/time.h override system API
> +    'USE_HEADERMAP' => 'NO',
> +  }
> +
>    s.prepare_command = <<-END_OF_COMMAND
>      set -e
>      # Add a module map and an umbrella header
> ```

Thanks, I added the patch to this PR, let's see what the test say.

jtattermusch:
One more issue to resolve is the "Bazel BinderTransport APK building/testing" test:
https://source.cloud.google.com/results/invocations/17247ee8-2b0c-4573-a1c6-57130ee430e3/log

```
external/androidndk/ndk/toolchains/x86_64-4.9/prebuilt/linux-x86_64/lib/gcc/x86_64-linux-android/4.9.x/../../../../x86_64-linux-android/bin/ld: error: cannot find -lpthread
clang: error: linker command failed with exit code 1 (use -v to see invocation)
INFO: Elapsed time: 481.943s, Critical Path: 42.15s
INFO: 2577 processes: 155 internal, 2410 processwrapper-sandbox, 12 worker.
FAILED: Build did NOT complete successfully
```

jtattermusch:
> One more issue to resolve is the "Bazel BinderTransport APK building/testing" test: https://source.cloud.google.com/results/invocations/17247ee8-2b0c-4573-a1c6-57130ee430e3/log
> 
> ```
> external/androidndk/ndk/toolchains/x86_64-4.9/prebuilt/linux-x86_64/lib/gcc/x86_64-linux-android/4.9.x/../../../../x86_64-linux-android/bin/ld: error: cannot find -lpthread
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> INFO: Elapsed time: 481.943s, Critical Path: 42.15s
> INFO: 2577 processes: 155 internal, 2410 processwrapper-sandbox, 12 worker.
> FAILED: Build did NOT complete successfully
> ```

This might have to do with this boringssl change: https://github.com/google/boringssl/commit/c5179c693097d0c47b7646543bc7156e894275f7  Is the android platform not being detected in this build for some reason and thus `-lpthread` is being passed?

jtattermusch:
> > Yeah, add `'USE_HEADERMAP' => 'NO'` works
> > patch to the podspec:
> > ```
> > diff --git a/src/objective-c/BoringSSL-GRPC.podspec b/src/objective-c/BoringSSL-GRPC.podspec
> > index 529dacc37b..8c3e5e1c6e 100644
> > --- a/src/objective-c/BoringSSL-GRPC.podspec
> > +++ b/src/objective-c/BoringSSL-GRPC.podspec
> > @@ -151,6 +151,11 @@ Pod::Spec.new do |s|
> >      ss.dependency "#{s.name}/Interface", version
> >    end
> > 
> > +  s.pod_target_xcconfig = {
> > +    # Do not let src/include/openssl/time.h override system API
> > +    'USE_HEADERMAP' => 'NO',
> > +  }
> > +
> >    s.prepare_command = <<-END_OF_COMMAND
> >      set -e
> >      # Add a module map and an umbrella header
> > ```
> 
> Thanks, I added the patch to this PR, let's see what the test say.

Some ObjC test jobs were fixed by the patch, but "ObjC examples" is still failing:
https://source.cloud.google.com/results/invocations/ecdb2569-856d-406a-bdf0-b97717f24f1e/targets/github%2Fgrpc%2Frun_tests%2Fobjc_macos_opt_native%2Fios-buildtest-example-sample-frameworks/tests

```
Ditto /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Build/Build/Build/Intermediates.noindex/Pods.build/Release-iphoneos/BoringSSL-GRPC.build/module.modulemap /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Build/Build/Build/Products/Release-iphoneos/BoringSSL-GRPC/openssl_grpc.framework/Modules/module.modulemap (in target 'BoringSSL-GRPC' from project 'Pods')
/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Pods/BoringSSL-GRPC/src/crypto/x509/x_spki.c:61:10: fatal error: 'openssl_grpc/asn1t.h' file not found
#include <openssl_grpc/asn1t.h>
         ^~~~~~~~~~~~~~~~~~~~~~
/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Pods/BoringSSL-GRPC/src/crypto/x509/x_spki.c:61:10: note: did not find header 'asn1t.h' in framework 'openssl_grpc' (loaded from '/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Build/Build/Build/Products/Release-iphoneos/BoringSSL-GRPC')
```

davidben:
> This might have to do with this boringssl change: https://github.com/google/boringssl/commit/c5179c693097d0c47b7646543bc7156e894275f7 Is the android platform not being detected in this build for some reason and thus -lpthread is being passed?

Wow, https://source.cloud.google.com/results/invocations/17247ee8-2b0c-4573-a1c6-57130ee430e3/log sure got enthusiastic about linking in pthreads. There's like a page full of `-pthread`! It's irritating that Bazel doesn't just have a standards pthreads target. That would avoid a lot of mess.

Anyway, based on https://stackoverflow.com/questions/23250863/difference-between-pthread-and-lpthread-while-compiling, it sounds like we should just always be using `-pthread` anyway, which would hopefully be simpler and more portable. I'm OOO this week, but adding @agl if he has cycles for it this week.

That said, that your Android build isn't picking up the (allegedly) Bazel-standard `@platforms//os:android` seems a problem. E.g. you'll miss out on the assembly optimizations added in https://boringssl.googlesource.com/boringssl/+/99eac424d84d34395c7eea7a90d419e23c00dfc2%5E%21/. So while it'd be good for us to use `-pthread` uniformly, you might need to look into that.

> Some ObjC test jobs were fixed by the patch, but "ObjC examples" is still failing:
> https://source.cloud.google.com/results/invocations/ecdb2569-856d-406a-bdf0-b97717f24f1e/targets/github%2Fgrpc%2Frun_tests%2Fobjc_macos_opt_native%2Fios-buildtest-example-sample-frameworks/tests

I don't think anything's changed around those headers. Perhaps this is some consequence of turning off header maps? You all will probably need to look into the cocoapods issue, as that sounds like more gRPC-specific issues.

jtattermusch:
@davidben  @agl   I'm trying to make our android build work with bazel `@platforms`, but it's proving tricky.
Because we're using `--fat_apk_cpu=x86_64,arm64-v8a` https://github.com/grpc/grpc/blob/a017e9b7f20743c69627b94d7d101e4e6baadb44/tools/internal_ci/linux/grpc_binder_transport_apk_build_in_docker.sh#L31, it seems that I must use bazel's `--platform_mappings`, but so far I haven't been able to get things to work.
CC @sifmelcara  since you added the android binder example - can you please help with getting the examples work with the new version of boringssl.

sifmelcara:
I tried building binder transport APK locally with this PR by running the following command, but see a lot of compilation error in `external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S`, is this the same issue you have?

```
# Use https://github.com/grpc/grpc/blob/a017e9b7f20743c69627b94d7d101e4e6baadb44/tools/dockerfile/test/binder_transport_apk/Dockerfile#L67-L69 to install NDK
export ANDROID_NDK_HOME=/opt/android-sdk/ndk/21.4.7075529
export ANDROID_HOME=/opt/android-sdk/
bazel build --define=use_strict_warning=true --extra_toolchains=@rules_python//python:autodetecting_toolchain_nonstrict --experimental_ui_max_stdouterr_bytes=-1 //examples/android/binder/java/io/grpc/binder/cpp/exampleclient:app
```

davidben:
@jtattermusch Fun. I've no idea what any of that means. I gather this part of Bazel doesn't actually work yet?

@sifmelcara I don't see an error about poly_rq_mul.S anywhere in your comment. What's that error? The one folks are running into here seem to be something else.

sifmelcara:
The poly_rq_mul.S looks like this (there are like 6000 of these error)

<details>
  <summary>Click to expand</summary>

```

external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:28:8: error: invalid alignment value
.align 32
       ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:302:20: error: expected STT_<TYPE_IN_UPPER_CASE>, '#<type>', '%<type>' or "<type>"
.type poly_Rq_mul, @function
                   ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:303:1: error: unknown directive
.att_syntax prefix
^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:306:6: error: unexpected token in operand
push %rbp
     ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:309:6: error: unexpected token in operand
movq %rsp, %rbp
     ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:311:6: error: unexpected token in operand
push %r12
     ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:320:5: error: unexpected token in operand
mov %rcx, %r8
    ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:321:31: error: unexpected token in operand
addq $6144+12288+512+9408+32, %r8
                              ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:322:5: error: unexpected token in operand
mov %r8, %rax

... <redacted> ...

external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8465:20: error: unexpected token in variant, expected ')'
vpand mask_mod8192(%rip), %ymm11, %ymm11
                   ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8466:9: error: unexpected token in operand
vmovdqu %ymm11, 1232(%rdi)
        ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8467:12: error: unexpected token in argument list
vmovdqu 616(%rdi), %ymm11
           ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8468:12: error: unexpected token in argument list
vpaddw 2240(%r8), %ymm11, %ymm11
           ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8469:20: error: unexpected token in variant, expected ')'
vpand mask_mod8192(%rip), %ymm11, %ymm11
                   ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8470:9: error: unexpected token in operand
vmovdqu %ymm11, 616(%rdi)
        ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8471:12: error: unexpected token in argument list
vmovdqu 968(%rdi), %ymm11
           ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8472:12: error: unexpected token in argument list
vpaddw 2496(%r8), %ymm11, %ymm11
           ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8473:20: error: unexpected token in variant, expected ')'
vpand mask_mod8192(%rip), %ymm11, %ymm11
                   ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8474:9: error: unexpected token in operand
vmovdqu %ymm11, 968(%rdi)
        ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8475:13: error: unexpected token in argument list
vmovdqu 1320(%rdi), %ymm11
            ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8476:12: error: unexpected token in argument list
vpaddw 2752(%r8), %ymm11, %ymm11
           ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8477:20: error: unexpected token in variant, expected ')'
vpand mask_mod8192(%rip), %ymm11, %ymm11
                   ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8478:9: error: unexpected token in operand
vmovdqu %ymm11, 1320(%rdi)
        ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8479:5: error: unexpected token in operand
pop %r12
    ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8481:5: error: unexpected token in operand
pop %rbp
    ^
external/boringssl/src/crypto/hrss/asm/poly_rq_mul.S:8485:1: error: invalid instruction, did you mean: rev?
ret
^
Target //examples/android/binder/java/io/grpc/binder/cpp/exampleclient:app failed to build
```

</details>

The way I'm testing this PR is `git checkout 15f3256d6d5b582a4b08df44fc3e440ba30ccce1 ; git submodule update --init --recursive`

davidben:
That sounds like something has gone very badly wrong in gRPC's build. That's an x86_64 assembly file, which shouldn't be used when building for Arm targets. Perhaps more fallout from Bazel's "platforms" machinery not working right?

gRPC folks, this is the second of your buildsystems that has run into mysterious issues so far. Who on your team is familiar with your build?

davidben:
One thing that might help in future: https://boringssl.googlesource.com/boringssl/+/cdccbe121fa5dc8c6df7102eaedb62c99eec8273 tried to wrap every asm file in ifdefs corresponding to the platform.

The hope was that we can reduce the amount of platform conditions in the build and just rely on the preprocessor to drop the files that aren't used. That would probably have avoided the poly_rq_mul.S issue. But I haven't gotten that to work on the build side yet, so for now we still depend on detecting the platform in the build.

HannahShiSFB:
> > > Yeah, add `'USE_HEADERMAP' => 'NO'` works
> > > patch to the podspec:
> > > ```
> > > diff --git a/src/objective-c/BoringSSL-GRPC.podspec b/src/objective-c/BoringSSL-GRPC.podspec
> > > index 529dacc37b..8c3e5e1c6e 100644
> > > --- a/src/objective-c/BoringSSL-GRPC.podspec
> > > +++ b/src/objective-c/BoringSSL-GRPC.podspec
> > > @@ -151,6 +151,11 @@ Pod::Spec.new do |s|
> > >      ss.dependency "#{s.name}/Interface", version
> > >    end
> > > 
> > > +  s.pod_target_xcconfig = {
> > > +    # Do not let src/include/openssl/time.h override system API
> > > +    'USE_HEADERMAP' => 'NO',
> > > +  }
> > > +
> > >    s.prepare_command = <<-END_OF_COMMAND
> > >      set -e
> > >      # Add a module map and an umbrella header
> > > ```
> > 
> > 
> > Thanks, I added the patch to this PR, let's see what the test say.
> 
> Some ObjC test jobs were fixed by the patch, but "ObjC examples" is still failing: https://source.cloud.google.com/results/invocations/ecdb2569-856d-406a-bdf0-b97717f24f1e/targets/github%2Fgrpc%2Frun_tests%2Fobjc_macos_opt_native%2Fios-buildtest-example-sample-frameworks/tests
> 
> ```
> Ditto /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Build/Build/Build/Intermediates.noindex/Pods.build/Release-iphoneos/BoringSSL-GRPC.build/module.modulemap /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Build/Build/Build/Products/Release-iphoneos/BoringSSL-GRPC/openssl_grpc.framework/Modules/module.modulemap (in target 'BoringSSL-GRPC' from project 'Pods')
> /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Pods/BoringSSL-GRPC/src/crypto/x509/x_spki.c:61:10: fatal error: 'openssl_grpc/asn1t.h' file not found
> #include <openssl_grpc/asn1t.h>
>          ^~~~~~~~~~~~~~~~~~~~~~
> /Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Pods/BoringSSL-GRPC/src/crypto/x509/x_spki.c:61:10: note: did not find header 'asn1t.h' in framework 'openssl_grpc' (loaded from '/Volumes/BuildData/tmpfs/altsrc/github/grpc/workspace_objc_macos_opt_native/src/objective-c/examples/Sample/Build/Build/Build/Products/Release-iphoneos/BoringSSL-GRPC')
> ```

ObjC examples tests have passed in #31869, not sure which parts are different, but will help to look into it. I'm currently working on the android problem.

jtattermusch:
Ok, so I think I've been able to fix the android binder example build. Basically, since we use `android_library()` together with `--fat_apk_cpu` and our bazel build currently isn't compatible with bazel `@platforms` (at least AFAIK), the only possible workaround is to use the bazel [platform mappings](https://bazel.build/concepts/platforms#platform-mappings) feature, which basically allows us to define which set of bazel flags should be interpreted as a given "platform".

I think I was able to figure out the right platform_mappings file to fix the build, but the problem with boringssl switching to bazel `@platforms` seems more complicated.

- it seems that we'll have to define the right platform_mappings in the grpc bazel build for all the supported platforms, otherwise we will lose the asm optimizations (or get a broken build) whenever we are crosscompiling (e.g. for android, ios, or for a different cpu arch), as by default the platform definitions (such as `@platforms//os:linux`, `@platforms//os:android`) will be set to the local machine's CPU and OS (IIUC what https://bazel.build/concepts/platforms#default-platforms says) 

- what's worse is that all the grpc users who use bazel to build with grpc as a dependency will have to do the same, since platform_mappings are only applied once at the top-level project and all the dependencies simply use the top-level platform_mappings (same situation when grpc has to provide their own platform_mappings to make sure the platform constraints values are set correctly for the boringssl builds).

I think eventually grpc will need to switch to `@platform`-aware builds by default, but that seems quite difficult, especially given that it requires all the grpc users to adapt their builds accordingly.

jtattermusch:
@HannahShiSFB looks like I have a solution for the android binder example (or a temporary workaround), but I'd still need help with the ObjC Examples.

sampajano:
> @HannahShiSFB looks like I have a solution for the android binder example (or a temporary workaround), but I'd still need help with the ObjC Examples.

@jtattermusch Hey Jan, since Hannah has provided her boring SSL update PR (https://github.com/grpc/grpc/pull/31869) where the test is working, could you try to find out what are the discrepancies on why it's passing there but failing here? It might give you a clue on how to improve the script here? (As far as i understand the only difference between yours and her PR is that she did the upgrade manually and you did it via a script?) Thanks!

davidben:
> [...]  it sounds like we should just always be using -pthread anyway, which would hopefully be simpler and more portable.

https://boringssl-review.googlesource.com/c/boringssl/+/55927, though it sounds like that doesn't address the broader `@platforms` mess.

HannahShiSFB:
> @HannahShiSFB looks like I have a solution for the android binder example (or a temporary workaround), but I'd still need help with the ObjC Examples.

It's great that you got the android binder example done!  Unfortunately, I don't have a quick answer for the ObjC Examples failure, will need extra time to look into it. (we need to fix this eventually, but maybe not right now? one step at a time... )

davidben:
Since the new error doesn't seem related to time.h, my guess is that disabling headermaps didn't so much fail to solve the a second time.h problem but unveiled/introduced (guess we'll figure out which when we understand this mess!) a problem elsewhere.

From what I can tell, headermaps seem to bypass the standard include path machinery. In the case of time.h, it bypassed it in a way that caused problems. But it's plausible that something else was relying on it and didn't have an equivalent setup with the standard mechanism. So my guess is, by setting that, we've exposed this and need to tidy that up. (We can probably confirm this guess by seeing what USE_HEADERMAPS does in a pre-time.h BoringSSL.)

jtattermusch:
While this PR is important and we do want to upgrade boringssl ASAP (definitely for the next release cycle 1.53.x), I'm not going to block the 1.52.x release on it (so removing the release blocker label).
Even though the boringssl version we're using currently is quite out of date (approx from Nov 2021), there have been no important security fixes in boringssl since then (as confirmed by the boringssl team) - so using the older version of boringssl is fine from security perspective.

HannahShiSFB:
'USE_HEADERMAP' => 'NO' does fix the time.h issue.

I couldn't reproduce the same error of not find 'asn1t.h' but werid `/Users/hannahshi/Library/Developer/Xcode/DerivedData/Sample-bhqfopwhkrtesjdjuiyqaalqeesd/Build/Products/Release-iphonesimulator/BoringSSL-GRPC/openssl_grpc.framework/Headers/base.h:59:10 Cyclic dependency in module 'Darwin': Darwin -> openssl -> Darwin` error.

Projects in `src/objective-c/tests` builds without an issue with the latest boringssl and 'USE_HEADERMAP' => 'NO' fix.

But I get error building samples in `src/objective-c/examples`, the differences are the `use_frameworks!` and `FRAMEWORKS` env in 
https://github.com/grpc/grpc/blob/master/src/objective-c/examples/Sample/Podfile#L9 and 
https://github.com/grpc/grpc/blob/master/src/objective-c/examples/RemoteTestClient/RemoteTest.podspec#L27

I however don't know why they are different form the test projects and how to fix them.

jtattermusch:
Closing in favor of https://github.com/grpc/grpc/pull/32492

PR #31939, opened 2022-12-20
round_robin: avoid spurious updates when switching subchannel lists

markdroth:
Previously, when RR saw a new address list, it would swap to that address list as soon as the first subchannel in the new list reported READY, even if there are still other subchannels in the list for which we have not yet seen the initial connectivity state notification.  However, that behavior is sub-optimal in the case where there are multiple subchannels in the new list that will report READY on their initial notification, especially if those subchannels are already present in the old subchannel list.  To fix this, we now wait until we have seen the initial connectivity state notification for all subchannels in the new list before swapping over to it.  This also avoids unnecessary picker churn.

Also add a basic unit test for RR that covers this case.  (Other tests still need to be added separately.)


eugeneo (approved):
LGTM! Thanks, should unblock me.

PR #31366, opened 2022-10-14
Ensure Full Build is Run for each Bazel Version in Bazel Distribtest

gnossen:
It looks like there were two issues here:

1. Bazel was not respecting the `tools/bazel` script. Bazel is supposed to check for this directory and run the script there as `bazel`. This should have been running the requested version. This appears not to have been happening. I have switched to invoking `tools/bazel` directly and have added a check to fail the test if Bazel does not report itself as running the expected version.

2. Bazel seems to have been caching _across versions_. That is, after a test was run on version A, it would not be re-run on version B but would instead report as passing. `bazel clean --expunge` did not resolve this issue. I had to resort to manually deleting the Bazel cache directory.

[Manual run layered on top of a commit known to be broken for Bazel 4.X failing as expected.](https://source.cloud.google.com/results/invocations/95010c23-07ae-4eef-aa40-cd65546b449c/log)

PR #31940, opened 2022-12-21
Add Bazel 6 to the list of supported bazel versions

veblush:
Simply adding Bazel 6 to the list of supported Bazel versions.


jtattermusch:
IIRC, the bazel distribtests are currently unrealiable, so I wouldn't bet on the build passing with bazel 6.
I think we should first finalize https://github.com/grpc/grpc/pull/31366 so that the bazel distribtests work as they should and only after that we should mess with adjusting bazel versions.

jtattermusch:
Force re-running all the tests because https://github.com/grpc/grpc/pull/31366 was merged.

jtattermusch:
https://github.com/grpc/grpc/pull/31999 also needs to be merged first.

gnossen (approved):

PR #31942, opened 2022-12-21
[EventEngine] RunAfter migration: Chttp2Server

yijiem:


src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -407,29 +410,33 @@ void Chttp2ServerListener::ActiveConnection::HandshakingState::Start(
...
-    }
     grpc_transport_perform_op(&transport->base, op);
   }
-  self->Unref();
 }
 
 void Chttp2ServerListener::ActiveConnection::HandshakingState::
     OnReceiveSettings(void* arg, grpc_error_handle /* error */) {
   HandshakingState* self = static_cast<HandshakingState*>(arg);
-  grpc_timer_cancel(&self->timer_);
```
drfloob, line 432: Check my logic, please, on behavior deviations: Even with `grpc_timer_cancel` called here, I think there was possibly a race in the previous implementation where `OnTimeout` may have been called with an OK status already (imagine ~concurrently with `OnReceiveSettings`), in which case the `grpc_transport_perform_op` in OnTimeout would have proceeded. It seems like this was previously incorrect behavior, and this PR fixes that by resetting the timer handle under a mutex here, and checking it in `OnTimeout` before committing to performing the op.

yijiem: Good case! I didn't think about this case. I think it's possible. But I'm wondering what the correct behavior in this case would be. I feel like from the server's point of view it might rightfully think that the client has timed out to send the settings, send goaway and disconnect while at the same time the client sends the settings...

drfloob: In this race, presuming timeout and receipt of settings are happening concurrently on the server, I think I would err on the side of leniency towards the client, which is what this PR is doing. But I'm not 100% certain that's correct either.

@markdroth would you mind sharing your insights?

src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -407,29 +410,33 @@ void Chttp2ServerListener::ActiveConnection::HandshakingState::Start(
...
+void Chttp2ServerListener::ActiveConnection::HandshakingState::OnTimeout() {
+  grpc_chttp2_transport* transport = nullptr;
+  {
+    MutexLock lock(&connection_->mu_);
+    if (timer_handle_.has_value()) {
+      transport = connection_->transport_;
+      timer_handle_.reset();
+    }
+  }
+  if (transport) {
```
markdroth, line 422: Please write `!= nullptr`, to make it clear to the reader that this is a pointer, not a boolean.

src/core/ext/transport/chttp2/server/chttp2_server.cc
markdroth: Yes, this looks right to me.  If we've actually seen the settings, then we shouldn't close the connection when the callback runs.

Looking at the existing code, I'm actually surprised this built, because `timer_` is marked as being guarded by the mutex, but here we were cancelling the timer without holding the mutex.  Not sure why that was working, but I guess it doesn't matter now.

src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -578,20 +590,17 @@ void Chttp2ServerListener::ActiveConnection::SendGoAway() {
...
-                      .value_or(Duration::Minutes(10))),
-          &on_drain_grace_time_expiry_);
-      drain_grace_timer_expiry_callback_pending_ = true;
+      drain_grace_timer_handle_ = event_engine_->RunAfter(
+          std::max(Duration::Zero(),
+                   listener_->args_
+                       .GetDurationFromIntMillis(
+                           GRPC_ARG_SERVER_CONFIG_CHANGE_DRAIN_GRACE_TIME_MS)
+                       .value_or(Duration::Minutes(10))),
+          [self = Ref()] {
```
markdroth, line 599: I think this ref needs to be released while the `ExecCtx` is in scope, just like we're doing on line 519 for the other timer.

src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -170,8 +172,9 @@ class Chttp2ServerListener : public Server::ListenerInterface {
           ABSL_GUARDED_BY(&connection_->mu_);
       // State for enforcing handshake timeout on receiving HTTP/2 settings.
       Timestamp const deadline_;
-      grpc_timer timer_ ABSL_GUARDED_BY(&connection_->mu_);
-      grpc_closure on_timeout_ ABSL_GUARDED_BY(&connection_->mu_);
+      absl::optional<EventEngine::TaskHandle> timer_handle_
+          ABSL_GUARDED_BY(&connection_->mu_);
+      EventEngine* event_engine_ ABSL_GUARDED_BY(&connection_->mu_);
```
markdroth, line 177: I don't think we need to store this pointer again in this object.  We can instead use `connection_->event_engine_`.

src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -636,30 +646,31 @@ void Chttp2ServerListener::ActiveConnection::OnClose(
...
-      MutexLock lock(&self->mu_);
-      transport = self->transport_;
+  {
+    MutexLock lock(&mu_);
+    if (drain_grace_timer_handle_.has_value()) {
+      transport = transport_;
+      drain_grace_timer_handle_.reset();
     }
+  }
+  if (transport) {
```
markdroth, line 668: `!= nullptr`

markdroth (approved):
This looks good overall!  My comments are all minor, so I'll go ahead and approve.

I would also like @yashykt to review this before you merge it, since this code is fairly complex, and he might catch something that I missed.

src/core/ext/transport/chttp2/server/chttp2_server.cc
yijiem: Good point. Removed. I also changed to grab the `ActiveConnection::event_engine_` in its constructor instead of inside `ActionConnection::Start()`. I don't remember why I did that before...

yijiem (commented):
Thanks Mark!

src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -206,10 +208,9 @@ class Chttp2ServerListener : public Server::ListenerInterface {
     // is created.
     grpc_chttp2_transport* transport_ ABSL_GUARDED_BY(&mu_) = nullptr;
     grpc_closure on_close_;
-    grpc_timer drain_grace_timer_;
-    grpc_closure on_drain_grace_time_expiry_;
-    bool drain_grace_timer_expiry_callback_pending_ ABSL_GUARDED_BY(&mu_) =
-        false;
+    absl::optional<EventEngine::TaskHandle> drain_grace_timer_handle_
+        ABSL_GUARDED_BY(&mu_);
+    EventEngine* event_engine_ ABSL_GUARDED_BY(&mu_);
```
yashykt, line 213: How do we guarantee the lifetime of this EventEngine object?

src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -407,29 +410,33 @@ void Chttp2ServerListener::ActiveConnection::HandshakingState::Start(
...
-      MutexLock lock(&self->connection_->mu_);
-      transport = self->connection_->transport_;
-    }
     grpc_transport_perform_op(&transport->base, op);
   }
-  self->Unref();
 }
 
 void Chttp2ServerListener::ActiveConnection::HandshakingState::
     OnReceiveSettings(void* arg, grpc_error_handle /* error */) {
```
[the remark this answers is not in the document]
yashykt: `OnReceiveSettings` is invoked when HTTP2 receives settings. That would be the most common case.

src/core/ext/transport/chttp2/server/chttp2_server.cc
yashykt: There is always a race given that the network is involved and that the SETTINGS frame can be received at any time, but I think this is a good change. Speaking purely from the Chttp2Server's perspective, if it has already seen the settings, it should ignore the timer even if the timer was fired with an OK status.

src/core/ext/transport/chttp2/server/chttp2_server.cc
```diff
@@ -579,20 +590,18 @@ void Chttp2ServerListener::ActiveConnection::SendGoAway() {
...
-          &on_drain_grace_time_expiry_);
-      drain_grace_timer_expiry_callback_pending_ = true;
+      drain_grace_timer_handle_ = event_engine_->RunAfter(
+          std::max(Duration::Zero(),
+                   listener_->args_
+                       .GetDurationFromIntMillis(
+                           GRPC_ARG_SERVER_CONFIG_CHANGE_DRAIN_GRACE_TIME_MS)
+                       .value_or(Duration::Minutes(10))),
+          [self = Ref(DEBUG_LOCATION, "drain_grace_timer")]() mutable {
+            ApplicationCallbackExecCtx callback_exec_ctx;
```
yashykt, line 600: question for @drfloob : Do all timers now need ExecCtx to be declared on the stack?

drfloob: _Need_ is a string word, but yes, the policy is that all callbacks must create both ExecCtx types locally at the moment. Debugging missing ExecCtxs in EventEngine callbacks has been complicated, and it will be trivially easy to remove them all once we've migrate away from `ExecCtx::Run` (a long-term thing).

src/core/ext/transport/chttp2/server/chttp2_server.cc
yijiem: Good callout! This event_engine_ is grabbed from a copy of the listener_'s ChannelArgs. And it also held a `RefCounterPtr<Chttp2ServerListener> listener_;` at line 202. Added a comment to clarify that.

yijiem (commented):
Thanks Yash!

src/core/ext/transport/chttp2/server/chttp2_server.cc
yashykt: Slightly incorrect. We make a copy of the listener args, modify it and then pass it to `ActiveConnection`. I would suggest getting the pointer from the channel args in the listener and then explicitly pass the EventEngine pointer as a parameter to the constructor of ActiveConnection.

yijiem: Makes sense! Done.

yashykt (approved):
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/31942)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @drfloob, @markdroth, and @yijiem)

PR #31969, opened 2022-12-22
Fix hang on fork when the security context is destroyed in the Executor

drfloob:
Attempting to create an ExecCtx while forking will hang forever. In the apparently rare case that the transport is destroyed in an Executor thread, this blocks indefinitely in the Executor pre-fork handler

I don't believe this ExecCtx is needed regardless, see `grpc_server_security_context_destroy` for comparison.


ctiller (approved):
`grpc_client_security_context_destroy` is a private interface, so this should be ok.

Issue #30976, opened 2022-09-14
Undefined symbol `_ruby_abi_version` regression with Xcode 13 -> 14


agrieser-healthie:
### What version of gRPC and what language are you using?
grpc: 1.48.0
language: ruby 2.7.6

### What operating system (Linux, Windows,...) and version?
MacOS 12.6
Apple M1 chip

### What runtime / compiler are you using (e.g. python version or version of gcc)
Xcode 14

### What did you do?
`gem install grpc -v 1.48.0`

### What did you expect to see?
Successfully install

### What did you see instead?
```
Undefined symbols for architecture arm64:
  "_ruby_abi_version", referenced from:
     -exported_symbol[s_list] command line option
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [grpc_c.bundle] Error 1
```

### Anything else we should know about your project / environment?
Worked correctly a few days ago with Xcode 13

technicalpickles:
That constant was added in https://github.com/grpc/grpc/pull/28976 . Is there a way to make this conditional on the ruby version if it's only required for Ruby 3.2+?

technicalpickles:
@ethankhall found a workaround for this:

```
export DLDFLAGS=-Wl,-undefined,dynamic_lookup
# however you are installing... gem directly
gem install grpc
# or ...
bundle install
```

cc @ngan @noizwaves

astanimirov:
@technicalpickles I have the same issue and it is blocking me to do my work. The provided workaround is not working for me as well. Could we give this issue a higher priority as the project cannot build and I do not know what to do?

solofeed:
Same issue with 
- ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [arm64-darwin21]
- google-protobuf-3.21.6
- grpc-1.48.0

solofeed:
Log

```
[...]

PR #31151, opened 2022-09-28
[Ruby] Stop stripping exported symbols

peterzhu2118:
This commit stops stripping exported symbols in the compiled binary. This will ensure that functions like `Init_grpc` and `ruby_abi_version` are not removed from the generated binary.

Fixes #30976.

Issue #28627, opened 2022-01-20
Support pre-built binaries for Ruby 3.1


erikaxel:
### Is your feature request related to a problem? Please describe.
It seems grpc does not support pre-built binaries for Ruby 3.1. This causes the gem-folder to balloon to almost 1gb. Locally in development this is not a huge issue, but it causes issues for CI and for the images created for deployment.

### Describe the solution you'd like
Add pre-built binaries for Ruby 3.1

### Describe alternatives you've considered
N/A

### Additional context
Ruby 3.1 release note: https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/

cirdes:
This issue is preventing me to upgrade my Rails app to Ruby 3.1 on Heroku. My app compressed image went from 290.9M to more than 500M (heroku hard limit). After narrow down the problem I found grpc with more than 1gb.
`1018M	/app/vendor/bundle/ruby/3.1.0/gems/grpc-1.43.1`

In my case this is a relevant issue, I hope this could be fixed soon.

simi:
:thinking: Could this be fixed by using this PR [in RubyGems](https://github.com/rubygems/rubygems/pull/4017)?

Stex:
For me it's not the image sizes that become a problem, but rather the time it takes to actually build them. Even on my M1 Max Macbook Pro, it took over 10 minutes to actually install grpc's native extensions using the `ruby:3.1` image with 5 cores used.

On a shared gitlab.com runner in CI, the build went from 2 minutes to 20 minutes.

vitalinfo:
@cirdes did you found any workarounds for this problem?

cirdes:
@vitalinfo unfortunately not. I'm waiting for a fix. 😞

ErikSchierboom:
We've run into the same issue. The huge size of the gem is also causing issues for us.

bouk:
If you want to save some space (in our case it's like 800MB) you can run the following after installing the gem:

```shell
grpc_path="$(bundle show --paths grpc)/src/ruby/ext/grpc"
make -C $grpc_path clean
rm -rf "$grpc_path"/{libs,objs}
```

vitalinfo:
version 1.45 increases our slug size even more (+43Mb) :(

gsar:
running it the same issue as well, hoping this has the attention of the maintainers.

whithajess:
This issue has us considering removing a google integration in our app because of the speed to build issue.

[...]

PR #31970, opened 2022-12-22
[Ruby] build: make exported symbol files platform-specific

apolcyn:
This is an alternative to https://github.com/grpc/grpc/pull/31151, and should fix https://github.com/grpc/grpc/issues/30976.

The idea of this PR is to use separate export files depending on whether or not we expect `ruby_abi_version` to be defined (i.e. depending on whether or not we're using ruby 3.2+).

I lean towards this PR over https://github.com/grpc/grpc/pull/31151 because it maintains the current symbol-stripping behavior:

Mainly, it looks like https://github.com/grpc/grpc/pull/31151 will cause a size increase of an uncompressed `grpc_c.so` file on linux from ~8MB to ~11MB due to the extra symbols. Compressed size correspondingly increases from ~3MB to ~4MB.

The latest pre-compiled linux gem, `grpc-1.50.0-x86_64-linux.gem`, is ~21MB. So the size increase might be noticeable.

cc @peterzhu2118

### DIFF
--- a/src/ruby/ext/grpc/ext-export-truffleruby.clang
+++ b/src/ruby/ext/grpc/ext-export-truffleruby.clang
@@ -1,2 +1,1 @@
 _Init_grpc_c
-_rb_tr_abi_version
--- a/src/ruby/ext/grpc/ext-export-truffleruby-with-ruby-abi-version.clang
+++ b/src/ruby/ext/grpc/ext-export-truffleruby-with-ruby-abi-version.clang
@@ -1,0 +1,2 @@
+_Init_grpc_c
+_rb_tr_abi_version
--- a/src/ruby/ext/grpc/ext-export-truffleruby-with-ruby-abi-version.gcc
+++ b/src/ruby/ext/grpc/ext-export-truffleruby-with-ruby-abi-version.gcc
@@ -1,6 +1,7 @@
 grpc_1.0 { 
   global: 
     Init_grpc_c;
+    rb_tr_abi_version;
   local: 
     *;
 };
--- a/src/ruby/ext/grpc/ext-export-with-ruby-abi-version.clang
+++ b/src/ruby/ext/grpc/ext-export-with-ruby-abi-version.clang
@@ -1,0 +1,2 @@
+_Init_grpc_c
+_ruby_abi_version
--- a/src/ruby/ext/grpc/ext-export-with-ruby-abi-version.gcc
+++ b/src/ruby/ext/grpc/ext-export-with-ruby-abi-version.gcc
@@ -1,6 +1,7 @@
 grpc_1.0 { 
   global: 
     Init_grpc_c;
+    ruby_abi_version;
   local: 
     *;
 };
--- a/src/ruby/ext/grpc/ext-export.clang
+++ b/src/ruby/ext/grpc/ext-export.clang
@@ -1,2 +1,1 @@
 _Init_grpc_c
-_ruby_abi_version
--- a/src/ruby/ext/grpc/ext-export.gcc
+++ b/src/ruby/ext/grpc/ext-export.gcc
@@ -1,7 +1,6 @@
 grpc_1.0 { 
   global: 
     Init_grpc_c;
-    ruby_abi_version;
   local: 
     *;
 };
--- a/src/ruby/ext/grpc/extconf.rb
+++ b/src/ruby/ext/grpc/extconf.rb
@@ -113,8 +113,30 @@ end
 $CFLAGS << ' -DGRPC_RUBY_WINDOWS_UCRT' if windows_ucrt
 $CFLAGS << ' -I' + File.join(grpc_root, 'include')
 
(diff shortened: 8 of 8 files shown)


johnnyshields:
@apolcyn looks good to me!

stanley-cheung (approved):

src/ruby/ext/grpc/ext-export-truffleruby.clang
```diff
@@ -1,2 +1 @@
 _Init_grpc_c
-_rb_tr_abi_version
```
eregon, line 2: This doesn't seem correct as it will break the gem on TruffleRuby if I read the PR correctly. TruffleRuby has rb_tr_abi_version before RUBY_VERSION 3.2, specifically it's 3.0 in the 22.3 release currently for TruffleRuby and it already had it before that. We can consider TruffleRuby always had this symbol, versions before are not supported anymore.

eregon:
IMHO https://github.com/grpc/grpc/pull/31151 is much safer, future proof and a lot simpler  A 1MB increase for the existing 23MB precompiled gem seems minor.

peterzhu2118:
I agree with @eregon. There isn't really much benefits to stripping symbols (I don't think most users of the gem are too worried about the binary size) and having different behaviour depending on the Ruby version will make it difficult to maintain, especially if CI doesn't test the different Ruby versions.

johnnyshields:
I agree with @eregon and @peterzhu2118

apolcyn:
> This doesn't seem correct as it will break the gem on TruffleRuby if I read the PR correctly. TruffleRuby has rb_tr_abi_version before RUBY_VERSION 3.2, specifically it's 3.0 in the 22.3 release currently for TruffleRuby and it already had it before that. We can consider TruffleRuby always had this symbol, versions before are not supported anymore.

I've opened https://github.com/grpc/grpc/pull/31995 to try to unbreak TruffleRuby.

> IMHO https://github.com/grpc/grpc/pull/31151 is much safer, future proof and a lot simpler. A 1MB increase for the existing 23MB precompiled gem seems minor.

I agree that #31151 is simpler, but the actual package size for pre-compiled gems will be multiplied by the ~4 supported ruby versions of each gem (there's a separate native library for each ruby version). I.e., total size increase should be closer to ~4MB.

Overall, I agree that the size increase in itself is probably not too bad. But I also don't think that these export files add *too much* complexity at the moment. I'm open to revisit this down the road if we find more problems stemming from these, but I'd prefer to avoid a size regression for now if it's easy enough. For example, one option is to remove symbol-stripping *only* from TruffleRuby.

> I agree with @eregon. There isn't really much benefits to stripping symbols (I don't think most users of the gem are too worried about the binary size) and having different behaviour depending on the Ruby version will make it difficult to maintain, especially if CI doesn't test the different Ruby versions.

We do have tests to-be-published gems that they'll work on each supported ruby version. We'll add such tests for 3.2 as a part of adding 3.2 support. I'm also adding some better test coverage for different ruby versions in https://github.com/grpc/grpc/pull/31991.

I don't know of data to side one way or the other, and a size increase will naturally affect downloads, storage, etc., so I'd just like to err on the side of caution here if it's easy enough. There's some evidence that package size is an issue for users in https://github.com/grpc/grpc/issues/28627.

stanhu:
@apolcyn @eregon It seems that `_ruby_abi_version` appears to be defined only for development builds for Ruby. It seems we may want to check `RUBY_PATCHLEVEL`: https://github.com/redis-rb/redis-client/issues/58#issuecomment-1367667097.

stanhu:
Submitted https://github.com/grpc/grpc/pull/31997 to fix this.

eregon:
> I've opened https://github.com/grpc/grpc/pull/31995 to try to unbreak TruffleRuby.

Thank you.

> Overall, I agree that the size increase in itself is probably not too bad. But I also don't think that these export files add too much complexity at the moment. I'm open to revisit this down the road if we find more problems stemming from these, but I'd prefer to avoid a size regression for now if it's easy enough. For example, one option is to remove symbol-stripping only from TruffleRuby.

I think it's getting significant and the perfect example is https://github.com/grpc/grpc/pull/31970#issuecomment-1367727911 i.e. yet another brittle check which might change in the future (e.g. I'm not sure why CRuby doesn't include the symbol on releases, it feels like it should be safer and detect bad gem homes, it might change).
Also TruffleRuby might rename that ABI symbol at some point, e.g. to unify it with CRuby (but ensure distinct versions schemes or some prefix in the value).

> There's some evidence that package size is an issue for users in https://github.com/grpc/grpc/issues/28627.

That's a different order of magnitude ([13.8 MB](https://rubygems.org/gems/grpc/versions) vs [1018MB](https://github.com/grpc/grpc/issues/28627#issuecomment-1038196969)), that issue is about size and experience if compiled from source IIUC, a couple MB extra is not a big deal compared to 1GB.

eregon:
FYI I've filed https://bugs.ruby-lang.org/issues/19289 on the Ruby side to see how we could improve things related to `rb_abi_version`.

PR #31980, opened 2022-12-26
Fix str vs. bytes error in Python bindings for ALTS.

vandry:
It looks like nobody ever created ALTS redentials from Python with a list of accepted service accounts before.

Simple reproduction:

```
import grpc
grpc.alts_channel_credentials(None)  # works
grpc.alts_channel_credentials(['foo'])  # fails
```

Without this change, generates this error:

```
[...]
  File "src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi", line 414, in _cython.cygrpc.channel_credentials_alts
  File "src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi", line 403, in _cython.cygrpc.ALTSChannelCredentials.__cinit__
TypeError: expected bytes, str found
```

(And the error cannot be worked around by the caller by passing a bytes
object from the Python side: you still get the same error.)


gnossen:
Thank you for the fix!

gnossen (approved):

PR #31981, opened 2022-12-27
Include type_traits

MaskRay:
Latest libc++ removes many transitive includes of type_traits and files using features like std::aligned_storage need explicit includes.

### DIFF
--- a/src/core/lib/gprpp/manual_constructor.h
+++ b/src/core/lib/gprpp/manual_constructor.h
@@ -25,6 +25,7 @@
 
 #include <stddef.h>
 
+#include <type_traits>
 #include <utility>
 
 #include "src/core/lib/gprpp/construct_destruct.h"
--- a/src/core/lib/gprpp/no_destruct.h
+++ b/src/core/lib/gprpp/no_destruct.h
@@ -17,6 +17,7 @@
 
 #include <grpc/support/port_platform.h>
 
+#include <type_traits>
 #include <utility>
 
 #include "src/core/lib/gprpp/construct_destruct.h"
--- a/src/core/lib/gprpp/ref_counted_ptr.h
+++ b/src/core/lib/gprpp/ref_counted_ptr.h
@@ -22,6 +22,7 @@
 #include <grpc/support/port_platform.h>
 
 #include <iosfwd>
+#include <type_traits>
 #include <utility>
 
 #include "src/core/lib/gprpp/debug_location.h"


drfloob:
@ctiller looks like an IWYU issue. Can you investigate?

MaskRay:
> std::aligned_storage

This patch is trivially correct. `std::aligned_storage` needs `<type_traits>` and it breaks with latest libc++. Can an owner approve and accept this patch?

veblush:
IWYU test says

```
diff --git a/src/core/lib/gprpp/manual_constructor.h b/src/core/lib/gprpp/manual_constructor.h
index 786edf3a17..eb769271d4 100644
--- a/src/core/lib/gprpp/manual_constructor.h
+++ b/src/core/lib/gprpp/manual_constructor.h
@@ -22,10 +22,7 @@
 // manually construct a region of memory with some type

-
 #include <stddef.h>
-
-#include <type_traits>
 #include <utility>

 #include "src/core/lib/gprpp/construct_destruct.h"
diff --git a/src/core/lib/gprpp/no_destruct.h b/src/core/lib/gprpp/no_destruct.h
index e8692e16d8..510d3b89ba 100644
--- a/src/core/lib/gprpp/no_destruct.h
+++ b/src/core/lib/gprpp/no_destruct.h
@@ -16,8 +16,6 @@
 #define GRPC_CORE_LIB_GPRPP_NO_DESTRUCT_H

-
-#include <type_traits>
 #include <utility>

 #include "src/core/lib/gprpp/construct_destruct.h"
diff --git a/src/core/lib/gprpp/ref_counted_ptr.h b/src/core/lib/gprpp/ref_counted_ptr.h
index 3e5b71a904..185e706d50 100644
--- a/src/core/lib/gprpp/ref_counted_ptr.h
+++ b/src/core/lib/gprpp/ref_counted_ptr.h
@@ -20,9 +20,7 @@
 #define GRPC_CORE_LIB_GPRPP_REF_COUNTED_PTR_H

-
 #include <iosfwd>
-#include <type_traits>
 #include <utility>

 #include "src/core/lib/gprpp/debug_location.h"
+ exit 1
```

To get this merged, either this PR or IWYU script needs a change.

MaskRay:
That is a problem of the configured iwyu in the CI. Can it be ignored? I don't think it can precise results if it doesn't track the libc++ upstream that closely...

After https://reviews.llvm.org/D140426  `<type_traits>` no longer includes all `<utility>`.

MaskRay:
Ping. This is needed to fix grpc with latest libc++. The configured IWYU does not use latest libc++, so the IWYU mismatch is expected.

alexfh (approved):
LG, though I don't have any rights for the project.

alexfh:
BTW, it seems like there's an explicit IWYU mapping from type_traits to utility:
https://github.com/grpc/grpc/blob/master/tools/distrib/iwyu_mappings.imp#L7

Should this be removed in this PR to fix the IWYU warning?

MaskRay:
Subsumed in https://github.com/grpc/grpc/pull/32008 . Closing

Issue #31227, opened 2022-10-04
UnboundLocalError: local variable 'raw_response' referenced before assignment


kovan:
### What version of gRPC and what language are you using?

1.49.0, Python

### What operating system (Linux, Windows,...) and version?

Linux on WSL

### What runtime / compiler are you using (e.g. python version or version of gcc)

Python 3.10.6

### What did you do?

I have no idea.

### What did you expect to see?

I am just going to leave the stack trace.
My code is failing and maybe it has something to do with this error:

```python

[generate] ----------------------------
[generate] Traceback (most recent call last):
[generate]   File "/app/videosdb/downloader.py", line 172, in _process_video
[generate]     _, video = await self.api.get_video_info(video_id)
[generate]   File "/app/videosdb/youtube_api.py", line 111, in get_video_info
[generate]     return await self._request_one(url, params)
[generate]   File "/app/videosdb/youtube_api.py", line 156, in _request_one
[generate]     item = await anext(generator)
[generate]   File "/app/videosdb/youtube_api.py", line 163, in generator
[generate]     async for page in pages:
[generate]   File "/app/videosdb/youtube_api.py", line 196, in _request_with_cache
[generate]     async for page in self.db.stream("cache/%s/pages" % cache_id):
[generate]   File "/home/appuser/.cache/pypoetry/virtualenvs/videosdb-9TtSrW0h-py3.10/lib/python3.10/site-packages/google/cloud/firestore_v1/async_collection.py", line 241, in stream
[generate]     async for d in query.stream(transaction=transaction, **kwargs):
[generate]   File "/home/appuser/.cache/pypoetry/virtualenvs/videosdb-9TtSrW0h-py3.10/lib/python3.10/site-packages/google/cloud/firestore_v1/async_query.py", line 264, in stream
[generate]     async for response in response_iterator:
[generate]   File "/home/appuser/.cache/pypoetry/virtualenvs/videosdb-9TtSrW0h-py3.10/lib/python3.10/site-packages/google/api_core/grpc_helpers_async.py", line 102, in _wrapped_aiter
[generate]     async for response in self._call:  # pragma: no branch
[generate]   File "/home/appuser/.cache/pypoetry/virtualenvs/videosdb-9TtSrW0h-py3.10/lib/python3.10/site-packages/grpc/aio/_call.py", line 323, in _fetch_stream_responses
[generate]     message = await self._read()
[...]

PR #30568, opened 2022-08-12
Fix occasional UnboundLocalError exception on grpc aio call cancellation

skaslev:
Fix occasional UnboundLocalError exception on grpc aio call cancellation  

  File "/venv/lib/python3.10/site-packages/grpc/aio/_call.py", line 323, in _fetch_stream_responses
    message = await self._read()
  File "/venv/lib/python3.10/site-packages/grpc/aio/_call.py", line 346, in _read
    if raw_response is cygrpc.EOF:
UnboundLocalError: local variable 'raw_response' referenced before assignment

release notes: no

PR #31982, opened 2022-12-27
fix: Re-raise python asyncio.CancelledErrors

njhill:
This is motivated by #31227 which I also encountered recently. I'm not familiar with the call lifecycle details, but it seems clear that there is some kind of race condition between user-initiated cancellations and other completion status updates.

If a user cancels a call before being notified of its completion, I believe it should be considered cancelled unconditionally, regardless of whether the call actually completed successfully before its cancellation status could be set.

So it seems better to re-raise explicitly rather than using `self._raise_for_status()`, which is a no-op in this race condition and can lead to an `UnboundedLocalError`.

Also related: #30568


gnossen (approved):
Thanks for the contribution! I'll get the unit test for this in in a subsequent PR.

Issue #31880, opened 2022-12-13
Wrong return type for abort()


Dreamsorcerer:
The method is documented as always raising an exception, so it should not return None.

This leads to typing errors where mypy is unaware that code after an abort() call will not be run.

Created a PR (#31879) to change this to `NoReturn`, but the CI wants CLA and other things sorted, which I don't have the time to look into. The change is not copyrightable anyway, so feel free to recreate or update that one.

Dreamsorcerer:
> Are there any instances where a user's current type checking could be broken by this change?

Unlikely. The only errors that could be introduced are legitimate errors that the user wants to fix (or from workarounds they'll want to  cleanup).

> Also, in your PR, you say "This leads to typing errors where mypy is unaware that code after an abort() call will not be run." Can you please help me understand these errors a bit better?

The rough flow that we encountered looks like:
```
x: str | None
if x is None:
    abort()
x + "foo"   # error: NoneType is not compatible with str
```
Or something to that extent. In order to convince mypy that this code is correct, we needed to put a `raise Exception()` or similar after the `abort()` so it can tell that the code never passes that point.

So, if users have done similar workarounds, they may get type errors to cleanup those workarounds. e.g. This workaround would later trigger an unreachable code error on the workaround line, which would then need to be removed.

Dreamsorcerer:
I've just tried this out and just realised that this doesn't really resolve the issue yet (and that the discussion about backwards-compatibility was pointless), because the annotations are not being exposed at all currently. (#29041)

PR #31984, opened 2022-12-27
Change Aio abort() function return type to NoReturn

XuanWang-Amos:
`abort()` function always raises an exception thus return type should be `NoReturn`.

Fix: https://github.com/grpc/grpc/issues/31880


gnossen (approved):
Thanks for taking care of this!

PR #31985, opened 2022-12-27
xds_override_host LB: add support for draining state

eugeneo:


test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -18,39 +18,55 @@
...
+using AddressAttributes =
+    std::map<const char*, std::unique_ptr<ServerAddress::AttributeInterface>>;
 
 class XdsOverrideHostTest : public LoadBalancingPolicyTest {
  protected:
   XdsOverrideHostTest()
       : policy_(MakeLbPolicy("xds_override_host_experimental")) {}
 
   RefCountedPtr<LoadBalancingPolicy::Config> MakeXdsOverrideHostConfig(
+      bool exclude_draining_status = false,
```
markdroth, line 57: Instead of using a bool here, I suggest passing in the set of override host statuses, since we will need tests to verify that we honor the override host status set even for non-draining statuses -- e.g., if the override host status set includes only HEALTHY, then the picker needs to ignore hosts with status UNKNOWN.

Please make sure that for each of the three health statuses, we have two tests: one that shows that an override host with the status will be used if it is present in the override host status set, and another that shows that an override host with the status will *not* be used if it is *not* present in the override host status set.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -79,6 +95,30 @@ class XdsOverrideHostTest : public LoadBalancingPolicyTest {
...
+                  std::make_unique<XdsEndpointHealthStatusAttribute>(
+                      XdsHealthStatus(status)));
+    return {MakeAddress(address), {}, std::move(attrs)};
+  }
+
+  void ApplyUpdateWithHealthStatuses(
+      absl::Span<const std::pair<const absl::string_view,
+                                 XdsHealthStatus::HealthStatus>>
+          addresses_and_statuses,
+      bool skip_draining = false) {
```
markdroth, line 111: Same as above: Please change this to pass in the full override host status set.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -213,6 +253,92 @@ TEST_F(XdsOverrideHostTest, ConnectingSubchannelIsQueued) {
...
+  ApplyUpdateWithHealthStatuses(
+      {{kAddresses[0], XdsHealthStatus::HealthStatus::kUnknown},
+       {kAddresses[2], XdsHealthStatus::HealthStatus::kHealthy}});
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
+  // Gone!
+  ExpectRoundRobinPicks(picker.get(), {kAddresses[0], kAddresses[2]}, pick_arg);
+}
+
+TEST_F(XdsOverrideHostTest, DrainingNotDrainingGone) {
```
markdroth, line 283: I don't understand the purpose of this test.  It looks like the only thing being covered here that isn't covered elsewhere is that we honor the override host when its status is healthy, but that has nothing to do with draining, so the test name and structure are confusing here.

As I mentioned above, I think that we should just have tests for each health status to show that we honor the override host status appropriately.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -213,6 +253,92 @@ TEST_F(XdsOverrideHostTest, ConnectingSubchannelIsQueued) {
   picker = ExpectState(GRPC_CHANNEL_READY);
   ExpectPickQueued(picker.get(), pick_arg);
 }
+
+TEST_F(XdsOverrideHostTest, DrainingState) {
```
markdroth, line 257: This test looks really good!

I think we also need a test similar to the `ConnectingSubchannelIsQueued` test above, but for DRAINING state.  We need to verify that we are still monitoring the subchannel's connectivity state even when the child has unreffed the subchannel and is not watching its connectivity state.

markdroth (commented):
This looks like a good start!

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -132,6 +140,8 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
 
     XdsOverrideHostLb* policy() { return policy_.get(); }
 
+    void detach() { key_ = absl::nullopt; }
+
    private:
     class ConnectivityStateWatcher : public ConnectivityStateWatcherInterface {
```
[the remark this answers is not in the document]
eugeneo: Please take a look at a new implementation.

Note that `interested_parties` is recomputed on every invocation. It is done in case the watchers have dynamic `interested_parties`

eugeneo (commented):
Thank you for the review! This was fairly substantial rework, I am afraid...

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -646,79 +619,58 @@ XdsOverrideHostLb::SubchannelWrapper::SubchannelWrapper(
...
-    const JsonArgs&) {
-  static const auto kJsonLoader =
-      JsonObjectLoader<XdsOverrideHostLbConfig>()
-          // Child policy config is parsed in JsonPostLoad
-          .Finish();
-  return kJsonLoader;
+grpc_pollset_set* XdsOverrideHostLb::SubchannelWrapper::
+    ConnectivityStateWatcher::interested_parties() {
+  if (interested_parties_ != nullptr) {
+    grpc_pollset_set_destroy(interested_parties_);
```
markdroth, line 651: I don't think this is safe.  The watcher will already have returned its interested_parties to the subchannel, and the subchannel may try to notify those interested_parties at any time.  We cannot delete the interested_parties out from under it.  Basically, the API contract here is that `interested_parties()` should return a single object that exists for the lifetime of the watcher.

I don't think it should be necessary to recreate the interested_parties set this way in the first place.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -646,79 +619,58 @@ XdsOverrideHostLb::SubchannelWrapper::SubchannelWrapper(
...
-        auto status = XdsHealthStatus::FromString(value.string_value());
-        if (status.has_value()) {
-          override_host_status_mask_ |= HealthStatusBitMask(*status);
-        }
-      }
-    }
+void XdsOverrideHostLb::SubchannelWrapper::ConnectivityStateWatcher::
+    OnConnectivityStateChange(grpc_connectivity_state state,
+                              absl::Status status) {
+  connectivity_state_ = state;
```
markdroth, line 665: Please use `connectivity_state_.store(state)`, so that it's clear to the reader that this is an atomic.

test/core/client_channel/lb_policy/xds_override_host_lb_config_parser_test.cc
```diff
@@ -42,12 +42,16 @@ TEST(XdsOverrideHostConfigParsingTest, ValidConfig) {
...
+      "      ],\n"
+      "      \"overrideHostStatus\": [\n"
+      "        \"DRAINING\", \"HEALTHY\", \"UNKNOWN\""
+      "      ]"
       "    }\n"
       "  }]\n"
       "}\n";
   auto service_config =
       ServiceConfigImpl::Create(ChannelArgs(), service_config_json);
+  EXPECT_EQ(service_config.status(), absl::OkStatus());
```
markdroth, line 54: This line is not necessary.  The next line already checks the same thing.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.h
```diff
@@ -49,12 +49,16 @@ class XdsOverrideHostLbConfig : public LoadBalancingPolicy::Config {
     return child_config_;
   }
 
+  int override_host_status_mask() const { return override_host_status_mask_; }
```
markdroth, line 52: Instead of exposing this as an integer, how about adding an interface to provide set semantics?  That would make the calling code (both in the LB policy and in the config tests) more readable.

I'm thinking that we could add something like this in src/core/ext/xds/xds_health_status.h:

```
class XdsHealthStatusSet {
 public:
  explicit XdsHealthStatusSet(const std::vector<XdsHealthStatus>& statuses) {
    for (XdsHealthStatus status : statuses) {
      Add(status);
    }
  }

  bool operator==(const XdsHealthStatusSet& other) { return status_mask_ == other.status_mask_; }

  void Clear() { status_mask_ = 0; }

  void Add(XdsHealthStatus status) {
    status_mask_ |= (0x1 << status.status());
  }

  bool Contains(XdsHealthStatus status) {
    return status_mask_ & (0x1 << status.status());
  }

 private:
  int status_mask_ = 0;
};
```

Then we could have this method return `XdsHealthStatusSet`, and the callers could check whether the set contains the expected statuses.

test/core/client_channel/lb_policy/xds_override_host_lb_config_parser_test.cc
```diff
@@ -151,6 +218,51 @@ TEST(XdsOverrideHostConfigParsingTest, ReportsEmptyChildPolicyArray) {
                 "error:errors validating xds_override_host LB policy config: "
                 "[field:childPolicy error:No known policies in list: ]]"));
 }
+
+TEST(XdsOverrideHostConfigParsingTest, ParsesOverrideHostStatusList) {
```
markdroth, line 222: This test looks like a duplicate of the one above.  I don't think it's needed.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -54,15 +54,9 @@ class XdsOverrideHostTest : public LoadBalancingPolicyTest {
       : policy_(MakeLbPolicy("xds_override_host_experimental")) {}
 
   RefCountedPtr<LoadBalancingPolicy::Config> MakeXdsOverrideHostConfig(
-      bool exclude_draining_status = false,
+      Json::Array override_host_status = {"UNKNOWN", "HEALTHY", "DRAINING"},
```
markdroth, line 57: If you take my suggestion elsewhere about adding an `XdsHealthStatusSet` class, then we could use that type here as well.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: I believe that you are misunderstanding the spec.  As I mentioned elsewhere, the picker needs to look at the health status of the override host, and if that status is not in the configured override host status set, then it needs to ignore the override host.

I think we need to test all of the following cases:

- override host status set contains UNKNOWN, override host status is UNKNOWN => override host will be used
- override host status set does *not* contain UNKNOWN, override host is UNKNOWN => override host will not be used
- override host status set contains HEALTHY, override host status is HEALTHY => override host will be used
- override host status set does *not* contain HEALTHY, override host is HEALTHY => override host will not be used
- override host status set contains DRAINING, override host status is DRAINING => override host will be used
- override host status set does *not* contain DRAINING, override host is DRAINING => override host will not be used (in this case, the host won't exist in the map at all)

Note that all of this is orthogonal to the other tests that cover the special behavior for DRAINING state related to whether the subchannel gets passed to the child and whether we trigger a connection attempt when we pick a DRAINING override host that is in state IDLE.

Once you add the above 6 tests, I don't think we need this `DrainingNotDrainingGone` test at all.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: Yes, I think that update is expected.  The way you've written the test, the address is starting out in a state other than DRAINING, and then you are switching it to state DRAINING.  This will trigger an update to the child policy that no longer contains this address, which will cause the child policy to return a new picker that will no longer return the removed address.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -280,6 +274,30 @@ TEST_F(XdsOverrideHostTest, DrainingState) {
...
+       {kAddresses[1], XdsHealthStatus::HealthStatus::kDraining},
+       {kAddresses[2], XdsHealthStatus::HealthStatus::kHealthy}});
+  auto subchannel = FindSubchannel(kAddresses[1]);
+  ASSERT_NE(subchannel, nullptr);
+  subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ExpectPickQueued(picker.get(), pick_arg);
+  EXPECT_TRUE(subchannel->ConnectionRequested());
+  subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+  ExpectPickQueued(picker.get(), pick_arg);
```
markdroth, line 298: Prior to this line, you need to again get a new picker:

```
```

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -646,79 +619,58 @@ XdsOverrideHostLb::SubchannelWrapper::SubchannelWrapper(
...
-      auto array = it->second.array_value();
-      for (const auto& value : array) {
-        auto status = XdsHealthStatus::FromString(value.string_value());
-        if (status.has_value()) {
-          override_host_status_mask_ |= HealthStatusBitMask(*status);
-        }
-      }
-    }
+void XdsOverrideHostLb::SubchannelWrapper::ConnectivityStateWatcher::
+    OnConnectivityStateChange(grpc_connectivity_state state,
```
markdroth, line 663: For a subchannel with health status DRAINING, if the subchannel reports READY, we need to trigger another picker update to the channel.  Otherwise, the queued pick will never be retried when the subchannel becomes connected again.

It looks like the tests aren't catching this case.  I'll add a comment elsewhere indicating what I think we need to change to catch this.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -280,6 +274,30 @@ TEST_F(XdsOverrideHostTest, DrainingState) {
...
+}
```
markdroth, line 299: We also need to test what happens when the connection attempt succeeds, to make sure that we're properly returning a new picker in this case.  I think we need to do something like this:

```
EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
```

markdroth (commented):

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -153,9 +163,10 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
      private:
       std::unique_ptr<ConnectivityStateWatcherInterface> delegate_;
       RefCountedPtr<SubchannelWrapper> subchannel_;
+      XdsHealthStatus health_status_{XdsHealthStatus::kUnknown};
```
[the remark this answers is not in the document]

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -132,6 +140,8 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
 
     XdsOverrideHostLb* policy() { return policy_.get(); }
 
+    void detach() { key_ = absl::nullopt; }
```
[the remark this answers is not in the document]
eugeneo: It is now "Shutdown"

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -621,18 +683,34 @@ const JsonLoaderInterface* XdsOverrideHostLbConfig::JsonLoader(
...
+  }
+  {
+    ValidationErrors::ScopedField field(errors, ".overrideHostStatus");
+    auto it = json.object_value().find("overrideHostStatus");
+    if (it != json.object_value().end()) {
+      override_host_status_mask_ = 0;
+      auto array = it->second.array_value();
+      for (const auto& value : array) {
+        auto status = XdsHealthStatus::FromString(value.string_value());
+        if (status.has_value()) {
```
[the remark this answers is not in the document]
eugeneo: Ack

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -287,26 +290,42 @@ TEST_F(XdsOverrideHostTest, DrainingSubchannelIsConnecting) {
...
   ASSERT_NE(subchannel, nullptr);
-  subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
+  // There are two notifications - one from child policy and one from the parent
+  // policy due to draining channel update
   picker = ExpectState(GRPC_CHANNEL_READY);
-  ExpectPickQueued(picker.get(), pick_arg);
+  EXPECT_EQ(ExpectPickComplete(picker.get(), pick_arg), kAddresses[1]);
+  ExpectRoundRobinPicks(picker.get(), {kAddresses[0], kAddresses[2]});
+  subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
+  ExpectStateAndQueuingPicker(GRPC_CHANNEL_READY, absl::OkStatus(), pick_arg);
```
markdroth, line 303: I think we need to check the behavior both with the `pick_arg` and without.  Suggest replacing this line with:

```
ExpectPickQueued(picker.get(), pick_arg);
ExpectRoundRobinPicks(picker.get(), {kAddresses[0], kAddresses[2]});
```

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -287,26 +290,42 @@ TEST_F(XdsOverrideHostTest, DrainingSubchannelIsConnecting) {
...
   EXPECT_TRUE(subchannel->ConnectionRequested());
+  ExpectQueueEmpty();
   subchannel->SetConnectivityState(GRPC_CHANNEL_CONNECTING);
+  picker = ExpectState(GRPC_CHANNEL_READY);
   ExpectPickQueued(picker.get(), pick_arg);
+  ExpectRoundRobinPicks(picker.get(),
+                        {"ipv4:127.0.0.1:441", "ipv4:127.0.0.1:443"});
+  subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  ASSERT_NE(picker, nullptr);
```
markdroth, line 313: I don't think this line is needed, because the following lines will already check for this.

markdroth (commented):
This is getting closer!

eugeneo (commented):
Thank you for the reviews. PTAL.

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: Huh, I thought we had already changed `ExpectPickComplete()` to check whether the picker is null.  Let's go ahead and do that so that the callers don't need to do it themselves.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -615,57 +633,96 @@ void XdsOverrideHostLb::Helper::AddTraceEvent(TraceSeverity severity,
...
+  auto it = watchers_.find(watcher);
+  if (it != watchers_.end()) {
+    watchers_.erase(it);
+  }
+}
+
+void XdsOverrideHostLb::SubchannelWrapper::UpdateConnectivityState(
+    grpc_connectivity_state state, absl::Status status) {
+  connectivity_state_.store(state);
+  // Assumption is that watchers may come and go as the events are processed
```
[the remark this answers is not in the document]
markdroth: Hmm, that's interesting.  Can you point me to the tests that were failing, with log output?

In any case, if the issue here is that sending the connectivity state notification causes the child to change the set of watchers, then let's change the comment to be more specific about what the actual issue is here: "Sending connectivity state notifications to the watchers may cause the set of watchers to change, so we can't be iterating over the set of watchers while we send the notifications."

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -471,77 +490,72 @@ OrphanablePtr<LoadBalancingPolicy> XdsOverrideHostLb::CreateChildPolicyLocked(
...
+  if (!addresses.ok()) {
+    return addresses;
+  }
+  ServerAddressList return_value;
+  std::map<const std::string, XdsHealthStatus> addresses_for_map;
+  for (const auto& address : *addresses) {
+    XdsHealthStatus status = GetAddressHealthStatus(address);
+    auto key = grpc_sockaddr_to_string(&address.address(), false);
+    if (key.ok()) {
+      addresses_for_map.insert({std::move(*key), status});
```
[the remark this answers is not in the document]
markdroth: I think this is incorrect.  We need to add entries even for hosts whose health status is not present in the override status set, because if the entry is not present in the map, then we won't wrap the subchannel, which means that we won't have a way to update the connectivity state later if the host's health status changes to a value that *is* in the set.

I believe the code I sent above is correct.

markdroth (commented):
This is getting closer!

eugeneo (commented):
Thank you for the review.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -242,21 +243,18 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
...
         : eds_health_status_(eds_health_status) {}
 
-    ~SubchannelEntry() {
-      SubchannelWrapper* subchannel = GetSubchannel();
-      if (subchannel != nullptr) subchannel->Shutdown();
-    }
-
-    void SetSubchannel(RefCountedPtr<SubchannelWrapper> subchannel) {
+    void SetSubchannel(SubchannelWrapper* subchannel) {
+      ResetSubchannel();
```
markdroth, line 247: Now that we've changed the subchannel to shut down automatically when it loses all of its strong refs, I think we no longer need this line.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -239,15 +302,21 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
...
   RefCountedPtr<SubchannelWrapper> LookupSubchannel(absl::string_view address);
 
-  void UpdateAddressMap(const absl::StatusOr<ServerAddressList>& addresses);
+  absl::StatusOr<ServerAddressList> UpdateAddressMap(
+      absl::StatusOr<ServerAddressList> addresses);
 
   RefCountedPtr<SubchannelWrapper> AdoptSubchannel(
       ServerAddress address, RefCountedPtr<SubchannelInterface> subchannel);
 
   void ResetSubchannel(absl::string_view key, SubchannelWrapper* subchannel);
```
markdroth, line 311: Suggest renaming this to `UnsetSubchannel()` as well.

test/core/client_channel/lb_policy/xds_override_host_test.cc
```diff
@@ -18,46 +18,57 @@
...
 #include "src/core/lib/json/json.h"
 #include "src/core/lib/load_balancing/lb_policy.h"
+#include "src/core/lib/resolver/server_address.h"
 #include "test/core/client_channel/lb_policy/lb_policy_test_lib.h"
 #include "test/core/util/test_config.h"
 
 namespace grpc_core {
 namespace testing {
 namespace {
+using AddressAttributes =
```
markdroth, line 48: This type is used in only one place, so it's probably better to just inline it instead of defining a type.

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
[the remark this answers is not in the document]
markdroth: Ah, okay -- it looks like this is happening during test shutdown.  The test shuts down the servers before shutting down the client, and when pick_first (which we use for LOGICAL_DNS clusters) sees the connection to the backend fail, it goes into IDLE state, at which point it stops watching the subchannel.

I guess this makes sense then.  It would be nice not to need it, but it's also not horrible.  I'll see if I can simplify this later via some structural changes to the `SubchannelInterface` API that I have in mind.

markdroth (commented):
This looks really good!  All remaining comments are minor cosmetic things.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Done

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -529,15 +523,15 @@ absl::StatusOr<ServerAddressList> XdsOverrideHostLb::UpdateAddressMap(
     XdsHealthStatus status = GetAddressHealthStatus(address);
     if (status.status() != XdsHealthStatus::kDraining) {
       return_value.push_back(address);
+    } else if (!config_->override_host_status_set().Contains(
+                   XdsHealthStatus(status))) {
+      continue;
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -529,15 +523,15 @@ absl::StatusOr<ServerAddressList> XdsOverrideHostLb::UpdateAddressMap(
     XdsHealthStatus status = GetAddressHealthStatus(address);
     if (status.status() != XdsHealthStatus::kDraining) {
       return_value.push_back(address);
+    } else if (!config_->override_host_status_set().Contains(
+                   XdsHealthStatus(status))) {
```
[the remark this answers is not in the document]

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -242,21 +243,18 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
...
-        subchannel_ = subchannel.get();
+        subchannel_ = subchannel;
       }
     }
 
-    void ResetSubchannel(SubchannelWrapper* expected) {
-      if (GetSubchannel() == expected) {
+    void ResetSubchannel() {
+      SubchannelWrapper* subchannel = GetSubchannel();
+      if (subchannel != nullptr) {
```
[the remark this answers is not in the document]

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
```diff
@@ -242,21 +243,18 @@ class XdsOverrideHostLb : public LoadBalancingPolicy {
...
+        subchannel_ = subchannel->Ref();
       } else {
-        subchannel_ = subchannel.get();
+        subchannel_ = subchannel;
       }
     }
 
-    void ResetSubchannel(SubchannelWrapper* expected) {
-      if (GetSubchannel() == expected) {
+    void ResetSubchannel() {
```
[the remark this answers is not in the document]

src/core/ext/filters/client_channel/lb_policy/xds/xds_override_host.cc
eugeneo: Yep.

eugeneo (commented):

test/core/client_channel/lb_policy/xds_override_host_test.cc
markdroth: Looks like you inlined it but didn't remove this (now unused) type definition.

markdroth (approved):
Just one minor nit remaining -- otherwise, looks great!

Feel free to merge after addressing the remaining comment.

test/core/client_channel/lb_policy/xds_override_host_test.cc
eugeneo: Oops! Thanks for catching it.

PR #31986, opened 2022-12-28
Update bazel_support.md

Clement-Jean:
Typo where 'k' should be a comma. We can now copy paste the code as is.

### DIFF
--- a/doc/bazel_support.md
+++ b/doc/bazel_support.md
@@ -19,7 +19,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 http_archive(
     name = "com_github_grpc_grpc",
     strip_prefix = "grpc-1.45.0",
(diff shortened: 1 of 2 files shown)


yijiem:
Not really familiar with this, but I believe this file is generated from the template here: [bazel_support.md.template](../blob/136055b043dcf2b15f69f535f659e4090cf25b9f/templates/doc/bazel_support.md.template)

Can you change that template and run `tools/buildgen/generate_projects.sh` to see if it works? You can find more info here: https://github.com/grpc/grpc/tree/master/templates#regenerating-project-files

jtattermusch (approved):
Good catch, looks like the typo was introduced by https://github.com/grpc/grpc/pull/29191.

jtattermusch:
I'll merge once the tests are green.

jtattermusch:
The bazel distribtest failure is also happening on master so it is unrelated. Other tests are green.

PR #31987, opened 2022-12-28
[ObjC] upgrade build_bazel_rules_apple to x.x.x, build_bazel_apple_support to x.x.x

HannahShiSFB:
Follow up of https://github.com/grpc/grpc/pull/31555
for sanitizer support from https://github.com/bazelbuild/rules_apple/pull/1590

needed by bazel 7.0 due to removal of `apple_common.multi_arch_split`
needs bazel >= 6.0

With this update, we can no longer support bazel 5 or earlier versions.

--
Removed empty `src/core/ext/transport/binder/java/WORKSPACE` file due to the following error:
```
ERROR: /Users/hannahshi/work/grpc/src/core/BUILD:35:14: Label '//src/core:ext/transport/binder/java/WORKSPACE' is invalid because '@@binder_transport_android_helper//' is a subpackage
ERROR: /Users/hannahshi/work/grpc/src/core/BUILD:1448:16: Cannot compute config conditions
```

--
@sampajano

### DIFF
--- a/bazel/grpc_deps.bzl
+++ b/bazel/grpc_deps.bzl
@@ -419,20 +419,20 @@ def grpc_deps():
     if "build_bazel_rules_apple" not in native.existing_rules():
         http_archive(
             name = "build_bazel_rules_apple",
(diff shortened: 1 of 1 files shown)


HannahShiSFB:
in favor of #35374

Issue #31772, opened 2022-11-30
gRPC extension prevents PHP execution


donis:
### What version of gRPC and what language are you using?
grpc 1.51.1
PHP 7.4.33

### What operating system (Linux, Windows,...) and version?
Linux

### What runtime / compiler are you using (e.g. python version or version of gcc)
Whatever `php:7.4-fpm-buster` docker image is running.

### What did you do?
Compiled a new gRPC php extension version and when running anything with PHP - we get this error:

### What did you expect to see?
normally running PHP process

### What did you see instead?
any PHP process hangs with this error message:
`E1130 11:14:11.870998498      29 thread_pool.cc:253]                   Waiting for thread pool to idle before forking`

rickyyx:
We running into this error logging as well. When we have multiple threads trying to do some computation. 

Piggybacking here as it seems to be the same issue - or happy to open another if needed. 

**What version of gRPC and what language are you using?**
grpc 1.51.1

**What operating system (Linux, Windows,...) and version?**
macOS Catalina

**What runtime / compiler are you using (e.g. python version or version of gcc)**
python 3.7

**What did you do?**
We have a workload where I believe multiple threads perform some computation (serialization) - we use grpc for cross process communication. I wasn't sure exactly what the callstack is when this logging happens though. 

**What did you expect to see?**
Normal run 

**What did you see instead?**
workload hanging with: 
```
E1207 01:40:20.292858000 4513988032 thread_pool.cc:253]                Waiting for thread pool to idle before forking
```

**Anything else we should know about your project / environment?**

Our job is here if this is helpful in anyway: https://buildkite.com/ray-project/oss-ci-build-branch/builds/1435#0184e9ed-2010-4a26-a83d-6b3918b006d8

scv119:
hi Grpc team, does this warrant yanking the release? It will cause downstream project hanging on Mac, as manifested in our CI.

stanley-cheung:
May I ask how can I reproduce this? Reason I am asking is we have tests running on the `php:7.4-buster` docker image and I was able to run some basic RPCs examples just fine. Is there something specific to `fpm`? 

Also, did this work in any previous release? Was it only 1.51.1 + fpm that breaks?

wouter-toppy:
I have the same error since december 7.

PHP 8.1.13
RPC 1.51.1

downgrated to 1.49.0 solves that issue, but getting random `Segmentation Fault` errors on CLI commands. which we didnt have before. 

[...]

PR #31994, opened 2022-12-29
[EventEngine] ThreadPool: pause closure execution on fork events

drfloob:
See #31885 and #31772 for context.

I'm not sure this is the right solution, but if it resolves Python's issues, we should try it.

Why this may be the wrong solution: in a post-fork child process for example, any pre-existing closures will likely (but not necessarily) want to do things it shouldn't, the post-fork child should essentially be a clean slate with respect to timers, polling, in-flight RPCs, etc.. 

Better potential solutions are much more complicated:

* ensuring no callback can possibly instantiate an ExecCtx during a fork event
* removing ExecCtx's special handling of fork events altogether
* removing ExecCtxs altogether

### DIFF
--- a/src/core/lib/event_engine/thread_pool.cc
+++ b/src/core/lib/event_engine/thread_pool.cc
@@ -128,8 +128,9 @@ bool ThreadPool::Queue::Step() {
   switch (state_) {
     case State::kRunning:
       break;
-    case State::kShutdown:
     case State::kForking:
+      return false;
+    case State::kShutdown:
       if (!callbacks_.empty()) break;
       return false;
   }


test/core/event_engine/thread_pool_test.cc
```diff
@@ -85,21 +82,6 @@ void ScheduleSelf(ThreadPool* p) {
   p->Run([p] { ScheduleSelf(p); });
 }
 
-TEST(ThreadPoolDeathTest, CanDetectStucknessAtFork) {
```
ctiller, line 88: Let's leave this in with a `DISABLED_` prefix to remind us to do something about this in the distant future.

ctiller (approved):

test/core/event_engine/thread_pool_test.cc
drfloob: Good idea. I've disabled it and added a note about why, and when it can be considered for re-enabling.

PR #31995, opened 2022-12-29
[ruby] TruffleRuby support: Always export _rb_tr_abi_version on truffle ruby

apolcyn:
Followup from https://github.com/grpc/grpc/pull/31970, specifically [this comment](https://github.com/grpc/grpc/pull/31970#pullrequestreview-1228356119) from @eregon:

"""
This doesn't seem correct as it will break the gem on TruffleRuby if I read the PR correctly. TruffleRuby has rb_tr_abi_version before RUBY_VERSION 3.2, specifically it's 3.0 in the 22.3 release currently for TruffleRuby and it already had it before that. We can consider TruffleRuby always had this symbol, versions before are not supported anymore.
"""

### DIFF
--- a/src/ruby/ext/grpc/ext-export-truffleruby.clang
+++ b/src/ruby/ext/grpc/ext-export-truffleruby.clang
@@ -1,1 +1,0 @@
-_Init_grpc_c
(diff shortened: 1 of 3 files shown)


eregon (approved):
Thanks!

veblush (approved):

PR #31999, opened 2023-01-02
Fix bazel distribtest

jtattermusch:
Broken by an omission in addressing https://github.com/grpc/grpc/pull/31366#discussion_r1055678526 (wrong copy paste that wasn't fixed).
Also improved formatting of the "FAILED_TESTS" variable for better readability of the results. (Now prints messages such as "Failed tests at version 4.2.2: buildtest distribtest_python" which IMHO is easier to parse by humans).

### DIFF
--- a/test/distrib/bazel/test_single_bazel_version.sh
+++ b/test/distrib/bazel/test_single_bazel_version.sh
@@ -54,14 +54,14 @@ export OVERRIDE_BAZEL_WRAPPER_DOWNLOAD_DIR=/tmp
 
 ACTION_ENV_FLAG="--action_env=bazel_cache_invalidate=version_${VERSION}"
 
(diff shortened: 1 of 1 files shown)


jtattermusch:
Looks like this makes bazel distribtest green again.

ctiller (approved):

ctiller:
Main reason for not making things blocking historically has been infrastructure flakiness, or other low rate flakiness that we hadn't gotten under control.

jtattermusch:
> Main reason for not making things blocking historically has been infrastructure flakiness, or other low rate flakiness that we hadn't gotten under control.

Another reason that is just as important is that the set of tests that runs on PR changes over time (jobs get added / removed / renamed) and it's a huge pain to keep the list of "required" tests in github's repo configuration up-to-date to match the tests that are supposed to run on master (or any given branch). So the tests marked as "required" are either jobs that have existed for a very long time and they are unlikely to be removed/renamed (e.g. sanity tests, portability tests) or tests that are of special importance (e.g. very important not to break them).

PR #32007, opened 2023-01-03
[bugfixes] Fix vulnerability for PyYAML

ralphchung:
Internal vulnerability report ([internal link](http://b/257242165)) recommends using `yaml.safe_load()` over `yaml.load()`. This PR simply find and replace those `load()` with `safe_load()`.


ctiller:
@gnossen - could you review this one, I'm out of the loop a bit on Python ecosystem stuff.
I think LGTM, but regardless low priority since the YAML usage we see is in developer tooling, not anything user or internet facing.

gnossen (approved):

PR #32010, opened 2023-01-03
Update default ruby version to 3.2 on linux/macos ruby builds

apolcyn:
As title.

We no longer support ruby < 2.7 since those versions have been EOL. Meanwhile, in order to start testing on ruby 3.2, we'll need to upgrade some dependencies to new versions which no longer support those EOL versions.


apolcyn:
Bazel distribtest failure is a pre-existing CBF

stanley-cheung (approved):

PR #32016, opened 2023-01-04
Remove objc cfstream tests from run_tests.py

jtattermusch:
Followup for https://github.com/grpc/grpc/pull/31555.

objc CFStream tests now run under bazel.
- there is no need for the complexity of "ios-test-cfstream-tests" under run_tests.py
- ios-test-cfstream-tests is the last test task under the grpc_basictests_objc_ios test job, which means we'll be able to get rid of the job completely (the job is slow so this is welcome)
- right now the actual test implementation in CFStreamClientTests.mm and CFStreamEndpointTests.mm is duplicated (there are 2 almost identical copies of each of the files in the repository).


jtattermusch:
CC @sampajano @HannahShiSFB

sampajano:
I'm not familiar with the test being removed. The rationale looks great to me and i'm happy to approve if this looks good to @HannahShiSFB.

HannahShiSFB:
According to https://github.com/bazelbuild/rules_apple/releases, "Add sanitizer support for iOS tests (https://github.com/bazelbuild/rules_apple/pull/1590)" is released, but we can not delete all these code before grpc updates bazel to 6.0.0

jtattermusch:
> According to https://github.com/bazelbuild/rules_apple/releases, "Add sanitizer support for iOS tests ([bazelbuild/rules_apple#1590](https://github.com/bazelbuild/rules_apple/pull/1590))" is released, but we can not delete all these code before grpc updates bazel to 6.0.0

It is not clear to me that we want to delay the simplifications until the point where we can run the sanitizers with bazel. I agree that sanitizers for cfstream tests are nice to have, but I'm not sure how many bugs they have caught in the past and whether radically simplifying the objc build stack isn't actually the primary goal (given that we're working on a pure-objC implementation and that the C-core based version of ObjC stack has suffered from many infrastructure problems for years - e.g. slow and unrealiable builds and tests and lot of complexity involved in the build in general).

sampajano:
> given that we're working on a pure-objC implementation

It's true that we're working on a pure-objC implementation but as some of the team members have pointed out in the past, until it has been completed (and we have migrated all google3 users), we still rely on the C-core based implementation as it's still depended by internal google teams.

> I'm not sure how many bugs they have caught in the past and whether radically simplifying the objc build stack isn't actually the primary goal

I'm not sure how many bugs they've caught in the past either; although, i'd error on the side of not removing tests unless they're **proven** to be ineffective.

I also don't know how "radical" we should be simplifying the objc build stack (although i'm 100% onboard with the direction) given the previously mentioned points.

> the C-core based version of ObjC stack has suffered from many infrastructure problems for years - e.g. slow and unrealiable builds and tests and lot of complexity involved in the build in general

I do not have personal experiences with them but i've definitely heard it many times.

I'll ask a few basic questions:

1. In this particular case, is the test being removed actively slowing down our build/release process, or has done so recently? Is it quantifiable?
2. Can we prove that this test has been ineffective and is not useful to keep?
3. Regarding _"until the point where we can run the sanitizers with bazel."_, is it a remote goal or achievable in the near term?

Hopefully by getting the answers it clarifies the situation better for us. Thanks! :)

Issue #31975, opened 2022-12-23
Grpc.Tools nuget package breaks builds on Windows_arm64 (see bottom for workaround)


ScottSedgwick:
### What version of gRPC and what language are you using?

Grpc.Tools Nuget Package version 2.5.0

### What operating system (Linux, Windows,...) and version?

Windows ARM64 (in a VM on Apple M1)

### What runtime / compiler are you using (e.g. python version or version of gcc)

dotnet 7.0.100

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.

We have a C# project with developers using Windows x64, and Windows ARM64 (because they have Mac M1s).
Any project that includes the Grpc.Tools nuget package compiles correctly on Win64, but fails on ARM64.

### What did you expect to see?

Compilation completes correctly.

### What did you see instead?

Compilation fails with this message:
```
C:\Users\scott\.nuget\packages\grpc.tools\2.50.0\build\_protobuf\Google.Protobuf.Tools.targets(280,5): error MSB6004: The specified task executable location "C:\Users\scott\.nuget\packages\grpc.tools\2.50.0\tools\windows_arm64\protoc.exe" is invalid. [C:\dev\ngf\src\Servers\Device\Protos\Cochlear.Ngf.Servers.Device.Protos.csproj]
```

### Anything else we should know about your project / environment?

Making a copy of the `C:\Users\scott\.nuget\packages\grpc.tools\2.50.0\tools\windows_x86` folder and renaming it to `C:\Users\scott\.nuget\packages\grpc.tools\2.50.0\tools\windows_arm86` fixes this problem.  Windows ARM seems to be capable of running the x86 tools in compatibility mode.

tonydnewell:
@ScottSedgwick as you have observed we do not support Windows ARM64 in Grpc.Tools.

Another workaround is to set the environment variables `PROTOBUF_PROTOC` and `GRPC_PROTOC_PLUGIN` to point to the x86 executables before running the build (see [here](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Tools/implementation_notes.md#protobuf-compiler-and-c-grpc-plugin-binaries)) 

e.g.
```
[...]

PR #32017, opened 2023-01-04
Grpc.Tools: Use x86 protoc binaries on arm64 Windows

tonydnewell:
A workaround for https://github.com/grpc/grpc/issues/31975 - use x86 protoc binaries on Windows arm64.

Unable to test as I don't have access to Windows-arm64.

Also not sure how NUnit PlatformAttribute handles arm64 (could not find any information).  It is possible that tests will fail if run on Windows-arm64 with full .NET framework if NUnit doesn't distinguish between x64 and arm64 (just identifies a 64 bit platform). Not sure that we run tests on Windows-arm64.

### DIFF
--- a/src/csharp/Grpc.Tools.Tests/ProtoToolsPlatformTaskTest.cs
+++ b/src/csharp/Grpc.Tools.Tests/ProtoToolsPlatformTaskTest.cs
@@ -79,6 +79,11 @@ namespace Grpc.Tools.Tests
                 {
                     Assert.AreEqual("x64", _task.Cpu);
                 }
+                // On windows arm64, x86 is used until a native protoc is shipped
+                else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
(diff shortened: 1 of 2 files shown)


jtattermusch:
Within the next 24 hours after merging, a Grpc.Tools nuget with the fix should show up in https://packages.grpc.io/ (and also the daily nuget feed: https://github.com/grpc/grpc-dotnet#grpc-nuget-feed).
It would be good to get a verification from the user.

tonydnewell:
@jtattermusch the Grpc.Tools nuget in the nightly builds has dummy binaries (zero length) for most platforms so cannot be used for testing.

jtattermusch:
> @jtattermusch the Grpc.Tools nuget in the nightly builds has dummy binaries (zero length) for most platforms so cannot be used for testing.

Filed https://github.com/grpc/grpc/issues/32179

PR #32019, opened 2023-01-05
StatsPluginEnd2EndTest: Add QueueOnceLoadBalancingPolicy to remove flakiness from test

yashykt:
Passes 1000x runs

### DIFF
--- a/test/core/util/test_lb_policies.cc
+++ b/test/core/util/test_lb_policies.cc
@@ -750,7 +750,7 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
 
 //
 // QueueOnceLoadBalancingPolicy - a load balancing policy that provides a Queue
-// PickResult atleast once, after which it delegates to PickFirst.
+// PickResult at least once, after which it delegates to PickFirst.
 //
 
 constexpr char kQueueOncePolicyName[] = "queue_once";
@@ -764,24 +764,29 @@ class QueueOnceLoadBalancingPolicy : public ForwardingLoadBalancingPolicy {
             std::move(args), "pick_first",
             /*initial_refcount=*/2) {}
 
-  ~QueueOnceLoadBalancingPolicy() override = default;
+  // We use the standard QueuePicker which invokes ExitIdleLocked() on the first
+  // pick.
+  void ExitIdleLocked() override {
+    bool needs_update = false;
+    {
+      MutexLock lock(&mu_);
+      needs_update = !std::exchange(seen_pick_queued_, true);
+    }
+    if (needs_update && state_to_update_.has_value()) {
+      channel_control_helper()->UpdateState(
+          state_to_update_->state, state_to_update_->status,
+          std::move(state_to_update_->picker));
+    }
+  }
+
+  bool SeenQueuedPick() {
+    MutexLock lock(&mu_);
+    return seen_pick_queued_;
+  }
 
   absl::string_view name() const override { return kQueueOncePolicyName; }
 
  private:
-  class Helper;
-
-  // An always queueing picker
-  class QueueingPicker : public SubchannelPicker {
-   public:
-    QueueingPicker(Helper* helper) : helper_(helper) {}
-
-    PickResult Pick(PickArgs /*args*/) override;
-
-   private:
-    Helper* helper_;
-  };
-
   class Helper : public ChannelControlHelper {
    public:
     Helper(RefCountedPtr<QueueOnceLoadBalancingPolicy> parent)
@@ -795,23 +800,18 @@ class QueueOnceLoadBalancingPolicy : public ForwardingLoadBalancingPolicy {
 
     void UpdateState(grpc_connectivity_state state, const absl::Status& status,
                      RefCountedPtr<SubchannelPicker> picker) override {
-      bool pick_queued;
-      {
-        MutexLock lock(&mu_);
-        pick_queued = pick_queued_;
-        // If a pick has not been queued yet, we need to save the update from
-        // PickFirst to propagate later.
-        if (!pick_queued) {
-          state_to_update_ = {state, status, std::move(picker)};
-        }
+      // If we've already seen a queued pick, just propagate the update
+      // directly.
+      if (parent_->SeenQueuedPick()) {
+        parent_->channel_control_helper()->UpdateState(state, status,
+                                                       std::move(picker));
+        return;
       }
-      if (!pick_queued) {
-        return parent_->channel_control_helper()->UpdateState(
-            GRPC_CHANNEL_CONNECTING, absl::OkStatus(),
-            MakeRefCounted<QueueingPicker>(this));
-      }
-      parent_->channel_control_helper()->UpdateState(state, status,
-                                                     std::move(picker));
+      // Otherwise, store the update in the LB policy, to be propagated later,
+      // and return a queueing picker.
+      parent_->state_to_update_ = {state, status, std::move(picker)};
+      parent_->channel_control_helper()->UpdateState(
+          state, status, MakeRefCounted<QueuePicker>(parent_->Ref()));
     }
 
     void RequestReresolution() override {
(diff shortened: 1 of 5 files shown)


test/core/util/test_lb_policies.cc
```diff
@@ -746,6 +748,173 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
...
+
+  absl::string_view name() const override { return kQueueOncePolicyName; }
+
+ private:
+  class Helper;
+
+  // An always queueing picker
+  class QueueingPicker : public SubchannelPicker {
+   public:
+    QueueingPicker(Helper* helper) : helper_(helper) {}
```
yashykt, line 777: @markdroth Please check if this raw pointer usage is safe..

test/core/util/test_lb_policies.cc
```diff
@@ -746,6 +748,173 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
...
+
+    RefCountedPtr<SubchannelInterface> CreateSubchannel(
+        ServerAddress address, const ChannelArgs& args) override {
+      return parent_->channel_control_helper()->CreateSubchannel(
+          std::move(address), args);
+    }
+
+    void UpdateState(grpc_connectivity_state state, const absl::Status& status,
+                     RefCountedPtr<SubchannelPicker> picker) override {
+      bool pick_queued;
```
markdroth, line 798: I think the logic here can be a bit simpler:

```
// If we've already seen a queued pick, just propagate the update directly.
if (parent_->SeenQueuedPick()) {
  parent_->channel_control_helper()->UpdateState(
      state, status, std::move(picker));
  return;
}
// Otherwise, store the update in the LB policy, to be propagated later, and return a queueing picker.
parent_->state_ = state;
parent_->status_ = status;
parent_->child_picker_ = std::move(picker);
parent_->channel_control_helper()->UpdateState(
    state, status, MakeRefCounted<QueuePicker>(parent_->Ref()));
```

The synchronization can happen inside of `SeenQueuedPick()`.

test/core/util/test_lb_policies.cc
```diff
@@ -746,6 +748,173 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
...
+    void AddTraceEvent(TraceSeverity severity,
+                       absl::string_view message) override {
+      parent_->channel_control_helper()->AddTraceEvent(severity, message);
+    }
+
+    void SetPickQueued() {
+      MutexLock lock(&mu_);
+      pick_queued_ = true;
+      if (state_to_update_.has_value()) {
+        // Propagate the update from PickFirst but do not do it inline to avoid
```
markdroth, line 838: I don't think there's any need to worry about sequencing here, since we'll be caching the update from the child, so we'll always return the most recent picker.

test/core/util/test_lb_policies.cc
```diff
@@ -746,6 +748,173 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
...
+        // deadlocks.
+        parent_->work_serializer()->Schedule(
+            [helper = parent_->channel_control_helper(),
```
markdroth, line 841: This is a little unsafe, because you're not holding ownership of the helper here.  I think it probably works out okay in practice, because you're holding a ref to the LB policy (which owns the helper) in the `ExecCtx` callback below, but I'd prefer not to see this pattern in our code.

As mentioned above, I think the picker should just hop into the `WorkSerializer` to invoke a method on the LB policy, and that callback should hold a ref to the LB policy, so it's explicitly safe.

test/core/util/test_lb_policies.cc
```diff
@@ -746,6 +748,173 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
...
+    grpc_event_engine::experimental::EventEngine* GetEventEngine() override {
+      return parent_->channel_control_helper()->GetEventEngine();
+    }
+
+    void AddTraceEvent(TraceSeverity severity,
+                       absl::string_view message) override {
+      parent_->channel_control_helper()->AddTraceEvent(severity, message);
+    }
+
+    void SetPickQueued() {
```
markdroth, line 834: This should be a method on the LB policy, not on the helper.  If you take my suggestion above about using the standard queueing picker, then this can probably go into `ExitIdleLocked()`; otherwise, it can go in a separate method on the LB policy.

Either way, I think the logic can be a lot simpler.  The caller should worry about hopping into the `WorkSerializer`, so all we should need to do here is this:

```
bool needs_update = false;
{
  MutexLock lock(&mu_);
  needs_update = !std::exchange(seen_pick_queued_, true);
}
if (needs_update) {
  channel_control_helper()->UpdateState(
      state_, status_, child_picker_);
}
```

test/core/util/test_lb_policies.cc
```diff
@@ -746,6 +748,159 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
   std::atomic<int>* pick_counter_;
 };
 
+//
+// QueueOnceLoadBalancingPolicy - a load balancing policy that provides a Queue
+// PickResult atleast once, after which it delegates to PickFirst.
```
markdroth, line 753: "at least" is two words, not one. :)

test/core/util/test_lb_policies.h
```diff
@@ -91,6 +91,9 @@ void RegisterFailLoadBalancingPolicy(CoreConfiguration::Builder* builder,
                                      absl::Status status,
                                      std::atomic<int>* pick_counter = nullptr);
 
+// Registers an LB policy called "queue_once" that queues atleast one pick, and
```
markdroth, line 94: "at least"

markdroth (commented):
This approach looks good!  Just a few comments about the implementation of the new LB policy.

test/core/util/test_lb_policies.cc
yashykt: haha we should probably enable a spell-check github action

test/core/util/test_lb_policies.cc
yashykt: Suggestion taken. Thanks!

test/core/util/test_lb_policies.cc
yashykt: Thanks! I didn't spot QueuePicker earlier. I'm just gonna reuse that as suggested

test/core/util/test_lb_policies.cc
```diff
@@ -746,6 +748,173 @@ class FailLbFactory : public LoadBalancingPolicyFactory {
...
+class QueueOnceLoadBalancingPolicy : public ForwardingLoadBalancingPolicy {
+ public:
+  QueueOnceLoadBalancingPolicy(Args args)
+      : ForwardingLoadBalancingPolicy(
+            std::make_unique<Helper>(
+                RefCountedPtr<QueueOnceLoadBalancingPolicy>(this)),
+            std::move(args), "pick_first",
+            /*initial_refcount=*/2) {}
+
+  ~QueueOnceLoadBalancingPolicy() override = default;
```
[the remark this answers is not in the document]
yashykt: Removed

test/core/util/test_lb_policies.cc
yashykt: Right, I didn't realize that the state could be synchronized through WorkSerializer by popping into it earlier from the picker, so we don't need to protect that with a mutex.

test/core/util/test_lb_policies.cc
yashykt: I like it! Suggestion taken

test/core/util/test_lb_policies.cc
yashykt: Right, I was actually just trying to synchronize access to `state_to_update_`. With the new logic of popping into the WorkSerializer from the picker, we just use the WorkSerializer to synchronize that access, and it all works.

yashykt (commented):
All suggestions taken

test/core/util/test_lb_policies.cc
```diff
@@ -831,55 +830,20 @@ class QueueOnceLoadBalancingPolicy : public ForwardingLoadBalancingPolicy {
...
-        false;  // Has a pick been queued yet
     RefCountedPtr<QueueOnceLoadBalancingPolicy> parent_;
   };
+  struct StateToUpdate {
+    grpc_connectivity_state state;
+    absl::Status status;
+    RefCountedPtr<SubchannelPicker> picker;
+  };
+  absl::optional<StateToUpdate> state_to_update_;
+  Mutex mu_;
```
markdroth, line 842: Now that we're accessing `seen_pick_queued_` only from inside the `WorkSerializer`, I don't think we actually need this mutex anymore.

test/core/util/test_lb_policies.cc
```diff
@@ -831,55 +830,20 @@ class QueueOnceLoadBalancingPolicy : public ForwardingLoadBalancingPolicy {
...
-    bool pick_queued_ ABSL_GUARDED_BY(mu_) =
-        false;  // Has a pick been queued yet
     RefCountedPtr<QueueOnceLoadBalancingPolicy> parent_;
   };
+  struct StateToUpdate {
+    grpc_connectivity_state state;
+    absl::Status status;
+    RefCountedPtr<SubchannelPicker> picker;
+  };
+  absl::optional<StateToUpdate> state_to_update_;
```
markdroth, line 841: I don't think we need to use `absl::optional<>` here.  There shouldn't be any case where we use these values where they are not already set.

markdroth (approved):
Looks great!  Remaining two comments are minor, feel free to merge after addressing.

test/core/util/test_lb_policies.cc
yashykt: You're right. I originally wanted to push the update in the constructor so that it happens asap, but that didn't work which is why I just moved the update to when we get an update from PickFirst.

test/core/util/test_lb_policies.cc
yashykt: Thanks!

PR #32022, opened 2023-01-05
Fix order issue in setting grpc_tcp_client_impl in api_fuzzer

yijiem:
`iomgr` sets the `grpc_tcp_client_impl` vtable [here](https://github.com/grpc/grpc/blob/f99b8b5bc4a89d37f86bf063999ec22e44312663/src/core/lib/iomgr/iomgr_posix.cc#L78) as part of the first `grpc_init` which overwrites the `fuzz_tcp_client_vtable` set by api_fuzzer. This caused inconsistent behavior between whether api_fuzzer is used to run one test case or multiple test cases (since later test case run will overwrite the vtable again and use the intended one).

### DIFF
--- a/test/core/end2end/fuzzers/api_fuzzer.cc
+++ b/test/core/end2end/fuzzers/api_fuzzer.cc
@@ -815,7 +815,6 @@ DEFINE_PROTO_FUZZER(const api_fuzzer::Msg& msg) {
   if (squelch && !grpc_core::GetEnv("GRPC_TRACE_FUZZER").has_value()) {
     gpr_set_log_function(dont_log);
   }
-  grpc_set_tcp_client_impl(&fuzz_tcp_client_vtable);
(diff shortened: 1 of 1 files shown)


yijiem:
To add more to this, it seems like we are not resetting the [iomgr_platform_vtable](https://github.com/grpc/grpc/blob/f99b8b5bc4a89d37f86bf063999ec22e44312663/src/core/lib/iomgr/iomgr_internal.cc#L32) during an actual `grpc_shutdown`. This causes `grpc_init` in subsequent test cases to skip overwriting `grpc_tcp_client_impl` vtable [here](https://github.com/grpc/grpc/blob/f99b8b5bc4a89d37f86bf063999ec22e44312663/src/core/lib/iomgr/iomgr.cc#L56-L58) (which explains the inconsistency between different runs).

PR #32027, opened 2023-01-05
Fix flakiness in xds_cluster_end2end_test: XdsTest/CdsTest.CircuitBreaking/V3

Vignesh2208:
Flakiness appears to be caused by immediately trying a new RPC after cancelling a previous one.

Link:https://source.cloud.google.com/results/invocations/1e9f67e9-7888-41c8-9a20-cf0475272e13/targets/%2F%2Ftest%2Fcpp%2Fend2end%2Fxds:xds_cluster_end2end_test@poller%3Dpoll/tests

### DIFF
--- a/test/cpp/end2end/xds/xds_cluster_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_cluster_end2end_test.cc
@@ -218,6 +218,10 @@ TEST_P(CdsTest, CircuitBreaking) {
                       "circuit breaker drop");
   // Cancel one RPC to allow another one through.
   rpcs[0].CancelRpc();
+  // Add a sleep here to ensure the RPC cancellation has completed correctly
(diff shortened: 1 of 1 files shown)


markdroth:
Can you say more about why you think we need this delay?  Note that calling `LongRunningRpc::CancelRpc()` will not return until the status has been returned to the application, so the call counter should definitely have been decremented by that point.  Or is there some subtle atomic thing going on here that means that the decremented value might not be visible to another thread right away?

Vignesh2208:
> Looks like the call tracker's Finish() method can wind up being called after status is returned the application in the cancellation case.
> 
> This is fine, assuming that you've run this 1000x on RBE for all of the {a,t,m,ub}san builds and verified that it fixes the flake.

Thanks I verified it by running 10K times.

PR #32028, opened 2023-01-05
[༺ EventEngine ༻] Specify requirements for Run* immediate execution

drfloob:
Also adds test suite tests that may catch non-conforming implementations (Run called ~3200 times in case it's non-deterministic).

### DIFF
--- a/include/grpc/event_engine/event_engine.h
+++ b/include/grpc/event_engine/event_engine.h
@@ -380,10 +380,10 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
   /// \a Closures scheduled with \a Run cannot be cancelled. The \a closure will
   /// not be deleted after it has been run, ownership remains with the caller.
   ///
-  /// Implementations must not immediately execute the closure in a blocking
-  /// manner on the same thread before the current thread stack is unwound. For
-  /// example, if the caller must release a lock before the closure can proceed,
-  /// running the closure immediately would cause a deadlock.
+  /// Implementations must not execute the closure in the calling thread before
+  /// \a Run returns. For example, if the caller must release a lock before the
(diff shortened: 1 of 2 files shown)


markdroth (approved):
Thanks for doing this!

Vignesh2208 (approved):

include/grpc/event_engine/event_engine.h
```diff
@@ -379,6 +379,11 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
   ///
   /// \a Closures scheduled with \a Run cannot be cancelled. The \a closure will
   /// not be deleted after it has been run, ownership remains with the caller.
+  ///
+  /// Implementations must not immediately execute the closure in a blocking
```
ctiller, line 383: "Implementations must not execute the closure in the calling thread before Run() returns"?

I think there's a few too many extra words here that make it less clear. I'd certainly strike 'in a blocking manner'.

drfloob: updated.

drfloob:
Regarding the test failures: The posix implementation complies with the spec, but the 1-second delay in thread startup under backlog, along with a race that rarely makes the thread pool stuck when no new work is being added, is causing the test to time out in 5% of runs, more so on constrained systems. Watering the tests down (10 concurrent runs, 10 iterations) was not sufficient to resolve the issues completely. I've got fixes for both issues coming with the work stealing changes, but for now I've disabled the tests. Let's land the documentation changes and work on the tests/implementations separately.

Also TODO: implement XFAIL without googletest support. It would be a better solution than disabling these for everyone.

gnossen:
I am deeply disappointed that "[༺ EventEngine ༻]" did not become the new official EventEngine PR tag.

PR #32026, opened 2023-01-05
Pin version of rubygems-update (on v1.46.x)

jtattermusch:
To unbreak macos artifact build on v1.46.x branch.

```
+ gem install rubygems-update
+ typeset result
+ typeset rvmrc
+ rvm_rvmrc_files=("/etc/rvmrc" "$HOME/.rvmrc")
+ [[ -n /Users/kbuilder ]]
+ [[ /Users/kbuilder/.rvmrc -ef /Users/kbuilder/.rvmrc ]]
+ rvm_rvmrc_files+=("${rvm_prefix}/.rvmrc")
+ for rvmrc in '"${rvm_rvmrc_files[@]}"'
+ [[ -s /etc/rvmrc ]]
+ true
+ for rvmrc in '"${rvm_rvmrc_files[@]}"'
+ [[ -s /Users/kbuilder/.rvmrc ]]
+ true
+ for rvmrc in '"${rvm_rvmrc_files[@]}"'
+ [[ -s /Users/kbuilder/.rvmrc ]]
+ true
+ unset rvm_rvmrc_files
+ command gem install rubygems-update
+ gem install rubygems-update
ERROR:  Error installing rubygems-update:
	The last version of rubygems-update (>= 0) to support your Ruby & RubyGems was 3.3.26. Try installing it with `gem install rubygems-update -v 3.3.26`
	rubygems-update requires Ruby version >= 2.6.0. The current ruby version is 2.5.0.0.
```

https://source.cloud.google.com/results/invocations/14d23e11-f883-4b69-8fd3-da20acfe0591

PR #32029, opened 2023-01-05
Disable ruby artifact builds on v1.46.x

apolcyn:
**Note this PR should not be up-merged to master**

Ruby artifact builds are broken on this branch which is preventing patch releases for other languages e.g. C#

Since we don't actually plan to do any patch releases on v1.46.x for ruby, let's just disable ruby builds on this branch

### DIFF
--- a/tools/internal_ci/macos/grpc_build_artifacts.sh
+++ b/tools/internal_ci/macos/grpc_build_artifacts.sh
@@ -23,7 +23,9 @@ cd $(dirname $0)/../../..
 
 export PREPARE_BUILD_INSTALL_DEPS_CSHARP=true
 export PREPARE_BUILD_INSTALL_DEPS_PYTHON=true
-export PREPARE_BUILD_INSTALL_DEPS_RUBY=true
+# Ruby builds are disabled on this branch because they're broken.
+# TODO(apolcyn): re-enable and fix if a ruby patch release on 1.46.x is needed
+export PREPARE_BUILD_INSTALL_DEPS_RUBY=false
 export PREPARE_BUILD_INSTALL_DEPS_PHP=true
 source tools/internal_ci/helper_scripts/prepare_build_macos_rc
 
@@ -35,9 +37,6 @@ python3.6 -m pip install -U cython setuptools==44.1.1 wheel --user
 python3.7 -m pip install -U cython setuptools==44.1.1 wheel --user
 python3.8 -m pip install -U cython setuptools==44.1.1 wheel --user
 python3.9 -m pip install -U cython setuptools==44.1.1 wheel --user
-
-gem install rubygems-update
-update_rubygems
 
 tools/run_tests/task_runner.py -f artifact macos ${TASK_RUNNER_EXTRA_FILTERS} || FAILED="true"
 
--- a/tools/run_tests/artifacts/artifact_targets.py
+++ b/tools/run_tests/artifacts/artifact_targets.py
@@ -486,12 +486,8 @@ def targets():
         PythonArtifact('windows', 'x64', 'Python38'),
         PythonArtifact('windows', 'x64', 'Python39'),
         PythonArtifact('windows', 'x64', 'Python310', presubmit=True),
-        RubyArtifact('linux', 'x86-mingw32', presubmit=True),
-        RubyArtifact('linux', 'x64-mingw32', presubmit=True),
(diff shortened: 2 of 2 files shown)


drfloob (approved):

drfloob:
cc @jtattermusch

drfloob:
Merging this will mean we should never upmerge 1.46.x, correct? Given that it's our LTS branch, we do risk disabling the ruby build accidentally in master if we land this here, and someone accidentally upmerges later on.

apolcyn:
My personal guess is that we'll end up with patches on this branch that we won't want to upstream, either way. For example, although more minor I think we also won't want to upstream https://github.com/grpc/grpc/pull/32026.

The version of ruby used in the scripts here is EOL so it's possible that other things involved in the ruby build will break sooner or later.

@jtattermusch if I'm reading things correctly, we actually disabled ruby macos artifact builds back in https://github.com/grpc/grpc/pull/29304, which was before this branch was cut. But we unnecessarily left ruby dependencies in the macos scripts.

jtattermusch:
> Merging this will mean we should never upmerge 1.46.x, correct? Given that it's our LTS branch, we do risk disabling the ruby build accidentally in master if we land this here, and someone accidentally upmerges later on.

IMHO The solution for this is to do a "proper" upmerge of 1.46.x proactively. That means we merge a change we only want in 1.46.x to that branch and immediately after that we do a manual upmerge and do it in a way that only the desired changes end up in master (e.g. we discard some of the changes as part of the merge commit). Any subsequent upmerges will then automatically do the right thing (since the unwanted changes will be considered as "already included" by git, even though we've edited them away).
One interesting question is what our current "squash and merge" way of merging PRs does to this approach - I think things are still going to work just fine, but I haven't actually tried this.

tools/run_tests/artifacts/artifact_targets.py
```diff
@@ -485,12 +485,8 @@ def targets():
         PythonArtifact('windows', 'x64', 'Python38'),
         PythonArtifact('windows', 'x64', 'Python39'),
         PythonArtifact('windows', 'x64', 'Python310', presubmit=True),
-        RubyArtifact('linux', 'x86-mingw32', presubmit=True),
```
jtattermusch, line 488: If removing the ruby artifact build on linux, distribtest_targets.py (and possibly package_targets.py) also need to be updated.

With this change, the Ruby Distribtests Linux are failing
https://source.cloud.google.com/results/invocations/4ea9f8c2-f310-4a9d-8d3e-ec1d5dc0ef26/targets/github%2Fgrpc%2Fdistribtests/tests
(some of them already are failing on v1.46.x but this change seems to make it worse).

tools/internal_ci/macos/grpc_build_artifacts.sh
```diff
@@ -23,7 +23,7 @@ cd $(dirname $0)/../../..
 
 export PREPARE_BUILD_INSTALL_DEPS_CSHARP=true
 export PREPARE_BUILD_INSTALL_DEPS_PYTHON=true
-export PREPARE_BUILD_INSTALL_DEPS_RUBY=true
```
jtattermusch, line 26: yes, you're right, I checked and there are no ruby artifact being built on mac in v1.46.x, so the ruby dependencies can be removed from this script. Changes to this file are a better version of https://github.com/grpc/grpc/pull/32026/files.

jtattermusch (approved):
Added a comment about removing the linux ruby artifact builds, but since the ruby linux distribtests are already broken on this branch, I'm ok with merging this PR.

tools/run_tests/artifacts/artifact_targets.py
apolcyn: Oh good catch. Removed distrib test targest

jtattermusch:
Hm, unfortunately with no artifacts/distribtests to build, the ruby distribtest on linux says "Nothing to build" which produces an error (it's a safety mechanism for not accidentally skipping important checks when wrong task filters are passed).

https://source.cloud.google.com/results/invocations/b9f75cad-55a8-4afa-aaf1-73b428c943d1/log

jtattermusch:
Still LGTM.

apolcyn:
@jtattermusch @drfloob I don't have permission to merge.

Please merge if you do and looks good.

PR #32039, opened 2023-01-07
Make grpc++_test target publicly visible

### LINKED ISSUE #31296: mock_stream.h is not externally visible when using Bazel
v1.49.1, C++ 

Linux

LLVM / gcc with Bazel

### What did you do?
Depend on the target `@com_github_grpc_grpc//:grpc++_test` in my `BUILD.bazel` file, see screenshot.

For things to compile.

### What did you see instead?
The following error message:
```
ERROR: /home/runner/work/pins-infra/pins-infra/p4_pdpi/BUILD.bazel:312:11: in cc_library rule //p4_pdpi:p4_runtime_session_mocking: target '@com_github_grpc_grpc//:grpc++_test' is not visible from target '//p4_pdpi:p4_runtime_session_mocking'. Check the visibility declaration of the former target if you think the dependency is legitimate
```

I am a Google SWE working on a project that is used both internally and in open source.

This issue seems to be similar to https://github.com/grpc/grpc/issues/20208 and https://github.com/grpc/grpc/pull/20369.

### Screenshot

image

yashykt:
Fix #31296

I think Craig's concern here was the stability of the APIs. We could potentially add a disclaimer at the very top of all these files that we do not guarantee the stability of these files or this build target. What do y'all think?

### DIFF
--- a/bazel/grpc_build_system.bzl
+++ b/bazel/grpc_build_system.bzl
@@ -97,7 +97,7 @@ def _update_visibility(visibility):
         "grpc_opencensus_plugin": PUBLIC,
         "grpcpp_gcp_observability": PUBLIC,
         "grpc_resolver_fake": PRIVATE,
-        "grpc++_test": PRIVATE,
+        "grpc++_test": PUBLIC,
         "http": PRIVATE,
         "httpcli": PRIVATE,
         "iomgr_timer": PRIVATE,


ctiller (changes requested):
I don't think adding a comment would materially affect the maintainability here.

We've made it public internally already - my recommendation would be to remove the annotation from the build system entirely and commit to maintaining this API as part of the public interface, but @markdroth should probably make the final call on that.

rdesgroppes:
Is there a suggested way to overcome the visibility issue?

PR #32049, opened 2023-01-09
Disable logging test

yashykt:
The logging tests currently flake ~1 out 100 runs. The issue is that the client-side logging filter is have trouble closing the send message pipe. Without the send message pipe being closed, half-close on the client side is not logged and the overall promise does not complete either.


jtattermusch:
Adding a "release blocker" label since we don't want the flake in the release branch (which is going to be cut soon).

ctiller (approved):

PR #32067, opened 2023-01-10
WRR: xDS LB policy support

markdroth:
Adds support for configuring the WRR LB policy via xDS, using the xDS config introduced in https://github.com/envoyproxy/envoy/pull/24520.


src/proto/grpc/testing/xds/v3/client_side_weighted_round_robin.proto
```diff
@@ -0,0 +1,67 @@
...
+  google.protobuf.Duration blackout_period = 3;
+
+  // If a given endpoint has not reported load metrics in this long,
+  // then we stop using the reported weight. This ensures that we do
+  // not continue to use very stale weights. Once we stop using a stale
+  // value, if we later start seeing fresh reports again, the
+  // blackout_period applies. Defaults to 3 minutes.
+  google.protobuf.Duration weight_expiration_period = 4;
+
+  // How often endpoint weights are recalculated.  Default is 1 second.
```
yousukseung, line 65: I think we can mention that it's lower-bounded by 100ms here.

yousukseung (approved):

src/proto/grpc/testing/xds/v3/client_side_weighted_round_robin.proto
markdroth: This file is just a local copy of the xDS proto file, used only for tests.  We unfortunately need this due to some circular dependency problems with protos whose BUILD rules depend on gRPC itself.  No one is going to actually look at the comments in this file, so there's no point in changing them.

PR #32071, opened 2023-01-11
Fix a transport leak in subchannel

yijiem:
This should fix b/264740459.


yijiem:
Adding more to this, there seems to be a transport leak detected by ClusterFuzz in b/264740459 and encountered locally when I ran `api_fuzzer` test at https://github.com/grpc/grpc/pull/32034. The transport leak could be naively fixed by something like https://github.com/grpc/grpc/compare/master...yijiem:grpc:transport_leak. However this change uses a `std::unique_ptr` with custom deleter to automatically destroy transport in error paths until the ownership is released to the channel stack in the success path.

src/core/ext/filters/client_channel/connector.h
```diff
@@ -48,16 +51,22 @@ class SubchannelConnector : public InternallyRefCounted<SubchannelConnector> {
...
-    grpc_transport* transport = nullptr;
+    TransportPointer transport;
     // Channel args to be passed to filters.
     ChannelArgs channel_args;
     // Channelz socket node of the connected transport, if any.
     RefCountedPtr<channelz::SocketNode> socket_node;
 
     void Reset() {
-      transport = nullptr;
+      transport.reset();
```
markdroth, line 69: Instead of using `unique_ptr<>` here, I suggest just changing this to the following:

```
if (transport != nullptr) {
  grpc_transport_destroy(transport);
  transport = nullptr;
}
```

I think it will make more sense to use a smart pointer in the future, when we change the transport API to be an idiomatic C++ API.

markdroth (commented):
I think the right fix here is closer to what you had in https://github.com/grpc/grpc/compare/master...yijiem:grpc:transport_leak.  I think we should just change subchannel.cc line 883 to call `connecting_result_.Reset()`.

src/core/ext/filters/client_channel/connector.h
yijiem: Makes sense! Done. Thanks for the review!

markdroth (approved):

PR #32077, opened 2023-01-11
[ObjC] CF EventEngine client

HannahShiSFB:
@sampajano


src/core/lib/event_engine/cf_engine/cf_engine.cc
```diff
@@ -89,12 +109,12 @@ std::unique_ptr<EventEngine::DNSResolver> CFEventEngine::GetDNSResolver(
   grpc_core::Crash("unimplemented");
 }
 
-void CFEventEngine::Run(EventEngine::Closure* /* closure */) {
-  grpc_core::Crash("unimplemented");
+void CFEventEngine::Run(EventEngine::Closure* closure) {
+  executor_->Run(closure);
```
sampajano, line 113: there's a null check here, do we need it?
https://github.com/grpc/grpc/blob/1bcaa3e9351a9b00f5ce2a4b68c5e3bf61f6be8d/src/core/lib/event_engine/posix_engine/posix_engine.cc#L301-L312

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+    return;
+  }
+
+  buffer->RemoveLastNBytes(buffer->Length() - read_size);
+  on_read(absl::OkStatus());
+}
+
+void CFStreamEndpoint::Write(absl::AnyInvocable<void(absl::Status)> on_writable,
+                             SliceBuffer* data, const WriteArgs* /* args */) {
+  write_event_.NotifyOn(new PosixEngineClosure(
```
drfloob, line 250: I'm a bit lost here, what is actually writing the data? I believe the lockfree event here is triggered when a write finishes, and no write has been initiated yet. Does this work?

test/core/event_engine/test_suite/BUILD
```diff
@@ -119,8 +120,10 @@ grpc_cc_test(
         "no_linux",
         "no_windows",
     ],
-    uses_polling = False,
+    uses_polling = True,
```
drfloob, line 123: This should be `False`. When `uses_polling = True`, the same test is replicated for every iomgr poller that works on the platform. We don't care about iomgr for these tests.

test/core/event_engine/test_suite/cf_event_engine_test.cc
```diff
@@ -19,15 +19,22 @@
...
   grpc::testing::TestEnvironment env(&argc, argv);
+  gpr_set_log_verbosity(GPR_LOG_SEVERITY_DEBUG);
+
   auto factory = []() {
     return std::make_unique<grpc_event_engine::experimental::CFEventEngine>();
   };
-  SetEventEngineFactories(factory, factory);
+  auto oracle_factory = []() {
+    return std::make_unique<
+        grpc_event_engine::experimental::PosixOracleEventEngine>();
```
drfloob, line 35: This may mean that the posix engine is used for the `Listener` implementation on whatever platform is running the iOS tests. Is that ok?

HannahShiSFB: Yes, client test needs a server which we don't have in cf_engine

test/core/event_engine/test_suite/BUILD
HannahShiSFB: I believe it caused additional tests with polling=none (or something similar) which fails, let me change it back and test it again

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+      WriteCallback, &cf_context);
+  CFReadStreamSetDispatchQueue(cf_read_stream_,
+                               dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0));
+  CFWriteStreamSetDispatchQueue(
+      cf_write_stream_, dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0));
+
+  if (!CFReadStreamOpen(cf_read_stream_)) {
+    auto status = CFErrorToStatus(CFReadStreamCopyError(cf_read_stream_));
+    on_connect(std::move(status));
+    delete this;
```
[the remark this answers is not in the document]
HannahShiSFB: I can do it here and when CFWriteStreamOpen fails, but not that in open_event_ closure which are executed after this function returns.

src/core/lib/event_engine/cf_engine/cf_engine.cc
HannahShiSFB: I can add the check but it won't help anything as
1. unlike that in PosixEnginePollerManager, here executor_ is created in engine constructor so will not be null 
2. if it were null then closures cannot run so it will stuck which is not better than crash where we have stack trace
3. can we instead somehow annotate it not null or assert here?

src/core/lib/event_engine/cf_engine/cf_engine.cc
```diff
@@ -72,14 +76,30 @@ CFEventEngine::CreateListener(
...
 CFEventEngine::ConnectionHandle CFEventEngine::Connect(
-    OnConnectCallback /* on_connect */, const ResolvedAddress& /* addr */,
-    const EndpointConfig& /* args */, MemoryAllocator /* memory_allocator */,
-    Duration /* timeout */) {
-  grpc_core::Crash("unimplemented");
+    OnConnectCallback on_connect, const ResolvedAddress& addr,
+    const EndpointConfig& /* args */, MemoryAllocator memory_allocator,
+    Duration timeout) {
+  auto addr_uri = ResolvedAddressToURI(addr);
+  gpr_log(GPR_INFO, "CFEventEngine::connect: %s", addr_uri.value().c_str());
```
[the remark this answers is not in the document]
HannahShiSFB: I don't think there is an ios specific one, I can use the same event engine category for logging

src/core/lib/event_engine/cf_engine/cf_engine.cc
```diff
@@ -72,14 +76,30 @@ CFEventEngine::CreateListener(
...
+    return {0, 0};
+  }
+
+  auto endpoint_ptr = new CFStreamEndpoint(
+      std::static_pointer_cast<CFEventEngine>(shared_from_this()),
+      std::move(memory_allocator));
+
+  endpoint_ptr->Connect(std::move(on_connect), addr, std::move(timeout));
+
+  return {reinterpret_cast<intptr_t>(endpoint_ptr), 0};
```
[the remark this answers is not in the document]
drfloob: I see, thanks for explaining. It should be valid to have an EventEngine instance that does not support connection cancellation, I imagine there may be cases where things like gRPC shutdown takes longer. We can try it without that support.

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  auto connect_timeout_timer = engine_->RunAfter(timeout, [this]() {
+    open_event_.SetShutdown(absl::DeadlineExceededError("Connect timed out"));
+  });
+
+  open_event_.NotifyOn(new PosixEngineClosure(
+      [this, on_connect = std::move(on_connect),
+       connect_timeout_timer](absl::Status status) mutable {
+        engine_->Cancel(connect_timeout_timer);
```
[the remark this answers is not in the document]
HannahShiSFB: Mmm, it sounds that I'll have to do some tracking or reference counting after all, even if we don't support cancel connection, to prevent from this being deleted because of connection error while the timeout timer is being fired.
I'll need sometime on this.

src/core/lib/event_engine/cf_engine/cftype_unique_ref.h
```diff
@@ -0,0 +1,76 @@
...
+
+#ifdef GPR_APPLE
+
+#include <CoreFoundation/CoreFoundation.h>
+
+namespace grpc_event_engine {
+namespace experimental {
+
+template <typename T>
+class CFTypeUniqueRef {
```
sampajano, line 26: @drfloob Hey AJ I wonder there are existing utilities like this in the grpc codebase?

This class seems quite generic to me so i'm hoping we could reuse if there exists one. Thanks!

drfloob: It looks like this class provides some sugar specifically for `CFTypeRef` objects (implicit conversions and output-parameter specializations). It looks to me like a `std::unique_ptr` could do this job, albeit without the sugar on top. I'm mostly ambivalent about it, but we don't have anything `CFTypeRef`-specific in the core codebase.

ctiller: This looks like `template <typename T> using CFTypeUniqueRef = std::unique_ptr<T, CFRelease>;`

HannahShiSFB: I tries with something like
```
template <typename T>
using CFTypeUniqueRef =
    std::unique_ptr<std::remove_cvref_t<decltype(*std::declval<T>())>,
                    decltype(& CFRelease)>;
```
Which is very difficult to use because:
1. Unlike typical unique_ptr use case, `T` (CFStringRef, CFErrorRef, etc.) is the pointer type thus the complex alias
2. Deleter (&CFRelease) needs to be passed everywhere creating a pointer
2. I don't do dereference (operator* and operator->) but only convert back and forth when using with CF APIs.
3. I need `operator&` for using as output parameter

HannahShiSFB: Alternative to this, I can also try warp a unique pointer for auto release then add implicit conversions and `operator&` if you prefer. It'll be something like
```
template <typename T>
class CFTypeUniqueRef {
public:
  // conversions, operator&
private:
  std::unique_ptr<std::remove_cvref_t<decltype(*std::declval<T>())>,
                    decltype(& CFRelease)> ref_;
};
```

HannahShiSFB:
@drfloob Hi AJ, I think I have addressed all the comments, could you take a look and review it again? Thank you

src/core/lib/event_engine/cf_engine/cf_engine.cc
drfloob: That's a fine justification. Thanks.

drfloob:
@HannahShiSFB it looks like the `EventEngineClientTest.MultipleIPv6ConnectionsToOneOracleListenerTest` test crashes on Mac with a SIGBUS error.  See the `Basic Tests C/C++ MacOS ` check. If it reproduces locally, the command will be something like `bazel test --config=dbg --test_env=GRPC_EXPERIMENTS=event_engine_client --test_env=GRPC_VERBOSITY=debug --test_filter='*MultipleIPv6ConnectionsToOneOracleListenerTest*' //test/core/event_engine/test_suite:cf_event_engine_test`.

src/core/lib/gprpp/time.h
```diff
@@ -38,7 +38,8 @@
     uint64_t now = grpc_core::Timestamp::FromTimespecRoundDown( \
                        gpr_now(GPR_CLOCK_MONOTONIC))            \
                        .milliseconds_after_process_epoch();     \
-    if ((now - prev.exchange(now)) > (n)*1000) {                \
+    if ((now - prev) > (n)*1000) {                              \
```
drfloob, line 41: I suggest reverting this change, explicit atomic operations are preferred in this codebase, and we use `atomic::exchange` a fair amount. This change removes an explicit call, and adds an implicit load and store.

test/core/event_engine/test_suite/BUILD
drfloob: Did you re-test this?

src/core/BUILD
```diff
@@ -2057,16 +2057,31 @@ grpc_cc_library(
...
+    srcs = [
+        "lib/event_engine/cf_engine/cf_engine.cc",
+        "lib/event_engine/cf_engine/cfstream_endpoint.cc",
+    ],
+    hdrs = [
+        "lib/event_engine/cf_engine/cf_engine.h",
+        "lib/event_engine/cf_engine/cfstream_endpoint.h",
+        "lib/event_engine/cf_engine/cftype_unique_ref.h",
+    ],
+    defines = ["GRPC_CFSTREAM=1"],
```
drfloob, line 2069: Is this necessary? `port_platform.h` should set this automatically for the right platforms. And further, we would have to be careful to ensure that no general target accidentally depends on `cf_event_engine` (even if it's unused), because the defines here will propagate to every rule that depends on it ... on every platform. https://bazel.build/reference/be/c-cpp#cc_binary.defines

HannahShiSFB: That's what I wondered. I initially tried to set it to the test target only but grpc_cc_test doesn't have it.
Still trying other configurations for running tests.
Besides, it appears some tests don't pass with GRPC_CFSTREAM with iomgr, still looking

HannahShiSFB:
> the EventEngineClientTest.MultipleIPv6ConnectionsToOneOracleListenerTest test crashes on Mac with a SIGBUS error

Passing locally, investigating

src/core/lib/gprpp/time.h
drfloob: It looks like https://github.com/grpc/grpc/pull/32204 fixes the bug as well, and only uses one `exchange`. That will likely land today. Thanks for catching the bug!

test/core/event_engine/test_suite/BUILD
```diff
@@ -82,6 +83,8 @@ grpc_cc_test(
     uses_polling = False,
```
HannahShiSFB, line 83: This fails with `GRPC_POLL_STRATEGY=none`

https://source.cloud.google.com/results/invocations/4cead0a6-9484-4fd7-80ec-2fac56cf8746/targets/github%2Fgrpc%2Frun_tests%2Fcpp_macos_dbg_native/tests

CC: @drfloob

HannahShiSFB:
@drfloob I noticed that https://github.com/grpc/grpc/pull/32253 completely disabled experimental on mobile platforms. How should I enable cf event engine in our ios tests in "src/objective-c/tests/"?

PS, I cannot enable it in mac tests as many of the tests are using socket files instead of ip:port, which is not supported by cfstream.

drfloob:
> How should I enable cf event engine in our ios tests in "src/objective-c/tests/"?

Pinging @ctiller for considerations into a partial rollback of https://github.com/grpc/grpc/pull/32253. The PR has no description, I presume this solved some actual bloat problems.

test/core/event_engine/test_suite/BUILD
drfloob: I see, thanks.

HannahShiSFB:
Android CI failure is not related.
All review comments addressed.
Will enable cf event engine in iOS tests in a separate PR once experimental features can be enabled.

drfloob:
@HannahShiSFB after chatting with @ctiller, we'll need another solution for enabling experiments on iOS - the way it's set up today resulted in 10s of MB of binary bloat on iOS, which caused some problems internally.

We can set it up differently for iOS, though, and instead of a runtime flag, add the ability to build gRPC with `bazel test --config=event_engine_client [targets]`. We would need to set up a separate CI job for it though, since it would build gRPC from scratch twice, and we'd risk making presubmit CI suite take much longer otherwise. @sampajano do you have experience setting up new pull request CI jobs?

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,294 @@
...
+                                     CFStreamEventType type,
+                                     void* client_callback_info) {
+  auto self = static_cast<CFStreamEndpoint*>(client_callback_info);
+
+  switch (type) {
+    case kCFStreamEventOpenCompleted:
+      self->open_event_.SetReady();
+      break;
+    case kCFStreamEventCanAcceptBytes:
+    case kCFStreamEventEndEncountered:
```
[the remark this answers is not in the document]
sampajano: Borrowing this thread: I wonder if `kCFStreamEventEndEncountered` should also trigger `write_event_` (which triggers a write), or `kCFStreamEventCanAcceptBytes` alone is sufficient?

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,343 @@
...
+    auto slice = data->RefSlice(i);
+    size_t written_size =
+        CFWriteStreamWrite(cf_write_stream_, slice.begin(), slice.size());
+
+    total_written_size += written_size;
+    if (written_size < slice.size()) {
+      SliceBuffer written;
+      data->MoveFirstNBytesIntoSliceBuffer(total_written_size, written);
+
+      write_event_.NotifyOn(new PosixEngineClosure(
```
sampajano, line 324: Am I reading it correctly, that this is recursively scheduling a write, if it cannot finish in one `CFWriteStreamWrite` call?

I wonder if this is susceptible to a race condition if `CFStreamEndpoint::Write` is called again and now there are 2 scheduled writes?

src/core/lib/event_engine/cf_engine/cftype_unique_ref.h
```diff
@@ -0,0 +1,76 @@
...
+    return old;
+  }
+
+  void reset(T other = nullptr) {
+    T old = cf_type_ref_;
+    cf_type_ref_ = other;
+    if (old) {
+      CFRelease(old);
+    }
+  }
```
sampajano, line 65: I asked ChatGPT(3.5) if there's any bug in this class (i have to say i love i can just do that 😂) and it says if you add a `if (cf_type_ref_ != other)` check it could prevent a potential issue if both are the same.. i don't know if it truly matters but FYI.. 😄

sampajano (commented):
@HannahShiSFB Hi Hannah, thanks SO MUCH for working on this change!! Looks fantastic!! Just a bunch of relatively minor questions and comments inline! Thanks again for working on this!!! ❤️

sampajano:
> @HannahShiSFB after chatting with @ctiller, we'll need another solution for enabling experiments on iOS - the way it's set up today resulted in 10s of MB of binary bloat on iOS, which caused some problems internally.
> 
> We can set it up differently for iOS, though, and instead of a runtime flag, add the ability to build gRPC with `bazel test --config=event_engine_client [targets]`. We would need to set up a separate CI job for it though, since it would build gRPC from scratch twice, and we'd risk making presubmit CI suite take much longer otherwise. @sampajano do you have experience setting up new pull request CI jobs?

@drfloob Sorry i forgot to reply here. As we discussed offline earlier, i do not have previous experiences with new CI jobs for grpc, but would be happy to work with you on figuring that out! Will reach out once this PR is merged. Thanks!

src/core/lib/event_engine/cf_engine/cf_engine.cc
```diff
@@ -72,14 +76,76 @@ CFEventEngine::CreateListener(
...
+  auto deadline_timer =
+      RunAfter(timeout, [handle, that = std::static_pointer_cast<CFEventEngine>(
+                                     shared_from_this())]() {
+        that->CancelConnectInternal(
+            handle, absl::DeadlineExceededError("Connect timed out"));
+      });
+
+  auto on_connect2 =
+      [that = std::static_pointer_cast<CFEventEngine>(shared_from_this()),
+       deadline_timer = std::move(deadline_timer), handle,
```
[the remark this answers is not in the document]
HannahShiSFB: Not really as deadline_timer is a plain struct, copy and move are the same. So I guess it's more like a documentation of moving the ownership into the lambda then.

src/core/lib/event_engine/cf_engine/cf_engine.cc
```diff
@@ -72,14 +76,76 @@ CFEventEngine::CreateListener(
...
+
+        {
+          grpc_core::MutexLock lock(&that->conn_mu_);
+          that->conn_handles_.erase(handle);
+        }
+
+        auto endpoint_ptr = reinterpret_cast<CFStreamEndpoint*>(handle.keys[0]);
+
+        if (!status.ok()) {
+          on_connect(std::move(status));
```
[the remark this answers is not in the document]
HannahShiSFB: I don't know what "valid but unspecified" state mean, but in any case we should not (and we didn't) access the moved-from object.

src/core/lib/event_engine/cf_engine/cf_engine.cc
```diff
@@ -72,14 +76,76 @@ CFEventEngine::CreateListener(
...
+        "Unknown connection handle: %s",
+        HandleToString<EventEngine::ConnectionHandle>(handle).c_str());
+    return false;
+  }
+  conn_handles_.erase(handle);
+
+  // keep the `conn_mu_` lock to prevent endpoint_ptr from being deleted
+
+  auto endpoint_ptr = reinterpret_cast<CFStreamEndpoint*>(handle.keys[0]);
+  return endpoint_ptr->CancelConnect(status);
```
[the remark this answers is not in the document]
HannahShiSFB: This will trigger the shutdown of open_event_ in CFStreamEndpoint, then the on_connect callback, it'll be deleted there.

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,343 @@
...
+
+absl::Status CFErrorToStatus(CFTypeUniqueRef<CFErrorRef> cf_error) {
+  CFErrorDomain cf_domain = CFErrorGetDomain((cf_error));
+  CFIndex code = CFErrorGetCode((cf_error));
+  CFTypeUniqueRef<CFStringRef> cf_desc = CFErrorCopyDescription((cf_error));
+  char domain_buf[256];
+  char desc_buf[256];
+  CFStringGetCString(cf_domain, domain_buf, 256, kCFStringEncodingUTF8);
+  CFStringGetCString(cf_desc, desc_buf, 256, kCFStringEncodingUTF8);
+  return absl::Status(absl::StatusCode::kUnknown,
```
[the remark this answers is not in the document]
HannahShiSFB: Don't know, this is from the old cfstream implementation

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,343 @@
...
+void CFStreamEndpoint::WriteCallback(CFWriteStreamRef stream,
+                                     CFStreamEventType type,
+                                     void* client_callback_info) {
+  auto self = static_cast<CFStreamEndpoint*>(client_callback_info);
+  GRPC_EVENT_ENGINE_ENDPOINT_TRACE(
+      "CFStreamEndpoint::WriteCallback, type: %lu, this: %p", type, self);
+
+  switch (type) {
+    case kCFStreamEventOpenCompleted:
+      self->open_event_.SetReady();
```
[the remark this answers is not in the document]
HannahShiSFB: In my test it's always read becomes ready first then write is ready. For any reason if read is not ready yet, it's unlikely to be an issue as client will send request first.
Old cfstream_handle have it in both read and write which means which ever happens first triggers the event which seems worked fine as well.

open_event is connected to the on_connect callback, it'll either be set ready (connected) or shutdown with a status (connection failed or canceled or timed out)

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
HannahShiSFB: "There can be at most one outstanding write per Endpoint at any given time", this is how c-core uses endpoint.

src/core/lib/event_engine/cf_engine/cftype_unique_ref.h
HannahShiSFB: It's actually right, to prevent self-reset

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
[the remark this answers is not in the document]
HannahShiSFB: I think kCFStreamEventCanAcceptBytes alone should be sufficient, the old cfstream does this way, borrowed from there

src/core/lib/iomgr/endpoint_cfstream.cc
```diff
@@ -240,7 +240,7 @@ static void WriteAction(void* arg, grpc_error_handle error) {
 }
 
 static void CFStreamRead(grpc_endpoint* ep, grpc_slice_buffer* slices,
-                         grpc_closure* cb, bool urgent,
+                         grpc_closure* cb, bool /*urgent*/,
```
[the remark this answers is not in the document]
sampajano: ah thanks for explaining!

src/core/lib/event_engine/cf_engine/cf_engine.cc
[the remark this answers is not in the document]
sampajano: i see! i'm ok with this if this is consistent with the rest of the grpc code base (if it is). :)

src/core/lib/event_engine/cf_engine/cf_engine.cc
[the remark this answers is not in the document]
sampajano: aha ok! i guess this is fine.. (i've seen this used in other classes too so i assume this is how it's typically used.. :))

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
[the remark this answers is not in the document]
sampajano: ok sg thanks :)

src/core/lib/event_engine/cf_engine/cf_engine.cc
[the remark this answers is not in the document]
sampajano: aha ok! just want to double check. thanks for explaining!

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
sampajano: aha ok got it! Good to note! Thanks for clarifying!

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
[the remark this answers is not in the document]
sampajano: aha ok! thanks for explaining!

Could you maybe put a little bit comment somewhere to make this assumption more clear? Thanks!

HannahShiSFB:
All comments addressed except the [grpc_event_engine_endpoint_data_trace](https://github.com/grpc/grpc/pull/32077#discussion_r1148519440) one.

src/core/lib/event_engine/cf_engine/cfstream_endpoint.cc
```diff
@@ -0,0 +1,343 @@
...
+    auto status = CFErrorToStatus(CFReadStreamCopyError(cf_read_stream_));
+    GRPC_EVENT_ENGINE_ENDPOINT_TRACE("CFStream read error: %s",
+                                     status.ToString().c_str());
+    on_read(status);
+    return;
+  }
+
+  buffer->RemoveLastNBytes(buffer->Length() - read_size);
+
+  if (grpc_event_engine_endpoint_data_trace.enabled()) {
```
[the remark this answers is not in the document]
drfloob: No, it's not necessary, probably not useful on iOS.

drfloob:
> As we discussed offline, i do not have previous experiences with new CI jobs for grpc, but would be happy to work with you on figuring that out! Will reach out once this PR is merged. Thanks!

@sampajano I put together a CI test that should work for you, https://github.com/grpc/grpc/pull/32782. And please refer to our offline chat for the other piece required to land it. I'll be afk until next week, hopefully this is sufficient to get you started with EventEngine CI tests on iOS.

sampajano:
@drfloob Thanks SOO much for helping putting together a change here!! I have asked @HannahShiSFB to help testing them. 😃 

@HannahShiSFB Maybe you could provide an update once you have a good idea on whether the tests are working, and if not, why. Thanks! 😃

sampajano:
> All comments addressed except the [grpc_event_engine_endpoint_data_trace](https://github.com/grpc/grpc/pull/32077#discussion_r1148519440) one.

@HannahShiSFB Thanks Hannah!!

So, after you've done fixing it, could you go over the comments again and ensure all open comments are either:

1) Addressed AND resolved
2) Addressed AND responded, and waiting for confirmation from reviewer?

I've still seeing some comments like [this one](https://github.com/grpc/grpc/pull/32077/files#r1152609051) and [this one](https://github.com/grpc/grpc/pull/32077/files#r1152171360) which are unclear whether they're addressed or not.

Thanks!!

HannahShiSFB:
Enabled cf engine in #32791 in ios and test passes.
Tests were flaky because c-core may delete an endpoint at any time, fixed by [use ref counting for CFStreamEndpointImpl](https://github.com/grpc/grpc/pull/32077/commits/ca4115094d1d5595b00e7baa2525ae18dc077901)
Stress run InteropTestsLocalCleartext 500 times and passing.

HannahShiSFB:
> I've still seeing some comments like [this one](https://github.com/grpc/grpc/pull/32077/files#r1152609051) and [this one](https://github.com/grpc/grpc/pull/32077/files#r1152171360) which are unclear whether they're addressed or not.

All comments addressed, added comment for first one, no change to the second (follow the old implementation)

sampajano:
> All comments addressed ...
> 
> Failed CI "Distribution Tests Python Windows" is not related

Woohoo! Thanks so much for all the hard work on pulling this through, Hannah!! 😃

@drfloob Hey AJ do you want to take a last pass? If not, this can be merged? Thanks!

drfloob (approved):
Looks good. Thanks for the hard work, Hannah!

sampajano:
> Looks good. Thanks for the hard work, Hannah!

Thanks for the hard work indeed, Hannah!! Please feel free to merge the PR whenever you feel ready to! Thanks!! 🎉 🥳 😃

---

@drfloob Also thanks so much for being super helpful during the review process too, AJ! 😃

PR #32084, opened 2023-01-12
Update boringssl-with-bazel readme

jtattermusch:
The docs change is extracted from https://github.com/grpc/grpc/pull/31869 and https://github.com/grpc/grpc/pull/31938.
The actual upgrade of boringssl is in progress, but in the meantime we can at least make sure the instructions are up-to-date.
I'll also update the internal counterpart (cl/501499368)


jtattermusch:
CC @HannahShiSFB @davidben  @agl

PR #32093, opened 2023-01-13
xds interop: Fix buildscripts not continuing on a failed test suite

sergiitk:
Apparently there's a difference between bash 3 and bash 4. OSX comes with bash 3 out-of-box, so for whoever wrote this logic it "worked on my machine" :)

Proof of concept fix: https://source.cloud.google.com/results/invocations/f076d25c-105b-48ee-a722-b17126b83604

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -149,6 +149,11 @@ config_setting(
 )
 
 config_setting(
+    name = "mac_arm64",
+    values = {"cpu": "darwin_arm64"},
+)
+
+config_setting(
     name = "use_strict_warning",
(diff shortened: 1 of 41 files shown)


gnossen (approved):
Bash version incompatibilities are the stuff of nightmares 🙂

sergiitk:
Changed the syntax per https://github.com/grpc/grpc-node/pull/2323#pullrequestreview-1248043432

sergiitk:
Updated the description with the root cause details. Thank you @murgatroid99 for figuring it out!

PR #32096, opened 2023-01-13
Add MakeChannelArgumentOption() for void*

yousukseung:


yousukseung:
It looks the sanity check failure is from src/core/lib/event_engine/cf_engine/cf_engine.cc and is unrelated.

ctiller (approved):

PR #32099, opened 2023-01-13
[EventEngine] Update EventEngine client experiment

drfloob:
This bumps the expiry timeline, and guards certain poller behavior behind the flag. The Epoll1 poller has been shown to not support fork events, which is a problem for Python. Fixes for that are incoming, but for now, the poller instantiation is guarded behind the flag as well.


drfloob:
CC @gnossen

Vignesh2208 (approved):

Issue #31738, opened 2022-11-22
gRPC server thread gets stuck after creating an Alarm object


ashu-ciena:
### What version of gRPC and what language are you using?
grpc 1.42, C++

### What operating system (Linux, Windows,...) and version?
Linux 5168 5.10.147-oneos-standard #1 SMP PREEMPT Mon Nov 21 04:26:32 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux

### What runtime / compiler are you using (e.g. python version or version of gcc)
g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44.0.3)

### What did you do?
Using a Go based gnmi client gnmi_cli, subscribed to a Path to a C++ based grpc/gnmi server.  After the user authentication, 
 created an Alarm object as below to return the tag of a C++ object from completion queue to proceed further.

m_alarm = new Alarm();
 if (m_alarm)
 {
    m_alarm->Set(m_cq.get(), gpr_now(GPR_CLOCK_MONOTONIC), this);
 }

### What did you expect to see?
I expected an immediate tag to the returned from completion queue which usually happens in good working case.
[Working_Log.txt](https://github.com/grpc/grpc/files/10068529/Working_Log.txt)

### What did you see instead?
The GRPC server gets stuck and the logs suggests, it has either cancelled the timer or it is stuck and not proceeding.
[Problem_Log.txt](https://github.com/grpc/grpc/files/10068534/Problem_Log.txt)

The good and problematic logs are attached.  gRPC process restarts resolves the issue. There is no segfault, so no backtrace.

### Anything else we should know about your project / environment?
Not yet.

ashu-ciena:
@yashykt Would you please take a look ?

yijiem:
Can you provide a complete minimal example that repros this issue? gRPC internally sets and cancels timers frequently, so it's not easy to distinguish the alarm timer from the rest just from the log. And for your example, the timer log should show the `SET` time < `now`, since the `SET` time happens before the `now` time of the logging, e.g.:

```
I1123 15:38:45.253072109  769743 timer_generic.cc:342]       TIMER 0x55555658b368: SET 5668 now 5679 call 0x55555658b3b0[0x5555555f6b47]
```

But I couldn't find any timer with this property in your logs.

ashu-ciena:
Hello @yijiem 
[...]

PR #32101, opened 2023-01-13
Add a test case for infinite past or unix epoch expiry in alarm_test.cc

yijiem:
We have seen users using infinite past timespec or unix epoch time_point to set an alarm to be fired immediately. See https://github.com/grpc/grpc/issues/31738#issuecomment-1382087585.

### DIFF
--- a/test/cpp/common/alarm_test.cc
+++ b/test/cpp/common/alarm_test.cc
@@ -266,6 +266,30 @@ TEST(AlarmTest, NegativeExpiry) {
   EXPECT_EQ(junk, output_tag);
 }
 
+// Infinite past or unix epoch should fire immediately.
+TEST(AlarmTest, InfPastExpiry) {
+  CompletionQueue cq;
+  void* junk = reinterpret_cast<void*>(1618033);
+  Alarm alarm;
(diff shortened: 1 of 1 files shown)


test/cpp/common/alarm_test.cc
```diff
@@ -266,6 +266,30 @@ TEST(AlarmTest, NegativeExpiry) {
...
+TEST(AlarmTest, InfPastExpiry) {
+  CompletionQueue cq;
+  void* junk = reinterpret_cast<void*>(1618033);
+  Alarm alarm;
+  alarm.Set(&cq, gpr_inf_past(GPR_CLOCK_REALTIME), junk);
+
+  void* output_tag;
+  bool ok;
+  CompletionQueue::NextStatus status =
+      cq.AsyncNext(&output_tag, &ok, grpc_timeout_seconds_to_deadline(1));
```
drfloob, line 279: On a heavily loaded system with low resources, or maybe running this 10,000 times, I do wonder if we'll see flakes due to the short deadline. It may be safer to extend the deadline to 10s, given we expect it to complete immediately regardless.

yijiem: Makes sense. Done.

drfloob (approved):

PR #32104, opened 2023-01-13
interop_matrix: Update matrix for grpc-java 1.45.3 release

ejona86:


ejona86:
So that I don't lose it (it is still running): https://sponge2.corp.google.com/2a59dd1f-4fc2-45a1-ab06-409fdefa2f82

ejona86:
Interop tests passed. This is could use a review.

PR #32114, opened 2023-01-17
Add an additional < marker in comment blocks after the documented functions

yijiem:
A user is seeing out-of-order documentation for these functions in https://github.com/grpc/grpc/issues/31738#issuecomment-1383861800. This seems to fix that.

### DIFF
--- a/include/grpc/support/time.h
+++ b/include/grpc/support/time.h
@@ -55,9 +55,9 @@ typedef struct gpr_timespec {
 
 /** Time constants. */
 GPRAPI gpr_timespec
-gpr_time_0(gpr_clock_type type); /** The zero time interval. */
-GPRAPI gpr_timespec gpr_inf_future(gpr_clock_type type); /** The far future */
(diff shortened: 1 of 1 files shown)


include/grpc/support/time.h
```diff
@@ -55,9 +55,9 @@ typedef struct gpr_timespec {
 
 /** Time constants. */
 GPRAPI gpr_timespec
-gpr_time_0(gpr_clock_type type); /** The zero time interval. */
-GPRAPI gpr_timespec gpr_inf_future(gpr_clock_type type); /** The far future */
-GPRAPI gpr_timespec gpr_inf_past(gpr_clock_type type);   /** The far past. */
+gpr_time_0(gpr_clock_type type); /**< The zero time interval. */
```
veblush, line 58: How about having `/** */` like other? Comments need to move to the front then.

veblush (approved):
LGTM.

include/grpc/support/time.h
yijiem: Done.

PR #32127, opened 2023-01-18
[EventEngine] Refactoring the EventEngine Test Suite: Part 1

drfloob:
This is a partial refactoring, shortened to support @ctiller's work improving fix_build_deps.

Known limitations:
* redundancy in test suite configuration
* ~temporary locations for some EventEngine test utils should likely be moved into the src tree~
* A few new TODOs, to be cleaned up in Part 2+


test/core/event_engine/event_engine_test_utils.cc
```diff
@@ -241,5 +217,29 @@ ConnectionManager::CreateConnection(std::string target_addr,
   return absl::CancelledError("Failed to create connection.");
 }
 
+// Returns a random message with bounded length.
+std::string GetNextSendMessage() {
```
Vignesh2208, line 221: Any specific reason why this function was moved to the end of this file from the beginning of the file ?

test/core/event_engine/event_engine_test_utils.h
```diff
@@ -132,7 +131,38 @@ class ConnectionManager {
   std::unique_ptr<EventEngine> oracle_event_engine_;
 };
 
+void AppendStringToSliceBuffer(SliceBuffer* buf, absl::string_view data);
+
+std::string ExtractSliceBufferIntoString(SliceBuffer* buf);
```
Vignesh2208, line 136: I think ExtractSliceBufferIntoString is declared twice: here and in line 136.

test/core/event_engine/test_suite/README.md
```diff
@@ -1,5 +1,7 @@
 A reusable test suite for EventEngine implementations.
 
+# Customizing tests for your EventEngine implementation
+
 To exercise a custom EventEngine, create a new bazel test target that links
 against the `//test/core/event_engine/test_suite:complete` library, and provide
```
Vignesh2208, line 6: //test/core/event_engine/test_suite:complete appears to be deleted. Does this line still hold true ?

test/core/event_engine/test_suite/tests/client_test.cc
```diff
@@ -44,10 +44,18 @@
...
-#include "test/core/event_engine/test_suite/event_engine_test.h"
-#include "test/core/event_engine/test_suite/event_engine_test_utils.h"
+#include "test/core/event_engine/event_engine_test_utils.h"
+#include "test/core/event_engine/test_suite/event_engine_test_framework.h"
 #include "test/core/util/port.h"
 
+namespace grpc_event_engine {
+namespace experimental {
+
+void InitClientTests() {}
```
Vignesh2208, line 54: What is expected to go in this function implementation ?

Vignesh2208 (commented):
Mostly lgtm. Left a few suggestions

test/core/event_engine/event_engine_test_utils.cc
drfloob: No, it was a merge artifact, we had both done similar changes here. I've moved it back up near the top.

test/core/event_engine/event_engine_test_utils.h
drfloob: Good catch, fixed

test/core/event_engine/test_suite/README.md
drfloob: Documentation updated, thanks.

test/core/event_engine/test_suite/tests/client_test.cc
drfloob: Nothing, unfortunately. This bit of code allows `fix_build_deps` to see a source file dependency between the custom test specifications and the test suite test implementations. They were previously only associated at link-time, which is fundamentally opposed to what `fix_build_deps` is trying to do. Without this hack, the tests themselves get removed from the test suite targets.

Vignesh2208 (approved):

PR #32136, opened 2023-01-18
Update third_party/protobuf to 3.21.12

jtattermusch:

- a few manual fixes have been added.
- Even though "protobuf@master" is currently broken, it's probably a good idea to upgrade protobuf to the latest released version in order to minimize the diff.


jtattermusch:
Python311 artifact build on windows failure is interesting:
https://source.cloud.google.com/results/invocations/435291ee-873b-4ca3-b8bd-94841beacf56/targets/github%2Fgrpc%2Fbuild_artifacts_python/tests;passed=true;group=tasks;test=build_artifact.python_windows_x64_Python311;row=4

jtattermusch:
The sanity test error seem to be due to a design flaw in `check_protobuf_pod_version.sh` which, takes `version1=$(git describe --tags | cut -f 1 -d'-')` 
```
+ '[' 21.12 '!=' 3.21.12 ']'
+ echo 'Protobuf version in src/objective-c/!ProtoCompiler.podspec does not match protobuf version in third_party/protobuf.'
Protobuf version in src/objective-c/!ProtoCompiler.podspec does not match protobuf version in third_party/protobuf.
```

jtattermusch:
> Python311 artifact build on windows failure is interesting: https://source.cloud.google.com/results/invocations/435291ee-873b-4ca3-b8bd-94841beacf56/targets/github%2Fgrpc%2Fbuild_artifacts_python/tests;passed=true;group=tasks;test=build_artifact.python_windows_x64_Python311;row=4



> Details

And it might have been a one-off flake.

jtattermusch:
CC @sampajano  for the `check_protobuf_pod_version.sh` fix.

tools/distrib/check_protobuf_pod_version.sh
```diff
@@ -22,6 +22,8 @@ pushd third_party/protobuf
 
 version1=$(git describe --tags | cut -f 1 -d'-')
 v1=${version1:1}
+# in case we got a 2-part "x.y" version number from the tag, convert it to version number in "3.x.y" format.
+v1=$(echo "$v1" | sed 's/^\([0-9]*\)\.\([0-9]*\)$/3.\1.\2/')
```
sampajano, line 26: Could you add a rationale on why the "3.x.y" format is chosen?

(i.e. maybe point to https://github.com/protocolbuffers/protobuf/releases if that's the reason?)

sampajano:
> CC @sampajano for the `check_protobuf_pod_version.sh` fix.

Thanks for cc! I'm not very familiar with protobuf versioning but i suppose this "3.x.y" format won't work forever.. hence i've added a comment for you to link to the source of why you're making this change so we can update it in the future when appropriate. Thanks!

tools/distrib/check_protobuf_pod_version.sh
jtattermusch: done.

gnossen (approved):

tools/distrib/check_protobuf_pod_version.sh
sampajano: Thanks so much for adding the detailed comments!

sampajano (approved):
Thanks a lot for the change! iOS related change LGTM thanks!

PR #32138, opened 2023-01-18
De-experimentalize compression

gnossen:
Removes all docstrings referencing compression as an experimental API.


XuanWang-Amos (approved):
Some python tests are failing but shouldn't be caused by this PR.

gnossen:
That flake is addressed by [this PR](https://github.com/grpc/grpc/pull/32140)

PR #32140, opened 2023-01-18
Fix test race condition on port binding

gnossen:
Spotted [here](https://source.cloud.google.com/results/invocations/eb046c1f-2328-45fd-8f8a-8c68c07183dd/details). It was previously possible for another process on the machine to take the server's port between when it went down and when it came back up. This change ensures that the port remains occupied throughout the duration of the test.


XuanWang-Amos (approved):
Interesting change

PR #32142, opened 2023-01-18
[EventEngine] Add ResolvedAddress default constructor

drfloob:


Vignesh2208 (approved):

ctiller (commented):
I think
```
char address_[MAX_SIZE_BYTES] = {};
```
has the same effect fwiw, and allow more inlining opportunities for the compiler

(it's also very unclear to me that it matters)

drfloob:
Going with your alternative, thanks.

> (it's also very unclear to me that it matters)

Vignesh got a bit hung up on some MSAN issues in another PR, code trying to use the array's garbage data. Giving this a sane default is a step. There are probably still some methods that don't respect the 0-sized address, which will need to be fixed.

PR #32147, opened 2023-01-19
Get some stats on disk usage for bazel RBE

yashykt:

### DIFF
--- a/tools/internal_ci/linux/grpc_bazel_rbe.sh
+++ b/tools/internal_ci/linux/grpc_bazel_rbe.sh
@@ -37,4 +37,4 @@ bazel_rbe/bazel_wrapper \
 
 # Print disk usage
 df -h
-du -h --max-depth=1 * | sort -rh | head -100
+du -h --max-depth=1 | sort -rh | head -100
--- a/tools/internal_ci/linux/grpc_bazel_rbe.sh
+++ b/tools/internal_ci/linux/grpc_bazel_rbe.sh
@@ -34,3 +34,7 @@ bazel_rbe/bazel_wrapper \
   $BAZEL_FLAGS \
   "$@" \
   -- ${BAZEL_TESTS:-//test/...}
+
+# Print disk usage
+df -h
+du -hs * | sort -rh | head -100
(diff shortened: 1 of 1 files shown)


tools/internal_ci/linux/grpc_bazel_rbe.sh
```diff
@@ -34,3 +34,7 @@ bazel_rbe/bazel_wrapper \
   $BAZEL_FLAGS \
   "$@" \
   -- ${BAZEL_TESTS:-//test/...}
+
+# Print disk usage
+df -h
+du -hs * | sort -rh | head -100
```
drfloob, line 40: `du -h --max-depth=1` will also capture hidden dirs. `.cache` and `.local` on my systems are generally always hefty, unclear what the build environment may be hiding where. It'll be interesting to see if the repo root takes up the majority of the space, too.

yashykt: but with max-depth, we won't get subdirectories correct?

drfloob:
The summary flag rolls up all subdirectories as well.

tools/internal_ci/linux/grpc_bazel_rbe.sh
```diff
@@ -34,3 +34,7 @@ bazel_rbe/bazel_wrapper \
   $BAZEL_FLAGS \
   "$@" \
   -- ${BAZEL_TESTS:-//test/...}
+
+# Print disk usage
+df -h
+du -h --max-depth=1 * | sort -rh | head -100
```
drfloob, line 40: should be `du -h --max-depth=1 .` ... as it is now, this will report 2 folders deep, and miss hidden folders in the cwd (I believe).

yashykt: you're right

drfloob (approved):

drfloob:
Commenting so we don't forget: this will mask build errors, we'll need to capture the build result without failing the script on build failures, and return it after the disk utilization commands are run.

jtattermusch:
note that you don't need to merge this, you can do experiments by running  the impacted jobs as ephemeral builds (see go/grpc-local-testing).

Also, I used this trick in the past with good success:
https://github.com/grpc/grpc/blob/17cadf868942df4d8cbebbd61afe7710125b1111/tools/internal_ci/helper_scripts/prepare_build_linux_rc#L47

PR #32156, opened 2023-01-19
CI: Move kokoro bazel output root to /tmpfs/bazel

yashykt:
More information on the options available here - https://bazel.build/remote/output-directories
Earlier, the output directory was going in /home/... which has limited space ~49GB as opposed to /tmpfs which has ~246GB.


veblush (approved):

tools/internal_ci/linux/grpc_bazel_rbe.sh
```diff
@@ -30,6 +30,7 @@ python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path bazel_
 
 bazel_rbe/bazel_wrapper \
   --bazelrc=tools/remote_build/linux_kokoro.bazelrc \
+  --output_user_root=/tmpfs/bazel \
```
jtattermusch, line 33: nice find and good short term fix, but we should probably address the issue in a more systemic way (we have many more bazel builds than just this).

PR #32159, opened 2023-01-20
[Build] tests: fix compilation on older macOS: add port_platform.h (to pull in __STDC_FORMAT_MACROS define)

barracuda156:
Some versions of MacOS (as well as some `glibc`-based platforms) require `__STDC_FORMAT_MACROS` to be defined prior to including `<cinttypes>` or `<inttypes.h>`, otherwise build fails with undeclared `PRIdMAX`, `PRIdPTR` etc. 

See note here: https://en.cppreference.com/w/cpp/types/integer

label: (release notes: no)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,6 +312,24 @@ if(UNIX AND NOT HAIKU)
   endif()
 endif()
 
+include(CheckCXXSourceCompiles)
+
+if(UNIX OR APPLE)
+  # Some systems require the __STDC_FORMAT_MACROS macro to be defined
+  # to get the fixed-width integer type formatter macros.
+  check_cxx_source_compiles("#include <inttypes.h>
+  #include <cstdio>
+  int main()
+  {
+    int64_t i64{};
+    std::printf(\"%\" PRId64, i64);
+  }"
+  HAVE_STDC_FORMAT_MACROS)
+  if(NOT HAVE_STDC_FORMAT_MACROS)
+    add_definitions(-D__STDC_FORMAT_MACROS)
+  endif()
+endif()
+
 # configure ccache if requested
 include(cmake/ccache.cmake)
 
--- a/templates/CMakeLists.txt.template
+++ b/templates/CMakeLists.txt.template
@@ -338,6 +338,24 @@
     endif()
   endif()
 
+  include(CheckCXXSourceCompiles)
+
+  if(UNIX OR APPLE)
+    # Some systems require the __STDC_FORMAT_MACROS macro to be defined
+    # to get the fixed-width integer type formatter macros.
+    check_cxx_source_compiles("#include <inttypes.h>
+    #include <cstdio>
+    int main()
+    {
+      int64_t i64{};
+      std::printf(\"%\" PRId64, i64);
+    }"
+    HAVE_STDC_FORMAT_MACROS)
+    if(NOT HAVE_STDC_FORMAT_MACROS)
+      add_definitions(-D__STDC_FORMAT_MACROS)
+    endif()
+  endif()
+
   # configure ccache if requested
   include(cmake/ccache.cmake)
 


veblush:
I'm sure this change is desirable. I found this [doc](https://www.gnu.org/software/gnulib/manual/html_node/inttypes_002eh.html) indicating this macro is only required for pre C++11 compilers but gRPC requires C++14. If it's the case, I don't think we can accept this PR.

barracuda156:
That is not the case. It is required on MacOS with the latest GCC and some versions of Clang, all of which support modern C++.

A similar patch was recently merged to ICU (that is how I recognized the error when encountered it myself, with gcc12).

veblush:
@barracuda156 Would you share that information like one that ICU got recently? and this is more likely a GCC issue?

barracuda156:
> @barracuda156 Would you share that information like one that ICU got recently? and this is more likely a GCC issue?

@veblush Sure: https://github.com/unicode-org/icu/pull/2250/commits/c9e1c1343a7b25510462afea36c83260ee6a6da9

P. S. No, it is not GCC-specific.

veblush:
As those macros are defined in https://github.com/grpc/grpc/blob/dd07fd8669d4dc0a8c51f4d8841e1d6c8710b371/include/grpc/support/port_platform.h#L760-L762 

you may want to mare sure that `port_platform.h` is included in those sources.

barracuda156:
> As those macros are defined in
> https://github.com/grpc/grpc/blob/dd07fd8669d4dc0a8c51f4d8841e1d6c8710b371/include/grpc/support/port_platform.h#L760-L762
> 
> you may want to mare sure that `port_platform.h` is included in those sources.

@veblush So do you suggest including the said header instead in all affected files? https://github.com/grpc/grpc/pull/32159/files
If yes, I can do that, no issues.

barracuda156:
(Will try that option soon, gimme a bit of time, been busy.)

veblush:
> @veblush So do you suggest including the said header instead in all affected files? https://github.com/grpc/grpc/pull/32159/files If yes, I can do that, no issues.

Yes, exactly.

barracuda156:
> > @veblush So do you suggest including the said header instead in all affected files? https://github.com/grpc/grpc/pull/32159/files If yes, I can do that, no issues.
> 
> Yes, exactly.

@veblush Ok, will deal with this shortly.

barracuda156:
@veblush Sources changed since then, but I have dropped `__STDC_FORMAT_MACROS` now and added `grpc/support/port_platform.h` wherever it was missing.

veblush:
All files under `third_party` shouldn't be modified so you need to revert them.

barracuda156:
> All files under `third_party` shouldn't be modified so you need to revert them.

That will likely leave them broken, but sure, I will revert those soon.

barracuda156:
@veblush How about if we drop all per-file trickery and simply add the following check to CMakeLists? https://github.com/kcat/openal-soft/commit/118c729680d6664f793f8d88ff0b7548137847d3
That gonna fix everything at once, including third party, and no need to bother if some newly added files may need the fix too.

barracuda156:
@veblush Yes, sure, will do it today. (Sorry for a bit of a delay, been quite busy with other stuff.)

veblush:
Thanks for the update!

I triggered CI to run all tests. One thing it just found is that you also need to make the same change against `templates/CMakeLists.txt.template` which is a master template of `CMakeLists.txt`.

barracuda156:
> Thanks for the update!
> 
> I triggered CI to run all tests. One thing it just found is that you also need to make the same change against `templates/CMakeLists.txt.template` which is a master template of `CMakeLists.txt`.

@veblush Thank you. Will be done today.

veblush:
LGTM. Thank you for making this change!

yashykt (approved):

PR #32089, opened 2023-01-12
[ruby]: add pre-compiled binaries for ruby 3.2; drop them for ruby 2.6

apolcyn:
Ruby 3.2 came out at the end of 2022, and rake-compiler-dock just added support in 1.3.0. Ruby 2.6 is now EOL, so stop building pre-compiled targets for it.

Note:

- Test coverage is primarily being added in https://github.com/grpc/grpc/pull/31991. Besides that, we'll need to follow this up with new distrib tests as well.
- I did some cleanup to the `Rakefile` by removing the `if RUBY_PLATFORM =~ /darwin/` - we no longer build packages on macos hosts so there's no need for it.
- A hack is made to `extconf.rb` to work around https://github.com/rake-compiler/rake-compiler/issues/210

PR #32160, opened 2023-01-20
[ruby] Undrop ruby 2.6

apolcyn:
Follow-up from https://github.com/grpc/grpc/pull/32089

We can still build pre-compiled 2.6 gems after adding 3.2 support.


jtattermusch (approved):
LGTM on green tests.

apolcyn:
Ruby failure looks like pre-existing flake: b/266212253

PR #32161, opened 2023-01-20
[example] Fix bad example

molto0504:
Channel object has no method ``stop``, it supposed to be ``close`` here.


XuanWang-Amos:
Nice catch and thanks for the change!

antonkesy (approved):

XuanWang-Amos (approved):

PR #32170, opened 2023-01-21
Print useful logs when encountering ENOBUFS with tx-zero copy

Vignesh2208:


arjunroy (changes requested):
I think we can simplify the logic a bit.

void MaybeLogULimit() {
#ifdef GRPC_LINUX_ERRQUEUE
GRPC_LOG_EVERY_N_SEC(
            1,
            "Tx0cp encountered an ENOBUFS error possibly because one or "
            "both of RLIMIT_MEMLOCK or hard memlock ulimit values are too "
            "small for the intended user. Current system value of "
            "RLIMIT_MEMLOCK is %" PRIu64 " and hard memlock ulimit is %" PRIu64
            ".Consider increasing these values appropriately for the intended "
            "user.",
            GetRLimitMemLockMax(), GetUlimitHardMemLock());
#endif
}

bool UpdateZeroCopyOMemStateAfterSend(bool seen_enobuf) {
  if (seen_enobuf) {
    MaybeLogUlimit();
  } 
}

Vignesh2208:
> void MaybeLogULimit() {
> #ifdef GRPC_LINUX_ERRQUEUE
> GRPC_LOG_EVERY_N_SEC(
>             1,
>             "Tx0cp encountered an ENOBUFS error possibly because one or "
>             "both of RLIMIT_MEMLOCK or hard memlock ulimit values are too "
>             "small for the intended user. Current system value of "
>             "RLIMIT_MEMLOCK is %" PRIu64 " and hard memlock ulimit is %" PRIu64
>             ".Consider increasing these values appropriately for the intended "
>             "user.",
>             GetRLimitMemLockMax(), GetUlimitHardMemLock());
> #endif
> }

I updated the PR to simplify the logic a bit but I dont think we need to log the rlimit values every time we see an enobuf. I think we need to log only when we are resource constrained i.e only when there are no un-acked z-copy records but we still see an ENOBUFS error.

arjunroy:
> > void MaybeLogULimit() {
> > #ifdef GRPC_LINUX_ERRQUEUE
> > GRPC_LOG_EVERY_N_SEC(
> > 1,
> > "Tx0cp encountered an ENOBUFS error possibly because one or "
> > "both of RLIMIT_MEMLOCK or hard memlock ulimit values are too "
> > "small for the intended user. Current system value of "
> > "RLIMIT_MEMLOCK is %" PRIu64 " and hard memlock ulimit is %" PRIu64
> > ".Consider increasing these values appropriately for the intended "
> > "user.",
> > GetRLimitMemLockMax(), GetUlimitHardMemLock());
> > #endif
> > }
> 
> I updated the PR to simplify the logic a bit but I dont think we need to log the rlimit values every time we see an enobuf. I think we need to log only when we are resource constrained i.e only when there are no un-acked z-copy records but we still see an ENOBUFS error.

My concern is that we should not be adding additional state variables in a change where we just want to add debug info. If you can architect this patch such that the 'constrained' variable is not added (and the control flow does not change, beyond occasionally outputting a log statement) then I think it is enough.

If you're concerned about not logging it every time, we can add a log once statement instead (since it's unlikely that ulimit is changing that often, and we can just search for its existence at all in a log file rather than needing to see it multiple times).

PR #32178, opened 2023-01-23
Enforce POSIX-Sync on Linux

### LINKED ISSUE #32176: gRPC is blocking after system time changes
1.36.4 (but seems to be the same in newer versions)

Linux (built with Yocto Hardknott)

gcc

### What did you do?
I created a channel state observer which should be called whenever the channel state changes or at least every two seconds. As my code is running on an embedded device changes of the system time - either by setting it hard or by drifting can occur. According to the proposition in https://github.com/grpc/grpc/issues/22955 I made sure to specify the deadline not in system time, but as a monotonic time.

Here's a snippet of the most important methods within the class `ChannelStateObserver`:
```cpp
/**
 * mState is an instance of type grpc_connectivity_state
 */
ChannelStateObserver::ChannelStateObserver(std::shared_ptr<grpc::Channel> channel, grpc::CompletionQueue& cq)
    : mChannel(channel)
    , mCompletionQueue(&cq)
    , mState(mChannel->GetState(true))
{
  registerStateChange();
}

void ChannelStateObserver::channelStateChanged()
{
  std::cout << "callback" << std::endl;
  auto state = mChannel->GetState(true);
  if (state != mState) {
    std::cout << "channel state changed to " << state << std::endl;    
    mState = state;
  }
  registerStateChange();
}

void ChannelStateObserver::registerStateChange()
{
  mChannel->NotifyOnStateChange(
      mState, GrpcTimeUtil::makeSteadyDeadline(std::chrono::seconds { 2 }), mCompletionQueue, static_cast<ITag*>(this));
}
```
And the class `GrpcTimeUtil`:
```cpp
/**
 * By default gRPC is based on the wall clock. This leads to the fact that timeouts don't work properly
 * in case the wall clock ist changed.
 * When specifing deadlines they should be based on the monotonic clock. Then gRPC will treat them the right way.
 *
 * see: https://github.com/grpc/grpc/issues/22955
 * @param timeoutDuration Duration to be set in the deadline
 * @return Deadline based on the monotonic clock
 */
template<class Rep, class Ratio>
inline gpr_timespec makeSteadyDeadline(std::chrono::duration<Rep, Ratio> timeoutDuration)
{
  using namespace std::chrono;
  nanoseconds timeoutAsNanoseconds = duration_cast<nanoseconds>(timeoutDuration);
  seconds timeoutAsSeconds = duration_cast<seconds>(timeoutAsNanoseconds);
  nanoseconds timeoutFractionalNanoseconds = timeoutAsNanoseconds - timeoutAsSeconds;
  gpr_timespec timeoutGpr = gpr_now(GPR_CLOCK_MONOTONIC);
  timeoutGpr.tv_sec += timeoutAsSeconds.count();
  timeoutGpr.tv_nsec += timeoutFractionalNanoseconds.count();
  return timeoutGpr;
}
```

In my main loop I'm blocking on the completion queue until I get any callback:
```cpp
void* tag = nullptr;
bool readSuccessful = false;
while (mCompletionQueue->Next(&tag, &readSuccessful)) {
  auto type = static_cast<ITag*>(tag)->getType();
  if (type == ITag::Type::CHANNEL_STATE_OBSERVER) {
    IChannelStateObserver* observer = static_cast<IChannelStateObserver*>(tag);
    observer->channelStateChanged();
  } else {
    // handle other tags
  }
}
```

I'm expecting to see the debug-output `callback` at least every two seconds or sooner, in case the channel state actually changes.

### What did you see instead?
Exactly what I expect - until the moment I start changing the system time. In case I set the system time back for 1 minute, I don't see any output for exactly 1 minute. If I set the system time back for 2 minutes, I don't see any output for 2 minutes....

Even if the channel state actually changes within that minute or two, I don't get any callback at all: gRPC seems to be completely blocked. Even if I set the environment variables `GRPC_VERBOSITY=debug GRPC_TRACE=all` I don't get any output until that minute or two are passed.
Afterwards everything works as expected again.

This seems to be quite a low level problem and I highly assume not necessarily related to my use case with the channel state observer.

thomkopp:
This is needed in order to handle system time changes properly. As SYNC-Abseil and Abseil in general (see https://github.com/abseil/abseil-cpp/issues/172) is exclusively based on system-time, we need to use POSIX-Sync which uses the monotonic clock for timeouts

Fixes #32176

### DIFF
--- a/include/grpc/support/port_platform.h
+++ b/include/grpc/support/port_platform.h
@@ -41,6 +41,11 @@
 #if defined(__APPLE__)
 // This is disabled on Apple platforms because macos/grpc_basictests_c_cpp
 // fails with this. https://github.com/grpc/grpc/issues/23661
+#elif defined(__linux__)
+// This is disabled on Linux as Abseil's time is exclusively based on the system time
+// which leads to gRPC being completely blocked in case the system clock is set back.
+// On linux we need to use SYNC_POSIX which uses the monotonic clock.
+// https://github.com/grpc/grpc/issues/32176
 #else
 #define GPR_ABSEIL_SYNC 1
 #endif


thomkopp:
I can't decide if this is release nodes relvant or not. This should be set by one of the maintainers.

veblush:
Thank you for the PR but our plan is to migrate to Abseil one. So we want to keep this.

thomkopp:
> Thank you for the PR but our plan is to migrate to Abseil one. So we want to keep this.

Thanks for your feedback. That’s the impression I got by looking at the code. But I couldn’t find a way on how to make abseils condition variable (or better Abseils time implementation) use a monotonic clock as I referred in my comment above. Feel free to refer me to a link on how to do this and I‘ll adapt it.

Additionally please note that due to another bug Abseil sync has also been disabled on all __APPLE__ devices.

It’s not such an exotic use case that time changes can happen on embedded devices. We need a reliable solution here.

Issue #31483, opened 2022-10-27
Consider having separate timeout for wait_for_ready independent of RPC timeout


rlubke:
### Is your feature request related to a problem? Please describe.
I have a long lived bidirectional stream that ideally should have no timeout.   However, similarly, I don't want an indefinite wait_for_ready either. 

### Describe the solution you'd like
Ideally a timeout that can be applied to the wait_for_ready functionality that is independent of the RPC timeout.

gnossen:
@ejona86 @markdroth  Is this something that any other language has?

markdroth:
No, this is not something we support in any language, and we have no plans to do so.  I don't think this is a common enough use case that it makes sense to support this as a first-class citizen in our API.

The application can't *exactly* do this itself, but it can do things that are probably close enough for what it actually cares about.  For example, if what the application really cares about is whether the server has seen the start of the stream, if it can ensure that the server sends initial metadata as soon as the stream starts, and then the client can set a timer when it starts the stream, and if it does not receive initial metadata from the server by the time the timer fires, then it can cancel the call.

XuanWang-Amos:
As Mark mentioned, we'll create an example for this use case too.

PR #32183, opened 2023-01-24
Add wait_for_ready with client timeout example

XuanWang-Amos:
### Description
Adding an example for `wait_for_ready` with client side timeout. (Requested from this issue: https://github.com/grpc/grpc/issues/31483)

### Details
For stream server, if client set `wait_for_ready` but server never actually starts, client will wait indefinitely, this example will do the following steps to set a timeout on client side:
 1. Set server to return `initial_metadata` once it receives request.
 2. Client will set a timer (customized client timeout) waiting for `initial_metadata`.
 3. Client will timeout if it didn't receive `initial_metadata`.

### DIFF
--- a/examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
+++ b/examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
@@ -11,10 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""The Python client of utilizing wait-for-ready flag with client time out."""
+"""The Python client of utilizing wait-for-ready flag with client time out.
 
+For stream server, if client set wait_for_ready but server never actually starts,
+client will wait indefinitely, this example will do the following steps to set a
+timeout on client side:
+1. Set server to return initial_metadata once it receives request.
+2. Client will set a timer (customized client timeout) waiting for initial_metadata.
+3. Client will timeout if it didn't receive initial_metadata.
+"""
 import logging
 import threading
+from typing import Sequence, Tuple
 
 import grpc
 
@@ -26,7 +34,8 @@ _LOGGER.setLevel(logging.INFO)
 
 
 def wait_for_metadata(response_future, event):
-    for key, value in response_future.initial_metadata():
+    metadata: Sequence[Tuple[str, str]] = response_future.initial_metadata()
+    for key, value in metadata:
         print('Greeter client received initial metadata: key=%s value=%s' %
               (key, value))
     event.set()
@@ -37,7 +46,7 @@ def check_status(response_future, wait_success):
         print("received initial metadata before time out!")
         for response in response_future:
             message = response.message
-        print("Greeter client received: " + message)
+            print("Greeter client received: " + message)
     else:
         print("Timed out before receiving any initial metadata!")
         response_future.cancel()
@@ -61,7 +70,7 @@ def main():
         thread_with_delay.start()
 
         # Wait on client side with timeout
-        timeout = 3
+        timeout = 7
         check_status(response_future_delay, event_for_delay.wait(timeout))
 
         # Expected to timeout.
--- a/examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
+++ b/examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
@@ -43,28 +43,26 @@ def check_status(response_future, wait_success):
 
 def main():
     # Create gRPC channel
-    channel = grpc.insecure_channel('localhost:50051')
-    stub = helloworld_pb2_grpc.GreeterStub(channel)
+    with grpc.insecure_channel('localhost:50051') as channel:
+        stub = helloworld_pb2_grpc.GreeterStub(channel)
 
-    event_for_delay = threading.Event()
+        event_for_delay = threading.Event()
 
-    # Server will delay send initial metadata back for this RPC
-    response_future_delay = stub.SayHelloStreamReply(
-        helloworld_pb2.HelloRequest(name='you'), wait_for_ready=True)
+        # Server will delay send initial metadata back for this RPC
+        response_future_delay = stub.SayHelloStreamReply(
+            helloworld_pb2.HelloRequest(name='you'), wait_for_ready=True)
 
-    # Fire RPC and wait for metadata
-    thread_with_delay = threading.Thread(target=wait_for_metadata,
-        args=(response_future_delay, event_for_delay))
-    thread_with_delay.start()
+        # Fire RPC and wait for metadata
+        thread_with_delay = threading.Thread(target=wait_for_metadata,
+            args=(response_future_delay, event_for_delay))
+        thread_with_delay.start()
 
-    # Wait on client side with timeout
-    timeout = 3
-    check_status(response_future_delay, event_for_delay.wait(timeout))
+        # Wait on client side with timeout
+        timeout = 3
+        check_status(response_future_delay, event_for_delay.wait(timeout))
 
-    # Expected to timeout.
-    thread_with_delay.join()
-
-    channel.close()
+        # Expected to timeout.
+        thread_with_delay.join()
 
 if __name__ == '__main__':
     logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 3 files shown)


examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
```diff
@@ -0,0 +1,73 @@
...
+
+helloworld_pb2, helloworld_pb2_grpc = grpc.protos_and_services(
+    "helloworld.proto")
+
+_LOGGER = logging.getLogger(__name__)
+_LOGGER.setLevel(logging.INFO)
+
+
+def wait_for_metadata(response_future, event):
+    for key, value in response_future.initial_metadata():
```
gnossen, line 29: I think it would make it more clear what's happening here if you pull the call to `initial_metadata()` onto another line  and save it off to a typed variable. Something like:

```
metadata: Sequence[Tuple[str, str]] = response_future.initial_metadata()
```

The way this is currently written, it looks like `initial_metadata` might actual be a streaming method.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
```diff
@@ -0,0 +1,73 @@
...
+                                             args=(response_future_delay,
+                                                   event_for_delay))
+        thread_with_delay.start()
+
+        # Wait on client side with timeout
+        timeout = 3
+        check_status(response_future_delay, event_for_delay.wait(timeout))
+
+        # Expected to timeout.
+        thread_with_delay.join()
```
gnossen, line 68: Does the process successfully exit if the server never returns initial metadata?

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
```diff
@@ -0,0 +1,73 @@
...
+        print('Greeter client received initial metadata: key=%s value=%s' %
+              (key, value))
+    event.set()
+
+
+def check_status(response_future, wait_success):
+    if wait_success:
+        print("received initial metadata before time out!")
+        for response in response_future:
+            message = response.message
```
gnossen, line 39: It looks like you're just overwriting `message` each time. I don't know if this makes much sense as a streaming API if we're ignoring all but the last message.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
```diff
@@ -0,0 +1,73 @@
...
+import threading
+
+import grpc
+
+helloworld_pb2, helloworld_pb2_grpc = grpc.protos_and_services(
+    "helloworld.proto")
+
+_LOGGER = logging.getLogger(__name__)
+_LOGGER.setLevel(logging.INFO)
+
```
gnossen, line 26: Please include a multi-line comment explaining what this example demonstrates.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_server.py
```diff
@@ -0,0 +1,54 @@
...
+
+_INITIAL_METADATA = ((b'initial-md', 'initial-md-value'),)
+
+
+class Greeter(helloworld_pb2_grpc.GreeterServicer):
+
+    def SayHelloStreamReply(self, request, servicer_context):
+        # Send initial metadata back to indicate server is ready and running.
+        print("sleeping 5s before sending metadata back")
+        sleep(5)
```
gnossen, line 33: Maybe pull this sleep out to a function called `do_work` to indicate that this is just pretending to do real work?

gnossen: This sleep should come _after_ we send initial metadata. The point of this example is to use the initial metadata as an indication that the request has been received and being processed.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_server.py
```diff
@@ -0,0 +1,54 @@
...
+
+        # Initial metadata will be send back immediately after calling send_initial_metadata.
+        print("sending inital metadata back")
+        servicer_context.send_initial_metadata(_INITIAL_METADATA)
+
+        # Server can do whatever it wants here before send actual response.
+        yield helloworld_pb2.HelloReply(message='Hello, %s!' % (request.name))
```
gnossen, line 40: If this example is going to be a streaming RPC, we should probably be returning multiple messages. Honestly, I think this example would be fine as a unary RPC.

gnossen (commented):
Nice work!

gnossen:
Optional stretch goal for this example:

Many of our examples have [subprocess-based tests](https://github.com/grpc/grpc/blob/master/examples/python/cancellation/test/_cancellation_example_test.py). This ensures that they don't bitrot over time like many examples do. This isn't entirely necessary, but it would be a good way to get exposed to a technique that we use very frequently (e.g. [1](https://github.com/grpc/grpc/blob/3980ed706a5ad43abbe5ac5e4a3c7843188ca630/src/python/grpcio_tests/tests/unit/_signal_handling_test.py), [2](https://github.com/grpc/grpc/blob/3980ed706a5ad43abbe5ac5e4a3c7843188ca630/src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client_test.py))

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
XuanWang-Amos: Sure, will add comments on both client and server side.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
XuanWang-Amos: Got it.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
XuanWang-Amos: Server will only return one reply anyway but you're right, we should log every message received for stream reply.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
XuanWang-Amos: It will timeout.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_server.py
XuanWang-Amos: We can add sleep at both places, the sleep before send initial_metadata means server is starting up, and client should cancel RPC if server didn't finish starting up before client timeout, I'll add some comments too.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_server.py
XuanWang-Amos: It's easier on client side to process stream response, I'll change here to return multiple messages.

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
gnossen: Then let's make this a daemon thread so that the whole process obeys the time limit.

XuanWang-Amos: Ack

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
```diff
@@ -11,10 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""The Python client of utilizing wait-for-ready flag with client time out."""
```
gnossen: Suggestion: "An example of setting a server connection timeout independent from the overall RPC timeout."

examples/python/wait_for_ready/wait_for_ready_with_client_timeout_example_client.py
```diff
@@ -61,7 +70,7 @@ def main():
         thread_with_delay.start()
 
         # Wait on client side with timeout
-        timeout = 3
```
gnossen: Nit: Might want to mention that the units here are "seconds"

gnossen (approved):
LGTM with a couple of nits.

gnossen (approved):

Issue #32171, opened 2023-01-21
gRPC plugin crashes (or fails) if the service comments contain a single `$` 


coryan:
### What version of gRPC and what language are you using?

gRPC 1.51.1, C++.

### What operating system (Linux, Windows,...) and version?

Linux, Alpine Stable

### What runtime / compiler are you using (e.g. python version or version of gcc)

GCC 11

### What did you do?

Run the gRPC plugin over this file:

https://github.com/googleapis/googleapis/blob/master/google/cloud/certificatemanager/v1/certificate_manager.proto

At the time the file contains this comment:

https://github.com/googleapis/googleapis/blob/9fac84a6942bd3e95fca1ffe87018adbf560c514/google/cloud/certificatemanager/v1/certificate_manager.proto#L57

### What did you expect to see?

A successful run.

### What did you see instead?

The plugin crashes with:

```
[ 97%] Running gRPC C++ protocol buffer compiler on /workspace/cmake-out/external/googleapis/src/googleapis_download/google/cloud/certificatemanager/v1/certificate_manager.proto
[libprotobuf FATAL google/protobuf/io/printer.cc:142]  Unclosed variable name.
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():   Unclosed variable name.
--grpc_out: protoc-gen-grpc: Plugin killed by signal 6.
```

The problem seems to be in the code generator, it passes the leading comments of the service directly to the Protobuf printer:

https://github.com/grpc/grpc/blob/8cf04e9a5405c3f6923276eb01ca0122f4223533/src/compiler/cpp_generator.cc#L1334

The current Protobuf release treats single `$` as a fatal error:

https://github.com/protocolbuffers/protobuf/blob/9b3231002dcfb713e619d841fc9bfa5156c211fc/src/google/protobuf/io/printer.cc#L142-L147

drfloob:
`gcf-owl-bot` sure is an active bot.

Thanks for reporting this, @coryan. The fix is merged.

PR #32190, opened 2023-01-24
[codegen] Escape variable delimiters in raw comment printing

drfloob:
Fixes https://github.com/grpc/grpc/issues/32171

### DIFF
--- a/src/compiler/cpp_generator.cc
+++ b/src/compiler/cpp_generator.cc
@@ -54,6 +54,16 @@ std::string FilenameIdentifier(const std::string& filename) {
   }
   return result;
 }
+
+std::string EscapeVariableDelimiters(std::string&& mut_str) {
+  size_t index = 0;
+  while ((index = mut_str.find('$', index)) != std::string::npos) {
+    mut_str.replace(index, 1, "$$");
+    index += 2;
+  }
+  return mut_str;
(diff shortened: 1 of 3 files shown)


src/compiler/cpp_generator.cc
```diff
@@ -54,6 +54,16 @@ std::string FilenameIdentifier(const std::string& filename) {
   }
   return result;
 }
+
+std::string EscapeVariableDelimiters(std::string&& mut_str) {
+  size_t index = 0;
```
markdroth, line 59: Consider instead using `absl::StrReplaceAll()` for this.

markdroth (approved):

src/compiler/cpp_generator.cc
drfloob: That was my first instinct, but the `grpc_plugin_support` target has very few dependencies, and I'm not entirely sure why, it seems almost by design. Maybe for code bloat reasons? Anyhow, that's the only reason I didn't use abseil here.

veblush: I feel like we should be careful on adding more dependency to `grpc_plugin_support`.

veblush (approved):
Thank you for having this quick fix.

PR #32196, opened 2023-01-25
Fix Python epoll1 Fork Support

gnossen:
This PR resolves several issues in core impacting fork support in Python. It also reenables and improves the existing Python-layer fork test.


src/python/grpcio_tests/tests/fork/_fork_interop_test.py
```diff
@@ -131,25 +167,60 @@ def testInProgressBidiNewChannelBlockingCall(self):
...
+        streams = tuple(tempfile.TemporaryFile() for _ in range(2))
+        sys.stderr.write("Invoking gdb\n")
+        sys.stderr.flush()
+        process = subprocess.Popen(cmd, stdout=streams[0], stderr=streams[1])
+        try:
+            process.wait(timeout=_GDB_TIMEOUT_S)
+        except subprocess.TimeoutExpired:
+            sys.stderr.write("gdb stacktrace generation timed out.\n")
+        finally:
+            for stream_name, stream in zip(("STDOUT", "STDERR"), streams):
```
XuanWang-Amos, line 200: [Minor] Can we modify `_dump_streams` a little bit so that can be used here too?

src/python/grpcio_tests/tests/fork/methods.py
```diff
@@ -118,33 +124,131 @@ def __init__(self, task, args=None):
...
+        streams = tuple(tempfile.TemporaryFile() for _ in range(2))
+        sys.stderr.write("Invoking gdb\n")
+        sys.stderr.flush()
+        process = subprocess.Popen(cmd, stdout=sys.stderr, stderr=sys.stderr)
+        time.sleep(5)
 
     def start(self):
-        self._process.start()
+        import sys
+        sys.stderr.write("AAAAAAAAAAAAAAAAAAAAAA forking\n")
```
XuanWang-Amos, line 171: Don't think we need this.

src/python/grpcio_tests/tests/fork/methods.py
```diff
@@ -118,33 +124,131 @@ def __init__(self, task, args=None):
...
+            sys.stderr.flush()
 
     def finish(self):
-        self._process.join(timeout=_CHILD_FINISH_TIMEOUT_S)
-        if self._process.is_alive():
-            raise RuntimeError('Child process did not terminate')
-        if self._process.exitcode != 0:
-            raise ValueError('Child process failed with exitcode %d' %
-                             self._process.exitcode)
+        import sys
```
XuanWang-Amos, line 230: Are this and below lines intended? Looks like debug strings to me.

tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
```diff
@@ -27,3 +27,6 @@ python_bazel_tests_single_threaded_unary_streams/bazel_wrapper ${RESULTSTORE_RES
 
 python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests_poller_engine
 python_bazel_tests_poller_engine/bazel_wrapper ${RESULTSTORE_RESULTS_FLAG} test --config=python_poller_engine ${BAZEL_FLAGS} ${TEST_TARGETS}
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path python_bazel_tests_fork_support
+python_bazel_tests_poller_engine/bazel_wrapper ${RESULTSTORE_RESULTS_FLAG} test --config=fork_support --runs_per_test=16 ${BAZEL_FLAGS} //src/python/grpcio_tests/tests/fork:fork_test
```
XuanWang-Amos, line 32: Just curious, is there any conventions for magic numbers like this?

src/python/grpcio_tests/tests/fork/_fork_interop_test.py
```diff
@@ -34,23 +47,33 @@
     from grpc._cython import cygrpc
```
XuanWang-Amos, line 47: Won't let me comment on previous lines but looks like we don't need `import sys` here?

src/python/grpcio_tests/tests/fork/_fork_interop_test.py
gnossen: Indeed we can. Nice catch.

src/python/grpcio_tests/tests/fork/methods.py
gnossen: Right you are. I missed one.

src/python/grpcio_tests/tests/fork/methods.py
gnossen: You're right. Probably not necessary anymore.

tools/internal_ci/linux/grpc_python_bazel_test_in_docker.sh
gnossen: I probably should have added a comment here, but this is a number that:

- Is low enough that it won't take forever
- Is high enough that we'll catch _some_ nondetermistic issues if they slip in
- Is low enough that it won't catch the 1/256 issue that I _know_ still exists.

src/python/grpcio_tests/tests/fork/_fork_interop_test.py
gnossen: Ah, yes. Another missed debug line. Nice catch!

src/core/lib/gprpp/fork.cc
```diff
@@ -228,5 +235,6 @@ void Fork::AwaitThreads() {
 
 std::atomic<bool> Fork::support_enabled_(false);
 bool Fork::override_enabled_ = false;
-Fork::child_postfork_func Fork::reset_child_polling_engine_ = nullptr;
+std::vector<Fork::child_postfork_func>* Fork::reset_child_polling_engine_ =
```
drfloob, line 238: I don't think this needs to be a pointer, the nullptr check doesn't seem to have much purpose other than to lazy initialize the vector.

src/core/lib/gprpp/fork.h
```diff
@@ -58,7 +59,7 @@ class Fork {
   // reset the polling engine's internal state.
   static void SetResetChildPollingEngineFunc(
       child_postfork_func reset_child_polling_engine);
-  static child_postfork_func GetResetChildPollingEngineFunc();
+  static std::vector<child_postfork_func> GetResetChildPollingEngineFunc();
```
drfloob, line 62: I think we can return this by const ref instead of by copy, `static const std::vector<child_postfork_func>& GetResetChildPollingEngineFunc()`

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
```diff
@@ -570,6 +575,12 @@ Epoll1Poller* MakeEpoll1Poller(Scheduler* scheduler) {
   return nullptr;
 }
 
+void Epoll1Poller::PrepareFork() { Kick(); }
+
+void Epoll1Poller::PostforkParent() {}
```
drfloob, line 580: Please leave `TODO(vigneshbabu): implement` on these methods. Vignesh or I will eventually tackle them.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.h
```diff
@@ -46,7 +47,8 @@ namespace experimental {
 class Epoll1EventHandle;
 
 // Definition of epoll1 based poller.
-class Epoll1Poller : public PosixEventPoller {
+class Epoll1Poller : public PosixEventPoller,
+                     public grpc_event_engine::experimental::Forkable {
```
drfloob, line 51: This can just be `public Forkable` since we're already in that namespace.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
```diff
@@ -388,8 +390,11 @@ Epoll1Poller::~Epoll1Poller() {
       delete handle;
     }
   }
+  closed_ = true;
```
drfloob, line 393: In the rare occurrence that a Kick and a fork event happen at roughly the same time, there may be concurrent read/write on `closed_`. I think it'll work out if we guard the variable by `mu_` as well.

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -316,6 +317,8 @@ void PosixEnginePollerManager::Run(absl::AnyInvocable<void()> cb) {
 }
 
 void PosixEnginePollerManager::TriggerShutdown() {
+  GPR_DEBUG_ASSERT(trigger_shutdown_called_ == false);
```
drfloob, line 320: Can be `GPR_DEBUG_ASSERT(!trigger_shutdown_called_);`

Or you can merge these two lines into `GPR_ASSERT(!std::exchange(trigger_shutdown_called_, true));`, but it's debatably less readable.

src/core/lib/iomgr/fork_posix.cc
```diff
@@ -100,10 +100,11 @@ void grpc_postfork_child() {
   if (!skipped_handler) {
     grpc_core::Fork::AllowExecCtx();
     grpc_core::ExecCtx exec_ctx;
-    grpc_core::Fork::child_postfork_func reset_polling_engine =
-        grpc_core::Fork::GetResetChildPollingEngineFunc();
-    if (reset_polling_engine != nullptr) {
-      reset_polling_engine();
+    for (auto reset_polling_engine :
```
drfloob, line 103: Does `auto*` work here?

drfloob (commented):
This is great, Richard! Have you confirmed this iteration order fix yet with the external reproductions?

@Vignesh2208 we'll still need to do some work to decouple the PosixEventEngine from the iomgr fork handlers. But we'll soon have some tests that'll help us do that safely.

src/core/lib/gprpp/fork.cc
gnossen: Is `std::vector<T*>` trivially destructible? I know this is `static`, not technically global, but I figured the same rules applied.

drfloob: I didn't realize this was a global, you're right. You could use the `NoDestruct` types src/core/lib/gprpp/no_destruct.h, but this works as well.

gnossen:
> Have you confirmed this iteration order fix yet with the external reproductions?

I have specifically confirmed that this fixes the issue for Ray on Mac OS.

src/core/lib/gprpp/fork.h
gnossen: Ah yeah. I meant to go back and do that. Fixed.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
gnossen: Done.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.h
gnossen: Ah yes. Didn't make the connection.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
gnossen: Oh, duh. Fixed.

src/core/lib/event_engine/posix_engine/posix_engine.cc
gnossen: I opted for the readability since this is a string we could be reading in a test failure at some point.

src/core/lib/iomgr/fork_posix.cc
gnossen: Indeed it does.

src/python/grpcio_tests/tests/fork/methods.py
```diff
@@ -118,27 +124,114 @@ def __init__(self, task, args=None):
...
+            traceback.print_exc()
+            self._exceptions.put('RpcError: %s' % rpc_error)
+        except Exception as e:  # pylint: disable=broad-except
+            traceback.print_exc()
+            self._exceptions.put(e)
+        sys.exit(0)
+
+    def _orchestrate_child_gdb(self):
+        cmd = [
+            "gdb",
```
gnossen, line 151: Note to self: Somehow ensure GDB is installed in the test execution environment.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
```diff
@@ -375,7 +375,10 @@ void Epoll1Poller::Shutdown() {
...
 
 void Epoll1Poller::Close() {
-  if (closed_) return;
+  {
+    grpc_core::MutexLock lock(&mu_);
+    if (closed_) return;
+  }
 
   if (g_epoll_set_.epfd >= 0) {
     close(g_epoll_set_.epfd);
```
drfloob, line 384: nit: I wonder if this needs to be done without holding the mutex. If not, it may be cleaner to take the lock once at the beginning of the function and hold it throughout.

drfloob (approved):
Looks good!

src/python/grpcio_tests/tests/fork/methods.py
gnossen: It turns out this won't work at the moment given that we need to change a kernel setting. We can leave it for local development though. For reference, you need to run `sudo bash -c 'echo "0" > /proc/sys/kernel/yama/ptrace_scope'` to get this working locally.

PR #32139, opened 2023-01-18
Update Abseil to 20230125

veblush:
Ready for review. What's done here;
- Updated abseil-cpp to [Abseil LTS 20230125](https://github.com/abseil/abseil-cpp/releases/tag/20230125.0)
- Fix IWYU issues

PR #32200, opened 2023-01-25
Podspec specifying third_party directories rather than using wildcard

veblush:
I found a problem while attempting to upgrade Abseil in https://github.com/grpc/grpc/pull/32139. The issue was that two versions of abseil can be shown to podspec; one is in the third_party/abseil-cpp and the other is from abseil podspec. gRPC Podspec shouldn't see the one in the third_party/abseil-cpp but it happened to see because of `"$(PODS_TARGET_SRCROOT)/third_party/**"`, resulting in a bunch of link errors.

Instead, I changed it to specify certain directories which we believe safe to use so that only one version of abseil can be accessible to gRPC podspec.

### DIFF
--- a/templates/gRPC-C++.podspec.template
+++ b/templates/gRPC-C++.podspec.template
@@ -136,7 +136,9 @@
       'USER_HEADER_SEARCH_PATHS' => '"$(PODS_TARGET_SRCROOT)"'${"\\"}
           ' "$(PODS_TARGET_SRCROOT)/src/core/ext/upb-generated"'${"\\"}
           ' "$(PODS_TARGET_SRCROOT)/src/core/ext/upbdefs-generated"'${"\\"}
-          ' "$(PODS_TARGET_SRCROOT)/third_party/**"',
+          ' "$(PODS_TARGET_SRCROOT)/third_party/re2"'${"\\"}
+          ' "$(PODS_TARGET_SRCROOT)/third_party/upb"'${"\\"}
+          ' "$(PODS_TARGET_SRCROOT)/third_party/xxhash"',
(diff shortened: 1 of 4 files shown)


sampajano (approved):
@veblush Thanks for the change! LGTM!

@HannahShiSFB Could you also review as you're recently doing podspec releases and could possibly get affected? Thanks!

templates/gRPC-C++.podspec.template
```diff
@@ -136,7 +136,9 @@
       'USER_HEADER_SEARCH_PATHS' => '"$(PODS_TARGET_SRCROOT)"'${"\\"}
           ' "$(PODS_TARGET_SRCROOT)/src/core/ext/upb-generated"'${"\\"}
           ' "$(PODS_TARGET_SRCROOT)/src/core/ext/upbdefs-generated"'${"\\"}
-          ' "$(PODS_TARGET_SRCROOT)/third_party/**"',
+          ' "$(PODS_TARGET_SRCROOT)/third_party/re2"'${"\\"}
+          ' "$(PODS_TARGET_SRCROOT)/third_party/upb"'${"\\"}
+          ' "$(PODS_TARGET_SRCROOT)/third_party/xxhash"',
```
HannahShiSFB, line 141: There are more than 20 directories under third_party, I wonder how did you decide only these three are needed?

veblush: We used to manage a list of dependencies to third_party and but it changed by https://github.com/grpc/grpc/pull/27609 so I used it to find these dependencies.

HannahShiSFB: I see, thanks for the change! LGTM!

PR #32204, opened 2023-01-25
[iomgr and EE logging] Log ERROR if socket() returns EMFILE on posix

apolcyn:
There are potentially surprising deployment bugs that can cause `EMFILE` to be hit. For example, file descriptor limits can be easily reached if
- the round robin LB policy is used
- the load balancer hands out an assignment with a lot of backends
- using debian's default 1024 file descriptor limit.

To make such problems more apparent, we can pay special attention to this error and log ERROR when it happens.

Related: b/265199104

### DIFF
--- a/src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
+++ b/src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
@@ -30,6 +30,7 @@
 
 #include "src/core/lib/gpr/useful.h"
 #include "src/core/lib/gprpp/crash.h"  // IWYU pragma: keep
+#include "src/core/lib/gprpp/time.h"
 #include "src/core/lib/iomgr/port.h"
 
 #ifdef GRPC_POSIX_SOCKET_UTILS_COMMON
@@ -96,8 +97,20 @@ absl::Status ErrorForFd(
 
 int CreateSocket(std::function<int(int, int, int)> socket_factory, int family,
                  int type, int protocol) {
-  return socket_factory != nullptr ? socket_factory(family, type, protocol)
-                                   : socket(family, type, protocol);
+  int res = socket_factory != nullptr ? socket_factory(family, type, protocol)
+                                      : socket(family, type, protocol);
+  if (res < 0 && errno == EMFILE) {
+    int saved_errno = errno;
+    GRPC_LOG_EVERY_N_SEC(
+        10, GPR_ERROR,
+        "socket(%d, %d, %d) returned %d with error: |%s|. This process "
+        "might not have a sufficient file descriptor limit for the number "
+        "of connections we want to open (which is a function of the LB policy, "
+        "number of channels, and number of backends to load balance across).",
+        family, type, protocol, res, grpc_core::StrError(errno).c_str());
+    errno = saved_errno;
+  }
+  return res;
 }
 
 absl::Status PrepareTcpClientSocket(PosixSocketWrapper sock,
--- a/src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
+++ b/src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
@@ -105,8 +105,9 @@ int CreateSocket(std::function<int(int, int, int)> socket_factory, int family,
         10, GPR_ERROR,
         "socket(%d, %d, %d) returned %d with error: |%s|. This process "
         "might not have a sufficient file descriptor limit for the number "
-        "of connections we want to open (which is a function of the LB policy, "
-        "number of channels, and number of backends to load balance across).",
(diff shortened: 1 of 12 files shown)


src/core/lib/gprpp/time.h
```diff
@@ -26,20 +26,27 @@
...
 
-#define GRPC_LOG_EVERY_N_SEC(n, format, ...)                    \
+#define GRPC_LOG_EVERY_N_SEC(n, severity, format, ...)          \
   do {                                                          \
     static std::atomic<uint64_t> prev{0};                       \
+    static std::atomic<bool> first_call{true};                  \
     uint64_t now = grpc_core::Timestamp::FromTimespecRoundDown( \
                        gpr_now(GPR_CLOCK_MONOTONIC))            \
                        .milliseconds_after_process_epoch();     \
+    bool should_log = false;                                    \
```
Vignesh2208, line 41: Instead of another atomic, could we do 

```
uint64_t prev_tsamp = prev.exchange(now)
if (prev_tsamp == 0 || now - prev_tsamp > (n)*1000) {
  gpr_log(severity, format, __VA_ARGS__);
}

```

Vignesh2208:
LGTM just one minor nit

src/core/lib/gprpp/time.h
apolcyn: There's a minor oddity with that approach which is that it could be called multiple times within the first millisecond of the process.

WDYT?

apolcyn: Thinking about this more, I followed your suggestion. The simpler code is probably worth this edge case being mis-handled (and it could only cause an increase of logs at the very start of a process).

markdroth (approved):
This looks good to me.  However, please also get a review from @ctiller and @drfloob before merging.  Thanks!

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
```diff
@@ -96,8 +97,20 @@ absl::Status ErrorForFd(
...
-                                   : socket(family, type, protocol);
+  int res = socket_factory != nullptr ? socket_factory(family, type, protocol)
+                                      : socket(family, type, protocol);
+  if (res < 0 && errno == EMFILE) {
+    int saved_errno = errno;
+    GRPC_LOG_EVERY_N_SEC(
+        10, GPR_ERROR,
+        "socket(%d, %d, %d) returned %d with error: |%s|. This process "
+        "might not have a sufficient file descriptor limit for the number "
+        "of connections we want to open (which is a function of the LB policy, "
```
drfloob, line 108: This is pretty minor, but I noticed "we" here, and wondered how often gRPC logs might contain "you", "I", or "us". There's a good mix of them all. We should consider using a standard voice for the logs, but probably not anthropomorphising regardless. Anyway, I'd suggest "... number of connection the process is trying to open ..." or some such.

drfloob (approved):
LGTM with a few minor comments.

test/core/event_engine/posix/log_too_many_open_files_test.cc
```diff
@@ -0,0 +1,52 @@
...
+  const auto mock_socket_factory = [](int, int, int) {
+    errno = EMFILE;
+    return -1;
+  };
+  auto addr = grpc_event_engine::experimental::URIToResolvedAddress(
+      "ipv4:127.0.0.1:80");
+  ASSERT_TRUE(addr.ok());
+  grpc_event_engine::experimental::PosixSocketWrapper::DSMode dsmode;
+  absl::StatusOr<grpc_event_engine::experimental::PosixSocketWrapper> result =
+      grpc_event_engine::experimental::PosixSocketWrapper::
```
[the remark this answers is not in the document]
apolcyn: done

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
apolcyn: Changed this - what do you think about 3rd person "gRPC wants to open", to scope it to the gRPC library

drfloob: My position is totally subjective, and easily bike-shedable, but fun to engage for a minute :-). Unless gRPC is sentient, I probably wouldn't attribute "wants" to it. Things like "gRPC is trying to ..." is still a bit of anthropomorphising - like asking if submarines can swim, depending on how people define "swim". But I'd be fine with it.

apolcyn: Perhaps it can be tweaked - as long as we're bike-shedding, I'll leave as is :)

ctiller (approved):

PR #32210, opened 2023-01-26
xDS interop: Refresh k8s auth on 401 to pick up updated auth token

sergiitk:
This PR adds retries on create/get requests from the test driver to the K8s API when 401 Unauthorized error is encountered.  
K8S python library expects the ApiClient to be cycled on auth token refreshes.   

The problem is described in https://github.com/kubernetes-client/python/issues/741. Currently we don't have any hypotheses why we weren't affected by this problem before.

To force the ApiClient to pick up the new credentials, I shut down the current client, create a new one, and replace api_client properties on all k8s APIs we manage.

This should also work with the Watch-based log collector recovering from an error. To support that, I replace default Configuration so that the next time Watch creates ApiClient implicitly, the Configuration with updated token will be used.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/framework/infrastructure/k8s.py
+++ b/tools/run_tests/xds_k8s_test_driver/framework/infrastructure/k8s.py
@@ -15,7 +15,6 @@
 #   added to get around circular dependencies caused by k8s.py clashing with
 #   k8s/__init__.py
 import datetime
-import functools
 import json
 import logging
 import pathlib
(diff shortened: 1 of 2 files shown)


sergiitk:
Running the cleanup job with this change: [grpc/core/master/linux/grpc_xds_resource_cleanup](https://source.cloud.google.com/results/invocations/bafbb9d2-0311-42e2-8447-4df9fe698022).

gnossen:
It looks like it worked, but there was so much stuff to clean up that it hit the Kokoro VM time limit.

sergiitk:
Moar cleanup runs!

- [grpc/core/master/linux/grpc_xds_resource_cleanup](https://source.cloud.google.com/results/invocations/0fcd306f-7db1-46ec-baf0-ac962e31b478)

gnossen (approved):

sergiitk:
Allright, looks like a full cleanup has been completed!

PR #32214, opened 2023-01-26
Upgrading to protobuf 22.x: react to changes in C++ code

jtattermusch:
- `::google::protobuf::util::Status` was removed, so trying to replace it with `::absl::Status` (hopefully that makes sense)
- some public protobuf headers were removed, adapting code of grpc codegen plugins to the new names.

TODO: As is, this change would need to get merged atomically with the actual protobuf upgrade.
Potentially we could condition the changes being made on protobuf version, so that grpc plugins can still compile against an older version of protobuf:
(e.g. sth like this).
```
#include <google/protobuf/port_def.inc>
#if PROTOBUF_VERSION ...
#define IS_OLD
#endif
#include <google/protobuf/port_undef.inc>
#ifdef IS_OLD
#include <google/protobuf/old/path>
#else
#include <google/protobuf/new/path>
#endif
```

Ideally, protobuf would add forwarding headers for the headers that were removed, (e.g. like here: https://github.com/protocolbuffers/protobuf/pull/11613/files, but more adjustments would be needed to the protobuf bazel build).

### DIFF
--- a/include/grpcpp/impl/codegen/config_protobuf.h
+++ b/include/grpcpp/impl/codegen/config_protobuf.h
@@ -68,8 +68,9 @@
 #ifndef GRPC_CUSTOM_JSONUTIL
 #include <google/protobuf/util/json_util.h>
 #include <google/protobuf/util/type_resolver_util.h>
+#include "absl/status/status.h"
 #define GRPC_CUSTOM_JSONUTIL ::google::protobuf::util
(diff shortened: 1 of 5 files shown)


gnossen (approved):

veblush (approved):

jtattermusch:
Protobuf upgrade was merged here: https://github.com/grpc/grpc/pull/32606 (and it includes some of these early experiments)

PR #32215, opened 2023-01-26
Add info about ca cert used to verify chain.

gtcooke94:
The tsi_peer object will now contain the subject of the root/ca cert that was used to verify the peer's chain during a handshake.

We have to get this information into `tsi_peer` in a roundabout way - unfortunately BoringSSL does not have the function `SSL_get0_verified_chain`. The only way to get the verified chain (and thus the CA/root cert that was used to verify the peer's chain) is through the `X509_STORE_CTX` used during the handshake post-verification. So we tie into that callback, then we store the CA Cert on the SSL object as `ex_data`. The SSL object is what we have access to when setting the values on `tsi_peer`, thus we need the information on that object.

There was issue with overlap with a CRL-related callback, so I had to move that function further up in the file.

There will be a follow-up PR that adds this information to `grpc_tls_custom_verification_check_request` and `TlsCustomVerificationCheckRequest` (these structs/classes are constructed using information in `tsi_peer`), with the end goal of the grpc-user having access to the Subject information of the CA cert used to verify the peer chain during custom verification callbacks.

The main changes are in the `ssl_transport_security.cc` file.
I added a test that uses a certificate chain with an intermediate CA, so there are several file additions to the `test_creds` directory. The README changes explain how these files are generated.


gtcooke94:
@matthewstevenson88

src/core/tsi/ssl_transport_security.cc
```diff
@@ -845,6 +858,22 @@ static tsi_result build_alpn_protocol_name_list(
...
 
+// This callback is invoked when the CRL has been verified and will soft-fail
+// errors in verification depending on certain error types.
+static int verify_cb(int ok, X509_STORE_CTX* ctx) {
+  int cert_error = X509_STORE_CTX_get_error(ctx);
+  if (cert_error == X509_V_ERR_UNABLE_TO_GET_CRL) {
+    gpr_log(
+        GPR_INFO,
+        "Certificate verification failed to get CRL files. Ignoring error.");
+    return 1;
```
matthewstevenson88, line 869: Should we be returning `ok` instead of `1` here?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -853,6 +882,38 @@ static int NullVerifyCallback(int /*preverify_ok*/, X509_STORE_CTX* /*ctx*/) {
...
+  if (preverify_ok == 0) {
+    return 0;
+  }
+
+  // If we're here, verification was successful
+  // Get the verified chain from the X509_STORE_CTX and put it on the SSL object
+  // so that we have access to it when populating the tsi_peer
+  STACK_OF(X509)* chain = X509_STORE_CTX_get0_chain(ctx);
+  if (chain == nullptr) {
+    return 1;
```
matthewstevenson88, line 905: Shouldn't all return values continue to be `preverify_ok`? We don't want to change the existing behavior, just opportunistically extract this info.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -853,6 +882,38 @@ static int NullVerifyCallback(int /*preverify_ok*/, X509_STORE_CTX* /*ctx*/) {
...
+  }
+
+  // The ca cert is the last in the chain
+  X509* ca_cert = sk_X509_value(chain, sk_X509_num(chain) - 1);
```
matthewstevenson88, line 909: Please return early if `ca_cert` is nullptr.

matthewstevenson88, line 909: Shouldn't we check that `sk_X509_num` is `>0` before this line?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -853,6 +882,38 @@ static int NullVerifyCallback(int /*preverify_ok*/, X509_STORE_CTX* /*ctx*/) {
...
+
+  SSL* ssl = static_cast<SSL*>(
+      X509_STORE_CTX_get_ex_data(ctx, SSL_get_ex_data_X509_STORE_CTX_idx()));
+  SSL_set_ex_data(ssl, g_ssl_ex_ca_cert_index, ca_cert);
```
matthewstevenson88, line 913: Please check that `ssl` is non-null before this line.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1146,6 +1211,17 @@ static tsi_result ssl_handshaker_result_extract_peer(
       &peer->properties[peer->property_count]);
   if (result != TSI_OK) return result;
   peer->property_count++;
+
+  if (ca_cert != nullptr) {
+    result = peer_property_from_x509_subject(
+        ca_cert, &peer->properties[peer->property_count], true);
+    if (result != TSI_OK) {
+      gpr_log(GPR_ERROR, "resut: %d", static_cast<int>(result));
```
matthewstevenson88, line 1219: nit: typo in "result". Also can we add a more informative error message here?

matthewstevenson88, line 1219: I'm not sure this should be an error log, since this will occur routinely e.g. if a self-signed cert is used. Could we make it a debug log instead?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1146,6 +1211,17 @@ static tsi_result ssl_handshaker_result_extract_peer(
...
+      return result;
```
matthewstevenson88, line 1220: We shouldn't fail if there is a problem here, I think we should just log and continue.

test/core/tsi/ssl_transport_security_test.cc
```diff
@@ -191,6 +191,18 @@ static void ssl_test_setup_handshakers(tsi_test_fixture* fixture) {
       TSI_OK);
 }
 
+static void check_ca_cert_subject(ssl_tsi_test_fixture* /*ssl_fixture*/,
+                                  const tsi_peer* peer) {
+  const tsi_peer_property* ca_cert_subject =
+      tsi_peer_get_property_by_name(peer, TSI_X509_CA_SUBJECT_PEER_PROPERTY);
+  ASSERT_NE(ca_cert_subject, nullptr);
+  const char* expected_match =
```
matthewstevenson88, line 199: nit: Please use `std::string` here instead of `const char*.

test/core/tsi/ssl_transport_security_test.cc
```diff
@@ -360,6 +372,9 @@ static void ssl_test_check_handshaker_peers(tsi_test_fixture* fixture) {
     check_session_reusage(ssl_fixture, &peer);
     check_alpn(ssl_fixture, &peer);
     check_security_level(&peer);
+    if (!ssl_fixture->session_reused) {
```
matthewstevenson88, line 375: Can we check that this peer property is *not set* for resumed sessions?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -144,6 +144,7 @@ struct tsi_ssl_frame_protector {
 static gpr_once g_init_openssl_once = GPR_ONCE_INIT;
 static int g_ssl_ctx_ex_factory_index = -1;
 static const unsigned char kSslSessionIdContext[] = {'g', 'r', 'p', 'c'};
+static int g_ssl_ex_ca_cert_index = -1;
```
matthewstevenson88, line 147: Here and throughout: we seem to always refer to "ca cert", but really we're interested in the root CA cert that was used to verify the chain. Can we rename instead say (e.g.) "verified root cert"? For example, this variable would become `g_ssl_ex_verified_root_cert_index`?

test/core/tsi/ssl_transport_security_test.cc
```diff
@@ -191,6 +191,18 @@ static void ssl_test_setup_handshakers(tsi_test_fixture* fixture) {
       TSI_OK);
 }
 
+static void check_ca_cert_subject(ssl_tsi_test_fixture* /*ssl_fixture*/,
```
matthewstevenson88, line 194: IIUC, we are only testing in the case when verification is successful and there are no intermediate CAs in the chain. I think it would be good to test in the following cases as well:

1. When verification fails (e.g. we can use the self-signed cert in the /src/core/tsi/test_creds directory).
2. When verification succeeds but there is >=1 intermediate CA in the chain. (If there are no examples of this in /src/core/tsi/test_creds let me know and I can help you generate them.)

matthewstevenson88 (commented):
Thanks for the PR! Left a handful of comments, PTAL and feel free to reach out if you'd like to talk about any of them. :)

src/core/tsi/ssl_transport_security.cc
gtcooke94: For completeness, I didn't actually add this, just moved it further up in the file.
But to answer the question, no, these verification callbacks just return 1 on success and 0 on failure

src/core/tsi/ssl_transport_security.cc
gtcooke94: Above this is a line for `if preverify_ok == 0 return 0;`, so behaviorally here `preverify_ok` will always be 1, but it could be cleaner to just re-use `preverify_ok`.
One of the intents here was that, if `preverify_ok` is _not_ ok (`== 0`) then there won't be a verified chain so we should not continue.

Thoughts?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Good idea, I'll add the additional checks for this and for `nullptr`. I think if `chain` were empty but not `nullptr` this would have an issue

src/core/tsi/ssl_transport_security.cc
gtcooke94: Ack, will change

src/core/tsi/ssl_transport_security.cc
gtcooke94: Ack, will change

src/core/tsi/ssl_transport_security.cc
gtcooke94: Ack, will add check

test/core/tsi/ssl_transport_security_test.cc
gtcooke94: `const char* expected match` is used through this test file, [example](https://github.com/grpc/grpc/blob/a4e3a22f5179fabc1670dcfa6d58e4adfaf7b02a/test/core/tsi/ssl_transport_security_test.cc#L322)

test/core/tsi/ssl_transport_security_test.cc
gtcooke94: So you mean in an `else` block after this, explicitly check that there is no peer property with `TSI_X509_CA_SUBJECT_PEER_PROPERTY`?

src/core/tsi/ssl_transport_security.cc
gtcooke94: We had placed this log here for debugging - do you think we actually need it? Above a lot of the properties follow the pattern `if result != TSI_OK) return result;`. I think the `peer_chain` is the only property above that doesn't return result.

I guess if we reach here, what we're saying is "there is a root cert that is set in the ex_data, but adding it as a property to `tsi_peer` is failing." So what's our desired behavior here?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -2139,7 +2214,7 @@ tsi_result tsi_create_ssl_server_handshaker_factory_with_options(
         case TSI_REQUEST_AND_REQUIRE_CLIENT_CERTIFICATE_AND_VERIFY:
           SSL_CTX_set_verify(impl->ssl_contexts[i],
                              SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT,
-                             nullptr);
+                             RootCertExtractCallback);
```
rockspore, line 2217: I cann't comment right there but should l2207 also have this callback?

gtcooke94: Defer to @matthewstevenson88 on that, it's the case of `TSI_REQUEST_CLIENT_CERTIFICATE_AND_VERIFY`

test/core/tsi/ssl_transport_security_test.cc
gtcooke94: Added nullptr checks in the test for cases where this should be unset

Is case 2 a special case? The implementation just gets the last cert in the verified chain, should it matter if there are intermediate CAs?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Resolving because I think we will address this in this thread: https://github.com/grpc/grpc/pull/32215#discussion_r1089523074

src/core/tsi/ssl_transport_security.cc
```diff
@@ -472,7 +485,7 @@ static tsi_result peer_from_x509(X509* cert, int include_certificate_type,
     }
 
     result = peer_property_from_x509_subject(
-        cert, &peer->properties[current_insert_index++]);
+        cert, &peer->properties[current_insert_index++], false);
```
matthewstevenson88, line 488: nit: Please add an inline comment here /*is_verified_root_cert=*/ before the "false"

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Ah I see. Why did it need to be moved?

If there's a good reason to move it, can we leave it where it is in this PR and move it in a follow-up PR?

Rationale: This CL has some subtleties to it, and we want to avoid making unrelated changes that will obscure the real functional changes.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -853,6 +882,52 @@ static int NullVerifyCallback(int /*preverify_ok*/, X509_STORE_CTX* /*ctx*/) {
   return 1;
 }
 
+static int RootCertExtractCallback(int preverify_ok, X509_STORE_CTX* ctx) {
+  // There's a case where this function is set in SSL_CTX_set_verify and a CRL
+  // related callback is set with X509_STORE_set_verify_cb. They overlap and
+  // this will take precedence, thus we need to ensure the CRL related callback
```
matthewstevenson88, line 888: Why does this take precendence?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -853,6 +882,52 @@ static int NullVerifyCallback(int /*preverify_ok*/, X509_STORE_CTX* /*ctx*/) {
...
+  // is still called
+  X509_VERIFY_PARAM* param = X509_STORE_CTX_get0_param(ctx);
+  auto flags = X509_VERIFY_PARAM_get_flags(param);
+  if (flags & X509_V_FLAG_CRL_CHECK) {
+    preverify_ok = verify_cb(preverify_ok, ctx);
+  }
+
+  if (preverify_ok == 0) {
```
matthewstevenson88, line 896: Please add a comment explaining the rationale behind this if statement.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -853,6 +882,52 @@ static int NullVerifyCallback(int /*preverify_ok*/, X509_STORE_CTX* /*ctx*/) {
   return 1;
 }
 
+static int RootCertExtractCallback(int preverify_ok, X509_STORE_CTX* ctx) {
```
matthewstevenson88, line 885: Please check whether `ctx` is null when we enter the function. If we call `X509_STORE_CTX_get0_chain` and `ctx` is null, then we will segfault.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -853,6 +882,52 @@ static int NullVerifyCallback(int /*preverify_ok*/, X509_STORE_CTX* /*ctx*/) {
...
+  // If we're here, verification was successful
+  // Get the verified chain from the X509_STORE_CTX and put it on the SSL object
+  // so that we have access to it when populating the tsi_peer
+  STACK_OF(X509)* chain = X509_STORE_CTX_get0_chain(ctx);
+  if (chain == nullptr) {
+    return preverify_ok;
+  }
+
+  // The root cert is the last in the chain
+  auto chain_length = sk_X509_num(chain);
```
matthewstevenson88, line 909: nit: Please avoid the use of `auto`.

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Yup good call, we should have it in both cases.

test/core/tsi/ssl_transport_security_test.cc
matthewstevenson88: Maybe I misunderstood, but I didn't see any tests where verification fails?

I think 2 is important because the intermediates are sent by the peer but the root is not, it's from our trust store.

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Sounds reasonable.

test/core/tsi/ssl_transport_security_test.cc
gtcooke94: There are failed verifications in the [CRL related SSL tests](https://github.com/grpc/grpc/blob/master/test/core/tsi/crl_ssl_transport_security_test.cc)

I'll happily add more tests though, I'll discuss with you offline about setting up the chain for 2.

src/core/tsi/ssl_transport_security.cc
gtcooke94: Adding

src/core/tsi/ssl_transport_security.cc
gtcooke94: It's because of this [code in boringssl](https://boringssl.googlesource.com/boringssl/+/refs/heads/master/ssl/ssl_x509.cc#402).

In the chain of BoringSSL, if you set a verification callback with `SSL_CTX_set_verify` it ends up as the callback here on `hs->config->verify_callback`. So if that is set, it will override a callback set on `X509_STORE_CTX_set_verify` with whatever is there.

And this is the case here - we've maybe set a CRL-related verification callback with `X509_STORE_CTX_set_verify`, but we've also set the callback to extract the root cert with `SSL_CTX_set_verify`

src/core/tsi/ssl_transport_security.cc
gtcooke94: Will do

src/core/tsi/ssl_transport_security.cc
gtcooke94: Will do

src/core/tsi/ssl_transport_security.cc
gtcooke94: Will do

src/core/tsi/ssl_transport_security.cc
gtcooke94: Changed to `size_t`

src/core/tsi/test_creds/README
```diff
@@ -15,7 +15,7 @@ common name which is set to badserver.test.google.com.
 
 Valid test credentials:
 =======================
-
+er
```
matthewstevenson88, line 18: nit: Should this line be removed?

src/core/tsi/test_creds/README
```diff
@@ -99,6 +99,25 @@ multi-domain-openssl.cnf
 $ openssl req -x509 -new -extensions v3_req -key multi-domain.key -out
 multi-domain.pem -days 3650 -config multi-domain-openssl.cnf
 
+
+Intermediate is an intermediate CA used to sign leaf certs used to test having
```
matthewstevenson88, line 103: nit: This comment is a bit circular, could you clarify? Also please mention the other files (e.g. leaf_and_intermediate_chain.pem) that are being created.

src/core/tsi/test_creds/intermediate.cnf
```diff
@@ -0,0 +1,12 @@
...
+req_extensions = v3_req
+prompt = no
+
+[req_distinguished_name]
+CN = intermediatecert.example.com
+
+[v3_req]
+keyUsage = critical, digitalSignature, keyEncipherment, keyCertSign
+extendedKeyUsage = clientAuth, serverAuth
+basicConstraints = critical, CA:true
```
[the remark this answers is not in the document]
matthewstevenson88: Friendly ping :)

src/core/tsi/test_creds/leaf_signed_by_intermediate.cnf
```diff
@@ -0,0 +1,12 @@
...
+req_extensions = v3_req
+prompt = no
+
+[req_distinguished_name]
+CN = *.test.google.com.au
+
+[v3_req]
+keyUsage = critical, digitalSignature, keyEncipherment
+extendedKeyUsage = clientAuth, serverAuth
+basicConstraints = critical, CA:false
```
matthewstevenson88, line 12: Same here RE blank line to make the linter happy. :)

matthewstevenson88 (commented):
A few miscellaneous comments, and then this CL LGTM. If it's OK with you, I might ask Zhan to take a look as well, because this PR got quite technical.

We should find a reviewer from the gRPC team to take a look as well.

src/core/tsi/test_creds/leaf_signed_by_intermediate.cnf
gtcooke94: Done

src/core/tsi/test_creds/README
gtcooke94: Done

src/core/tsi/test_creds/README
gtcooke94: Done

src/core/tsi/test_creds/leaf_signed_by_intermediate.cnf
```diff
@@ -0,0 +1,17 @@
...
+subjectAltName = @alt_names
+
+[alt_names]
+URI = spiffe://trust-domain/path
+DNS = leafcert.example.com
```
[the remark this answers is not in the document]
gtcooke94: Done

src/core/tsi/test_creds/intermediate.cnf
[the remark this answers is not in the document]
gtcooke94: Done

gtcooke94:
Feel free to add Zhan as another pair of eyes, and I believe I've addressed your last round of comments.

There's a few tests failing, but I believe it's some infra/CI related issues (I made no material code changes since the last commit since everything passed). I'll keep an eye on it and re-run the tests.

matthewstevenson88:
@zeromath

zeromath:
LGTM on OpenSSL APIs.

gtcooke94:
@markdroth we've got LGTM from the security side, can we get a general language review?

markdroth (approved):

PR #32217, opened 2023-01-26
[backport][v1.48.x] Fix python tests on kokoro macos monterey (#31126)

drfloob:
* fix python build on kokoro macos monterey

* avoid using realpath

* fix grpc_distribtests_python on macos monterey

* try upgrade setuptools

* fix unused variable warning/error on macos


jtattermusch:
Looks like tests are red, I suspect you'd need to adjust this change to match the python versions that are being tested/built on the 1.48.x branch (recently we dropped support for some versions such as 3.6 and added some new such as 3.11 and the backport needs to reflect that).

PR #32226, opened 2023-01-27
Generate record instead of class

KentGrigo:
The motivation for this PR is that I'd like to make an adapter that's a record instead of a class.
But since gRPC generates a class instead of a record, it forces my hand to use a class as well, since a record can't inherit from a class.

I want to use a record in my adapter because it turns this:
```
public class XyzAdapter : XyzService.XyzServiceBase
{
	private readonly IAbc _abc;
        // x 10

	public XyzAdapter(
		IAbc abc,
                // x 10
	)
	{
		_abc = abc;
                // x 10
	}
```

into this

```
public record XyzAdapter(
	IAbc Abc,
        // x 10
) : XyzService.XyzServiceBase {
```

### DIFF
--- a/src/compiler/csharp_generator.cc
+++ b/src/compiler/csharp_generator.cc
@@ -436,14 +436,14 @@ void GenerateServiceDescriptorProperty(Printer* out,
 
 void GenerateServerClass(Printer* out, const ServiceDescriptor* service) {
   out->Print(
-      "/// <summary>Base class for server-side implementations of "
+      "/// <summary>Base record for server-side implementations of "
       "$servicename$</summary>\n",
       "servicename", GetServiceClassName(service));
   out->Print(
       "[grpc::BindServiceMethod(typeof($classname$), "
       "\"BindService\")]\n",
       "classname", GetServiceClassName(service));
-  out->Print("public abstract partial class $name$\n", "name",
+  out->Print("public abstract partial record $name$\n", "name",
              GetServerClassName(service));
   out->Print("{\n");
   out->Indent();


KentGrigo:
It seems that I'm not able to add labels. 
I think the appropriate labels are: `lang/c#` and `release notes: no`.

Can someone please add them?

KentGrigo:
@yashykt or @jtattermusch, will one of you please add the `release notes` label?
I'm not sure whether it should `yes` or `no`; but based on previous PRs, I think it should be `no`.

KentGrigo:
@yashykt and @jtattermusch, will you please look into this PR before it's automatically closed?

jtattermusch:
CC @tonydnewell

src/compiler/csharp_generator.cc
```diff
@@ -436,14 +436,14 @@ void GenerateServiceDescriptorProperty(Printer* out,
...
-      "/// <summary>Base class for server-side implementations of "
+      "/// <summary>Base record for server-side implementations of "
       "$servicename$</summary>\n",
       "servicename", GetServiceClassName(service));
   out->Print(
       "[grpc::BindServiceMethod(typeof($classname$), "
       "\"BindService\")]\n",
       "classname", GetServiceClassName(service));
-  out->Print("public abstract partial class $name$\n", "name",
+  out->Print("public abstract partial record $name$\n", "name",
```
jtattermusch, line 446: TBH, I don't understand why the base class for an implementation of a service would be a record rather than a class. I could see some benefits of the protobuf messages being records, but for service base class using a `record` doesn't make much sense IMHO?

PR #32228, opened 2023-01-27
[Draft][EventEngine] Pass Posix interface support information with ChannelArgs

drfloob:
Proposed alternative to #32194

Based on https://github.com/grpc/grpc/pull/32201, which may land first. 

Requires a cherrypick.


ctiller:
Do we need the fd stuff to be carried by channel args?

drfloob:
> Do we need the fd stuff to be carried by channel args?

As discussed offline, this is necessary if we continue with the current approach for fd support.

PR #32229, opened 2023-01-28
[EventEngine] Skip legacy fork handling for ExecCtx when in an EventEngine thread

drfloob:


gnossen (approved):

src/core/lib/gprpp/fork.cc
```diff
@@ -63,6 +64,13 @@ class ExecCtxState {
   }
 
   void IncExecCtxCount() {
+    // If we are in an EventEngine thread, we should not block any ExecCtx
```
ctiller, line 67: I don't think this is sound: event engine threads can ultimately end up executing anything that schedules on an `ExecCtx`, so I don't buy the claim that we don't need to worry about them.

src/core/lib/event_engine/posix_engine/posix_engine.cc
```diff
@@ -481,7 +481,9 @@ std::unique_ptr<EventEngine::DNSResolver> PosixEventEngine::GetDNSResolver(
   grpc_core::Crash("unimplemented");
 }
 
-bool PosixEventEngine::IsWorkerThread() { grpc_core::Crash("unimplemented"); }
+bool PosixEventEngine::IsWorkerThread() {
```
ctiller, line 484: My read here is that this returns true if the thread is a worker thread is from *any* event engine - if that's the case we should rephrase the API to make this not a member function - a member function implies "this is a worker thread from *this* event engine instance"

src/core/lib/gprpp/fork.cc
gnossen: I might be missing something here -- not an expert on EventEngine -- but I think the issue is just that the wording of the comment here is just slightly off. It's not that we're not worrying about them  it's that the ThreadPool implementation is expected to deal with the pre-fork behavior rather than this method.

drfloob: Richard is right. The wording is loose, the behavior is not.

src/core/lib/event_engine/posix_engine/posix_engine.cc
drfloob: That's a good point. Given that we have a solution that does not require `IsWorkerThread`, I will back out this implementation and leave decisions about the public API for another day. I imagine the "any EventEngine thread" version will be more useful, but we'll get a better idea of that in the next few months.

src/core/lib/gprpp/fork.cc
ctiller: Can we update this comment with more precise language then? I assume it's "EventEngine is expected to terminate all threads before fork, and so this extra work is unnecessary" or something along those lines?

src/core/lib/gprpp/fork.cc
```diff
@@ -63,6 +64,12 @@ class ExecCtxState {
   }
 
   void IncExecCtxCount() {
+    // EventEngine is expected to terminate all threads before fork, and so this
+    // extra work is unnecessary
+    if (grpc_event_engine::experimental::ThreadLocal::IsEventEngineThread()) {
+      gpr_atm_no_barrier_fetch_add(&count_, 1);
```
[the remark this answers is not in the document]
drfloob: No, @gnossen has tests with stack traces that show EventEngine threads getting stuck here. He also patched this PR into some work he's doing to debug the fork hang issue, and found this resolved the hang.

src/core/lib/gprpp/fork.cc
drfloob: Done

src/core/lib/gprpp/fork.cc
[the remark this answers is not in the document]
ctiller: Ok so this is papering over the problem and we expect some rarer problem to come up (and the comment needs to be fixed still - EventEngine evidently does not terminate all threads before the fork, but we believe ignoring that is better than the alternative)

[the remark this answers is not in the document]
drfloob: Not quite. EventEngines are expected to terminate all of their threads pre-fork (or otherwise confirm they are fork-safe). We are dealing with an intersection of fork handlers, and non-determinism around where and when callbacks are run. This code guarantees that the _blocking behavior_ of ExecCtx on fork only applies to the older fork handling code. If an EventEngine is running this bit of code, all current known implementations are going to wait on this code to finish before allowing the fork to proceed.

I don't see anything being papered over here. This older fork code is going away, ExecCtx is (hopefully, eventually) going away. This change allows fork handlers to coexist in the meantime.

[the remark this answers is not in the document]
ctiller: Is it the expectation then that the thread shutdown happens *after* the exec ctx completion?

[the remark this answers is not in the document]
drfloob: Correct. EventEngines are not aware of ExecCtxs, parking engine threads here otherwise would require we expose something to aid in engine pre-fork handling.

ctiller (approved):
Would recommend noting the ordering in some comment

PR #32239, opened 2023-01-30
[fix] Change expected checksums of Github archives

drfloob:
See https://github.com/google/googletest/issues/4135 for explanation and further reading.

This currently affects googletest and libprotobuf-mutator in our tests, but may eventually affect others.

I manually confirmed that both packages have identical contents by:

1) downloading both the new archives and the cached versions in google storage
2) confirming the different checksums matched the errors we're seeing in our CI (example: https://source.cloud.google.com/results/invocations/fcadc8ed-8b7c-4110-b5e2-073250b56e4a/log)
3) extracting both archive pairs, and running `find . -type f | xargs sha256sum > /tmp/checksums`, then in its twin, running `sha256sum --check /tmp/checksums`. No differences were found.

### DIFF
--- a/bazel/grpc_deps.bzl
+++ b/bazel/grpc_deps.bzl
@@ -251,10 +251,11 @@ def grpc_deps():
     if "com_google_googletest" not in native.existing_rules():
         http_archive(
             name = "com_google_googletest",
-            sha256 = "c8de6c60e12ad014a28225c5247ee735861d85cf906df617f6a29954ca05f547",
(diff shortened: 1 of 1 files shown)


drfloob:
Github is rolling back the change that caused the Checksum mismatch. This PR is unnecessary for now, but we (and many, many other bazel users) will likely want to revisit how we verify our dependencies.

See https://github.com/orgs/community/discussions/45830 and https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1409438954

PR #32240, opened 2023-01-30
[EventEngine] RunAfter migration: grpc_chttp2_transport

yijiem:


src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -2613,47 +2637,36 @@ void grpc_chttp2_config_default_keepalive_args(grpc_channel_args* args,
...
-  grpc_chttp2_transport* t = static_cast<grpc_chttp2_transport*>(arg);
+static void init_keepalive_ping(grpc_chttp2_transport* t) {
   t->combiner->Run(GRPC_CLOSURE_INIT(&t->init_keepalive_ping_locked,
                                      init_keepalive_ping_locked, t, nullptr),
-                   error);
+                   absl::OkStatus());
 }
 
 static void init_keepalive_ping_locked(void* arg, grpc_error_handle error) {
+  GPR_ASSERT(error.ok());
```
drfloob, line 2647: Maybe `GPR_DEBUG_ASSERT`? It seems impossible for it to be any other value.

drfloob (commented):
Looks pretty good. All of the extra `UNREF_TRANSPORT` calls are unfortunate, adds a bit of complexity to the ref counting if we need to debug problems, but I don't see a great way to avoid those at the moment.

src/core/ext/transport/chttp2/transport/writing.cc
```diff
@@ -152,14 +153,14 @@ static void maybe_initiate_ping(grpc_chttp2_transport* t) {
...
-    if (!t->ping_state.is_delayed_ping_timer_set) {
-      t->ping_state.is_delayed_ping_timer_set = true;
+    if (!t->ping_state.delayed_ping_timer_handle) {
       GRPC_CHTTP2_REF_TRANSPORT(t, "retry_initiate_ping_locked");
-      GRPC_CLOSURE_INIT(&t->retry_initiate_ping_locked,
-                        grpc_chttp2_retry_initiate_ping, t,
-                        grpc_schedule_on_exec_ctx);
-      grpc_timer_init(&t->ping_state.delayed_ping_timer, next_allowed_ping,
-                      &t->retry_initiate_ping_locked);
+      t->ping_state.delayed_ping_timer_handle =
```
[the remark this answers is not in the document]
yijiem: Hmmm, I didn't see the race... `maybe_initiate_ping` only runs in the transport's combiner lock, same for `close_transport_locked` I would think... So I don't think they will run concurrently or in parallel (correct me if I'm wrong though).

yijiem (commented):
Thanks AJ!

src/core/ext/transport/chttp2/transport/writing.cc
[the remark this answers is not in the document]
drfloob: You're right. I didn't realize `maybe_initiate_ping` was only ever called with the combiner lock held.

drfloob (approved):

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -451,14 +459,21 @@ static void init_keepalive_pings_if_enabled(grpc_chttp2_transport* t) {
...
+    // causing issue with overriding the handle. Thus use a mutex here.
+    // Hopefully this only happens once during init.
+    std::shared_ptr<absl::Mutex> mu = std::make_shared<absl::Mutex>();
+    absl::MutexLock lock(&*mu);
+    t->keepalive_ping_timer_handle =
+        t->event_engine->RunAfter(t->keepalive_time, [t, mu] {
+          grpc_core::ApplicationCallbackExecCtx callback_exec_ctx;
+          grpc_core::ExecCtx exec_ctx;
+          absl::MutexLock lock(&*mu);
+          init_keepalive_ping(t);
```
ctiller, line 472: Why not just re-enter the combiner here, eschewing the mutex?

yijiem: yep, that's the other approach I was considering... `init_keepalive_ping` does re-enter the combiner, we need to run `init_keepalive_pings_if_enabled` in the combiner, so that `init_keepalive_ping_locked` is guaranteed to run after `init_keepalive_pings_if_enabled` had returned.

I decide to make the change since your prompt has changed my mind. :) PTAL. And also I don't like introducing mutex to chttp2_transport and adding performance overhead potentially.

This approach might slightly delay the run of the keepalive_ping_timer though.

ctiller: I think that delay is probably ok, but I like that we keep a consistent locking story in this code (we want to change that story, and having uniformity ahead of time is a huge boon)

yijiem: Gotcha, thanks Craig!

PR #32241, opened 2023-01-30
Fix cmake on macos

veblush:
AppleClang's default C++ version is C++98 so it always needs `CMAKE_CXX_STANDARD` option but it might be too much for existing use-cases that rely on the old gRPC cmake behavior setting `CMAKE_CXX_STANDARD` by itself. So I revived this behavior only for MacOS with a warning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,6 +229,12 @@ if(WIN32)
   set(_gRPC_PLATFORM_WINDOWS ON)
 endif()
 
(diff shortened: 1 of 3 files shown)


XuanWang-Amos (approved):

veblush:
`Portability Tests Linux [Build Only] (internal CI)` is now flaky due to `not enough space` so I'm going to force this to get merged.

PR #32246, opened 2023-01-31
Changing parallelism of linux/grpc_portability_build_only

veblush:
Currently `linux/grpc_portability_build_only` became flaky because of out of space (300GB). This is because one docker run can use up to 80GB and 4 docker jobs can work in parallel. Sometimes 4 jobs end up consuming all available 300GB resulting in out of space error. So I reduced this parallelism to 1 docker run at a time but increased the inner parallelism so that total elapsed time won't suffer much from this change.

### DIFF
--- a/tools/internal_ci/linux/grpc_portability_build_only.cfg
+++ b/tools/internal_ci/linux/grpc_portability_build_only.cfg
@@ -26,5 +26,5 @@ action {
 
 env_vars {
   key: "RUN_TESTS_FLAGS"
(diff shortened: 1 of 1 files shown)


eugeneo (approved):

jtattermusch:
I understand the need for the change but making changes like this without checking the impact on the total duration of the job almost always leads to job slowdowns:

image

The avg duration of the job was just over 30minutes and now it's ~45mins, so not a trivial regression.

PR #32025, opened 2023-01-05
add back grpc_artifact_android_ndk image to unbreak C# artifact build

jtattermusch:
Fixup https://github.com/grpc/grpc/pull/31961/files in v1.46.x branch.
The backport PR has by mistake removed the docker image that's needed for building Grpc C# android artifacts.

Example failed build:
https://source.cloud.google.com/results/invocations/1a51e8f9-1595-4fc6-a118-cdff4dfdd2d6/targets/github%2Fgrpc/tests

PR #32247, opened 2023-01-31
add back grpc_artifact_android_ndk image to unbreak C# artifact build (#32025)

drfloob:
See https://github.com/grpc/grpc/pull/32025 for similar work done on the v1.46.x branch


drfloob:
Looks like this fixed the `Distribution Tests C# Linux` test, which was the hope.

PR #32249, opened 2023-01-31
xDS: fix bug when receiving an update for an aggregate cluster child

markdroth:


apolcyn (approved):

apolcyn:
Thanks, interesting issue. So I understand, we were clearing this `lb_config` state and breaking subsequent CDS updates?

veblush:
I used my admin power to get this merged as portability build issue is a known issue which is now fixed by https://github.com/grpc/grpc/pull/32246

markdroth:
@apolcyn This affects only the LB config, which always comes from the aggregate cluster itself, not the underlying clusters.  So this would happen only if the aggregate cluster did *not* change but one of the underlying clusters *did* change.  The test exercises this case.

PR #32252, opened 2023-02-01
remove go version v1.52.0 and keep only v1.52.3

arvindbr8:
As a reactive from #32205, we need to remove the 52.0 release and only keep v1.52.3 which is the latest patch version for that minor release


arvindbr8:
@jtattermusch @drfloob, PTAL when you can..

arvindbr8:
Friendly ping, :) @markdroth @jtattermusch

markdroth (approved):

PR #32255, opened 2023-02-01
[WIP] Update boringssl

veblush:


sampajano:
@HannahShiSFB You had some experiences with upgrading BoringSSL in https://github.com/grpc/grpc/pull/31869. Could you help comment on the iOS failures here (in what's the causes and solutions)? Thanks!

HannahShiSFB:
It's the same issue as in #31938, which I have some initial progress commented there, will follow up in that PR.

veblush:
> It's the same issue as in #31938, which I have some initial progress commented there, will follow up in that PR.

Thanks! Please make sure that you're using the latest commit of boringssl as they recently improved it so that we don't need bazel-side changes.

PR #32256, opened 2023-02-01
Add lb policy example

gnossen:
The design of this example is "short and sweet." Both C++ and Go's examples use APIs that Python simply does not have.

C++ has a first-class API that wraps the raw channel arg that we're supplying here, so that actually is pretty one-to-one. Go's example, however, uses a full service config _and_ it ensures that you can see the load balancing policy in action on a single local machine. Python is not capable of this.

Load balancing fundamentally requires more than one backend. This _is_ possible with DNS, but not with static hosts like `/etc/hosts`. Go gets around this by creating a simple fake resolver using its custom resolver API, which neither C++ nor Python has. So we do the best we can and just write a comment saying "swap out localhost for something else to see this in action."


XuanWang-Amos (approved):
Looks like we still have many gaps between languages.

PR #32262, opened 2023-02-01
Revert "EventEngine::RunAfter: GrpcLb"

markdroth:
Reverts grpc/grpc#30043

This PR seems to be causing a flake where grpclb_end2end_test times out in opt builds in the `SingleBalancerTest.Fallback` test.


drfloob (approved):
LGTM, presuming it can be rolled back cleanly.

PR #32269, opened 2023-02-02
Reland "EventEngine::RunAfter: GrpcLb" (#32262)"

drfloob:
There was a ~1% flake in grpclb end2end tests that was reproducible in opt builds, manifesting as a hang, usually in a the `SingleBalancerTest.Fallback` test. Through experimentation, I found that by skipping the death test in the grpclb end2end test suite, the hang was no longer reproducible in 10,000 runs. Similarly, moving this test to the end of the suite, or making it run first (as is the case in this PR) resulted in 0 failures in 3000 runs.

It's unclear to me yet why the death test causes things to be unstable in this way. It's clear from the logs that one test does affect the rest, `grpc_init` is done once for all tests, so all tests utilize the same EventEngine ... until the death test completes, and a new EventEngine is created for the next test.

I think this death test is sufficiently artificial that it's fine to change the test ordering itself, and ignore the wonky intermediate state that results from it.

Reproducing the flake:

```
tools/bazel --bazelrc=tools/remote_build/linux.bazelrc test \
  -c opt \
  --test_env=GRPC_TRACE=event_engine \
  --runs_per_test=5000 \
  --test_output=summary \
  test/cpp/end2end/grpclb_end2end_test@poller=epoll1 
```

### DIFF
--- a/test/cpp/end2end/grpclb_end2end_test.cc
+++ b/test/cpp/end2end/grpclb_end2end_test.cc
@@ -1016,19 +1016,6 @@ TEST_F(SingleBalancerTest, SecureNaming) {
   EXPECT_EQ("grpclb", channel_->GetLoadBalancingPolicyName());
 }
 
-TEST_F(SingleBalancerTest, SecureNamingDeathTest) {
-  GTEST_FLAG_SET(death_test_style, "threadsafe");
-  // Make sure that we blow up (via abort() from the security connector) when
-  // the name from the balancer doesn't match expectations.
-  ASSERT_DEATH_IF_SUPPORTED(
-      {
-        ResetStub(0, kApplicationTargetName_ + ";lb");
-        SetNextResolution({AddressData{balancers_[0]->port_, "woops"}});
-        channel_->WaitForConnected(grpc_timeout_seconds_to_deadline(1));
-      },
-      "");
-}
-
 TEST_F(SingleBalancerTest, InitiallyEmptyServerlist) {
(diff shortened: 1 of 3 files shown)


test/cpp/end2end/grpclb_end2end_test.cc
```diff
@@ -1899,6 +1886,19 @@ TEST_F(SingleBalancerTest, DropAll) {
   EXPECT_EQ(status.error_message(), "drop directed by grpclb balancer");
 }
 
+TEST_F(SingleBalancerTest, SecureNamingDeathTest) {
```
ctiller, line 1889: Can you add a comment letting us know to keep it at the end?

drfloob: Done

markdroth (approved):

test/cpp/end2end/grpclb_end2end_test.cc
```diff
@@ -764,6 +764,26 @@ class GrpclbEnd2endTest : public ::testing::Test {
   const std::string kApplicationTargetName_ = "application_target_name";
 };
 
+// This death test is kept separate from the rest to ensure that it's run before
+// any others. See https://github.com/grpc/grpc/pull/32269 for details.
+class SingleBalancerDeathTest : public GrpclbEnd2endTest {
```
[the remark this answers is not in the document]
drfloob: Done. To be safe, I reconfirmed the flake did not show up in 1000 more runs.

Issue #30927, opened 2022-09-09
Support 3.11 on aarch64


gnossen:
The latest version of dockcross aarch64 does not yet support Python 3.11, blocking us from creating these artifacts. Once they _do_ support it, we should start producing these artifacts. This image has not been updated on dockerhub for a year, so it's possible that this may require more than just waiting for a new release. I've raised [the issue here](https://github.com/dockcross/dockcross/issues/722) to find out.

mgu:
maybe an alternative could be to switch to pypa/cibuildwheel ?

samanta-widjaja:
is there a solution. i can't install due the same problem too

mgu:
@gnossen do you think I could open a PR to switch from dockcross to qemu+aarch64 image instead ?
Even if it is super slow, at least it would work

gnossen:
> **maybe an alternative could be to switch to pypa/cibuildwheel ?**

I haven't looked deeply into this, but it seems that integrates at the level of the CI system, e.g. Github Actions, Travis, Circle CI. For software supply chain security reasons, we use a private system called Kokoro and we would not be able to use these cibuildwheel directly, though we might be able to pull some components out of it to enable this case.

> do you think I could open a PR to switch from dockcross to qemu+aarch64 image instead ?

We'd have to play around with CI configuration to make sure that our PR presubmits don't get blocked by a 5 hour build, but in principle, yes, a contribution like that would be greatly appreciated. Thank you for the offer!

mgu:
I think the qemu is way too slow (the build timed out after 2h 😓 ).
I'll do some more tests to understand what can be done

mgu:
```
PASSED: build_artifact.python_manylinux2014_aarch64_cp311-cp311 [time=798.4sec, retries=0:0]
```
I've been able to build a dockcross image with python 3.11 🎉 
I'll try to create PRs for dockross & grpc in the next few days

kasium:
It seems that the latest release 1.15.3 still doesn't include a 3.11 aarch wheel. Can you please check?

mgu:
Hi @kasium 
From my understanding, the aarch wheel will be available from 1.53.x (1.51.3 is a maintenance release and 1.52 was yanked)

kasium:
@mgu I see. Thanks!

PR #32270, opened 2023-02-02
Support python 3.11 on aarch64

mgu:
Fixes https://github.com/grpc/grpc/issues/30927

I've build the artifact on amd64 and tested the wheel on Mac M1.

I'm not sure about what do do with grpc_artifact_python_manylinux2014_aarch64.current_version


mgu:
cc @gnossen

tools/dockerfile/grpc_artifact_python_manylinux2014_aarch64/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM dockcross/manylinux2014-aarch64:20210826-19322ba
+FROM dockcross/manylinux2014-aarch64:latest
```
gnossen, line 15: For reasons of reproducibilty, we prefer to pin to tags that are less likely to change.

gnossen:
Super excited to see this PR! Thank you for the contribution.

mgu:
Thanks for adding the tag, I was not sure because some images don't have one

gnossen:
@mgu I think what you've already done is above and beyond the call of duty. The last bit will be fiddly CI-specific stuff that I'll take care of. Thank you very much for the contribution!

ewdurbin:
Thanks so much to the many folks who were involved in seeing this through. This is a huge improvement in my day to day quality of life working on PyPI. 💜

PR #32273, opened 2023-02-02
Add enforcement for cpp-style comment license in check_copyright.py

yijiem:
Also fixing things during the interim.


tools/distrib/check_copyright.py
```diff
@@ -246,6 +281,13 @@ def write_copyright(license_text, file_text, filename):
     m = re.search(re_license, text)
     if m:
         pass
+    elif enforce_cpp_style_comment:
+        log(1, 'copyright missing or does not use cpp comment style', filename)
+        if args.fix:
+            # attempt fix; search for c-style comment license and replace it
+            # with cpp-style comment license
+            replace_copyright(license_text, text, filename)
```
markdroth, line 289: If there's no C-style copywrite, don't we still need to add the C++-style one?

markdroth (approved):

tools/distrib/check_copyright.py
yijiem: Good callout. Added `write_copyright` if `replace_copyright` did not work.

PR #32274, opened 2023-02-02
Use WaitWithTimeout for gpr_cv_wait

veblush:
This is a prerequisite for upcoming Abseil change using a monotonic clock for `WaitWithTimeout`. This change allows gRPC gpr_cv_wait to use `WaitWithTimeout` when it's given monotonic clock or timespan.

Caveat: This won't change the actual gRPC behavior until Abseil gets new change regarding this.


yijiem (approved):
Thanks Esun!

PR #32106, opened 2023-01-14
server: introduce ServerMetricRecorder API and move per-call reporting from a C++ interceptor to a C-core filter

yousukseung:
Support backend metric sampling and make orca intereptor a channel filter

PR #32279, opened 2023-02-03
Revert "Revert "server: introduce ServerMetricRecorder API and move per-call reporting from a C++ interceptor to a C-core filter (#32106)" (#32272)"

yousukseung:
PR #32106 caused msan errors in some tests while de-referencing the
server object where async calls are active after the server is
destroyed. Instead cache the ServerMetricRecorder pointer.


markdroth (approved):

ctiller:
Please cherry-pick this change in case of a second revert.

PR #32280, opened 2023-02-03
PSM Interop: Switch to the new GKE clusters (except the URL map)

sergiitk:
Our test clusters are in the broken/dirty state (except the URL map's "basic" cluster, which isn't affected).  
This PR switches to the newly created clusters to:
1. Get a data point on whether newly created clusters are affected by the same issues.
2. Allow for descriptive work on the old clusters.
3. Hopefully, bring our tests back to green.

Bonus: more sensible cluster names.


sergiitk:
Tests (WIP):

- [ ] [grpc/core/v1.52.x/branch/linux/psm-security](https://source.cloud.google.com/results/invocations/a6e4c839-5e36-4c6c-823f-e47eb65ba3a7)

gnossen (approved):

PR #32285, opened 2023-02-03
DRAFT: Grpc.Tools detect alpine linux

tonydnewell:
This contains changes to the Grpc.Tools package to detect running on alpine linux and adds placeholders for using x86 and x64 alpine linux protoc and grpc_csharp_plugin binaries in MSBuild.

NOTE: **It does not build or provide the alpine linux tools binaries**. It assumes the alpine artifacts are already built and are available in `${EXTERNAL_GIT_ROOT}"/input_artifacts/`

Thus there are the changes that would need to be made to Grpc.Tools if we built the alpine linux tools - so is marked as a draft PR until we do so.

The detection of alpine linux is done by reading the file `/etc/os-release` if it exists. This is done in a new class `LinuxDistroDetection` rather than adding to the existing `CommonPlatformDetection` class to separate it out (as the latter class is a copy from elsewhere) and make it easier to test. The code can be moved into  `CommonPlatformDetection` if that was felt to be a more appropriate place.

### DIFF
--- a/src/csharp/Grpc.Tools.Tests/LinuxDistroDetectionTest.cs
+++ b/src/csharp/Grpc.Tools.Tests/LinuxDistroDetectionTest.cs
@@ -1,0 +1,35 @@
+#region Copyright notice and license
+
+// Copyright 2023 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#endregion
+
+using System.IO;
+using System.Reflection;
+using NUnit.Framework;
+
+namespace Grpc.Tools.Tests
+{
+    public class LinuxDistroDetectionTest
+    {
+        [Test]
+        public void ReadLinuxDistroFile()
+        {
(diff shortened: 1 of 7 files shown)


tonydnewell:
This is how dotnet itself determines the Linux variant by reading `/etc/os-release` file:
https://github.com/dotnet/sdk/blob/033003f9fdbd09e33b3b2882c21d338d26bba88e/src/Cli/Microsoft.DotNet.Cli.Utils/RuntimeEnvironment.cs#L140
and
https://github.com/dotnet/arcade/blob/759bbe7e2a8fc570c8a742e52cf1605cb7d3e9c2/src/Microsoft.DotNet.XUnitConsoleRunner/src/common/AssemblyResolution/Microsoft.DotNet.PlatformAbstractions/Native/PlatformApis.cs#L102
and other places

There is no other way. There is no system call, library function, environment variable or command. E.g. `uname` only gives kernel info not distro info.  (We do call `uname` to determine if we're running on MacOS since "Darwin" is returned in the kernel info)

tonydnewell:
Some options:
- only detect alpine for .NET 5 or later and document workaround for older versions (and use `RuntimeInformation.RuntimeIdentifer`)
- or for older versions of .NET read the `/etc/os-release` file
- or always reading the  `/etc/os-release` file
- or don't try and detect alpine and just document workaround

Dotnet itself reads `/etc/os-release` for the runtime information is C++:
https://github.com/dotnet/dotnet/blob/66e89f0c81e2159f0ea69993ca244e4f1a4c4eb6/src/runtime/src/native/corehost/hostmisc/pal.unix.cpp#L801


Note: output from `RuntimeInformation` on alpine is:
```
~/rti # dotnet run
RuntimeIdentifier: alpine.3.17-x64
OSArchitecture:    X64
OSDescription:     Linux 5.15.0-58-generic #64~20.04.1-Ubuntu SMP Fri Jan 6 16:42:31 UTC 2023
```

tonydnewell:
Closing as there is a workaround for using alpine built binaries with Grpc.Tools

PR #32291, opened 2023-02-03
Schedule the cancellation on_complete callback in transport immediately if supported

Vignesh2208:


src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1311,14 +1314,16 @@ static void perform_stream_op_locked(void* stream_op,
     t->num_messages_in_next_write++;
     grpc_core::global_stats().IncrementHttp2SendMessageSize(
         op->payload->send_message.send_message->Length());
-    on_complete->next_data.scratch |= CLOSURE_BARRIER_MAY_COVER_WRITE;
```
ctiller, line 1314: it's probably worth adding a line up the top:

`const auto closure_barrier_may_cover_write_flags = grpc_event_engine_run_in_background()? 0 : CLOSURE_BARRIER_MAY_COVER_WRITE;`

so that you don't need to recheck this multiple times

(honestly it might also be worth storing those bits on the `grpc_transport` instance and checking just once at initialization)

Vignesh2208: Good point. Fixed.

PR #32294, opened 2023-02-03
GCP Observability : Framework for detecting the environment

yashykt:
This code is not plumbed through yet, but it provides the core infrastructure needed to detect the proper GCP environment resources needed to set up the labels/attributes/resources for stats, tracing and logging.

Details on how the various environment resources are setup has been derived by looking at java's cloud logging library and OpenTelemetry's future plans. (Could be better explained in an offline review since some links are internal).

Requesting @veblush for a full review and @markdroth for a structural review.


veblush (approved):

src/cpp/ext/filters/census/environment_autodetect.cc
```diff
@@ -0,0 +1,337 @@
...
+  struct Attribute {
+    std::string resource_attribute;
+    std::string metadata_server_atttribute;
+  };
+
+  const std::string project_id_;
+  grpc_polling_entity* pollent_;
+  absl::AnyInvocable<void(EnvironmentAutoDetect::ResourceType)> on_done_;
+  grpc_core::Mutex mu_;
+  std::list<Attribute> attributes_to_fetch_ ABSL_GUARDED_BY(mu_);
```
ctiller, line 241: This seems like it would be simpler & clearer if it was a `flat_hash_map<string /* metadata_server_attribute */, string /* resource_attribute */>`

yashykt: suggestion taken

src/cpp/ext/filters/census/environment_autodetect.cc
```diff
@@ -0,0 +1,421 @@
...
+      : InternallyRefCounted(/*trace=*/nullptr, /*initial_refcount=*/2),
+        project_id_(std::move(project_id)),
+        on_done_(std::move(on_done)) {
+    grpc_core::ExecCtx exec_ctx;
+    // TODO(yashykt): The pollset stuff should go away once the HTTP library is
+    // ported over to use EventEngine.
+    pollset_ = static_cast<grpc_pollset*>(gpr_zalloc(grpc_pollset_size()));
+    grpc_pollset_init(pollset_, &mu_poll_);
+    pollent_ = grpc_polling_entity_create_from_pollset(pollset_);
+    poller_ = grpc_core::Thread(
```
drfloob, line 223: This can be replaced by `GetDefaultEventEngine()->Run([this]{ PollLoop(); });`. 

Note that with either a thread like this or using EventEngine::Run, we don't strictly have to worry about fork support since this is in C++ code, not the core library where Python requires a clean fork story, but it might be worth adding a comment that this is not fork-safe for anyone who wants fork support in C++. If you want to make this fork-safe, we can have a short design chat.

yashykt: Suggestion accepted. Added comments to that affect when we actually care about fork support here.

src/cpp/ext/filters/census/environment_autodetect.cc
```diff
@@ -0,0 +1,425 @@
...
+constexpr const char kZoneAttribute[] = "/computeMetadata/v1/instance/zone";
+constexpr const char kClusterNameAttribute[] =
+    "/computeMetadata/v1/instance/attributes/cluster-name";
+constexpr const char kRegionAttribute[] = "/computeMetadata/v1/instance/region";
+constexpr const char kInstanceIdAttribute[] = "/computeMetadata/v1/instance/id";
+
+// Fire and Forget class (Cleans up after itself.)
+// Fetches the value of an attribute from the MetadataServer on a GCP
+// environment.
+class MetadataQuery {
```
markdroth, line 78: Since this code essentially already exists in the google-c2p resolver, I think we should move it into a common library instead of duplicating it.

src/cpp/ext/filters/census/environment_autodetect.cc
```diff
@@ -0,0 +1,426 @@
...
+        on_done_(std::move(on_done)) {
+    grpc_core::ExecCtx exec_ctx;
+    // TODO(yashykt): The pollset stuff should go away once the HTTP library is
+    // ported over to use EventEngine.
+    pollset_ = static_cast<grpc_pollset*>(gpr_zalloc(grpc_pollset_size()));
+    grpc_pollset_init(pollset_, &mu_poll_);
+    pollent_ = grpc_polling_entity_create_from_pollset(pollset_);
+    // TODO(yashykt): Note that using EventEngine::Run is not fork-safe. If we
+    // want to make this fork-safe, we might need some re-work here.
+    grpc_event_engine::experimental::GetDefaultEventEngine()->RunAfter(
```
drfloob, line 229: Note that this is potentially unsafe usage of `GetDefaultEventEngine`. It returns a `shared_ptr`, and will lazily create an instance if one does not already exist. If you do not save the engine locally, you risk having the engine created and destroyed on this line.

If you aren't certain the default engine's lifetime is guaranteed until your timer is complete, I'd recommend adding a `shared_ptr<EventEngine> engine_` member to `EnvironmentAutoDetectHelper` to cache it for the life of the object at least.

src/cpp/ext/filters/census/environment_autodetect.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "absl/base/thread_annotations.h"
+#include "absl/functional/any_invocable.h"
+
+#include "src/core/lib/gprpp/sync.h"
+
+namespace grpc {
+namespace internal {
+
+class EnvironmentAutoDetect {
```
markdroth, line 35: Does this code actually belong in the census filter directory?  It seems to me that it's really a more general GCP-specific autodetection thing, not really anything specific to the census filter.  Maybe this should go in its own library somewhere?

Note that we also have some other existing code to detect GCP platform here, which we use in ALTS and in GoogleDefaultCreds:

https://github.com/grpc/grpc/blob/0ecc18ef0f51eb38880835db34f38f99980367e4/src/core/lib/security/credentials/alts/check_gcp_environment.h#L55

@apolcyn may have some thoughts about the best way to structure all of this code into a single combined library.

src/cpp/ext/filters/census/environment_autodetect.h
```diff
@@ -0,0 +1,71 @@
...
+ public:
+  struct ResourceType {
+    // For example, "gce_instance", "gke_container", etc.
+    std::string resource_type;
```
markdroth, line 39: Suggest making this an enum instead of a string.

src/cpp/ext/filters/census/environment_autodetect.h
```diff
@@ -0,0 +1,71 @@
...
+    // Values for all the labels listed in the associated resource type.
+    std::map<std::string, std::string> labels;
+  };
+
+  // A Create() call properly sets up the environment detector with the
+  // project_id. All subsequent calls can use a Get() without needing to mention
+  // the project_id.
+  static EnvironmentAutoDetect& Create(std::string project_id);
```
markdroth, line 47: Splitting `Create()` and `Get()` into two different methods seems odd.  If there are multiple callers, how do any of them know which one is the first one?

Also, what happens if `Create()` is called twice, but with different project IDs?

src/cpp/ext/filters/census/environment_autodetect.cc
```diff
@@ -0,0 +1,425 @@
...
+        }
+        for (auto& callback : callbacks) {
+          callback();
+        }
+      });
+}
+
+void EnvironmentAutoDetect::NotifyOnDone(absl::AnyInvocable<void()> callback) {
+  {
+    grpc_core::ReleasableMutexLock lock(&mu_);
```
markdroth, line 413: I generally consider `ReleasableMutexLock` to be a code smell.  And in this case, since you're invoking the callback asynchronously, it looks like you don't actually need to release the lock early in the first place.

yashykt: Done. I was intending to reduce the critical region as much as I could, but the EventEngine::Run call is not so bad that we need to worry about it here.

src/cpp/ext/filters/census/environment_autodetect.cc
```diff
@@ -0,0 +1,349 @@
...
+std::string GetVersionId() {
+  return grpc_core::GetEnv("GAE_VERSION").value_or("");
+}
+
+// Fire and forget class
+class EnvironmentAutoDetectHelper
+    : public grpc_core::InternallyRefCounted<EnvironmentAutoDetectHelper>,
+      private internal::GrpcLibrary {
+ public:
+  explicit EnvironmentAutoDetectHelper(
```
drfloob, line 126: I don't see this constructor used with an initializer list, and I'm not sure you'd want to prevent that regardless, so removing `explicit` here seems the right thing to do.

src/core/ext/gcp/metadata_query.cc
```diff
@@ -0,0 +1,113 @@
...
+  http_request->Start();
+}
+
+MetadataQuery::~MetadataQuery() { grpc_http_response_destroy(&response_); }
+
+void MetadataQuery::OnDone(void* arg, absl::Status error) {
+  auto* self = static_cast<MetadataQuery*>(arg);
+  std::string result;
+  if (!error.ok()) {
+    gpr_log(GPR_ERROR, "MetadataServer Query failed for %s: %s",
```
drfloob, line 89: Small suggestion: guard the logging behind a trace flag. Trace flags can be set on by default, if that's your intention, but it will provide an easy toggle later if you want to disable this logging as on by default.

src/cpp/ext/filters/census/environment_autodetect.cc
```diff
@@ -0,0 +1,349 @@
...
+    if (!GRPC_LOG_IF_ERROR(
+            "pollset_work",
+            grpc_pollset_work(grpc_polling_entity_pollset(&pollent_), &worker,
+                              grpc_core::Timestamp::InfPast()))) {
+      notify_poller_ = true;
+    }
+    done = notify_poller_;
+    gpr_mu_unlock(mu_poll_);
+    if (!done) {
+      event_engine_->RunAfter(grpc_core::Duration::Milliseconds(100),
```
drfloob, line 183: The performance hit is probably negligible, but I wonder if this would be better written as a loop with a 100ms sleep, than a lambda-wrapped recursive function with a thread hop.

```
bool done = false;
while (!done) {
  // ...
}
Unref();
```

src/cpp/ext/filters/census/environment_autodetect.cc
yashykt: Done

src/core/ext/gcp/metadata_query.cc
yashykt: Note that all these logs are error logs which we generally do not guard by Trace Flags. I think we always want to see these logs.

src/cpp/ext/filters/census/environment_autodetect.cc
yashykt: Isn't it bad to do a blocking sleep in an event manager thread. That seems like something we definitely do not want to do. Also note that in this case, we really do not care about performance.

src/cpp/ext/filters/census/environment_autodetect.cc
yashykt: Thanks!

src/cpp/ext/filters/census/environment_autodetect.h
yashykt: Moved it to src/cpp/ext/gcp/... Alex, let's chat on whether we'd like to move some of this in Core if it's reusable.
I'd like to not block this PR though.

src/core/ext/gcp/metadata_query.h
```diff
@@ -0,0 +1,64 @@
...
+#include "src/core/lib/http/parser.h"
+#include "src/core/lib/iomgr/closure.h"
+#include "src/core/lib/iomgr/polling_entity.h"
+
+namespace grpc_core {
+
+// Fire and Forget class (Cleans up after itself.)
+// Fetches the value of an attribute from the MetadataServer on a GCP
+// environment.
+class MetadataQuery {
```
apolcyn, line 36: The existing `MetadataQuery` object used by the google-c2p resolver is cancellable, and e.g. the resolver will cancel outstanding MD server queries during shutdown.

Looks like cancellation is removed from the revised API, is that intentional?

src/core/ext/gcp/metadata_query.cc
```diff
@@ -0,0 +1,113 @@
...
+  memset(&request, 0, sizeof(grpc_http_request));
+  grpc_http_header header = {const_cast<char*>("Metadata-Flavor"),
+                             const_cast<char*>("Google")};
+  request.hdr_count = 1;
+  request.hdrs = &header;
+  // The http call is local. If it takes more than one sec, it is probably not
+  // on GCP.
+  auto http_request = grpc_core::HttpRequest::Get(
+      std::move(*uri), nullptr /* channel args */, pollent, &request,
+      grpc_core::Timestamp::Now() + grpc_core::Duration::Seconds(1), &on_done_,
```
apolcyn, line 76: 1 second seems like a fairly short default timeout. The google-c2p resolver currently uses 10 seconds. (I'm not sure what the "right" value is, but we may want to give some slack e.g. for packet loss b/t the client and MD server).

Maybe, make this timeout configurable?

src/core/ext/gcp/metadata_query.cc
```diff
@@ -0,0 +1,113 @@
...
+MetadataQuery::~MetadataQuery() { grpc_http_response_destroy(&response_); }
+
+void MetadataQuery::OnDone(void* arg, absl::Status error) {
+  auto* self = static_cast<MetadataQuery*>(arg);
+  std::string result;
+  if (!error.ok()) {
+    gpr_log(GPR_ERROR, "MetadataServer Query failed for %s: %s",
+            self->attribute_.c_str(), grpc_core::StatusToString(error).c_str());
+  } else if (self->response_.status != 200) {
+    gpr_log(GPR_ERROR,
```
apolcyn, line 92: We shouldn't log ERROR here, since there are some queries that are performed as "probes" for which we may or may not get a 200 response, but we can handle both cases.

E.g. the google-c2p resolver queries the MD server to see if the VM has an IPv6 address, to decide whether to prefer v4 or v6.

If we want to add extra error message context, suggest adding it to the error and letting the caller decide to log or not.

apolcyn (changes requested):

src/core/ext/gcp/metadata_query.cc
drfloob: Some are DEBUG, and the first one in the Poll loop is an ERROR and it maybe should not be? Your call though.

src/core/ext/gcp/metadata_query.h
yashykt: MetadataQuery requests are intended to be quick enough that I don't think that the cancellation matters so much but it is also trivial enough to add, so I don't mind adding it.

src/core/ext/gcp/metadata_query.cc
yashykt: Given that Alex had similar concerns, guarded it with TraceFlags

src/core/ext/gcp/metadata_query.cc
yashykt: Done!

src/core/ext/gcp/metadata_query.h
yashykt: Done

src/core/ext/gcp/metadata_query.cc
```diff
@@ -0,0 +1,127 @@
...
+  auto uri = URI::Create("http", "metadata.google.internal.", attribute_,
+                         {} /* query params */, "" /* fragment */);
+  GPR_ASSERT(uri.ok());  // params are hardcoded
+  grpc_http_request request;
+  memset(&request, 0, sizeof(grpc_http_request));
+  grpc_http_header header = {const_cast<char*>("Metadata-Flavor"),
+                             const_cast<char*>("Google")};
+  request.hdr_count = 1;
+  request.hdrs = &header;
+  // The http call is local. If it takes more than one sec, it is probably not
```
apolcyn, line 76: nit: stale comment now that timeout is a param?

src/core/ext/gcp/metadata_query.cc
```diff
@@ -0,0 +1,127 @@
...
+void MetadataQuery::OnDone(void* arg, absl::Status error) {
+  auto* self = static_cast<MetadataQuery*>(arg);
+  std::string result;
+  if (!error.ok()) {
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_metadata_query_trace)) {
+      gpr_log(GPR_INFO, "MetadataServer Query failed for %s: %s",
+              self->attribute_.c_str(), StatusToString(error).c_str());
+    }
+  } else if (self->response_.status != 200) {
+    if (GRPC_TRACE_FLAG_ENABLED(grpc_metadata_query_trace)) {
```
apolcyn, line 102: nit: as long as this tracer flag is enabled, it will probably be helpful to log all MD server query results, and the HTTP status will probably be useful to include in the log too.

e.g. suggest replacing:

```
if (!error.ok()) {
  ...
} else if (self->response_.status != 200) {
  ...
}
```

with just one log at the top:

```
if (GRPC_TRACE_FLAG_ENABLED(grpc_metadata_query_trace)) {
    gpr_log(GPR_INFO, "MetadataServer Query for %s: http status: %d, error: %s",
            self->attribute_.c_str(), self->response_.status, StatusToString(error).c_str());
}
```

src/core/ext/gcp/metadata_query.h
```diff
@@ -0,0 +1,70 @@
...
+ public:
+  static const char kZoneAttribute[];
+  static const char kClusterNameAttribute[];
+  static const char kRegionAttribute[];
+  static const char kInstanceIdAttribute[];
+  static const char kIPv6Attribute[];
+
+  MetadataQuery(std::string attribute, grpc_polling_entity* pollent,
+                absl::AnyInvocable<void(std::string /* attribute */,
+                                        std::string /* result */)>
```
apolcyn, line 48: `result` might be better as a `StatusOr<std::string>` in order to propagate debug info. E.g., ` StatusOr<std::string>` would give callers the reason for query failure without enabling tracers and correlating logs.

Note this API is fine as is for the google-c2p resolver, since the c2p resolver internally handles query failures without propagating failures, but I don't know about other callers.

src/core/ext/gcp/metadata_query.cc
yashykt: removed

src/core/ext/gcp/metadata_query.cc
yashykt: done

src/core/ext/gcp/metadata_query.h
yashykt: done

apolcyn (approved):
LGTM

src/core/BUILD
```diff
@@ -5310,6 +5304,37 @@ grpc_cc_library(
     ],
 )
 
+grpc_cc_library(
+    name = "metadata_query",
```
markdroth, line 5308: Suggest calling this `gcp_metadata_query`.

src/core/ext/gcp/metadata_query.cc
```diff
@@ -0,0 +1,131 @@
...
+          absl::StrFormat("MetadataServer Could not parse zone: %s",
+                          std::string(body).c_str()));
+      if (GRPC_TRACE_FLAG_ENABLED(grpc_metadata_query_trace)) {
+        gpr_log(GPR_INFO, "%s", result.status().ToString().c_str());
+      }
+    } else {
+      result = std::string(body.substr(pos + 1));
+    }
+  } else {
+    result = self->response_.body;
```
markdroth, line 123: Is the body guaranteed to be nul-terminated and to contain no nul characters before the end?  If not, this isn't safe.  I think instead it should be:

```
result = std::string(self->response_.body, self->response_.body_length);
```

src/core/ext/gcp/metadata_query.h
```diff
@@ -0,0 +1,72 @@
...
+                              absl::StatusOr<std::string> /* result */)>
+          callback,
+      Duration timeout);
+
+  ~MetadataQuery() override;
+
+  void Orphan() override;
+
+ private:
+  static void OnDone(void* arg, absl::Status error);
```
markdroth, line 59: I think we should use `grpc_error_handle` instead of `absl::Status` as the type for this parameter, since this is a C-style closure.  That will be a signal to us that this is something that needs to be cleaned up later.

src/cpp/ext/gcp/BUILD
```diff
@@ -123,3 +123,47 @@ grpc_cc_library(
...
 )
+
+grpc_cc_library(
+    name = "environment_autodetect",
+    srcs = [
+        "environment_autodetect.cc",
+    ],
+    hdrs = [
+        "environment_autodetect.h",
+        "//:include/grpcpp/ext/gcp_observability.h",
```
markdroth, line 134: Should this be in `public_hdrs`?

src/cpp/ext/gcp/environment_autodetect.h
```diff
@@ -59,7 +65,14 @@ class EnvironmentAutoDetect {
   }
 
  private:
+  friend absl::Status grpc::experimental::GcpObservabilityInit();
+
+  // GcpObservabilityInit() is responsible for setting up the singleton with the
+  // project_id.
+  static EnvironmentAutoDetect& Create(std::string project_id);
```
markdroth, line 72: I don't think `Create()` should return the instance anymore, since it's just for initializing the global singleton.  Instead, it should set a static data member (or just a global in the .cc file), and `Get()` should directly return that global instead of calling `Create()`.

test/cpp/ext/gcp/environment_autodetect_test.cc
```diff
@@ -55,25 +54,23 @@ class EnvironmentAutoDetectTest : public ::testing::Test {
...
-  absl::Notification notify;
+  grpc_core::Notification notify;
   GetNotifiedOnEnvironmentDetection(&env, &notify);
   notify.WaitForNotification();
 
   // Unless we test in a specific GCP resource, we should get "global" here.
-  EXPECT_EQ(env.resource()->resource_type, "global");
-  EXPECT_THAT(env.resource()->labels,
-              UnorderedElementsAre(Pair("project_id", "project")));
+  // EXPECT_EQ(env.resource()->resource_type, "global");
```
markdroth, line 62: Why is this commented out?

test/cpp/ext/gcp/environment_autodetect_test.cc
```diff
@@ -124,17 +118,17 @@ TEST_F(EnvironmentAutoDetectTest, AppEngine) {
...
   }
   for (int i = 0; i < 10; ++i) {
     notify[i].WaitForNotification();
   }
 
-  EXPECT_EQ(env.resource()->resource_type, "global");
-  EXPECT_THAT(env.resource()->labels,
-              UnorderedElementsAre(Pair("project_id", "project")));
+  // Unless we test in a specific GCP resource, we should get "global" here.
+  // EXPECT_EQ(env.resource()->resource_type, "global");
```
markdroth, line 130: Same question here.

markdroth (commented):
I like this structure much better!

test/cpp/ext/gcp/environment_autodetect_test.cc
yashykt: This is a problem of our testing environments. I can't be sure whether we'd be running on an environment that has MetadataServer or not, so unless I mock the MetadataServer, I can not get a stable result here across environments.

test/cpp/ext/gcp/environment_autodetect_test.cc
yashykt: This is a problem of our testing environments. I can't be sure whether we'd be running on an environment that has MetadataServer or not, so unless I mock the MetadataServer, I can not get a stable result here across environments.

src/core/ext/gcp/metadata_query.cc
yashykt: Done. Good catch! Thanks!

src/core/ext/gcp/metadata_query.h
yashykt: because we want c-style closures to go away? Done

src/cpp/ext/gcp/BUILD
yashykt: ah, yeah, removed

src/cpp/ext/gcp/environment_autodetect.h
yashykt: Done

markdroth (approved):

PR #32295, opened 2023-02-04
[EventEngine] WindowsEventEngine Listener

drfloob:
Test cherrypick is clean.


yijiem (commented):
Thanks AJ! This one is mostly learning for me (sorry for taking a long time to get back).

src/core/lib/event_engine/windows/iocp.cc
```diff
@@ -74,43 +73,36 @@ Poller::WorkResult IOCP::Work(EventEngine::Duration timeout,
...
-  if (GRPC_TRACE_FLAG_ENABLED(grpc_event_engine_trace)) {
-    gpr_log(GPR_DEBUG, "IOCP::%p got event on OVERLAPPED::%p", this,
-            overlapped);
+  GRPC_EVENT_ENGINE_POLLER_TRACE("IOCP::%p got event on OVERLAPPED::%p", this,
+                                 overlapped);
+  auto* socket = reinterpret_cast<WinSocket*>(completion_key);
+  if (socket->IsShutdown()) {
+    GRPC_EVENT_ENGINE_POLLER_TRACE("IOCP::%p deleting abandoned WinSocket::%p",
+                                   this, socket);
+    delete socket;
```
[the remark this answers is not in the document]
drfloob: Yes. IIUC, after closing a socket, an Overlapped event shows up, indicating to us that the kernel is done using the OVERLAPPED object, and the WinSocket is safe to destroy. This is what I've seen in my testing, but I could be wrong that the overlapped event arrives in every scenario. Once this lands, and we start to exercise more of the end2end tests, we'll see if memory leaks turn up.

src/core/lib/event_engine/windows/win_socket.cc
```diff
@@ -40,22 +43,21 @@ WinSocket::WinSocket(SOCKET socket, Executor* executor) noexcept
...
-    }
+    GRPC_EVENT_ENGINE_ENDPOINT_TRACE("WinSocket::%p already shutting down",
+                                     this);
     return;
   }
-  if (GRPC_TRACE_FLAG_ENABLED(grpc_event_engine_trace)) {
-    gpr_log(GPR_DEBUG, "WinSocket::%p shutting down now. Reason: %s", this,
-            why.ToString().c_str());
-  }
+  GRPC_EVENT_ENGINE_ENDPOINT_TRACE("WinSocket::%p shutting down now.", this);
```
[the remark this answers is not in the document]
drfloob: Added to `Orphan` as recommended, thanks.

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -0,0 +1,327 @@
...
+        GPR_ASSERT(GRPC_LOG_IF_ERROR("SinglePortSocketListener::Start",
+                                     StartLocked()));
+      };
+  auto read_info = listener_socket_->read_info();
+  if (read_info->wsa_error() != 0) {
+    gpr_log(GPR_INFO, "%s",
+            grpc_core::StatusToString(
+                GRPC_WSA_ERROR(read_info->wsa_error(),
+                               "Skipping on_accept due to error"))
+                .c_str());
```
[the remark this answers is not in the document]
drfloob: Good question, I think we are a bit inconsistent with it. `StatusToString` will give more information if there are any child errors / payloads in the Status, but `GRPC_WSA_ERROR` produces a status with no children or other payloads, so it doesn't much matter here. I've replaced this with a basic `Status::ToString`.

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -0,0 +1,327 @@
...
+    WindowsEventEngineListener* listener, LPFN_ACCEPTEX AcceptEx,
+    grpc_core::OrphanablePtr<WinSocket> win_socket, int port)
+    : AcceptEx(AcceptEx),
+      listener_(listener),
+      on_accept_([this]() { OnAcceptCallbackImpl(); }),
+      port_(port),
+      listener_socket_(std::move(win_socket)) {
+  EventEngine::ResolvedAddress tmp_addr;
+  int sockname_temp_len = sizeof(struct sockaddr_storage);
+  if (getsockname(listener_socket_->socket(),
```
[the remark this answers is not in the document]
drfloob: Done.

src/core/lib/event_engine/windows/iocp.cc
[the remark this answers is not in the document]
yijiem: Sounds good, thanks for the explanation!

src/core/lib/event_engine/windows/windows_listener.cc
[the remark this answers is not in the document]
yijiem: Looks good! Thanks!

src/core/lib/event_engine/windows/windows_listener.cc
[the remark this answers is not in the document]
yijiem: Gotcha, I missed the part that `grpc_core::StatusToString` seems to handle child status recursively.

yijiem (approved):

src/core/lib/event_engine/windows/windows_listener.h
```diff
@@ -0,0 +1,126 @@
...
+      IOCP* iocp, AcceptCallback accept_cb,
+      absl::AnyInvocable<void(absl::Status)> on_shutdown,
+      std::unique_ptr<MemoryAllocatorFactory> memory_allocator_factory,
+      Executor* executor, const EndpointConfig& config);
+  ~WindowsEventEngineListener() override;
+  absl::StatusOr<int> Bind(const EventEngine::ResolvedAddress& addr) override;
+  absl::Status Start() override;
+
+ private:
+  friend class SinglePortSocketListener;
```
yashykt, line 48: SinglePortSocketListener is an inner class right? Do you still need this?

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -284,7 +285,6 @@ EventEngine::ConnectionHandle WindowsEventEngine::Connect(
       Run([on_connect = std::move(on_connect), status]() mutable {
         on_connect(status);
       });
-      watched_socket->MaybeShutdown(status);
```
yashykt, line 287: @drfloob as discussed please fix this

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -0,0 +1,323 @@
...
+  // Shut down each port listener before destroying this EventEngine::Listener
+  for (auto& port_listener : port_listeners_) {
+    port_listener.reset();
+  }
+  on_shutdown_(absl::OkStatus());
+}
+
+absl::StatusOr<int> WindowsEventEngineListener::Bind(
+    const EventEngine::ResolvedAddress& addr) {
+  if (started_) {
```
yashykt, line 257: I think accesses to `started_` should either be synchronized, or this should really just be an assert. (again as discussed)

src/core/lib/event_engine/windows/windows_listener.h
```diff
@@ -0,0 +1,126 @@
...
+    EventEngine::ResolvedAddress listener_sockname_;
+  };
+  absl::StatusOr<SinglePortSocketListener*> AddSinglePortSocketListener(
+      SOCKET sock, EventEngine::ResolvedAddress addr);
+
+  IOCP* iocp_;
+  AcceptCallback accept_cb_;
+  absl::AnyInvocable<void(absl::Status)> on_shutdown_;
+  bool started_{false};
+  std::unique_ptr<MemoryAllocatorFactory> memory_allocator_factory_;
```
yashykt, line 113: maybe make it a const

src/core/lib/event_engine/windows/windows_listener.h
```diff
@@ -0,0 +1,126 @@
...
+    // This will hold the socket for the next accept.
+    SOCKET accept_socket_ ABSL_GUARDED_BY(mu_) = INVALID_SOCKET;
+    // The listener winsocket.
+    grpc_core::OrphanablePtr<WinSocket> listener_socket_ ABSL_GUARDED_BY(mu_);
+    EventEngine::ResolvedAddress listener_sockname_;
+  };
+  absl::StatusOr<SinglePortSocketListener*> AddSinglePortSocketListener(
+      SOCKET sock, EventEngine::ResolvedAddress addr);
+
+  IOCP* iocp_;
```
yashykt, line 109: maybe change the ordering of the data members to start with the stuff that doesn't change through the life of the object.

src/core/lib/event_engine/windows/windows_listener.h
```diff
@@ -0,0 +1,126 @@
...
+  AcceptCallback accept_cb_;
+  absl::AnyInvocable<void(absl::Status)> on_shutdown_;
+  bool started_{false};
+  std::unique_ptr<MemoryAllocatorFactory> memory_allocator_factory_;
+  grpc_core::Mutex socket_listeners_mu_;
+  std::list<std::unique_ptr<SinglePortSocketListener>> port_listeners_
+      ABSL_GUARDED_BY(socket_listeners_mu_);
+  const EndpointConfig& config_;
+  Executor* executor_;
```
yashykt, line 118: same here

src/core/lib/event_engine/windows/windows_listener.h
yashykt, line 109: same here

src/core/lib/event_engine/windows/iocp.cc
```diff
@@ -74,43 +73,36 @@ Poller::WorkResult IOCP::Work(EventEngine::Duration timeout,
...
     grpc_core::Crash(
         absl::StrFormat("Unknown custom completion key: %lu", completion_key));
   }
-  if (GRPC_TRACE_FLAG_ENABLED(grpc_event_engine_trace)) {
-    gpr_log(GPR_DEBUG, "IOCP::%p got event on OVERLAPPED::%p", this,
-            overlapped);
+  GRPC_EVENT_ENGINE_POLLER_TRACE("IOCP::%p got event on OVERLAPPED::%p", this,
+                                 overlapped);
+  auto* socket = reinterpret_cast<WinSocket*>(completion_key);
+  if (socket->IsShutdown()) {
```
yashykt, line 97: Let's check if the access to socket needs to be synchronized.
Consider cases of the time-of-check-to-time-of-use type

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -0,0 +1,323 @@
...
+        grpc_error_set_str(
+            GRPC_ERROR_CREATE_REFERENCING("Failed to prepare server socket",
+                                          &error, 1),
+            grpc_core::StatusStrProperty::kTargetAddress,
+            addr_uri.ok() ? *addr_uri : addr_uri.status().ToString()),
+        grpc_core::StatusIntProperty::kFd, static_cast<intptr_t>(sock));
+    if (sock != INVALID_SOCKET) closesocket(sock);
+    return error;
+  };
+
```
yashykt, line 209: channeling my Mark: replace the blank space with a comment, or remove it altogether

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -0,0 +1,323 @@
...
+    Executor* executor, const EndpointConfig& config)
+    : iocp_(iocp),
+      accept_cb_(std::move(accept_cb)),
+      on_shutdown_(std::move(on_shutdown)),
+      memory_allocator_factory_(std::move(memory_allocator_factory)),
+      config_(config),
+      executor_(executor) {}
+
+WindowsEventEngineListener::~WindowsEventEngineListener() {
+  auto shutdown_msg =
```
yashykt, line 245: just do this as part of the trace

src/core/lib/event_engine/windows/windows_engine.cc
drfloob: Good catch, fixed.

src/core/lib/event_engine/windows/windows_listener.cc
drfloob: blank removed :-). Hopefully the next line is sufficiently self-documenting.

src/core/lib/event_engine/windows/windows_listener.cc
drfloob: Done

src/core/lib/event_engine/windows/windows_listener.cc
drfloob: I made it an atomic, and also asserted that `Start` is only called once. Recapping the discussion a bit: if gRPC code is simultaneously calling `Bind` and `Start`, that's improper usage. This is a bit of defensive programming, but we might legitimately have improper usage that needs to be uncovered, so it's probably for the best to not crash here quite yet.

src/core/lib/event_engine/windows/windows_listener.h
drfloob: We do not need it, thanks. I had initially hidden the class declaration in the `cc` file, but opted to put it here instead.

src/core/lib/event_engine/windows/windows_listener.h
drfloob: I like that, done.

drfloob: iocp_ is now a const pointer to a non-const IOCP.

src/core/lib/event_engine/windows/windows_listener.h
drfloob: done

src/core/lib/event_engine/windows/windows_listener.h
drfloob: done

src/core/lib/event_engine/windows/iocp.cc
drfloob: Recap of the discussion: the concern here is if `socket->IsShutdown()` return false, but the socket is then shut down in some other thread while either A) the WinSocket callback is being run, or B) the subsequent IOCP::Work call deletes the WinSocket. 

This is valid, and I think points to the need for a ref counted WinSocket type. I'll need to think through it. Unfortunately at the moment, aside from crashes and UB, I don't think we have a good way to test for this, and it's probably a fairly rare occurrence.

drfloob:
This PR got a bit lanky, I'm going to split it up into a few different pieces for a cleaner commit/rollback story:
* the logging changes can be landed independently without much risk of rollbacks, 
* the endpoint and socket fixes are necessary, and discovered via the listener implementation work, but can be landed separately from the listener work to make the listener easier to review.
* the listener itself.

PR #32299, opened 2023-02-05
Make non-local static variables const

xuzhongxing:
Non-const non-local static variables are not guaranteed to be initialized before they are used.

This is hard to reproduce. I encounter this bug undeterministically in a CI task.

Reference: Embracing Modern C++ Safely

Despite C++11’s guarantee that each individual function-scope static initialization will occur at most once and before control can reach a point where the variable can be referenced, no analogous guarantees are made of nonlocal objects of static storage duration. Absence of this guarantee makes any interdependency in the initialization of such objects, especially across translation units (TUs), an abundant source of insidious errors.

Objects that undergo constant initialization have no such issue: Such objects will never be accessible at run time before having their initial values. Objects that are not constant initialized5 will instead be zero initialized until their constructors run, which itself might lead to undefined behavior that is not necessarily conspicuous.

### DIFF
--- a/src/core/lib/gprpp/global_config_env.h
+++ b/src/core/lib/gprpp/global_config_env.h
@@ -106,21 +106,21 @@ class GlobalConfigEnvString : public GlobalConfigEnv {
 // `help` argument is ignored for this implementation.
 
 #define GPR_GLOBAL_CONFIG_DEFINE_BOOL(name, default_value, help)         \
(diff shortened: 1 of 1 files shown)


xuzhongxingcas:
https://isocpp.org/wiki/faq/ctors#static-init-order

veblush:
These variables will be modified so it cannot get `const` here so it will get a segfault.

PR #32303, opened 2023-02-06
OpenCensus: Convert client filter to promises

yashykt:


src/cpp/ext/filters/census/client_filter.cc
```diff
@@ -68,53 +75,42 @@ constexpr uint32_t
...
-// OpenCensusClientChannelData
+// OpenCensusClientFilter
 //
 
-grpc_error_handle OpenCensusClientChannelData::Init(
-    grpc_channel_element* /*elem*/, grpc_channel_element_args* args) {
+const grpc_channel_filter OpenCensusClientFilter::kFilter =
+    grpc_core::MakePromiseBasedFilter<
+        OpenCensusClientFilter, grpc_core::FilterEndpoint::kClient,
+        grpc_core::kFilterExaminesServerInitialMetadata>("opencensus_client");
```
ctiller, line 84: I don't see where this examines server initial metadata?

yashykt: Thanks! Yeah, I think I misread it as client initial metadata. This makes more sense!

ctiller (approved):

PR #32329, opened 2023-02-08
[EventEngine] ThreadPool: manage fork and shutdown bits separately

drfloob:
The previous implementation assumed that shutdown and fork events could not occur at the same time, but that's not the case. This change adds separate tracking for fork and shutdown bits.

cc @gnossen


gnossen (approved):

gnossen:
I was finally able to reproduce and verify that this resolves the issue described in [this comment](https://github.com/grpc/grpc/issues/31885#issuecomment-1421904380).

PR #32334, opened 2023-02-09
grpc_build_system: add compile options "-std=c++14" to library definitions

dirkboye:
When importing com_github_grpc_grpc as an external dependency in a WORKSPACE which does not have "-std=c++14" as a setting in the bazelrc this leads to compiler errors as the default bazel toolchain uses c++0x.

Using c++14 for a complete project is not always desirable and with this change it's possible to import and use grpc++ libraries in such projects.

### DIFF
--- a/bazel/grpc_build_system.bzl
+++ b/bazel/grpc_build_system.bzl
@@ -163,6 +163,8 @@ def grpc_cc_library(
     copts = []
     if language.upper() == "C":
         copts = copts + if_not_windows(["-std=c11"])
+    if language.upper() == "C++":
+        copts = copts + ["-std=c++14"]
     linkopts = linkopts + if_not_windows(["-pthread"]) + if_windows(["-defaultlib:ws2_32.lib"])
     if select_deps:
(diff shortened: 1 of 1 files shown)


drfloob:
This change would eliminate optimizations and simplifications when compiling with newer standards, of which there are a few.

I did some quick research, and while it it appears _mostly_ safe for _most_ compilers to link objects built with _most_ standards (exception, for example: c++98 to c++11 compatibility [has some gotchas](https://gcc.gnu.org/wiki/Cxx11AbiCompatibility)), it's generally not recommended. I'm not sure we want to go out of our way to support that use case.

> Using c++14 for a complete project is not always desirable

If you have the c++14 toolchain available to build gRPC, why would you want to build parts of your code with an earlier standard?

veblush:
This is not what we're supposed to do when configuring to use c++14. Instead, your project needs to have the following in bazel.rc.

```
build --cxxopt='-std=c++14'
build --host_cxxopt='-std=c++14'
```

PR #32335, opened 2023-02-09
Added verified_root_cert_subject pass up through cpp api

gtcooke94:
PR #32215 added the verified root cert subject to the lower level `tsi_peer`. This PR is a companion to that and completes the feature by bubbling the information up to the `TsiCustomVerificationCheckRequest` which is part of the user facing API for implementing custom verification callbacks.


include/grpc/grpc_security.h
```diff
@@ -931,6 +931,9 @@ typedef struct grpc_tls_custom_verification_check_request {
      * grpc_security_constants.h.
      * TODO(ZhenLian): Consider fixing this in the future. */
     const char* peer_cert_full_chain;
+    /* The verified root cert subject.
+     * This value will only be filled if the validation was successful */
```
matthewstevenson88, line 935: nit: I'd suggest we say "cryptographic peer certificate verification" instead of "validation".

Rationale: I think it's important to disambiguate between the "validator" that we're modifying here (which occurs after the TLS handshake) and the cryptographic verification (which happens during the TLS handshake).

include/grpcpp/security/tls_certificate_verifier.h
```diff
@@ -62,10 +62,12 @@ class TlsCustomVerificationCheckRequest {
   grpc::string_ref peer_cert() const;
   grpc::string_ref peer_cert_full_chain() const;
   grpc::string_ref common_name() const;
+  grpc::string_ref verified_root_cert_subject() const;
```
matthewstevenson88, line 65: We need to be explicit about the contract/format for this API: for example,

* If verification fails or the peer cert is self-signed, this will be the empty string.
* If verification is successful, it is a comma-separated list, where the entries are of the form "FIELD_ABBREVIATION=string" and give a list of supported field abbreviations.

We also need tests to ensure that this contract is adhered to.

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,6 +161,32 @@ TEST(TlsCertificateVerifierTest,
...
 
+TEST(TlsCertificateVerifierTest, NoOpCertificateVerifierSucceedsWithVerifiedRootCertSubject) {
+  grpc_tls_custom_verification_check_request request;
+  memset(&request, 0, sizeof(request));
+  char* expected_subject = const_cast<char*>("CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU");
+  request.peer_info.verified_root_cert_subject = expected_subject;
+  auto verifier = std::make_shared<NoOpCertificateVerifier>();
+  TlsCustomVerificationCheckRequest cpp_request(&request);
+  EXPECT_EQ(cpp_request.verified_root_cert_subject(), expected_subject);
+  grpc::Status sync_status;
```
matthewstevenson88, line 172: Why is it necessary to call verifier->Verify here and below?

matthewstevenson88 (commented):
Thanks for the PR! Left a few comments. :)

test/cpp/security/tls_certificate_verifier_test.cc
gtcooke94: It's not necessary, I was just using the structure from other tests.
The real thing this is testing is whether or not the `TlsCustomVerificationCheckRequest` is properly populated from the `grpc_tls_custom_verification_check_request`.

The reason I didn't take it out is that calling Verify with an unset root cert does confirm that it doesn't cause a seg fault or something like that, so I figured it might have some value to leaving it in

include/grpc/grpc_security.h
gtcooke94: Changed

test/cpp/security/tls_certificate_verifier_test.cc
rockspore: It may be worth considering adding a test with an external verifier and make sure the root subject can be accessed inside the Verify() call

matthewstevenson88: Good call, makes sense.

include/grpcpp/security/tls_certificate_verifier.h
matthewstevenson88: SGTM! And please add unit tests somewhere that enforce this.

src/core/lib/security/security_connector/tls/tls_security_connector.cc
```diff
@@ -74,6 +74,7 @@ void PendingVerifierRequestInit(
   bool has_common_name = false;
   bool has_peer_cert = false;
   bool has_peer_cert_full_chain = false;
+  bool has_verified_root_cert_subject = false;
```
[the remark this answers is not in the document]
gtcooke94: I think I've added the relevant tests - existing tests should test the cases where it is unset, so I added tests where it is set. PTAL

include/grpcpp/security/tls_certificate_verifier.h
gtcooke94: Working on more unit tests for this, particularly the self-signed cert scenario.
In terms of listing all support field abbreviations, I hesitate to create a listing of them here - it should be a list of anything that is valid for the subject name of a certificate, so the best place to get that information is the user's certificate, or an RFC

test/cpp/util/tls_test_utils.cc
```diff
@@ -94,5 +94,17 @@ void AsyncCertificateVerifier::WorkerThread(void* arg) {
...
 }
 
+bool VerifiedRootCertSubjectVerifier::Verify(
+    TlsCustomVerificationCheckRequest* request,
+    std::function<void(grpc::Status)>, grpc::Status* sync_status) {
+  if (request->verified_root_cert_subject() != expected_subject_) {
+    *sync_status = grpc::Status(grpc::StatusCode::UNAUTHENTICATED,
+                                "VerifiedRootCertSubjectVerifier failed");
+  } else {
+    *sync_status = grpc::Status(grpc::StatusCode::OK, "");
```
rockspore, line 104: nit: I thought one could do grpc::Status::OK directly.

test/core/security/tls_security_connector_test.cc
```diff
@@ -737,6 +737,44 @@ TEST_F(TlsSecurityConnectorTest,
                             on_peer_checked);
 }
 
+TEST_F(TlsSecurityConnectorTest,
+       ChannelSecurityConnectorWithVerifiedRootCertSubjectSucceeds) {
+  auto* sync_verifier_ = new SyncExternalVerifier(true);
```
matthewstevenson88, line 742: Why do we need to use `new` here?

Can we just write:
```
SyncExternalVerifier sync_verifier(true);
```

matthewstevenson88, line 742: nit: The `_` suffix should not be used on this local variable, only for class member fields.

test/core/security/tls_security_connector_test.cc
```diff
@@ -737,6 +737,44 @@ TEST_F(TlsSecurityConnectorTest,
...
+      MakeRefCounted<TlsCredentials>(options);
+  ChannelArgs new_args;
+  RefCountedPtr<grpc_channel_security_connector> connector =
+      credential->create_security_connector(nullptr, kTargetName, &new_args);
+  EXPECT_NE(connector, nullptr);
+  TlsChannelSecurityConnector* tls_connector =
+      static_cast<TlsChannelSecurityConnector*>(connector.get());
+  EXPECT_NE(tls_connector->ClientHandshakerFactoryForTesting(), nullptr);
+  // Construct a basic TSI Peer.
+  char* expected_subject = const_cast<char*>(
```
matthewstevenson88, line 759: Can we store this as a `std::string` and call `.c_str()` when we need to convert to a `const char*`?

test/core/security/tls_security_connector_test.cc
```diff
@@ -737,6 +737,44 @@ TEST_F(TlsSecurityConnectorTest,
...
+      "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU");
+  tsi_peer peer;
+  GPR_ASSERT(tsi_construct_peer(2, &peer) == TSI_OK);
```
matthewstevenson88, line 762: Can we use EXPECT_EQ here and throughout?

test/core/security/tls_security_connector_test.cc
```diff
@@ -737,6 +737,44 @@ TEST_F(TlsSecurityConnectorTest,
...
+                 TSI_X509_VERIFIED_ROOT_CERT_SUBECT_PEER_PROPERTY,
+                 expected_subject, &peer.properties[1]) == TSI_OK);
+  RefCountedPtr<grpc_auth_context> auth_context;
+  ExecCtx exec_ctx;
+  grpc_closure* on_peer_checked = GRPC_CLOSURE_CREATE(
+      VerifyExpectedErrorCallback, nullptr, grpc_schedule_on_exec_ctx);
+  ChannelArgs args;
+  tls_connector->check_peer(peer, nullptr, args, &auth_context,
+                            on_peer_checked);
+}
```
matthewstevenson88, line 776: IIUC, we're trying to test the `PendingVerifierRequestInit` API. Where/how is that called by the above test?

test/core/security/tls_security_connector_test.cc
```diff
@@ -1089,6 +1127,38 @@ TEST_F(TlsSecurityConnectorTest,
   core_external_verifier->Unref();
 }
 
+TEST_F(TlsSecurityConnectorTest,
```
matthewstevenson88, line 1130: Similar comments here as above.

test/cpp/util/tls_test_utils.h
```diff
@@ -76,6 +76,25 @@ class AsyncCertificateVerifier
...
+
+  bool Verify(grpc::experimental::TlsCustomVerificationCheckRequest* request,
+              std::function<void(grpc::Status)> callback,
+              grpc::Status* sync_status) override;
+
+  void Cancel(grpc::experimental::TlsCustomVerificationCheckRequest*) override {
+  }
+
+ private:
+  char* expected_subject_ = nullptr;
```
matthewstevenson88, line 95: Please hold a `std::string` rather than `char*`, so we don't need to worry about memory management.

test/cpp/util/tls_test_utils.h
```diff
@@ -76,6 +76,25 @@ class AsyncCertificateVerifier
   std::deque<Request> queue_ ABSL_GUARDED_BY(mu_);
 };
 
+class VerifiedRootCertSubjectVerifier
+    : public grpc::experimental::ExternalCertificateVerifier {
+ public:
+  explicit VerifiedRootCertSubjectVerifier(char* expected_subject)
```
matthewstevenson88, line 82: Can we pass in an `absl::string_view` (preferred) or a `const std::string&` instead?

test/core/security/tls_security_connector_test.cc
rockspore, line 742: It doesn't seem to be a good pattern to introduce dependencies in test/cpp for test/core. It will be more tedious but I'd create a separate external verifier test class in core itself.

rockspore: Oh, nvm. I misread it.

gtcooke94: This is a pattern used in other tests in this file that I've re-used, I don't believe I've introduced a new dependency? Can you clarify what you mean?

include/grpcpp/security/tls_certificate_verifier.h
gtcooke94: Will add

test/cpp/util/tls_test_utils.cc
gtcooke94: Will change

include/grpcpp/security/tls_certificate_verifier.h
```diff
@@ -62,6 +62,13 @@ class TlsCustomVerificationCheckRequest {
   grpc::string_ref peer_cert() const;
   grpc::string_ref peer_cert_full_chain() const;
   grpc::string_ref common_name() const;
+  /**
+   * The subject name of the root certificate used to verify the peer chain
+   * If verification fails or the peer cert is self-signed, this will be an
+   * empty string If verification is successful, it is a comma-separated list,
```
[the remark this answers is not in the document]
gtcooke94: Will fix

include/grpcpp/security/tls_certificate_verifier.h
```diff
@@ -62,6 +62,13 @@ class TlsCustomVerificationCheckRequest {
...
+   * where the entries are of the form "FIELD_ABBREVIATION=string" that are
```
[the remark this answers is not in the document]
gtcooke94: Will fix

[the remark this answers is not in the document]
gtcooke94: Will add

test/core/security/tls_security_connector_test.cc
gtcooke94: This pattern copies the structure used in other tests - would we rather keep it matching all of the other tests or change it to this?

gtcooke94: My bad, fixed

rockspore: Since we don't see a explicit delete or free here. My intuition is the ownership of this is taken somewhere below. If you prepared this on a stack, it would be freed twice, which is not good.

test/core/security/tls_security_connector_test.cc
gtcooke94: We should be able to, but this is another of those patterns taken from the existing tests, so open question of changing just here, or keeping the pattern and doing a fix PR to clean the whole test file another time.

test/core/security/tls_security_connector_test.cc
gtcooke94: `PendingVerifierRequestInit` is only in the cpp file, not the header.

It's called indirectly through `connector->check_peer`. 

The stack is short - it's called [during the constructor of {Channel/Server}PendingVerifierRequest}](https://github.com/grpc/grpc/blob/d5fcbce4b4ff81f246bf63899f51f5a3bd56fb8c/src/core/lib/security/security_connector/tls/tls_security_connector.cc#L476), and those are created [during the associated check_peer](https://github.com/grpc/grpc/blob/d5fcbce4b4ff81f246bf63899f51f5a3bd56fb8c/src/core/lib/security/security_connector/tls/tls_security_connector.cc#L370).

I could simplify the test and just make sure that we can construct the `{Channel/Server}PendingVerifierRequest` object since `PendingVerifierRequestInit` is called during their constructors. Thoughts?

test/core/security/tls_security_connector_test.cc
gtcooke94: Addressed in [this comment chain](https://github.com/grpc/grpc/pull/32335#discussion_r1103285702)

test/cpp/util/tls_test_utils.h
gtcooke94: Will change

test/cpp/util/tls_test_utils.h
gtcooke94: Will change

test/cpp/util/tls_test_utils.cc
```diff
@@ -94,5 +94,18 @@ void AsyncCertificateVerifier::WorkerThread(void* arg) {
...
 }
 
+bool VerifiedRootCertSubjectVerifier::Verify(
+    TlsCustomVerificationCheckRequest* request,
+    std::function<void(grpc::Status)>, grpc::Status* sync_status) {
+  if (request->verified_root_cert_subject() != expected_subject_) {
+    *sync_status = grpc::Status(grpc::StatusCode::UNAUTHENTICATED,
+                                "VerifiedRootCertSubjectVerifier failed");
+  } else {
+    // *sync_status = grpc::Status(grpc::StatusCode::OK, "");
```
rockspore, line 104: nit: Remove the commented code, just in case you forgot.

gtcooke94: Good catch

test/core/security/tls_security_connector_test.cc
matthewstevenson88: I think we should change this now - using GPR_ASSERT is the old style and we should be migrating away from it in new code, IMO.

Same for char* to std::string.

test/core/security/tls_security_connector_test.cc
matthewstevenson88: This might be fine, I just want to make sure I understand. When were calling the `check_peer` API in this test, how do we know whether or not the peer property was present when we expect it to be present?

test/core/security/tls_security_connector_test.cc
```diff
@@ -1089,6 +1127,38 @@ TEST_F(TlsSecurityConnectorTest,
...
+       ServerSecurityConnectorWithVerifiedRootSubjectCertSucceeds) {
+  auto* sync_verifier = new SyncExternalVerifier(true);
+  ExternalCertificateVerifier core_external_verifier(sync_verifier->base());
+  RefCountedPtr<grpc_tls_credentials_options> options =
+      MakeRefCounted<grpc_tls_credentials_options>();
+  options->set_cert_request_type(GRPC_SSL_DONT_REQUEST_CLIENT_CERTIFICATE);
```
matthewstevenson88, line 1136: I'm guessing it doesn't make a big difference to the test, but this option should be set to something different, since if the server is not requesting the client cert then it doesn't make sense to get the verified root.

Can we just remove this line entirely? If it is necessary, can we change it to request and require verify?

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,6 +161,85 @@ TEST(TlsCertificateVerifierTest,
   EXPECT_EQ(sync_status.error_message(), "Hostname Verification Check failed.");
 }
 
+TEST(TlsCertificateVerifierTest,
+     NoOpCertificateVerifierSucceedsWithVerifiedRootCertSubject) {
+  grpc_tls_custom_verification_check_request request;
+  memset(&request, 0, sizeof(request));
```
matthewstevenson88, line 167: nit: Why is this line necessary? I'm guessing this is copied from existing code, but I think this line is NO-OP.

Similarly below.

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,6 +161,85 @@ TEST(TlsCertificateVerifierTest,
...
+  char* expected_subject = const_cast<char*>(
+      "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU");
+  request.peer_info.verified_root_cert_subject = expected_subject;
+  auto verifier = std::make_shared<NoOpCertificateVerifier>();
+  TlsCustomVerificationCheckRequest cpp_request(&request);
+  EXPECT_EQ(cpp_request.verified_root_cert_subject(), expected_subject);
+  grpc::Status sync_status;
+  verifier->Verify(&cpp_request, nullptr, &sync_status);
```
matthewstevenson88, line 175: Should we add an expectation on the return value of the `verifier->Verify` call?

Similarly throughout.

test/core/security/tls_security_connector_test.cc
gtcooke94: It's a good point. Right now the test essentially just makes sure it doesn't crash when the property is present, but it doesn't do anything to verify the actual value.

The issue is, it's all inside of private variables - the `{Channel/Server}PendingVerifierRequest` are private, and all of this happens within `check_peer`. The request is created, placed in a map `pending_verifier_requests_`, then removed from that map when completed, all without external accessibility.

I'll keep thinking on how to improve this, but I am open to suggestions.

test/core/security/tls_security_connector_test.cc
gtcooke94: Changed to request and require verify

test/cpp/security/tls_certificate_verifier_test.cc
gtcooke94: The return on this indicates whether it executed synchronously or not - actually checking the status of the verification is done in the line below checking `sync_status`.
Given this is a sync test, I've gone on and added a check that it returns `True`.

test/cpp/security/tls_certificate_verifier_test.cc
gtcooke94: Interestingly, the unset test`NoOpCertificateVerifierSucceedsWithoutVerifiedRootCertSubject` fails without this line.  I presume this clears the memory, and perhaps without it we are accessing garbage?

test/core/security/tls_security_connector_test.cc
gtcooke94: Updated the new tests to std::strings and EXPECT_EQ

test/core/security/tls_security_connector_test.cc
matthewstevenson88: Could we leverage the VerifiedRootCertSubjectVerifier to do this?

(Apologies again if I'm missing something obvious.)

gtcooke94: That's a test util added in `test/cpp` and uses the cpp API. This test we're looking at is a level lower in `test/core` in the `security_connector`

Unfortunately that higher level cpp test doesn't cover either - we want to make sure that, given a `tsi_peer`, the `security_connector` level code can construct a `grpc_tls_custom_verification_check_request` that properly pulls the information from the `tsi_peer`.

The cpp level test that uses the `VerifiedRootCertSubjectVerifier` starts by constructing a `grpc_tls_custom_verification_check_request` itself, so we still doesn't have coverage of the `tsi_peer` -> `grpc_tls_custom_verification_check_request`.

I could create a public test-only wrapper function for `PendingVerifierRequestInit`, and I think that would help us test this? Thoughts?

rockspore: Would it be more convenient to test this in an end2end test?

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,6 +161,86 @@ TEST(TlsCertificateVerifierTest,
...
+  grpc::Status sync_status;
+  bool is_sync = verifier->Verify(&cpp_request, nullptr, &sync_status);
+  EXPECT_TRUE(is_sync);
+  EXPECT_TRUE(sync_status.ok())
+      << sync_status.error_code() << " " << sync_status.error_message();
+}
+
+TEST(TlsCertificateVerifierTest, VerifiedRootCertSubjectVerifierFailsNull) {
+  grpc_tls_custom_verification_check_request request;
+  memset(&request, 0, sizeof(request));
```
matthewstevenson88, line 212: nit: I think this line is redundant and can be removed.

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,6 +161,86 @@ TEST(TlsCertificateVerifierTest,
...
+  grpc::Status sync_status;
+  verifier->Verify(&cpp_request, nullptr, &sync_status);
+  EXPECT_EQ(sync_status.error_code(), grpc::StatusCode::UNAUTHENTICATED);
+  EXPECT_EQ(sync_status.error_message(),
+            "VerifiedRootCertSubjectVerifier failed");
+}
+
+TEST(TlsCertificateVerifierTest, VerifiedRootCertSubjectVerifierFailsMismatch) {
+  grpc_tls_custom_verification_check_request request;
+  memset(&request, 0, sizeof(request));
```
matthewstevenson88, line 229: nit: I think this line is redundant and can be removed.

test/core/security/tls_security_connector_test.cc
matthewstevenson88: +1, good call Luwei. Let's modify the e2e tests for custom verification to check this.

gtcooke94: Added an E2E test, PTAL

I named the custom verifier `TestExternalVerifier` instead of something specific to root cert subject. I did this so that we have a modifiable custom verifier for checking any property if it is something we want to expand in the future.

test/core/end2end/fixtures/h2_tls_common.h
```diff
@@ -51,7 +51,8 @@ struct SecurityPrimitives {
   enum VerifierType {
     EXTERNAL_SYNC_VERIFIER = 0,
     EXTERNAL_ASYNC_VERIFIER = 1,
-    HOSTNAME_VERIFIER = 2
+    HOSTNAME_VERIFIER = 2,
+    TEST_VERIFIER = 3,
```
rockspore, line 55: I'd prefer this verifier name being more descriptive of what it does, e.g. ROOT_SUBJECT_VERIFIER or so. It's test code so it can always be renamed if needed in the future.

test/core/end2end/fixtures/h2_tls_test_external_verifier.cc
```diff
@@ -0,0 +1,70 @@
...
+  SetCertificateVerifier(ffd, SecurityPrimitives::VerifierType::TEST_VERIFIER);
+  f.fixture_data = ffd;
+  f.cq = grpc_completion_queue_create_for_next(nullptr);
+  return f;
+}
+
+static grpc_end2end_test_config config = {
+    // client: static data provider + sync external verifier
+    // server: static data provider + sync external verifier
+    // extra: TLS 1.2
```
rockspore, line 52: nit: The version is set to 1.3 here.

test/core/end2end/fixtures/h2_tls_common.h
gtcooke94: Not sure if you saw my comment in the other chain, copying here:
> I named the custom verifier TestExternalVerifier instead of something specific to root cert subject. I did this so that we have a modifiable custom verifier for checking any property if it is something we want to expand in the future.

Really, there's nothing that tests all of these properties yet, so this test verifier should get expanded to test all the different properties that we populate from `tsi_peer`

test/cpp/security/tls_certificate_verifier_test.cc
gtcooke94: Removed

test/core/end2end/fixtures/h2_tls_test_external_verifier.cc
gtcooke94: Fixed

test/cpp/security/tls_certificate_verifier_test.cc
matthewstevenson88: Once this PR is submitted, we should investigate this. I'm worried there's some order-dependence or shared state to these tests.

matthewstevenson88 (commented):
LGTM, modulo a few minor comments. I'll approve once they're resolved.

Thanks again for the PR! :)

test/core/end2end/fixtures/h2_tls_common.h
```diff
@@ -167,6 +168,19 @@ inline void SetCertificateVerifier(
           server_async_verifier->base());
       break;
     }
+    case SecurityPrimitives::VerifierType::TEST_VERIFIER: {
+      std::string expeected_root_cert_subject =
+          "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU";
```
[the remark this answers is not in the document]
gtcooke94: It comes from the `ca.pem` cert that is used in all of the tests, will add a comment

test/core/end2end/fixtures/h2_tls_common.h
gtcooke94: I like that, will update

test/core/end2end/fixtures/h2_tls_test_external_verifier.cc
```diff
@@ -0,0 +1,70 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//
+
+#include <string.h>
```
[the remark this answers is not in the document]
gtcooke94: Neither were actually needed, removed.

test/core/end2end/fixtures/h2_tls_common.h
```diff
@@ -167,6 +168,19 @@ inline void SetCertificateVerifier(
           server_async_verifier->base());
       break;
     }
+    case SecurityPrimitives::VerifierType::TEST_VERIFIER: {
+      std::string expeected_root_cert_subject =
```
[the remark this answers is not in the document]
gtcooke94: Fixed

matthewstevenson88 (commented):
There are some newly-failing tests, otherwise LGTM.

gtcooke94:
Looks like some of the CI was unhappy about removing the `<string>` and `string.h` import and it wants me to add them back. Doing that and letting the tests re-run

matthewstevenson88 (approved):

test/core/end2end/generate_tests.bzl
```diff
@@ -117,6 +117,7 @@ END2END_FIXTURES = {
     "h2_tls_static_async_tls1_3": _fixture_options(secure = True),
     "h2_tls_certwatch_sync_tls1_2": _fixture_options(secure = True),
     "h2_tls_certwatch_async_tls1_3": _fixture_options(secure = True),
+    "h2_tls_peer_property_external_verifier": _fixture_options(secure = True),
```
markdroth, line 120: Do we actually need to add a new test fixture for this?  Adding a test fixture is a pretty big multiplier for the set of test cases, because we basically run every core e2e test case in every fixture.  But it's not clear to me that this particular code is actually useful to run against every one of the e2e test cases -- it seems like we really need to test this functionality only once.

I suggest instead just adding a single one-off test to cover this case with one common config.  As an example of such a one-off e2e test, see:

https://github.com/grpc/grpc/blob/master/test/core/end2end/h2_ssl_session_reuse_test.cc

test/cpp/util/tls_test_utils.h
```diff
@@ -76,6 +76,25 @@ class AsyncCertificateVerifier
   std::deque<Request> queue_ ABSL_GUARDED_BY(mu_);
 };
 
+class VerifiedRootCertSubjectVerifier
+    : public grpc::experimental::ExternalCertificateVerifier {
+ public:
+  explicit VerifiedRootCertSubjectVerifier(const std::string& expected_subject)
+      : expected_subject_(expected_subject) {}
```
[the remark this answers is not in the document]
gtcooke94: When I add `std::move()` here I get a clang-tidy error `std::move of const variable has no effect`

include/grpcpp/security/tls_certificate_verifier.h
```diff
@@ -62,6 +62,15 @@ class TlsCustomVerificationCheckRequest {
   grpc::string_ref peer_cert() const;
   grpc::string_ref peer_cert_full_chain() const;
   grpc::string_ref common_name() const;
+  /**
```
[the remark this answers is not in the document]
gtcooke94: Fixed

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,6 +161,84 @@ TEST(TlsCertificateVerifierTest,
   EXPECT_EQ(sync_status.error_message(), "Hostname Verification Check failed.");
 }
 
+TEST(TlsCertificateVerifierTest,
```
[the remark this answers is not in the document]
gtcooke94: Good point, I wrote those before writing the new verifier. I think the case where it is unset (`NoOpCertificateVerifierSucceedsWithoutVerifiedRootCertSubject`) is a different code path and worth keeping?

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,6 +161,84 @@ TEST(TlsCertificateVerifierTest,
...
+  grpc::Status sync_status;
+  bool is_sync = verifier->Verify(&cpp_request, nullptr, &sync_status);
+  EXPECT_TRUE(is_sync);
+  EXPECT_TRUE(sync_status.ok())
+      << sync_status.error_code() << " " << sync_status.error_message();
+}
+
+TEST(TlsCertificateVerifierTest, VerifiedRootCertSubjectVerifierSucceeds) {
+  grpc_tls_custom_verification_check_request request;
+  std::string expected_subject =
```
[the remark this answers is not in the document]
gtcooke94: I original had these as `char*`, similar to the other tests in this file, but was suggested to use `std::string` to be more modern. I'll change it to `constexpr char[]`, but I want to make sure I understand why this is preferred here?

test/core/end2end/generate_tests.bzl
gtcooke94: I like the single one-off test better. I wasn't sure if all of the end2end tests for `core` had to be done with the fixture setup. I'll change this out for a one-off

gtcooke94: Changed, PTAL

test/cpp/util/tls_test_utils.h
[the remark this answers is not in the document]
markdroth: Option 1: The way this is currently written:
- If the caller has a `std::string` object, then we'll get a ref to that object passed into this ctor, and we'll make a copy of the `std::string` for the new object.
- If the caller has a `char*` instead of a `std::string`, then the compiler will construct a temporary `std::string` object from the `char*` in the caller, so that it has an object that it can pass down a reference to, then we'll make a copy of that `std::string` for the new object, and then the compiler will immediately destroy the temporary `std::string` it created in the caller.  So we'll create two `std::string` objects and then immediately delete one of them.
- If the caller has an `absl::string_view`, the caller will need to explicitly convert it to a `std::string`.  The result in this case is basically the same as the `char*` case, but it's a little more annoying to write, because there is no explicit conversion from `absl::string_view` to `std::string`.

Option 2: If you change the parameter to `std::string` and use `std::move()` here:
- If the caller has a `std::string` object, then a copy will be made to pass the string in by value, and then we'll move that copy into place in the new object, so we're just creating a single `std::string`.
- If the caller has a `char*`, the compiler will construct a new `std::string` object from the `char*`, it will pass that in by value, and we'll move it into place in the new object, so we're just creating a single `std::string`.
- If the caller has an `absl::string_view`, the caller will need to explicitly convert it to `std::string`.  The result in this case is basically the same as the `char*` case, but it's a little more annoying to write, because there is no explicit conversion from `absl::string_view` to `std::string`.
- If the caller happens to have a `std::string` that they don't need to hold on to, then the caller can use `std::move()` to pass in the parameter, in which case we're not creating *any* new `std::string` objects.  (This case isn't necessarily that common, but it can be a useful optimization at times.)

Option 3: Another option is to use `absl::string_view` as the parameter type, in which case:
- Regardless of whether the caller has a `std::string`, `char*`, or `absl::string_view`, we will construct a new `absl::string_view` to pass in by value (which is much cheaper than constructing a `std::string`).  We will then create a new `std::string` from the `absl::string_view` for this object, so we're creating just one `std::string`.

In general, `absl::string_view` is probably the right thing to use for any C++ code that takes a string as input.  But using `std::string` and then using `std::move()` to store it does allow optimizing for the occasional case where the caller already has a `std::string` that they no longer need.  So either of those options seems like a reasonable choice here.

To be clear, the performance does not matter at all in this particular case, but I find that it's generally a good idea to consider the performance implications of the various ways of writing this sort of thing, so that you're in good habits when you do encounter a case where it really matters. :)

test/core/end2end/h2_tls_peer_property_external_verifier_test.cc
```diff
@@ -0,0 +1,345 @@
...
+  const char* server_cert =
+      reinterpret_cast<const char*> GRPC_SLICE_START_PTR(cert_slice);
+  const char* server_key =
+      reinterpret_cast<const char*> GRPC_SLICE_START_PTR(key_slice);
+
+  grpc_tls_credentials_options* options = grpc_tls_credentials_options_create();
+  // Set credential provider.
+  grpc_tls_identity_pairs* server_pairs = grpc_tls_identity_pairs_create();
+  grpc_tls_identity_pairs_add_pair(server_pairs, server_key, server_cert);
+  auto* server_provider =
```
rockspore, line 83: The provider needs to be released eventually to void memory leak, which should be why asan is failing.. So you probably need a separate function to retain the pointers. Same for the client.

test/core/end2end/h2_tls_peer_property_external_verifier_test.cc
```diff
@@ -0,0 +1,345 @@
...
+
+  std::string server_addr = grpc_core::JoinHostPort("localhost", port);
+
+  grpc_completion_queue* cq = grpc_completion_queue_create_for_next(nullptr);
+  grpc_ssl_session_cache* cache = grpc_ssl_session_cache_create_lru(16);
+
+  grpc_server* server = server_create(cq, server_addr.c_str());
+
+  do_round_trip(cq, server, server_addr.c_str(), cache, false);
+  //   do_round_trip(cq, server, server_addr.c_str(), cache, true);
```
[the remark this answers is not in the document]
gtcooke94: Fixed

test/cpp/security/tls_certificate_verifier_test.cc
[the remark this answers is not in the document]
gtcooke94: Good idea, will change

test/cpp/security/tls_certificate_verifier_test.cc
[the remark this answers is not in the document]
gtcooke94: Thank you for the explanation - fixed here to your suggestion.

I'll note that change in `grpc_tls_custom_verification_check_request` as something for us to do as well

test/cpp/util/tls_test_utils.h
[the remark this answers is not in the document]
gtcooke94: Thanks for the thorough explanation! I've swapped the parameter to an `absl::string_view`

test/core/end2end/h2_tls_peer_property_external_verifier_test.cc
gtcooke94: Just pushed changes to deal with this, good catch.

Both the provider and the verifier needed to be managed. I added double pointer args to the functions so I could retain and free them at the end.

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -161,28 +161,15 @@ TEST(TlsCertificateVerifierTest,
...
-  std::string expected_subject =
+  constexpr char kExpectedSubject[] =
       "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU";
-  request.peer_info.verified_root_cert_subject = expected_subject.c_str();
+  request.peer_info.verified_root_cert_subject =
+      const_cast<char*>(kExpectedSubject);
   auto verifier =
       ExternalCertificateVerifier::Create<VerifiedRootCertSubjectVerifier>(
-          expected_subject);
+          const_cast<char*>(kExpectedSubject));
```
markdroth, line 172: The `const_cast<>` should not be needed here.  It's fine to create an `absl::string_view` from a non-const `char*`.

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -193,13 +180,14 @@ TEST(TlsCertificateVerifierTest, VerifiedRootCertSubjectVerifierSucceeds) {
...
   grpc_tls_custom_verification_check_request request;
-  char* expected_subject = const_cast<char*>(
-      "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU");
+  constexpr char kExpectedSubject[] =
+      "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU";
   request.peer_info.verified_root_cert_subject = nullptr;
   auto verifier =
       ExternalCertificateVerifier::Create<VerifiedRootCertSubjectVerifier>(
-          expected_subject);
+          const_cast<char*>(kExpectedSubject));
```
markdroth, line 188: Same here: `const_cast<>` should not be needed.

test/cpp/security/tls_certificate_verifier_test.cc
```diff
@@ -209,12 +197,12 @@ TEST(TlsCertificateVerifierTest, VerifiedRootCertSubjectVerifierFailsNull) {
...
   grpc_tls_custom_verification_check_request request;
-  char* expected_subject = const_cast<char*>(
-      "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU");
+  constexpr char kExpectedSubject[] =
+      "CN=testca,O=Internet Widgits Pty Ltd,ST=Some-State,C=AU";
   request.peer_info.verified_root_cert_subject = "BAD_SUBJECT";
   auto verifier =
       ExternalCertificateVerifier::Create<VerifiedRootCertSubjectVerifier>(
-          expected_subject);
+          const_cast<char*>(kExpectedSubject));
```
markdroth, line 205: Same here: `const_cast<>` should not be needed.

markdroth (approved):
Only a few minor nits remaining, feel free to merge once they're addressed.

test/cpp/security/tls_certificate_verifier_test.cc
gtcooke94: Removed them

test/cpp/security/tls_certificate_verifier_test.cc
gtcooke94: Removed them

test/cpp/security/tls_certificate_verifier_test.cc
gtcooke94: Removed them

PR #32340, opened 2023-02-10
Add missing dependency for tsi_alts_credentials.

matthewstevenson88:
While creating an internal CL that depends directly on tsi_alts_credentials, I was getting linker errors saying ` error: backward reference detected: grpc_channel_credentials_release`, because `alts_tsi_handshaker.cc` uses the `grpc_channel_credentials_release` API, which is defined in the `grpc_security_base` target.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -3130,6 +3130,7 @@ grpc_cc_library(
         "alts_util",
         "gpr",
         "grpc_base",
+        "grpc_security_base",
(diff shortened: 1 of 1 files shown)


ctiller (approved):
this is a bit weird... would like to know how we were getting access to the bad api (which include path) so we can factor our build a little bit better and avoid this kind of thing

PR #32350, opened 2023-02-10
Disable testKeepaliveWithV2API test case for objc interop test

yijiem:
This test has a race and is flaky, see [b/268379869](https://b.corp.google.com/issues/268379869). It sends an RPC to the interop server with 0s `keepaliveTimeout` and expects the keepalive watchdog timer to fire immediately before the server acks the ping.

### DIFF
--- a/src/objective-c/tests/InteropTests/InteropTests.m
+++ b/src/objective-c/tests/InteropTests/InteropTests.m
@@ -1612,48 +1612,48 @@ static dispatch_once_t initGlobalInterceptorFactory;
     }
 
     /* DISABLES CODE */ (
-      RMTTestService *service = [RMTTestService serviceWithHost:[[self class] host]];
-      if ([[self class] transport] == gGRPCCoreCronetID) {
-      // Cronet does not support keepalive
+    RMTTestService *service = [RMTTestService serviceWithHost:[[self class] host]];
+    if ([[self class] transport] == gGRPCCoreCronetID) {
+    // Cronet does not support keepalive
+    return;
+    }
+    __weak XCTestExpectation *expectation = [self expectationWithDescription:@"Keepalive"];
+
+    const NSTimeInterval kTestTimeout = 5;
+    NSNumber *kRequestSize = @27182;
+    NSNumber *kResponseSize = @31415;
+
+    id request = [RMTStreamingOutputCallRequest messageWithPayloadSize:kRequestSize
+                                                requestedResponseSize:kResponseSize];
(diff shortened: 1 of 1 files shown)


src/objective-c/tests/InteropTests/InteropTests.m
```diff
@@ -1603,7 +1603,13 @@ - (void)testKeepaliveWithV2API {
   GRPCTestRunWithFlakeRepeats(self, ^(GRPCTestWaiter waiterBlock, GRPCTestAssert assertBlock) {
     // The test is highly flaky when ran as part of InteropTestsRemote
     // TODO(jtattermusch): fix and re-enable the test.
-    if ([[self class] isRemoteTest]) {
+    // if ([[self class] isRemoteTest]) {
+    //   return;
+    // }
+    // This test has a race and is flaky in any configurations. See b/268379869.
```
drfloob, line 1609: You can write `// TODO(b/268379869) This test ...` here, there's some tooling available around comments with that style.

src/objective-c/tests/InteropTests/InteropTests.m
```diff
@@ -1603,7 +1603,13 @@ - (void)testKeepaliveWithV2API {
   GRPCTestRunWithFlakeRepeats(self, ^(GRPCTestWaiter waiterBlock, GRPCTestAssert assertBlock) {
     // The test is highly flaky when ran as part of InteropTestsRemote
     // TODO(jtattermusch): fix and re-enable the test.
-    if ([[self class] isRemoteTest]) {
+    // if ([[self class] isRemoteTest]) {
```
drfloob, line 1606: Instead of leaving code commented out (it's a smell, since the code could get out of sync if it's not being compiled) ... I'd recommend moving the TODO up to the top of the function, and having an unconditional `return`.

drfloob (approved):
LGTM with nits, presuming @sampajano is willing to disable this test. Please wait for his review as well.

sampajano (approved):
Thanks for the change, @yijiem and thanks for the review, @drfloob!

This LGTM in particular based on the fact that this only tests a 0 timeout case (as @yijiem has pointed out).

We should still hopefully work on restoring it soon. (Will be tracked by the bug.)

Thanks!

src/objective-c/tests/InteropTests/InteropTests.m
yijiem: Done.

src/objective-c/tests/InteropTests/InteropTests.m
yijiem: Done. Good point.

PR #32363, opened 2023-02-13
[EventEngine] Fix the shims for iOS

drfloob:
The shims assumed that all platforms with Posix socket support would use the PosixEventEngine. This is not the case for iOS.


sampajano (approved):
@drfloob Thanks for making this change for iOS!

PR #32364, opened 2023-02-13
web: make HTTP field names case-insensitive

akshayjshah:
I'm opening this PR to (hopefully!) stimulate a discussion. In brief, I'd like to amend the gRPC-Web protocol docs to encourage implementations to follow HTTP semantics and compare HTTP field names case-insensitively.

The gRPC-Web specification is a nicely-designed way for proxies to expose standard HTTP/2 gRPC servers to clients using less tightly-controlled HTTP stacks, such as web browsers. To serve that goal, it seems valuable to have the gRPC-Web specification follow [RFC 9110 (HTTP Semantics)](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names). Like previous RFCs, 9110 specifies that "field names are case-insensitive." However, the current gRPC-Web specification requires that servers and proxies "use lower-case header/trailer names" on the wire. In principle, mandating casing on the wire is normal for HTTP/2 and fine (if unusual) for HTTP/1.1; however, it encourages implementations to violate HTTP semantics and require lower-case names when _reading_ headers and trailers.

I'd like to loosen the gRPC-Web specification to permit any casing on the wire for HTTP/1.1. I'd also like to emphasize that gRPC-Web implementations ought to follow standard HTTP semantics when _reading_ fields and compare names case-insensitively.  Implementations that can't treat names case-insensitively without breaking backward compatibility should instead normalize field names to lowercase. Among the Google-maintained gRPC implementations, at least `grpc-go` and `grpc-java` already compare names case-insensitively (even though they're HTTP/2-only). `grpc-dart` does the opposite and compares names case-sensitively. `grpc-web` is sometimes case-insensitive (when reading `grpc-status` and `grpc-message` from trailers-only responses) and sometimes case-sensitive (when hand-parsing a block of length-prefixed trailers).

The proposed amendment does not affect the correctness of Envoy (which may continue to use lower-case field names). It partially affects `grpc-web`, which would require a small patch to always normalize names. (Both patched and unpatched versions of `grpc-web` would work with Envoy.) `grpc-dart` would need to either begin treating field names case-insensitively or normalize names, depending on what's possible in Dart without breaking backward compatibility.

Relates to https://github.com/improbable-eng/grpc-web/issues/228, https://github.com/bufbuild/connect-go/issues/453/, and https://github.com/grpc/grpc-dart/issues/594.

### DIFF
--- a/doc/PROTOCOL-WEB.md
+++ b/doc/PROTOCOL-WEB.md
@@ -48,14 +48,9 @@ Content-Type
 HTTP wire protocols
 
 1. support any HTTP/*, with no dependency on HTTP/2 specific framing
-2. send lower-case header/trailer names over HTTP/2 (following [RFC
-   9113](https://www.rfc-editor.org/rfc/rfc9113#section-8.2-2)) and send
-   header/trailer names using any case over HTTP/1.1 (following [RFC
-   9112](https://www.rfc-editor.org/rfc/rfc9112.html#section-5-1))
-3. compare header/trailer names case-insensitively (following [RFC
-   9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1-3)) where
-   possible, and normalize header/trailer names to lower-case otherwise
-4. use EOF (end of body) to close the stream
+2. header names may be upper- or mixed-case over HTTP/1.1, but trailers encoded in
+   the last length-prefixed message must always use lower-case names.
+3. use EOF (end of body) to close the stream
 
 ---
 
--- a/doc/PROTOCOL-WEB.md
+++ b/doc/PROTOCOL-WEB.md
@@ -48,8 +48,14 @@ Content-Type
 HTTP wire protocols
 
 1. support any HTTP/*, with no dependency on HTTP/2 specific framing
-2. use lower-case header/trailer names
-3. use EOF (end of body) to close the stream
+2. send lower-case header/trailer names over HTTP/2 (following [RFC
+   9113](https://www.rfc-editor.org/rfc/rfc9113#section-8.2-2)) and send
+   header/trailer names using any case over HTTP/1.1 (following [RFC
+   9112](https://www.rfc-editor.org/rfc/rfc9112.html#section-5-1))
+3. compare header/trailer names case-insensitively (following [RFC
+   9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1-3)) where
+   possible, and normalize header/trailer names to lower-case otherwise
+4. use EOF (end of body) to close the stream
 
 ---
 


akshayjshah:
@wenbozhu I've amended the diff here - it's hopefully a bit clearer now. Appreciate you taking a look!

The mergeability check is currently failing because of missing labels. I'm not sure which of the labels is appropriate here, or if this change should be included in the changelog. Happy to take your advice :)

akshayjshah:
@wenbozhu Any thoughts on the contents of this PR?

doc/PROTOCOL-WEB.md
```diff
@@ -48,8 +48,14 @@ Content-Type
 HTTP wire protocols
 
 1. support any HTTP/*, with no dependency on HTTP/2 specific framing
-2. use lower-case header/trailer names
```
wenbozhu, line 51: How about 

2. header names may not be lower-case over HTTP/1.1; and trailers are not supported

wenbozhu (commented):
Thanks for the suggestion.

doc/PROTOCOL-WEB.md
akshayjshah: I believe the current language is referring to the block of trailers encoded in the last length-prefixed message in the stream, not HTTP trailers. Those trailers aren't visible to any generic HTTP systems (proxies, middleware, etc.) so I think it's fine to mandate lower-case field names on the wire.

How about this?

> HTTP header and trailer names may not be lower-case over HTTP/1.1; trailers encoded in the final length-prefixed message of the stream must use lower-case names

akshayjshah:
@wenbozhu Thanks for the feedback - I've updated the PR!

wenbozhu (approved):

akshayjshah:
:tada: Thanks for the stamp! What do I need to do to get CI passing and merge this in?

sampajano:
@akshayjshah Thanks for the proposal and the change! I wonder if would provide a patch for https://github.com/grpc/grpc-web too, if it's within your interest? If not, we can do a follow up change there. Thanks! :)

akshayjshah:
@sampajano Unfortunately, I'm probably not the right person to make a patch to any JS project 🤦🏽‍♂️

sampajano:
@akshayjshah aha ok! perfectly fine! Thanks for considering! :)

I'll make a follow-up change once this change is submitted then! thanks!

akshayjshah:
@wenbozhu or @sampajano - is there anything else I need to do to get this merged in? I'm not sure how to add the appropriate label to get the `Mergeable` CI check passing, and none of the other checks have run.

sampajano:
> @wenbozhu or @sampajano - is there anything else I need to do to get this merged in? I'm not sure how to add the appropriate label to get the `Mergeable` CI check passing, and none of the other checks have run.

@akshayjshah Hi! I've added the missing label so i think the "Mergeable" CI should pass.

I'd wait for @wenbozhu for approval since there's a comment from him. I'll ping the thread too!

doc/PROTOCOL-WEB.md
sampajano: @wenbozhu Could you take a look and approve the PR if the updated text looks good? Thanks!

akshayjshah:
Thanks for merging this in, @sampajano! 😻

PR #32365, opened 2023-02-13
Minimum time file watcher

gtcooke94:
Enforce a minimum value for the `refresh_interval_sec_` for the `FileWatcherCertificateProvider`. There have been issues found when this is set to 0, and the security team discussed and agreed that 0 should not be a valid value for this use-case.

I made the `refresh_interval_sec_` public to make it easy to test - I didn't immediately see an easy way around this. I found `FRIEND_TEST` exists for accessing private members, but I didn't see that used anywhere in grpc. If there is a better solution to this, please let me know.

### DIFF
--- a/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
+++ b/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
@@ -32,7 +32,6 @@
 #include <openssl/x509.h>
 
 #include "absl/status/status.h"
-#include "absl/strings/str_format.h"
 
 #include <grpc/slice.h>
 #include <grpc/support/log.h>
--- a/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
+++ b/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
@@ -392,7 +392,7 @@ FileWatcherCertificateProvider::ReadIdentityKeyCertPairFromFiles(
   return absl::nullopt;
 }
 
-const int64_t FileWatcherCertificateProvider::TestOnlyGetRefreshIntervalSecond()
+int64_t FileWatcherCertificateProvider::TestOnlyGetRefreshIntervalSecond()
     const {
   return refresh_interval_sec_;
 }
--- a/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
+++ b/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
@@ -117,7 +117,7 @@ gpr_timespec TimeoutSecondsToDeadline(int64_t seconds) {
 
 }  // namespace
 
-static const int64_t MINIMUM_FILE_WATCHER_REFRESH_INTERVAL_SECONDS = 1;
+static constexpr int64_t kMinimumFileWatcherRefreshIntervalSeconds = 1;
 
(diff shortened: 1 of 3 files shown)


gtcooke94:
@rockspore

src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h
```diff
@@ -150,6 +150,7 @@ class FileWatcherCertificateProvider final
   }
 
   UniqueTypeName type() const override;
+  int64_t refresh_interval_sec_ = 0;
```
rockspore, line 153: If we really have to expose it (indeed I don't know of any other good way to assert this), I think we should have a getter method instead of exposing the raw field.

src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.cc
```diff
@@ -117,14 +117,26 @@ gpr_timespec TimeoutSecondsToDeadline(int64_t seconds) {
 
 }  // namespace
 
+static const int64_t MINIMUM_FILE_WATCHER_REFRESH_INTERVAL_SECONDS = 1;
```
rockspore, line 120: Is this how constants are defined elsewhere? I think people usually use `kVar` for constants. Capital letters with underscore are usually macros.

rockspore: I also feel like `static` is not necessary here. `constexpr` usually suffices.

gtcooke94: I'm not sure exactly what the style is, I'll dig around some more code

I saw these - https://source.corp.google.com/github/grpc/grpc/src/core/tsi/ssl_transport_security.cc;rcl=1d8fac36b3914fc17296158d4cb90b3fe51f4654;l=68 but those _are_ macros so I suppose I shouldn't be re-using that naming pattern

gtcooke94: Changed to `static constexpr kVar`

src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h
gtcooke94: Swapped to an accessor - `TestOnlyGetRefreshIntervalSecond`

src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h
```diff
@@ -151,6 +151,8 @@ class FileWatcherCertificateProvider final
 
   UniqueTypeName type() const override;
 
+  int64_t TestOnlyGetRefreshIntervalSecond() const;
```
rockspore, line 154: Most getters in the repo follow the snake style. Also this member function is not exposed as a public API since it's not in any include headers. Following the distributor() style above, do you think refresh_interval_seconds() would make more sense?

gtcooke94: Discussed style offline with other grpc folks, we'll keep it this way at their suggestion

rockspore (approved):

markdroth (approved):

PR #32376, opened 2023-02-14
[EventEngine] Add more granular trace flags

drfloob:
The set of trace flags is now:
* event_engine
* event_engine_endpoint
* event_engine_endpoint_data: additionally log all sent/received data, similar to what the shims do.
* event_engine_poller

### DIFF
--- a/src/core/lib/event_engine/trace.cc
+++ b/src/core/lib/event_engine/trace.cc
@@ -16,3 +16,9 @@
 #include "src/core/lib/debug/trace.h"
 
 grpc_core::TraceFlag grpc_event_engine_trace(false, "event_engine");
+grpc_core::TraceFlag grpc_event_engine_endpoint_trace(false,
+                                                      "event_engine_endpoint");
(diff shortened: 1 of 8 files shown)


ctiller:
Not for this PR, but to put it out there:
What do we feel about hierarchical trace flags?

```
GRPC_TRACE=event_engine/endpoint
GRPC_TRACE=event_engine/{endpoint,poller}
GRPC_TRACE=event_engine/*
```

drfloob:
> What do we feel about hierarchical trace flags?

I'd use it here for sure. I think there are other small (<= 3 sub-traces) potential uses with call, client_channel, subchannel. XDS also has quite a few, but I'm not sure how valuable it would be to enable the bulk of them at once.

Regarding the change, the trace flags are sort of public API, right? Changing existing flags could upset some folks who rely on them already. If we did any flag restructuring, we may want to ensure the old flags are still supported.

Vignesh2208:
Thanks for creating this AJ. LGTM

Vignesh2208 (approved):

PR #32421, opened 2023-02-17
Skip TSAN run for bm_fullstack_unary_ping_pong

yijiem:
`//test/cpp/microbenchmarks:bm_fullstack_unary_ping_pong@poller=epoll1` seems to be OOMKilled sometimes with exit code 137: 

image

We could try disabling TSAN run for it if we think it's okay to do so, since:
> Typical memory overhead introduced by ThreadSanitizer is about 5x-10x (https://clang.llvm.org/docs/ThreadSanitizer.html)

### DIFF
--- a/test/cpp/microbenchmarks/BUILD
+++ b/test/cpp/microbenchmarks/BUILD
@@ -315,6 +315,7 @@ grpc_cc_test(
     tags = [
         "no_mac",  # to emulate "excluded_poll_engines: poll"
         "no_windows",
+        "notsan",
     ],
     deps = [":fullstack_unary_ping_pong_h"],
 )


ctiller:
This benchmark should run for 1/10 of a millisecond for each test - so usually just a handful of iterations.

Do we know how it's OOMing so quickly?

yijiem:
> This benchmark should run for 1/10 of a millisecond for each test - so usually just a handful of iterations.
> 
> Do we know how it's OOMing so quickly?

Good point. I haven't verified (this is just a speculation). I will try profile it locally and report back.

yijiem:
> > This benchmark should run for 1/10 of a millisecond for each test - so usually just a handful of iterations.
> > Do we know how it's OOMing so quickly?
> 
> Good point. I haven't verified (this is just a speculation). I will try profile it locally and report back.

I'm not sure what killed the test... and for what reason. In my local run, I observed the test's peak resident memory usage was ~`2500MB` for a `--config=tsan` and `--config=dbg` build and was ~`900MB` for a `--config=opt` build.

The test is configured to have a [large size](https://github.com/grpc/grpc/blob/master/test/cpp/microbenchmarks/BUILD#L310). According to bazel, it should use 300MB RAM: https://bazel.build/reference/be/common-definitions#common-attributes-tests. But I'm not sure if bazel will actually kill the test if it runs passed that limit.

ctiller:
How did you run it? With `bazel run` or with `bazel-bin/.../bm_fullstack_unary_ping_pong` directly?

yijiem:
> How did you run it? With `bazel run` or with `bazel-bin/.../bm_fullstack_unary_ping_pong` directly?

Just tried `$ bazel run --config=tsan --config=dbg --nocache_test_results //test/cpp/microbenchmarks:bm_fullstack_unary_ping_pong@poller=epoll1` locally a few times and seems like bazel is killing the test consistently for some reason.. Seems like the SIGKILL signal is sent from `test-setup.sh` as:
```

PR #32429, opened 2023-02-17
xDS fault injection e2e test: fix flakes caused by processing queued calls in parallel

markdroth:
The `XdsFaultInjectionMaxFault` test has seen a few flakes since #32326 was merged.  I believe the flakiness is caused by the fact that when a large number of RPCs are queued up before the resolver result comes in, those RPCs are now re-processed in parallel instead of sequentially, which can cause us to delay more RPCs than we should due to the `max_faults` setting.  To fix this, we change the test to ensure that the channel is connected (i.e., the resolver result has already been returned) before we start sending a large number of concurrent RPCs.

Although this is the only test that I've seen flakes in, I've made this same change consistently to all fault injection tests that are creating a large number of concurrent RPCs, since the same flake could affect any of them.


yijiem (approved):
Thanks Mark!

PR #32438, opened 2023-02-21
Move trans stream receiver callbacks out of `mu_`

rock1246:
[Binder Tansport] This prevents deadlock against wire writer issues.

Currently, there are some `transport_stream_receiver_` callbacks triggered by NDK binder may acquire `WireReaderImpl::mu_` first then `WireWriterImpl::write_mu_`. We don't like see this.

We have this problem since some client and server are in the same process. The behavior of NDK binder seems more aggressive when the Tx and Rx are in the same process.

### DIFF
--- a/src/core/ext/transport/binder/wire_format/wire_reader_impl.cc
+++ b/src/core/ext/transport/binder/wire_format/wire_reader_impl.cc
@@ -24,6 +24,7 @@
 #include <utility>
 #include <vector>
 
+#include "absl/functional/any_invocable.h"
 #include "absl/memory/memory.h"
 #include "absl/status/statusor.h"
 
@@ -248,40 +249,24 @@ absl::Status WireReaderImpl::ProcessStreamingTransaction(
     transaction_code_t code, ReadableParcel* parcel) {
   bool need_to_send_ack = false;
   int64_t num_bytes = 0;
+  // Indicates which callbacks should be cancelled. It will be initialized as
+  // the flags the in-coming transaction carries, and when a particular
+  // callback is completed, the corresponding bit in cancellation_flag will be
+  // set to 0 so that we won't cancel it afterward.
+  int cancellation_flags = 0;
+  // The queue saves the actions needed to be done "WITHOUT" `mu_`.
+  // It prevents deadlock against wire writer issues.
+  std::queue<absl::AnyInvocable<void() &&>> deferred_func_queue;
   absl::Status tx_process_result;
+
   {
     grpc_core::MutexLock lock(&mu_);
     if (!connected_) {
       return absl::InvalidArgumentError("Transports not connected yet");
     }
 
-    // Indicate which callbacks should be cancelled. It will be initialized as
-    // the flags the in-coming transaction carries, and when a particular
-    // callback is completed, the corresponding bit in cancellation_flag will be
-    // set to 0 so that we won't cancel it afterward.
-    int cancellation_flags = 0;
-    tx_process_result =
-        ProcessStreamingTransactionImpl(code, parcel, &cancellation_flags);
-    if (!tx_process_result.ok()) {
-      gpr_log(GPR_ERROR, "Failed to process streaming transaction: %s",
-              tx_process_result.ToString().c_str());
-      // Something went wrong when receiving transaction. Cancel failed
-      // requests.
-      if (cancellation_flags & kFlagPrefix) {
-        gpr_log(GPR_INFO, "cancelling initial metadata");
-        transport_stream_receiver_->NotifyRecvInitialMetadata(
-            code, tx_process_result);
-      }
-      if (cancellation_flags & kFlagMessageData) {
-        gpr_log(GPR_INFO, "cancelling message data");
-        transport_stream_receiver_->NotifyRecvMessage(code, tx_process_result);
-      }
-      if (cancellation_flags & kFlagSuffix) {
-        gpr_log(GPR_INFO, "cancelling trailing metadata");
-        transport_stream_receiver_->NotifyRecvTrailingMetadata(
-            code, tx_process_result, 0);
-      }
-    }
+    tx_process_result = ProcessStreamingTransactionImpl(
+        code, parcel, &cancellation_flags, deferred_func_queue);
     if ((num_incoming_bytes_ - num_acknowledged_bytes_) >=
         kFlowControlAckBytes) {
       need_to_send_ack = true;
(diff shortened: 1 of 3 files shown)


sifmelcara:
The change looks good to me

Have you test the change internally to make sure we don't accidentally break other use cases (e.g. use BinderTransport as gRPC server)? What's the plan to make sure this patch fixes the bug? Maybe stress test in real use case for a few hours? (please ping me in the chat room instead of replying here publicly)

Some thoughts (no need to implement them right now but it would be good to have to prevent this kind of issue in the future):

1. We should somehow enforce that we don't run combiner (i.e. destructing last `grpc::ExecCtx` in the stack) while holding any binder transport mutex, as it will make it harder to reason about thread-safety and potential deadlock situations

2. We should probably move `grpc_core::ExecCtx`s in our code (e.g. https://github.com/grpc/grpc/blob/d43e35738ad9c624bd9eb2f9c56b93fb4cf849a7/src/core/ext/transport/binder/transport/binder_transport.cc#L491 and https://github.com/grpc/grpc/blob/d43e35738ad9c624bd9eb2f9c56b93fb4cf849a7/src/core/ext/transport/binder/wire_format/wire_writer.cc#L289
) to somewhere that is closer to entrypoint of our code. The entrypoint of NdkBinder callback is probably a good place https://github.com/grpc/grpc/blob/d43e35738ad9c624bd9eb2f9c56b93fb4cf849a7/src/core/ext/transport/binder/wire_format/binder_android.cc#L62

This would make it easier for us make sure we comply with (1). IIRC there are some assertions in place to make sure combiner API is only used while there are `grpc_core::ExecCtx` in the stack, if not, we need to replace those `grpc_core::ExecCtx` declarations with checks that verify we already have `grpc_core::ExecCtx` in the callstack, to ensure the tasks will be run.

3. Need to document the intended acquisition order somewhere (or.. refactor the code to avoid mutex at all) (  `WireReaderImpl::mu_` should not be acquired after `WireWriterImpl::write_mu_` ) but probably without using `ABSL_ACQUIRED_BEFORE` since it could have false positive when there are multiple instance of the classes in the same process

src/core/ext/transport/binder/wire_format/wire_reader_impl.h
```diff
@@ -99,9 +101,11 @@ class WireReaderImpl : public WireReader {
...
   absl::Status ProcessStreamingTransaction(transaction_code_t code,
                                            ReadableParcel* parcel);
-  absl::Status ProcessStreamingTransactionImpl(transaction_code_t code,
-                                               ReadableParcel* parcel,
-                                               int* cancellation_flags)
+  absl::Status ProcessStreamingTransactionImpl(
+      transaction_code_t code, ReadableParcel* parcel, int* cancellation_flags,
+      // The queue saves the actions needed to be done "WITHOUT" `mu_`.
+      // It prevents deadlock against wire writer issues.
+      std::queue<std::function<void()>>& deferred_func_queue)
```
sifmelcara, line 108: Have you considered inline this function and reconsider how to extract it in a logically way? Currently it requires caller to acquire `mu_` while calling it but it needs to do some tasks without holding`mu_`, which shows current way we structure the code probably doesn't make sense anymore.

sifmelcara, line 108: Not sure if it will be cleaner, but it is possible to combine multiple callbacks in that function so it can return single callback instead of a queue of callbacks.

sifmelcara, line 108: Nit: Use `absl::AnyInvocable<void()&&>` to make sure the function can only be called once

src/core/ext/transport/binder/wire_format/wire_reader_impl.cc
```diff
@@ -380,8 +392,12 @@ absl::Status WireReaderImpl::ProcessStreamingTransactionImpl(
                                                 "binder.authority");
       }
     }
-    transport_stream_receiver_->NotifyRecvInitialMetadata(
-        code, *initial_metadata_or_error);
+    deferred_func_queue.emplace([this, code,
+                                 initial_metadata_or_error = std::move(
+                                     initial_metadata_or_error)]() mutable {
```
sifmelcara, line 397: Do you need `mutable` here because `this` is captured? If yes, would it make sense to pass `this` as argument to avoid `mutable`?

sifmelcara (commented):
(I will let Mao stamp this)

sifmelcara:
> Have you test the change internally to make sure we don't accidentally break other use cases (e.g. use BinderTransport as gRPC server)? What's the plan to make sure this patch fixes the bug? Maybe stress test in real use case for a few hours? (please ping me in the chat room instead of replying here publicly)

Oh I see someone has already tested this. Then we should probably avoid making changes that could affect the behavior of the code, do some quick test to make sure this don't break other features, then merge this PR. Not sure if we should think about how we are going to test this kind of change more efficiently in the future (it is probably acceptable to do manual testing every time, as we only have to fix 2-3 bugs in the last 6 months so it is probably not too bad).

ctiller:
Just noting that the transport interfaces are going to change soon (once my
current project clears up a little I'll start work on doing that) and I
don't expect to be manually testing anything.

src/core/ext/transport/binder/wire_format/wire_reader_impl.cc
rock1246: Yes. We need the `mutable` here. If there is no `mutable`, `initial_metadata_or_error` will be const and NOT movable to the next function call.

src/core/ext/transport/binder/wire_format/wire_reader_impl.h
rock1246: I personally think multiple callbacks in the queue will be easier to understand/maintain for this case.

rock1246: I think it's fine that we do somethings needing `mu_` in `ProcessStreamingTransactionImpl()` and do something without `mu_` in `ProcessStreamingTransaction()`?

rock1246: Done! Thanks!

sifmelcara: Yeah it is fine, just looks strange. The function declaration is currently annotated with `ABSL_EXCLUSIVE_LOCKS_REQUIRED(mu_)`, and we made it a separate function because we want to put codes that need to be executed while holding `mu_` in it. Now, if we just inline the function, we can simply use local variables and will not need all these callback stuff that is hard to read

sifmelcara:
@ctiller Thanks for the information!

LittleCVR (approved):

PR #32442, opened 2023-02-21
Remove protobuf 4.22.0 from testing

gnossen:
This version broke backward compatibility in `plugin_pb2.py`, which is presumably a relatively minor regression, since we have  not yet heard any complaints about it. This PR:

- Excludes `4.22.0` from installation
- _Includes_ protobuf pre-releases into testing so this can be caught more quickly in the future.

When bad prereleases are caught, we can exclude them from testing in a similar manner to this PR. We may eventually want to invest into a system where we can define these bad versions centrally.

### DIFF
--- a/src/python/grpcio_channelz/setup.py
+++ b/src/python/grpcio_channelz/setup.py
@@ -64,7 +64,7 @@ PACKAGE_DIRECTORIES = {
 }
 
 INSTALL_REQUIRES = (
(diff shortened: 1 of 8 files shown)


XuanWang-Amos (approved):
LGTM, maybe open an issue to tract define bad versions centrally?

gnossen:
https://github.com/grpc/grpc/issues/32443

PR #21929, opened 2020-02-06
Autogenerate build.yaml equivalent from bazel BUILD metadata (with bazel query xml)

jtattermusch:
Improved version of https://github.com/grpc/grpc/pull/21889.

The main ideas:
- get rid of build.yaml and replace it by two files: build_autogenerated.yaml and build_handwritten.yaml
- A python script (currently named `extract_metadata_from_bazel_xml.py`) runs `bazel query deps(  ) --output xml` and postprocesses the xml rules data to generate a build_autogenerated.yaml file with the same structure as the original build.yaml
- build_handwritten.yaml is some extra configuration taken from the original build.yaml file

Current status:
- this is capable of generating a `build*.yaml` which can generate all public libraries, a bunch of helper libraries and a subset of run_tests.py tests.

- both Makefile and CMakeLists.txt have a subset of original tests. The two main categories of tests that are missing are the generated end2end tests and the generated qps tests. The rest of tests should be enough to verify that the build is sane (there's still a relatively large number of tests) and strictly speaking we don't need full coverage - the primary source of truth for testing are the bazel tests. 

After this change is merged, a lot of followup work can be made to simplify / cleanup the run_tests.py / make / cmake stack - in this PR I'm mostly focusing on getting as much parity with the old build.yaml as possible (so that we don't miss any tricky build issues).

PR #32444, opened 2023-02-21
h2_ssl_cert_test: re-add to build files and fix so that it builds

markdroth:
Looks like this was accidentally dropped from our build files in https://github.com/grpc/grpc/pull/21929, which means that this test hasn't actually been built or run in almost 3 years.  Unsurprisingly after all that time, I had to make some changes to the test to get it to actually build.


erm-g (approved):
LGTM

PR #32448, opened 2023-02-22
[build] Fix clang warnings in Windows code

drfloob:
This fixes the problems identified while building with clang-cl on Windows, with build arguments `/std:c++14 /W4`

Passes internal checks: cl/511562057

----

We can't yet enable a clang-cl build as part of our continuous integration tests due to a few issues:

protobuf fails an `unused-parameter` warning check in v4.21 (the current pinned version) on Windows. The upgrade to v4.22 is evidently painful and in progress. Without maintaining a patch against protobuf, or disabling warnings-as-errors somehow for the protobuf code alone, we'll need to upgrade our dependency before we can automate the clang-cl build for Windows.

Next, our Windows CI environment does not have clang installed. There has been some work over the past year to create custom kokoro images, but that work has apparently stalled after trading hands a few times. Using our current images, installing clang every time we run the job may be our best bet (likely from precompiled binaries that we host ourselves), but it will eat up more CI resources.

Finally, some of the default build configurations are incorrect for clang-cl. For example `-Wall` in clang-cl translates roughly to `-Weverything` in clang linux, whereas `-W4` in clang-cl translates more closely to `-Wall -Wextra`. This configuration in the gRPC bazel build is not currently platform-specific, it will need to be updated. Similarly, `-std=c++14` is an unknown argument on Windows (should be `/std:c++14`), and should not be in the bazelrc. This will likely need the same platform-specific support.


drfloob:
cc @jtattermusch @veblush

yijiem (approved):

PR #32449, opened 2023-02-22
[fix] Remove unnecessary gpr_atm dependency in grpc_opencensus_plugin

drfloob:


yashykt:
The title seems misleading

yashykt (approved):
fix the title please. We are removing gpr_atm dependency, not adding it

drfloob:
> fix the title please. We are removing gpr_atm dependency, not adding it

Fixed, thanks.

PR #26966, opened 2021-08-11
[3PI] Validate external account credentials token urls

renkelvin:
Hey @markdroth , this PR added an additional validation to the external account credentials urls. Could you please take a look and let me know if you have any questions. Thanks!
For external account credentials, the token url and service account impersonation url must follow the url patterns as follows:
```
*.sts.googleapis.com
sts.googleapis.com
sts.*.googleapis.com
*-sts.googleapis.com

*.iamcredentials.googleapis.com
iamcredentials.googleapis.com
iamcredentials.*.googleapis.com
*-iamcredentials.googleapis.com
```

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/26966)

PR #32450, opened 2023-02-22
feat: Auth lib: Remove 3PI config url validation

BigTailWolf:
To support TPC feature for BYOID (3PI), we need to remove the validation the pattern of impersonation endpoints, sts endpoints and token info endpoints since they are different in TPC regions.

A security review is already passed at b/261634871


eugeneo:
@BigTailWolf please sign the CLA. Login to the EasyCLA using your @google.com email.

@veblush I do not see any issues with this PR, can you take a look as well. Should this be added to the release notes?

BigTailWolf:
@eugeneo Signed in.

veblush:
Thanks for the PR!

I think this is the rollback of https://github.com/grpc/grpc/pull/26966 and the corresponding unit test might need to be removed.

veblush (approved):

PR #32452, opened 2023-02-22
Remove the if() to save some cpu cycles.

LeiW000:
The "reserved" var was not used in the function. So the if() can be removed. It saves some cpu cycles.

### DIFF
--- a/src/core/lib/surface/call.cc
+++ b/src/core/lib/surface/call.cc
@@ -3645,13 +3645,9 @@ grpc_call_error grpc_call_start_batch(grpc_call* call, const grpc_op* ops,
       "reserved=%p)",
       5, (call, ops, (unsigned long)nops, tag, reserved));
 
-  if (reserved != nullptr) {
-    return GRPC_CALL_ERROR;
-  } else {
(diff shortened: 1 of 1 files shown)


ctiller:
The reserved arguments were introduced at a time when we thought we'd need to offer ABI stability for gRPC Core.
In such a world, you need to check whether the reserved argument is null so that *when you start to use it* in a future version there's a guarantee that nobody is successfully passing non-null in that position -- although this should have been a crash and not an error.

History has taught us that we don't actually need to provide ABI stability at this layer: we never used that and nobody asked AFAIK.

All that said: I think we should keep the check for now, but open a gRFC to delete all `void* reserved` arguments from public APIs, and clean this up properly - and not in a haphazard fashion.

LeiW000:
> All that said: I think we should keep the check for now, but open a gRFC to delete all `void* reserved` arguments from public APIs, and clean this up properly - and not in a haphazard fashion.

Thanks for sharing the background. Do you want me to file the gRFC? Or, someone else will do it?

yijiem:
I opened a new issue for this at https://github.com/grpc/grpc/issues/32554. Closing this PR.

PR #32457, opened 2023-02-22
[build] Require Windows-specific code to be cherrypicked

drfloob:
Internal Windows builds will catch issues that we cannot yet catch in OSS. This will be mostly remedied once we have clang-cl in our CI (See a rough roadmap in https://github.com/grpc/grpc/pull/32448). For now, this PR identifies folders where most Windows-specific code is developed, and requires cherrypicks for PRs that touch anything inside those folders. 

This PR also refactors gpr and gprpp source files to better isolate  all platform-specific code ~the Windows-only code~. ~I will reorganize the other platform-specific files using this structure if there are no objections.~ 

This subset of folders covers about half of the `#ifdef GPR_WINDOWS` usages in gRPC, but nearly all of the actively-developed Windows code locations.


ctiller:
approach lgtm

drfloob:
> approach lgtm

Thanks. I did the rest of the folder migration for gpr & gprpp. All tests pass, auto-submit is enabled.

.github/labeler.yml
```diff
@@ -51,3 +51,7 @@ lang/ruby:
 - src/compiler/csharp*
 - src/csharp/**
 
+"disposition/Needs Internal Changes":
```
yijiem, line 54: Maybe add `- test/core/event_engine/windows/**` since that one seems to be standalone Windows code as well (though it's testing)?

yijiem (approved):

.github/labeler.yml
drfloob: Thanks, will do in a followup PR.

PR #32459, opened 2023-02-23
[build] Use platform-specific -std=c++14 compile flags

drfloob:
With the `--copt="-std=c++14"` setting in the bazel.rc file as it is today, MSVC builds have complained for every cc file: 

```
cl : Command line warning D9002 : ignoring unknown option '-std=c++14'
```

This adds thousands of lines of noise to Windows builds, and hides useful warnings. Using the `/std:c++14` flag on MSVC (and clang-cl) gets us the desired result.

### DIFF
--- a/test/distrib/bazel/cpp/.bazelrc
+++ b/test/distrib/bazel/cpp/.bazelrc
@@ -1,4 +1,3 @@
 # bazelrc file
 
-build --cxxopt='-std=c++14'
 build --host_cxxopt='-std=c++14'
--- a/test/distrib/bazel/cpp/.bazelrc
+++ b/test/distrib/bazel/cpp/.bazelrc
@@ -1,3 +1,15 @@
 # bazelrc file
+common --enable_platform_specific_config
 
-build --host_cxxopt='-std=c++14'
+build:windows --cxxopt='/std:c++14'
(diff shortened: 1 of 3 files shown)


test/distrib/bazel/cpp/.bazelrc
```diff
@@ -1,4 +1,15 @@
 # bazelrc file
+common --enable_platform_specific_config
```
gnossen, line 2: I'm surprised that this wasn't necessary for the Python Bazel distribtests. Any idea why? I guess `-std=c++14` might just be the default? But then I wouldn't expect the C++ Bazel distribtest to need this bazelrc update.

gnossen (approved):

test/distrib/bazel/cpp/.bazelrc
drfloob: Python bazel distribtests only run on linux. The garden-variety distribtests hide all output unless there's an error, but luckily we've had an error recently, and the logs contain Megabytes of `cl : Command line warning D9002 : ignoring unknown option '-std=c++14'`.

drfloob: I just noticed which file you commented on. Good question, this is mostly unnecessary at the moment, I'm going to revert this file.

PR #32461, opened 2023-02-23
[EventEngine] Revert "Revert "EventEngine::RunAt: C++ Alarm (#30024)" (#30147)"

drfloob:
This reverts commit 0fc0384b5aab5f05165534840087ffc843dcf5e3.

Major changes: this code calls `GetDefaultEventEngine` once on Alarm init instead of 7 times throughout.

I will run benchmarks to ensure b/237283941 is not reproduced.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1918,6 +1918,7 @@ grpc_cc_library(
         "//src/core:channel_args",
         "//src/core:channel_init",
         "//src/core:closure",
+        "//src/core:default_event_engine",
         "//src/core:error",
         "//src/core:gpr_atm",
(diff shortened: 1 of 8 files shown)


drfloob:
Update on benchmark results: this appears to have no effect on client & server QPS on the 1-byte-1-channel callback scenarios compared to running the benchmarks @ HEAD. It's an expensive test to run, so the results are not statistically significant (3 runs each), but the baseline@HEAD client QPS is **44,787 +/- 1,253 stdev**, and this PR's results are **44,489 +/- 1038 stdev**.

ctiller (approved):

PR #32467, opened 2023-02-24
Logging: Move filter to Core

yashykt:
This filter was originally written only for the C++ wrapped layer, but we have plans to use this for Python (and maybe other wrapped languages too in the future.)


XuanWang-Amos (approved):
Moving it to core helps a lot for Python and other wrapped languages, thanks for doing it!

PR #32468, opened 2023-02-24
[EventEngine] PosixEventEngine: Replace on_shutdown InternalError with more appropriate type

drfloob:
I've replaced all use of `InternalError` here because none of these scenarios would necessarily merit a bug or outage report.

Identified in the fuchsia test suite: calling the Listener's `on_shutdown` method with anything other than `absl::OkStatus()` would fail some assertions in the Posix-specialized client test suite if the Oracle were implemented similarly. It _should_ fail the same way in the listener test suite, but the statuses are ignored. I've fixed that.

### DIFF
--- a/src/core/lib/event_engine/posix_engine/posix_endpoint.cc
+++ b/src/core/lib/event_engine/posix_engine/posix_endpoint.cc
@@ -360,9 +360,9 @@ bool PosixEndpointImpl::TcpDoRead(absl::Status& status) {
       // 0 read size ==> end of stream
       incoming_buffer_->Clear();
       if (read_bytes == 0) {
-        status = TcpAnnotateError(absl::AbortedError("Socket closed"));
+        status = TcpAnnotateError(absl::InternalError("Socket closed"));
       } else {
-        status = TcpAnnotateError(absl::UnknownError(
+        status = TcpAnnotateError(absl::InternalError(
             absl::StrCat("recvmsg:", grpc_core::StrError(errno))));
       }
       return true;
@@ -889,7 +889,8 @@ void PosixEndpointImpl::UnrefMaybePutZerocopySendRecord(
 // release operations needed can be performed on the arg.
 void PosixEndpointImpl::TcpShutdownTracedBufferList() {
   if (outgoing_buffer_arg_ != nullptr) {
-    traced_buffers_.Shutdown(outgoing_buffer_arg_, absl::OkStatus());
+    traced_buffers_.Shutdown(outgoing_buffer_arg_,
+                             absl::InternalError("TracedBuffer list shutdown"));
     outgoing_buffer_arg_ = nullptr;
   }
 }
@@ -1139,7 +1140,7 @@ bool PosixEndpointImpl::Write(
   if (data->Length() == 0) {
     TcpShutdownTracedBufferList();
     if (handle_->IsHandleShutdown()) {
-      status = TcpAnnotateError(absl::AbortedError("EOF"));
+      status = TcpAnnotateError(absl::InternalError("EOF"));
       engine_->Run([on_writable = std::move(on_writable), status]() mutable {
         on_writable(status);
       });
--- a/src/core/lib/event_engine/posix_engine/posix_endpoint.cc
+++ b/src/core/lib/event_engine/posix_engine/posix_endpoint.cc
@@ -360,9 +360,9 @@ bool PosixEndpointImpl::TcpDoRead(absl::Status& status) {
       // 0 read size ==> end of stream
       incoming_buffer_->Clear();
       if (read_bytes == 0) {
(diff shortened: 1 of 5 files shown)


drfloob:
cc @tamird

test/core/event_engine/test_suite/tests/server_test.cc
```diff
@@ -174,7 +175,8 @@ TEST_F(EventEngineServerTest,
   args = args.Set(GRPC_ARG_RESOURCE_QUOTA, quota);
   ChannelArgsEndpointConfig config(args);
   auto listener = *test_ee->CreateListener(
-      std::move(accept_cb), [](absl::Status /*status*/) {}, config,
+      std::move(accept_cb),
+      [](absl::Status status) { ASSERT_TRUE(status.ok()); }, config,
```
tamird, line 179: in the event that these assertions fail, we probably want to discover the status?

drfloob: Good idea, done.

src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
```diff
@@ -246,7 +246,7 @@ absl::Status PosixEngineListenerImpl::HandleExternalConnection(
 void PosixEngineListenerImpl::AsyncConnectionAcceptor::Shutdown() {
   // The ShutdownHandle whould trigger any waiting notify_on_accept_ to get
   // scheduled with the not-OK status.
-  handle_->ShutdownHandle(absl::InternalError("Shutting down acceptor"));
+  handle_->ShutdownHandle(absl::AbortedError("Shutting down acceptor"));
```
tamird, line 249: does this merit a bug?

tamird (approved):

src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
drfloob: When this line runs, it is under normal operation. However, the implementation requires this be a non-ok status at the moment. It should be cleaned up, using something like an additional shutdown bit.

tamird: Is that _long_ for "yes"?

ctiller:
This needs to be combined with an audit of all endpoint consuming code
likely: there are likely places we forward endpoint errors into the request
path, and we have strict rules about what errors grpc is allowed to
generate on a request.

We'll likely need to build a status code firewall and convert some of these
back to internal along those paths.

src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
drfloob: Not really, it's an explanation, your question was ambiguous. I've added a TODO anyhow. It would be a whole project to systematically track the cleanup opportunities in this codebase. A valuable one, but not in scope for this PR.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -889,8 +889,7 @@ void PosixEndpointImpl::UnrefMaybePutZerocopySendRecord(
 // release operations needed can be performed on the arg.
 void PosixEndpointImpl::TcpShutdownTracedBufferList() {
   if (outgoing_buffer_arg_ != nullptr) {
-    traced_buffers_.Shutdown(outgoing_buffer_arg_,
-                             absl::InternalError("TracedBuffer list shutdown"));
+    traced_buffers_.Shutdown(outgoing_buffer_arg_, absl::OkStatus());
```
Vignesh2208, line 892: Are we sure using an OkStatus here wouldn't cause any problems ? The error passed to traced buffers shutdown in the iomgr tcp_posix.cc is a non OK status

drfloob:
> This needs to be combined with an audit of all endpoint consuming code

That is out of scope for this PR. The `InternalErrors` here are fundamentally incorrect, the PosixEventEngine would not pass the test suite as its own oracle. Some of this PR is necessary to get the tests to pass.

I ran this PR through our CI with the event_engine_client and event_engine_listener experiments enabled, and no new breakages were identified. Hopefully that is sufficient for now.

src/core/lib/event_engine/posix_engine/posix_endpoint.cc
drfloob: That's a good question. I think this is the _correct_ implementation here, but it may nonetheless be problematic. I'm going to pare this down to just the pieces we need at the moment to get the tests passing.

drfloob:
I have reduced this PR to just the parts that are needed to get the EventEngine to pass tests as its own oracle, and move on with Fuchsia. This affects the listener's on_shutdown callback alone, which in iomgr is only ever called with absl::OkStatus anyhow. I believe that to be a safe change.

PR #32475, opened 2023-02-24
[aio types] Fix some grpc.aio python types

Tasssadar:
With these, it is actually possible to have typed client stubs where the return type is correctly inferred.

It's only for the non-streaming calls, because there is `RequestIterableType` for the streaming ones (but it's just Any with extra steps and would require much more work).

### DIFF
--- a/src/python/grpcio/grpc/aio/_call.py
+++ b/src/python/grpcio/grpc/aio/_call.py
@@ -255,8 +255,7 @@ class _APIStyle(enum.IntEnum):
 class _UnaryResponseMixin(Call, Generic[ResponseType]):
     _call_response: asyncio.Task
 
-    def _init_unary_response_mixin(self,
-                                   response_task: asyncio.Task):
+    def _init_unary_response_mixin(self, response_task: asyncio.Task):
         self._call_response = response_task
 
     def cancel(self) -> bool:
@@ -572,6 +571,7 @@ class UnaryStreamCall(_StreamResponseMixin, Call, _base_call.UnaryStreamCall):
         await self._send_unary_request_task
         if self.done():
             await self._raise_for_status()
+
 
 # pylint: disable=too-many-ancestors
 class StreamUnaryCall(_StreamRequestMixin, _UnaryResponseMixin, Call,
--- a/src/python/grpcio/grpc/aio/_call.py
+++ b/src/python/grpcio/grpc/aio/_call.py
@@ -253,9 +253,10 @@ class _APIStyle(enum.IntEnum):
 
 
 class _UnaryResponseMixin(Call, Generic[ResponseType]):
-    _call_response: asyncio.Task[ResponseType]
+    _call_response: asyncio.Task
 
-    def _init_unary_response_mixin(self, response_task: asyncio.Task[ResponseType]):
+    def _init_unary_response_mixin(self,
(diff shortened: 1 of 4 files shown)


XuanWang-Amos:
Hi, thanks for the PR, the change looks fine to me except for one concern about changing `Awaitable` to `Generator` since it's not backward compatible and might break some peoples integration. Can you share more context on the issue you're having now and how this PR can help address it?

Tasssadar:
Hello, thanks for taking a look.

The `Awaitable` type for `__await__` is wrong. As per [mypy docs](https://mypy.readthedocs.io/en/stable/more_types.html#typing-async-await) and [PEP492](https://peps.python.org/pep-0492/), it's supposed to return a `Generator`, and it simply does not work (typecheck nor infer return type) with `Awaitable`.

Besides, if you look at the actual implementation of  `__await__`, that's what the method is returning, the `Awaitable` was never correct: https://github.com/grpc/grpc/pull/32475/files#diff-986f9aff481adca09f2b4e49c2d5daa8971a6c5e113bdd96002681e208c0aa1fR268-R293

Now, as for what I'm trying to achieve: `mypy-protobuf` generates type stubs like this for gRPC services:
```python
class FooServiceStub:
    def __init__(self, channel: grpc.Channel) -> None: ...

    GetFoo: grpc.UnaryUnaryMultiCallable[FooRequest,FooResponse]
```

It generates non-aio stubs, true, but if you put `import grpc.aio as grpc` on top, plus the changes in this PR, it actually typechecks and infers argument and return value types correctly.

XuanWang-Amos:
Hi, sorry for late reply and thanks for the additional context.

We agree with the proposed change, but we would like to verify that it does not break any `mypy` or `pytype` checks before merging it. We will update this thread once we had a chance to verify that.

In the meantime, please let us know if the change has already been tested.

Tasssadar:
Hi, thank you. I'm already using this in some projects, but they are not open-source so I can't provide any examples unfortunately.

src/python/grpcio/grpc/aio/_call.py
```diff
@@ -265,7 +265,7 @@ def cancel(self) -> bool:
         else:
             return False
 
-    def __await__(self) -> ResponseType:
+    def __await__(self) -> Generator[Any, None, ResponseType]:
```
XuanWang-Amos, line 268: Nit: We might need add `Generic` to this class too.

XuanWang-Amos:
Sorry about the delay. I just verified with both `mypy` and `pytype`, good news is we're now sure this change is good to merge, but bad news is we're [not checking aio files in our `pytype` workflow](https://github.com/grpc/grpc/blob/master/setup.cfg#L35), we might need fix that later.

I added a small comment, can you help address that and fix merge conflict?

Thanks again for the contribution!

Tasssadar:
Hello, sorry for the long turnaround, but I added the `Generic` and rebased the PR.

XuanWang-Amos:
Looks like some bazel tests are failing, you can check them locally using:
`bazel test //src/python/grpcio_tests/tests/unit:all`.
For sanity (format) related fixes, you can run:
`./tools/distrib/sanitize.sh` (Or `./tools/distrib/yapf_code.sh` and `./tools/distrib/isort_code.sh` etc manually).

XuanWang-Amos (approved):

PR #32487, opened 2023-02-27
Add verbatim if method's name collides with c# keyword

rmainil:
Add verbatim symbol (@) when a method's name collides with a c# keyword.

Without doing that, c# generated code is invalid. I ran into that issue with a rpc service definition containing a "goto" method : 

` rpc goto(stream Goto) returns (GotoResult); // Go to a specific timepoint`

That's obviously not a good name but in case we are not allowed to modify the service definition, generated code should be valid.


rmainil:
@jtattermusch, can you review that PR please ? (Sorry if you are not the right person, but I saw you were previously tagged on multiple other lang/C# PR, so I take my chance ;))

jtattermusch:
@tonydnewell  the issue reported by user is a fair concern, but we'd have to see about what's the right way of fixing.

I think we'd have to also fix in sync with protobuf C# (since it will suffer from pretty much the same problem).

I'm mostly worried about the next extra complexity it adds (and that over time the name mangling in grpc_csharp_plugin and protoc) will diverge even more. 

So this is something to think about.

Btw, I think if we wanted a C# generator that's fully keyword-safe, it's much more involved that simply introducing `GetSafeMethodName`, there are potential clashes in other names as well? And we don't want a half baked solution IMHO.

CC @jskeet  since this potentially impacts protobuf as well.

jskeet:
We don't have this precise issue in protobuf, as we don't generate anything for services. All our generated names *within* a message either have a `_` suffix (to avoid this) or are PascalCased, IIRC.

I believe we *would* have an issue with keywords being used as message names. (See also https://github.com/protocolbuffers/protobuf/issues/12291 which is about ToString, GetHashCode etc rather than C# keywords. I hadn't thought before, but the gRPC plugin may need to be aware of any fixes there - if it uses the helpers provided by the C# compiler plugin, we should be okay.)

tonydnewell:
@jtattermusch @jskeet I've looked at the C# code generation for gRPC services to see if there is anything in common with the protocol buffers code generation and suggestions made in https://github.com/grpc/grpc/pull/32487

As far as I can tell there is very little in common with the suggested changes to protocol buffers compiler. There are only a few things we have to do to prevent name clashes with services and methods in the services.

For **service names**, things we should _not_ allow are:

- C# keywords such as `for`, `enum` etc
    - these can be fixed by prefixing with `@` in the generated code
- "_Descriptor_" as it is the name of one of the methods that we generate
    - No obvious fix apart from maybe adding `_` after the name
- There are some pathological cases such as naming the service `__Helper_SerializeMessage` that we may wish to ignore

All other reserved names as defined in https://github.com/protocolbuffers/protobuf/issues/12291#issuecomment-1490110631 are OK and do not cause a problem.

For **method names**, things we should _not_ allow are:

- C# keywords such as `for`, `enum` etc
    - these can be fixed by prefixing with `@` in the generated code
- method names that append _Base_ or _Client_ to the service name, .e.g.
```
service Greeter {
	// These generate code that won't compile
	rpc GreeterBase(HelloRequest) returns (HelloReply) {}
	rpc GreeterClient(HelloRequest) returns (HelloReply) {}
}
```
Method names such as `ToString`, `Equals`, `Descriptor`, etc all seem to be OK.

As an aside - sharing functionality with the protocol buffers compiler and plugin seems to be broken at the moment in some circumstances - some functions are not visible depending on how the _grpc_plugin_support_ library is being built.

DmitryNaumov:
> All other reserved names as defined in [protocolbuffers/protobuf#12291 (comment)](https://github.com/protocolbuffers/protobuf/issues/12291#issuecomment-1490110631) are OK and do not cause a problem.

Same issue with messages named Parser, MergeFrom...

Issue #28433, opened 2021-12-27
Not compatible with Apple Silicon


m-yukio:
Not compatible with Apple Silicon .
For example, Apple Silicon does not exist in the definition below.

# grpc/src/python/grpcio/grpc_core_dependencies.py:

```
ASM_SOURCE_FILES = {
    ...
    'crypto_linux_x86_64': [
        ...
    ],
    'crypto_mac_x86': [
        ...
    ],
    'crypto_mac_x86_64': [
        ...
    ],
    'crypto_win_aarch64': [
        ...
    ],
    ...
```

# grpc/setup.py:

```
asm_key = ''
if BUILD_WITH_BORING_SSL_ASM and not BUILD_WITH_SYSTEM_OPENSSL:
    boringssl_asm_platform = BUILD_OVERRIDE_BORING_SSL_ASM_PLATFORM if BUILD_OVERRIDE_BORING_SSL_ASM_PLATFORM else util.get_platform(
    )
    LINUX_X86_64 = 'linux-x86_64'
    LINUX_ARM = 'linux-arm'
    LINUX_AARCH64 = 'linux-aarch64'
    if LINUX_X86_64 == boringssl_asm_platform:
        asm_key = 'crypto_linux_x86_64'
    elif LINUX_ARM == boringssl_asm_platform:
        asm_key = 'crypto_linux_arm'
    elif LINUX_AARCH64 == boringssl_asm_platform:
        asm_key = 'crypto_linux_aarch64'
    elif "mac" in boringssl_asm_platform and "x86_64" in boringssl_asm_platform:
        asm_key = 'crypto_mac_x86_64'
    else:
        print("ASM Builds for BoringSSL currently not supported on:",
              boringssl_asm_platform)
```

Maybe AARCH64 or arm64 or arm64e.

muhammadbilal20130:
...

teamzz111:
Same here

`
"Error loading native library. Not found in any of the possible locations: /Users/andreslargo/Develop/Chamba/chamba-business/Chamba-Business-Feed/bin/Debug/net6.0/libgrpc_csharp_ext.arm64.dylib,/Users/andreslargo/Develop/Chamba/chamba-business/Chamba-Business-Feed/bin/Debug/net6.0/
`

`
[...]

PR #32492, opened 2023-02-28
Update third_party/boringssl-with-bazel

jtattermusch:
Upgrade boringssl to the latest "master-with-bazel"
- use the `'USE_HEADERMAP' => 'NO'` fix for ObjC
- update the key for asm optimizations on mac/apple in python's setup.py

This PR depends on monterey fixes here: https://github.com/grpc/grpc/pull/32493 and the boringssl's build simplification https://boringssl-review.googlesource.com/c/boringssl/+/56465.


jtattermusch:
The "Basic Tests Ruby" failure in `grpc_class_init_test.rb` on Linux and MacOS seems unrelated to this PR, since I saw occurrences of it on master branch as well.

setup.py
```diff
@@ -375,7 +375,9 @@ def _quote_build_define(argument):
     elif LINUX_AARCH64 == boringssl_asm_platform:
         asm_key = 'crypto_linux_aarch64'
     elif "mac" in boringssl_asm_platform and "x86_64" in boringssl_asm_platform:
-        asm_key = 'crypto_mac_x86_64'
+        asm_key = 'crypto_apple_x86_64'
+    elif "mac" in boringssl_asm_platform and "arm64" in boringssl_asm_platform:
```
jtattermusch, line 379: Is the `"mac" in boringssl_asm_platform and "arm64" in boringssl_asm_platform` combination actually used anywhere at this point or is it just a proactive change?
Also is "arm64" the right value to use for mac? (linux uses "aarch64")

CC @HannahShiSFB @gnossen

veblush (approved):

setup.py
HannahShiSFB: Needed for M1 mac.
This should fix https://github.com/grpc/grpc/issues/28433 as well

jtattermusch: AFAIK we don't publish python wheels for M1 Macs currently, so it sounds like this is being proactive.
Anyway, the entire section that select the asm_key can be simplified as a followup (since we the asm files are `#ifdef` conditioned to only kick in on the right platform, it's no longer important so select only the asm files that exactly match the target platforms, we can simply determine whether we want the asm optimizations or not, set OPENSSL_NO_ASM accordingly and include all gcc-style asm files in the build).

Since likely want the followup simplification, it's fine to leave this change as is.

jtattermusch:
In order to doublecheck that the boringssl build change won't result in silently dropping asm optimizations for some configurations where they've previously been enabled (which would be a performance regression),
I also did an experiment with a patched version of boringssl (which breaks the build if asm optimizations aren't enabled).
The results are in the comment [here](https://github.com/grpc/grpc/pull/32506#issuecomment-1451478059).

Based on the results, it seems that there are no build configurations where boringssl optimizations were enabled and became disabled by the borignssl change. We can thus proceed with the boringssl upgrade.

PR #32357, opened 2023-02-12
Update third_party/boringssl-with-bazel

HannahShiSFB:
Update boringssl to fd3e770 on Feb 11, 2023

PR #32493, opened 2023-02-28
Make remaining objc test jobs to pass on MacOS monterey

jtattermusch:
Make remaining objC jobs compatible with kokoro monterey workers and prepare for boringssl upgrade.

The changes here are taken from https://github.com/grpc/grpc/pull/32357, but they should be merged in a separate PR
(we need the changes to be able to upgrade to monterey anyway and there's no reason to make the boringssl upgrade PR more complicated by bundling more fixes into it).

I've checked that the grpc_basictests_objc_examples and grpc_ios_binary_size are green if switched to monterey.
Unfortunately it's hard to make grpc_basictests_objc_examples pass on both monterey and mojave, so I suggest merging this PR at the same time as CL to upgrade the kokoro jobs to monterey.
- that way both PR and continuous runs will remain green
- older branches would need a backport anyway

### DIFF
--- a/src/objective-c/BoringSSL-GRPC.podspec
+++ b/src/objective-c/BoringSSL-GRPC.podspec
@@ -103,11 +103,6 @@ Pod::Spec.new do |s|
   # following lets users write `#include <openssl/ssl.h>`.
   s.header_dir = name
 
-  # The module map and umbrella header created automatically by Cocoapods don't work for C libraries
-  # like this one. The following file, and a correct umbrella header, are created on the fly by the
-  # `prepare_command` of this pod.
-  s.module_map = 'src/include/openssl/BoringSSL.modulemap'
-
   # We don't need to inhibit all warnings; only -Wno-shorten-64-to-32. But Cocoapods' linter doesn't
   # want that for some reason.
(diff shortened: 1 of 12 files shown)


jtattermusch:
CC @HannahShiSFB

jtattermusch:
The CL to switch remaining objC jobs to monterey is here: https://critique.corp.google.com/cl/512999186

veblush (approved):

sampajano (approved):
@jtattermusch Thanks! Makes sense to unblock the upgrade to monterey first!

HannahShiSFB:
LGTM, Thanks Jan @jtattermusch

jtattermusch:
Adhoc runs on macos monterey(both green)

iOS binary size:
https://source.cloud.google.com/results/invocations/def82992-25c5-4274-ba3c-1fac19726387
ObjC examples:
https://source.cloud.google.com/results/invocations/5239887c-308d-498b-ae4b-700aac2f3ede

jtattermusch:
The ObjC swiftsample is failing in presubmit since the PR job is still running on macos mojave  (cl/512999186 will switch it to monterey).
https://source.cloud.google.com/results/invocations/fca57f00-17d3-4105-ab1c-74ca8509d289/targets

PR #32491, opened 2023-02-27
client channel: fix TSAN flake in FilterBasedLoadBalancedCall

markdroth:
This fixes a very strange TSAN flake seen in an internal test (b/268292646), which seems to have been introduced in https://github.com/grpc/grpc/pull/32326.

I've included an example of the TSAN failure below, for future reference.  I don't fully understand what is causing this failure.  It looks like the async callback from a previous picker update caused the pick to be re-queued but then got stuck draining queued callbacks in its `ExecCtx` instance, while the async callback from the next picker update had already failed the call.  But the part I don't understand is why the call combiner cancellation callback wound up scheduled on the `ExecCtx` instance from the first thread -- that seems likely to be caused by either the `WorkSerializer` or maybe `ExecCtx` work-stealing, but the exact details are likely to be hard to nail down.

Switching from `EventEngine::Run()` back to `ExecCtx::Run()` seems to fix the flake, so that's what I'm doing in this PR.  I find this work-around deeply dissatisfying, especially since I do not fully understand the root cause of the problem.  However, given that we are working toward eliminating the `FilterBasedLoadBalancedCall` code entirely as part of the promise conversion, this problem is probably not worth further investigation.

```
WARNING: ThreadSanitizer: data race (pid=7406)
  Read of size 8 at 0x7b8800037558 by thread T47:
    #0 grpc_core::ClientChannel::FilterBasedLoadBalancedCall::~FilterBasedLoadBalancedCall() third_party/grpc/src/core/ext/filters/client_channel/client_channel.cc:2780:5 (libST-a5842b3303c3_third_Uparty_Sgrpc_Sgoogle_Slibgrpc_Uclient_Uchannel.so+0x530ed) (BuildId: a8369742d3c435f7f1a244a535f4907c)
    #1 Delete third_party/grpc/src/core/lib/gprpp/ref_counted.h:248:31 (libST-a5842b3303c3_third_Uparty_Sgrpc_Sgoogle_Slibgrpc_Uclient_Uchannel.so+0x64a73) (BuildId: a8369742d3c435f7f1a244a535f4907c)
    #2 Unref third_party/grpc/src/core/lib/gprpp/orphanable.h:102:7 (libST-a5842b3303c3_third_Uparty_Sgrpc_Sgoogle_Slibgrpc_Uclient_Uchannel.so+0x64a73)
[...]

PR #32499, opened 2023-02-28
c2p resolver test: override the XDS server address, point it to a fake server that immediatley closes the connection

apolcyn:
Alongside https://github.com/grpc/grpc/pull/32496, this makes this test behave the same on all platforms.

FWIW, I verified this causes us to see the previous lock cycle problem in https://github.com/grpc/grpc/pull/32491 on linux - originally that lock cycle was only on mac, because of environmental differences between mac and linux in CI.


markdroth (approved):

apolcyn:
bazel failures are "tool failure"s

PR #32500, opened 2023-03-01
Propose a fix for GRP_DEBUG_ASSERT failures seen in Google-internal l…

GPacenza:
…ibrary

Context: A Google-internal change to the Scaffolding framework included a third party grpc header file (`grpc/include/grcpp/security/server_credentials.h`) into some google3 code, and an internal library's build started to break with the following error:
```
third_party/grpc/include/grpcpp/support/callback_common.h:210:33: error: reference to non-static member function must be called; did you mean to call it with no arguments?
    GPR_DEBUG_ASSERT(ignored == ops);
                                ^~~
third_party/grpc/include/grpc/support/log.h:103:40: note: expanded from macro 'GPR_DEBUG_ASSERT'
#define GPR_DEBUG_ASSERT(x) GPR_ASSERT(x)
                                       ^
third_party/grpc/include/grpc/support/log.h:97:24: note: expanded from macro 'GPR_ASSERT'
    if (GPR_UNLIKELY(!(x))) {                       \
                       ^
./third_party/grpc/include/grpc/support/port_platform.h:754:43: note: expanded from macro 'GPR_UNLIKELY'
#define GPR_UNLIKELY(x) __builtin_expect((x), 0)
```
It seems that when compiling, NDEBUG is defined, so the line `auto* ops = ops_;` is removed by the preprocessor. However, the GPR_DEBUG_ASSERT macro still expands to something like the following:
```
#ifndef NDEBUG
  do { ... } while (0);
#endif
```
When compiled in "non-optimized mode" with blaze (i.e. without -c opt), it seems like the contents of the loop are evaluated, so the symbol will need to be declared.

This proposed fix unblocks the original internal Scaffolding change (which had to be rolled back) and has been confirmed to fix the internal build error mentioned above.

### DIFF
--- a/include/grpcpp/support/callback_common.h
+++ b/include/grpcpp/support/callback_common.h
@@ -203,11 +203,10 @@ class CallbackWithSuccessTag : public grpc_completion_queue_functor {
     void* ignored = ops_;
     // Allow a "false" return value from FinalizeResult to silence the
     // callback, just as it silences a CQ tag in the async cases
-#ifndef NDEBUG
-    auto* ops = ops_;
-#endif
+    auto* ops_tag = ops_;
+
     bool do_callback = ops_->FinalizeResult(&ignored, &ok);
-    GPR_DEBUG_ASSERT(ignored == ops);
+    GPR_DEBUG_ASSERT(ignored == ops_tag);
 
     if (do_callback) {
       CatchingCallback(func_, ok);


ctiller:
I don't understand this.

In log.h:
```
#ifndef NDEBUG
#define GPR_DEBUG_ASSERT(x) GPR_ASSERT(x)
#else
#define GPR_DEBUG_ASSERT(x)
#endif
```

so if `NDEBUG` is defined, `GPR_DEBUG_ASSERT(...)` expands to nothing.

This looks like a build configuration problem of some sort.

GPacenza:
Thanks for the feedback. Do you have any thoughts on what kind of build configuration problem this could be? I am not very familiar with this code, or the library that was broken, so I admittedly don't have a great understanding of how to move forward with debugging here.

In an email thread with us and tonyliaoss@ (titled "cl/494018870 broke our tests") from December, Tony listed some steps (which I've just now reproduced) in which he compiled one of the broken Hawksbill library's dependencies and showed that, when it seemed that `NDEBUG` was not defined, the `auto* ops = ops_;` was removed as expected, but the `GPR_DEBUG_ASSERT` macro still expands to `do { if (__builtin_expect((!(ignored == ops_)), 0)) { gpr_assertion_failed("third_party/grpc/include/grpcpp/support/callback_common.h", 121, "ignored == ops_"); } } while (0);`. (There are some slight differences in what he found then and what I've reproduced now since the original macro was `GPR_CODEGEN_DEBUG_ASSERT`, which is now replaced with `GPR_DEBUG_ASSERT`). So it seems that `NDEBUG` may be defined when the definition of `GPR_DEBUG_ASSERT` is getting evaluated, but not when initializing `ops` in callback_common.h?

Followup question from tonyliaoss@ in the referenced email thread:

"By including the header into google3 code from third_party, maybe the toolchain changed? E.g. compiler moves from gcc to clang, or some of the flags/compiler options are changed from less strict to more strict. Craig, could you confirm this?"

ctiller:
So I don't know what's triggering this, other than it seems very likely something somewhere is redefining NDEBUG. That's ridiculously dangerous and I think the root cause should be found rather than this problem papered over.

I don't believe we will generate a valid build if NDEBUG is defined in some places but not others.

GPacenza:
Hi Craig, could you please take a look at the follow-up email Tony sent a few days ago? I think he provided more details about the error we are seeing

yashykt:
@ctiller is this still relevant?

ctiller:
I think this can be closed.

PR #32506, opened 2023-03-01
Boringssl noasm experiment

jtattermusch:
Inject a version of boringssl that breaks build if asm optimizations are not enabled.


jtattermusch:
boringssl assembly optimization presence based on test results:
- PHP linux/macos: NO
- Python wheels Linux x86 (32-bit): NO
- Python wheels MacOS: NO 
- Python wheels windows: NO
- Ruby windows/linux/macos: NO
- C/C++ RBE Linux MSAN and UBSAN: NO (since for msan/ubsan we disable asm optimizations)
- C/C++ RBE Windows: NO
- win C++ distribtests: x86 as external project: NO

PR #32466, opened 2023-02-23
GCP Observability testing for C++

stanley-cheung:
This PR enables GCP Observability testing for C++.

This re-uses the existing interop test client and server and adds an `--enable_observability=true` flag.

This also adds a `Dockerfile` and `run.sh` script to allow the test framework to execute the test. 

Framework PR: https://github.com/GoogleCloudPlatform/grpc-gcp-tools/pull/18
Java PR: https://github.com/grpc/grpc-java/pull/9858
Golang PR: https://github.com/grpc/grpc-go/pull/5979

PR #32509, opened 2023-03-01
Refactor C++ interop client flags

stanley-cheung:
Refactor C++ interop test client flags into the common `client_helper.h/cc`. This is needed by the observability testing PR #32466 

We need the `ABSL_DECLARE_FLAG` in the header file so that we can share that across different implementation.


yashykt (approved):
ok, as long as tests pass

PR #32524, opened 2023-03-02
[testing]: Add "orca_per_rpc" test case

eugeneo:


test/cpp/interop/interop_client.cc
```diff
@@ -116,11 +136,30 @@ void InteropClient::ServiceStub::ResetChannel() {
...
+}
+
+InteropClient::InteropClient(LoadReportsFromLbPolicy load_reports,
+                             ChannelCreationFunc channel_creation_func,
                              bool new_stub_every_test_case,
                              bool do_not_abort_on_transient_failures)
     : serviceStub_(std::move(channel_creation_func), new_stub_every_test_case),
-      do_not_abort_on_transient_failures_(do_not_abort_on_transient_failures) {}
+      do_not_abort_on_transient_failures_(do_not_abort_on_transient_failures),
+      load_reports_(load_reports) {}
```
markdroth, line 162: `std::move()`

test/cpp/interop/interop_client.h
```diff
@@ -39,11 +39,17 @@ typedef std::function<std::shared_ptr<Channel>(void)> ChannelCreationFunc;
 
 class InteropClient {
  public:
+  using LoadReportsFromLbPolicy =
+      std::shared_ptr<std::vector<xds::data::orca::v3::OrcaLoadReport>>;
+
+  static LoadReportsFromLbPolicy InitializeCustomLbPolicy();
```
markdroth, line 45: Instead of exposing this type and passing it into the ctor, why not just register the LB policy inside the ctor and have the callback directly write to the data member of `InteropClient`, so that it doesn't need to be a `shared_ptr<>`?  Even if you need this behavior to be configurable for the Android client, you can pass in a bool indicating whether to configure the LB policy.

test/cpp/interop/interop_server.cc
```diff
@@ -136,6 +139,26 @@ bool CheckExpectedCompression(const ServerContext& context,
...
+    recorder->RecordCpuUtilizationMetric(request_metrics.cpu_utilization());
+  }
+  if (request_metrics.mem_utilization() > 0) {
+    recorder->RecordMemoryUtilizationMetric(request_metrics.mem_utilization());
+  }
+  if (request_metrics.rps_fractional() > 0) {
+    recorder->RecordQpsMetric(request_metrics.rps_fractional());
+  }
+  for (const auto& p : request_metrics.request_cost()) {
+    recorder->RecordRequestCostMetric(p.first.c_str(), p.second);
```
markdroth, line 155: Are you sure this works correctly?  This is storing `absl::string_view` keys in the `CallMetricRecorder` that point to storage in the request proto, but I don't think that request proto is guaranteed to live until the ORCA metadata is serialized to be sent back to the client, so it seems like this could leave pointers dangling into unsafe memory.

In our e2e tests, we deal with this by arena-allocating a copy of these strings:

https://github.com/grpc/grpc/blob/c5213ceff9edc723cbf56a5f917083ff4a025bf0/test/cpp/end2end/client_lb_end2end_test.cc#L141

test/cpp/interop/interop_server.cc
```diff
@@ -360,6 +386,9 @@ void grpc::testing::interop::RunServer(
   if (absl::GetFlag(FLAGS_max_send_message_size) >= 0) {
     builder.SetMaxSendMessageSize(absl::GetFlag(FLAGS_max_send_message_size));
   }
+  auto metric_recorder = experimental::ServerMetricRecorder::Create();
```
markdroth, line 389: In this particular case, we don't need a `ServerMetricRecorder`, so I don't think it's necessary to create this.  You can just pass null to `EnableCallMetricRecording()`.

markdroth (commented):
Thanks for doing this.  This looks really good overall!

My main comment is the one about exposing the type of the map to store the backend metrics in.  I think we can just do that internally inside of `InteropClient`, so that it doesn't need to be exposed to callers.

src/proto/grpc/testing/BUILD
```diff
@@ -184,17 +190,33 @@ grpc_proto_library(
...
+    srcs = ["orca_load_report_gen_proto_file"],
+    has_services = False,
+)
+
 genrule(
     name = "messages_gen_proto_file",
     srcs = ["messages.proto"],
     outs = ["messages_gen.proto"],
-    cmd = "cp $< $@",
+    cmd = "sed 's/xds\\/v3\\/orca_load_report.proto/orca_load_report_gen.proto/' $< > $@",
```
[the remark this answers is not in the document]
eugeneo: Done

test/cpp/interop/client.cc
```diff
@@ -124,6 +125,9 @@ ABSL_FLAG(
     bool, log_metadata_and_status, false,
     "If set to 'true', will print received initial and trailing metadata, "
     "grpc-status and error message to the console, in a stable format.");
+ABSL_FLAG(absl::optional<std::string>, service_config_json, absl::nullopt,
```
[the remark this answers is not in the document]
eugeneo: I like optionals - compiler makes sure I don't forget to check if it is set...

I switched to string.

test/core/util/test_lb_policies.h
```diff
@@ -62,7 +65,9 @@ using InterceptRecvTrailingMetadataCallback =
 // invokes cb when trailing metadata is received for each call.
 void RegisterInterceptRecvTrailingMetadataLoadBalancingPolicy(
     CoreConfiguration::Builder* builder,
-    InterceptRecvTrailingMetadataCallback cb);
+    InterceptRecvTrailingMetadataCallback cb,
+    absl::string_view factory_name =
+        kInterceptRecvTrailingMetadataLbPolicyName);
```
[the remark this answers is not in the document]
eugeneo: This policy is referred by name [here](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#orca_per_rpc) - I am assuming this config will be provided from outside. I did entertain a though of changing the default name for the policy, lmk if you'd rather have that.

test/cpp/interop/client_helper.cc
```diff
@@ -87,7 +87,12 @@ std::shared_ptr<Channel> CreateChannelForTestCase(
     const std::string& test_case,
     std::vector<
         std::unique_ptr<experimental::ClientInterceptorFactoryInterface>>
-        interceptor_creators) {
+        interceptor_creators,
+    absl::optional<std::string> service_config_json) {
```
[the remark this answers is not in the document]
eugeneo: It would require `ABSL_FLAG` in each client, that would make this PR much bigger. Another option would be to have that declaration in `client_helper.cc` which seems to be discouraged.

src/android/test/interop/app/src/main/cpp/grpc-interop.cc
```diff
@@ -39,7 +40,9 @@ std::shared_ptr<grpc::testing::InteropClient> GetClient(const char* host,
   grpc::testing::ChannelCreationFunc channel_creation_func =
       std::bind(grpc::CreateChannel, host_port, credentials);
   return std::shared_ptr<grpc::testing::InteropClient>(
-      new grpc::testing::InteropClient(channel_creation_func, true, false));
+      new grpc::testing::InteropClient(
+          grpc::testing::InteropClient::LoadReportsFromLbPolicy(),
```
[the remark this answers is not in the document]
eugeneo: I am not familiar with Android at this point. This line only initializes the custom LB policy.

test/cpp/interop/interop_client.h
eugeneo: `serviceStub_` creation opens a channel, so LB policy needs to be created before that one is created. I considered multiple approaches but then decided that passing it here will offer more flexibility as I am not familiar enough with all the tests that use this code.

src/android/test/interop/app/src/main/cpp/grpc-interop.cc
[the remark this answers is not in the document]
eugeneo: Reverted

test/core/util/BUILD
```diff
@@ -289,7 +289,6 @@ grpc_cc_test(
 
 grpc_cc_library(
     name = "test_lb_policies",
-    testonly = 1,
```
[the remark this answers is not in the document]
eugeneo: There are more executables than just interop_client. I am not sure what are the implications of marking them all as `testonly`.

test/cpp/interop/interop_client.h
eugeneo: Please take a look at a new implementation

test/cpp/interop/interop_server.cc
eugeneo: Copied. Was not familiar with this API.

test/cpp/interop/interop_server.cc
eugeneo: Done.

eugeneo (commented):

test/cpp/interop/interop_client.h
markdroth: Thanks, I like this a lot better, since the ugliness is now hidden inside the client code.

In the long run, I think we should probably change this LB policy to get the address of the callback from a channel arg instead of it being globally set at LB policy registration time.  But that doesn't need to block this PR.

eugeneo:
This was an extensive change. I did a full reset to revert the files that no longer need to be changed. Please take another look.

I don't think failed checks should block the review at this point.

test/cpp/interop/interop_client.h
eugeneo: Done

test/cpp/interop/client_helper.cc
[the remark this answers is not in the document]
eugeneo: No longer needed. Clients need to configure all args.

test/core/util/test_lb_policies.h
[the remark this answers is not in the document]
eugeneo: Now this is a distinct LB policy

test/cpp/interop/backend_metrics_lb_policy.cc
```diff
@@ -0,0 +1,240 @@
...
+    delegate_args.work_serializer = work_serializer();
+    delegate_args.args = channel_args();
+    delegate_args.channel_control_helper =
+        std::make_unique<Helper>(RefCountedPtr<BackendMetricsLbPolicy>(this));
+    load_report_tracker_ = delegate_args.args.GetPointer<LoadReportTracker>(
+        kMetricsTrackerAttribute);
+    delegate_ =
+        CoreConfiguration::Get().lb_policy_registry().CreateLoadBalancingPolicy(
+            "pick_first", std::move(delegate_args));
+    grpc_pollset_set_add_pollset_set(delegate_->interested_parties(),
```
markdroth, line 78: We need to remove this pollset_set linkage in the dtor.

test/cpp/interop/backend_metrics_lb_policy.cc
```diff
@@ -0,0 +1,240 @@
...
+class BackendMetricsLbPolicy : public LoadBalancingPolicy {
+ public:
+  explicit BackendMetricsLbPolicy(Args args)
+      : LoadBalancingPolicy(std::move(args), /*initial_refcount=*/2) {
+    Args delegate_args;
+    delegate_args.work_serializer = work_serializer();
+    delegate_args.args = channel_args();
+    delegate_args.channel_control_helper =
+        std::make_unique<Helper>(RefCountedPtr<BackendMetricsLbPolicy>(this));
+    load_report_tracker_ = delegate_args.args.GetPointer<LoadReportTracker>(
```
markdroth, line 73: Suggest moving this to the top of the function, before you start to build delegate_args, since it's really not related to constructing the delegate policy.  Note that it's fine to call `channel_args()` twice, since that method returns a const reference.

Also, please add an assertion that `load_report_tracker_` is non-null.

test/cpp/interop/backend_metrics_lb_policy.cc
```diff
@@ -0,0 +1,240 @@
...
+      std::make_unique<BackendMetricsLbPolicyFactory>());
+}
+
+void LoadReportTracker::SetupOnChannel(ChannelArguments* arguments) {
+  arguments->SetPointer(kMetricsTrackerAttribute, this);
+}
+
+void LoadReportTracker::RecordPerRpcLoadReport(
+    const grpc_core::BackendMetricData* backend_metric_data) {
+  if (backend_metric_data != nullptr) {
```
markdroth, line 222: I don't think we need to check whether it's null here, because `BackendMetricDataToOrcaLoadReport()` will return nullopt in that case.  I think we can just unconditionally add the result of `BackendMetricDataToOrcaLoadReport()` to the queue.

test/cpp/interop/interop_client.h
```diff
@@ -109,8 +121,9 @@ class InteropClient {
    public:
     // If new_stub_every_call = true, pointer to a new instance of
     // TestServce::Stub is returned by Get() everytime it is called
-    ServiceStub(ChannelCreationFunc channel_creation_func,
-                bool new_stub_every_call);
+    ServiceStub(
+        std::function<std::shared_ptr<grpc::Channel>()> channel_creation_func,
```
markdroth, line 125: Shouldn't this still use `ChannelCreationFunc` rather than spelling out the type?

test/cpp/interop/interop_client.h
```diff
@@ -119,7 +132,7 @@ class InteropClient {
     void ResetChannel();
 
    private:
-    ChannelCreationFunc channel_creation_func_;
+    std::function<std::shared_ptr<Channel>()> channel_creation_func_;
```
markdroth, line 135: Same here.

test/cpp/interop/interop_client.h
```diff
@@ -154,6 +167,9 @@ class InteropClient {
   ServiceStub serviceStub_;
   /// If true, abort() is not called for transient failures
   bool do_not_abort_on_transient_failures_;
+  // Load Orca metrics captured by the custom LB policy. Needs to be initialized
```
markdroth, line 170: I don't think we need the "Needs to be initialized first" comment anymore.

test/cpp/interop/client.cc
```diff
@@ -225,7 +238,15 @@ int main(int argc, char** argv) {
         factories.emplace_back(
             new grpc::testing::MetadataAndStatusLoggerInterceptorFactory());
       }
-      return CreateChannelForTestCase(test_case, std::move(factories));
+      grpc::ChannelArguments arguments;
+      std::string service_config_json =
+          absl::GetFlag(FLAGS_service_config_json);
+      if (!service_config_json.empty()) {
+        arguments.SetServiceConfigJSON(service_config_json);
```
markdroth, line 245: `std::move()`

test/cpp/interop/interop_client.h
```diff
@@ -35,17 +38,25 @@ typedef std::function<void(const InteropClientContextInspector&,
                            const SimpleRequest*, const SimpleResponse*)>
     CheckerFn;
 
-typedef std::function<std::shared_ptr<Channel>(void)> ChannelCreationFunc;
+typedef std::function<std::shared_ptr<Channel>()> ChannelCreationFunc;
+
+typedef std::function<std::shared_ptr<Channel>(
```
markdroth, line 43: Instead of using a function that takes another function as a parameter, I think it would be simpler to just pass in the channel args from the caller, like this:

```
typedef std::function<std::shared_ptr<Channel>(ChannelArguments)>
    ChannelCreationFuncWithCustomArgs;
```

Then the new `InteropClient` ctor can look like this:

```
InteropClient::InteropClient(
    ChannelCreationFuncWithCustomArgs channel_creation_func,
    bool new_stub_every_test_case, bool do_not_abort_on_transient_failures)
    : serviceStub_(
          [&]() {
            InitializeCustomLbPolicyIfNeeded();
            ChannelArguments args;
            load_report_tracker_->SetupOnChannel(&args);
            return channel_creation_func(std::move(args));
          },
          new_stub_every_test_case),
      do_not_abort_on_transient_failures_(do_not_abort_on_transient_failures) {}
```

And the lambdas in client.cc can just add their own args to the passed-in args.

test/cpp/interop/interop_client.h
```diff
@@ -35,17 +38,25 @@ typedef std::function<void(const InteropClientContextInspector&,
...
   /// created for every test case
-  /// If do_not_abort_on_transient_failures is true, abort() is not called in
-  /// case of transient failures (like connection failures)
+  /// If do_not_abort_on_transient_failures is true, abort() is not called
+  /// in case of transient failures (like connection failures)
   explicit InteropClient(ChannelCreationFunc channel_creation_func,
                          bool new_stub_every_test_case,
                          bool do_not_abort_on_transient_failures);
+  /// Same as ctor above, allows customizing channel arguments
+  explicit InteropClient(
```
markdroth, line 57: Instead of adding a second ctor, why not just change the existing callers to use the new parameter?

test/cpp/interop/client.cc
```diff
@@ -194,18 +199,26 @@ int main(int argc, char** argv) {
...
       if (absl::GetFlag(FLAGS_log_metadata_and_status)) {
         factories.emplace_back(
             new grpc::testing::MetadataAndStatusLoggerInterceptorFactory());
       }
-      return CreateChannelForTestCase(test_case, std::move(factories));
+      grpc::ChannelArguments arguments;
+      std::string service_config_json =
+          absl::GetFlag(FLAGS_service_config_json);
+      if (!service_config_json.empty()) {
+        arguments.SetServiceConfigJSON(service_config_json);
```
markdroth, line 217: `std::move()`

markdroth (commented):
The new LB policy looks great!

Just one substantive comment here, which is the one about simplifying the way that channel args are passed through the interop code.  The current approach seems more complicated than it needs to be, and it took me a while to understand.

test/cpp/interop/backend_metrics_lb_policy.h
```diff
@@ -0,0 +1,54 @@
...
+
+class LoadReportTracker {
+ public:
+  void SetupOnChannel(ChannelArguments* arguments);
+  void RecordPerRpcLoadReport(
+      const grpc_core::BackendMetricData* backend_metric_data);
+
+  void ClearPerRpcLoadReports() { per_rpc_load_reports_.clear(); }
+
+  void AssertHasSinglePerRpcLoadReport(
```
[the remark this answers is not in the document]
eugeneo: Done

test/cpp/interop/backend_metrics_lb_policy.h
```diff
@@ -0,0 +1,54 @@
...
+      const xds::data::orca::v3::OrcaLoadReport& expected);
+
+ private:
+  std::vector<absl::optional<xds::data::orca::v3::OrcaLoadReport>>
```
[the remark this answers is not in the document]
eugeneo: Done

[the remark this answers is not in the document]
eugeneo: Done

test/cpp/interop/backend_metrics_lb_policy.cc
```diff
@@ -0,0 +1,240 @@
...
+using grpc_core::CoreConfiguration;
+using grpc_core::LoadBalancingPolicy;
+using grpc_core::MakeRefCounted;
+using grpc_core::OrphanablePtr;
+using grpc_core::RefCountedPtr;
+
+constexpr absl::string_view kBackendMetricsLbPolicyName =
+    "test_backend_metrics_load_balancer";
+
+const std::string kMetricsTrackerAttribute = "orca_metrics_tracker";
```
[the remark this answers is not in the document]
eugeneo: Done (it lives in the header now)

test/cpp/interop/backend_metrics_lb_policy.cc
eugeneo: Done

test/cpp/interop/backend_metrics_lb_policy.cc
eugeneo: Done, as I understand it... Went to ShutdownLocked, as that's where the delegate is reset.

test/cpp/interop/backend_metrics_lb_policy.cc
eugeneo: Done

eugeneo (commented):
Done, please take another look.

test/cpp/interop/interop_client.h
markdroth: Looks like you made this change below but not here in the argument to the `ServiceStub` ctor.

test/cpp/interop/interop_client.h
markdroth: I didn't mean to remove the whole comment, just the second sentence.

test/cpp/interop/interop_client.h
markdroth: I don't understand what you mean about not wanting to bury the service config flag.  The approach I proposed here would not require changing where the flag is handled; that would still happen in the lambdas that are created in client.cc.

test/cpp/interop/interop_client.h
markdroth: I only see 3 other callers, and they look trivial to change.  Avoiding this doesn't seem to save much time.  And unless you plan to write a subsequent PR to clean this up, this is just leaving unnecessary technical debt.

The down-side of this approach is that any caller that uses the old interface will not get the `LoadReportTracker` channel arg.

eugeneo: Every caller also needs to be retrofitted with the JSON command line arg and such. I understand this is not necessary for current pull request and figuring out all the build dependencies and such delays this pull request.

I removed the second ctor but would add JSON config flag in a separate PR if needed. I feel like some refactoring effort is needed to reduce code duplication.

test/cpp/interop/backend_metrics_lb_policy.h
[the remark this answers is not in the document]
eugeneo: Done

test/cpp/interop/interop_client.h
eugeneo: Done

test/cpp/interop/interop_client.h
eugeneo: Done

test/cpp/interop/backend_metrics_lb_policy.h
```diff
@@ -0,0 +1,54 @@
...
+#include "src/core/lib/config/core_configuration.h"
+#include "src/proto/grpc/testing/xds/v3/orca_load_report.pb.h"
+
+namespace grpc {
+namespace testing {
+class OrcaMetricsProcessor {};
+
+class LoadReportTracker {
+ public:
+  void SetupOnChannel(ChannelArguments* arguments);
```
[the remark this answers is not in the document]
eugeneo: Done

eugeneo (commented):
Thank you for the review. I addressed the comments. PTAL.

test/cpp/interop/backend_metrics_lb_policy.h
```diff
@@ -29,23 +29,19 @@
 namespace grpc {
 namespace testing {
 
-constexpr absl::string_view kMetricsTrackerArgument = "orca_metrics_tracker";
+using xds::data::orca::v3::OrcaLoadReport;
```
markdroth, line 32: The style guide says not to use this kind of "using" statement in a .h file:

https://google.github.io/styleguide/cppguide.html#Namespaces

test/cpp/interop/backend_metrics_lb_policy.h
```diff
@@ -29,23 +29,19 @@
...
-  }
-
+  ChannelArguments GetChannelArguments();
+  void ResetCollectedLoadReports();
   void RecordPerRpcLoadReport(
       const grpc_core::BackendMetricData* backend_metric_data);
-
-  std::deque<absl::optional<xds::data::orca::v3::OrcaLoadReport>>
-  GetCollectedPerRpcLoadReports();
+  absl::StatusOr<absl::optional<OrcaLoadReport>> GetFirstLoadReport();
```
markdroth, line 40: I don't think we need a `StatusOr<>` here, just an `optional<>`.  I realize that makes two levels of optionals, which is a little awkward, but I think you can make it clearer with a typedef:

```
// A load report, or nullopt if the call had no load report.
using LoadReportEntry = absl::optional<xds::data::orca::v3::OrcaLoadReport>;

// Returns the next load report, or nullopt if the queue is empty.
absl::optional<LoadReportEntry> GetNextLoadReport();
```

test/cpp/interop/backend_metrics_lb_policy.cc
```diff
@@ -221,12 +223,26 @@ void LoadReportTracker::RecordPerRpcLoadReport(
...
+absl::StatusOr<absl::optional<OrcaLoadReport>>
+LoadReportTracker::GetFirstLoadReport() {
   absl::MutexLock lock(&per_rpc_load_reports_mu_);
-  std::deque<absl::optional<xds::data::orca::v3::OrcaLoadReport>> result;
-  std::swap(result, per_rpc_load_reports_);
-  return result;
+  if (per_rpc_load_reports_.empty()) {
+    return absl::NotFoundError("No per-call load reports received");
+  }
+  auto report = per_rpc_load_reports_.front();
```
markdroth, line 232: `std::move()`

test/cpp/interop/backend_metrics_lb_policy.cc
```diff
@@ -27,6 +27,8 @@ namespace testing {
 
 namespace {
 
+constexpr absl::string_view kMetricsTrackerArgument = "orca_metrics_tracker";
```
markdroth, line 30: Suggest moving this down next to the other constant definition, on line 38.

test/cpp/interop/client.cc
```diff
@@ -222,17 +222,13 @@ int main(int argc, char** argv) {
...
         if (!service_config_json.empty()) {
-          arguments.SetServiceConfigJSON(std::move(service_config_json));
-        }
-        for (const auto& arg : extra_arguments) {
-          arguments.SetPointer(arg.first, arg.second);
+          arguments.SetServiceConfigJSON(service_config_json);
         }
         return CreateChannelForTestCase(test_case, std::move(factories),
-                                        std::move(arguments));
+                                        arguments);
```
markdroth, line 231: This should still use `std::move()`.

test/cpp/interop/interop_client.h
```diff
@@ -118,12 +111,13 @@ class InteropClient {
 
  private:
   class ServiceStub {
+    typedef std::function<std::shared_ptr<Channel>()> ChannelCreationFunc;
```
markdroth, line 114: This should be public within `ServiceStub`.

test/cpp/interop/interop_client.h
markdroth: I don't think the JSON command line flag is needed in the other binaries, so I don't think there's any need for a follow-on PR here.  This looks great!

markdroth (approved):
This looks really good!

Only minor comments remaining; feel free to merge after addressing.

eugeneo (commented):
Thank you for the review. I will work on cherry-picking.

test/cpp/interop/backend_metrics_lb_policy.h
```diff
@@ -28,20 +28,21 @@
...
   void RecordPerRpcLoadReport(
       const grpc_core::BackendMetricData* backend_metric_data);
-  absl::StatusOr<absl::optional<OrcaLoadReport>> GetFirstLoadReport();
+  // Returns the next load report, or nullopt if the queue is empty.
+  absl::optional<LoadReportEntry> GetNextLoadReport();
 
  private:
-  std::deque<absl::optional<OrcaLoadReport>> per_rpc_load_reports_
-      ABSL_GUARDED_BY(per_rpc_load_reports_mu_);
+  std::deque<absl::optional<xds::data::orca::v3::OrcaLoadReport>>
```
markdroth, line 44: This can be `std::deque<LoadReportEntry>`.

test/cpp/interop/client.cc
eugeneo: clang-tidy does not allow it...
```
test/cpp/interop/client.cc:231:41: error: passing result of std::move() as a const reference argument; no move will actually happen [performance-move-const-arg,-warnings-as-errors]
                                        std::move(arguments));
                                        ^~~~~~~~~~         ~
Suppressed 9286 warnings (9210 in non-user code, 76 NOLINT).
```

markdroth (approved):

test/cpp/interop/client.cc
markdroth: Ah, I didn't realize that `CreateChannelForTestCase()` takes the argument as a const reference.  It might make sense to change it to do so, but the way you have it now is fine too.

markdroth (approved):

stanley-cheung:
@eugeneo Hi, this PR might have broken the observability interop test. Those tests are ready to be run as PR tests yet so there wasn't a signal there. I noticed that the observability tests start failing after this PR is merged. Here's the logs that fails:

https://fusion2.corp.google.com/ci/kokoro/prod:grpc-gcp%2Ftools%2Fobservability%2Fmaster%2Fcontinuous_cpp/activity/f1566fe2-1017-4514-b828-c1c28be2ebf2/invocations/f1566fe2-1017-4514-b828-c1c28be2ebf2/targets/integration%2Fcpp:cpp:test_monitoring_basic/log

The observability tests also uses the interop server/client. I also see you make some changes to `observability_client.cc` as well. 

Do you mind taking a look? Hopefully it's something obvious. If it's not, I can help set up some way for you to reproduce this. It's basically just docker pull an image from `gcr.io`, set some environment variables and run a `docker run` command. 

Thanks.

stanley-cheung:
> It is fairly obvious. I will try to fix it.

Thanks! Let me know if you have a PR, I can run it through.

eugeneo:
@stanley-cheung #32630 is in review now.

stanley-cheung:
Thanks, I just kicked off an observability interop test run. Will report back. Thanks!

eugeneo:
That PR did not build, sorry. Please wait for checks to pass. Thanks!

stanley-cheung:
Ah, cool, will wait

eugeneo:
@stanley-cheung that PR was checked in.

stanley-cheung:
@eugeneo That fixed the issue! Thanks a lot!

PR #32525, opened 2023-03-02
[fork] Log when pthread_create fails

gnossen:
There was evidence that a `pthread_create` failed in [this comment](https://github.com/grpc/grpc/issues/31885#issuecomment-1433917201), but without logging, we couldn't diagnose any further. This PR logs in this case.

### DIFF
--- a/src/core/lib/gprpp/posix/thd.cc
+++ b/src/core/lib/gprpp/posix/thd.cc
@@ -19,10 +19,8 @@
 // Posix implementation for gpr threads.
 
 #include <grpc/support/port_platform.h>
-
-#include <string.h>
-
 #include <grpc/support/time.h>
+#include <string>
 
 #ifdef GPR_POSIX_SYNC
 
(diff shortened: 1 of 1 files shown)


src/core/lib/gprpp/posix/thd.cc
```diff
@@ -106,49 +106,52 @@ class ThreadInternalsPosix : public internal::ThreadInternalsInterface {
...
+          return nullptr;
+        },
+        info);
+    *success = (pthread_create_err == 0);
 
     GPR_ASSERT(pthread_attr_destroy(&attr) == 0);
 
     if (!(*success)) {
+      gpr_log(GPR_ERROR, "pthread_create failed: %s",
+              strerror(pthread_create_err));
```
drfloob, line 154: You'll want to use `StrError` from src/core/lib/gprpp/strerror.cc here. `strerror` is not thread safe.

gnossen: Oh the irony. Fixed.

drfloob (approved):

PR #32531, opened 2023-03-02
GCP Observability testing for C++

stanley-cheung:
Third try for #32466. 

This adds an interop client / server for GCP Observability integration testing.

Everything is new here with no refactor. Plan is to get this in first before trying to refactor out the flags.


stanley-cheung:
cl/513682048 is a trial run of the import and it finally shows no presubmit error now.

yashykt (approved):
Thanks for doing this!

PR #32536, opened 2023-03-03
[fork] Entirely opt EE threads out of ExecCtx counting

gnossen:
Follow-up to https://github.com/grpc/grpc/pull/32229.

https://github.com/grpc/grpc/pull/32229 incremented the `ExecCtx` count unconditionally. It was previously impossible for a thread to exit  `IncExecCtxCount` while `fork_complete_` was `false`. These same threads then went on to _decrement_ `count_` while the fork was still in progress, putting `count_` well below its expected range ([0, 1] while blocking and [2, inf) while not blocking). This resulted in cases where `count_` would be stuck at a negative number with a thread infinitely looping through  `IncExecCtxCount`.

This PR instead opts EE threads out of ExecCtx counting entirely. They handle clean-up of their threads separately through a separate set of handlers registered by an entirely separate invocation of `pthread_atfork`. This resolves the issue pointed out in [this comment](https://github.com/grpc/grpc/issues/31885#issuecomment-1426445192).

### DIFF
--- a/src/core/lib/gprpp/fork.cc
+++ b/src/core/lib/gprpp/fork.cc
@@ -67,7 +67,6 @@ class ExecCtxState {
     // EventEngine is expected to terminate all threads before fork, and so this
     // extra work is unnecessary
     if (grpc_event_engine::experimental::ThreadLocal::IsEventEngineThread()) {
-      gpr_atm_no_barrier_fetch_add(&count_, 1);
       return;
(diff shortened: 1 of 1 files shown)


drfloob (approved):
Thanks! We had discussed this a while ago, I thought we had executed on it already ... still looks right to me.

PR #32544, opened 2023-03-03
TlsCreds: Support revocation of intermediate in chain.

gtcooke94:
This PR is a small code change with a lot of new test data.
[In OpenSSL, there are two flags that configure CRL checks. Coping relevant section:](https://www.openssl.org/docs/man1.0.2/man3/X509_VERIFY_PARAM_get_depth.html)

> - X509_V_FLAG_CRL_CHECK enables CRL checking for the certificate chain leaf certificate. An error occurs if a suitable CRL cannot be found.
> - X509_V_FLAG_CRL_CHECK_ALL enables CRL checking for the entire certificate chain.

We currently only set `X509_V_FLAG_CRL_CHECK`, so we will only ever check if the leaf certificate is revoked. We should check the whole chain. I am open to making this a user configuration if we want to do it that way, but we certainly need to be able to check the whole chain.

So, this PR contains the small code change in `ssl_transport_security.cc` to use the `X509_V_FLAG_CRL_CHECK_ALL` flag.
Then the rest of the changes are in tests. I've added all the necessary files to have a chain built that looks as follows
`Root CA -> Revoked Intermediate CA -> Leaf Certificate`, and added a test for this case as well.
You can verify that on master this new test will fail (i.e. the handshake will succeed even though the intermediate CA is revoked) by checking out this branch, running `git checkout master -- ./src/core/tsi/ssl_transport_security.cc`, then running the test.

I also slightly reorganized test/core/tsi/test_creds/ so that the CRLs are in their own directory, which is the way our API intends to accept CRLs.


test/core/tsi/crl_ssl_transport_security_test.cc
```diff
@@ -185,12 +238,27 @@ class CrlSslTransportSecurityTest
...
+      // 1. CRL Directory with CA's CRL and Intermediate CA's CRL -> Handshake
+      // fails due to revoked cert
+      // 2. CRL Directory with CA's CRL but missing Intermediate CA's CRL ->
+      // Handshake fails due to revoked cert
+      // 3. CRL Directory without CA's CRL with but Intermediate CA's CRL ->
+      // Handshake succeeds because the CRL that revokes the cert is not
+      // present.
       bool expect_server_success =
-          !(use_revoked_server_cert_ || use_revoked_client_cert_);
+          !(use_revoked_server_cert_ || use_revoked_client_cert_ ||
```
rockspore, line 254: I think it'd be cleaner with
`!use_revoked_server_cert_ && !use_revoked_client_cert && !(use_revoked_intermediate_ && !use_missing_root_crl_);`. WDYT? Same below.

nit: I guess `&` for bit operation works well here as well but normally it should be `&&` for logical `and`.

test/core/tsi/test_creds/crl_data/current.crl
```diff
@@ -1,12 +1,15 @@
...
-CAwKU29tZS1TdGF0ZTEhMB8GA1UECgwYSW50ZXJuZXQgV2lkZ2l0cyBQdHkgTHRk
-MQ8wDQYDVQQDDAZ0ZXN0Y2EXDTIyMDIwNDIyMTI1MFoXDTMyMDIwMjIyMTI1MFow
-JzAlAhRSILpknGPXIWdLvLVIY7DlId/IQRcNMjIwMjA0MjIxMTExWqAPMA0wCwYD
-VR0UBAQCAhABMA0GCSqGSIb3DQEBCwUAA4IBAQAZXNfxSjT/EZDTGV71eE0jKKsg
-Ur8TNkRGypZXbV8cQ+YFlqt2Zp+dsWEP2FLsc048QGVe4sRuJrPOm7eSmvgZUHSX
-l1yI2T6si1wxhX2DKIKDZGWWYx6rOyocL9EjhxZjLSeJ43eLxzD6TnGE29cbDLXv
-bs9slsGyc+UZaD9KY9RpeJjQEV0Yh7+iwIG3PBVKNSFR2R2m8XK+Ioc5Z7zsWwZu
-7gCZ/CW/07/SXfFxM6Q8XTRFStIjqb3cuwivc+ig/X+RbpxDygy0SQA/oJXuABIh
-0y4+YxRUQauODK4vxLPeC1m7tHyh1poVY/jnp3tAn3ax6uFbwDz1wbNtTOeT
+MIICUDCCATgCAQEwDQYJKoZIhvcNAQELBQAwVjELMAkGA1UEBhMCQVUxEzARBgNV
```
rockspore, line 2: This file is not exported in BUILD. Is it redundant?

gtcooke94: Yes - `openssl rehash` creates a symlink to this, and the symlink is exported in BUILD

test/core/tsi/crl_ssl_transport_security_test.cc
gtcooke94: I think it's just a personal preference - the way I read it I see something like
"expect server success = NOT any of the revoked bools being true" So if anything bool flag is true in all of those ORs it'll be an expected failure.

IMO I like the way it is a little more because I don't have the think about and negate every individual condition

Overall, I'd rather pull some of this shared logic out and hard code it in each individual test since it's the known expected result, I think it would make it a lot cleaner, but that's a refactor for later.

test/core/tsi/test_creds/crl_data/current.crl
gtcooke94: To explain further, this was already in the repo. The generation script copies it around to different sub folders that hold different CRLs, then I run rehash on each of those folders (it's near the end of `intermediate_gen.sh`)

rockspore: Makes sense. Thanks!

test/core/tsi/crl_ssl_transport_security_test.cc
rockspore: Sure. I don't have a strong opinion either.

rockspore (approved):

markdroth (approved):

Issue #32356, opened 2023-02-11
UnimplementedAsyncRequest constructor races with GenericAsyncRequest causing crashes


snaury:
### What version of gRPC and what language are you using?
The bug seems to be very old, so the exact version does not matter.

### What operating system (Linux, Windows,...) and version?
Linux

### What runtime / compiler are you using (e.g. python version or version of gcc)
We use clang, but the compiler doesn't matter.

### What did you do?
We call cq->Next(&tag, ok) from multiple threads.

### What did you expect to see?
We expect to only ever see values of tags that we provided.

### What did you see instead?
Recently we started to occasionally see our grpc servers crashing right after the call to cq->Next. Debugging showed, that Next returns true, indicating the event is available, however tag turned out to be nullptr. Since we cast it to our interface and call a method on it, this caused crashes when loading a vptr. Further debugging showed that the call to FinalizeResult in CompletionQueue::AsyncNextInternal returns true, but the tag is nullptr. When inspecting core_cq_tag we observed the object is actually UnimplementedAsyncRequest, however UnimplementedAsyncRequest::FinalizeResult should always returns false, so it was puzzling at first how it could ever return true.

Turns out there's a race in UnimplementedAsyncRequest constructor. When it calls to GenericAsyncRequest, it starts the request right there, however the vptr is pointing to GenericAsyncRequest at that time. Later UnimplementedAsyncRequest updates vptr to its vtable, but it looks like the request may already start being handled on another thread, and AsyncNextInternal manages to catch it in a state when vptr is still pointing to GenericAsyncRequest, calling GenericAsyncRequest::FinalizeResult. It returns true, bypassing UnimplementedAsyncRequest override, and it indeed has tag initialized to nullptr by default.

It looks like our users tried using a new grpc service, which was not served by our server yet, this triggered a lot of unimplemented requests, which started triggering the race under heavy load. Workaround of delaying grpc_server_request_call in GenericAsyncRequest until UnimplementedAsyncRequest constructor body stopped crashes for us, confirming that the problem is indeed with UnimplementedAsyncRequest.

snaury:
Btw, I can see that something similar was previously detected by tsan in #17044, but it was marked as stale for some reason?

yijiem:
To ease debugging, is it possible for you to provide a minimal client/server example that repros or might repro this issue?

snaury:
[...]

PR #32547, opened 2023-03-03
Fix a race on vptr for UnimplementedAsyncRequest

yijiem:
It is reported in https://github.com/grpc/grpc/issues/32356 that there is a race on vptr for `UnimplementedAsyncRequest` which would cause crashes for multi-threaded server if clients send unimplemented RPC request to the server.

The cause is that the server requests a call for `UnimplementedAsyncRequest` in its base class `GenericAsyncRequest` when the `vptr` still points to the base class's `vtable`. If the call went in and another server thread picks up the tag before the `vptr` points back to the derived class's `vtable`, it would call the wrong virtual function and also this is a data race. This fix makes the request of the call inside the derived class's constructor.

### DIFF
--- a/include/grpcpp/server_interface.h
+++ b/include/grpcpp/server_interface.h
@@ -301,9 +301,12 @@ class ServerInterface : public internal::CallHook {
                         internal::ServerAsyncStreamingInterface* stream,
                         grpc::CompletionQueue* call_cq,
                         grpc::ServerCompletionQueue* notification_cq, void* tag,
-                        bool delete_on_finalize);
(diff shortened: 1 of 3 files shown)


veblush (approved):

ctiller (approved):
I'm good with this. @markdroth please confirm that we're good with no gRFC for the API change here (I am, but I'd like an explicit ack).

markdroth:
Yeah, I'm okay with this change without a gRFC.  It shouldn't actually break anything for anyone.

yijiem:
Thanks all!

PR #32548, opened 2023-03-04
[test-infra] Enable end2end test experiments to run on Windows

drfloob:
Because bazel runfile symlinks are not created by default on Windows, end2end test experiments had been explicitly disabled on Windows. To get end2end test experiments working on Windows, I had to replace the sh_test with something that would allow the test runner to dynamically look up runfile locations. I chose to use a short py_test script instead.

I've confirmed this works with bazel 4.2.3 and 5.4.0 on a Windows 10 Server.


jtattermusch:
I'm not sure I follow. AFAIK we do run the generated end2end tests on windows RBE
(see e.g. https://source.cloud.google.com/results/invocations/054bac56-6f84-455b-a3fa-ea691f09ba51/targets;query=test%2Fcore%2Fend2end;collapsed=/%2F%2Ftest%2Fcore%2Fend2end:h2_census_test@authority_not_supported/log).
So it's not clear to my what makes "experiments" different. I thought experiments are just another dimension we add to the test generator.

I'd need to understand better what is the problem with experiments on windows. The introduction of `grpc_py_test` seems like an adhoc solution for a problem that might have better (and less intrusive) solutions.
- the new python wrapper is of significant complexity compared to the old wrapper .sh script (which is  one line basically).
- I'm afraid this introduces python dependency to our already-complex bazel C/C++ builds
- ad "bazel runfile symlinks are not created by default on Windows" - is there an bazel option that enables them?

drfloob:
With experiments enabled, the current shell script does not find the test binary. Here is the full Windows RBE result: https://source.cloud.google.com/results/invocations/db102bb8-2af0-4dc5-8b83-c73ffccdfa28/targets
Example error:

```
C:\botcode\...\test\core\end2end\h2_census_test@registered_call@experiment=promise_based_server_call: line 17: test/core/end2end/h2_census_test.exe: Bad address
```

Regarding symlinks:

> ad "bazel runfile symlinks are not created by default on Windows" - is there an bazel option that enables them?

Yes, but enabling symlinks requires developer mode to be enabled, and only on Windows 10+ (v1703 or later). It would be nice to have a solution that did not require special system versions or privileges. https://bazel.build/reference/command-line-reference#flag--windows_enable_symlinks (and `--enable_runfiles`)

Error: 
```
Creating symlinks on Windows requires one of the following:
    1. Bazel is run with administrator privileges.
    2. The system version is Windows 10 Creators Update (1703) or later and developer mode is enabled. 
```

drfloob:
Depends on https://github.com/grpc/grpc/pull/32569

jtattermusch:
Hm, I'm planning to enable developer mode (and long filenames) on the next generation of our windows kokoro image. That work is currently in progress, but it's unclear when it will land (perhaps a week or two from now?).

Would it be difficult to create a .sh version of the logic you need (it doesn't seem complicated and bash is reasonably good at manipulating paths), include it in the already-existing test/core/end2end/run.sh and use it when running on windows?

We could add a TODO to get rid of the hack when symlinks are enabled on windows (which hopefully wouldn't take too long).

drfloob:
I wasn't able to get the bazel shell runfiles library to work on Windows, I
believe it assumes that symlinks are present. That was the first thing I
attempted here. Manually parsing the runfile metadata file will be more
hackish and error prone than this solution. We know that the fix in this PR
is short-lived for a number of reasons, I would recommend we accept the
temporary Python dependency here, especially given that it is already
satisfied in our CI.

drfloob:
*A note on lifetime:* To be clear, this is intended to be a stopgap measure until a better solution is available. @jtattermusch is working on a set of custom kokoro images that will support symlinked runfiles on Windows, which means we can effectively roll this PR back. And there has been work towards rewriting the end2end test framework itself, albeit slowly over the past 6 months or so. When either land, this Python script can likely go away.

veblush (approved):
I approved this to unblock others depending on this as this is a temporary solution to be removed later once new Kokoro Windows image lands. @jtattermusch please chime in if we need some fixes (or even rollback) after this gets merged.

drfloob:
This isn't going to work internally, we'll need to investigate another approach.

PR #32553, opened 2023-03-06
Document using other binaries in Grpc.Tools

tonydnewell:
Added documentation on how to use `Grpc.Tools` MSBuild integration with *unsupported* architectures - i.e. third party built `protoc` and `grpc_csharp_plugin`.

### DIFF
--- a/src/csharp/BUILD-INTEGRATION.md
+++ b/src/csharp/BUILD-INTEGRATION.md
@@ -36,7 +36,7 @@ Support is provided for the following platforms:
 * MacOS (x64 and arm64 via using the x64 binaries)
 * Linux (x86, x64, and arm64)
 
-You can still use the MSBuild integration provided by `Grpc.Tools` for other architectures.
+You may still use the MSBuild integration provided by `Grpc.Tools` for other architectures provided you can supply the codegen binaries for that platform/architecture.
 See [Using Grpc.Tools with unsupported architectures](#unsupported-arch) below.
 
 ## Adding `.proto` files to a project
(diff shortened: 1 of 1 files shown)


tonydnewell:
@jtattermusch for review

src/csharp/BUILD-INTEGRATION.md
```diff
@@ -485,6 +489,42 @@ then in your project file:
...
+For example, there are community provided packages for the gRPC plugins for Alpine Linux using the *musl* C standard library:
+[https://pkgs.alpinelinux.org/packages?name=grpc-plugins](https://pkgs.alpinelinux.org/packages?name=grpc-plugins)
+
+To use these:
+```bash
+# Build or install the binaries for your architecture.
+
+# e.g. for Alpine Linux the grpc-plugins package can be used
+#        See https://pkgs.alpinelinux.org/package/edge/community/x86_64/grpc-plugins
+apk add grpc-plugins  # Alpine Linux specific package installer
```
jtattermusch, line 515: add a note somewhere that when providing your own binaries, you're also responsible they're the right version of protobuf a grpc plugin.
The binaries bundled in Grpc.Tools already ensure that the correct and mutually compatible version of `protoc` and `grpc_csharp_plugin` will be chosen, but when providing them yourself, you're in charge.
- protoc and grpc_csharp_plugin can be mutually incompatible
- the code generated by the wrong version of many not work with your application (e.g. breaks the build).
- specifically, older version of plugins may generate incompatible code or may not contain patches/fixes.

src/csharp/BUILD-INTEGRATION.md
```diff
@@ -36,6 +36,9 @@ Support is provided for the following platforms:
 * MacOS (x64 and arm64 via using the x64 binaries)
 * Linux (x86, x64, and arm64)
 
+You can still use the MSBuild integration provided by `Grpc.Tools` for other architectures.
```
jtattermusch, line 39: + "provided you can supply the codegen binaries for that platform/architecture"

tonydnewell:
@jtattermusch updated from your comments

jtattermusch (approved):

PR #32556, opened 2023-03-07
Improve "ExecCtx::RunList()" with O(1) implementation.

LeiW000:
We don't have to iterate through the list for combining the two lists. They can be combined directly. It can improve the performance.

### DIFF
--- a/src/core/lib/iomgr/exec_ctx.cc
+++ b/src/core/lib/iomgr/exec_ctx.cc
@@ -106,27 +106,18 @@ void ExecCtx::Run(const DebugLocation& location, grpc_closure* closure,
 }
 
 void ExecCtx::RunList(const DebugLocation& location, grpc_closure_list* list) {
-  (void)location;
-  grpc_closure* c = list->head;
-  while (c != nullptr) {
-    grpc_closure* next = c->next_data.next;
-#ifndef NDEBUG
-    if (c->scheduled) {
-      Crash(absl::StrFormat(
-          "Closure already scheduled. (closure: %p, created: [%s:%d], "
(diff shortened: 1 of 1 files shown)


LeiW000:
Is there anyone, who can help to review the PR?

LeiW000:
Is there anything else I need to do?

LeiW000:
@yashykt , can the PR be merged?

LeiW000:
Can anyone help to take look at the PR?

yashykt:
Sorry for the delay! I will take a look at this PR this week.
@ctiller would appreciate another pair of eyes here.

src/core/lib/iomgr/exec_ctx.cc
```diff
@@ -106,27 +106,18 @@ void ExecCtx::Run(const DebugLocation& location, grpc_closure* closure,
...
+
+  if (list->head != nullptr) {
+    if (destlist->head == nullptr) {
+      destlist->head = list->head;
+      destlist->tail = list->tail;
+    } else {
+      destlist->tail->next_data.next = list->head;
+      destlist->tail = list->tail;
     }
-    c->scheduled = true;
```
yashykt, line 121: I think the approach probably works but you you are no longer setting `scheduled` any more

yashykt: nor are the other fields being set

LeiW000: The fields are defined in "#ifndef NDEBUG" for debugging. I'm wondering if we can ignore them here.

LeiW000:
@yashykt @ctiller , thanks for your reviews. I decide to close the PR as I can't find a good way to make the change while keeping the debug code.

PR #32559, opened 2023-03-07
Switch linux RBE to ubuntu18.04, get rid of rbe_autoconfig

jtattermusch:
First step in the modernization of our RBE stack (see go/rbe-tech-debt-notes).

- Get rid of the deprecated rbe_autoconfig and start using  [rbe_configs_gen](https://github.com/bazelbuild/bazel-toolchains#rbe_configs_gen---cli-tool-to-generate-configs) + check in the generated toolchain configs.
- Switch from marketplace.gcr.io/google/rbe-ubuntu16-04 to marketplace.gcr.io/google/rbe-ubuntu18-04 (this image is still not owned by us, but at least it's newer and demonstrates how a switch to a newer docker image is done).
- provide script for generating the linux RBE toolchain configs.
- cleanup RBE configuration in the bazelrc files used for remote build


jtattermusch:
There seem to be some MSAN failures.
https://source.cloud.google.com/results/invocations/ba10e80e-f16d-4d3d-b65b-117ff7d83bf1/targets

@veblush  can you take a quick look and let me know whether you think those are genuine MSAN failures reporting problems (e.g. because we upgraded from clang-11 to clang-12) or whether those are red herrings?

veblush (approved):

gnossen (approved):

jtattermusch:
@drfloob   FYI so that you are aware of this PR in case it has any unexpected consequences for https://github.com/grpc/grpc/pull/32548 (it should not).

PR #32561, opened 2023-03-07
Add random early rejection for metadata

ananda1066:
Add new channel arg `GRPC_ARG_ABSOLUTE_MAX_METADATA_SIZE` as hard limit for metadata. Change `GRPC_ARG_MAX_METADATA_SIZE` to be a soft limit. Behavior is as follows:

**Hard limit**
(1) if hard limit is explicitly set, this will be used.
(2) if hard limit is not explicitly set, maximum of default and soft limit * 1.25 (if soft limit is set) will be used.

**Soft limit**
(1) if soft limit is explicitly set, this will be used.
(2) if soft limit is not explicitly set, maximum of default and hard limit * 0.8 (if hard limit is set) will be used.


src/core/ext/transport/chttp2/transport/http2_settings.h
```diff
@@ -34,10 +34,11 @@ typedef enum {
   GRPC_CHTTP2_SETTINGS_MAX_HEADER_LIST_SIZE = 5,             // wire id 6
   GRPC_CHTTP2_SETTINGS_GRPC_ALLOW_TRUE_BINARY_METADATA = 6,  // wire id 65027
   GRPC_CHTTP2_SETTINGS_GRPC_PREFERRED_RECEIVE_CRYPTO_FRAME_SIZE =
-      7,  // wire id 65028
+      7,                                                      // wire id 65028
+  GRPC_CHTTP2_SETTINGS_GRPC_HEADER_LIST_SIZE_SOFT_LIMIT = 8,  // write id 65029
```
ctiller, line 38: We don't need to publish this as a HTTP2 setting...

Instead we should put the hard limit as the one we advertise as a HTTP2 setting and keep the soft setting private, per the http2 rfc:

```
SETTINGS_MAX_HEADER_LIST_SIZE (0x06):
This advisory setting informs a peer of the maximum field section size that the sender is prepared to accept, in units of octets. The value is based on the uncompressed size of field lines, including the length of the name and value in units of octets plus an overhead of 32 octets for each field line.

For any given request, a lower limit than what is advertised MAY be enforced. The initial value of this setting is unlimited.
```

ctiller (changes requested):
Looking really good... we should drop the settings publishing (it's unnecessary) and then this should be good to go.

ananda1066:
Rebasing seems to have messed with things - going to close this and try again.

PR #32574, opened 2023-03-09
[EventEngine] Fix PosixEventEngine IPv4 support

drfloob:
Connections cannot be made in IPv4-only environments. To test, hard-code `IsIpv6LoopbackAvailable` to return false.

Example Error:
`
D0309 00:29:49.514359445     235 tcp_client.cc:67]  (event_engine) EventEngine::Connect Status: INTERNAL: socket: Address family not supported by protocol
`

This can also be reproduced in gRPC's benchmark environment, which does not have IPv6 enabled.

### DIFF
--- a/src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
+++ b/src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
@@ -620,9 +620,9 @@ absl::Status PosixSocketWrapper::ApplySocketMutatorInOptions(
   return SetSocketMutator(usage, options.socket_mutator);
 }
 
-bool PosixSocketWrapper::SetSocketDualStack() {
+bool PosixSocketWrapper::SetSocketDualStack(int fd) {
   const int off = 0;
-  return 0 == setsockopt(fd_, IPPROTO_IPV6, IPV6_V6ONLY, &off, sizeof(off));
+  return 0 == setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, &off, sizeof(off));
 }
 
 bool PosixSocketWrapper::IsIpv6LoopbackAvailable() {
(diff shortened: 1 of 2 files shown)


src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
```diff
@@ -620,9 +620,9 @@ absl::Status PosixSocketWrapper::ApplySocketMutatorInOptions(
   return SetSocketMutator(usage, options.socket_mutator);
 }
 
-bool PosixSocketWrapper::SetSocketDualStack() {
+bool PosixSocketWrapper::SetSocketDualStack(int fd) {
   const int off = 0;
-  return 0 == setsockopt(fd_, IPPROTO_IPV6, IPV6_V6ONLY, &off, sizeof(off));
+  return 0 == setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, &off, sizeof(off));
```
yashykt, line 625: what happens to the old socket fd_?

yashykt (approved):

src/core/lib/event_engine/posix_engine/tcp_socket_utils.cc
```diff
@@ -620,9 +620,9 @@ absl::Status PosixSocketWrapper::ApplySocketMutatorInOptions(
   return SetSocketMutator(usage, options.socket_mutator);
 }
 
-bool PosixSocketWrapper::SetSocketDualStack() {
+bool PosixSocketWrapper::SetSocketDualStack(int fd) {
```
yashykt, line 623: this could just be in an anonymous namespace

drfloob: Very true, done.

yashykt (approved):

PR #32583, opened 2023-03-09
make_grpcio_tools.py improvements: more readable diffs

jtattermusch:
- sort source files to ensure stable ordering
- generate one source file per line

together this should produce diffs that are much more readable by humans when sources get added/removed to/from protobuf (and make_grpcio_tools.py is used to regenerate).


tools/distrib/python/make_grpcio_tools.py
```diff
@@ -125,8 +127,8 @@ def get_deps():
     ]
     commit_hash = protobuf_submodule_commit_hash()
     deps_file_content = DEPS_FILE_CONTENT.format(
-        cc_files=cc_files,
-        proto_files=proto_files,
+        cc_files=json.dumps(sorted(cc_files), indent=4),
```
gnossen, line 130: Nit: [`pprint`](https://docs.python.org/3/library/pprint.html) is guaranteed to produce valid Python. This should be okay for lists, but if there's a bug before this, the error could be a bit confusing.

gnossen (approved):

tools/distrib/python/make_grpcio_tools.py
jtattermusch: Thanks for the tip. Done (but needed an adjustment to end up with the format I wanted).

PR #32585, opened 2023-03-09
[testing]: remove server-side check on number of concurrent RPCs in alts_concurrent_connectivity_test

apolcyn:
This check only works if all handshake RPCs have an OK status, and it's racey e.g. if the client is cancelling handshake RPCs (being when an RPC is cancelled, termination of the RPC at the client is asynchronous from termination at the server, so the client can resume the queue before the server RPC completes).

### DIFF
--- a/test/core/tsi/alts/fake_handshaker/fake_handshaker_server.cc
+++ b/test/core/tsi/alts/fake_handshaker/fake_handshaker_server.cc
@@ -58,7 +58,7 @@ namespace gcp {
 // It is thread-safe.
 class FakeHandshakerService : public HandshakerService::Service {
  public:
-  FakeHandshakerService(const std::string& peer_identity)
+  explicit FakeHandshakerService(const std::string& peer_identity)
       : peer_identity_(peer_identity) {}
 
   Status DoHandshake(
--- a/test/core/tsi/alts/fake_handshaker/fake_handshaker_server.cc
+++ b/test/core/tsi/alts/fake_handshaker/fake_handshaker_server.cc
@@ -251,8 +251,8 @@ class FakeHandshakerService : public HandshakerService::Service {
 
 std::unique_ptr<grpc::Service> CreateFakeHandshakerService(
     const std::string& peer_identity) {
-  return std::unique_ptr<grpc::Service>{new grpc::gcp::FakeHandshakerService(
-      peer_identity)};
+  return std::unique_ptr<grpc::Service>{
+      new grpc::gcp::FakeHandshakerService(peer_identity)};
 }
 
 }  // namespace gcp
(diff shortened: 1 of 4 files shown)


matthewstevenson88 (approved):
Thanks for the quick fix Alex!

test/core/tsi/alts/fake_handshaker/fake_handshaker_server.cc
```diff
@@ -58,15 +58,12 @@ namespace gcp {
 // It is thread-safe.
 class FakeHandshakerService : public HandshakerService::Service {
  public:
-  FakeHandshakerService(int expected_max_concurrent_rpcs,
-                        const std::string& peer_identity)
-      : expected_max_concurrent_rpcs_(expected_max_concurrent_rpcs),
-        peer_identity_(peer_identity) {}
+  FakeHandshakerService(const std::string& peer_identity)
```
markdroth, line 61: Single-arg ctors need to be `explicit`.

https://google.github.io/styleguide/cppguide.html#Implicit_Conversions

test/core/tsi/alts/handshaker/alts_concurrent_connectivity_test.cc
```diff
@@ -104,17 +104,10 @@ grpc_channel* create_secure_channel_for_test(
 
 class FakeHandshakeServer {
  public:
-  explicit FakeHandshakeServer(bool check_num_concurrent_rpcs) {
+  explicit FakeHandshakeServer() {
```
markdroth, line 107: No need for `explicit` if no args.

test/core/tsi/alts/handshaker/alts_concurrent_connectivity_test.cc
```diff
@@ -139,8 +132,7 @@ class FakeHandshakeServer {
 
 class TestServer {
  public:
-  explicit TestServer()
-      : fake_handshake_server_(true /* check num concurrent rpcs */) {
+  explicit TestServer() {
```
markdroth, line 135: No need for `explicit`.

markdroth (approved):
Thanks for fixing this!

markdroth:
Also looks like there's a build failure due to an unused constant.

test/core/tsi/alts/fake_handshaker/fake_handshaker_server.cc
apolcyn: done

test/core/tsi/alts/handshaker/alts_concurrent_connectivity_test.cc
apolcyn: done

test/core/tsi/alts/handshaker/alts_concurrent_connectivity_test.cc
apolcyn: done

apolcyn:
> Also looks like there's a build failure due to an unused constant.

Fixed, thanks!

markdroth (approved):
Looks great!

PR #32600, opened 2023-03-11
Add random early rejection for metadata

ananda1066:
(hopefully last try)

Add new channel arg GRPC_ARG_ABSOLUTE_MAX_METADATA_SIZE as hard limit for metadata. Change GRPC_ARG_MAX_METADATA_SIZE to be a soft limit. Behavior is as follows:

Hard limit
(1) if hard limit is explicitly set, this will be used.
(2) if hard limit is not explicitly set, maximum of default and soft limit * 1.25 (if soft limit is set) will be used.

Soft limit
(1) if soft limit is explicitly set, this will be used.
(2) if soft limit is not explicitly set, maximum of default and hard limit * 0.8 (if hard limit is set) will be used.

Requests between soft and hard limit will be rejected randomly, requests above hard limit will be rejected.


src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -375,6 +376,21 @@ static void read_channel_args(grpc_chttp2_transport* t,
                 .GetObjectRef<grpc_core::channelz::SocketNode::Security>());
   }
 
+  const int soft_limit =
+      channel_args.GetInt(GRPC_ARG_MAX_METADATA_SIZE).value_or(-1);
+  if (soft_limit == -1) {
```
ctiller, line 381: maybe make this `<=0`

src/core/ext/transport/chttp2/transport/hpack_parser.h
```diff
@@ -121,7 +124,7 @@ class HPackParser {
   uint8_t dynamic_table_updates_allowed_;
   // Length of frame so far.
   uint32_t frame_length_;
-  uint32_t metadata_size_limit_;
+  std::unique_ptr<RandomEarlyDetection> metadata_early_detection_ = nullptr;
```
ctiller, line 127: I don't think this needs a `unique_ptr`... let's just inline the type here.

ctiller: (like... `RandomEarlyDetection metadata_early_detection_;`)

ctiller (changes requested):
looks good... a couple of things we should take care of and then I think this is gtg

src/core/ext/transport/chttp2/transport/hpack_parser.h
ananda1066: Done - I hadn't done this because there was no default constructor for RandomEarlyDetection, but I've added one now.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
ananda1066: Hm I'm worried about consistency then - if the value is 0, that means the user set the soft limit to that. I think we should treat any user-set value the same way; otherwise, they would see different behavior for a 0 vs. some other value. The user can also set the hard limit to 0 right now, so whatever we decide, we should do for both.

ctiller (approved):

test/core/transport/chttp2/hpack_parser_fuzzer_test.cc
```diff
@@ -69,12 +69,16 @@ DEFINE_PROTO_FUZZER(const hpack_parser_fuzzer::Msg& msg) {
         priority = grpc_core::HPackParser::Priority::Included;
       }
       int max_length = 1024;
+      int absolute_max_length = 1024;
       if (frame.max_metadata_length() != 0) {
         max_length = frame.max_metadata_length();
       }
+      if (frame.absolute_max_metadata_length() != 0) {
+        absolute_max_length = frame.absolute_max_metadata_length();
+      }
```
ctiller, line 78: maybe:
```
if (absolute_max_length < max_length) std::swap(absolute_max_length, max_length);
```

ananda1066: sure done

PR #32607, opened 2023-03-13
[time] Fix startup stall in Linux, using the same approach as #31844.

dbcode:
With this change, gpr_now(GPR_CLOCK_MONOTONIC) adds 5 seconds of padding to the returned timestamp on Linux as well as iOS. This makes gRPC start quickly on recently started machines (e.g. freshly booted micro-VMs).

### DIFF
--- a/src/core/lib/gpr/posix/time.cc
+++ b/src/core/lib/gpr/posix/time.cc
@@ -80,6 +80,11 @@ static gpr_timespec now_impl(gpr_clock_type clock_type) {
 #else
     clock_gettime(clockid_for_gpr_clock[clock_type], &now);
 #endif
+    if (clock_type == GPR_CLOCK_MONOTONIC) {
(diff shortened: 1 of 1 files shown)


dbcode:
Note: I created this PR but had not joined the google organization at that time.  I've now joined the google organization, but the warning above about EasyCLA hasn't gone away. Let me know if I should abandon this PR and make a new one, or if there's another way to resolve that warning. I tried the "Please click here to be authorized", but that didn't seem to work.

yashykt:
if it's indeed an undefined epoch and there's no hyrum law effect going on over here, this looks fine to me..

ctiller (approved):

dbcode:
Hi - is the LABEL issue above something I am supposed to address?

Just looking for anything I can do to help get this PR merged :)

PR #32610, opened 2023-03-13
[EventEngine] RunAfter migration: RetryFilter

yijiem:


src/core/ext/filters/client_channel/retry_filter.cc
```diff
@@ -548,8 +551,9 @@ class RetryFilter::CallData {
   // If server_pushback is nullopt, retry_backoff_ is used.
   void StartRetryTimer(absl::optional<Duration> server_pushback);
 
-  static void OnRetryTimer(void* arg, grpc_error_handle error);
-  static void OnRetryTimerLocked(void* arg, grpc_error_handle error);
+  void OnRetryTimer();
+  static void OnRetryTimerLocked(void* arg,
+                                 GRPC_UNUSED grpc_error_handle error);
```
drfloob, line 556: `GRPC_UNUSED` isn't portable, right? I suspect that MSVC builds might complain if unused variable warnings were enabled.

https://github.com/grpc/grpc/blob/43c3d1f932da02c334e71dc5a9cd998755be5ff7/include/grpc/support/port_platform.h#L589-L595

drfloob (commented):
This looks pretty good! As long we're certain the handles are only ever created and accessed when the combiner lock is held, looks right to me.

markdroth (approved):
This looks good to me, modulo one minor nit.

src/core/ext/filters/client_channel/retry_filter.cc
```diff
@@ -1241,68 +1244,67 @@ void RetryFilter::CallData::CallAttempt::Abandon() {
...
+  GPR_DEBUG_ASSERT(error.ok());
   auto* call_attempt = static_cast<CallAttempt*>(arg);
   auto* calld = call_attempt->calld_;
   if (GRPC_TRACE_FLAG_ENABLED(grpc_retry_trace)) {
     gpr_log(GPR_INFO,
             "chand=%p calld=%p attempt=%p: perAttemptRecvTimeout timer fired: "
-            "error=%s, per_attempt_recv_timer_pending_=%d",
-            calld->chand_, calld, call_attempt, StatusToString(error).c_str(),
-            call_attempt->per_attempt_recv_timer_pending_);
+            "error=%s",
```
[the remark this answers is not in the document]
yijiem: Done.

src/core/ext/filters/client_channel/retry_filter.cc
yijiem: Thanks for catching it! `GRPC_UNUSED` is not portable and I couldn't find an equivalent for MSVC easily. There is the `[[maybe_unused]]` C++ attribute but it needs C++17.

So I went with this for now (based on our discussion):
```
GPR_DEBUG_ASSERT(error.ok());
#ifdef NDEBUG
  (void)error;
#endif
```

src/core/ext/filters/client_channel/retry_filter.cc
```diff
@@ -442,18 +445,18 @@ class RetryFilter::CallData {
...
 
     CallData* calld_;
     AttemptDispatchController attempt_dispatch_controller_;
     OrphanablePtr<ClientChannel::FilterBasedLoadBalancedCall> lb_call_;
     bool lb_call_committed_ = false;
 
-    grpc_timer per_attempt_recv_timer_;
     grpc_closure on_per_attempt_recv_timer_;
-    bool per_attempt_recv_timer_pending_ = false;
+    absl::optional<EventEngine::TaskHandle> per_attempt_recv_timer_handle_;
```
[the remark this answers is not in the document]
yijiem: Ack; thanks for checking!

src/core/ext/filters/client_channel/retry_filter.cc
markdroth: The way we normally deal with this is to just comment out the parameter name, like this:

```
static void OnRetryTimerLocked(void* arg, grpc_error_handle /*error*/);
```

yijiem: In this case the argument is used only in debug build with `GPR_DEBUG_ASSERT`. Actually I should remove `GRPC_UNUSED` now since it is used in both debug and release build, just differently.

markdroth: Alternatively, consider removing the `GPR_DEBUG_ASSERT()`.  I'm not sure it's actually adding any value, since we're only ever scheduling this closure with OK status.

yijiem: Got it, done! Thanks for the suggestion!

PR #32618, opened 2023-03-14
OpenCensus: Use new CallTracer interfaces

yashykt:
This change mostly aims to get OpenCensus to use the new ServerCallTracer interface. Note that the interfaces nor the code are in their final states. There are a bunch of moving pieces, but I thought this might be a nice mid-step to check-in and make sure that our internal traces can also work with these changes.

Overall changes -
1) call_tracer.h shows what the hierarchy of new call tracer interfaces looks like. Open to renaming suggestions.
2) Moved most of the common interface between `CallAttemptTracer` and `ServerCallTracer` into a common `CallTracerInterface`. We should be able to eventually move `RecordReceivedTrailingMetadata` and `RecordEnd` as well to these common interfaces, but it requires some additional work.
3) The compression filter is now responsible for recording the recv and send messages for both the subchannel call and the server, and adds in ability to record compressed and decompressed messages as well.
4) The OpenCensus server filter now uses the new `ServerCallTracer` interface, and so doesn't need to be a filter anymore.
5) A new ServerCallTracerFilter was added. Ideally, we should be able to move it to the current connected filter, but it is in a bit of an interesting state right now, so I would prefer making those changes in a separate PR with Craig's eyes on it.
6) A new context element `GRPC_CONTEXT_CALL_TRACER_ANNOTATION_INTERFACE` was created that replaces the old `GRPC_CONTEXT_CALL_TRACER`, and the new `GRPC_CONTEXT_CALL_TRACER` is mainly to pass the `CallAttemptTracer` down the stack. This should go away in the new promise-based world.

- - -
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/32618)


src/core/plugin_registry/grpc_plugin_registry.cc
```diff
@@ -109,6 +110,7 @@ void BuildCoreConfiguration(CoreConfiguration::Builder* builder) {
 #ifndef GRPC_NO_RLS
   RegisterRlsLbPolicy(builder);
 #endif  // !GRPC_NO_RLS
+  RegisterServerCallTracerFilter(builder);
```
markdroth, line 113: Is this positioning ensuring that this filter will be the last one in the stack, right before the connected_channel filter?

src/cpp/ext/filters/census/grpc_plugin.cc
```diff
@@ -51,13 +54,6 @@ void RegisterOpenCensusPlugin() {
                   &grpc::internal::OpenCensusClientFilter::kFilter);
               return true;
             });
-        builder->channel_init()->RegisterStage(
```
markdroth, line 54: Why do we no longer need the server side to be a filter but we still need the client side to be a filter?

src/cpp/ext/filters/census/server_filter.h
markdroth, line 1: This file should be renamed, since it's no longer a filter.

Same for the .cc file.

test/core/xds/xds_channel_stack_modifier_test.cc
```diff
@@ -111,44 +110,6 @@ TEST(XdsChannelStackModifierTest, XdsHttpFiltersInsertion) {
   grpc_shutdown();
 }
 
-// Test filters insertion with OpenCensus plugin registered
```
markdroth, line 114: Shouldn't we still test this with "census_server", since the code is still checking for that case?

markdroth (commented):

yashykt (commented):
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/32618)* status: 0 of 33 files reviewed, 7 unresolved discussions (waiting on @ctiller, @markdroth, and @veblush)

___
*[`src/core/lib/channel/call_tracer.h` line 60 at r1](https://reviewable.io/reviews/grpc/grpc/32618#-NQaT_aUz5XJysFB59ry-r1-60:-NQaUNoF1cmwiQF4JWeB:b8do4i0) ([raw file](https://github.com/grpc/grpc/blob/3461c11972b7b8f7eccf2cc8736ffc8835e71160/src/core/lib/channel/call_tracer.h#L60)):*

I am raising the following idea just to make sure that we've considered alternatives -- my guess is that we don't actually want to do this.

Since we're now going to be using this same interface on both the client and server side, would it make sense to phrase these methods in terms of client and server rather than send and recv?  For example, we could have a method like `RecordClientInitialMetadata()`, which would be called for send_initial_metadata on the client side and recv_initial_metadata on the server side.

The motivation for this idea is that a lot of the promise code is phrased this way, so I'm wondering if this approach would be more consistent in the long run.

When evaluating this idea, please consider both (a) how the tracer implementations will work and (b) how the `RpcTracerInterface` API will be called in various parts of the code (e.g., filters, transport, etc).  The overall goal here should be minimizing `if (is_client)` throughout the code base.  (Looking at the compression filter, it looks like the send/recv pattern is actually better than the client/server pattern would be.  But I don't know if that's also the case for the other call sites that we'll have in the future.)

One other factor to consider is that I think there are cases where the mutability of the parameters will want to be different on the client and server side.  For example, client initial metadata needs to be mutable on the client side but can be const on the server side, whereas server trailing(?) metadata needs to be mutable on the server side but can be const on the client side.  If we use the same interface for both sides, then we need to make the parameter mutable even on the side where we don't really want to support mutation.  However, if we stick with the current send/recv interface, then I think we can make the parameters mutable only for the send operations.

I'd like to get @ctiller's input on this as well.
</blockquote></details>

I did consider this, and had similar considerations as you - 
1) For example for compression/decompression, we only need to care about whether we are receiving and need to decompress, or whether we are sending and need to decompress.
2) As you suggest, the mutability of parameters is also a good point. We wouldn't need to modify metadata on the receive side, just on the send side.
___
*[`src/core/lib/channel/call_tracer.h` line 95 at r1](https://reviewable.io/reviews/grpc/grpc/32618#-NQaT_aUz5XJysFB59ry-r1-95:-NQaVC4B7DNvosW9KAOC:bfxhcdk) ([raw file](https://github.com/grpc/grpc/blob/3461c11972b7b8f7eccf2cc8736ffc8835e71160/src/core/lib/channel/call_tracer.h#L95)):*

What additional work is required to move these last two methods to `RpcTracerInterface`?
</blockquote></details>

The first change would be to split the final status and stats from `RecordReceivedTrailingMetadata` and move it to a separate call (maybe potentially move it to `RecordEnd`). `RecordEnd` would also need to change accordingly. This part seems potentially tricky enough that I would want a dedicated PR that has reviewers just look at that change.
___
*[`src/core/lib/channel/context.h` line 45 at r1](https://reviewable.io/reviews/grpc/grpc/32618#-NQaT_aUz5XJysFB59rz-r1-45:-NQaXjAuCs73J9GFC_ZZ:b-sh1mqz) ([raw file](https://github.com/grpc/grpc/blob/3461c11972b7b8f7eccf2cc8736ffc8835e71160/src/core/lib/channel/context.h#L45)):*

Suggest renaming `GRPC_CONTEXT_CALL_TRACER` to `GRPC_CONTEXT_CALL_ATTRIBUTE_TRACER` and then using `GRPC_CONTEXT_CALL_TRACER` for this new one.
</blockquote></details>

What does `GRPC_CONTEXT_CALL_ATTRIBUTE_TRACER` even mean? :)
This is messy since with these changes GRPC_CONTEXT_CALL_ATTRIBUTE_TRACER would hold a  client-side`CallTracer`, while a `GRPC_CONTEXT_CALL_TRACER` would now hold a `CallAttemptTracer`
___
*[`src/core/lib/surface/call.cc` line 749 at r1](https://reviewable.io/reviews/grpc/grpc/32618#-NQaT_aUz5XJysFB59s0-r1-749:-NQaYpbl74Mh5BNCDfIh:b-jzyr5x) ([raw file](https://github.com/grpc/grpc/blob/3461c11972b7b8f7eccf2cc8736ffc8835e71160/src/core/lib/surface/call.cc#L749)):*

I think we decided that we want the tracer to be created in the AcceptStream callback, right?  Can't we do it there immediately?
</blockquote></details>

No, since we need the arena to create the `ServerCallTracer`, we said we'll have AcceptStream return the `ServerCallTracer` object to the transport (not yet done).
___
*[`src/core/plugin_registry/grpc_plugin_registry.cc` line 113 at r1](https://reviewable.io/reviews/grpc/grpc/32618#-NQaT_aUz5XJysFB59s1-r1-113:-NQaZ6D29IzzJjjh0gbx:bnp90j3) ([raw file](https://github.com/grpc/grpc/blob/3461c11972b7b8f7eccf2cc8736ffc8835e71160/src/core/plugin_registry/grpc_plugin_registry.cc#L113)):*

Is this positioning ensuring that this filter will be the last one in the stack, right before the connected_channel filter?
</blockquote></details>

That's a good point. Hmm, I got a smart idea, I can actually register it along with the "connected" filter for now as a built_in filter.

I still want to avoid having this filter and move it to the connected filter, but the connected filter seems to be in a messy state right now, so this might still be the right thing to do for now.

yashykt (commented):
> *[Reviewable](https://reviewable.io/reviews/grpc/grpc/32618)* status: 0 of 33 files reviewed, 7 unresolved discussions (waiting on @ctiller, @markdroth, and @veblush)

___
*[`src/core/ext/filters/client_channel/client_channel.cc` line 2529 at r1](https://reviewable.io/reviews/grpc/grpc/32618#-NQaT_aUz5XJysFB59rs-r1-2529:-NQagL-G-wkmz-vGI0Nr:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/3461c11972b7b8f7eccf2cc8736ffc8835e71160/src/core/ext/filters/client_channel/client_channel.cc#L2529)):*

Now that we're storing this in the call context, I don't think we need this data member anymore.  Instead, we can get it from `call_context()[GRPC_CONTEXT_RPC_TRACER]`.
</blockquote></details>

Done.
___
*[`src/core/lib/channel/call_tracer.h` line 54 at r1](https://reviewable.io/reviews/grpc/grpc/32618#-NQaT_aUz5XJysFB59ry-r1-54:-NQagLlB1Uhkv_XDhKkg:b-896fix) ([raw file](https://github.com/grpc/grpc/blob/3461c11972b7b8f7eccf2cc8736ffc8835e71160/src/core/lib/channel/call_tracer.h#L54)):*

I suggest renaming `CallTracerInterface` to something like `CallTracerAnnotationInterface` and then renaming `RpcTracerInterface` to `CallTracerInterface`.
</blockquote></details>

Done.

src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -3046,25 +3037,12 @@ void ClientChannel::FilterBasedLoadBalancedCall::RecvInitialMetadataReady(
...
     // recv_initial_metadata_flags is not populated for clients
     self->call_attempt_tracer()->RecordReceivedInitialMetadata(
-        self->recv_initial_metadata_, 0 /* recv_initial_metadata_flags */);
+        self->recv_initial_metadata_);
   }
   Closure::Run(DEBUG_LOCATION, self->original_recv_initial_metadata_ready_,
                error);
 }
 
-void ClientChannel::FilterBasedLoadBalancedCall::RecvMessageReady(
```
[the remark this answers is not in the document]
yashykt: Thanks! Removed!

src/core/lib/channel/call_tracer.h
```diff
@@ -35,48 +33,74 @@
...
+ public:
+  virtual ~CallTracerInterface() {}
+  // Records an annotation on the call attempt.
+  // TODO(yashykt): If needed, extend this to attach attributes with
+  // annotations.
+  virtual void RecordAnnotation(absl::string_view annotation) = 0;
+};
+
+// The base class for CallAttemptTracer and ServerCallTracer.
+// TODO(yashykt): What's a better name for this?
```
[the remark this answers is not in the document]
yashykt: (please view discussion on reviewable)

src/core/lib/channel/call_tracer.h
```diff
@@ -35,48 +33,74 @@
...
+class RpcTracerInterface : public CallTracerInterface {
+ public:
+  ~RpcTracerInterface() override {}
+  // Please refer to `grpc_transport_stream_op_batch_payload` for details on
+  // arguments.
+  virtual void RecordSendInitialMetadata(
```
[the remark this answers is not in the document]
yashykt: (please view discussion on reviewable)

src/core/lib/channel/call_tracer.h
```diff
@@ -35,48 +33,74 @@
...
-    // methods should only be invoked when the metadata/message was
-    // successfully received, i.e., without any error.
-    virtual void RecordReceivedInitialMetadata(
-        grpc_metadata_batch* recv_initial_metadata, uint32_t flags) = 0;
-    virtual void RecordReceivedMessage(const SliceBuffer& recv_message) = 0;
+    ~CallAttemptTracer() override {}
     // If the call was cancelled before the recv_trailing_metadata op
     // was started, recv_trailing_metadata and transport_stream_stats
     // will be null.
     virtual void RecordReceivedTrailingMetadata(
```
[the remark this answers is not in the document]
yashykt: (please view discussion on reviewable)

src/core/lib/channel/context.h
```diff
@@ -36,9 +36,16 @@ typedef enum {
...
   GRPC_CONTEXT_TRACING,
 
-  /// Value is a CallTracer object.
+  /// Value is a CallTracer object. (Only valid on the client-side call, or
+  /// ServerCallTracer on the server-side.)
   GRPC_CONTEXT_CALL_TRACER,
 
+  /// Value is a RpcTracerInterface (ServerCallTracer on the server-side,
+  /// CallAttemptTracer on a subchannel call.)
+  /// TODO(yashykt): Maybe come up with a better name. This will go away in the
```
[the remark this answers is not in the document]
yashykt: Done

src/core/lib/surface/call.cc
```diff
@@ -743,6 +744,24 @@ grpc_error_handle FilterStackCall::Create(grpc_call_create_args* args,
     global_stats().IncrementServerCallsCreated();
     call->final_op_.server.cancelled = nullptr;
     call->final_op_.server.core_server = args->server;
+    // TODO(yashykt): In the future, we want to also enable stats and trace
+    // collecting from when the call is created at the transport. The idea is
+    // that the transport would create the call tracer and pass it in as part of
```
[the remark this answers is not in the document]
yashykt: (please view discussion on reviewable)

src/cpp/ext/filters/census/grpc_plugin.cc
yashykt: As discussed in the design review, the method of creating call tracers are going to be different on the client-side and the server-side.
On the client-side, we are going to have an interceptor inject this. (I know it's a filter right now, but it'll move to an interceptor). On the server-side, we will have a factory that gets invoked by the surface call.

test/core/xds/xds_channel_stack_modifier_test.cc
yashykt: This test just checks for OpenCensus unfortunately :( For the internal census, we will need a test in google3.

src/core/lib/channel/server_call_tracer.h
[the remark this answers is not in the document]
yashykt: Done

src/core/plugin_registry/grpc_plugin_registry.cc
yashykt: (please view discussion on reviewable)

src/cpp/ext/filters/census/server_filter.h
yashykt: Done

src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -2525,7 +2527,7 @@ ClientChannel::LoadBalancedCall::LoadBalancedCall(
       chand_(chand),
       call_dispatch_controller_(call_dispatch_controller),
       call_attempt_tracer_(
```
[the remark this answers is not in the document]
yashykt: Done.

test/core/xds/xds_channel_stack_modifier_test.cc
yashykt: cl/516905576 moves it to an internal test

src/core/lib/channel/call_tracer.h
```diff
@@ -80,13 +84,13 @@ class RpcTracerInterface : public CallTracerInterface {
 // Interface for a tracer that records activities on a call. Actual attempts for
 // this call are traced with CallAttemptTracer after invoking RecordNewAttempt()
 // on the CallTracer object.
-class CallTracer : public CallTracerInterface {
+class CallTracer : public CallTracerAnnotationInterface {
```
markdroth, line 87: Suggest renaming this to `ClientCallTracer` for clarity.

src/core/lib/channel/context.h
```diff
@@ -36,15 +36,16 @@ typedef enum {
   /// Value is a \a census_context.
   GRPC_CONTEXT_TRACING,
 
-  /// Value is a CallTracer object. (Only valid on the client-side call, or
+  /// Value is a CallTracerAnnotationInterface. (CallTracer object on the
+  /// client-side call, or
   /// ServerCallTracer on the server-side.)
```
markdroth, line 41: This can move up to the previous line.

src/core/lib/channel/call_tracer.h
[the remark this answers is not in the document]
markdroth: Suggest adding a TODO about this.

markdroth (approved):
This LGTM, modulo a few remaining comments.

I'd also like Craig to weigh in on the design question, but that does not have to block this PR.

src/core/lib/channel/call_tracer.h
yashykt: Done

src/core/lib/channel/context.h
yashykt: Done

ctiller:
If you're waiting for my input, please feel free to ping me on chat.
It'd be especially nice to get a heads up on reviews on promise based calls right now.

src/core/lib/channel/call_tracer.h
```diff
@@ -21,77 +21,141 @@
...
 #include "src/core/lib/transport/metadata_batch.h"
 #include "src/core/lib/transport/transport.h"
 
 namespace grpc_core {
 
+// The interface hierarchy is as follows -
+//                 CallTracerAnnotationInterface
+//                      /          \
+//        ClientCallTracer       CallTracerInterface
+//                                /             \
```
devbww, line 44: A backslash before a newline is probably something you want to avoid.

PR #32622, opened 2023-03-15
Fix race conditions with windows endpoint shutdown.

Vignesh2208:
The current endpoint semantics is as follows:
- On endpoint shutdown, the socket is immediately closed regardless and any pending registered read/write closures are not immediately executed. 
- The pending read/write closures get executed with aborted error whenever the next grpc_iocp_work method runs.

However the grpc_iocp_work may run only during grpc_shutdown and grpc_shutdown may only get scheduled after the pending registered read/write closures execute.

This PR changes the shutdown semantics to match shutdown semantics used in posix - i.e On endpoint shutdown, the socket is immediately closed and any pending registered read/write closures are executed immediately. Additional care is taken to ensure that the socket is not immediately deleted because the pending I/O ops still need to be flushed later during grpc_shutdown.

### DIFF
--- a/src/core/lib/iomgr/iocp_windows.cc
+++ b/src/core/lib/iomgr/iocp_windows.cc
@@ -43,6 +43,7 @@ static ULONG g_iocp_kick_token;
 static OVERLAPPED g_iocp_custom_overlap;
 
 static gpr_atm g_custom_events = 0;
+static gpr_atm g_pending_socket_shutdowns = 0;
 
 static HANDLE g_iocp;
 
@@ -90,6 +91,7 @@ grpc_iocp_work_status grpc_iocp_work(grpc_core::Timestamp deadline) {
   } else {
     abort();
   }
+  gpr_mu_lock(&socket->state_mu);
   if (socket->shutdown_called) {
     info->bytes_transferred = 0;
     info->wsa_error = WSA_OPERATION_ABORTED;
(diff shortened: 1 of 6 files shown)


src/core/lib/iomgr/iocp_windows.cc
```diff
@@ -155,4 +163,20 @@ void grpc_iocp_add_socket(grpc_winsocket* socket) {
   GPR_ASSERT(ret == g_iocp);
 }
 
+void grpc_iocp_register_socket_shutdown(grpc_winsocket* socket) {
```
drfloob, line 166: Does this method need to be guarded by the state_mu lock? The lock is currently when calling this method, but `shutdown_registered` is not consistently protected across all uses.

src/core/lib/iomgr/tcp_windows.cc
```diff
@@ -252,13 +259,15 @@ static void win_read(grpc_endpoint* ep, grpc_slice_buffer* read_slices,
     gpr_log(GPR_INFO, "TCP:%p win_read", tcp);
   }
 
-  if (tcp->shutting_down) {
+  gpr_mu_lock(&tcp->read_mu);
```
drfloob, line 262: I don't believe this lock is necessary here. An endpoint should never have more than one thread concurrently calling `grpc_tcp_read`, and because of that, there's no chance that any of this read state could be accessed concurrently. Race potential comes after the on_read callback is registered, which is the last line of this function.

Vignesh2208:
Closing this since it was landed in #33099

PR #32630, opened 2023-03-15
[testing] Update orca-* tests for better compatibility

eugeneo:
1. Make channel creation lazy. This allows test cases to update the configuration before the connection is made.
2. Pass load reports tracker when creating the policy. This way other test cases do not see any changes to ChannelArguments.

Using grpc_core::CoreConfiguration::RunWithSpecialConfiguration was considered but did not work as it removes other builders setup prior to starting the test cases.

### DIFF
--- a/test/cpp/interop/interop_client.cc
+++ b/test/cpp/interop/interop_client.cc
@@ -29,6 +29,7 @@
 #include "absl/strings/match.h"
 #include "absl/strings/str_format.h"
 #include "absl/strings/str_join.h"
+#include "absl/types/optional.h"
 
 #include <grpc/grpc.h>
 #include <grpc/support/alloc.h>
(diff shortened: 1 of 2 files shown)


eugeneo:
@markdroth I reverted most changes so the tracker is now passed through the channel arguments. Please take another look.

test/cpp/interop/interop_client.cc
```diff
@@ -161,44 +145,40 @@ absl::optional<std::string> OrcaLoadReportsDiff(
...
   }
   return unimplemented_service_stub_.get();
 }
 
 void InteropClient::ServiceStub::ResetChannel() {
-  channel_ = channel_creation_func_();
-  if (!new_stub_every_call_) {
-    stub_ = TestService::NewStub(channel_);
-  }
+  channel_ = absl::nullopt;
```
markdroth, line 173: Can just say:

```
channel_.reset();
stub_.reset();
```

test/cpp/interop/interop_client.h
```diff
@@ -127,9 +127,9 @@ class InteropClient {
 
    private:
     ChannelCreationFunc channel_creation_func_;
-    std::unique_ptr<TestService::Stub> stub_;
+    absl::optional<std::unique_ptr<TestService::Stub>> stub_;
```
markdroth, line 130: There's no need for `absl::optional<>` for the stub or the channel, since we don't need to differentiate between nullptr and nullopt.

markdroth (approved):
Just a couple of minor comments, otherwise looks great!

test/cpp/interop/interop_client.h
eugeneo: Done

test/cpp/interop/interop_client.cc
eugeneo: Done

eugeneo (commented):
I addressed the comments. Thank you for the review.

Issue #32628, opened 2023-03-15
C core improperly forked testing.proto


ejona86:
The canonical testing.proto is in grpc-proto. It seems #32599 forked the proto. The [upstream grpc-proto](https://github.com/grpc/grpc-proto/blob/master/grpc/testing/messages.proto) should always be changed first.

When we added orca testing earlier, the orca dependency caused trouble so we purposefully avoided the orca dependency. That's why we added `TestOrcaReport` and avoided the dependency. I don't know why we thought this time it'd be less of a problem.

CC @markdroth

ejona86:
It also seems that [orca_oob](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#orca_oob) would not be compatible with other languages. The grpc-proto version should be used; that's what the interop-test-descriptions is talking about.

markdroth:
Yes, I agree, copying the grpc-proto version is the right thing to do.  I completely forgot about that when Eugene was doing this, so it's my bad.  Sorry for the oversight, and thanks for the correction!

PR #32631, opened 2023-03-15
[testing] Synchronize interop test protos.

eugeneo:
Fixes: issue #32628


test/cpp/interop/interop_server.cc
```diff
@@ -73,6 +73,18 @@ const char kEchoInitialMetadataKey[] = "x-grpc-test-echo-initial";
 const char kEchoTrailingBinMetadataKey[] = "x-grpc-test-echo-trailing-bin";
 const char kEchoUserAgentKey[] = "x-grpc-test-echo-useragent";
 
+namespace {
+xds::data::orca::v3::OrcaLoadReport TestOrcaReportToXdsOrcaReport(
```
markdroth, line 77: Instead of adding this conversion function, I suggest just changing `RecordCallMetrics()` and `RecordServerMetrics()` to take their input as a `TestOrcaReport` proto.

markdroth (approved):

test/cpp/interop/interop_server.cc
eugeneo: Done

eugeneo (commented):
Thank you for the review. Comment was fixed. Please take another look.

PR #32634, opened 2023-03-16
Merge per-request and per-server named metrics field-wise

yousukseung:
We currently take named metrics recorded per-request only. Instead we should merge field-wise.


markdroth:
Can you please add a test for this case?  I think you can just modify one of the test cases in client_lb_end2end_test.

markdroth (approved):

Issue #28663, opened 2022-01-23
C# protoc codegen plugin: Add base_namespace option to grpc_csharp_plugin


seniorquico:
### Is your feature request related to a problem? Please describe.

This is a follow-up to #13724. It was closed because a workaround was introduced in the `Grpc.Tools` NuGet package. However, I have a use case that isn't addressed by #17672.

I'm using [Buf](https://buf.build/) to generate sources (among other things), and it works for all languages except C# gRPC services (it _does_ work for C# protobuf messages). Buf generates sources directly using the `grpc_csharp_plugin` through `protoc`, and this ends up generating duplicate files (leaving me with incomplete generated sources) when two services are in files with the same name but in different directories. For example:

- `proto\v1\service.proto`
- `proto\v2\service.proto`

This isn't a problem with protobuf's `csharp_plugin` thanks to the `base_namespace` option. However, adding a similar option to `grpc_csharp_plugin` was rejected in #24426.

### Describe the solution you'd like

I'd love to see a native fix to this issue in `grpc_csharp_plugin` for scenarios where generating sources happens outside MSBuild/through `protoc`. A couple possible solutions:

- Add a `base_namespace` option that works like `csharp_plugin`'s (i.e. reconsider #24426).
- Add any other file name deconfliction algorithm (like random or hash-based file suffixes) to `grpc_csharp_plugin`, given how the C# compiler cares very little about the actual file names of the generated sources.

### Describe alternatives you've considered

I could assign unique file names to the different services. For example:

- `proto\v1\service_v1.proto`
- `proto\v2\service_v2.proto`

However, this would impact the output of other language plugins (e.g. Javascript) and introduce friction elsewhere.

Alternatively-- I could file this as an issue with the Buf project and request they investigate/implement their own workaround for this limitation. I thought about doing that first, but reconsidered after identifying that there are options in most (all?) other protobuf/gRPC plugins to work around duplicate file name issues like this. This just appears to be a limitation of the C# gRPC plugin.

nahk-ivanov:
+1

Doing it through Buf as well, trying to generalize the build of client/server SDKs for different languages, and gRPC C# plugin is giving me troubles. 😞 Would be more versatile if it was solved in the generator plugin, rather than collateral build tools.

etiennecl:
+1

[...]

PR #32414, opened 2023-02-16
[csharp proto plugin] Apply Obsolete attribute to deprecated services and methods in C# generated code

tonydnewell:
Apply Obsolete attribute to deprecated services and methods in C# generated code

Fix for https://github.com/grpc/grpc/issues/28597

- Deprecated support for enums and enum values is already fixed by https://github.com/protocolbuffers/protobuf/pull/10520 but this is not yet released. It is fixed in  Protocol Buffers v22.0-rc1 but the gRPC repo currently has 21.12 as the protocol buffers submodule.

- Deprecated support for messages and fields already exists in the protocol buffers compiler.

The fix in this PR adds `Obsolete` attribute to classes and methods for deprecated services and methods within services. e.g.
```
service Greeter {
  option deprecated=true; // service level deprecated
  // Sends a greeting
  rpc SayHello (HelloRequest) returns (HelloReply) {
    option deprecated=true; // method level deprecated
  }
}
```

I couldn't find any protocol buffers plugin tests to update. Tested locally.

PR #32636, opened 2023-03-16
[CSHARP] Add base_namespace experimental option to C# plugin

tonydnewell:
Added `base_namespace` experimental option to `grpc_csharp_plugin` as this has been requested several times by
people not using `Grpc.Tools` to generate their code - see https://github.com/grpc/grpc/issues/28663

Notes:
- it should not be used with `Grpc.Tools`. That has a different way of handling duplicate proto file names in different directories. Using this option will break those builds.  It can only be used on the `protoc` command line.
- it uses common code with the `base_namespace` option for C# in `protoc`, which unfortunately has a slightly different name mangling algorithm for converting proto file names to C# camel case names. This only affects files with punctation or numbers in the name. This should not matter unless you are expecting specific file names
- See https://protobuf.dev/reference/csharp/csharp-generated/#compiler_options for an explanation of this option


tonydnewell:
@jtattermusch for review

src/compiler/csharp_generator_helpers.h
```diff
@@ -22,13 +22,30 @@
...
+  // If base_namespace is provided then slightly different name mangling
+  // is used to generate the service file name. This is because this
+  // uses common code with protoc. For most file names this will not
+  // make a difference (only files with punctuation or numbers in the
+  // name.)
+  // Otherwise the behavor remains the same as before.
+  if (base_namespace.empty()) {
+    out_file = grpc_generator::FileNameInUpperCamel(file, false) + file_suffix;
+  } else {
+    out_file = GetOutputFile(file, file_suffix, true, base_namespace, error);
```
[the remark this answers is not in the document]
jtattermusch: Also, will using GetOutputFile be still usable in protobuf 22.x (which we are going to upgrade to soon)?

Looks like it would? 
https://github.com/protocolbuffers/protobuf/blob/3a871acca05ca4cdf79fc14ea9b701be6fe745f1/src/google/protobuf/compiler/csharp/names.h#L121
https://github.com/protocolbuffers/protobuf/blob/3a871acca05ca4cdf79fc14ea9b701be6fe745f1/src/google/protobuf/compiler/csharp/BUILD.bazel#L15

src/csharp/BUILD-INTEGRATION.md
```diff
@@ -170,12 +170,21 @@ to perform code generation. Here is an overview of the available `grpc_csharp_pl
...
+| base_namespace | none       | *Experimental - may change or be removed.* Same as `base_namespace` for `protoc` C# options.  **Can only be used on the command line.** |
 
-Note that the protocol buffer compiler has a special commandline syntax for plugin options.
-Example:
-```
-protoc --plugin=protoc-gen-grpc=grpc_csharp_plugin --csharp_out=OUT_DIR \
-    --grpc_out=OUT_DIR --grpc_opt=lite_client,no_server \
+To use these options with `Grpc.Tools` specify them in the __GrpcOutputOptions__
+metadata in the `<Protobuf>` item.
+- Note: `file_suffix` and `base_namespace` should not be used with `Grpc.Tools`. Using them will break the build.
```
jtattermusch, line 178: add a note that when base_namespace is used, it changes the algorithm for generating file name from the name of the proto file slightly (this might be quite important for users and currently it's only mentioned in a comment in the implementation.

src/csharp/BUILD-INTEGRATION.md
```diff
@@ -170,12 +170,21 @@ to perform code generation. Here is an overview of the available `grpc_csharp_pl
...
+
+To use these options on the command line specify them using the `--grpc_opt`
```
jtattermusch, line 180: Unless that's mentioned somewhere, also mention that the "csharp" codegen itself (for protobuf messages), uses --csharp_opt flag.
Also, probably add --csharp_opt to the example, since it's more realistic that actual users will want to set base_namespace for both --grpc_opt and --csharp_opt shall they use the `base_namespace` option.

src/csharp/BUILD-INTEGRATION.md
jtattermusch: nit: add link to "https://protobuf.dev/reference/csharp/csharp-generated/#compiler_options" in "same as base_namespace for protoc C# options"?

jtattermusch (commented):
Overall this looks good but opens up some discussion. I'm also wondering whether we should add bazel tests for codegen first, so that we can add this feature in a more controlled way.

jtattermusch (approved):
I think this looks good, but I'd prefer to hold until we have the bazel codegen tests in place.

jtattermusch:
I just merged https://github.com/grpc/grpc/pull/32734, so I think you should rebase and provide a codegen test for when the base_namespace option is enabled? (something simple is enough, e.g. run grep to check that the important parts of the generated file contain the expected stuff there when the option is set and aren't there if it's not set)

tonydnewell:
@jtattermusch 
Added bazel tests.
This should probably be merged _**after**_ https://github.com/grpc/grpc/pull/32414
Both modify the `BUILD` bazel file so some merging/editing may need to be done before this can be merged.

src/compiler/csharp_generator_helpers.h
```diff
@@ -22,13 +22,30 @@
 #include "src/compiler/config.h"
 #include "src/compiler/generator_helpers.h"
 
+using google::protobuf::compiler::csharp::GetOutputFile;
```
apolcyn, line 25: @tonydnewell we needed to revert this in https://github.com/grpc/grpc/pull/32636 because of some internal build errors, basically the following:

```
src/compiler/csharp_generator_helpers.h:25] error: no member named 'compiler' in namespace 'google::protobuf'
using google::protobuf::compiler::csharp::GetOutputFile;
```

I have not gotten too far into the details but I'm wondering if the `google::protobuf::compiler::csharp::GetOutputFile` is a new API, only in some older versions of protobuf, or something else like that.

PR #32637, opened 2023-03-16
[EventEngine] Return correct value EventEngine::IsWorkerThread and use this value in completion queue handling to correctly offload work

Vignesh2208:


drfloob:
Looks good. The PR name is now a bit misleading, this is now more of an optimization for completion queue callback execution.

drfloob (approved):

Issue #32587, opened 2023-03-10
Python gRPC AIO Client Interceptor Iteration Issue


cpvandehey:
### What version of gRPC and what language are you using?
`1.51.3`
Python

### What operating system (Linux, Windows,...) and version?
Mac/linux

### What runtime / compiler are you using (e.g. python version or version of gcc)
`python 3.9.6`

### What did you do?
My team and I are working on a custom async gRPC client with a set of interceptors that is supposed to be in lock-step parity with its current synchronous version. We have found nearly all of the interceptors to be straight forward **except** the retry interceptor. We have found a few reasons why that might be.

First off, heres what our super general async retry interceptor looks like 
```
class AsyncRetryInterceptor(grpc.aio._interceptor.UnaryUnaryClientInterceptor):
    async def intercept_unary_unary(
        self,
        continuation,
        client_call_details,
        request_or_iterator
    ) :
      tries_remaining = 3
      while 0 < tries_remaining:
            try:
                print("Attempting retry")
                call = await continuation(client_call_details, request_or_iterator)
                result = await call
            except Exception:
                tries_remaining -= 1

        return result
```
Essentially, this runs the continuation of the other `N` interceptors and the grpc method multiple times until it receives a result or runs out of tries. Delightful :)

`AsyncRetryInterceptor` works as expected when it is the only interceptor (_or is the final interceptor in the chain_), however, it behaves in an unexpected way when it is ordered differently.

Suppose I introduce another interceptor like so:
```
class AsyncOtherInterceptor(grpc.aio._interceptor.UnaryUnaryClientInterceptor):
    async def intercept_unary_unary(
        self,
        continuation,
        client_call_details,
        request_or_iterator,
    ) ->:
        print("before")
        call =  await continuation(client_call_details, request_or_iterator)  # returns an awaitable
        result = await call  # returns result of the grpc method
        print("after")
        return result
```
If we happen put this interceptor **AFTER** the retry interceptor, we see this odd behavior.
```
Attempting retry
before
after
Attempting retry
Attempting retry
...
```
When we expect to see:
```
Attempting retry
before
after
Attempting retry
before
after
Attempting retry
before
after
...
```

There seems to be some fundamental difference in the way/order the interceptors are executed. In the current async world, an interceptor appears awaited once (_and for all_)
[Our leading clue is in the `_run_interceptor` logic linked here](https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/aio/_interceptor.py#L606)
[...]

PR #32641, opened 2023-03-16
Fix an issue that AIO interceptors can only be called once.

XuanWang-Amos:
Prior to this change, we invoke aycnio interceptors by converting them to `iterator` first, which means we can only call them once before they're exhausted.

This PR changes the implementation to use `list`, thus the interceptors can be called multiple times.

### DIFF
--- a/src/python/grpcio/grpc/aio/_interceptor.py
+++ b/src/python/grpcio/grpc/aio/_interceptor.py
@@ -17,8 +17,8 @@ from abc import abstractmethod
 import asyncio
 import collections
 import functools
-from typing import (AsyncIterable, Awaitable, Callable, Iterator, Optional,
-                    Sequence, Union)
+from typing import (AsyncIterable, Awaitable, Callable, Iterator, List,
+                    Optional, Sequence, Union)
 
 import grpc
 from grpc._cython import cygrpc
@@ -571,7 +571,7 @@ class InterceptedUnaryUnaryCall(_InterceptedUnaryResponseMixin, InterceptedCall,
     _channel: cygrpc.AioChannel
 
     # pylint: disable=too-many-arguments
-    def __init__(self, interceptors: Sequence[UnaryUnaryClientInterceptor],
+    def __init__(self, interceptors: List[UnaryUnaryClientInterceptor],
                  request: RequestType, timeout: Optional[float],
                  metadata: Metadata,
                  credentials: Optional[grpc.CallCredentials],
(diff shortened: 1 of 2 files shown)


src/python/grpcio_tests/tests_aio/unit/client_unary_unary_interceptor_test.py
```diff
@@ -224,6 +224,51 @@ async def intercept_unary_unary(self, continuation,
             self.assertEqual(await interceptor.calls[1].code(),
                              grpc.StatusCode.OK)
 
+    async def test_retry_with_multiple_interceptors(self):
+
+        class RetryInterceptor(aio.UnaryUnaryClientInterceptor):
+            """a Retry Interceptor which calls another interceptor"""
```
gnossen, line 230: Technically speaking, this isn't true. This interceptor only calls another interceptor because of the `interceptors` kwarg to `aio.secure_channel` below. I think it would be more explanatory to move a comment like this down there.

src/python/grpcio/grpc/aio/_interceptor.py
```diff
@@ -599,16 +599,14 @@ async def _invoke(
...
                 client_call_details: ClientCallDetails,
                 request: RequestType) -> _base_call.UnaryUnaryCall:
 
-            interceptor = next(interceptors, None)
-
-            if interceptor:
-                continuation = functools.partial(_run_interceptor, interceptors)
-
-                call_or_response = await interceptor.intercept_unary_unary(
+            if len(interceptors) > 0:
```
gnossen, line 606: Nit: According to the style guide (which I know we don't technically follow), this should be `if interceptors:`

src/python/grpcio/grpc/aio/_interceptor.py
```diff
@@ -599,16 +599,14 @@ async def _invoke(
...
+                continuation = functools.partial(_run_interceptor,
+                                                 interceptors[1:])
```
gnossen, line 608: Nice. I like this approach better than copying.

src/python/grpcio/grpc/aio/_interceptor.py
```diff
@@ -571,7 +571,7 @@ class InterceptedUnaryUnaryCall(_InterceptedUnaryResponseMixin, InterceptedCall,
     _channel: cygrpc.AioChannel
 
     # pylint: disable=too-many-arguments
-    def __init__(self, interceptors: Sequence[UnaryUnaryClientInterceptor],
```
gnossen, line 574: This `__init__` is part of our public interface, so we can't change the type. Instead, we'll want to convert the argument to a `list` within the method.

gnossen (commented):
Nice job! This is an important fix. I'm surprised it wasn't caught earlier.

src/python/grpcio_tests/tests_aio/unit/client_unary_unary_interceptor_test.py
XuanWang-Amos: That's right, I'll move this down.

gnossen:
Fixes https://github.com/grpc/grpc/issues/32587

gnossen (approved):

PR #32647, opened 2023-03-17
Optimize absl::StrFormat calls showing up on profiles

scwhittle:
- construct string for tracing in chttp2_transport only if tracing is enabled
- use equivalent absl::StrCat/StrAppend instead of absl::StrFormat/StrJoin for formatting showing up on cpu profiles

### DIFF
--- a/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
+++ b/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
@@ -205,20 +205,14 @@ static void keepalive_watchdog_fired_locked(
 static void maybe_reset_keepalive_ping_timer_locked(grpc_chttp2_transport* t);
 
 namespace {
-void MaybeRecordTransportAnnotation(grpc_chttp2_stream* s,
-                                    absl::string_view annotation) {
-  GPR_ASSERT(s->context);
-  if (!grpc_core::IsTraceRecordCallopsEnabled()) {
-    return;
+grpc_core::CallTracerInterface* CallTracerIfEnabled(grpc_chttp2_stream* s) {
+  if (!s->context || !grpc_core::IsTraceRecordCallopsEnabled()) {
+    return nullptr;
   }
-  auto* call_tracer = static_cast<grpc_core::CallTracerInterface*>(
+  return static_cast<grpc_core::CallTracerInterface*>(
       static_cast<grpc_call_context_element*>(
           s->context)[GRPC_CONTEXT_CALL_TRACER_ANNOTATION_INTERFACE]
           .value);
-  if (!call_tracer) {
-    return;
-  }
-  call_tracer->RecordAnnotation(annotation);
 }
 }  // namespace
 
@@ -369,8 +363,8 @@ static void read_channel_args(grpc_chttp2_transport* t,
         grpc_core::MakeRefCounted<grpc_core::channelz::SocketNode>(
             std::string(grpc_endpoint_get_local_address(t->ep)),
             std::string(t->peer_string.as_string_view()),
-            absl::StrFormat("%s %s", get_vtable()->name,
-                            t->peer_string.as_string_view()),
+            absl::StrCat(get_vtable()->name, " ",
+                         t->peer_string.as_string_view()),
             channel_args
                 .GetObjectRef<grpc_core::channelz::SocketNode::Security>());
   }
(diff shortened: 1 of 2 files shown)


scwhittle:
It looks like creating trace string even if tracing not enabled was recently introduced in https://github.com/grpc/grpc/pull/32503
@ananda1066 @Vignesh2208

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -205,20 +205,14 @@ static void keepalive_watchdog_fired_locked(
 static void maybe_reset_keepalive_ping_timer_locked(grpc_chttp2_transport* t);
 
 namespace {
-void MaybeRecordTransportAnnotation(grpc_chttp2_stream* s,
-                                    absl::string_view annotation) {
-  GPR_ASSERT(s->context);
-  if (!grpc_core::IsTraceRecordCallopsEnabled()) {
-    return;
+grpc_core::CallTracerInterface* CallTracerIfEnabled(grpc_chttp2_stream* s) {
+  if (!s->context || !grpc_core::IsTraceRecordCallopsEnabled()) {
```
drfloob, line 209: nit: prefer explicit checks against nullptr, such as `if (s->context == nullptr || ...)`.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1310,12 +1305,12 @@ static void perform_stream_op_locked(void* stream_op,
...
 
-  if (s->context != nullptr) {
-    MaybeRecordTransportAnnotation(
-        s, absl::StrFormat(
-               "perform_stream_op_locked[s=%p; op=%p]: %s; on_complete = %p", s,
-               op, grpc_transport_stream_op_batch_string(op, true).c_str(),
-               op->on_complete));
+  if (auto* tracer = CallTracerIfEnabled(s)) {
+    tracer->RecordAnnotation(absl::StrCat(
+        "perform_stream_op_locked[s=0x", absl::Hex(s, absl::kZeroPad16),
```
drfloob, line 1310: Same suggestion here regarding `%p` for consistency.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1232,10 +1226,11 @@ void grpc_chttp2_complete_closure_step(grpc_chttp2_transport* t,
...
   }
 
-  if (s->context != nullptr) {
-    MaybeRecordTransportAnnotation(
-        s, absl::StrFormat("on_complete: s=%p %p desc=%s err=%s", s, closure,
-                           desc, grpc_core::StatusToString(error).c_str()));
+  if (auto* tracer = CallTracerIfEnabled(s)) {
+    tracer->RecordAnnotation(
+        absl::StrCat("on_complete: s=0x", absl::Hex(s, absl::kZeroPad16), " 0x",
+                     absl::Hex(closure, absl::kZeroPad16), " desc=", desc,
```
drfloob, line 1232: Given that the string is now only generated when tracing is enabled, that should sufficiently mitigate the performance impact, yes?

If so, I'd recommend sticking with `StrFormat("%p", s)` here for readability, and consistency with the ~1700 other places in the codebase where `%p` is used to print addresses.

If not, and this StrCat + `absl::Hex` trick is necessary for performance, we may want to make this change more broadly.

drfloob (commented):
Thank you for the improvement!

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
scwhittle: Done. I changed the tracer check below to compare with nullptr too. AFAICT grpc supports c++ versions before 17 so I didn't use
if (auto* tracer = ....; tracer != nullptr)

If that's not the case let me know since that seems cleaner if allowed.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
scwhittle: Done

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
scwhittle: Reverted. I left the changes to transport_op_string to move to appending instead of building up a vector and joining but switched some of them back to format string via StrAppendFormat.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
drfloob: > grpc supports c++ versions before 17

Indeed, gRPC still needs to support c++14. This looks good, thanks.

scwhittle:
updated based upon iwyu failing check

drfloob:
> updated based upon iwyu failing check

I kicked off the CI again.

FYI, for the next time you contribute, we have a helpful bot that fixes many of the sanity checks for you and files a pull request against your branch. You can usually just merge it. https://github.com/scwhittle/grpc/pull/1/files

drfloob (approved):

PR #32652, opened 2023-03-17
[PHP] Add @deprecated tag in PHP generated methods and services

negram:
When we deprecate services in proto-file:
```proto
syntax = "proto3";

option php_namespace = "Test";

service MyService {
  option deprecated = true;
  rpc Call(Test) returns (Test) {
    option deprecated = true;
  };
}

message Test {
}
```

It might be a good idea to propagate this flat in the generated classes so that the final users will be aware of the changes coming soon.

Example of the generated code:
```php
<?php
// GENERATED CODE -- DO NOT EDIT!

namespace Test;

/**
 * @deprecated
 */
class MyServiceClient extends \Grpc\BaseStub {

    /**
     * @param string $hostname hostname
     * @param array $opts channel options
     * @param \Grpc\Channel $channel (optional) re-use channel object
     */
    public function __construct($hostname, $opts, $channel = null) {
        parent::__construct($hostname, $opts, $channel);
    }

    /**
     * @deprecated
     * @param \Test\Test $argument input argument
     * @param array $metadata metadata
     * @param array $options call options
     * @return \Grpc\UnaryCall
     */
    public function Call(\Test\Test $argument,
      $metadata = [], $options = []) {
        return $this->_simpleRequest('/MyService/Call',
        $argument,
        ['\Test\Test', 'decode'],
        $metadata, $options);
    }

}
```

These tags will cause IDE to highlight the class/method to bring the developer's attention.

This is consistent with how "protoc" handles the "deprecated" option for PHP

### DIFF
--- a/src/compiler/php_generator.cc
+++ b/src/compiler/php_generator.cc
@@ -77,6 +77,9 @@ void PrintMethod(const MethodDescriptor* method, Printer* out) {
       GeneratedClassName(output_type), output_type->file());
 
   out->Print("/**\n");
+  if (method->options().deprecated()) {
+    out->Print(" * @deprecated\n");
+  }
   out->Print(GetPHPComments(method, " *").c_str());
   if (method->client_streaming()) {
     if (method->server_streaming()) {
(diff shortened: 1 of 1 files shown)


negram:
Could anyone take a look at this? Maybe I did something wrong?..
I will be happy to fix if there is an issue

negram:
@alto-php, @stanley-cheung, would it be possible to take a quick look at this, please?

alto-php (approved):

alto-php:
Sorry for late response, LGTM, thanks for the PR.

PR #32656, opened 2023-03-20
Revert "[EventEngine] Return correct value EventEngine::IsWorkerThread and use this value in completion queue handling to correctly offload work"

ctiller:
Reverts grpc/grpc#32637

Apologies for not catching this before submission.

I continue to think that IsWorkerThread as an EventEngine member function is a poor API choice: it implies that the thread is a thread for the referenced event engine, and that's not what we're returning, and also rarely useful.

If we want this concept in the API it should be a static function with a way of registering a thread into the set. I however firmly believe we'll do better long term if we abandon the concept of being able to identify from outside who created a thread and instead solve problems in other ways.


ctiller:
This is responsible for b/274531212 internally.

PR #32673, opened 2023-03-21
[fix] Use signed millis in grpc_core::Timestamp math

drfloob:
Most uses of `Timestamp::milliseconds_after_process_epoch()` are already implicitly converting the result to `int64_t`. Using unary `operator-` on unsigned integers can be problematic, as done in Timestamp's operator- implementation - the result may be unsigned.

See 
* https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=msvc-170
* https://google.github.io/styleguide/cppguide.html#Integer_Types

### DIFF
--- a/src/core/lib/event_engine/posix_engine/timer.h
+++ b/src/core/lib/event_engine/posix_engine/timer.h
@@ -176,7 +176,7 @@ class TimerList {
   const size_t num_shards_;
   grpc_core::Mutex mu_;
   // The deadline of the next timer due across all timer shards
-  std::atomic<uint64_t> min_timer_;
+  std::atomic<int64_t> min_timer_;
   // Allow only one FindExpiredTimers at once (used as a TryLock, protects no
   // fields but ensures limits on concurrency)
   grpc_core::Mutex checker_mu_;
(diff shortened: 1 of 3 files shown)


ctiller:
Is there a bug being fixed here?

I ask mostly because whilst unsigned wraparound can be problematic it's at least defined, signed overflow/underflow is undefined behavior - and we commonly have values at INT64_MAX or INT64_MIN - and so may be subject to that compiler chosen behavior.

drfloob:
To summarize the link I posted above, the warning is `unary minus operator applied to unsigned type, result still unsigned`. The salient bit of the description:

> If the numeric value fits in the range of an unsigned integer type, but not the corresponding signed integer type, the compiler interprets the value as unsigned. An unsigned value is unchanged by the unary negation operator.

The bug here - whether we've seen it cause a problem or not - is a large millisecond addition instead of subtraction if the process has some timestamp that has been alive long enough (~25 days).

ctiller:
I'm gonna note that the ubsan failures here bear out my worries.

drfloob:
I added a few lines to avoid the UB. I prefer this PR for readability reasons, compared to the currently obscured casts. I think the current implementation behaves correctly though, so I'll let you decide whether this should go in.

Also, I think `MillisAdd` and friends can move from `time.h` to `time.cc`.

PR #32674, opened 2023-03-21
fix: gRPC: Remove AWS URL validation

BigTailWolf:
As of adding support in future for TPC support. We removed all url validations from Credentials.
AWS url validation doesn't make sense here either.


BigTailWolf:
@veblush Please check

yijiem (approved):

veblush (approved):

PR #24785, opened 2020-11-18
csharp - proposed fix for #24784 - expose marshaller faults in DebugException

mgravell:
context from #24784:

If the marshaller faults, the [original exception is lost](https://github.com/grpc/grpc/blob/a1ad4800da2bb02c843bdaea8fb06518916b6fba/src/csharp/Grpc.Core/Internal/AsyncCallBase.cs#L349-L355) - it is never exposed to the client and a pre-defined [status is reported](https://github.com/grpc/grpc/blob/a1ad4800da2bb02c843bdaea8fb06518916b6fba/src/csharp/Grpc.Core/Internal/AsyncCallBase.cs#L41).

This makes it very hard to diagnose underlying causes, as seen [in this Stack Overflow question](https://stackoverflow.com/questions/64868856/serializing-an-ienumerable-in-protobuf-net?noredirect=1#comment114732115_64868856).

For server-side faults, the original exception [is exposed](https://github.com/grpc/grpc/blob/a1ad4800da2bb02c843bdaea8fb06518916b6fba/src/csharp/Grpc.Core/Internal/AsyncCallBase.cs#L372).

Suggestion: allow client-code to expose the fault in the `DebugException`; this is a trivial change.

Issue #24784, opened 2020-11-18
csharp - expose DebugException in client-side response marshaller exceptions 


mgravell:
If the marshaller faults, the [original exception is lost](https://github.com/grpc/grpc/blob/a1ad4800da2bb02c843bdaea8fb06518916b6fba/src/csharp/Grpc.Core/Internal/AsyncCallBase.cs#L349-L355) - it is never exposed to the client and a pre-defined [status is reported](https://github.com/grpc/grpc/blob/a1ad4800da2bb02c843bdaea8fb06518916b6fba/src/csharp/Grpc.Core/Internal/AsyncCallBase.cs#L41).

This makes it very hard to diagnose underlying causes, as seen [in this Stack Overflow question](https://stackoverflow.com/questions/64868856/serializing-an-ienumerable-in-protobuf-net?noredirect=1#comment114732115_64868856).

For server-side faults, the original exception [is exposed](https://github.com/grpc/grpc/blob/a1ad4800da2bb02c843bdaea8fb06518916b6fba/src/csharp/Grpc.Core/Internal/AsyncCallBase.cs#L372).

Suggestion: allow client-code to expose the fault in the `DebugException`; this is a trivial change:

``` txt
-        protected static readonly Status DeserializeResponseFailureStatus = new Status(StatusCode.Internal, "Failed to deserialize response message.");
...
-            CancelWithStatus(DeserializeResponseFailureStatus);
+            CancelWithStatus(new Status(StatusCode.Internal, "Failed to deserialize response message.", deserializeException));
```

Would a PR of the same be reasonable?

PR #32681, opened 2023-03-22
Including underlying deserialization errors in Status (as a string)

tonydnewell:
This a fix for https://github.com/grpc/grpc/issues/24784 that supersedes proposed PR https://github.com/grpc/grpc/pull/24785

The original request was to expose `DebugException` in client-side response marshaller exceptions. 

However the native API only accepts status code and description (and not the `DebugException` in the `Status`). 

This PR create a new `Status` to encapsulate the deserialization exception.  A simple string representation of the `DebugException` is also provided  as part of the description so that the information from the exception is not lost.

In some situations the `DebugException` may be retained and available to the API user, such as a blocking unary call, so it is also included.

Tests for marshalling errors have also been updated


tonydnewell:
@jtattermusch for review
There seems to be lots of CI checks failing but none look relevant.

PR #32606, opened 2023-03-13
[protobuf] Upgrade third_party/protobuf to 22.x

jtattermusch:
The very non-trivial upgrade of third_party/protobuf to 22.x
This PR strives to be as small as possible and many changes that were compatible with protobuf 21.x and didn't have to be merged atomically with the upgrade were already merged. 
Due to the complexity of the upgrade, this PR wasn't created automatically by a tool, but manually. Subsequent upgraded of third_party/protobuf with our OSS release script should work again once this change is merged.

This is best reviewed commit-by-commit, I tried to group changes in logical areas.

Notable changes:
- the upgrade of third_party/protobuf submodule, the bazel protobuf dependency itself
- upgrade of UPB dependency to 22.x  (in the past, we used to always upgrade upb to "main", but upb now has release branch as well). UPB needs to be upgraded atomically with protobuf since there's a de-facto circular dependency (new protobuf depends on new upb, which depends on new protobuf for codegen).
- some protobuf and upb bazel rules are now aliases, so ` extract_metadata_from_bazel_xml.py` and `gen_upb_api_from_bazel_xml.py` had to be modified to be able to follow aliases and reach the actual aliased targets.
- some protobuf public headers were renamed, so especially `src/compiler` needed to be updated to use the new headers.
- protobuf and upb now both depend on utf8_range project, so since we bundle upb with grpc in some languages, we now have to bundle utf8_range as well (hence changes in build for python, PHP, objC, cmake etc).
- protoc now depends on absl and utf8_range  (previously protobuf had absl dependency, but not for the codegen part), so python's make_grpcio_tools.py required partial rewrite to be able to handle those dependencies in the grpcio_tools build.
- many updates and fixes required for C++ distribtests (currently they all pass, but we'll probably need to follow up, make protobuf's and grpc's handling of dependencies more aligned and revisit the distribtests)
- bunch of other changes mostly due to overhaul of protobuf's and upb's internal build layout.

TODOs:
- [DONE] make sure IWYU and clang_tidy_code pass
- create a list of followups (e.g. work to reenable the few tests I had to disable and to remove workaround I had to use)
- [DONE in cl/523706129] figure out problem(s) with internal import

PR #32682, opened 2023-03-22
Cleanup: Remove unused protobuf WKT headers

jtattermusch:
I've been getting warnings about unused headers when trying to upgrade protobuf 22.x, so better get this off the table in advance.


yashykt (approved):

yashykt:
i would be fine as long as iwyu and build succeeds

jtattermusch:
This was dealt with as part of https://github.com/grpc/grpc/pull/32606

PR #32686, opened 2023-03-22
Add custom LB xDS test description

temawi:
@eugeneo

### DIFF
--- a/doc/xds-test-descriptions.md
+++ b/doc/xds-test-descriptions.md
@@ -813,6 +813,12 @@ The `backendService` will have the following `localityLbPolicies` entry:
 ]
 ```
 
+The client **should not** implement the `test.ThisLoadBalancerDoesNotExist`, but
+it **should** implement `test.RpcBehaviorLoadBalancer`. The
+`RpcBehaviorLoadBalancer` implementation should set the rpcBehavior request
+header based on the configuration it is provided. The `rpcBehavior` field value
+in the config should be used as the header value.
+
 Assert:
 
 1. The first custom policy is ignored as the client does not have an
--- a/doc/xds-test-descriptions.md
+++ b/doc/xds-test-descriptions.md
@@ -780,3 +780,49 @@ backends and all requests end with the `OK` status.
 3.  The test driver removes the client configuration to send metadata. The
 driver asserts that during some 10-second interval, traffic is equally
 distributed among the five backends, and all requests end with the `OK` status.
+
+### custom_lb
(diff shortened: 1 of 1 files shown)


doc/xds-test-descriptions.md
```diff
@@ -779,4 +779,50 @@ that during some 10-second interval, all traffic goes to the other four
...
+ignore a bad one and try the next one on the list until it finds a good one.
+
+Client parameters:
+
+1.  --num_channels=1
+2.  --qps=100
+
+Load balancer configuration:
+
+One MIG with a single backend.
```
eugeneo, line 796: Please clarify... I have not seen this acronym before so it may confuse other people implementing tests as well.

doc/xds-test-descriptions.md
```diff
@@ -779,4 +779,50 @@ that during some 10-second interval, all traffic goes to the other four
...
+
+The `backendService` will have the following `localityLbPolicies` entry:
```
eugeneo, line 798: I am not sure this is portable. Maybe this should be reworded, e.g. mention that `test.RpcBehaviorLoadBalancer` custom LB policy should be defined and understand the config and `test.ThisLoadBalancerDoesNotExist` should not be defined.

I wonder if this should pass `--service_config_json` to client and said config should configure the policies.

eugeneo (changes requested):
I have some comments that may originate from my lack of familiarity with this part of the testing framework.

doc/xds-test-descriptions.md
temawi: MIG = Managed Instance Group, used to manage multiple VMs as a group
https://cloud.google.com/compute/docs/instance-groups

It's used elsewhere in this file so I just followed the pattern.

eugeneo (approved):

doc/xds-test-descriptions.md
temawi: I added language that explains what the LB should do. For reference you can look at the Java implementation here: https://github.com/grpc/grpc-java/blob/master/interop-testing/src/main/java/io/grpc/testing/integration/RpcBehaviorLoadBalancerProvider.java

We can't use service config directly like this because the idea is to test things end to end with Traffic Director. The test will set up the GCP backendService with this custom LB config and TD passes it down to the client that will read it and set up load balancing accordingly. The test itself is here: https://github.com/grpc/grpc/blob/master/tools/run_tests/xds_k8s_test_driver/tests/custom_lb_test.py

PR #32690, opened 2023-03-23
xDS: Include orca named_metrics in LRS load reports

yousukseung:


include/grpcpp/ext/call_metric_recorder.h
```diff
@@ -73,6 +78,16 @@ class CallMetricRecorder {
...
                                                       double value) = 0;
+
+  /// Records an application-specific opaque metric measurement.
+  /// Multiple calls to this method with the same name will
+  /// override the corresponding stored value. The lifetime of the
+  /// name string needs to be longer than the lifetime of the RPC
+  /// itself, since it's going to be sent as trailers after the RPC
+  /// finishes. It is assumed the strings are common names that
+  /// are global constants.
+  virtual CallMetricRecorder& RecordNamedMetric(string_ref name,
```
yousukseung, line 89: > Let's just call this RecordNamedMetric().
Renamed.

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_impl.cc
```diff
@@ -332,7 +333,13 @@ class XdsClusterImplLb::Picker::SubchannelCallTracker
     }
     // Record call completion for load reporting.
     if (locality_stats_ != nullptr) {
-      locality_stats_->AddCallFinished(!args.status.ok());
+      auto* backend_metric_data =
+          args.backend_metric_accessor->GetBackendMetricData();
+      const std::map<absl::string_view, double>* named_metrics = nullptr;
+      if (backend_metric_data != nullptr) {
```
yousukseung, line 339: > args.backend_metric_accessor will never be null. Also, it seems cleaner to not call GetBackendMetricData() multiple times.

Removed the null check and redundant calls.

src/core/ext/xds/xds_client_stats.h
```diff
@@ -215,27 +216,32 @@ class XdsClusterLocalityStats : public RefCounted<XdsClusterLocalityStats> {
   Snapshot GetSnapshotAndReset();
 
   void AddCallStarted();
-  void AddCallFinished(bool fail = false);
+  void AddCallFinished(const std::map<absl::string_view, double>* named_metrics,
```
yousukseung, line 219: > Suggest making this parameter a reference instead of a pointer.

I switched to a pointer to make it nullable so that I can avoid allocating an empty map when there's no load report, which I think is worth it here. From what I read GetBackendMetricData() wil return null if "endpoint-load-metrics-bin" trailing metadata is missing.

src/core/ext/xds/xds_client_stats.cc
```diff
@@ -131,29 +131,43 @@ XdsClusterLocalityStats::~XdsClusterLocalityStats() {
...
-      GetAndResetCounter(&total_successful_requests_),
-      // Don't reset total_requests_in_progress because it's
-      // not related to a single reporting interval.
-      total_requests_in_progress_.load(std::memory_order_relaxed),
-      GetAndResetCounter(&total_error_requests_),
-      GetAndResetCounter(&total_issued_requests_),
-      {}};
-  MutexLock lock(&backend_metrics_mu_);
-  snapshot.backend_metrics = std::move(backend_metrics_);
+  Snapshot snapshot;
```
yousukseung, line 134: > No need for the = {}.
I don't see we initialize Snapshot fields, so believe total_* counters are undefined today. Instead I updated to explicitly initialize fields in the definition, I see no reason not to.

src/core/ext/xds/xds_client_stats.cc
```diff
@@ -131,29 +131,43 @@ XdsClusterLocalityStats::~XdsClusterLocalityStats() {
...
+    Snapshot percpu_snapshot = {
+        GetAndResetCounter(&percpu_stats.total_successful_requests),
+        // Don't reset total_requests_in_progress because it's
+        // not related to a single reporting interval.
+        percpu_stats.total_requests_in_progress.load(std::memory_order_relaxed),
+        GetAndResetCounter(&percpu_stats.total_error_requests),
+        GetAndResetCounter(&percpu_stats.total_issued_requests),
+        {}};
+    {
+      MutexLock lock(&percpu_stats.backend_metrics_mu);
```
yousukseung, line 145: > These two lines should go in their own nested scope, so that we're not still holding the lock while we add the per-CPU snapshot to the aggregated snapshot on line 146.

Fixed.

src/core/lib/gprpp/per_cpu.h
```diff
@@ -29,16 +31,20 @@ namespace grpc_core {
 template <typename T>
 class PerCpu {
  public:
-  T& this_cpu() { return data_[ExecCtx::Get()->starting_cpu()]; }
+  explicit PerCpu(size_t max = std::numeric_limits<size_t>::max())
```
yousukseung, line 34: > What's the purpose of this change? If there are actually more CPUs, don't we actually want that many more instances?
> If we do impose a limit on the total number of instances, what happens if this_cpu() gets called from a CPU that is greater than the limit? It seems like that would just cause a crash.

Thanks for catching this, fixed. Craig was concerned when there are too many cores. We actually did the same in Fathom in prod where we capped it at 16~32, so I picked 32 here, which I think will be fine.

test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -769,6 +783,12 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType> {
       return *this;
     }
 
+    RpcOptions& set_backend_metrics(
```
yousukseung, line 786: > I think this approach is inconsistent with the existing API pattern here...

Updated

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -1425,17 +1434,54 @@ TEST_P(ClientLoadReportingTest, Vanilla) {
                              ::testing::Pair("locality1", ::testing::_)));
   size_t num_successful_rpcs = 0;
   size_t num_failed_rpcs = 0;
+  std::map<std::string, ClientStats::LocalityStats::LoadMetric>
```
yousukseung, line 1437: > Instead of having two maps here, I suggest a single map with both values:

Updated.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -1425,17 +1434,54 @@ TEST_P(ClientLoadReportingTest, Vanilla) {
...
+                          (kNumFailuresPerAddress * backends_.size()) * 0.3)))),
+          ::testing::Pair(
+              "bar", ::testing::AllOf(
+                         ::testing::Field(
+                             &ClientStats::LocalityStats::LoadMetric::
+                                 num_requests_finished_with_metric,
+
+                             (kNumRpcsPerAddress + kNumFailuresPerAddress) *
+                                 backends_.size()),
+                         ::testing::Field(
```
yousukseung, line 1478: Empty line removed.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -1515,6 +1608,11 @@ TEST_P(ClientLoadReportingTest, BalancerRestart) {
   EXPECT_EQ(0U, client_stats.total_requests_in_progress());
   EXPECT_EQ(0U, client_stats.total_error_requests());
   EXPECT_EQ(0U, client_stats.total_dropped_requests());
+  ASSERT_THAT(client_stats.locality_stats(),
+              ::testing::ElementsAre(::testing::Pair(
```
yousukseung, line 1612: > Why not just replace the ::testing::_ with ::testing::IsEmpty(), so that you don't need a separate expectation immediately below?

Updated.

src/core/ext/xds/xds_client_stats.h
yousukseung: But it requires allocating an empty string unnecessarily unless we make this nullable in some way.
In XdsClusterImplLb::Picker::SubchannelCallTracker::Finish() where we call AddCallFinished():

```
auto* backend_metric_data = args.backend_metric_accessor->GetBackendMetricData();
const std::map<absl::string_view, double>* named_metrics = nullptr;
if (backend_metric_data != nullptr) {
  named_metrics = &backend_metric_data->named_metrics;
}
locality_stats_->AddCallFinished(named_metrics, !args.status.ok());
```

vs. Something like (pseudo-code)
```
auto* backend_metric_data = args.backend_metric_accessor->GetBackendMetricData();
std::map<absl::string_view, double> named_metrics
if (backend_metric_data == nullptr) {
  locality_stats_->AddCallFinished(std::map<..>(), !args.status.ok());
} else {
  locality_stats_->AddCallFinished(
    backend_metric_data->named_metrics, !args.status.ok());
}
```

src/core/ext/xds/xds_client_stats.h
```diff
@@ -215,27 +216,32 @@ class XdsClusterLocalityStats : public RefCounted<XdsClusterLocalityStats> {
...
  private:
+  struct Stats {
+    std::atomic<uint64_t> total_successful_requests{0};
+    std::atomic<uint64_t> total_requests_in_progress{0};
+    std::atomic<uint64_t> total_error_requests{0};
+    std::atomic<uint64_t> total_issued_requests{0};
+
+    // Protects backend_metrics. A mutex is necessary because the length of
+    // backend_metrics_ can be accessed by both the callback intercepting the
+    // call's recv_trailing_metadata (not from the control plane work
```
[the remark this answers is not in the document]
yousukseung: Done.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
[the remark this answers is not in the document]
yousukseung: Yes, will merge after #32657. For future PRs, what's the best practice to chain PRs in github, or do we discourage it?

test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -710,6 +722,8 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType> {
     int client_cancel_after_us = 0;
     bool skip_cancelled_check = false;
     StatusCode server_expected_error = StatusCode::OK;
+    absl::optional<xds::data::orca::v3::OrcaLoadReport> backend_metrics =
+        absl::nullopt;
```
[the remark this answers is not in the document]
yousukseung: Removed.

test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -769,6 +783,12 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType> {
       return *this;
     }
 
+    RpcOptions& set_backend_metrics(
+        const xds::data::orca::v3::OrcaLoadReport& metrics) {
```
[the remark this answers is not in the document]
yousukseung: Updated. We want to copy each time for sequential tests, so made a const ref of an optional.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -1424,17 +1433,54 @@ TEST_P(ClientLoadReportingTest, Vanilla) {
...
   }
   EXPECT_EQ(num_successful_rpcs, total_successful_rpcs_sent);
   EXPECT_EQ(num_failed_rpcs, total_failed_rpcs_sent);
   EXPECT_EQ(num_successful_rpcs + num_failed_rpcs, total_rpcs_sent);
+  EXPECT_THAT(
+      named_metrics_total,
+      ::testing::UnorderedElementsAre(
+          ::testing::Pair(
+              "foo",
+              ::testing::AllOf(
```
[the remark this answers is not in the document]
yousukseung: Done.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -1514,6 +1607,11 @@ TEST_P(ClientLoadReportingTest, BalancerRestart) {
   EXPECT_EQ(0U, client_stats.total_requests_in_progress());
   EXPECT_EQ(0U, client_stats.total_error_requests());
   EXPECT_EQ(0U, client_stats.total_dropped_requests());
+  ASSERT_THAT(client_stats.locality_stats(),
```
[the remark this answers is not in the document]
yousukseung: Done.

test/cpp/end2end/xds/xds_end2end_test_lib.h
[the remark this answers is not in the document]
markdroth: I don't understand the benefit of using a const reference here.

There are two possible cases for the caller:
1. They want to retain their own copy of the data they pass in.
2. They do not need to retain their own copy of the data they pass in.

In case 1, it does not matter if we use a const reference or not, because either way we will make our own copy.  If we use a const reference, we make the copy inside of the method where we assign to the data member; if we take the parameter by value, the copy is made when the method is invoked, and the method's body can use `std::move()` to move that copy to the data member.

In case 2, if we take the argument by value, then the caller can use `std::move()` when they invoke the method, and we can still use `std::move()` inside the method, so there is no copy at all.  However, if we use a const reference, then we always make a copy, even though we don't actually need one.

This isn't a big deal either way in this case, but it seems strictly better to pass it by value.  I'll leave it up to you.

src/core/ext/xds/xds_client_stats.h
yousukseung: I'll keep it then. Certainly no dynamic allocation, but I found this option looks slightly better: https://screenshot.googleplex.com/B5csTZpq6Eda2Gv

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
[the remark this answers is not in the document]
yousukseung: Ack.

test/cpp/end2end/xds/xds_end2end_test_lib.h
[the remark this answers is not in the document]
yousukseung: Makes sense, updated.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -1375,6 +1375,17 @@ INSTANTIATE_TEST_SUITE_P(
     ::testing::Values(XdsTestType().set_enable_load_reporting()),
     &XdsTestType::Name);
 
+MATCHER_P(LoadMetricEq, other, "equals LoadMetric") {
```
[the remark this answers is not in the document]
yousukseung: Done.

src/core/ext/xds/xds_client_stats.h
markdroth: Note that the first version in that screenshot is not the same as the version that I proposed in my last comment.  The version you're using imposes the overhead of a `std::move()`, which is not present in the version I proposed, so I think it should have less overhead.

In any case, though, as I said, this isn't worth holding up this PR over.

src/core/ext/filters/client_channel/lb_policy/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -445,7 +445,7 @@ float WeightedRoundRobin::AddressWeight::GetWeight(
     gpr_log(GPR_INFO,
             "[WRR %p] subchannel %s: getting weight: now=%s "
             "weight_expiration_period=%s blackout_period=%s "
-            "last_update_time_=%s non_empty_since_=%s weight_=%f",
+            "last_update_time=%s non_empty_since=%s weight=%f",
```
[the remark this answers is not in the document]
yousukseung: Something went wrong during `merge master` which gave me a "local copy is outdated" error, which I had to force-push. I think this change was related. Fixed and ack on force-push, I think in this case I can `merge origin` to avoid force-pushing later.

src/core/ext/xds/xds_client_stats.h
yousukseung: Right that we should only compare the empty case here, and it comes down to https://screenshot.googleplex.com/329NtyXoDS7yPZX. That said, I overestimated the cost of allocating an empty map in the stack which should be very cheap, however like you said I think we can keep it, which won't hurt. I will resolve thisl.

yousukseung:
The EPS PR #32657 has been rolled back, will hold until we roll it forward.

yousukseung:
> The EPS PR #32657 has been rolled back, will hold until we roll it forward.

@markdroth PR is now fully ready (pending tests).

markdroth (approved):
Looks great!

yousukseung:
> Looks great!

I removed arg name comments for the matcher instead of adding NOLINT()'s, it looks clang still cannot correctly handle MATCHER macros.

PR #32692, opened 2023-03-23
[round robin] delegate to pick_first instead of creating subchannels directly

markdroth:
More work on the dualstack backend design:
- Change round_robin to delegate to pick_first instead of creating subchannels directly.
- Change pick_first such that when it is the child of a petiole policy, it will unconditionally start a health watch.
- Change the client-side health checking code such that if client-side health checking is not enabled, it will return the subchannel's raw connectivity state.
- As part of this, we introduce a new endpoint_list library to be used by petiole policies, which is intended to replace the existing subchannel_list library.  The only policy that will still directly interact with subchannels is pick_first, so the relevant parts of the subchannel_list functionality have been copied directly into that policy.  The subchannel_list library will be removed after all petiole policies are updated to delegate to pick_first.


src/core/ext/filters/client_channel/lb_policy/endpoint_list.cc
```diff
@@ -0,0 +1,176 @@
...
+  // child policy. This will make the child policy progress upon activity on
+  // this policy, which in turn is tied to the application's call.
+  grpc_pollset_set_add_pollset_set(
+      child_policy_->interested_parties(),
+      endpoint_list_->policy_->interested_parties());
+  // Update child policy.
+  LoadBalancingPolicy::UpdateArgs update_args;
+  update_args.addresses.emplace().emplace_back(address);
+  update_args.args = child_args;
+  // TODO(roth): If the child reports a non-OK status with the update,
```
eugeneo, line 117: Should it at least be logged?

src/core/ext/filters/client_channel/lb_policy/endpoint_list.h
```diff
@@ -0,0 +1,211 @@
...
+namespace grpc_core {
+
+// A list of endpoints for use in a petiole LB policy.  Each endpoint may
+// have one or more addresses, which will be passed down to a pick_first
+// child policy.
+//
+// To use this, a petiole policy must define its own subclass of both
+// EndpointList and EndpointList::Endpoint, like so:
+/*
+class MyEndpointList : public EndpointList {
```
eugeneo, line 52: nit: Commented out code (I believe linter will point it out too)

eugeneo (approved):

src/core/ext/filters/client_channel/lb_policy/endpoint_list.cc
markdroth: In this situation, there's no point in logging it, because a non-OK status isn't actually a real failure; it's really just an indication of whether the resolver should consider the update a failure.  We actually have a broader pattern of this issue that occurs in many places, as introduced by #30809.  When we eventually find a good solution for this, we'll fix it everywhere it occurs.

In this particular case, it's actually currently impossible for this to return non-OK anyway, because we are creating one endpoint for each address, and the only case where pick_first will return non-OK is if the address list is empty.  However, that will change when we add support for multiple addresses per endpoint, so I've added a TODO in round_robin.cc to make sure that we properly handle that case.

src/core/ext/filters/client_channel/lb_policy/endpoint_list.h
markdroth: This commented-out code is intentional; it's showing an example of how to use this API.  I'm not aware of any linter that would consider this a problem.

eugeneo (approved):

PR #32697, opened 2023-03-23
Add Java v1.54.0 to the client matrix

temawi:
@sergiitk


sergiitk (changes requested):
Take it back - we need to run the adhoc test first.

sergiitk:
Test started - https://source.cloud.google.com/results/invocations/3d942ccb-010a-444b-ae1d-096b8d832e0e/targets

sergiitk (approved):

PR #32699, opened 2023-03-24
Gcp Observability Logging: The Call ID should use a UUIDv4 format

yashykt:
Earlier, we were simply using a 64 bit random number, but the spec actually calls for UUIDv4.


veblush (approved):
LGTM. Would you add unit tests for GenerateUUIDv4 function?

yashykt:
Done! Thanks to @ejona86 as well for checking the UUIDv4 generator code

PR #32706, opened 2023-03-24
Invalidate RBE linux cache by setting a new cache silo key

jtattermusch:
An attempt to fix b/275035917 (but I have no idea if this helps or not).


drfloob (approved):
Let's try. It might also be worth checking if artifacts with those hashes even exist in the cache bucket.

ctiller (approved):

ctiller:
If this helps it might be worth embedding the year/month in the key and then forgetting about this forever.

PR #32707, opened 2023-03-24
[Interop] Enable Orca test cases for C++ (#32684)

eugeneo:


ejona86 (approved):
https://github.com/grpc/grpc-java/pull/9989 has been merged in Java, so this should be safe.

PR #32709, opened 2023-03-24
[client channel] move health checking code out of subchannel and into LB policies

markdroth:
This paves the way for making pick_first the universal leaf policy (see #32692), which will be needed for the dualstack design.  That change will require changing pick_first to see both the raw connectivity state and the health-checking connectivity state of a subchannel, so that we can enable health checking when pick_first is used underneath round_robin without actually changing the pick_first connectivity logic (currently, pick_first always disables health checking).  To make it possible to do that, this PR moves the health checking code out of the subchannel and into a separate API using the same data-watcher mechanism that was added for ORCA OOB calls.


src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -130,6 +131,10 @@ class PickFirst : public LoadBalancingPolicy {
     void set_attempting_index(size_t index) { attempting_index_ = index; }
 
    private:
+    std::shared_ptr<WorkSerializer> work_serializer() const override {
```
eugeneo, line 134: Can this be moved to a base class?

eugeneo (approved):

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
markdroth: Unfortunately, it can't.  The problem is that `LoadBalancingPolicy::work_serializer()` is protected, so it can be accessed only in subclasses.  And the base `SubchannelList` class is not nested inside of a subclass of `LoadBalancingPolicy`, but the individual subclasses of `SubchannelList` are nested inside of subclasses of `LoadBalancingPolicy`, so this is where it needs to be.

I agree that this is ugly.  However, I plan to clean this up as part of the changes I'm going to be making for the dualstack design.  Once I make pick_first the universal leaf policy, it will be the only policy actually using `SubchannelList`, and at that point I will move the code directly into pick_first so that it no longer needs to be a separate base class.

PR #32710, opened 2023-03-24
Gcp Observability: Add example

yashykt:

### DIFF
--- a/examples/cpp/gcp_observability/helloworld/README.md
+++ b/examples/cpp/gcp_observability/helloworld/README.md
@@ -1,0 +1,30 @@
+# gRPC C++ GCP Observability Hello World Example
+
+This example consists of a hello world client and a hello world server instrumented with GCP Observability for logs, metrics and tracing. Note that familiarity with the [basic hello world][] example is assumed.
+
+Please refer to GCP's Microservices Observability user guide for setup 
+instructions.
+
+[basic hello world]: https://grpc.io/docs/languages/cpp/quickstart
+
+### Run the example with configuration
+
+To use Observability, you should first setup and configure authorization as mentioned in the Microservices Observability user guide. 
+
+You need to set the `GRPC_GCP_OBSERVABILITY_CONFIG_FILE` environment variable to point to the gRPC GCP Observability configuration file (preferred) or alternatively set `GRPC_GCP_OBSERVABILITY_CONFIG` environment variable to gRPC GCP Observability configuration value. This is needed by both
+`build/install/example-gcp-observability/bin/gcp-observability-client` and
+`build/install/example-gcp-observability/bin/gcp-observability-server`.
+
+Sample configurations are provided with the example.
+
+1. To start the observability-enabled example server on its default port of 50051, run:
+```
+$ export GRPC_GCP_OBSERVABILITY_CONFIG_FILE=${pwd}/examples/cpp/gcp_observability/helloworld/server_config.json
+$ bazel run examples/cpp/gcp_observability/helloworld:greeter_server
+```
+
+2. In a different terminal window, run the observability-enabled example client:
+```
+$ export GRPC_GCP_OBSERVABILITY_CONFIG_FILE=${pwd}/examples/cpp/gcp_observability/helloworld/client_config.json
+$ bazel run examples/cpp/gcp_observability/helloworld:greeter_client
+```
--- a/examples/cpp/gcp_observability/helloworld/greeter_client.cc
+++ b/examples/cpp/gcp_observability/helloworld/greeter_client.cc
@@ -1,0 +1,115 @@
+//
+//
+// Copyright 2023 gRPC authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//
+
+#include <chrono>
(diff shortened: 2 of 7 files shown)


examples/cpp/gcp_observability/helloworld/greeter_client.cc
```diff
@@ -0,0 +1,115 @@
...
+  std::string target_str = absl::GetFlag(FLAGS_target);
+  // Turn on GCP Observability for the whole binary. Note that this should be
+  // done before any other gRPC operation.
+  auto status = grpc::experimental::GcpObservabilityInit();
+  if (!status.ok()) {
+    std::cerr << "GcpObservabilityInit() failed: " << status.ToString()
+              << std::endl;
+    return static_cast<int>(status.code());
+  }
+  std::cout << "Initialized GCP Observability" << std::endl;
```
zasweq, line 96: If we print something to standard out we should keep it consistent across the 3 languages.

examples/cpp/gcp_observability/helloworld/greeter_server.cc
```diff
@@ -0,0 +1,95 @@
...
+
+  // Wait for the server to shutdown. Note that some other thread must be
+  // responsible for shutting down the server for this call to ever return.
+  server->Wait();
+}
+
+int main(int argc, char** argv) {
+  absl::ParseCommandLine(argc, argv);
+  // Turn on GCP Observability for the whole binary. Note that this should be
+  // done before any other gRPC operation.
```
zasweq, line 81: Same comment here as for server.

examples/cpp/gcp_observability/helloworld/greeter_server.cc
```diff
@@ -0,0 +1,95 @@
...
+    std::cerr << "GcpObservabilityInit() failed: " << status.ToString()
+              << std::endl;
+    return static_cast<int>(status.code());
+  }
+  std::cout << "Initialized GCP Observability" << std::endl;
+  RunServer(absl::GetFlag(FLAGS_port));
+  // Note that RunServer() never returns and hence there is no need for a
+  // GcpObservabilityClose() here, but as on the client side, we need to wait
+  // atleast 25seconds after the last RPC to make sure that stats and tracing
+  // are flushed.
```
zasweq, line 93: See the golang example. We catch ctrl c signals (and another exit signal), block on that (since that's only way to kill server binary) and then use that even to trigger server.Stop() exiting. server.Stop() exiting causes main() to finish and thus run our o11y cleanup (which flushes all 3) in a defer from main.

examples/cpp/gcp_observability/helloworld/README.md
```diff
@@ -0,0 +1,34 @@
...
+### Run the example with configuration
+
+To use Observability, you should first setup and configure authorization as
+mentioned in the Microservices Observability user guide.
+
+You need to set the `GRPC_GCP_OBSERVABILITY_CONFIG_FILE` environment variable to
+point to the gRPC GCP Observability configuration file (preferred) or
+alternatively set `GRPC_GCP_OBSERVABILITY_CONFIG` environment variable to gRPC
+GCP Observability configuration value. This is needed by both
+`build/install/example-gcp-observability/bin/gcp-observability-client` and
```
DNVindhya, line 21: This looks like java examples binary PATH, you might need to update this?

examples/cpp/gcp_observability/helloworld/greeter_client.cc
```diff
@@ -0,0 +1,115 @@
...
+  // Flush out any pending Observability data
+  std::cout << "Closing GCP Observability" << std::endl;
+  grpc::experimental::GcpObservabilityClose();
+  std::cout << "Sleeping for 25 seconds to make sure Observability stats and "
+               "tracing are flushed. Don't shut off server either."
+            << std::endl;
+  // Currently, GcpObservabilityClose() only supports flushing logs. Stats and
+  // tracing get automatically flushed at a regular interval, so sleep for an
+  // interval to make sure that those are flushed too.
+  std::this_thread::sleep_for(std::chrono::seconds(25));
```
DNVindhya, line 113: Java decided to do add sleep in `close()`, which can be removed later once we have a fix so that shutdown would run faster.

examples/cpp/gcp_observability/helloworld/greeter_client.cc
yashykt: I think that's nice but not necessary

examples/cpp/gcp_observability/helloworld/greeter_client.cc
yashykt: After talking with @fengli79, the decision was not to include the sleep in `close()`

examples/cpp/gcp_observability/helloworld/greeter_server.cc
yashykt: Alright, I'll take your comment.. was feeling lazy earlier.

examples/cpp/gcp_observability/helloworld/README.md
yashykt: Yeah, thanks!

examples/cpp/gcp_observability/helloworld/greeter_client.cc
```diff
@@ -0,0 +1,115 @@
...
+  std::unique_ptr<Greeter::Stub> stub_;
+};
+
+int main(int argc, char** argv) {
+  absl::ParseCommandLine(argc, argv);
+  // Instantiate the client. It requires a channel, out of which the actual RPCs
+  // are created. This channel models a connection to an endpoint specified by
+  // the argument "--target=" which is the only expected argument.
+  std::string target_str = absl::GetFlag(FLAGS_target);
+  // Turn on GCP Observability for the whole binary. Note that this should be
```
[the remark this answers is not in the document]
yashykt: added some more

examples/cpp/gcp_observability/helloworld/greeter_server.cc
yashykt: added some more

PR #32711, opened 2023-03-24
XDS: enable XDS federation by default

apolcyn:
Integration tests have been green so let's enable this (verification of test results in https://b.corp.google.com/issues/262593165#comment30).

### DIFF
--- a/src/core/ext/xds/xds_bootstrap.cc
+++ b/src/core/ext/xds/xds_bootstrap.cc
@@ -25,8 +25,8 @@
 
 namespace grpc_core {
 
-// TODO(donnadionne): check to see if federation is enabled, this will be
-// removed once federation is fully integrated and enabled by default.
+// TODO(roth,apolcyn): remove this federation env var after the 1.55
+// release.
 bool XdsFederationEnabled() {
   auto value = GetEnv("GRPC_EXPERIMENTAL_XDS_FEDERATION");
   if (!value.has_value()) return true;
--- a/src/core/ext/xds/xds_bootstrap.cc
+++ b/src/core/ext/xds/xds_bootstrap.cc
@@ -29,7 +29,7 @@ namespace grpc_core {
 // removed once federation is fully integrated and enabled by default.
 bool XdsFederationEnabled() {
   auto value = GetEnv("GRPC_EXPERIMENTAL_XDS_FEDERATION");
-  if (!value.has_value()) return false;
+  if (!value.has_value()) return true;
   bool parsed_value;
   bool parse_succeeded = gpr_parse_bool_value(value->c_str(), &parsed_value);
   return parse_succeeded && parsed_value;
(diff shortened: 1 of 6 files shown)


src/core/ext/xds/xds_bootstrap.cc
```diff
@@ -29,7 +29,7 @@ namespace grpc_core {
 // removed once federation is fully integrated and enabled by default.
```
markdroth, line 29: Historically, in C-core, we've just completely removed the env var instead of changing the default to true, on the theory that once we pass interop tests, we have confidence that there won't be any more need to disable the feature.  However, in this particular case, since we've already had a fairly serious bug with this code, it probably does make sense to be more cautious here, so I'm okay with this approach.

However, we need to remember to eventually come back and remove this completely, or else we'll just keep accumulating env var controls that we don't actually need.  Please update the TODO here to indicate that this should be removed after the 1.55 release, just so that we have a target date for removing this.

markdroth (commented):
There's at least one test that will need to be changed to explicitly disable federation:

https://github.com/grpc/grpc/blob/4537dc4dc3a5d76f0e5d345e864ebeb7a283d15c/test/cpp/end2end/xds/xds_core_end2end_test.cc#L1156

There may be others as well -- please make sure everything is passing with this PR.

src/core/ext/xds/xds_bootstrap.cc
apolcyn: done, updated the todo

apolcyn:
Manually tested the previously-failed tests - they should pass now.

PTAL

test/core/xds/xds_client_test.cc
```diff
@@ -2533,11 +2517,15 @@ TEST_F(XdsClientTest, FederationWithUnparseableXdstpResourceName) {
...
 }
 
+// TODO(roth,apolcyn): remove this test when the
+// GRPC_EXPERIMENTAL_XDS_FEDERATION env var is removed.
 TEST_F(XdsClientTest, FederationDisabledWithNewStyleName) {
   // We will use this xdstp name, whose authority is not present in
   // the bootstrap config.  But since federation is not enabled, we
   // will treat this as an opaque old-style name, so we'll send it to
   // the default server.
+  grpc_core::testing::ScopedEnvVar env_var("GRPC_EXPERIMENTAL_XDS_FEDERATION",
```
markdroth, line 2527: Nit: Please move this up to line 2523, so it's not between the comment above and the line that the comment refers to.

markdroth (approved):

test/core/xds/xds_client_test.cc
apolcyn: done

apolcyn:
Android failure `fatal error: 'src/core/lib/security/security_connector/ssl_utils_config.h' file not found` is unrelated

PR #32712, opened 2023-03-24
[Infra] Use Bazel 6 and drop Bazel 4

veblush:
Let's move on as we'd agreed to drop 4.x


jtattermusch (approved):
Fair enough, but I'd say let's switch to bazel 6 first (since we have to do that anyway because of b/275035917) and if everything looks good, merge this.

PR #32715, opened 2023-03-24
[GCP Observability  C++] De-experimentalize API

yashykt:
This PR aims to de-experimentalize the APIs for GCP Observability. 

We would have ideally wanted public feedback before declaring the APIs stable, but we need stable APIs for GA.

Changes made after API review with @markdroth @veblush, @ctiller and the entire Core/C++ team -
* The old experimental APIs `grpc::experimental::GcpObservabilityInit` and `grpc::experimental::GcpObservabilityClose` are now deprecated and will be deleted after v.1.55 release.
* The new API gets rid of the Close method and follows the RAII idiom with a single `grpc::GcpObservability::Init()` call that returns an `GcpObservability` object, the lifetime of which controls when observability data is flushed.
* The `GcpObservability` class could in the future add more methods. For example, a debug method that shows the current configuration.
* Document that GcpObservability initialization and flushing (on `GcpObservability` destruction) are blocking calls.
* Document that gRPC is still usable if GcpObservability initialization failed. (Added a test to prove the same).
* Since we don't have a good way to flush stats and tracing with OpenCensus, the examples required users to sleep for 25 seconds. This sleep is now part of `GcpObservability` destruction.

Additional Implementation details -
* `GcpObservability::Init` is now marked with `GRPC_MUST_USE_RESULT` to make sure that the results are used. We ideally want users to store it, but this is better than nothing.
* Added a note on GCP Observability lifetime guarantees.


veblush (approved):

include/grpcpp/ext/gcp_observability.h
```diff
@@ -17,25 +17,76 @@
...
 
+#include <grpc/support/port_platform.h>
+
 #include "absl/status/status.h"
+#include "absl/status/statusor.h"
 
 namespace grpc {
-namespace experimental {
+
+namespace grpc_gcp {
```
ctiller, line 27: I think if we name it `grpc_gcp` we should probably just drop the top-level namesapce.

include/grpcpp/ext/gcp_observability.h
```diff
@@ -17,25 +17,76 @@
...
+// Note that this is a blocking call which properly sets up gRPC Observability
+// to work with GCP and might take a few seconds to return.  Similarly, the
+// destruction of a non-moved-from `Observability` object is also blocking since
+// it flushes the observability data to GCP.
+//
 // As an implementation detail, this properly initializes the OpenCensus stats
 // and tracing plugin, so applications do not need to perform any additional
 // gRPC C++ OpenCensus setup/registration to get GCP Observability for gRPC.
-absl::Status GcpObservabilityInit();
+absl::StatusOr<Observability> ObservabilityInit();
```
ctiller, line 78: Ok with this, but unconvinced it needs to be `absl::StatusOr` -- my expectation here is that `absl::optional` would be just find for real usage, and mean that folks don't need to interpret status codes.

If we keep `StatusOr` here, I'd encourage some documentation around how to interpret the status codes.

include/grpcpp/ext/gcp_observability.h
```diff
@@ -17,25 +17,76 @@
...
+// return an `Observability` instance. Observability data is flushed at regular
+// intervals, and also when this instance goes out of scope and its destructor
+// is invoked.
+class Observability {
+ public:
+  ~Observability();
+
+  // Move constructor. The moved-from object will no longer be valid and will
+  // not cause GCP Observability stats, tracing and logging data to flush.
+  Observability(Observability&& other) noexcept;
```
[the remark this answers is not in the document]
yashykt: The main reason for the move ctor is to allow people to return the observability object from a function or save it somewhere.

Providing a move-assignment operator assumes that the user was able to create an Observability object and they are now overwriting it with a new Observability object, which does not make sense. (Note that the constructor is private.)

include/grpcpp/ext/gcp_observability.h
yashykt: I have the same opinion as Mark when preferring absl::StatusOr<> over absl::optional<> here because of the human-readable error message.
I do think though that in the future, we *might* be able to provide some explicit guidance based on the status code, for example, retry if unvailable. For now, though, I'll add that users should not make any assumptions about the behavior based on the resulting status code.

include/grpcpp/ext/gcp_observability.h
veblush: Craig suggestion might make sense. Either would work.

Having a different namespace would be helpful because there is a clear boundary between vanilla gRPC and GCP specific extensions especially those extensions require an additional dependency. I hope that we can also put `GoogleDefaultCredentials` under gcp namespace but we may have to keep this for backward-compatibility issue. (or we can have both)

yashykt: Moved to `grpc::GcpObservability::Init` as discussed

include/grpcpp/ext/gcp_observability.h
```diff
@@ -22,15 +22,35 @@
...
+// initialization goes away, we might still want to keep gRPC Event Engine
+// initialized, just in case, we need to perform some IO operations during
+// observability close.
+// Note that the lifetime guarantees are only one way, i.e., GcpObservability
+// object guarantees that gRPC will not shutdown while the object is still in
+// scope, but the other way around does not hold true. Even though that is not
+// the expected usage, GCP Observability can shutdown before gRPC shuts down. It
+// follows that gRPC should not hold any callbacks from GcpObservability. A
+// change in this restriction should go through a design review.
+class GcpObservabilityImpl : private GrpcLibrary {
```
markdroth, line 42: This class can be private within `GcpObservability` rather than being in a separate `internal` namespace.

markdroth (approved):

include/grpcpp/ext/gcp_observability.h
yashykt: Done. Thanks!

yashykt: Done

veblush (approved):

include/grpcpp/ext/gcp_observability.h
```diff
@@ -17,25 +17,91 @@
...
+class GcpObservability {
+ public:
+  // Initialize GCP Observability for gRPC.
+  // This should be called before any other gRPC operations like creating a
+  // channel, server, credentials etc.
+  // The most common usage would call this at the top (or near the top) in
+  // main(). The return value helps determine whether observability was
+  // successfully enabled or not. On success, an object of class `Observability`
+  // is returned. When this object goes out of scope, GCP Observability stats,
+  // tracing and logging data is flushed. On failure, the status message can be
```
ctiller, line 43: Is it safe to do this (drop `GcpObservability`) whilst still making calls with gRPC?
What is the expected behavior?

yashykt: Current behavior - Just sleep for 25 sec to make sure current traces and stats are flushed. Logs are flushed and no more stuff is logged.

Possible future behavior with OpenTelemetry - Flush and stop exporting metrics for traces and stats. Same behavior for logs.

yashykt: I'll update the docs.

yashykt:
Thanks for providing reviews at such short notices! :pray:

PR #32724, opened 2023-03-27
[EventEngine] Always initialize PosixEventEngine when created

drfloob:
The `NeedPosixEngine' check broke external users of `CreateEventEngine` and `GetDefaultEventEngine` in release v1.53. As a workaround, users can set the environment variable `GRPC_EXPERIMENTS=event_engine_client` to use the engine, presuming gRPC has been initialized. But fundamentally, if an engine is created, it should be fully functional regardless of whether gRPC-core experiments are on or off.


Vignesh2208 (approved):

drfloob:
All python fork tests look good. https://source.cloud.google.com/results/invocations/f822425d-837c-4f10-9d7c-8634d2d783cb/targets;query=fork

PR #32726, opened 2023-03-27
[EventEngine] RunAfter migration: grpc_chttp2_transport round 2

yijiem:
Missed this one in round 1...

### DIFF
--- a/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
+++ b/src/core/ext/transport/chttp2/transport/chttp2_transport.cc
@@ -1691,8 +1691,7 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
           // The ref will be unreffed in the combiner.
           auto* ptr = self.release();
           ptr->t_->combiner->Run(
-              GRPC_CLOSURE_INIT(&ptr->on_timer_, OnTimerLocked, ptr,
-                                nullptr),
+              GRPC_CLOSURE_INIT(&ptr->on_timer_, OnTimerLocked, ptr, nullptr),
               absl::OkStatus());
         });
   }


src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1725,31 +1729,32 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
...
 
-  static void OnTimer(void* arg, grpc_error_handle error) {
-    auto* self = static_cast<GracefulGoaway*>(arg);
-    if (!error.ok()) {
-      self->Unref();
-      return;
-    }
-    self->t_->combiner->Run(
-        GRPC_CLOSURE_INIT(&self->on_timer_, OnTimerLocked, self, nullptr),
+  void OnTimer() {
```
yijiem, line 1742: Additional note: after the timer callback fired (a), the handle is erased from the engine's known handles, but we won't reset `timer_handle_` until inside the combiner (b). If `OnPingAckLocked` happens between (a) and (b), this shouldn't cause issue with `Cancel` cancelling the wrong task because of the aba_token in the handle.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1683,9 +1684,12 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
...
     grpc_chttp2_initiate_write(t, GRPC_CHTTP2_INITIATE_WRITE_GOAWAY_SENT);
     Ref().release();  // Ref for the timer
-    grpc_timer_init(
-        &timer_, grpc_core::Timestamp::Now() + grpc_core::Duration::Seconds(20),
-        GRPC_CLOSURE_INIT(&on_timer_, OnTimer, this, nullptr));
+    timer_handle_ =
+        t_->event_engine->RunAfter(grpc_core::Duration::Seconds(20), [this] {
+          grpc_core::ApplicationCallbackExecCtx callback_exec_ctx;
+          grpc_core::ExecCtx exec_ctx;
+          OnTimer();
```
drfloob, line 1691: Since `OnTimer` is just one statement, and isn't called from anywhere else, you could move its body here and delete that function.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1725,31 +1729,32 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
 
   static void OnPingAckLocked(void* arg, grpc_error_handle /* error */) {
     auto* self = static_cast<GracefulGoaway*>(arg);
-    grpc_timer_cancel(&self->timer_);
+    if (self->timer_handle_ != TaskHandle::kInvalid) {
```
drfloob, line 1732: Once you let the callback manage the ref lifetime, I think this whole chunk of code can condense down to just:
```
self->t_->event_engine->Cancel(std::exchange(self->timer_handle_, TaskHandle::kInvalid));
```

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
yijiem: Done.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1683,9 +1684,12 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
         t, nullptr, GRPC_CLOSURE_INIT(&on_ping_ack_, OnPingAck, this, nullptr));
     grpc_chttp2_initiate_write(t, GRPC_CHTTP2_INITIATE_WRITE_GOAWAY_SENT);
     Ref().release();  // Ref for the timer
-    grpc_timer_init(
-        &timer_, grpc_core::Timestamp::Now() + grpc_core::Duration::Seconds(20),
-        GRPC_CLOSURE_INIT(&on_timer_, OnTimer, this, nullptr));
+    timer_handle_ =
+        t_->event_engine->RunAfter(grpc_core::Duration::Seconds(20), [this] {
```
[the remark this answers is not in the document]
yijiem: Yep, done. I need to grab a Ref again inside the callback for the combiner->Run().

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
yijiem: This is slick! Thanks!

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1675,17 +1674,26 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
...
-    Ref().release();  // Ref for the timer
-    grpc_timer_init(
-        &timer_, grpc_core::Timestamp::Now() + grpc_core::Duration::Seconds(20),
-        GRPC_CLOSURE_INIT(&on_timer_, OnTimer, this, nullptr));
+    timer_handle_ = t_->event_engine->RunAfter(
+        grpc_core::Duration::Seconds(20),
+        [self = Ref(DEBUG_LOCATION, "GoawayTimer")] {
+          grpc_core::ApplicationCallbackExecCtx callback_exec_ctx;
+          grpc_core::ExecCtx exec_ctx;
+          self->Ref().release();
```
yashykt, line 1691: something like `auto* ptr = self.release();` would be better and avoid the unnecessary Ref() and then Unref()?

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -1725,31 +1733,22 @@ class GracefulGoaway : public grpc_core::RefCounted<GracefulGoaway> {
...
-      return;
-    }
-    self->t_->combiner->Run(
-        GRPC_CLOSURE_INIT(&self->on_timer_, OnTimerLocked, self, nullptr),
-        absl::OkStatus());
-  }
-
   static void OnTimerLocked(void* arg, grpc_error_handle /* error */) {
     auto* self = static_cast<GracefulGoaway*>(arg);
+    self->timer_handle_ = TaskHandle::kInvalid;
```
yashykt, line 1744: what's the significance of setting `timer_handle_` to `kInvalid`? Is it just a sanity thing? If so, can you add a quick comment please?

yashykt (approved):
Approved with comments.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
yijiem: hmmm, I thought closures pushed onto the combiner might be got picked up and ran from another thread (not necessarily when the current exec_ctx destroys)? If that happens, we would want to take another Ref here. (and Unref in the combiner to keep the transport alive) Does my understanding correct?

drfloob (approved):

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
yijiem: Great point! Just realize it (after the discussion) that if the timer closure is invoked, it's convenient to just release the RefCounterPtr in the closure and preserve the ref for the combiner (since the combiner will guarantee to be called).

If the timer closure is not invoked (e.g. cancelled), the RefCountedPtr will be destroyed when the closure is destroyed.

Issue #32700, opened 2023-03-24
Much CPU time consumed on LockedMultiProducerSingleConsumerQueue::TryPop. 


LeiW000:
Testing Environment
gRPC version: 1.52.1
[grpc_bench](https://github.com/LesnyRumcajs/grpc_bench)

Configs of grpc_bench:
GRPC_BENCHMARK_DURATION  120s     
GRPC_BENCHMARK_WARMUP 30s     
GRPC_SERVER_RAM 512m      
GRPC_CLIENT_CONNECTIONS 50     
GRPC_CLIENT_CONCURRENCY 1000      
GRPC_CLIENT_CPUS 10 
GPRC_SERVER_CPUS [40 | 150]   
GRPC_REQUEST_SCENARIO=complex_proto     

I run [cpp_grpc_mt_bench](https://github.com/LesnyRumcajs/grpc_bench/tree/master/cpp_grpc_mt_bench) of grpc_bench to benchmark the performance of gRPC multi-threads on a system. I saw that lots of CPU time were consumed on LockedMultiProducerSingleConsumerQueue::TryPop() with the thread increasing. Please see below. 
40 threads:   5.12%    
150 threads: 33.7% 

I generated the [FlameGraph](https://github.com/brendangregg/FlameGraph). Please see the attached files. It seems to me that there is a locking contention. If it can be reduced, the RPS could be improved. 

![perf-40-threads-cpp-mt](https://user-images.githubusercontent.com/51774176/227403340-4e5c75df-0095-4d1b-a2c2-4b44b06e0a45.svg)
![perf-150-threads-cpp-mt](https://user-images.githubusercontent.com/51774176/227403356-4b36193d-38f2-436c-8ee2-e71bfff2b62a.svg)

LeiW000:
I suspect that the code below in MatchOrQueue() causes the lock contention. All threads try to call TryPop(), which calls mu_.TryLock().

```
     for (size_t i = 0; i < requests_per_cq_.size(); i++) {
       size_t cq_idx = (start_request_queue_index + i) % requests_per_cq_.size();
       RequestedCall* rc =
           reinterpret_cast<RequestedCall*>(requests_per_cq_[cq_idx].TryPop());
       if (rc != nullptr) {
         calld->SetState(CallData::CallState::ACTIVATED);
         calld->Publish(cq_idx, rc);
         return;
       }
     }
```

LeiW000:
[...]

PR #32732, opened 2023-03-28
The changes to MatchOrQueue() to improve multi-threads performace.

LeiW000:
If multiple threads call LockedMultiProducerSingleConsumerQueue::TryPop() at the same time, much of cpu time will be consumed on Mutex::TryLock().

For the details of the issue, please see #32700.

To reduce calls of LockedMultiProducerSingleConsumerQueue::TryPop(), don't try all the queues for a request, but only one, which belongs to the current thread. If fails, goes to the slow path.

With the change, I see ~36% performance gain with multi-threads.

I run cpp_grpc_mt_bench of grpc_bench with gRPC 1.52.1. The following is the data from my tests.
Base: gRPC 1.52.1
Opt: gRPC 1.52.1 + the PR
image


veblush:
This doesn't seem right to me as it just checked the single CQ. @ctiller PTAL.

LeiW000:
All the CQs are still checked in the slow path.

ctiller:
I think the benchmarking done so far is flawed, and I'd ask for a reproduction to be done using the qps suite in this repository.

Specifically the benchmark mentioned has a maximum of one outstanding call request per thread: https://github.com/LesnyRumcajs/grpc_bench/blob/master/cpp_grpc_mt_bench/main.cpp#L153

This is the worst possible case for the gRPC C++ implementation: we rely on a pool of call requests that never depletes for performance - and use the depletion of said pool to signal it's time for pushback. So the grpc_bench benchmark is continuously flipflopping between pushback and non-pushback mode.

A better designed benchmark would request a large number of requests (1000?, 10000?, somewhere around there) and replenish each time it completes one.

PR #32734, opened 2023-03-28
[csharp] Initial C# codegen tests.

tonydnewell:
Initial bazel tests for C# protoc and grpc_protoc_plugin.

This initial test just generated code from the proto file and compares the generated code against expected files.

I've put the tests in `test/csharp/codegen` as that is similar to where the C++ tests are placed, but they could be moved to 
`src\csharp` if that is a better place.

Further tests can be added once the initial framework for the tests is agreed.


test/csharp/codegen/BUILD
```diff
@@ -0,0 +1,28 @@
...
+
+licenses(["notice"])
+
+grpc_sh_test(
+    name = "csharp_codegen_simple_test",
+    srcs = ["csharp_codegen_simple_test.sh"],
+    data = [
+        "//src/compiler:grpc_csharp_plugin",
+        "@com_google_protobuf//:protoc",
+    ] + glob(["simple/**"]),
```
jtattermusch, line 25: I think glob() is considered an antipattern with bazel. It's better to simply list the actual files you'll use.

test/csharp/codegen/simple/proto/helloworld.proto
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "io.grpc.examples.helloworld";
+option java_outer_classname = "HelloWorldProto";
+option objc_class_prefix = "HLW";
+
+package helloworld;
```
jtattermusch, line 22: I think you'll need to use package `test.csharp.codegen.simple.proto.helloworld` to avoid internal conflict between .proto files.

test/csharp/codegen/csharp_codegen_simple_test.sh
```diff
@@ -0,0 +1,74 @@
...
+}
+[ -e ${PROTO_OUT}/HelloworldGrpc.cs ] || {
+    echo >&2 "missing generated output, expecting HelloworldGrpc"
+    exit 1
+}
+
+DIFFCMD="diff --strip-trailing-cr"
+
+# diff expected files
+[ "`$DIFFCMD ${DATA_DIR}/expected/Helloworld.cs \
```
jtattermusch, line 61: I think for now, we can get rid of Helloworld.cs and only test the code generated by grpc_csharp_plugin - which is what we should primarily be testing in this repo (we can add output of protoc as a followup if we want to).

test/csharp/codegen/BUILD
```diff
@@ -0,0 +1,28 @@
...
+    uses_polling = False,
```
jtattermusch, line 26: Adding 
```
tags = [
        "no_windows",
],
```

will automatically skip the test on windows (and I think we want that for now).

jtattermusch: while at it, you should probably also add `noasan`, `notsan`, `noubsan`, `nomsan` since the sanitizers are failing on the CI and we don't need any of them (at least initially).

test/csharp/codegen/csharp_codegen_simple_test.sh
```diff
@@ -0,0 +1,74 @@
...
+             ${PROTO_OUT}/Helloworld.cs`" ] && {
+    echo >&2 "Generated code does not match for Helloworld.cs"
+    exit 1
+}
+[ "`$DIFFCMD ${DATA_DIR}/expected/HelloworldGrpc.cs \
```
jtattermusch, line 66: style: diff returns exit code depending on whether the files match or not.  checking whether diff output is empty feels dirty.

jtattermusch (commented):
overall looks good, but added a few comments to address. Also note that the BUILD file needs to be properly formatted  (see failing sanity test) and the shell script needs to pass shellcheck (also see failing sanity test).

tonydnewell:
@jtattermusch I've updated from the review comments.
The current CI check that is failing doesn't seem to be related.

test/csharp/codegen/BUILD
```diff
@@ -0,0 +1,36 @@
...
+    srcs = ["csharp_codegen_simple_test.sh"],
+    data = [
+        "simple/expected/HelloworldGrpc.cs",
+        "simple/proto/helloworld.proto",
+        "//src/compiler:grpc_csharp_plugin",
+        "@com_google_protobuf//:protoc",
+    ],
+    tags = [
+        "no_windows",
+        "noasan",
```
jtattermusch, line 30: FTR the asan, tsan, msan and ubsan issues can also be resolved by adding `//test/core/util:grpc_suppressions` as a dependency I think.

jtattermusch: But for now we can leave as is.

jtattermusch (approved):
LGTM.

The ASAN failure is unrelated.

I think this will become mergeable once the internal change cl/524263844 is submitted (I'll then recheck and merge).

jtattermusch:
(I rebased and force-pushed to get a fresh test run).

PR #32748, opened 2023-03-29
[test-infra] Switch RBE linux build to a new custom image rbe_ubuntu2004.

jtattermusch:
- Add a new docker image "rbe_ubuntu2004" that is built in a way that's analogous to how our other testing docker images are built (this gives us control over what exactly is contained in the docker image and ability to fine-tune our RBE configuration)
- Switch RBE on linux to the new image (which gives us ubuntu20.04-based builds)

For some reason, RBE seems to have trouble pulling the docker image from Google Artifact Registry (GAR), which is where our public testing images normally live, so for now, I used a workaround and I upload a copy of the rbe_ubuntu2004 docker image to GCR as well, and that makes RBE works just fine (see comment in the `renerate_linux_rbe_configs.sh` script).

More followup items (config cleanup, getting local sanitizer builds working etc.) are in go/grpc-rbe-tech-debt-2023


tools/remote_build/linux.bazelrc
```diff
@@ -54,6 +54,10 @@ build:msan --test_timeout=60,900,1800,3600
 build:msan --test_tag_filters=-no_linux,-nomsan,-json_run_localhost
 build:msan --cxxopt=--stdlib=libc++
 build:msan --linkopt=--stdlib=libc++
+# use MSAN-instrumented version of libc++
+# TODO(jtattermusch): include the modified LD_LIBRARY_PATH in an msan-specific
+# toolchain configuration, instead of needing to explicitly set it here.
+build:msan --action_env=LD_LIBRARY_PATH=/usr/local/libcxx-msan/lib/x86_64-unknown-linux-gnu
```
jtattermusch, line 60: qq: until now, we haven't been setting any special flavor of libc++ for TSAN. Since the rbe_ubuntu2004 image now has also the TSAN-instrumented version of libstdc++ installed, it might be a good idea to use it. I'm leaving that as a possible followup (I don't want to change more stuff than necessary in this PR).

jtattermusch:
Issues to resolve:
The `examine_stack_test` is failing with ASAN (for unknown reason, perhaps something missing in the RBE docker image?)
```
test/core/gprpp/examine_stack_test.cc:71
Value of: stack_trace->find("GetCurrentStackTrace") != std::string::npos
  Actual: false
Expected: true
```
https://source.cloud.google.com/results/invocations/b07c5649-3972-476d-a581-06d76b5d67d1/targets;collapsed=/%2F%2Ftest%2Fcore%2Fgprpp:examine_stack_test/tests

veblush (approved):

veblush:
It looks like it's now passing all Bazel RBE tests so it's worth being merged.

jtattermusch:
Thanks @ctiller  for unblocking this!

PR #32750, opened 2023-03-29
[Audit Logging] Logger and factory APIs in C-Core and C++.

rockspore:
Audit logging APIs for both built-in loggers and third-party logger implementations.

C-Core and C++ have independent base classes of loggers and factories for users to implement.


rockspore:
@ctiller @markdroth PTAL. Is this what you had in your mind yesterday?

include/grpc/grpc_audit_logging.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+
+namespace grpc_core {
+namespace experimental {
+
+// The base struct for audit context.
+typedef struct CoreAuditContext {
```
ctiller, line 30: `class AuditContext {` -- no need for a typedef

include/grpcpp/security/audit_logging.h
```diff
@@ -0,0 +1,106 @@
...
+      : core_context_(core_context) {}
+
+  grpc::string_ref rpc_method() const;
+  grpc::string_ref principal() const;
+  grpc::string_ref policy_name() const;
+  grpc::string_ref matched_rule() const;
+  bool authorized() const;
+
+ private:
+  const grpc_core::experimental::CoreAuditContext* core_context_;
```
ctiller, line 45: I'd make this a reference

include/grpcpp/security/audit_logging.h
```diff
@@ -0,0 +1,106 @@
...
+};
+
+// The base class for audit logger implementations.
+// Users are expected to inherit this class and implement the Log() function.
+class AuditLogger : public grpc_core::experimental::CoreAuditLogger {
```
ctiller, line 50: I don't think we need inheritance here: this should be its own type.

include/grpc/grpc_audit_logging.h
```diff
@@ -0,0 +1,71 @@
...
+  absl::string_view principal() const;
+  absl::string_view policy_name() const;
+  absl::string_view matched_rule() const;
+  bool authorized() const;
+} AuditContext;
+
+// This base class for audit logger implementations.
+class CoreAuditLogger {
+ public:
+  virtual void CoreLog(const CoreAuditContext& audit_context) = 0;
```
ctiller, line 42: This should just be Log(...), and we'll have an instance that has a `unique_ptr<grpc::AuditLogger> log_` that we call `log_->Log(...)` on.

rockspore: So it means this class defined in C-Core can have a dependence on the C++ class, right? I initially was about to do this but wasn't sure if it's fine to have such a reversed dependency.

rockspore: Oh. is the following flow correct?

```C++
// C-Core
class AuditLogger {
 public:
  virtual void Log() = 0;
};
```

```
// C++
class AuditLogger {
 public:
  virtual void Log() = 0;
};

// non-public API
class CoreLogger : public grpc_core::AuditLogger {
 public:
  void Log() { logger_->Log(); }
 private:
  std::unique_ptr<AuditLogger> logger_;
}
```

BUILD
```diff
@@ -1210,15 +1210,15 @@ grpc_cc_library(
...
         "//src/core:lib/channel/channel_stack.cc",
         "//src/core:lib/channel/channel_stack_builder_impl.cc",
         "//src/core:lib/channel/channel_trace.cc",
         "//src/core:lib/channel/channelz.cc",
         "//src/core:lib/channel/channelz_registry.cc",
         "//src/core:lib/channel/connected_channel.cc",
-        "//src/core:lib/channel/server_call_tracer_filter.cc",
         "//src/core:lib/channel/promise_based_filter.cc",
-        "//src/core:lib/channel/call_tracer.cc",
+        "//src/core:lib/channel/server_call_tracer_filter.cc",
```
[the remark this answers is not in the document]
rockspore: `buildifier` automatically formatted this by sorting the deps. Nothing new.

src/core/lib/security/audit_logging/grpc_audit_logging.h
```diff
@@ -0,0 +1,95 @@
...
+  AuditContext(absl::string_view rpc_method, absl::string_view principal,
+               absl::string_view policy_name, absl::string_view matched_rule,
+               bool authorized)
+      : rpc_method_(rpc_method),
+        principal_(principal),
+        policy_name_(policy_name),
+        matched_rule_(matched_rule),
+        authorized_(authorized) {}
+
+  absl::string_view rpc_method() const { return rpc_method_; }
```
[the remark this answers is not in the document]
rockspore: We will guarantee it outlives the `Log()` invocation it is given to. Users will have to make copies if they need the info for longer (e.g. invoking some async processes inside `Log()`)

rockspore:
Since I put C++ APIs in `grpc/include/`, many checks are meant to fail for now due to the failure of `//test/core/surface:public_headers_must_be_c89`.

include/grpc/grpc_audit_logging.h
```diff
@@ -0,0 +1,95 @@
...
+
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+
+#include "src/core/lib/json/json.h"
+
+namespace grpc_core {
+namespace experimental {
+
+// The base struct for audit context.
```
gtcooke94, line 35: nit: struct -> class

src/core/lib/security/audit_logging/audit_logging.cc
```diff
@@ -0,0 +1,69 @@
...
+}
+
+absl::StatusOr<AuditLoggerFactory*> AuditLoggerRegistry::GetAuditLoggerFactory(
+    absl::string_view name) {
+  MutexLock lock(&mu_);
+  auto it = logger_factories_map_.find(std::string(name));
+  if (it != logger_factories_map_.end()) {
+    return it->second.get();
+  }
+  return absl::NotFoundError("factory does not exist");
```
gtcooke94, line 50: Consider adding more info to error, something like
`AuditLoggerFactory does not exist for name {name}`

src/core/lib/security/audit_logging/audit_logging.h
```diff
@@ -0,0 +1,61 @@
...
+
+class AuditLoggerRegistry {
+ public:
+  void RegisterAuditLoggerFactory(std::unique_ptr<AuditLoggerFactory>);
+  absl::StatusOr<AuditLoggerFactory*> GetAuditLoggerFactory(absl::string_view);
+
+  // Factories are registered during initialization. They should never be
+  // unregistered since they will be looked up at any time till the program
+  // exits. This function should only be used in tests.
+  void UnregisterAuditLoggerFactory(absl::string_view name);
```
gtcooke94, line 47: Move the `This function should only be used in tests` to the beginning of the comment.
Maybe add something to the actual name like `TestOnlyUnregisterAuditLoggerFactory` to be extremely clear, I think that is done sometimes in the repo.

src/cpp/server/audit_logging.h
```diff
@@ -0,0 +1,89 @@
...
+#include <grpc/grpc_audit_logging.h>
+#include <grpcpp/security/audit_logging.h>
+
+#include "src/core/lib/json/json.h"
+
+namespace grpc {
+namespace experimental {
+
+// Unregisters an audit logger factory. This should only be used in tests.
+void UnregisterAuditLoggerFactory(absl::string_view name);
```
gtcooke94, line 38: Same things about it being TestOnly, would move that to the beginning of the comment

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -0,0 +1,106 @@
...
+}  // namespace
+
+TEST(AuditLoggingTest, FactoryRegistrationAndLoggerCreation) {
+  RegisterAuditLoggerFactory(std::make_unique<TestAuditLoggerFactory>());
+  auto& registry = GetAuditLoggerRegistry();
+  auto result = registry.GetAuditLoggerFactory(kName);
+  ASSERT_TRUE(result.ok());
+  auto* factory = result.value();
+  EXPECT_EQ(factory->name(), kName);
+  auto result2 = factory->ParseAuditLoggerConfig(Json());
```
gtcooke94, line 83: nit: name `result2` something more descriptive `parse_result` maybe?

gtcooke94 (commented):
Few small things

test/core/security/grpc_audit_logging_test.cc
rockspore: Done.

src/cpp/server/audit_logging.h
rockspore: This function is gone now. But for the member function of the registry, I added `TestOnly` prefix.

include/grpc/grpc_audit_logging.h
rockspore: Done. Modified the description.

src/core/lib/security/audit_logging/audit_logging.cc
rockspore: Done.

include/grpc/grpc_audit_logging.h
```diff
@@ -0,0 +1,96 @@
...
+  virtual void Log(const AuditContext& audit_context) = 0;
+};
+
+// This is the base class for audit logger factory implementations.
+class AuditLoggerFactory {
+ public:
+  class Config {
+   public:
+    virtual ~Config() = default;
+    virtual const char* name() const = 0;
```
markdroth, line 75: This should probably use `absl::string_view` as well, now that we've decided we're okay with that.

include/grpc/grpc_audit_logging.h
```diff
@@ -0,0 +1,96 @@
...
+    virtual std::string ToString() = 0;
```
markdroth, line 76: This method should be `const`.

include/grpc/grpc_audit_logging.h
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  virtual ~AuditLoggerFactory() = default;
+  virtual const char* name() const = 0;
```
markdroth, line 80: This should return `absl::string_view`.

include/grpcpp/security/audit_logging.h
```diff
@@ -0,0 +1,43 @@
...
+#include <grpcpp/support/string_ref.h>
+
+namespace grpc {
+namespace experimental {
+
+using AuditContext = grpc_core::experimental::AuditContext;
+using AuditLogger = grpc_core::experimental::AuditLogger;
+using AuditLoggerFactory = grpc_core::experimental::AuditLoggerFactory;
+
+void RegisterAuditLoggerFactory(std::unique_ptr<AuditLoggerFactory> factory);
```
markdroth, line 38: I think this can be a `using` statement as well.

src/core/BUILD
```diff
@@ -2716,6 +2716,27 @@ grpc_cc_library(
     ],
 )
 
+grpc_cc_library(
+    name = "grpc_audit_logging",
+    srcs = [
+        "lib/security/audit_logging/audit_logging.cc",
```
markdroth, line 2722: I think the audit_logging directory should be under the authorization directory.

src/core/lib/security/audit_logging/audit_logging.h
```diff
@@ -0,0 +1,61 @@
...
+  void TestOnlyUnregisterAuditLoggerFactory(absl::string_view name);
+
+ private:
+  Mutex mu_;
+  std::map<std::string, std::unique_ptr<AuditLoggerFactory>>
+      logger_factories_map_ ABSL_GUARDED_BY(mu_);
+};
+
+// Get the static registry instance.
+AuditLoggerRegistry& GetAuditLoggerRegistry();
```
markdroth, line 56: Suggest making this a static `Get()` method inside of `AuditLoggerRegistry`.

src/core/lib/security/audit_logging/audit_logging.h
```diff
@@ -0,0 +1,61 @@
...
+ public:
+  void RegisterAuditLoggerFactory(std::unique_ptr<AuditLoggerFactory>);
+  absl::StatusOr<AuditLoggerFactory*> GetAuditLoggerFactory(absl::string_view);
+
+  // Factories are registered during initialization. They should never be
+  // unregistered since they will be looked up at any time till the program
+  // exits. This function should only be used in tests.
+  void TestOnlyUnregisterAuditLoggerFactory(absl::string_view name);
+
+ private:
```
markdroth, line 49: Suggest making the ctor private.

src/core/lib/security/audit_logging/audit_logging.h
```diff
@@ -0,0 +1,61 @@
...
+  Mutex mu_;
+  std::map<std::string, std::unique_ptr<AuditLoggerFactory>>
```
markdroth, line 51: This can be keyed by `absl::string_view` instead of `std::string`.  It can point into the string returned by the factory's `name()` method.

src/core/lib/security/audit_logging/audit_logging.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include "src/core/lib/gprpp/sync.h"
+
+namespace grpc_core {
+namespace experimental {
+
+class AuditLoggerRegistry {
+ public:
+  void RegisterAuditLoggerFactory(std::unique_ptr<AuditLoggerFactory>);
+  absl::StatusOr<AuditLoggerFactory*> GetAuditLoggerFactory(absl::string_view);
```
markdroth, line 42: Instead of just returning the factory and having the caller interact with the factory, it will probably wind up being more convenient to have `ParseAuditLoggerConfig()` and `CreateAuditLogger()` methods that look up the right factory and then invoke the corresponding method on the factory.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -0,0 +1,106 @@
...
+
+namespace grpc_core {
+namespace testing {
+
+const char kName[] = "test_logger";
+
+using experimental::AuditContext;
+using experimental::AuditLogger;
+using experimental::AuditLoggerFactory;
+using Config = experimental::AuditLoggerFactory::Config;
```
markdroth, line 44: I don't think the `Config =` part is needed.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -0,0 +1,106 @@
...
+
+#include "src/core/lib/json/json.h"
+#include "src/core/lib/security/audit_logging/audit_logging.h"
+#include "test/core/util/test_config.h"
+#include "test/core/util/tls_utils.h"
+
+namespace grpc_core {
+namespace testing {
+
+const char kName[] = "test_logger";
```
markdroth, line 39: This can be `absl::string_view`.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -0,0 +1,106 @@
...
+  auto& registry = GetAuditLoggerRegistry();
+  auto result = registry.GetAuditLoggerFactory(kName);
+  ASSERT_TRUE(result.ok());
+  auto* factory = result.value();
+  EXPECT_EQ(factory->name(), kName);
+  auto parse_result = factory->ParseAuditLoggerConfig(Json());
+  ASSERT_TRUE(parse_result.ok());
+  std::unique_ptr<Config> config = std::move(parse_result.value());
+  ASSERT_NE(factory->CreateAuditLogger(std::move(config)), nullptr);
+  registry.TestOnlyUnregisterAuditLoggerFactory(kName);
```
markdroth, line 87: The intent of this seems to be to undo the changes that the test made to the registry.  However, the `ASSERT_*` checks above will immediately return from this function if the check fails, in which case this will never be run.

I suggest replacing the `TestOnlyUnregisterAuditLoggerFactory()` method with a `TestOnlyResetRegistry()` method that simple resets the global registry to a default instance.  Then you can create a test suite class that calls that method in `TearDown()`, so that the registry will always be completely reset after each test.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -0,0 +1,106 @@
...
+}
+
+TEST(AuditLoggingTest, FactoryNotFound) {
+  auto& registry = GetAuditLoggerRegistry();
+  auto result = registry.GetAuditLoggerFactory(kName);
+  ASSERT_EQ(result.ok(), false);
```
markdroth, line 93: Please check the error message in this case.  We need to make sure we're providing useful error messages to users.

test/cpp/server/audit_logging_test.cc
markdroth, line 1: Now that we're no longer wrapping the C-core API, I don't think we need this duplicate set of tests.

markdroth (commented):
This looks good overall!  Comments are mostly minor.

include/grpc/grpc_audit_logging.h
rockspore: Changed to `absl::string_view`. Same applies to the other places.

include/grpcpp/security/audit_logging.h
rockspore: Fixed.

src/core/BUILD
rockspore: Moved them over.

test/core/security/grpc_audit_logging_test.cc
rockspore: IDE warns me that using statements cannot refer to class members. But I removed this and just use `AuditLoggerFactory::Config` subsequently now.

test/core/security/grpc_audit_logging_test.cc
rockspore: Done. Thanks for the suggestion!

test/cpp/server/audit_logging_test.cc
rockspore: Removed this test.

src/core/lib/security/audit_logging/audit_logging.h
rockspore: Fixed.

src/core/lib/security/audit_logging/audit_logging.h
rockspore: Done. I made all the public methods static now so that there is no need to grab a reference of the registry outside the class ever. And this one is now private.

src/core/lib/security/audit_logging/audit_logging.h
rockspore: Done. And I made them static.

src/core/lib/security/audit_logging/audit_logging.h
rockspore: Fixed.

test/core/security/grpc_audit_logging_test.cc
rockspore: Done.

(Thinking out loud) Also by adding `ParseAuditLoggerConfig`, it made me realize that more tests need to be added here since now we will directly use this API to parse a config like:
```
{"stdout_logger": { actual config }}
```

whereas I previously planned to find the factory by the key above and pass that actual config into the factory's `ParseAuditLoggerConfig` config.

rockspore: Actually, on second thought, I changed my mind and let this function take the name and the "actual config" as two parameters. Because neither in SDK or xDS will we receive such a json object directly.

I also removed those added tests.

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -0,0 +1,101 @@
...
+    return it->second.get();
+  }
+  return absl::NotFoundError(
+      absl::StrFormat("audit logger factory %s does not exist", name));
+}
+
+void AuditLoggerRegistry::TestOnlyResetRegistry() {
+  auto& registry = GetAuditLoggerRegistry();
+  MutexLock lock(&registry.mu_);
+  registry.logger_factories_map_.clear();
```
markdroth, line 93: Note that we are going to want to automatically register all of the built-in logger factories (i.e., the stdout logger) in the `AuditLoggerRegistry` ctor, since we won't want applications to need to explicitly register built-in loggers.

I think that means that this needs to be structured a little differently, because we'll want this to reset the registry to its default state, which will include the built-in loggers.  So I think what we want here is something like this:

```
namespace {
Mutex* g_mu = new Mutex();  // Replaces AuditLoggerRegistry::mu_
AuditLoggerRegistry* g_registry ABSL_GUARDED_BY(g_mu) =
    new AuditLoggerRegistry();
}  // namespace

void AuditLoggerRegistry::TestOnlyResetRegistry() {
  MutexLock lock(g_mu);
  delete g_registry;
  g_registry = new AuditLoggerRegistry();
}
```

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -0,0 +1,101 @@
...
+  return registry.logger_factories_map_.find(name) !=
+         registry.logger_factories_map_.end();
+}
+
+absl::StatusOr<std::unique_ptr<AuditLoggerFactory::Config>>
+AuditLoggerRegistry::ParseAuditLoggerConfig(absl::string_view name,
+                                            const Json& json) {
+  auto factory_or = AuditLoggerRegistry::GetAuditLoggerFactory(name);
+  if (!factory_or.ok()) {
+    return absl::InvalidArgumentError("unsupported audit logger type");
```
markdroth, line 61: This should just return `factory_or.status()` instead of creating a new status.

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -0,0 +1,101 @@
...
+  static AuditLoggerRegistry& registry = *new AuditLoggerRegistry();
+  return registry;
+}
+
+absl::StatusOr<AuditLoggerFactory*> AuditLoggerRegistry::GetAuditLoggerFactory(
+    absl::string_view name) {
+  auto& registry = GetAuditLoggerRegistry();
+  MutexLock lock(&registry.mu_);
+  auto it = registry.logger_factories_map_.find(name);
+  if (it != registry.logger_factories_map_.end()) {
```
markdroth, line 83: Suggest handling the error case first, since that's a bit more readable:

```
if (it == registry.logger_factories_map_.end()) {
  return absl::NotFoundError(
      absl::StrFormat("audit logger factory %s does not exist", name));
}
return it->second.get();
```

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -0,0 +1,101 @@
...
+    return absl::InvalidArgumentError("unsupported audit logger type");
+  }
+  return factory_or.value()->ParseAuditLoggerConfig(json);
+}
+
+std::unique_ptr<AuditLogger> AuditLoggerRegistry::CreateAuditLogger(
+    std::unique_ptr<AuditLoggerFactory::Config> config) {
+  auto factory_or = AuditLoggerRegistry::GetAuditLoggerFactory(config->name());
+  GPR_ASSERT(factory_or.ok());
+  return factory_or.value()->CreateAuditLogger(std::move(config));
```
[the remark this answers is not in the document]
rockspore: Thanks for spotting the case! I followed your suggested pattern with a global `g_mu` and `g_registry`. Since the registry read is not guarded by the mutex, I think this scenario is addressed as well.

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -0,0 +1,101 @@
...
+#include "src/core/lib/json/json.h"
+
+namespace grpc_core {
+namespace experimental {
+
+void AuditLoggerRegistry::RegisterAuditLoggerFactory(
+    std::unique_ptr<AuditLoggerFactory> factory) {
+  auto& registry = GetAuditLoggerRegistry();
+  MutexLock lock(&registry.mu_);
+  registry.logger_factories_map_[factory->name()] = std::move(factory);
```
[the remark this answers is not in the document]
rockspore: Added the assertion to require the factory name to not be registered before.

src/core/lib/security/authorization/audit_logging.cc
rockspore: Fixed.

src/core/lib/security/authorization/audit_logging.cc
rockspore: Done.

src/core/lib/security/authorization/audit_logging.cc
rockspore: I followed this pattern and restructured all the relevant methods. I also made the ctor public again since this namespace variable needs to call it. Is this pattern using global variables fine? The only alternative I can think of still making it local to the function-static would be to have the `GetAuditLoggerRegistry` function conditionally reset the instance.

src/core/lib/security/authorization/audit_logging.h
```diff
@@ -32,30 +31,46 @@
...
+#include "src/core/lib/json/json.h"
 
 namespace grpc_core {
 namespace experimental {
 
 class AuditLoggerRegistry {
  public:
-  void RegisterAuditLoggerFactory(std::unique_ptr<AuditLoggerFactory>);
-  absl::StatusOr<AuditLoggerFactory*> GetAuditLoggerFactory(absl::string_view);
+  static void RegisterAuditLoggerFactory(std::unique_ptr<AuditLoggerFactory>);
```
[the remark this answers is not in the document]
rockspore: Done.

src/core/lib/security/authorization/audit_logging.cc
rockspore: (Had the response typed for a while but forgot to hit Comment).

I followed your suggested pattern and restructured all the methods.

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -39,62 +40,65 @@
...
   return factory_or.value()->CreateAuditLogger(std::move(config));
 }
 
-AuditLoggerRegistry& AuditLoggerRegistry::GetAuditLoggerRegistry() {
-  static AuditLoggerRegistry& registry = *new AuditLoggerRegistry();
-  return registry;
-}
-
 absl::StatusOr<AuditLoggerFactory*> AuditLoggerRegistry::GetAuditLoggerFactory(
     absl::string_view name) {
```
markdroth, line 85: Suggest adding an `ABSL_EXCLUSIVE_LOCKS_REQUIRED(g_mu)` annotation here.

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -39,62 +40,65 @@
...
+}  // namespace
+
+void AuditLoggerRegistry::RegisterFactory(
     std::unique_ptr<AuditLoggerFactory> factory) {
-  auto& registry = GetAuditLoggerRegistry();
-  MutexLock lock(&registry.mu_);
-  registry.logger_factories_map_[factory->name()] = std::move(factory);
+  if (factory == nullptr) return;
+  MutexLock lock(g_mu);
+  GPR_ASSERT(g_registry->logger_factories_map_.find(factory->name()) ==
```
markdroth, line 55: Suggest writing this as follows, to avoid doing the lookup twice:

```
absl::string_view name = factory->name();
GPR_ASSERT(
    g_registry->logger_factories_map_.emplace(name, std::move(factory))
        .second);
```

src/core/lib/security/authorization/audit_logging.cc
markdroth: Instead of making the ctor public, how about something like this:

```
namespace {
Mutex* g_mu = new Mutex();
AuditLoggerRegistry* g_registry ABSL_GUARDED_BY(g_mu) = nullptr;
}  // namespace

// Private static method in AuditLoggerRegistry.
AuditLoggerRegistry& AuditLoggerRegistry::Get() {
  MutexLock lock(g_mu);
  if (g_registry == nullptr) g_registry = new AuditLoggerRegistry();
  return g_registry;
}
```

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -39,62 +40,65 @@
...
+    new AuditLoggerRegistry();
+
+}  // namespace
+
+void AuditLoggerRegistry::RegisterFactory(
     std::unique_ptr<AuditLoggerFactory> factory) {
-  auto& registry = GetAuditLoggerRegistry();
-  MutexLock lock(&registry.mu_);
-  registry.logger_factories_map_[factory->name()] = std::move(factory);
+  if (factory == nullptr) return;
```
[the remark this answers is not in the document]
rockspore: Fixed.

src/core/lib/security/authorization/audit_logging.cc
rockspore: Done.

src/core/lib/security/authorization/audit_logging.cc
rockspore: I removed this function. There are only two places that use it and given the if condition there, putting this logic inline doesn't lead to more LOC. It also eliminates the problem of locking both registry reads and map I/O.

src/core/lib/security/authorization/audit_logging.cc
rockspore: With both the mutex and registry as static class members, the ctor can remain private now.

src/core/lib/security/authorization/audit_logging.h
```diff
@@ -56,12 +56,16 @@ class AuditLoggerRegistry {
   static void TestOnlyResetRegistry();
 
  private:
-  absl::StatusOr<AuditLoggerFactory*> GetAuditLoggerFactory(
-      absl::string_view name);
+  // TODO(lwge): Add built-in logger registrations once avaialble.
+  AuditLoggerRegistry() = default;
+
+  static Mutex mu;
```
markdroth, line 62: Note that the style guide dictates that static and global variables must be trivially destructible:

https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

I think this means that we'll need both of these static variables to be pointers, and we can dynamically allocate the value at initialization time.

src/core/lib/security/authorization/audit_logging.cc
```diff
@@ -40,61 +39,47 @@
...
     std::unique_ptr<AuditLoggerFactory> factory) {
-  if (factory == nullptr) return;
-  MutexLock lock(g_mu);
-  GPR_ASSERT(g_registry->logger_factories_map_.find(factory->name()) ==
-             g_registry->logger_factories_map_.end());
-  g_registry->logger_factories_map_[factory->name()] = std::move(factory);
+  GPR_ASSERT(factory != nullptr);
+  MutexLock lock(&mu);
+  GPR_ASSERT(registry.logger_factories_map_
+                 .emplace(factory->name(), std::move(factory))
```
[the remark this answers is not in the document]
rockspore: I see. Fixed.

src/core/lib/security/authorization/audit_logging.h
markdroth: > should we only construct the registry when it's being used?

I'd be fine with that change if you want to make it, but I don't think it's strictly necessary.  We're not allocating a huge amount of memory here.

markdroth (approved):
This looks great!

src/core/lib/security/authorization/audit_logging.h
rockspore: Sounds good. I'll leave it as is then.

PR #32757, opened 2023-03-30
[PSM Interop] Add custom_lb test client

eugeneo:


test/cpp/interop/rpc_behavior_lb_policy.cc
```diff
@@ -0,0 +1,196 @@
...
+  void ExitIdleLocked() override { delegate_->ExitIdleLocked(); }
+
+  void ResetBackoffLocked() override { delegate_->ResetBackoffLocked(); }
+
+ private:
+  class Picker : public SubchannelPicker {
+   public:
+    Picker(RefCountedPtr<SubchannelPicker> delegate_picker,
+           RefCountedPtr<RpcBehaviorLbPolicy> parent)
+        : delegate_picker_(std::move(delegate_picker)), parent_(parent) {}
```
markdroth, line 100: `std::move(parent)`

test/cpp/interop/rpc_behavior_lb_policy.cc
```diff
@@ -0,0 +1,196 @@
...
+  absl::string_view name() const override { return kRpcBehaviorLbPolicyName; }
+
+  grpc_core::OrphanablePtr<LoadBalancingPolicy> CreateLoadBalancingPolicy(
+      LoadBalancingPolicy::Args args) const override {
+    return grpc_core::MakeOrphanable<RpcBehaviorLbPolicy>(std::move(args));
+  }
+
+  absl::StatusOr<RefCountedPtr<LoadBalancingPolicy::Config>>
+  ParseLoadBalancingConfig(const Json& json) const override {
+    if (json.type() == Json::Type::kNull) {
```
markdroth, line 177: You can remove this block.  This case can't happen anymore, and I've been meaning to remove this check from all of our existing policies.

tools/run_tests/xds_k8s_test_driver/tests/custom_lb_test.py
```diff
@@ -40,6 +40,8 @@ class CustomLbTest(xds_k8s_testcase.RegularXdsKubernetesTestCase):
     def is_supported(config: skips.TestConfig) -> bool:
         if config.client_lang == _Lang.JAVA:
             return config.version_gte('v1.47.x')
+        if config.client_lang == _Lang.CPP:
+            return config.version_gte('v1.54.x')
```
markdroth, line 44: The 1.54.x branch has already been cut, so this probably won't actually be supported without env var protection until 1.55.

For now, if you want to test this *with* the env var protection, you can do that with any version since 1.51, but you'll need a way to set the env var when you run the client.

markdroth (commented):
This looks good!  Only one substantive comment, which is the one about accessing `rpc_behavior_` from the picker.

test/cpp/interop/rpc_behavior_lb_policy.cc
eugeneo: Done

test/cpp/interop/rpc_behavior_lb_policy.cc
```diff
@@ -0,0 +1,196 @@
...
+ private:
+  class Picker : public SubchannelPicker {
+   public:
+    Picker(RefCountedPtr<SubchannelPicker> delegate_picker,
+           RefCountedPtr<RpcBehaviorLbPolicy> parent)
+        : delegate_picker_(std::move(delegate_picker)), parent_(parent) {}
+
+    PickResult Pick(PickArgs args) override {
+      args.initial_metadata->Add(kRpcBehaviorMetadataKey,
+                                 parent_->rpc_behavior_);
```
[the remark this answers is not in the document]
eugeneo: Done

test/cpp/interop/rpc_behavior_lb_policy.cc
```diff
@@ -0,0 +1,196 @@
...
+
+  class Helper : public ChannelControlHelper {
+   public:
+    explicit Helper(RefCountedPtr<RpcBehaviorLbPolicy> parent)
+        : parent_(std::move(parent)) {}
+
+    RefCountedPtr<grpc_core::SubchannelInterface> CreateSubchannel(
+        grpc_core::ServerAddress address,
+        const grpc_core::ChannelArgs& args) override {
+      auto subchannel = parent_->channel_control_helper()->CreateSubchannel(
```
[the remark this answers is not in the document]
eugeneo: Done

test/cpp/interop/rpc_behavior_lb_policy.cc
eugeneo: Done

tools/run_tests/xds_k8s_test_driver/tests/custom_lb_test.py
eugeneo: This is version of Docker images with XDS interop test client. I updated to 1.55 as I don't think we update older images with the new XDS interop test client.

Environment variables are set here - [Kubernates config](https://github.com/grpc/grpc/blob/master/tools/run_tests/xds_k8s_test_driver/kubernetes-manifests/client.deployment.yaml#L51)

eugeneo (commented):

test/cpp/interop/rpc_behavior_lb_policy.cc
```diff
@@ -96,19 +96,22 @@ class RpcBehaviorLbPolicy : public LoadBalancingPolicy {
...
+           absl::string_view rpc_behavior)
+        : delegate_picker_(std::move(delegate_picker)),
+          rpc_behavior_(rpc_behavior) {}
 
     PickResult Pick(PickArgs args) override {
-      args.initial_metadata->Add(kRpcBehaviorMetadataKey,
-                                 parent_->rpc_behavior_);
+      char* rpc_behavior_copy = static_cast<char*>(
+          args.call_state->Alloc(rpc_behavior_.length() + 1));
+      strncpy(rpc_behavior_copy, rpc_behavior_.c_str(), rpc_behavior_.length());
```
markdroth, line 106: I think this should use `strcpy()` here, not `strncpy()`.  The latter has some edge cases that don't properly add the trailing nul character, and in this case, the bounds-checking isn't necessary anyway, because we're allocating the right amount of space before we do the copy.

tools/run_tests/xds_k8s_test_driver/tests/custom_lb_test.py
markdroth: If this requires that the env var protection be removed, then maybe hold off on making this part of the change until we actually remove the env var protection?  If we wind up hitting a delay and don't remove env var protection before 1.55, this could break.

markdroth (approved):

test/cpp/interop/rpc_behavior_lb_policy.cc
eugeneo: Done

tools/run_tests/xds_k8s_test_driver/tests/custom_lb_test.py
eugeneo: This code does not require removing the env var protection as env var is set in Kubernetes deployment config.

eugeneo (commented):
Thank you for the review. Please note that this test does not require us removing the env var protection.

tools/run_tests/xds_k8s_test_driver/tests/custom_lb_test.py
markdroth: If we're always setting the env var, then this test should work with version 1.51 or newer.

eugeneo: I will talk with Terry, but my understanding is this would require a backport to those branches.

markdroth: (This discussion doesn't need to block this PR from being merged at this point; whatever we set here, we can change later if this isn't exactly right.)

Does this control whether the test runs automatically for each branch?  If so, 1.55 is fine; we don't actually need it running on earlier branches, since it's still env var protected on the earlier branches.  But in that case, I still don't understand why we want to enable this on 1.55 in this PR, since we can't actually guarantee yet that the env var protection will go away by the time we cut the 1.55 branch (although we certainly intend that to happen).

Is there a way to enable this test on master without enabling it on any release branches?

eugeneo: This is a version of the image and not the branch. I.e. I can run this script from 1.55 branch but have it pull 1.50 image for test client. That test client will not have the code from this PR so this check currently only allows master and future branches (such as 1.55)

PR #32761, opened 2023-03-30
[gprpp] Add ValidationErrors::message retrieval

drfloob:
For cases where an absl::Status is unnecessary.

### DIFF
--- a/src/core/lib/gprpp/validation_errors.cc
+++ b/src/core/lib/gprpp/validation_errors.cc
@@ -23,7 +23,6 @@
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_join.h"
 #include "absl/strings/strip.h"
-#include "validation_errors.h"
 
 namespace grpc_core {
 
--- a/src/core/lib/gprpp/validation_errors.cc
+++ b/src/core/lib/gprpp/validation_errors.cc
@@ -43,13 +43,14 @@ bool ValidationErrors::FieldHasErrors() const {
   return field_errors_.find(absl::StrJoin(fields_, "")) != field_errors_.end();
 }
 
-absl::Status ValidationErrors::status(absl::string_view prefix,
-                                      absl::StatusCode code) const {
+absl::Status ValidationErrors::status(absl::StatusCode code,
+                                      absl::string_view prefix) const {
   if (field_errors_.empty()) return absl::OkStatus();
   return absl::Status(code, message(prefix));
 }
 
 std::string ValidationErrors::message(absl::string_view prefix) const {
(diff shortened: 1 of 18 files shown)


src/core/lib/gprpp/validation_errors.cc
```diff
@@ -42,8 +43,13 @@ bool ValidationErrors::FieldHasErrors() const {
   return field_errors_.find(absl::StrJoin(fields_, "")) != field_errors_.end();
 }
 
-absl::Status ValidationErrors::status(absl::string_view prefix) const {
+absl::Status ValidationErrors::status(absl::string_view prefix,
```
ctiller, line 46: Can we add `absl::StatusCode` as the first argument? (to mirror the `absl::Status` constructor)

src/core/lib/gprpp/validation_errors.cc
```diff
@@ -23,6 +23,7 @@
 #include "absl/strings/str_cat.h"
 #include "absl/strings/str_join.h"
 #include "absl/strings/strip.h"
+#include "validation_errors.h"
```
markdroth, line 26: This looks wrong.  It's a relative path, and it's a dup of the include on line 17 above.

src/core/lib/gprpp/validation_errors.cc
markdroth: I agree, it would probably make sense to make this a mandatory parameter and put it first.  It shouldn't be hard to update the callers.

src/core/lib/gprpp/validation_errors.cc
drfloob: Indeed. I haven't found a way to yet to teach my IDE to stop doing that, it sneaks in. Fixed.

src/core/lib/gprpp/validation_errors.cc
drfloob: Good idea. Done.

src/core/lib/gprpp/validation_errors.h
```diff
@@ -100,7 +100,12 @@ class ValidationErrors {
   bool FieldHasErrors() const GPR_ATTRIBUTE_NOINLINE;
 
   // Returns the resulting status of parsing.
-  absl::Status status(absl::string_view prefix) const;
+  absl::Status status(
+      absl::string_view prefix,
+      absl::StatusCode code = absl::StatusCode::kInvalidArgument) const;
+
+  // Returns the resulting error message
```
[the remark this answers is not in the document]
drfloob: Done

markdroth (approved):
Looks great!

PR #32662, opened 2023-03-20
Switch all VS2017 builds to VS2019

jtattermusch:
(part of removing support for VS2017)

Also see https://github.com/grpc/grpc/pull/32649
Also see https://github.com/grpc/grpc/pull/32615

The switch to grpc-win2019 windows workers has already happened: (cl/517400022).

Once this PR lands, I'll backport to 1.53.x branch as well (since that release removes the VS2017 support).

PR #32764, opened 2023-03-31
[cmake] Workaround missing stdalign.h in windows portability test

jtattermusch:
Tentative fix for b/275694647. Also see https://github.com/grpc/grpc/pull/32662

Adhoc windows/grpc_portability run: http://sponge/bb75e14c-fc77-45bb-89bd-4c4685746490

Once merged, I'll backport to 1.54.x and 1.53.x (these have switched to VS2019)

### DIFF
--- a/tools/run_tests/helper_scripts/build_cxx.bat
+++ b/tools/run_tests/helper_scripts/build_cxx.bat
@@ -29,6 +29,20 @@ If "%GRPC_BUILD_ACTIVATE_VS_TOOLS%" == "2019" (
   echo on
 )
 
+If "%GRPC_CMAKE_GENERATOR%" == "Visual Studio 16 2019" (
+  @rem Force using a new-enough Windows 10 SDK that supports C++11's stdalign.h
+  @rem For some reason, cmake together with visual studio generator by default
+  @rem pick an old version of Win SDK even when a newer version is installed on the system.
+  @rem When using Ninja generator, this problem doesn't happen.
+  @rem The argument corresponds to Microsoft.VisualStudio.Component.Windows10SDK.20348
+  @rem See b/275694647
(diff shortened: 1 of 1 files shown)


jtattermusch:
The adhoc windows/grpc_portability run: http://sponge/bb75e14c-fc77-45bb-89bd-4c4685746490 is green so this is ready for review&merge.

jtattermusch:
FYI @gnossen

davidben:
Oh, that's fun. :-/ Wonder how MSVC or CMake picks the default.

davidben:
> Oh, that's fun. :-/ Wonder how MSVC or CMake picks the default.

I think I found it!
https://github.com/Kitware/CMake/blob/f52a2fd26d13fa33d34dd1320f82d3f8aa49f3c3/Source/cmGlobalVisualStudio14Generator.cxx#L301

Though it's odd. The very last bit grabs the latest version, so it seems like we're not hitting that case. But otherwise it seems to just look for a match to `SystemVersion`, which I presume is `CMAKE_SYSTEM_VERSION`. I don't suppose it's possible your has set `CMAKE_SYSTEM_VERSION` elsewhere??

davidben:
Ugh. Looks like CMake defaults to the Windows SDK version that matches the host Windows version, rather than the latest one:

https://github.com/Kitware/CMake/blob/f52a2fd26d13fa33d34dd1320f82d3f8aa49f3c3/Source/cmGlobalGenerator.cxx#L624
https://github.com/Kitware/CMake/blob/f52a2fd26d13fa33d34dd1320f82d3f8aa49f3c3/Modules/CMakeDetermineSystem.cmake#L176

See also this discussion, and this comment in particular:
https://gitlab.kitware.com/cmake/cmake/-/issues/16202#note_140259

So I guess that means `-DCMAKE_SYSTEM_VERSION=10.0` may be cleaner way to force CMake to pick the latest one. But this whole situation is quite a mess. :-)

PR #32765, opened 2023-03-31
[xds interop tests] implement error-code- RPC behavior

eugeneo:
Implements handling for the `error-code-xx` `rcp-behavior` header. Compare to similar implementation in [java-server](https://github.com/grpc/grpc-java/blob/42b4c61d5e869b1534808204bc71df84720f5087/interop-testing/src/main/java/io/grpc/testing/integration/XdsTestServer.java#L379)

PR #32769, opened 2023-04-01
Revert "[xds interop tests] implement error-code- RPC behavior"

ctiller:
Reverts grpc/grpc#32765


sergiitk:
This was restored in https://github.com/grpc/grpc/pull/32810.

PR #32787, opened 2023-04-03
[metadata] Fix fuzzer bug with metadata arg.

ananda1066:
Bug: b/276525236.


ctiller:
please add the fuzzer example to the corpus here (it's the equivalent of adding a presubmit unit test, and that seems good)

ananda1066:
Good idea, done.

ctiller (approved):

PR #32800, opened 2023-04-05
O11y init

XuanWang-Amos:

### DIFF
--- a/src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
+++ b/src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
@@ -59,7 +59,6 @@ class PySpan:
 
 
 def observability_init() -> None:
-  gcpObservabilityInit() # remove print buffer
   _start_exporting_thread()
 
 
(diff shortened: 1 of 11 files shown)


src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
```diff
@@ -0,0 +1,199 @@
...
+import sys
+import os
+import logging
+from threading import Thread
+from typing import List, Tuple, Mapping
+
+from grpc_observability import open_census
+from grpc_observability import measures
+
+cdef const char* CLIENT_CALL_TRACER = "client_call_tracer"
```
gnossen, line 26: This name should probably be more specific in case there end up being other implementations. Something like `gcp_opencensus_client_call_tracer`. Same for the server call tracer factory string.

gnossen (commented):
Overall, the structure looks good, but I think we need to firm up the dependencies between the `grpc_observability` module and the `grpc` module.

PR #32806, opened 2023-04-05
[JSON] remove mutable accessor methods.

markdroth:


ctiller (approved):
I think this is fine for now... Like I mentioned earlier probably worth adding a `JsonBuilder` kind of type at some point, I think it'd clean up a lot of our deserialization code - but not here & now.

PR #32808, opened 2023-04-05
[JSON] use absl::variant for JSON values

markdroth:
This changes the `JSON` class to use `absl::variant<>` internally for storing the type and value.

The only API-visible change here is that now the accessor methods will crash if you use them on the wrong type, courtesy of `absl::get<>`.  This does not appear to break any tests, so it appears that we have already cleaned up all of the cases we had in the past where we might have done something like that.


src/core/lib/json/json.h
```diff
@@ -86,140 +86,107 @@ class Json {
...
   Json& operator=(bool b) {
-    type_ = b ? Type::kTrue : Type::kFalse;
+    value_ = b;
     return *this;
   }
 
   // Construct from any numeric type.
   template <typename NumericType>
   // NOLINTNEXTLINE(google-explicit-constructor)
-  Json(NumericType number)
```
ctiller, line 108: For a follow-up PR (I didn't catch this earlier, sorry!)

This function provides implicit conversion for anything that `std::to_string` supports - which may change in the future and bite us... would suggest making an explicit set of types for a public API.

Worse though:
```
std::to_string relies on the current locale for formatting purposes, and therefore concurrent calls to std::to_string from multiple threads may result in partial serialization of calls.
```
(source: https://en.cppreference.com/w/cpp/string/basic_string/to_string)

I'm not terribly fussed about the serialization (scary, but not terrible) - the locale dependency is worrisome though - I don't want to be dealing with EU specific bugs in gRPC ever.

ctiller (approved):

src/core/lib/json/json.h
markdroth: Good catch!  I will fix this in a subsequent PR.

PR #32799, opened 2023-04-04
[PSM Interop] Enable custom_lb_test for CXX

eugeneo:

PR #32811, opened 2023-04-05
[PSM Interop] Disable custom_lb test for cpp

sergiitk:
The code needed for this test was reverted in https://github.com/grpc/grpc/pull/32769.  
Let's uncomment this code in https://github.com/grpc/grpc/pull/32799, once the custom_lb support is brought back.


sergiitk:
The code was restored in https://github.com/grpc/grpc/pull/32810. No need to disable the test now.

PR #32815, opened 2023-04-05
Revert "XDS: enable XDS federation by default (#32711)"

apolcyn:
Backports https://github.com/grpc/grpc/pull/32814 to 1.54.x

cc @stanley-cheung


markdroth:
Please wait for #32814 to be merged and then back-port that merge commit instead of just copying the diff, so that up-merges from this branch are easier later.

PR #32825, opened 2023-04-06
[Transport] Update Chttp2 ContextList to include relative offset of traced RPCs within outgoing buffer

Vignesh2208:
The PR also creates a separate BUILD target for:
- chttp2 context list
- iomgr buffer_list
- iomgr internal errqueue

This would allow the context list to be included as standalone dependencies for EventEngine implementations.

### DIFF
--- a/src/core/ext/transport/chttp2/transport/context_list.h
+++ b/src/core/ext/transport/chttp2/transport/context_list.h
@@ -23,7 +23,6 @@
 
 #include <stddef.h>
 
-#include "src/core/ext/transport/chttp2/transport/frame.h"
 #include "src/core/lib/iomgr/buffer_list.h"
 #include "src/core/lib/iomgr/error.h"
 
@@ -32,18 +31,46 @@ namespace grpc_core {
 class ContextList {
  public:
   // Creates a new element with \a context as the value and appends it to the
-  // list.
-  static void Append(ContextList** head, grpc_chttp2_stream* s);
+  // list. Each RPC/stream is associated with a unique \a context. This method
+  // is invoked when a chunk of data stored in an outgoing buffer is going to be
+  // sent over the wire. A data chunk being written over the wire is multiplexed
+  // with bytes from multiple RPCs. If one such RPC is traced, we store the
+  // following information about the traced RPC:
+  // - byte_offset_in_stream: Number of bytes belonging to that traced RPC which
+  // have been sent so far from the start of the RPC stream.
+  // - relative_start_pos_in_chunk: Starting offset of the traced RPC within
+  // the current chunk that is being sent.
+  // - num_traced_bytes_in_chunk: Number of bytes belonging to the traced RPC
+  // within the current chunk.
+  static void Append(ContextList** tail, void* context,
+                     size_t byte_offset_in_stream,
+                     int64_t relative_start_pos_in_chunk,
+                     int64_t num_traced_bytes_in_chunk);
 
-  // Executes a function \a fn with each context in the list and \a ts. It also
-  // frees up the entire list after this operation. It is intended as a callback
-  // and hence does not take a ref on \a error
+  // Executes the function set using grpc_http2_set_write_timestamps_callback
+  // method with each context in the list and \a ts. It also frees up the entire
+  // list after this operation. It is intended as a callback and hence does not
+  // take a ref on \a error. The fn receives individual contexts in the same
+  // order in which they were Appended.
   static void Execute(void* arg, Timestamps* ts, grpc_error_handle error);
+
+  // Executes the passed function \a cb with each context in the list. The
+  // arguments provided to cb include the trace_context_, byte_offset_,
+  // traced_bytes_relative_start_pos_ and num_traced_bytes_ for each context in
+  // the context list. It also frees up the entire list after this operation.
+  // The cb receives individual contexts in the same order in which they were
+  // Appended.
+  static void ForEachExecuteCallback(
+      ContextList* tail,
+      std::function<void(void*, size_t, int64_t, int64_t)> cb);
 
  private:
   void* trace_context_ = nullptr;
-  ContextList* next_ = nullptr;
-  size_t byte_offset_ = 0;
+  ContextList* next_ = nullptr;  // For iterating from head to tail.
+  ContextList* head_ = nullptr;  // Point to the first element in the list.
+  int64_t relative_start_pos_in_chunk_ = 0;
+  int64_t num_traced_bytes_in_chunk_ = 0;
+  size_t byte_offset_in_stream_ = 0;
 };
 
 void grpc_http2_set_write_timestamps_callback(
--- a/BUILD
+++ b/BUILD
@@ -1223,7 +1223,6 @@ grpc_cc_library(
         "//src/core:lib/compression/compression.cc",
         "//src/core:lib/compression/compression_internal.cc",
         "//src/core:lib/compression/message_compress.cc",
-        "//src/core:lib/iomgr/buffer_list.cc",
         "//src/core:lib/iomgr/call_combiner.cc",
         "//src/core:lib/iomgr/cfstream_handle.cc",
         "//src/core:lib/iomgr/dualstack_socket_posix.cc",
@@ -1242,7 +1241,6 @@ grpc_cc_library(
         "//src/core:lib/iomgr/gethostname_fallback.cc",
         "//src/core:lib/iomgr/gethostname_host_name_max.cc",
         "//src/core:lib/iomgr/gethostname_sysconf.cc",
-        "//src/core:lib/iomgr/internal_errqueue.cc",
         "//src/core:lib/iomgr/iocp_windows.cc",
         "//src/core:lib/iomgr/iomgr.cc",
         "//src/core:lib/iomgr/iomgr_posix.cc",
@@ -1334,7 +1332,6 @@ grpc_cc_library(
         "//src/core:lib/compression/compression_internal.h",
         "//src/core:lib/compression/message_compress.h",
         "//src/core:lib/iomgr/block_annotate.h",
-        "//src/core:lib/iomgr/buffer_list.h",
         "//src/core:lib/iomgr/call_combiner.h",
         "//src/core:lib/iomgr/cfstream_handle.h",
         "//src/core:lib/iomgr/dynamic_annotations.h",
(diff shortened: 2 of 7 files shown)


src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -32,18 +31,46 @@ namespace grpc_core {
...
 
+  // Executes the passed function \a cb with each context in the list. The
+  // arguments provided to cb include the trace_context_, byte_offset_,
+  // traced_bytes_relative_start_pos_ and num_traced_bytes_ for each context in
+  // the context list. It also frees up the entire list after this operation.
+  // The cb receives individual contexts in the same order in which they were
+  // Appended.
+  static void ForEachExecuteCallback(
+      ContextList* tail,
+      std::function<void(void*, size_t, int64_t, int64_t)> cb);
```
ctiller, line 65: Prefer `absl::FunctionRef` - we don't need to hold this function longer

src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -32,18 +31,46 @@ namespace grpc_core {
...
+  // sent over the wire. A data chunk being written over the wire is multiplexed
+  // with bytes from multiple RPCs. If one such RPC is traced, we store the
+  // following information about the traced RPC:
+  // - byte_offset_in_stream: Number of bytes belonging to that traced RPC which
+  // have been sent so far from the start of the RPC stream.
+  // - relative_start_pos_in_chunk: Starting offset of the traced RPC within
+  // the current chunk that is being sent.
+  // - num_traced_bytes_in_chunk: Number of bytes belonging to the traced RPC
+  // within the current chunk.
+  static void Append(ContextList** tail, void* context,
```
ctiller, line 45: This API is very confusing & C-like... if we're making changes can we consider doing something a little simpler here?

I'd recommend gathering up all the elements in a `std::vector` so that we get some better cache coherency properties than what's here... once we do that we can be talking about just a `std::vector<ContextEntry>` with `ContextEntry` containing our handful of members and no complicated pointer chasing code.

ctiller (changes requested):

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: Fixed

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: updated it to use std::vector type. It also required some changes to chttp2_transport

src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -32,18 +36,59 @@ namespace grpc_core {
 class ContextList {
```
ctiller, line 36: Next question - and I don't have an answer.

Do we need this class or would just having `ContextListEntry` be enough?

Argument against having it:

Append is just an `emplace_back` if `ContextListEntry` has the right constructor.
IsEmpty is trivial.
MakeNewContextList is redundant (just instantiate a `std::vector<ContextListEntry>`)
`ForEachExecuteCallback` is just a really complicated way of writing `for (const auto& ent : entries) { ... }`

yousukseung: Or can we refactor in a separate PR? I guest we can just remove ContextList and replace it with vector<ContextListEntry> with a little more work, but given we have a customer who wants to enable Dapper asap we don't need to rush this in this PR.

ctiller: So a few things:
- the code you submit usually lasts longer than you'd like
- there's always another urgent customer
- added complexity tends to breed more complexity

Taken together, if there's a really quick change to wipe out an entire module I think we should take that... It doesn't significantly delay things now, and it speeds us up on the next thing.

yousukseung: Sure, I think it's ultimately you and Vignesh's call. The reason I proposed to separate them out is that I expect we need the Dapper support enabled for the customer asap within days, then it's worth not mixing any changes.

But can we at least remove the two additional args here and make this a pure refactor PR? I think we can add them in the "Dapper" PR and make it one, I expect this will make (unlikely) rollback of this refactor easier too.

src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -32,18 +36,59 @@ namespace grpc_core {
...
-  // Executes a function \a fn with each context in the list and \a ts. It also
-  // frees up the entire list after this operation. It is intended as a callback
-  // and hence does not take a ref on \a error
+  // Interprets the passed arg as a pointer to ContextList type and executes the
+  // function set using grpc_http2_set_write_timestamps_callback method with
+  // each context in the list and \a ts. It also deletes/frees up the passed
+  // ContextList after this operation. It is intended as a callback and hence
+  // does not take a ref on \a error. The fn receives individual contexts in the
+  // same order in which they were Appended.
   static void Execute(void* arg, Timestamps* ts, grpc_error_handle error);
```
yousukseung, line 62: This can also be made non-static like Append, correct? IIUC arg is always "this". I see this will require a cherry-pick to update google's init.cc together but this looks simple.

src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -32,18 +36,59 @@ namespace grpc_core {
...
 
+  // Executes the passed function \a cb with each context in the list. The
+  // arguments provided to cb include the trace_context_, byte_offset_,
+  // traced_bytes_relative_start_pos_ and num_traced_bytes_ for each context in
+  // the context list. It also deletes/frees up the ContextList after this
+  // operation. The cb receives individual contexts in the same order in which
+  // they were Appended.
+  static void ForEachExecuteCallback(
```
yousukseung, line 70: Same here, I think this can be made non-static as well?

src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -32,18 +36,59 @@ namespace grpc_core {
...
+      ContextList* list,
+      absl::FunctionRef<void(void*, size_t, int64_t, int64_t)> cb);
+
+  // Use this function to create a new ContextList instead of creating it
+  // manually.
+  static ContextList* MakeNewContextList() { return new ContextList; }
```
yousukseung, line 76: Do we need this?

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: Its no longer relevant now in the latest commit since we are directly dealing with the std::vector type.

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: Its no longer relevant now in the latest commit since we are directly dealing with the std::vector type.

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: Its no longer relevant now in the latest commit since we are directly dealing with the std::vector type.

src/core/ext/transport/chttp2/transport/context_list.h
yousukseung: I think it's better to keep ContextList as a thin wrapper class instead of exposing the vector directly.

Both Append() and Execute() are non-trivial and the underlying vector is only supposed to be updated in specific ways, so I think they should remain in one place and hidden. I assume this simply requires moving a few lines of code back.

src/core/ext/transport/chttp2/transport/chttp2_transport.cc
```diff
@@ -214,6 +215,9 @@ grpc_core::CallTracerInterface* CallTracerIfEnabled(grpc_chttp2_stream* s) {
           s->context)[GRPC_CONTEXT_CALL_TRACER_ANNOTATION_INTERFACE]
           .value);
 }
+
+grpc_core::WriteTimestampsCallback write_timestamps_callback_g = nullptr;
```
ctiller, line 219: `g_write_timestamps_callback` - we always prefix the g for global, never suffix it

ctiller: (same below too)

Vignesh2208: Fixed.

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: The pre-existing APIs seem a bit non intuitive. I unrolled the Append() logic directly into the chttp2/writing.cc  file because its the only place its used. 

By dealing with the vector directly, other internal EventEngine implementations can directly iterate over the ContextListEntries instead of exposing another API to the wrapper ContextList class to iterate and process them.

@ctiller do you agree ?

yousukseung: It's your call, but I think this is a clear case to hide internals. There are only two operations that are expected and current APIs look reasonable to me. We don't have unit tests which I think we need, but if we had them it might be another good reason to hide unnecessary internals and only test the APIs. Append() being called from one place seems less relevant given this is more about the structure, and I don't think we have other places that need to iterate over them now.

src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -26,74 +26,43 @@
...
 
-  // Use this function to create a new ContextList instead of creating it
-  // manually.
-  static ContextList* MakeNewContextList() { return new ContextList; }
+  void* TraceContext() { return trace_context_; }
+  int64_t RelativeStartPosInChunk() { return relative_start_pos_in_chunk_; }
+  int64_t NumTracedBytesInChunk() { return num_traced_bytes_in_chunk_; }
+  size_t ByteOffsetInStream() { return byte_offset_in_stream_; }
 
  private:
```
yousukseung, line 57: Since you already refactored, can we explicitly delete the default ctor or make it private? i.e. ContextListEntry() = delete;

src/core/ext/transport/chttp2/transport/context_list.h
```diff
@@ -22,33 +22,47 @@
 #include <grpc/support/port_platform.h>
```
ctiller, line 22: Nit: this file should now be called context_list_entry.h

ctiller (approved):

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: Fixed

src/core/ext/transport/chttp2/transport/context_list.h
Vignesh2208: Fixed

Issue #24665, opened 2020-11-04
Bazel rules should use the python headers from the python toolchain


dhalperi:
### Is your feature request related to a problem? Please describe.
The recommend way to use Python with Bazel is by providing a [Python toolchain](https://github.com/bazelbuild/bazel/blob/master/tools/python/toolchain.bzl#L65).

gRPC's `cython_library` rule, however, uses the [system Python headers](https://github.com/grpc/grpc/blob/a951c7632a6a52c8dc7920105d0dd7295b795469/bazel/cython_library.bzl#L61) instead of the headers for the Python toolchain that is being used.

### Describe the solution you'd like
Use the headers for the Python toolchain selected for the given platform.

### Describe alternatives you've considered
The only alternative I've found is forcing users to keep their system environment in sync with whatever the Bazel Python toolchain they're using is. This, of course, defeats the purpose of adding a Bazel toolchain - isolation from the system environment!

### Additional context
Add any other context about the feature request here.

dhalperi:
This is stale because no one has engaged. I would love a response from the gRPC team :)

aaliddell:
This almost works and I think could be fixed with a one-line change, at least in the short-term.

Presently, the `python_configure` is run which autodiscovers the local Python, then the headers are bound to the name `python_headers`:

https://github.com/grpc/grpc/blob/afc040b07a8dcab704fdcab248c8f8b764fafd6c/bazel/grpc_python_deps.bzl#L56-L61

However, when Cython is run it is linked against the non-aliased headers:

https://github.com/grpc/grpc/blob/afc040b07a8dcab704fdcab248c8f8b764fafd6c/bazel/cython_library.bzl#L71-L76

Therefore, the proposed change would be for `pyx_library` to instead use the binding `@python_headers`, thereby allowing a user to override the binding of `python_headers` in their WORKSPACE to point to the correct headers location. This is the same as how protobuf works when building the C extension: https://github.com/protocolbuffers/protobuf/blob/master/util/python/BUILD

adam-azarchs:
One problem with that solution is that it perpetuates the problems inherent `bind`.

Another more serious problem is that it still won't work if, as in for example my case, the system does not have a version of python installed that matches the version from your toolchain.  For example our python toolchains are configured for python 3.8, so `local_config_python` fails with a message like
```
Error in fail: Python Configuration Error: Unable to find Python headers for /usr/bin/python3
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #32829, opened 2023-04-06
[compiler] python: don't generate empty files for sources that do not define any services

### LINKED ISSUE #32763: python: generator produces useless/empty files for proto files that do not define services
The generator for python generates mostly empty (and totally useless) output files when processing a Proto source file that does not define any services.

First reported to Buf in our Slack.
https://bufbuild.slack.com/archives/CRZ680FUH/p1679937428898619

The issues seems to be [here](https://github.com/grpc/grpc/blob/master/src/compiler/python_generator.cc#L840-L846): it gets to this point _unconditionally_ and then calls `context->Open`, so it always produces a file, even if it will not actually generate any useful code for it.

Before it gets that far, probably around [here](https://github.com/grpc/grpc/blob/master/src/compiler/python_generator.cc#L890), it should query `file->service_count()`: if the count is zero, it could just immediately return and generate nothing.

jhump:
Fixes #32763.

I can't seem to find a super-easy way to test this. I tried running `bazel test //test/...` and half of them fail even without this patch. And I also ended up running into #24665 and didn't see any trivial work-arounds in that issue.

Hopefully it's fair to assume that whatever CI jobs will run against this PR will suss out anything that this breaks (hopefully nothing 🤞)?

### DIFF
--- a/src/compiler/python_generator.cc
+++ b/src/compiler/python_generator.cc
@@ -887,6 +887,11 @@ bool PythonGrpcGenerator::Generate(const FileDescriptor* file,
                                    const std::string& parameter,
                                    GeneratorContext* context,
                                    std::string* error) const {
+  if (file->service_count() == 0) {
+    // Nothing to do be done for this file.
+    return true;
+  }
+
   // Get output file name.
   std::string pb2_file_name;
   std::string pb2_grpc_file_name;


jhump:
Of the three failing CI jobs, one does not have a link, so I can't see any details ("import/copybara"). One is about labels, which a maintainer must set ("Mergeable"). The other ("Bazel Distribtest") is confusing: it says i failed, but when I look at the test output, I can't see what failed. I see six headings that summarize running tests (alternating 1 test case, then 5, then 1, then 5, etc), and all are reported as successful. So I don't know if I actually broke something or not. (I think maybe not.)

@gnossen, could you advise on if any of the red CI jobs are real problems? If none of them are, can you take a look at this (very tiny) PR?

sergiitk:
Closing as stale.

PR #32830, opened 2023-04-07
[BUILD files] use grpc_package() everywhere

teeler:
consolidate back to the native package rule

`grpc_package` was just a thin wrapper, and the `name` attribute was unused. This has a bit less overhead, and allows for easier copybara transforms in/around/on packages


teeler:
hang on, gotta revisit this.

veblush:
I'm not sure whether this is acceptable as this `grpc_package` was made to support various visibility in Google3. You can search the internal one `/grpc/bazel/grpc_build_system.bzl`

PR #32833, opened 2023-04-07
[upb] Fixed mini_table.hpp's implicit capture of this for C++20

sulmone:
external/upb/upb/mini_table.hpp: In lambda function:
external/upb/upb/mini_table.hpp:78:9: error: implicit capture of 'this' via '[=]' is deprecated in C++20 [-Werror=deprecated]
   78 |         [=](char* buf) { return upb_MtDataEncoder_EndEnum(&encoder_, buf); });
      |         ^
external/upb/upb/mini_table.hpp:78:9: note: add explicit 'this' or '*this' capture


yashykt:
This is a third_party submodule. You would need to make the change in the upb repo itself https://github.com/protocolbuffers/upb

PR #32834, opened 2023-04-07
[JSON] Replace ctors with factory methods

markdroth:


ctiller (approved):
There's extra verbosity here but I think it's helping - having it immediately obvious that we're talking about a string vs an array during construction of some of this is actually huge, and I feel like I can follow what's going on better here.

PR #32836, opened 2023-04-07
[core] Add an assertion to catch the environment variable emptiness issue.

mikefan1991:


ctiller:
We can't use the assert functions defined in log.h here because they use configuration loaded using this mechanism.

That said this seems ill advised... Isn't this a normal situation we're turning into a crash? Albeit a debug only crash, but I thing the problem we saw was a release build?

mikefan1991:
What should be a better way to catch this instead of using assert? I would like to catch it under debug mode.

mikefan1991:
Although we saw the issue in release build, my change here is to prevent it from going into the release build. It can catch emptiness during development.

ctiller:
I think an assert this broad (every config var we have) is destined to be problematic - experiments and trace for example have default empty, and most often have their values empty too.

An assert is fine, but you need to narrow the scope down to just the property that you need (this specific config var is not empty?) - you also need to get the code compiling so that you can test this and demonstrate that it's doing the thing you think it is.

mikefan1991:
> I think an assert this broad (every config var we have) is destined to be problematic - experiments and trace for example have default empty, and most often have their values empty too.
> 
> An assert is fine, but you need to narrow the scope down to just the property that you need (this specific config var is not empty?) - you also need to get the code compiling so that you can test this and demonstrate that it's doing the thing you think it is.

For this specific function, I don't see the reason of passing in two empty strings. It will clearly return nothing if both parameters are empty. So for this scenario, I think this function should not be called at the first place. WDYT?

mikefan1991:
Hi @ctiller, can you review the PR so that I can proceed with the process. Thank you!

ctiller (commented):
This looks fine so far... please address the failing tests (especially look at the sanity tests) - looks like we need to regenerate projects (`tools/buildgen/generate_projects.sh`)

mikefan1991:
The projects are updated after running tools/buildgen/generate_projects.sh. And the only two failing tests are not related to my change. Please take a look again. Thanks!

ctiller (approved):

ctiller (approved):

ctiller (approved):

Issue #15411, opened 2018-05-16
Re-enable broken tests with gevent and unpin gevent on Windows


kpayson64:
When gevent 1.3.0 was released, our windows tests started hanging.

#15352

We should unpin once we figure out the issue on windows.

lidizheng:
Unit test `unit._auth_context_test.AuthContextTest.testSessionResumption` hangs forever when using higher version `gevent`. Pull request: #16995

ericgribkoff:
Using this as a catch-all for gevent disabled tests, which as of https://github.com/grpc/grpc/pull/17466 include the `ExitTests`.

lidizheng:
Disable gevent DNS resolver test on Windows `unit._dns_resolver_test.DNSResolverTest.test_connect_loopback`. Details in https://github.com/grpc/grpc/pull/18547

gnossen:
Still valid.

gnossen:
Afraid not. [The test is actually being skipped even though it's being run.](https://github.com/grpc/grpc/blob/20a53246961280edca136de72804ced71ca8419a/src/python/grpcio_tests/tests/unit/_dynamic_stubs_test.py#L129)

PR #32842, opened 2023-04-10
[Python fix-it][Gevent] Enable gevent _dynamic_stubs_test

XuanWang-Amos:
Re-enable this test since it's now passing after gevent reimplementation.

Fix: https://github.com/grpc/grpc/issues/15411


gnossen (approved):
🎉

PR #32849, opened 2023-04-11
Update fix_build_deps.py

ctiller:


ctiller:
Leaving this open as a teaching tool for a little bit.

PR #32853, opened 2023-04-11
[fuzzing] Define a common fuzzing interface and move API fuzzer to it

Vignesh2208:
Requires cherrypick for grpc_fuzzer.bzl file.


test/core/end2end/fuzzers/BUILD
```diff
@@ -12,31 +12,51 @@
...
 
-load("//bazel:grpc_build_system.bzl", "grpc_package")
+load("//bazel:grpc_build_system.bzl", "grpc_cc_library", "grpc_package", "grpc_proto_library")
 load("//test/core/util:grpc_fuzzer.bzl", "grpc_fuzzer", "grpc_proto_fuzzer")
 
 grpc_package(name = "test/core/end2end/fuzzers")
 
 licenses(["notice"])
 
+grpc_proto_library(
```
ctiller, line 22: What's the argument for a single top level message across disparate fuzzers vs the proposed different top-level fuzzer message per fuzzer that depends-on the api-fuzzer method bits?

yijiem (commented):
It would be nice to run the fuzzer with the current corpus before and after the change and make sure we still get the same coverage. I tried a bit with Clang Coverage (https://clang.llvm.org/docs/SourceBasedCodeCoverage.html), but it does not seem to work properly (I can keep looking).

test/core/end2end/fuzzers/BUILD
Vignesh2208: I think it will be cleaner to just encode the different top level fuzzer message into a separate fuzzer action and have different fuzzers return whether they support/don't support the specific action. We would then even be able to keep the fuzzing loop common among all fuzzers.

Vignesh2208:
> It would be nice to run the fuzzer with the current corpus before and after the change and make sure we still get the same coverage. I tried a bit with Clang Coverage (https://clang.llvm.org/docs/SourceBasedCodeCoverage.html), but it does not seem to work properly (I will keep looking).

I did run the fuzzer against the corpus with `bazel test //test/core/end2end/fuzzers:api_fuzzer --test_output=all` command and it passed all test-cases

yijiem (approved):

PR #32863, opened 2023-04-13
[OpenCensus C++ Test] Don't depend on compression algorithm

yashykt:
We shouldn't depend on how much the compression algorithm compresses the bytes to. This is causing flakiness internally.


test/cpp/ext/filters/census/stats_plugin_end2end_test.cc
```diff
@@ -800,9 +800,9 @@ TEST_F(StatsPluginEnd2EndTest, TestMessageSizeWithCompressionAnnotations) {
   EXPECT_TRUE(
       IsAnnotationPresent(server_span_data, "Send message: 1026 bytes"));
   EXPECT_TRUE(IsAnnotationPresent(attempt_span_data,
-                                  "Send compressed message: 31 bytes"));
-  EXPECT_TRUE(
-      IsAnnotationPresent(server_span_data, "Received message: 31 bytes"));
+                                  "Send compressed message:" /*  31 bytes */));
```
drfloob, line 803: Suggest removing the byte number comment, which we know is occasionally incorrect, and it may drift further since it's not being tested.

yashykt: done

PR #32869, opened 2023-04-14
[EventEngine] Implement work-stealing in the EventEngine ThreadPool

drfloob:
This PR implements a work-stealing thread pool for use inside EventEngine implementations. Because of historical risks here, I've guarded the new implementation behind an experiment flag: `GRPC_EXPERIMENTS=work_stealing`. Current default behavior is the original thread pool implementation.

Benchmarks look very promising:

```
bazel test \
--test_timeout=300 \
--config=opt -c opt \
--test_output=streamed \
--test_arg='--benchmark_format=csv' \
--test_arg='--benchmark_min_time=0.15' \
--test_arg='--benchmark_filter=_FanOut' \
--test_arg='--benchmark_repetitions=15' \
--test_arg='--benchmark_report_aggregates_only=true' \
test/cpp/microbenchmarks:bm_thread_pool
```

2023-05-04: `bm_thread_pool` benchmark results on my local machine (64 core ThreadRipper PRO 3995WX, 256GB memory), comparing this PR to master:

image

2023-05-04: `bm_thread_pool` benchmark results in the Linux RBE environment (unsure of machine configuration, likely small), comparing this PR to master. 

image


src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -0,0 +1,459 @@
...
+      grpc_core::Thread::Options().set_tracked(false).set_joinable(false))
+      .Start();
+}
+
+void WorkStealingThreadPool::WorkStealingThreadPoolImpl::Lifeguard::
+    LifeguardMain() {
+  thread_running_.store(true);
+  while (true) {
+    absl::SleepFor(absl::Milliseconds(
+        (backoff_.NextAttemptTime() - grpc_core::Timestamp::Now()).millis()));
```
Vignesh2208, line 237: Why do we backoff exponentially here instead of having a fixed sleep duration ?

Vignesh2208 (commented):
Hi AJ, reviewed the core logic and left a few suggestions. I haven't reviewed the tests yet.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -0,0 +1,459 @@
...
+
+void WorkStealingThreadPool::TheftRegistry::Unenroll(WorkQueue* queue) {
+  grpc_core::MutexLock lock(&mu_);
+  queues_.erase(queue);
+}
+
+EventEngine::Closure* WorkStealingThreadPool::TheftRegistry::StealOne() {
+  grpc_core::MutexLock lock(&mu_);
+  EventEngine::Closure* closure;
+  for (auto* queue : queues_) {
```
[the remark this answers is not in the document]
drfloob: The bucketing idea is interesting. We could also institute queue priorities, I've seen some good results there. I just wonder if the overhead of queue juggling would negate the benefits of optimizing for queue time.

> One potential future optimization here could be that if we have a data structure that returns the queue with the highest backlog (in-terms of queue length), it might be preferable to steal from that queue first.

Maybe so. In my chats with @soheilhy, optimizing for queue time was not terribly fruitful in their experiments. Further, this is pop-most-recent because we have a high performance queue implementation that avoids mutexes for LIFO operations. It has some very rare atomic flake issues, so I planned to land that as a subsequent improvement.

But these are all things we can experiment with if the performance difference is meaningful.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
drfloob: Benchmarks and efficiency :-)

If the pool is highly active, we want a vigilant lifeguard because it will wake idle workers faster than the workers will wake themselves. If gRPC is idle, having the lifeguard wake up every 50 millis is needlessly expensive.

test/cpp/microbenchmarks/bm_basic_work_queue.cc
```diff
@@ -0,0 +1,205 @@
...
+    int cnt = 0;
+    do {
+      if (++pop_attempts && globalWorkQueue.PopMostRecent() != nullptr) ++cnt;
+    } while (cnt < element_count);
+  }
+  state.counters["added"] = element_count * state.iterations();
+  state.counters["pop_rate"] = benchmark::Counter(
+      element_count * state.iterations(), benchmark::Counter::kIsRate);
+  state.counters["pop_attempts"] = pop_attempts;
+  state.counters["hit_rate"] =
```
Vignesh2208, line 89: Can you add a comment on what the hit_rate is measuring ?

test/core/event_engine/thread_pool_test.cc
```diff
@@ -12,117 +12,204 @@
...
   // Our first thread pool implementation tried to create ~1M threads for this
   // test.
-  ScheduleTwiceUntilZero(&p, 20);
+  ScheduleTwiceUntilZero(&p, runcount, 20);
+  p.Quiesce();
+  ASSERT_EQ(runcount.load(), pow(2, 21) - 1);
+}
+
+TYPED_TEST(ThreadPoolTest, ScalesWhenBackloggedFromSingleThreadLocalQueue) {
+  int pool_thread_count = 8;
```
Vignesh2208, line 154: Is there an upper limit on the number of threads spawned by the thread pool ? If not, can you create some k* pool_thread_count callbacks in line 161 and line 187 to verify the lifeguard is able to create more and more threads as necessary ?

drfloob: There is no upper limit. As is, this test ensures that scaling happens, which is essential. I'm not sure if we can safely specify a minimum scale amount to assert for all implementations. Do you think some N threads would make this a better test?

The biggest issue with "more and more (N) threads" is that the pool implementation creates at most 1 thread per second, so the test could take a while depending on N.

test/core/event_engine/thread_pool_test.cc
```diff
@@ -12,117 +12,204 @@
...
-  gpr_log(GPR_INFO, "wait for notification");
+  p.Run([&n2] { n2.Notify(); });
   n2.WaitForNotification();
   p.Quiesce();
 }
 
-void ScheduleSelf(ThreadPool* p) {
-  p->Run([p] { ScheduleSelf(p); });
+TYPED_TEST(ThreadPoolTest, ForkStressTest) {
+  // Runs a large number of closures and multiple simulated fork events,
```
Vignesh2208, line 85: Can you add more comments for this test ? Is it testing that a Fork operation is not blocked indefinitely ?

drfloob: Done

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
Vignesh2208: Doesn't this mean that through line 267, the backoff timer keeps multiplying even for the case where there is some idle thread AND work to be done (i.e work queue is not empty) ? 

Shouldn't the backoff timer multiply only for the case where there are idle threads but no work needs to be done (i.e the pool is empty) ?

Vignesh2208 (commented):
Mostly looks good. Left a few clarifications.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
drfloob: Indeed, good catch. Fixed.

drfloob:
> Mostly looks good. Left a few clarifications.

Thanks Vignesh! I've updated the PR and responded to all comments.

Vignesh2208 (approved):

drfloob:
A test cherrypick was clean, no need for it. Thanks for the review!

PR #32870, opened 2023-04-14
[Core] Fix the build error caused by the incorrect logical operator.

James-Bao:


yashykt:
which platform are you running on to get this error?

James-Bao:
> which platform are you running on to get this error?

I was using x86_64 GNU/Linux on Amazon EC2 instance.

eugeneo:
What compiler are you using? This is compatible with the spec - https://en.cppreference.com/w/cpp/keyword/and

PR #32871, opened 2023-04-14
Support LLVM-Mingw

hongnod:
make Grpc built with LLVM-Mingw


drfloob:
Thanks for the contribution, @hongnod. Two things:

1) CMakeLists.txt is a generated file and will be overwritten. You'll need to change the template to make it permanent: https://github.com/grpc/grpc/blob/master/templates/CMakeLists.txt.template

2) Why is this not necessary for GCC? Seems like it would be worth adding this define for any mingw build https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/strsafe.h#l1928

hongnod:
> Thanks for the contribution, @hongnod. Two things:
> 
> 1. CMakeLists.txt is a generated file and will be overwritten. You'll need to change the template to make it permanent: https://github.com/grpc/grpc/blob/master/templates/CMakeLists.txt.template
> 2. Why is this not necessary for GCC? Seems like it would be worth adding this define for any mingw build https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/strsafe.h#l1928

As I know, grpc can be built with Mingw-64-ucrt64 with modification. the error message shows it possibly relates to libc++.

Issue #32564, opened 2023-03-08
Support .pyi generation in py_proto_library


ferstlf:
Hi!

### Is your feature request related to a problem? Please describe.
The general problem: When working with protos/gRPC and Bazel in an IDE i'd like to have proper language support, i.e. import resolution, auto-completion etc. should work for automatically generated proto classes. 

With Bazel, import resolution is a bit tricky, but essentially I can configure in my IDE/language server the relevant Python import paths inside Bazels temporary folders (e.g. `bazel-bin` or `bazel-bin/external/com_github_grpc_grpc`). However, the .py files generated by `protoc` use Metaclasses and language servers such as Pylance in VS Code do not understand these to offer proper auto-completion and more. 

Since [Protobuf v3.20.0](https://github.com/protocolbuffers/protobuf/releases/tag/v3.20.0), `protoc` is natively able to generate python stubs (.pyi) with the `--pyi_out` flag. If one generates a *_pb2.pyi file for every *_pb2.py file this should be understood by most language servers automatically and things like auto-completion and static type-checking with proto classes should just work  (I can at least confirm for Pylance/VS Code).

### Describe the solution you'd like
We use [gRPC's `py_proto_libary` rule](https://github.com/grpc/grpc/blob/cd4154f21d80b278a08f5de2a3450a0150744c69/bazel/python_rules.bzl#L154) in our Bazel workspace and also expect our users to use it in their workspaces (which depend on our workspace). It would be great if the `py_proto_libary` rule would generate .pyi files in addition to .py files - either by default (since it is not very costly) or enabled via a boolean flag on the `py_proto_libary` rule.

It should be as simple as adding the `--pyi_out` flag to the `protoc` params [here](https://github.com/grpc/grpc/blob/cd4154f21d80b278a08f5de2a3450a0150744c69/bazel/python_rules.bzl#L67) and then making sure that `py_proto_libary` correctly includes or excludes the additionally generated .pyi files from the output files, runfiles and PyInfo which it declares as its output.

### Additional context

Let me know if this would make sense to you. I already have this working as a patch to our Bazel import of the gRPC workspace. I could try to turn this into a PR myself.

XuanWang-Amos:
[...]

PR #32872, opened 2023-04-14
[bazel] Auto-generate .pyi files in py_proto_library rule

ferstlf:
With some delay, this is a PR for https://github.com/grpc/grpc/issues/32564 (and previously https://github.com/grpc/grpc/pull/31791). 

I looked into adding a regular `py_test` for this change [as suggested](https://github.com/grpc/grpc/pull/31791#issuecomment-1423245116) but I am not aware of any effect that the presence of a .pyi stub file would have at runtime and where some sort of type-checking in a .py script would be affected. Stub files are only for use by type checkers & IDE's. I mean, something like this would work:

```
import helloworld_pb2
py_file = helloworld_pb2.__file__ 
pyi_file = py_file + 'i’
self.assertTrue(os.path.exists(pyi_file))
```

But that seems really hacky to me. Instead I created a simple rule test for `py_proto_library` with Bazel Skylib which tests the declared outputs for an example `py_proto_library`  target. Indirectly, this also tests that the declared output files are actually generated. Please let me know if this is sufficient.


ferstlf:
Friendly bump!

gnossen:
Unfortunately, it looks like adding a dependency on Bazel Skylib introduces a thorny issue:

```
(venv) rbellevi@rbellevi:~/dev/grpc/test/distrib/bazel/python$ tools/bazel test -c dbg //:py_proto_library_provider_contents_test
INFO: Running bazel wrapper (see //tools/bazel for details), bazel version 6.1.2 will be used instead of system-wide bazel installation.
ERROR: /usr/local/google/home/rbellevi/.cache/bazel/_bazel_rbellevi/6b245ae2ccca35fac920f4dfd855cf78/external/bazel_tools/platforms/BUILD:89:6: in alias rule @bazel_tools//platforms:windows: Constraints from @bazel_tools//platforms have been removed. Please use constraints from @platforms repository embedded in Bazel, or preferably declare dependency on https://github.com/bazelbuild/platforms. See https://github.com/bazelbuild/bazel/issues/8622 for details.
ERROR: /usr/local/google/home/rbellevi/.cache/bazel/_bazel_rbellevi/6b245ae2ccca35fac920f4dfd855cf78/external/bazel_tools/platforms/BUILD:89:6: Analysis of target '@bazel_tools//platforms:windows' failed
ERROR: /usr/local/google/home/rbellevi/dev/grpc/test/distrib/bazel/python/BUILD:160:24: While resolving toolchains for target //:py_proto_library_provider_contents_test: invalid registered toolchain '@bazel_skylib//toolchains/unittest:cmd_toolchain': 
ERROR: Analysis of target '//:py_proto_library_provider_contents_test' failed; build aborted: 
INFO: Elapsed time: 0.355s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 2 targets configured)
ERROR: Couldn't start the build. Unable to run tests
```

[This only happens on Bazel 6+](https://github.com/bazelbuild/bazel/issues/8622), but we need to run these tests against all supported Bazel versions, including 6+. We don't use platform constraints ourself, but several of our dependencies do:

```
(venv) rbellevi@rbellevi:~/dev/grpc$ grep -rn "bazel_tools//platforms"
third_party/bloaty/third_party/protobuf/third_party/googletest/BUILD.bazel:41:    constraint_values = ["@bazel_tools//platforms:windows"],
third_party/bloaty/third_party/googletest/BUILD.bazel:41:    constraint_values = ["@bazel_tools//platforms:windows"],
third_party/bloaty/third_party/abseil-cpp/absl/BUILD.bazel:47:        "@bazel_tools//platforms:osx",
third_party/bloaty/third_party/abseil-cpp/absl/BUILD.bazel:54:        "@bazel_tools//platforms:ios",
third_party/bloaty/third_party/abseil-cpp/absl/time/internal/cctz/BUILD.bazel:29:        "@bazel_tools//platforms:osx",
third_party/bloaty/third_party/abseil-cpp/absl/time/internal/cctz/BUILD.bazel:36:        "@bazel_tools//platforms:ios",
```

So fixing _that_ issue would become a big yak shaving exercise.

Instead of using Bazel Skylib, could we do something more end-to-end? How are these `.pyi` files going to be consumed by you under Bazel builds? Presumably you also have a type checker like `pytype` or `mypy` integrated with your Bazel build? Could we just have a smoke check that a simple proto passes type check with this change?

ferstlf:
Thanks for the reply, I can reproduce this problem. I was using Bazel 5.4.0 when creating this PR.

Currently we don't consume the .pyi files with Bazel (or a type checker integrated with Bazel) but with Pylance/VS Code to get import resolution and auto-completion. There is a bit more to our setup, but in a nut-shell we add Bazel's output paths such as `bazel-bin/external/com_github_grpc_grpc/src/python/grpcio` to the analysis paths for Pylance and then, of course, Pylance understands seeing `foo.py` next to `foo.pyi`.

I haven't worked with type checkers in Bazel but I could give it a shot. But before I do: would it be an option to simply update to a newer version of Skylib? It seems like Skylib migrated their use of platform constraints in [this commit](https://github.com/bazelbuild/bazel-skylib/commit/327d61b5eaa15c11a868a1f7f3f97cdf07d31c58). All we'd need is Skylib version 1.0.3 instead of 1.0.2 in [grpc_deps()](https://github.com/grpc/grpc/blob/a90239f381e9c95b51aa8b33728923f71b971120/bazel/grpc_deps.bzl#L361-L369). I tentatively updated this PR with the newer skylib version - just so you can try it out (if bumping the version is in fact an option). Tested successfully with Bazel 5.4.0 and 6.1.2.

gnossen:
> would it be an option to simply update to a newer version of Skylib

Sure! As long as it works with all of our versions of Bazel. I just triggered CI so we can find out.

ferstlf:
AFAICT the newer Skylib version seems to have solved the issue?!

gnossen (approved):

PR #32878, opened 2023-04-14
[Fuzzing] Migrate client and server_fuzzer to structured fuzzing

yijiem:
- Added `fuzzer_input.proto` and `NetworkInput` proto message
- Migrated client_fuzzer and server_fuzzer to proto fuzzer
- Migrated the existing corpus and verified that the code coverage (e.g. chttp2) stays the same

Probably need to cherrypick due to amount of files changed.


ctiller (commented):
LGTM. I don't see any need to cherry pick - yes lots of files, but they're all fuzzer examples - these are not high traffic things.

Vignesh2208 (approved):

Issue #32885, opened 2023-04-17
gRPC install fails when gRPC_BUILD_CODEGEN is OFF during cross platform build


pankaj9um:
During a cross platform build of gRPC, none of the plugin targets are created if gRPC_BUILD_CODEGEN  is set to OFF. 
But here https://github.com/grpc/grpc/blob/master/CMakeLists.txt#L26104 gRPCPluginTargets are installed without checking gRPC_BUILD_CODEGEN variable. This leads to failure while doing install of gRPC build.

matthewjmiller1:
I just ran across this too. I think it broke at this commit:
https://github.com/grpc/grpc/commit/831d2a6855da36998e24f46e6b2a5b6f6042ad1a

Wrote the Dockerfile below to demonstrate the issue. 1.51.0 will succeed but 1.52.0 and the latest release (1.54.0 as of today) fail with the message: `CMake Error: INSTALL(EXPORT) given unknown export "gRPCPluginTargets"`.

It breaks our cross compile since we only build the libraries on the target system and don't build the plugins or codegen for the target system.

```
FROM ubuntu:22.04

ENV DEBIAN_FRONTEND=noninteractive
ENV INSTALL_DIR=/usr/local

RUN apt-get update && \
    apt-get -y install \
    build-essential \
    git \
    cmake \
    vim \
    && rm -rf /var/lib/apt/lists/*

WORKDIR /git_repos

RUN git clone -b v1.51.0 https://github.com/grpc/grpc.git grpc-1.51.0
WORKDIR /git_repos/grpc-1.51.0
RUN git submodule update --init --recursive

WORKDIR /git_repos
RUN git clone -b v1.52.0 https://github.com/grpc/grpc.git grpc-1.52.0
WORKDIR /git_repos/grpc-1.52.0
RUN git submodule update --init --recursive

WORKDIR /git_repos
RUN git clone -b v1.54.0 https://github.com/grpc/grpc.git grpc-1.54.0
WORKDIR /git_repos/grpc-1.54.0
RUN git submodule update --init --recursive

WORKDIR /git_repos/grpc-1.51.0
RUN cmake -DgRPC_BUILD_CODEGEN="OFF" .

WORKDIR /git_repos/grpc-1.52.0
RUN cmake -DgRPC_BUILD_CODEGEN="OFF" .

WORKDIR /git_repos/grpc-1.54.0
RUN cmake -DgRPC_BUILD_CODEGEN="OFF" .

[...]

PR #32886, opened 2023-04-18
[cmake] fix install without gRPC_BUILD_CODEGEN 

clementperon:
When compiling without codegen cmake give the following error: CMake Error: INSTALL(EXPORT) given unknown export "gRPCPluginTargets".

Check gRPC_BUILD_CODEGEN variable is before trying to install gRPCPluginTargets.

Fix https://github.com/grpc/grpc/issues/32885

@herbrechtsmeier ok with this fix ?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26101,10 +26101,12 @@ if(gRPC_INSTALL)
     DESTINATION ${gRPC_INSTALL_CMAKEDIR}
     NAMESPACE gRPC::
   )
-  install(EXPORT gRPCPluginTargets
-    DESTINATION ${gRPC_INSTALL_CMAKEDIR}
-    NAMESPACE gRPC::
-  )
+  if(gRPC_BUILD_CODEGEN)
+    install(EXPORT gRPCPluginTargets
+      DESTINATION ${gRPC_INSTALL_CMAKEDIR}
+      NAMESPACE gRPC::
+    )
+  endif()
 endif()
 
 include(CMakePackageConfigHelpers)


matthewjmiller1:
One other potential use case to consider is if `gRPC_BUILD_CODEGEN=ON` but every `gRPC_BUILD_GRPC_*_PLUGIN=OFF` (e.g., `gRPC_BUILD_GRPC_CPP_PLUGIN`, `gRPC_BUILD_GRPC_PYTHON_PLUGIN`), I think you would still end up with the gRPCPluginTargets export being unknown.

It looks like the export is only set in the blocks like:
```
if(gRPC_BUILD_CODEGEN AND gRPC_BUILD_GRPC_CPP_PLUGIN)
```

I'm not sure if there is a use case for that or not where some of the codegen stuff is built but none of the plugins.

clementperon:
> One other potential use case to consider is if `gRPC_BUILD_CODEGEN=ON` but every `gRPC_BUILD_GRPC_*_PLUGIN=OFF` (e.g., `gRPC_BUILD_GRPC_CPP_PLUGIN`, `gRPC_BUILD_GRPC_PYTHON_PLUGIN`), I think you would still end up with the gRPCPluginTargets export being unknown.
> 
> It looks like the export is only set in the blocks like:
> 
> ```
> if(gRPC_BUILD_CODEGEN AND gRPC_BUILD_GRPC_CPP_PLUGIN)
> ```
> 
> I'm not sure if there is a use case for that or not where some of the codegen stuff is built but none of the plugins.

if (TARGET gRPCPluginTargets) maybe ?

veblush:
Superseded by https://github.com/grpc/grpc/pull/32986

PR #32890, opened 2023-04-19
[Fuzzing] Extend FuzzingEventEngine to create mock endpoints

Vignesh2208:
supports specifying timestamps at which specified bytes are read through the endpoint

### DIFF
--- a/test/core/event_engine/fuzzing_event_engine/fuzzing_event_engine.cc
+++ b/test/core/event_engine/fuzzing_event_engine/fuzzing_event_engine.cc
@@ -19,18 +19,22 @@
 #include <algorithm>
 #include <chrono>
 #include <limits>
+#include <memory>
 #include <ratio>
 #include <type_traits>
 #include <vector>
 
 #include "absl/memory/memory.h"
+#include "absl/status/status.h"
 #include "absl/strings/str_cat.h"
 
 #include <grpc/event_engine/slice.h>
+#include <grpc/status.h>
 #include <grpc/support/log.h>
 #include <grpc/support/time.h>
 
 #include "src/core/lib/event_engine/tcp_socket_utils.h"
+#include "src/core/lib/gprpp/status_helper.h"
 #include "src/core/lib/gprpp/time.h"
 #include "test/core/event_engine/fuzzing_event_engine/fuzzing_event_engine.pb.h"
 #include "test/core/util/port.h"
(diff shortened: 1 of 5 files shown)


test/core/event_engine/fuzzing_event_engine/fuzzing_event_engine.h
```diff
@@ -224,6 +233,48 @@ class FuzzingEventEngine : public EventEngine {
...
+  // bytes to be read when the absolute time advances and triggers the action.
+  class MockEndpoint final : public Endpoint {
+   public:
+    MockEndpoint(int local_port, int peer_port,
+                 const MockEndpointActions& actions) {
+      local_address_ = *URIToResolvedAddress(
+          absl::StrFormat("ipv4:127.0.0.1:%d", local_port));
+      peer_address_ = *URIToResolvedAddress(
+          absl::StrFormat("ipv4:127.0.0.1:%d", peer_port));
+      Init(actions);
```
ctiller, line 252: Can we lose the Init method and just put everything in the constructor?

test/core/event_engine/fuzzing_event_engine/fuzzing_event_engine.h
```diff
@@ -224,6 +233,48 @@ class FuzzingEventEngine : public EventEngine {
...
+      return peer_address_;
+    }
+    const ResolvedAddress& GetLocalAddress() const override {
+      return local_address_;
+    }
+
+   private:
+    ResolvedAddress local_address_;
+    ResolvedAddress peer_address_;
+    std::vector<std::tuple<Time, std::string>> actions_by_time_;
```
ctiller, line 275: I think there's a simpler implementation available:

If we have a `std::vector<uint8_t> bytes_available_to_read_;` here then our initialization loop can look like:

```
for (const MockEndpointAction& action : actions) {
  event_engine->RunAfter(action.time, [this]() {
    bytes_available_to_read_.extend(action.bytes);
  });
}
```

and `Read` itself becomes a fairly trivial copy.

(ok sans delayed reads, but I think they can fall out fairly easily with an `on_complete_` member also, since `AnyInvocable` is nullable)

test/core/event_engine/fuzzing_event_engine/fuzzing_event_engine.cc
```diff
@@ -534,5 +538,97 @@ FuzzingEventEngine::ListenerInfo::~ListenerInfo() {
...
+      on_readable(tcp_annotate_error(absl::InternalError("Socket Closed")));
+    } else {
+      // Present the first set of bytes as the data read at this time.
+      do {
+        buffer->Append(Slice::FromCopiedString(
+            std::get<std::string>(actions_by_time_.front())));
+        actions_by_time_.erase(actions_by_time_.begin());
+      } while (!actions_by_time_.empty() &&
+               g_fuzzing_event_engine->Now() >=
+                   std::get<Time>(actions_by_time_.front()));
```
yijiem, line 600: This seems to imply that `actions_by_time_` is a serial of actions sorted by timepoints (earlier one happens first). But the data structure doesn't guarantee that.

Maybe a `std::map<Time, std::string> actions_by_time_;` could work? Then this part would become:
```
while (!actions_by_time_.empty() &&
       actions_by_time_.begin()->first <= g_fuzzing_event_engine->Now()) {
  buffer->Append(Slice::FromCopiedString(actions_by_time_.begin()->second));
  actions_by_time_.erase(actions_by_time_.begin());
}
```

yijiem: If this is doable, I believe the insertion and concatenation part of the code in `Init()` can be changed accordingly. Let me know what you think!

yijiem: Oh sorry, I think I missed the point that `actions_by_time_` **is** a serial of actions sorted by timepoints. Please feel free to disregard this comment!

PR #32894, opened 2023-04-19
[Fix fuzzer error] Memory address points to zero page.

nanahpang:
Found memory access error in frame_fuzzer_test. Located the root cause in ExecCtx::Get(), where ExecCtx needs to be initialized before using HPackParser:ParseInput().

Error logs:
MemorySanitizer:DEADLYSIGNAL
==2812845==ERROR: MemorySanitizer: SEGV on unknown address 0x000000000030 (pc 0x55869275574e bp 0x7fffd7d9fb50 sp 0x7fffd7d9fb20 T2812845)
==2812845==The signal is caused by a READ memory access.
==2812845==Hint: address points to the zero page.
    #0 0x55869275574e in starting_cpu [third_party/grpc/src/core/lib/iomgr/exec_ctx.h:129](https://cs.corp.google.com/piper///depot/google3/third_party/grpc/src/core/lib/iomgr/exec_ctx.h?l=129&ws=ladynana/2900&snapshot=42):9
    #1 0x55869275574e in grpc_core::PerCpu<grpc_core::GlobalStatsCollector::Data>::this_cpu() [third_party/grpc/src/core/lib/gprpp/per_cpu.h:38](https://cs.corp.google.com/piper///depot/google3/third_party/grpc/src/core/lib/gprpp/per_cpu.h?l=38&ws=ladynana/2900&snapshot=42):48
    #2 0x558692753cda in IncrementHttp2MetadataSize [third_party/grpc/src/core/lib/debug/stats_data.h:265](https://cs.corp.google.com/piper///depot/google3/third_party/grpc/src/core/lib/debug/stats_data.h?l=265&ws=ladynana/2900&snapshot=42):11
    #3 0x558692753cda in grpc_core::HPackParser::ParseInput(grpc_core::HPackParser::Input, bool) [third_party/grpc/src/core/ext/transport/chttp2/transport/hpack_parser.cc:933](https://cs.corp.google.com/piper///depot/google3/third_party/grpc/src/core/ext/transport/chttp2/transport/hpack_parser.cc?l=933&ws=ladynana/2900&snapshot=42):20


ctiller:
I don't think this is the root cause. For the specific fuzzer failure this will stop the error, but the fuzzer will soon give us a 1-byte long frame and this will trip again.

I wrote https://github.com/grpc/grpc/pull/32896 to demonstrate that incrementing a zero frame size always works.

Here it looks like we're trying to access a null execctx because there's none on the stack. Probably just adding a `grpc_core::ExecCtx exec_ctx` at the top of the fuzzer entry function would be sufficient.

nanahpang:
Yes, it is due to the ExecCtx not initialized before use. I have updated the files and running the sanitizer script right now. Thanks!

ctiller (approved):
Will need to appease the sanitizer, but this looks good - thankyou!

ctiller (approved):

PR #32910, opened 2023-04-20
[Infra] Use Bazel 6 and drop Bazel 4 (Part 2)

veblush:
Oops I missed important changes from https://github.com/grpc/grpc/pull/32712. And it turned out that there are two problems that I couldn't fix at this point.
 - Windows Bazel RBE Linker Error: This may be caused by how new Bazel 6 invokes build tools chain but it's not clear. I put workaround to use Bazel 5 by using `OVERRIDE_BAZEL_VERSION=5.4.1`
 - Rule `rules_pods` to fetch CronetFramework from CocoaPod has incompatibility with sort of built-in apple toolchain. (https://github.com/bazel-xcode/PodToBUILD/issues/232): I couldn't find a workaround to fix this so I ended up disabling all tests depending this target.


yashykt (approved):

veblush:
@jtattermusch Windows RBE issue: I think we can live with this workaround for now but we need to fix it before long.

@sampajano As described above, I ended up disabling Cronet-related iOS Bazel tests. I'm not sure when rule_pod author can fix this problem with Bazel 6. Is it okay to disable these tests for now? If there is a workaround or fix to run Cronet tests without using rule_pod, it'd be great.

sampajano:
> @jtattermusch Windows RBE issue: I think we can live with this workaround for now but we need to fix it before long.
> 
> @sampajano As described above, I ended up disabling Cronet-related iOS Bazel tests. I'm not sure when rule_pod author can fix this problem with Bazel 6. Is it okay to disable these tests for now? If there is a workaround or fix to run Cronet tests without using rule_pod, it'd be great.

@veblush Thanks for checking! Yes, i think it's very much OK to disable these tests for now, as Cronet on iOS is being deprecated by the end of 2023, and hence i think our official support should also end soon.

Thanks!

PR #32911, opened 2023-04-21
[Compiler Fix] Bring declarations and definitions to be in sync

yashykt:


mint570:
This PR fixes the linking issue in our setup.

veblush (approved):

PR #32917, opened 2023-04-21
[build] Ignore possibly-leaked memory in PHP Valgrind tests

drfloob:
Valgrind will now only fail the build on definite leaks, not "possible" leaks. A trivial example that fails the PHP valgrind test as it is configured today:

```
namespace {
grpc_core::NoDestruct<grpc_core::BackOff> g_backoff{
    grpc_core::BackOff::Options()};
}  // namespace
```

Valgrind detects a possible leak because BackOff contains an absl::BitGen, which calls `new` through a chain of ownership indirection. This is what Valgrind calls an [interior pointer](https://valgrind.org/docs/manual/mc-manual.html#mc-manual.options:~:text=%22Possibly%20lost%22.%20This,have%20interior%2Dpointers.). Our CI will no longer fail them


drfloob:
Automerge is enabled.

stanley-cheung (approved):
LGTM

PR #32922, opened 2023-04-22
[PSM Interop] Temporary remedy for the issue with pod log dups

sergiitk:
While a proper fix is on the way, this mitigates the number of duplicated container logs in the xds test server/client pod logs.

The issue is that we only wait between stream restarts when an exception is caught, which isn't always the reason the stream gets broken. Another reason is the main container being shut down by k8s. In this situation, we essentially do

```py
while True:
  try:
    restart_stream()
    read_all_logs_from_pod_start()
  except Exception:
    logger.warning('error')
    wait_seconds(1)
```

This PR makes it

```py
while True:
  try:
    restart_stream()
    read_all_logs_from_pod_start()
  except Exception:
    logger.warning('error')
  finally:
    wait_seconds(5)
```


yashykt (approved):

sergiitk:
Adhoc test run:

* [grpc/java/v1.54.x/branch/psm-security](https://source.cloud.google.com/results/invocations/77363aae-2dd5-4dbb-88fc-64a8c79b06c9)

PR #32930, opened 2023-04-24
[build] Add Bazel user-defined build setting for `grpc_no_rls`.

msjarrett:
This can be used to disable RLS to decrease binary size, on platforms
that don't disable it automatically.


markdroth (approved):
This looks good to me!  I've triggered the tests, so let's see how they look.

I'm going to be OOO tomorrow through Friday, but @drfloob may be able to get this merged while I'm out, assuming the tests look good.

drfloob:
@msjarrett it appears our build dependency fixer script needs to be updated to support `bool_flag`. https://source.cloud.google.com/results/invocations/be814217-d13f-4e14-8d3a-36439ecafc49/targets/github%2Fgrpc%2Frun_tests%2Fsanity_linux_dbg_native%2Ftools%2Fbuildgen%2Fgenerate_projects.sh/tests

The relevant code is here. It parses the BUILD files as python, substituting custom python code for starlark code that wouldn't otherwise be defined in local scope (mostly with no-op implementations). Please let me know if you need some help. https://github.com/grpc/grpc/blob/da5dbc068d86762eb0ac27d946abc90bc0c5ac53/tools/distrib/fix_build_deps.py#L462-L487

drfloob:
The iOS test failures are unrelated, looks good. I'll alter the title to get a broken check to pass, then submit.

Issue #18075, opened 2019-02-15
Client-side Python fork tests are flaky on Python 3.7


ericgribkoff:
https://github.com/grpc/grpc/pull/16513 adds fork-without-exec tests for Python clients using `GRPC_ENABLE_FORK_SUPPORT=1` and the `multiprocessing` library. The tests are flaky on Kokoro using Python 3.7, leading to what appear to be hanging child processes.

I have been unable to reproduce the problem locally to attach a debugger and determine the exact problem, but based on a series of trials on Kokoro it looks the child's post-fork Python handler is deadlocking either when [obtaining the GIL](https://github.com/grpc/grpc/blob/04cdb1826666e0de67cfd11c1f5214e0fd53a018/src/python/grpcio/grpc/_cython/_cygrpc/fork_posix.pyx.pxi#L59) and or at some point *before* beginning to [clean-up inherited channels](https://github.com/grpc/grpc/blob/04cdb1826666e0de67cfd11c1f5214e0fd53a018/src/python/grpcio/grpc/_cython/_cygrpc/fork_posix.pyx.pxi#L66). This was determined via inserting calls to C++'s `abort()` or Python's `os._exit` at various points within the post-fork handler: aborting the process before obtaining the GIL deterministically avoided the deadlock, but even simply exiting the process with the GIL (before doing anything potentially suspicious, such as obtaining a condition variable's lock) could still result in a hung child process.

Since it appears that the deadlock is present on Python 3.7 and not on Python 2, and the hang does not appear to occur as a result of an error in our fork-handlers (such as trying to obtain a lock that was held by another thread pre-fork), I suspect that the issue is a result of our pre-fork handlers only *pausing* active gRPC Python threads in [`block_if_fork_in_progress`](https://github.com/grpc/grpc/blob/04cdb1826666e0de67cfd11c1f5214e0fd53a018/src/python/grpcio/grpc/_cython/_cygrpc/fork_posix.pyx.pxi#L113), where threads block on a condition variable until the fork completes. My guess is that these threads, since they are abruptly terminated at fork, are still holding some internal Python lock or otherwise leave the interpreter's state in a broken condition post-fork. I couldn't find a corresponding known issue on the Python issue tracker, nor have I confirmed this to be the case, but it seems the most likely cause at this point - it's also not clear that this would even indicate a bug in Python, as the documentation is unclear as to whether it's even intended to be safe to fork with living (but blocking) threads via the multiprocessing module.

[...]

PR #32935, opened 2023-04-25
[Python Fork] Use os.register_at_fork instead of pthread_atfork.

XuanWang-Amos:
Fix: https://github.com/grpc/grpc/issues/18075

From comments in https://github.com/grpc/grpc/issues/18075, `CPython` reinitialize the `GIL` after `pthread_atfork` child handler, thus we shouldn't use any `GIL` related functions in child handler which is what we're currently doing, this PR uses `os.register_at_fork` to replace `pthread_atfork` to prevent any undesired bevahior.

This also seems to fixes a thread hanging issue cased by changes in core: https://github.com/grpc/grpc/pull/32869

### Testing:
* Passed existing fork tests. (Note that due to some issues in `Bazel`, this change was not verified by `Bazel runs_per_test`).
* Tested by patch the core PR, was able to fix Python fork tests: https://github.com/grpc/grpc/pull/32933

### DIFF
--- a/src/python/grpcio/grpc/_cython/_cygrpc/fork_posix.pxd.pxi
+++ b/src/python/grpcio/grpc/_cython/_cygrpc/fork_posix.pxd.pxi
@@ -12,14 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-cdef extern from "pthread.h" nogil:
-    int pthread_atfork(
-        void (*prepare)() nogil,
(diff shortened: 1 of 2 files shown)


drfloob (approved):
LGTM, pending Python tests passing. I can confirm that this fix removes the fork-hang problem I was able to reproduce in my work. I'm comfortable landing this, but @gnossen please also sanity check this before the next release at the end of May, just to be safe.

PR #32938, opened 2023-04-25
[xDS] Implement cluster locking by ClusterSelectionFilter

eugeneo:
Part of the work needed for [gRFC A60](https://github.com/grpc/proposal/blob/master/A60-xds-stateful-session-affinity-weighted-clusters.md).


src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -154,6 +211,19 @@ class XdsResolver : public Resolver {
     if (xds_client_ != nullptr) xds_client_->ResetBackoff();
   }
 
+  void MaybeRemoveUnusedClusters();
+
+  void Run(std::function<void()> callback, DebugLocation location) {
+    work_serializer_->Run(std::move(callback), location);
+  }
+
+  ClusterState::ClusterStateMap::iterator Emplace(
```
markdroth, line 220: I don't think we should need this method.  It looks like there's some legacy complexity here related to how we're storing the cluster name in the map keys.  I've sent you #32989 to simplify this.  Once that change has been merged, I don't think we'll need this method anymore.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -99,6 +108,54 @@ UniqueTypeName XdsClusterAttribute::TypeName() {
...
+  void Orphan() override;
+
+  const std::string& cluster() const { return it_->first; }
+
+ private:
+  RefCountedPtr<XdsResolver> resolver_;
+  ClusterStateMap::iterator it_;
+};
+
+class XdsClusterMap : public RefCounted<XdsClusterMap> {
```
markdroth, line 138: Please add a comment here explaining what this class is for.  In particular, it's the cluster map being held by the `XdsConfigSelector`, not the one being held by the resolver.  Refs to this will also be held by individual calls between the time that the `XdsConfigSelector` is called and the time that the `ClusterSelectionFilter` runs.

markdroth (commented):
This looks like a good start!

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
```diff
@@ -21,16 +21,39 @@
...
+
+  explicit XdsClusterLbDataAttribute(RefCountedPtr<XdsClusterMap> cluster_map);
+
+  bool LockClusterConfig(absl::string_view cluster_name);
+
+  UniqueTypeName type() const override { return TypeName(); }
+
+ private:
+  RefCountedPtr<XdsClusterMap> cluster_map_;
+  RefCountedPtr<ClusterState> locked_cluster_config_;
```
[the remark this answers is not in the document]
markdroth: Note that I'm adding a method to set the `on_commit` callback in `ServiceConfigCallData` in #33002.

eugeneo (commented):
Thanks for the review!

Note that this has SetOnCommit from your commit that was reverted.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -803,20 +826,16 @@ XdsResolver::XdsConfigSelector::GetCallConfig(GetCallConfigArgs args) {
...
   hash_value[hash_string.size()] = '\0';
   call_config.call_attributes[RequestHashAttribute::TypeName()] =
       args.arena->New<RequestHashAttribute>(hash_value);
-  call_config.on_commit = [cluster_state = it->second->Ref()]() mutable {
-    cluster_state.reset();
-  };
-  auto lb_data =
-      args.arena->ManagedNew<XdsClusterLbDataAttribute>(cluster_map_);
-  call_config.call_attributes[XdsClusterLbDataAttribute::TypeName()] = lb_data;
+  auto lb_data = args.arena->ManagedNew<XdsClusterDataAttribute>(cluster_map_);
```
markdroth, line 836: Please call this variable `cluster_map`.  It actually has nothing to do with the LB policies.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -1177,50 +1196,37 @@ void RegisterXdsResolver(CoreConfiguration::Builder* builder) {
...
-      it_(resolver_->Emplace(cluster_name, WeakRef())) {}
+                           absl::string_view cluster_name)
+    : resolver_(std::move(resolver)), cluster_name_(cluster_name) {}
 
 void ClusterState::Orphan() {
-  auto* resolver = resolver_.get();
-  resolver->Run([resolver = std::move(
-                     resolver_)]() { resolver->MaybeRemoveUnusedClusters(); },
-                DEBUG_LOCATION);
+  resolver_->MaybeRemoveUnusedClustersInWorkSerializer();
```
markdroth, line 1211: I just noticed that this change means that we're no longer unreffing the resolver from inside the `WorkSerializer`, which I think we need to be doing.

I suggest changing `MaybeRemoveUnusedClustersInWorkSerializer()` to take a `RefCountedPtr<XdsResolver>` as an argument, which it will `std::move()` into the lambda, so that we're no longer holding a ref here in `ClusterState`.  This means that `MaybeRemoveUnusedClustersInWorkSerializer()` can also be a static method.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -505,58 +525,72 @@ XdsResolver::XdsConfigSelector::XdsConfigSelector(
...
             auto result = CreateMethodConfig(route_entry.route, nullptr);
             if (!result.ok()) {
-              *status = result.status();
-              return;
+              return result.status();
             }
             route_entry.method_config = std::move(*result);
-            MaybeAddCluster(
-                absl::StrCat("cluster:", cluster_name.cluster_name));
+            return std::vector<std::string>{std::string(
```
markdroth, line 537: `absl::StrCat()` returns `std::string`, so there's no need to construct a new `std::string` from its result.

Same thing throughout.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -505,58 +525,72 @@ XdsResolver::XdsConfigSelector::XdsConfigSelector(
         route_action->max_stream_duration =
             resolver_->current_listener_.http_max_stream_duration;
       }
-      Match(
+      absl::StatusOr<std::vector<std::string>> status_or_cluster_names = Match(
```
markdroth, line 527: Using a `std::vector<>` here causes unnecessary allocation and makes the code less readable.  I think we're better off sticking with the previous approach of adding each cluster as we identify it, inside the lambdas.

I suggest adding a lambda like this on line 510 above:

```
auto maybe_add_cluster = [&](absl::string_view cluster_name) {
  if (clusters.find(cluster_name) != clusters.end()) return;
  auto cluster_state = resolver_->GetOrCreateClusterState(cluster_name);
  absl::string_view name = cluster_state->cluster_name();
  clusters.emplace(name, std::move(cluster_state));
};
```

Then we don't need any changes to this `Match()` call except for changing `MaybeAddCluster` to `maybe_add_cluster`.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -1177,50 +1196,37 @@ void RegisterXdsResolver(CoreConfiguration::Builder* builder) {
...
-    return false;
+    return nullptr;
   }
-  auto cluster_info = cluster_map_->Find(cluster_name);
-  if (cluster_info.has_value()) {
-    locked_cluster_config_ = cluster_info->second;
-    cluster_map_.reset();
-    return true;
+  auto cluster = cluster_map_->Find(cluster_name);
+  if (cluster == nullptr) {
```
markdroth, line 1224: This should never happen.  But if it did, I think we still want to unref `cluster_map_`, since there's no point in holding on to refs to any cluster until this call completes.  So I think we don't need this block at all.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
```diff
@@ -31,23 +31,23 @@ namespace grpc_core {
 class XdsClusterMap;
 class ClusterState;
 
-class XdsClusterLbDataAttribute
+class XdsClusterDataAttribute
```
markdroth, line 34: Suggest calling this `XdsClusterMapAttribute`.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
```diff
@@ -31,23 +31,23 @@ namespace grpc_core {
...
   static UniqueTypeName TypeName() {
     static UniqueTypeName::Factory factory("xds_cluster_lb_data");
     return factory.Create();
   }
 
-  explicit XdsClusterLbDataAttribute(RefCountedPtr<XdsClusterMap> cluster_map);
+  explicit XdsClusterDataAttribute(RefCountedPtr<XdsClusterMap> cluster_map);
 
-  bool LockClusterConfig(absl::string_view cluster_name);
+  RefCountedPtr<ClusterState> LockAndGetClusterConfig(
```
markdroth, line 44: Please document that this method should be called exactly once, at which point we drop our ref to the cluster map, so this attribute will no longer contain anything useful.

markdroth (commented):
This is looking pretty good!

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
eugeneo: I believe this is a better code if we split it into `GetClusterState` and `ReleaseClusterMap` - it makes code easier to follow and more logical.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
eugeneo: Done, in the filter code.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
eugeneo: Done. Note that this method needs access to work_serializer_ so it can't be static (or the work_serializer_ needs to be made static.
I added releasing the resolver to the Orphan method.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -367,11 +370,26 @@ class XdsResolver : public Resolver {
...
+        auto* cluster = static_cast<XdsClusterAttribute*>(
+            service_config_call_data->GetCallAttribute(
+                XdsClusterAttribute::TypeName()));
+        if (cluster != nullptr) {
+          service_config_call_data->SetOnCommit(
+              [locked_cluster = cluster_data->LockAndGetClusterConfig(
+                   cluster->cluster())]() mutable { locked_cluster.reset(); });
+        }
+      }
+
```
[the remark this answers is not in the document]
eugeneo: Oops.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -211,17 +209,23 @@ class XdsResolver : public Resolver {
     if (xds_client_ != nullptr) xds_client_->ResetBackoff();
   }
 
-  void MaybeRemoveUnusedClusters();
-
-  void Run(std::function<void()> callback, DebugLocation location) {
-    work_serializer_->Run(std::move(callback), location);
+  RefCountedPtr<ClusterState> GetClusterState(absl::string_view cluster_name) {
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -124,33 +125,30 @@ class ClusterState : public DualRefCounted<ClusterState> {
...
-  const std::string& cluster() const { return it_->first; }
+  const std::string& cluster_name() const { return cluster_name_; }
 
  private:
   RefCountedPtr<XdsResolver> resolver_;
-  ClusterStateMap::iterator it_;
+  std::string cluster_name_;
 };
 
+// Cluster map held by the XdsConfigSelector. It is passed in the call arguments
```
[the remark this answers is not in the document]
eugeneo: Done.

eugeneo (commented):
Thank you for the review!

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
markdroth: It doesn't fix the problem to just reset `cluster_` in this function.  The `ClusterState` object may be unreffed from outside the `WorkSerializer`, which is why we're hopping into the `WorkSerializer` to begin with, so we can't unref the resolver here.

The actual goal here is to ensure that the resolver is destroyed inside the `WorkSerializer`, and any ref that we're holding to the resolver might be the last one, which means that we always need to release those refs inside the `WorkSerializer`.  Keep in mind that the lambda that we pass to the `WorkSerializer` may run in another thread, and it's entirely possible that that other thread will finish executing it (thus releasing the ref it's holding) before this thread releases its ref, so this could very well be the last ref to the resolver.

Static methods of a class *do* have access to the class's private members; they just need a pointer to an object of that class, because they aren't called on an object of the class (i.e., they have no `this`).  So what we need to do in `XdsResolver` is this:

```
  static void MaybeRemoveUnusedClustersInWorkSerializer(
      RefCountedPtr<XdsResolver> resolver) {
    auto* resolver_ptr = resolver.get();
    resolver_ptr->work_serializer_->Run(
        [resolver = std::move(resolver)]() {
          resolver->MaybeRemoveUnusedClusters();
        },
        DEBUG_LOCATION);
  }
```

Then this method becomes:

```
void ClusterState::Orphan() {
  XdsResolver::MaybeRemoveUnusedClustersInWorkSerializer(
      std::move(resolver_));
}
```

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
markdroth: We still don't need this `if` block.  If this block wasn't here, then we'd return null anyway on line 1228.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
markdroth: I think splitting this into two methods actually makes this API easier to misuse.  There is no case in which we want a caller to lookup a cluster without unreffing the map, so the API shouldn't allow that.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
```diff
@@ -21,11 +21,36 @@
...
 #include "src/core/lib/gprpp/unique_type_name.h"
 #include "src/core/lib/service_config/service_config_call_data.h"
 
 namespace grpc_core {
 
+// Forward declaration only, implementation is private
+class XdsClusterMap;
+class ClusterState;
+
+class XdsClusterMapAttribute
```
markdroth, line 34: It occurs to me that this class can actually be moved into the .cc file.  The only two pieces of code that ever access this are the `XdsConfigSelector` code that sets the attribute and the `ClusterSelectionFilter` code that uses the attribute, and they are both in the .cc file, so it doesn't need to be exported anywhere else.

markdroth (commented):
This is very close!

eugeneo (commented):
Thank you for the review. I believe moving classes to inner made change easier to follow.

eugeneo (commented):
Thank you for the review. I believe moving classes to inner made change easier to follow.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -296,6 +239,63 @@ class XdsResolver : public Resolver {
...
 
+  // An entry in the map of clusters that need to be present in the LB
+  // policy config.  The map holds a weak ref.  One strong ref is held by
+  // the ConfigSelector, and another is held by each call assigned to
+  // the cluster by the ConfigSelector.  The ref for each call is held
+  // until the call is committed.  When the strong refs go away, we hop
+  // back into the WorkSerializer to remove the entry from the map.
+  class ClusterState : public DualRefCounted<ClusterState> {
+   public:
+    using ClusterStateMap =
```
markdroth, line 250: This typedef is no longer needed as of #32989.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
markdroth: Looks like you missed this comment from my last pass.  Please recombine `GetClusterState()` and `ReleaseClusterMap()` into a single method, so that it's not possible for the caller to misuse the API by calling one without the other.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
eugeneo: Done

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.h
eugeneo: I did that. Should the null check be replaced with an assert?

eugeneo (commented):
Thank you for the review. Should the nullptr check in `LockAndGetCluster` be replaced with an assert?

markdroth (approved):

markdroth:
Returning null is fine.  The call will just fail in the LB policy, which is better than crashing the process.

PR #32940, opened 2023-04-26
Add RecordMetadataInformation function to CallTracer API

ananda1066:
Add new function to CallTracer API in order to track metadata sizes.

### DIFF
--- a/src/core/lib/channel/call_tracer.h
+++ b/src/core/lib/channel/call_tracer.h
@@ -54,6 +54,9 @@ class CallTracerAnnotationInterface {
   // TODO(yashykt): If needed, extend this to attach attributes with
   // annotations.
   virtual void RecordAnnotation(absl::string_view annotation) = 0;
+  // Records an annotation for metadata information. `metadata` should be in
+  // proto string format.
+  virtual void RecordMetadataInformation(absl::string_view metadata_info) = 0;
   virtual std::string TraceId() = 0;
   virtual std::string SpanId() = 0;
   virtual bool IsSampled() = 0;
--- a/src/cpp/ext/filters/census/client_filter.cc
+++ b/src/cpp/ext/filters/census/client_filter.cc
@@ -288,6 +288,11 @@ void OpenCensusCallTracer::OpenCensusCallAttemptTracer::RecordAnnotation(
   context_.AddSpanAnnotation(annotation, {});
 }
 
+void OpenCensusCallTracer::OpenCensusCallAttemptTracer::
+    RecordMetadataInformation(absl::string_view metadata_info) {
+  RecordAnnotation(metadata_info);
+}
+
 //
 // OpenCensusCallTracer
 //
--- a/src/cpp/ext/filters/census/client_filter.cc
+++ b/src/cpp/ext/filters/census/client_filter.cc
@@ -288,6 +288,11 @@ void OpenCensusCallTracer::OpenCensusCallAttemptTracer::RecordAnnotation(
   context_.AddSpanAnnotation(annotation, {});
 }
 
+void OpenCensusCallTracer::OpenCensusCallAttemptTracer::
+    RecordMetadataInformation(absl::string_view metadata_info) {
+  RecordAnnotation(metadata_info);
+}
+
 //
 // OpenCensusCallTracer
 //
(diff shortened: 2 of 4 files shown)


src/core/lib/channel/call_tracer.h
```diff
@@ -54,6 +54,9 @@ class CallTracerAnnotationInterface {
   // TODO(yashykt): If needed, extend this to attach attributes with
   // annotations.
   virtual void RecordAnnotation(absl::string_view annotation) = 0;
+  // Records an annotation for metadata information. `metadata` should be in
+  // proto string format.
+  virtual void RecordMetadataInformation(absl::string_view metadata_info) = 0;
```
markdroth, line 59: Why do we need a new method here?  Can we get this info via the existing methods to intercept metadata?

ananda1066: We want to serialize the string into a proto in google3 for easier parsing to collect the actual metadata sizes, but we can't take a proto dependency here in OSS, so this new function will get overridden in google3 with the serialization.  I haven't added anything besides the API change right now because we don't want to touch metadata code at the moment.

@yashykt @ctiller have some context here.

markdroth: Why can't you deal with that proto in the `CallTracer` implementation itself, rather than adding a new method to the API?

ctiller: I just spied those methods (hadn't noticed/remembered them previously) - I think they'll do.

@ananda1066 we can probably just keep the existing methods in place:
```
  virtual void RecordSendInitialMetadata(
      grpc_metadata_batch* send_initial_metadata) = 0;
  virtual void RecordSendTrailingMetadata(
      grpc_metadata_batch* send_trailing_metadata) = 0;
```

and plumb our own logic internally to recalculate the encoded sizes and publish them out to the sink we were talking about

ananda1066: @ctiller I think we would just want to extend the `RecordAnnotation` in the google3 call tracer for this purpose, so that we can get metadata sizes in OSS as well (the info could be useful). 

I think the main thing is knowing that it is a metadata annotation and should be serialized into the specific proto type - I guess we could just do a simple StartsWith.

@yashykt we could add something, but would that actually be useful in OSS? Otherwise, I think Mark's suggestion is good; we can then abandon this PR and just make the changes in google3 CallTracer's `RecordAnnotation` function.

markdroth: Note that if we require the caller of `CallTracer` to construct and serialize the proto, then we still have the problem of how we call that API differently internally than in OSS, since all of the code that calls `CallTracer` is shared between the two.  The only way to do this such that it happens internally only is to do it inside the `CallTracer` impl, since that's the only thing that differs between OSS and internal.

ananda1066: Yup, the prod CallTracer implementation will serialize the proto itself. The caller just passes in a string that is in the correct format for serialization, but this same format works well in OSS directly (it's more readable).

markdroth: I don't understand what you mean by "a string that is in the correct format for serialization".  The string form *is* the serialized form -- if you have that already, then you've already done the serialization, which is exactly what we don't want to do in OSS.

There are two forms of a protobuf:
1. The serialized form, which is a string of bytes that contains the proto as it would be stored or sent on the wire.
2. The non-serialized form, which is in the form of a C++ object whose code was generated by the protobuf codegen.  This object has methods for setting and getting each of its fields and typically stores each field in a separate data member.

When constructing a new proto, you typically first build it in the non-serialized form, and then you use the protobuf API to serialize it into the serialized form, in which it can be stored or put on the wire.  (Conversely, when parsing a serialized proto from storage or from the wire, you deserialize it from the string form into the non-serialized form, which allows you to easily access each field.)  The problem here is that that transformations requires using the protobuf library, and we can't depend on that in OSS.  So if you want that code to be specific to our internal implementation, then you need to do that inside the `CallTracer` implementation.

That means you need to pass the non-serialized data into the `CallTracer` instead of a serialized protobuf.  And I think the existing methods here are already doing that.  I don't think we need a new method here.

ananda1066: I see, I'm using the term serialized incorrectly; it's a proto-formatted string I guess. I agree we don't need a new function.

This is how I've currently implemented it: The caller passes in a raw string to `RecordAnnotation`. The OSS call tracer simply records the raw string. The google3 call tracer implementation does something like `absl::Substitute(R"pb($0)pb", annotation)` and then `proto2::TextFormat::ParseFromString(annotation, &metadata_proto)` to create the proto.

There may be a better way to do this, I'm not very familiar with proto parsing.

markdroth: Looks like you're talking about the textproto form.  That form is not stable and not appropriate for use in an API; it's intended only to be a human-readable form for debugging purposes only.  And parsing it on a per-RPC basis seems likely to be inefficient from a performance perspective.

I don't think this is the approach we want.  I think we just want to construct the proto inside of the `CallTracer` implementation using data passed into the existing methods.

I don't think we need this PR.  The existing `CallTracer` methods should already provide what we need here.

ananda1066: Ok, I will close this and try to use the existing methods.

PR #32944, opened 2023-04-26
[Audit Logging] Authz policy support for audit logging

rockspore:
Add audit condition and audit logger config into `grpc_core::Rbac`. Support translation of audit logging options from authz policy to it.

Audit logging options in authz policy looks like:
```json
{
  "audit_logging_options": {
    "audit_condition": "ON_DENY",
    "audit_loggers": [
      {
        "name": "logger",
        "config": {},
        "is_optional": false
      }
    ]
  }
}
```
which is consistent with what's in the xDS RBAC proto but a little flattened.


src/core/lib/security/authorization/rbac_translator.cc
```diff
@@ -328,24 +335,140 @@ absl::StatusOr<std::map<std::string, Rbac::Policy>> ParseRulesArray(
...
+  } else {
+    return absl::InvalidArgumentError(
+        absl::StrFormat("\"audit_loggers[%d].config\" is required.", pos));
+  }
+  if (!AuditLoggerRegistry::FactoryExists(name)) {
+    if (is_optional) {
+      return nullptr;
+    }
+    return absl::InvalidArgumentError(absl::StrFormat(
+        "\"audit_loggers[%d].name\" %s is not supported.", pos, name));
```
gtcooke94, line 404: `is not supported` vs. `is not in the registry` or something like
".... is not supported. Ensure a factory for it is in the AuditLoggerRegistry"

src/core/lib/security/authorization/rbac_translator.cc
```diff
@@ -328,24 +335,140 @@ absl::StatusOr<std::map<std::string, Rbac::Policy>> ParseRulesArray(
...
+  }
+  auto result = AuditLoggerRegistry::ParseConfig(name, config);
+  if (!result.ok()) {
+    return absl::InvalidArgumentError(absl::StrFormat(
+        "\"audit_loggers[%d]\" %s", pos, result.status().message()));
```
gtcooke94, line 409: Maybe add something like this so it's clear the failing bit is from `AuditLoggerRegistry::ParseConfig`
"AuditLoggerRegistry::ParseConfig failed for audit_loggers[index] , <status message>"

test/core/security/rbac_translator_test.cc
```diff
@@ -53,9 +68,39 @@ MATCHER_P4(EqualsHeader, expected_name, expected_matcher_type,
...
+class TestAuditLoggerFactory : public AuditLoggerFactory {
+ public:
+  class TestAuditLoggerConfig : public AuditLoggerFactory::Config {
+    absl::string_view name() const override { return kLoggerName; }
+    std::string ToString() const override { return ""; }
+  };
+  absl::string_view name() const override { return kLoggerName; }
+  absl::StatusOr<std::unique_ptr<AuditLoggerFactory::Config>>
+  ParseAuditLoggerConfig(const Json& json) override {
+    if (json.object().find("bad") != json.object().end()) {
```
gtcooke94, line 80: I think it's worth a comment here that "bad" comes from hard-coding the json like that for a test

test/core/security/rbac_translator_test.cc
```diff
@@ -53,9 +68,39 @@ MATCHER_P4(EqualsHeader, expected_name, expected_matcher_type,
...
+      return absl::InvalidArgumentError("bad logger config.");
+    }
+    return std::make_unique<TestAuditLoggerConfig>();
+  }
+  std::unique_ptr<AuditLogger> CreateAuditLogger(
+      std::unique_ptr<AuditLoggerFactory::Config>) override {
+    Crash("unreachable");
```
gtcooke94, line 87: Can you add a comment explaining this, it's not immediately clear to me why `CreateAuditLogger` should Crash?

test/core/security/rbac_translator_test.cc
```diff
@@ -984,6 +1038,347 @@ TEST(GenerateRbacPoliciesTest, UnknownFieldInHeaders) {
...
+      "      }"
+      "    ]"
+      "  }"
+      "}";
+  auto rbacs = GenerateRbacPolicies(authz_policy);
+  EXPECT_EQ(rbacs.status().code(), absl::StatusCode::kInvalidArgument);
+  EXPECT_THAT(rbacs.status().message(),
+              "\"audit_loggers[0].name\" unknown_logger is not supported.");
+}
+
```
gtcooke94, line 1381: Can you add a test where you parse a non-empty `config`?
The `{"bad": true}` case parses it, but having a test where we pass a valid custom config and ensure that the values are accessible is worth it IMO

gtcooke94, line 1381: Can you add a test that exercises the path for `NONE` as the audit condition?

gtcooke94, line 1381: Can you add a test where there are multiple audit loggers in the array?

gtcooke94 (commented):
Looks good! A few comments/questions

src/core/lib/security/authorization/rbac_policy.h
```diff
@@ -38,6 +41,13 @@ struct Rbac {
     kDeny,
   };
 
+  enum class AuditCondition {
```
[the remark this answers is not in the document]
rockspore: I think in this directory, nothing should touch the xDS definitions. The Rbac action is also defined as enum

src/core/lib/security/authorization/rbac_translator.cc
```diff
@@ -328,24 +335,140 @@ absl::StatusOr<std::map<std::string, Rbac::Policy>> ParseRulesArray(
...
+      deny_condition = Rbac::AuditCondition::kOnDeny;
+      allow_condition = Rbac::AuditCondition::kOnDenyAndAllow;
+    } else {
+      return absl::InvalidArgumentError(absl::StrFormat(
+          "Unsupported \"audit_condition\" value %s.", condition));
+    }
+    if (rbacs.deny_policy != absl::nullopt) {
+      rbacs.deny_policy->audit_condition = deny_condition;
+    }
+    rbacs.allow_policy.audit_condition = allow_condition;
```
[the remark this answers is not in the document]
rockspore: Yes the authz policy will be rejected if no allow policy is present, whereas deny policy is designed to be optional

src/core/lib/security/authorization/rbac_translator.cc
```diff
@@ -328,24 +335,140 @@ absl::StatusOr<std::map<std::string, Rbac::Policy>> ParseRulesArray(
...
+      }
+      // Check the value since the unsupported logger config can also
+      // return ok when marked as optional.
+      if (result.value() != nullptr) {
+        // Only move the logger config over if audit condition is not NONE.
+        if (rbacs.allow_policy.audit_condition != Rbac::AuditCondition::kNone) {
+          rbacs.allow_policy.logger_configs.push_back(
+              std::move(result.value()));
+        }
+        if (rbacs.deny_policy != absl::nullopt &&
```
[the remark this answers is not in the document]
rockspore: Answered above.

rockspore (commented):
Thanks for the review! While I am working on the actionable comments, just wanted to answer your questions first. @gtcooke94

src/core/lib/security/authorization/rbac_translator.cc
rockspore: We do not expose `AuditLoggerRegistry` class as a public API. There is only a namespace function `RegisterAuditLoggerFactory`. So I don't think we want to reveal the implementation details in the error message. I changed to `... is not supported natively or registered`. Let me know what you think.

src/core/lib/security/authorization/rbac_translator.cc
rockspore: Similar to the reasoning above, `AuditLoggerRegistry` is not something people should know, they should only know which logger config has an error and it's the factory parser's job to return actionable errors, whether built-in or custom ones.

src/core/lib/security/authorization/rbac_translator.cc
```diff
@@ -328,24 +335,140 @@ absl::StatusOr<std::map<std::string, Rbac::Policy>> ParseRulesArray(
...
+      rbacs.deny_policy->audit_condition = deny_condition;
+    }
+    rbacs.allow_policy.audit_condition = allow_condition;
+  }
+  it = json.object().find("audit_loggers");
+  if (it != json.object().end()) {
+    if (it->second.type() != Json::Type::kArray) {
+      return absl::InvalidArgumentError("\"audit_loggers\" is not an array.");
+    }
+    for (size_t i = 0; i < it->second.array().size(); ++i) {
```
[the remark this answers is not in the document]
rockspore: Done. It's not expensive at all since array() returns a const reference.

src/core/lib/security/authorization/rbac_translator.cc
```diff
@@ -398,11 +521,25 @@ absl::StatusOr<RbacPolicies> GenerateRbacPolicies(
       }
       rbacs.allow_policy = std::move(*allow_policy_or);
       has_allow_rbac = true;
+    } else if (object.first == "audit_logging_options") {
+      // We need to process this till policies are all parsed.
```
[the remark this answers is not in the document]
rockspore: Thanks. I don't know what I was thinking.

test/core/security/rbac_translator_test.cc
rockspore: Added the comment.

test/core/security/rbac_translator_test.cc
rockspore: `rbac_translator.cc` shouldn't call this method. So having it crash in this test confirms this fact.

test/core/security/rbac_translator_test.cc
rockspore: Done.

rockspore: This bad config test verifies that `{"bad":  true}` is passed into the parser appropriately, since `bad` is found by it. It also confirms that the translator can propagate the error from the parser properly. IMO, this should suffice as a unit test for the rbac translator.

Parsing valid config and returning a good logger is something the factory/registry unit test should do. I think we need to do that for our built-in loggers.

rockspore: I converted the single logger test to two loggers. I think a case with multiple loggers covers single-logger case.

src/core/lib/security/authorization/rbac_translator.cc
gtcooke94: Makes sense, the implementations of `ParseConfig` should be returning helpful errors that we pass up

test/core/security/rbac_translator_test.cc
gtcooke94: Ah so it's just a method that has to be there, but since it shouldn't be called we crash to ensure it isn't called

test/core/security/rbac_translator_test.cc
gtcooke94: What I was thinking is that the tests here just all ensure that the config is parsable, but this doesn't check that it correctly plumbs along those parsed values.

For example if there were a mistake in the code that made an empty config always get put on the returned object vs. this parsed config, would we have a test that caught that?

src/core/lib/security/authorization/rbac_policy.cc
```diff
@@ -29,22 +29,46 @@ namespace grpc_core {
...
+    case AuditCondition::kOnAllow:
+      condition_str = "OnAllow";
+      break;
+    case AuditCondition::kOnDenyAndAllow:
+      condition_str = "OnDenyAndAllow";
+      break;
+  }
   contents.push_back(absl::StrFormat(
-      "Rbac action=%s{", action == Rbac::Action::kAllow ? "Allow" : "Deny"));
+      "Rbac name=%s action=%s audit_condition=%s{", name,
```
erm-g, line 70: Would it make sense to change the order of `name` and `action` here to mimic the previous structure?

test/core/security/rbac_translator_test.cc
```diff
@@ -984,6 +1040,371 @@ TEST(GenerateRbacPoliciesTest, UnknownFieldInHeaders) {
...
+      "  \"name\": \"authz\","
+      "  \"allow_rules\": ["
+      "    {"
+      "      \"name\": \"allow_policy\""
+      "    }"
+      "  ],"
+      "  \"audit_logging_options\": {"
+      "    \"audit_loggers\": ["
+      "      {"
+      "        \"name\": \"test_logger\""
```
erm-g, line 1372: Is that a correct behavior in this case? Can we treat it as an empty config or does my idea violate some best practices?

src/core/lib/security/authorization/rbac_policy.h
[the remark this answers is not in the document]
rockspore: Thanks. Changed to 1.26

test/core/security/rbac_translator_test.cc
```diff
@@ -984,6 +1040,371 @@ TEST(GenerateRbacPoliciesTest, UnknownFieldInHeaders) {
...
+      "      \"name\": \"allow_policy\""
+      "    }"
+      "  ],"
+      "  \"deny_rules\": ["
+      "    {"
+      "      \"name\": \"deny_policy\""
+      "    }"
+      "  ],"
+      "  \"audit_logging_options\": {"
+      "    \"audit_condition\": \"ON_DENY\""
```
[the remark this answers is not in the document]
rockspore: Absence of `audit_loggers` is fine. It simply won't have any loggers to invoke. What's invalid is if users have the field `audit_loggers` but with wrong format or invalid fields in it.

test/core/security/rbac_translator_test.cc
rockspore: Done.

rockspore: I think I see your point now. I made the config store the json string and return it in `ToString()`. The two-logger test case is verifying that now.

erm-g (changes requested):
I really like the test coverage but have few more suggesions:

1. add tests for rbac_policy.cc , especially for Rbac::ToString() - we added extra fields there and I think it's worth testing the resulting strings
2. for rbac_translator_test.cc - let's add a test for an empty audit_logger_options (`{}`) and for the situation when it's a customer's logger and we don't know if the config has a correct value or not (like `"config":{"bad_value_but_we_can't_check_it":"ok"}`
3. are we planning to create e2e tests where we can check the real output later?

test/core/security/rbac_translator_test.cc
rockspore: That's a good point. It got to me rethink how this maps to the xDS config. I think `config` should be allowed to be absent. Modified the behavior and test.

rockspore:
> I really like the test coverage but have few more suggesions:
> 
> 1. add tests for rbac_policy.cc , especially for Rbac::ToString() - we added extra fields there and I think it's worth testing the resulting strings
> 2. for rbac_translator_test.cc - let's add a test for an empty audit_logger_options (`{}`) and for the situation when it's a customer's logger and we don't know if the config has a correct value or not (like `"config":{"bad_value_but_we_can't_check_it":"ok"}`
> 3. are we planning to create e2e tests where we can check the real output later?

1. In one of my tests, I added a prefix checking of the `ToString()`. I don't want to add coverage for the existing behavior of `ToString()`, which is not directly called anywhere, and possibly only used for logging purpose somewhere (if any). I suppose there was a reason why tests didn't exist in the first place for it.
2. Added a test for empty logging options. Also added coverage by checking `Config.ToString()` which I have modified to store the config json.
3. e2e test is definitely required. We will do that once the last PR to complete the feature is done.

src/core/lib/security/authorization/rbac_policy.cc
rockspore: I feel name should always be the first thing if available. And the resulting string is gonna be different from before anyway, i.e. we are are falling back to existing behavior if name is absent or so.

gtcooke94 (approved):
This LGTM

src/core/lib/security/authorization/rbac_policy.h
```diff
@@ -162,15 +172,22 @@ struct Rbac {
   };
 
   Rbac() = default;
-  Rbac(Rbac::Action action, std::map<std::string, Policy> policies);
+  Rbac(Rbac::Action action, std::map<std::string, Policy> policies,
+       absl::string_view name);
```
markdroth, line 176: Please make `name` the first parameter, since it seems a little more logical that way.

src/core/lib/security/authorization/rbac_policy.h
```diff
@@ -162,15 +172,22 @@ struct Rbac {
...
   std::string ToString() const;
 
   Action action;
+  AuditCondition audit_condition;
+
   std::map<std::string, Policy> policies;
+  std::vector<std::unique_ptr<experimental::AuditLoggerFactory::Config>>
+      logger_configs;
+  // The authorization policy name or the HTTP RBAC filter name.
+  std::string name;
```
markdroth, line 190: Similarly, let's make `name` the first data member.

src/core/lib/security/authorization/rbac_policy.h
```diff
@@ -162,15 +172,22 @@ struct Rbac {
...
+  Rbac(Rbac::Action action, std::map<std::string, Policy> policies,
+       absl::string_view name);
 
   Rbac(Rbac&& other) noexcept;
   Rbac& operator=(Rbac&& other) noexcept;
 
   std::string ToString() const;
 
   Action action;
+  AuditCondition audit_condition;
```
markdroth, line 184: Please move this down next to the `logger_configs` field, since they're closely related.

src/core/lib/security/authorization/rbac_policy.cc
```diff
@@ -29,22 +29,46 @@ namespace grpc_core {
...
+  audit_condition = other.audit_condition;
   policies = std::move(other.policies);
+  logger_configs = std::move(other.logger_configs);
+  name = std::move(other.name);
   return *this;
 }
 
 std::string Rbac::ToString() const {
   std::vector<std::string> contents;
+  std::string condition_str;
```
markdroth, line 54: This can be `absl::string_view` to avoid an unnecessary allocation.

markdroth (commented):
This looks good overall!  Comments are mostly minor.

src/core/lib/security/authorization/rbac_policy.cc
```diff
@@ -29,29 +29,29 @@ namespace grpc_core {
 // Rbac
 //
 
-Rbac::Rbac(Rbac::Action action, std::map<std::string, Policy> policies,
-           absl::string_view name)
-    : action(action), policies(std::move(policies)), name(name) {}
+Rbac::Rbac(absl::string_view name, Rbac::Action action,
```
markdroth, line 32: Might as well take the name parameter as `std::string` instead of `absl::string_view`, since we're making a copy anyway.  That way, if a caller happens to have a temp string that they don't need to keep, they can pass it in with `std::move()` and avoid the copy.

src/core/lib/security/authorization/rbac_translator.cc
```diff
@@ -394,7 +392,8 @@ ParseAuditLogger(const Json& json, size_t pos) {
         absl::StrFormat("\"audit_loggers[%d].name\" is not a string.", pos));
   }
   absl::string_view name = it->second.string();
-  Json config = Json::Object();
+  // The config defaults to an empty object.
+  Json config = Json::FromObject(Json::Object());
```
markdroth, line 396: Can just say `Json::FromObject({})`.  The compiler is smart enough to know the type of the `{}`.

markdroth (approved):
Looks great!  Feel free to merge after addressing the remaining comments.

erm-g (approved):

src/core/lib/security/authorization/rbac_translator.cc
rockspore: Changed.

src/core/lib/security/authorization/rbac_policy.cc
rockspore: Done. Thanks for the tip!

PR #32954, opened 2023-04-27
[xDS] Accept cpu_utilization over 100%

yousukseung:


include/grpcpp/ext/server_metric_recorder.h
```diff
@@ -43,9 +43,10 @@ class ServerMetricRecorder {
...
   // Factory method. Use this to create.
   static std::unique_ptr<ServerMetricRecorder> Create();
-  /// Records the server CPU utilization in the range [0, 1].
-  /// Values outside of the valid range are rejected.
-  /// Overrides the stored value when called again with a valid value.
+  /// Records the server CPU utilization in the range [0, infy).
+  /// Values may be larger than 1.0 when the usage exceeds the reporter
+  /// dependent notion of soft limits. Values outside of the valid range are
+  /// rejected. Overrides the stored value when called again with a valid value.
   void SetCpuUtilization(double value);
```
markdroth, line 50: Please also update the comment on `CallMetricRecorder::RecordCpuUtilizationMetric()`:

https://github.com/grpc/grpc/blob/1267000bbb690353c7d5e05b32adcffdbbeed16c/include/grpcpp/ext/call_metric_recorder.h#L43

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -282,6 +282,7 @@ class FakeResolverResponseGeneratorWrapper {
       response_generator_;
 };
 
+using xds::data::orca::v3::OrcaLoadReport;
```
markdroth, line 285: Please move this up to line 87.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -2752,6 +2746,73 @@ TEST_F(ClientLbInterceptTrailingMetadataTest, BackendMetricDataMerge) {
...
+    ASSERT_TRUE(actual.has_value());
+    CheckLoadReportAsExpected(*actual, GetParam().received);
+  }
+  // Check LB policy name for the channel.
+  EXPECT_EQ("intercept_trailing_metadata_lb",
+            channel->GetLoadBalancingPolicyName());
+  EXPECT_EQ(kNumRpcs, num_trailers_intercepted());
+}
+
+INSTANTIATE_TEST_SUITE_P(
```
markdroth, line 2779: I think using a parameterized test is overkill here.  It makes the code harder to understand, and it generates test names that do not actually help a human understand the purpose of the test.

Instead, please just add a helper method to the existing `ClientLbInterceptTrailingMetadataTest` framework to run a test with specified input and expected output, like this:

```
static void RunPerRpcMetricReportingTest(
    const OrcaLoadReport& reported,
    const OrcaLoadReport& expected) {
  // ...run test...
}
```

Then write a separate `TEST_F()` block for each test case you want to test, with an appropriate test name, which calls the helper function with appropriate values.  For example:

```
TEST_F(ClientLbInterceptTrailingMetadataTest, IgnoresOutOfRangeValues) {
  RunPerRpcMetricReportingTest(...);
}

TEST_F(ClientLbInterceptTrailingMetadataTest, AllowsCpuUtilizationGreaterThan1) {
  RunPerRpcMetricReportingTest(...);
}
```

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -2345,30 +2345,66 @@ TEST_F(ClientLbPickArgsTest, Basic) {
...
+  }
+  OrcaLoadReportBuilder& SetUtilization(absl::string_view n, double v) {
+    (*report_.mutable_utilization())[n] = v;
+    return *this;
+  }
+  OrcaLoadReportBuilder& SetNamedMetrics(absl::string_view n, double v) {
+    (*report_.mutable_named_metrics())[n] = v;
+    return *this;
+  }
+  OrcaLoadReport Get() { return std::move(report_); }
```
markdroth, line 2381: Please call this method `Build()`.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -3140,20 +3200,19 @@ TEST_F(WeightedRoundRobinTest, CallAndServerMetric) {
...
   request.mutable_param()->mutable_backend_metrics()->set_eps(
       std::numeric_limits<double>::min());
   request.mutable_param()->mutable_backend_metrics()->set_rps_fractional(100);
   ExpectWeightedRoundRobinPicks(DEBUG_LOCATION, stub,
-                                /*expected_weights=*/{3, 6, 4},
+                                /*expected_weights=*/{15, 10, 2},
                                 /*total_passes=*/3, &request);
   // Now send requests without per-call reported QPS.
-  // This should change WRR picks back to 1:2:4.
+  // This should change WRR picks back to 2:1:8.
```
markdroth, line 3213: This comment doesn't match the numbers below.

yousukseung: Fixed.

test/cpp/end2end/client_lb_end2end_test.cc
yousukseung: Done.

test/cpp/end2end/client_lb_end2end_test.cc
yousukseung: Done.

include/grpcpp/ext/server_metric_recorder.h
yousukseung: Done.

test/cpp/end2end/client_lb_end2end_test.cc
markdroth: Please refactor as I suggested.  I think the parameterized test interface here is much harder to read, and we simply don't need the complexity in this case.

yousukseung: Updated.

markdroth (approved):
Looks great!

PR #32956, opened 2023-04-27
[fuzzing] Client channel resolver fuzzer

drfloob:


src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
```diff
@@ -495,6 +495,11 @@ absl::optional<Resolver::Result> EventEngineClientChannelDNSResolver::
     absl::Status status = errors_.status(
         absl::StatusCode::kUnavailable,
         absl::StrCat("errors resolving ", resolver_->name_to_resolve()));
+    if (status.ok()) {
```
drfloob, line 498: This is a fix for a crash identified with the fuzzer!

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
```diff
@@ -0,0 +1,68 @@
...
+enum TXTRecordType {
+    TXT_VALID = 0;
+    TXT_INVALID = 1;
+}
+message TXTRecords {
+    repeated TXTRecordType txt_records = 1;
+}
+
+message Msg {
+    ResolverArgs resolver_args = 1;
```
markdroth, line 54: Maybe just make this use a generic set of channel args here, so that the fuzzer can cover any additional channel args we might add to the resolver in the future?

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
```diff
@@ -0,0 +1,68 @@
...
+    bool request_service_config = 2;
+    bool enable_srv_queries = 3;
+}
+
+// Some error. The details are unimportant.
+message Error {}
+
+
+message ResolvedAddress {
+    uint32 port = 1;
```
markdroth, line 29: Why only the port and not the address?

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
```diff
@@ -0,0 +1,68 @@
...
+message SRVRecords {
+    repeated SRVRecord srv_records = 1;
+}
+
+enum TXTRecordType {
+    TXT_VALID = 0;
+    TXT_INVALID = 1;
+}
+message TXTRecords {
+    repeated TXTRecordType txt_records = 1;
```
markdroth, line 50: Shouldn't this include the actual text of the TXT record to return?

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
```diff
@@ -0,0 +1,68 @@
...
+}
+
+message Msg {
```
markdroth, line 53: Might be useful to also support re-resolution, and to test the backoff in case of a resolver failure.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
```diff
@@ -0,0 +1,68 @@
...
+    }
+    oneof srv {
+        SRVRecords srv_response = 4;
+        Error srv_error = 5;
+    }
+    oneof txt {
+        TXTRecords txt_response = 6;
+        Error txt_error = 7;
+    }
+    // TODO(hork): consider if cancellation responses can be fuzzed
```
markdroth, line 67: This would definitely be useful.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
```diff
@@ -0,0 +1,265 @@
...
+
+  // responses
+  absl::optional<std::vector<EventEngine::ResolvedAddress>> hostname_responses_;
+  absl::optional<std::vector<EventEngine::DNSResolver::SRVRecord>>
+      srv_responses_;
+  absl::optional<std::vector<std::string>> txt_responses_;
+
+  // base cases
+  const std::string txt_valid_config_ =
+      "grpc_config=[{\"serviceConfig\":{\"loadBalancingPolicy\":\"round_"
```
markdroth, line 192: I think this kind of input should be generated by the fuzzer, so that it can test that we correctly handle various types of malformed input.

markdroth (commented):
This looks like a great start!  I think we want the proto format to be a bit more generic though.

We don't necessarily have to have everything that I suggested in this PR, but I think we should consider how we would structure the proto when we add these things later, so that we can avoid breaking existing test cases when we add new functionality later.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
ctiller: If that happens... consider unifying API fuzzers channel args generation with this fuzzers channel args generation: there are subtleties in allowing free range generation of channel args that we've run into before.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
ctiller: Having a "this is good/this is bad" example for the fuzzer is still useful though: it allows additional checks to be made if the fuzzer understands that this must definitely pass or fail.

I'd probably encourage instead moving the proto to something like:

```
oneof {
  TXTRecordType enumerated_value = 1;
  string arbitrary_value = 2;
}
```

with a weaker check on `arbitrary_value` than is possible with `enumerated_value`

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
drfloob: I'm not sure the content of the address matters. I could have used the same ResolvedAddress instance for all of them, but figured I'd trivially add some variation.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
drfloob: This fuzzer reuses 2 cases which are hardcoded as base cases of the fuzzer: one TXT record that is parseable as a valid gRPC config, and the other "invalid" gRPC config. We could add more cases or have more content randomized, but I definitely wanted the fuzzer to test the happy path as well, and randomly generating a correct config is more complicated. If we wanted to fuzz the parser (maybe separately), we could.

There were problems early on where valid and invalid TXT responses would not mix, for example. This fuzzer exercises that.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
drfloob: Are you suggesting that `arbitrary_value` here would be the TXT record content if the enumerated value was "invalid"? If so, cool. If not, I may do that anyway.

drfloob: Oh, I missed the `oneof` ... that must be what you meant. SG.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
markdroth: I'm less interested in the service config parsing piece, since I think we can fuzz that separately (although I don't know if we are actually doing so -- we should look at that).  I'm more interested in testing the code in the client channel resolver that chooses which TXT entry to use, as per the changes we made to the EE TXT lookup API in #33030.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
```diff
@@ -0,0 +1,265 @@
...
+      "robin\",\"methodConfig\":[{\"name\":[{\"method\":\"Foo\",\"service\":"
+      "\"SimpleService\"}],\"waitForReady\":true}]}}]";
+  const std::string txt_invalid_config_ = "grpc_config=[{\"foops\":{}}]";
+  const absl::Status lookup_hostname_response_base_case_ =
```
yijiem, line 196: Random thought: would it be useful if we fuzz the error status as well? Maybe something like:

Proto:
```
message Error {
  uint16 code = 1;
  string message = 2;
}
```

And then in the code we can probably create the absl::Status with something like (not sure if the casting would work):
```
// clamp error.code() in (0, 16]
status = absl::Status(static_cast<absl::StatusCode>(error.code() % 16 + 1), error.message());
```

A little bit hacky though...

drfloob: That's a good point. There is no contract for a subset of valid statuses from the EventEngine DNS resolver APIs, so this could be valuable.

ctiller: Maybe use the rpc status proto here?

ctiller: (getting error codes outside range is actually valuable as defense against future changes)

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
drfloob: This PR now fuzzes the service config as well, so the parser is going to get some exercise too.

> I'm more interested in testing the code in the client channel resolver that chooses which TXT entry to use, as per the changes we made to the EE TXT lookup API in https://github.com/grpc/grpc/pull/33030.

That was already being fuzzed ... it was exercised without respect to whether it was doing "the right thing", but the fuzzer generated a series of valid & invalid records, and the resolver had to find one with a config prefix.

In any case, I've made this more robust, including options for non-config TXT records, non-parseable `grpc_config=` TXT records, and fuzzer-generated ServiceConfig JSON.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
drfloob: The rpc status proto would still need to be converted to an absl::Status for use in the resolver API, so I don't see how that's much of a win.

drfloob: It's implemented for now per Yijie's recommendation.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
drfloob: I think that's doable. IIUC, that amounts to calling the result health callback with bad statuses some number of times (capped at maybe 3, since that's a ~7s test already) before succeeding. We'd see timeouts if re-resolution failed.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_ops.proto
drfloob: It's a bit messy, but I think I see a path here. We'd want the fuzzer to decide when to inject a `Resolver::Orphan` call, and add real tracking of task handles.

```
message OrphanOp {
    oneof when {
        bool after_hostname_lookup = 4;
        bool after_srv_lookup = 5;
        bool after_txt_lookup = 6;
    }
}
```

drfloob:
@ctiller @markdroth In the interest of getting _some_ additional fuzzing landed early this week while we're focused on it, I'd like to propose iterative improvements, adding channel args generation/unification, backoff-retries, and Cancellation in subsequent PRs. If the fixit is quiet, I'll work on these this week.

I am definitely interested in thoughts regarding forward-compatibility of the proto. I made the contents of the `TXTRecordType` enum more specific to allow for variations we may think of later.

ctiller:
Yeah I think let's move forward.

Generally I think the advice I'm coming to wrt proto compatibility between fuzzer revisions is just don't sweat it. These things converge pretty fast, and the text from older versions remains around during the transition... so we're seeing days to small numbers of weeks to regenerate examples when we do swap proto formats -- the testing gap that's introduced by changing the format realistically isn't enough to justify excessive care authoring the format.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
ctiller: Fair. It's probably worth throwing some any's into the status... we do some funky things based on attributes on statuses throughout the stack (we need not to, but we do) and so it might be interesting to trigger them *at some point*

ctiller (approved):

Issue #30838, opened 2022-09-04
Distinguish between PUBLIC & PRIVATE linkage


h-vetinari:
Hi!

I help package grpc / grpcio for conda-forge (mostly centered around the python ecosystem), which has much more powerful infrastructure than what the default `pip` can do, mostly w.r.t. dealing with non-python dependencies, and tracking which projects are exposed to which projects' ABIs, and rebuilding the respective artefact once new versions of those dependencies get released.

This works very well in practice, but requires some curation, and metadata to distinguish which packages are a host dependence (need to be available at compile time, generally shared libs & their headers), and which need to be available at runtime (e.g. most python stuff, plus the shared libs so the linkage can succeed at runtime).

gRPC is obviously a pretty involved package (due to exposure to various other tricky packages like abseil & protobuf), and so we've done our best to just get & keep things running so far.

But in the process of trying to clean up some metadata for zlib (and to a degree, abseil), we ran into breakage builds consuming abseil, due to the way the link interface of grpc is defined.

For example, the linkage of gprc is defined as:
https://github.com/grpc/grpc/blob/d52ed193d11dee797c0d51dc8db06032998b33f4/CMakeLists.txt#L2345-L2368

even though many of those libraries are only used internally within grpc (in particular zlib, re2, c-ares and abseil; not 100% sure about openssl; address_sorting and upb are less critical because they are vendored and therefore always present).

To avoid that ever project using grpc now has to declare a dependence on zlib (to be able to link to grpc, which has it in its public link interface), we tried to patch this for our builds. This is not really "live" at the moment, and also not meant to be a definitive patches, but I've uploaded a [branch](https://github.com/h-vetinari/grpc/tree/private_1.48) displaying those changes; the most relevant is https://github.com/h-vetinari/grpc/commit/ebd16ea34611011183d80ce1180383846d1a0e3c, the other two then deal only with abseil (which more tedious/voluminous but still important because by default, abseil's ABI changes per C++ version, and we do not want to "infect" other packages with grpc's C++ version if abseil is only used internally, and not e.g. as part of the public API).

Would the grpc project be willing to distinguish which libraries are necessary internally / externally?

Thanks for your consideration. :)

PS. Here's a good (but long) [overview](https://cmake.org/pipermail/cmake/2016-May/063400.html) of public vs. private linkage in CMake.

yashykt:
@jtattermusch ping

jtattermusch:
CC @veblush this might have to do with your recent PR.

[...]

PR #32958, opened 2023-04-27
[Cmake] Deps are now either public or private

veblush:
To address https://github.com/grpc/grpc/issues/30838 which needs a distinction of either private or public for all deps of cmake targets, Cmake template now begins to specify this all the times. As `CMakeLists.txt` is automatically generated from `BUILD` which doesn't have a concept of public or private dependency, there is no way to annotate metadata which could be used to populate this property for CMake. Therefore, this template is now doing this by statically written property (a.k.a hard coding).


h-vetinari:
@veblush, any updates on the effort in this PR by any chance?

PR #32970, opened 2023-04-28
[Audit Logging] Custom audit logger parsing in xDS registry.

rockspore:


src/core/ext/xds/xds_audit_logger_registry.cc
```diff
@@ -29,17 +32,22 @@
...
 
+using experimental::AuditLoggerRegistry;
+
 class StdoutLoggerConfigFactory : public XdsAuditLoggerRegistry::ConfigFactory {
  public:
   Json::Object ConvertXdsAuditLoggerConfig(
       const XdsResourceType::DecodeContext& /*context*/,
       absl::string_view /*configuration*/,
       ValidationErrors* /*errors*/) override {
+    // Stdout logger has no configuration right now. So we don't need to invoke
```
gtcooke94, line 49: Should we still invoke and just let it be a no-op?

gtcooke94 (approved):
This LGTM, one small comment

src/core/ext/xds/xds_audit_logger_registry.cc
rockspore: Since the config is sent as protobuf, we should be pretty confident that it doesn't contain anything unlike the case of Json. Therefore, I don't think it's necessary. See https://github.com/grpc/grpc/blob/e5725e47300ce3a0cba42b322c36cd8e8881199a/src/core/ext/xds/xds_lb_policy_registry.cc#L46 as an example.

markdroth: While I agree that it's not strictly required, I think it would be a good idea to invoke the gRPC registry, just to be safe.  That way, if we ever did something crazy like add a required parameter to the config, we'd detect the error here, which is what we want.

I suggest doing this parsing on line 88 rather than doing it here, so that we don't have to do it in each individual audit logger factory implementation.

markdroth (commented):
Just one comment here -- the rest looks great!

src/core/ext/xds/xds_audit_logger_registry.cc
rockspore: Sounds good. In this case, instead of adding back the TODO, I will hold this PR off till I add the stdout logger impl.

rockspore: Sorry I wasn't thinking very carefully 2 weeks ago. Revisiting this, I think if we are to invoke the gRPC registry, we will need to convert the `serialized_value` into a stdout logger's proto message and then construct a json object from it via some proto2json API, since manual conversion here defeats the purpose of validating it.

So I just wanted to double check if this is indeed what we want. If so, I think we do need specific logic in each logger factory since configs need to be marshaled into different protobufs and thus cannot parsing everything on line 88.

markdroth: Yes, I think what we want here is to convert the xDS proto to JSON form, and then use the gRPC registry to parse the JSON.

I believe this PR already has all the code we need to convert to JSON.  There are two cases here:
- A registered audit logger type.  (Currently, this includes only the stdout logger, but there could be others in the future.)  In this case, we will have a registered converter that knows how to read the fields in the protobuf and knows how to generate the JSON config from the protobuf config.
- A custom audit logger type.  In this case, the xDS proto contains a `xds.type.v3.TypedStruct` proto, which is directly convertible to JSON.  (This is handled by `ExtractXdsExtension()`, called on line 87 below.)

Whichever case we see, we just take the resulting JSON and validate it via the gRPC registry.

rockspore: Thanks, Mark! This makes sense.

So I modified the `ConfigFactory::ConvertXdsAuditLoggerConfig` a little bit here so that it returns the json config ready to be parsed/validated by the gRPC registry and the factory here is also given a `name()` method to return the logger name used in the gRPC registry. As such, we are able to perform the validation outside each individual factory here. The return value from `XdsAuditLoggerRegistry::ConvertXdsAuditLoggerConfig` is still `{"<logger_name>": <config>}`. PTAL.

src/core/ext/xds/xds_audit_logger_registry.cc
markdroth, line 1: Not related to this PR, but I just noticed: the `else` on line 81 is not actually necessary, since we're returning in the block above it.

markdroth (commented):
Just a couple of nits remaining.  Otherwise, this looks great!

src/core/ext/xds/xds_audit_logger_registry.cc
```diff
@@ -93,9 +92,17 @@ Json XdsAuditLoggerRegistry::ConvertXdsAuditLoggerConfig(
       auto config_factory_it =
           audit_logger_config_factories_.find(extension->type);
       if (config_factory_it != audit_logger_config_factories_.end()) {
-        return Json::FromObject(
+        auto json_config = Json::FromObject(
             config_factory_it->second->ConvertXdsAuditLoggerConfig(
                 context, *serialized_value, errors));
+        auto result = AuditLoggerRegistry::ParseConfig(
```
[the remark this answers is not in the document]
rockspore: I basically took your suggestion. Feels great to learn nice patterns like this. Thanks for the tip!

src/core/ext/xds/xds_audit_logger_registry.cc
rockspore: Removed the block and saved one level of indentation.

markdroth (approved):

PR #32973, opened 2023-04-29
[xDS LB] Override cluster with value from cookie

eugeneo:


src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -158,29 +164,44 @@ ArenaPromise<ServerMetadataHandle> StatefulSessionFilter::MakeCallPromise(
               this, cookie_config->name->c_str(),
               std::string(*cookie_value).c_str());
     }
-    // We have a valid cookie, so add the call attribute to be used by the
-    // xds_override_host LB policy.
-    service_config_call_data->SetCallAttribute(
-        GetContext<Arena>()->New<XdsOverrideHostAttribute>(*cookie_value));
+    std::pair<absl::string_view, absl::string_view> host_cluster =
```
markdroth, line 167: Note that this functionality needs to be guarded by the env var specified in the gRFC.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -158,29 +164,44 @@ ArenaPromise<ServerMetadataHandle> StatefulSessionFilter::MakeCallPromise(
...
+        absl::StrSplit(*cookie_value, absl::MaxSplits(';', 1));
+    if (!host_cluster.first.empty()) {
+      // We have a valid cookie, so add the call attribute to be used by the
```
markdroth, line 170: Here we need to check that the cluster specified in the cookie is actually present in the selected route.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -87,16 +88,21 @@ namespace {
 void MaybeUpdateServerInitialMetadata(
     const StatefulSessionMethodParsedConfig::CookieConfig* cookie_config,
     absl::optional<absl::string_view> cookie_value,
-    ServerMetadata* server_initial_metadata) {
+    absl::string_view cluster_name, ServerMetadata* server_initial_metadata) {
   // Get peer string.
   Slice* peer_string = server_initial_metadata->get_pointer(PeerString());
   if (peer_string == nullptr) return;  // Nothing we can do.
+  std::string new_value;
+  if (cluster_name.empty()) {
```
markdroth, line 96: This also needs to be guarded by the env var.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -158,29 +189,46 @@ ArenaPromise<ServerMetadataHandle> StatefulSessionFilter::MakeCallPromise(
...
+  auto cluster_name_attribute = static_cast<XdsClusterAttribute*>(
+      service_config_call_data->GetCallAttribute(
+          XdsClusterAttribute::TypeName()));
+  if (cluster_name_attribute != nullptr) {
+    cluster_name = cluster_name_attribute->cluster();
   }
   // Intercept server initial metadata.
   call_args.server_initial_metadata->InterceptAndMap(
-      [cookie_config, cookie_value](ServerMetadataHandle md) {
+      [cookie_config, cookie_value, cluster_name](ServerMetadataHandle md) {
```
markdroth, line 214: I don't think it's safe to store the `absl::string_view` pointing to the cluster name in this lambda, because that `string_view` points to the storage inside the `ClusterRef` object, and that object may have been destroyed by the time the lambda runs, because the call will drop its ref to the `ClusterRef` as soon as it is committed (i.e., as soon as it's done with all LB picks), which happens before we see server initial metadata.  To fix this, I think we're going to need to allocate a copy of the cluster string on the arena.  See my comment above about moving the cluster string arena allocation into `OverrideClusterIfNeeded()`, which I am suggesting renaming to `GetClusterNameToUse()`.

With that change in place, I think the logic here should look something like this:

```
std::string cookie_value = GetOverrideCookie(
    call_args.client_initial_metadata, *cookie_config->name);
absl::string_view cluster_from_cookie;
// Cookie format is "host;cluster".
std::pair<absl::string_view, absl::string_view> host_cluster =
    absl::StrSplit(cookie_value, absl::MaxSplits(';', 1));
// Set override host attribute.  Allocate the string on the
// arena, so that it has the right lifetime.
if (!host_cluster.first.empty()) {
  char* host_str = GetContext<Arena>()->Alloc(host_cluster.first.size());
  memcpy(host_str, host_cluster.first.data(), host_cluster.first.size());  
  service_config_call_data->SetCallAttribute(
      GetContext<Arena>()->New<XdsOverrideHostAttribute>(
          absl::string_view(host_str, host_cluster.first.size())));
}
// Check if the cluster override is valid, and apply it if necessary.
// Note that cluster_name will point to an arena-allocated string
// that will still be alive when we see the server initial metadata.
// If the cluster name is empty, that means we cannot use a
// cluster override (i.e., the route uses a cluster specifier plugin).
absl::string_view cluster_name =
    GetClusterNameToUse(host_cluster.second, service_config_call_data);
```

test/cpp/end2end/xds/BUILD
```diff
@@ -156,6 +156,33 @@ grpc_cc_test(
     ],
 )
 
+grpc_cc_test(
+    name = "xds_cluster_override_host_end2end_test",
```
markdroth, line 160: Instead of creating a new file, let's just add these tests to the existing xds_override_host_end2end_test.cc file.  The functionality we're adding here is really part of that same feature, so it makes sense to have all of the tests in one place.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -158,29 +189,46 @@ ArenaPromise<ServerMetadataHandle> StatefulSessionFilter::MakeCallPromise(
...
               std::string(*cookie_value).c_str());
     }
-    // We have a valid cookie, so add the call attribute to be used by the
-    // xds_override_host LB policy.
-    service_config_call_data->SetCallAttribute(
-        GetContext<Arena>()->New<XdsOverrideHostAttribute>(*cookie_value));
+    std::pair<absl::string_view, absl::string_view> host_cluster =
+        absl::StrSplit(*cookie_value, absl::MaxSplits(';', 1));
+    if (!host_cluster.first.empty() &&
+        OverrideClusterIfNeeded(host_cluster.second, service_config_call_data,
```
markdroth, line 195: I think we should set the host override independently of the cluster override.  Even if we're not setting the cluster override, it's conceivably possible that the same host could exist in whatever cluster we choose, so we should still set the host override.  And I think it makes the logic here a bit simpler to decouple the two.

markdroth (commented):

I haven't yet reviewed the tests, but I didn't want to make you wait for the code comments.  I'll review the tests in the next round.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -1213,6 +1087,205 @@ class XdsResolverFactory : public ResolverFactory {
...
+    return nullptr;
+  }
+  auto cluster = route_data_->Find(cluster_name);
+  route_data_.reset();
+  return cluster;
+}
+
+bool XdsClusterDataAttribute::HasClusterForRoute(
+    absl::string_view cluster_name) const {
+  auto prefixless = absl::StripPrefix(cluster_name, "cluster:");
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -1213,6 +1087,205 @@ class XdsResolverFactory : public ResolverFactory {
...
+        for (const auto& cluster : clusters) {
+          if (cluster.name == prefixless) {
+            return true;
+          }
+        }
+        return false;
+      },
+      [&](const XdsRouteConfigResource::Route::RouteAction::
+              ClusterSpecifierPluginName& name) {
+        return absl::StrCat("cluster_specifier_plugin:",
```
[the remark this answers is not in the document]
eugeneo: I was not sure this code is specific to session affinity.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -162,6 +255,31 @@ class XdsResolver : public Resolver {
     if (xds_client_ != nullptr) xds_client_->ResetBackoff();
   }
 
+  void MaybeRemoveUnusedClustersOnWorkSerializer() {
+    work_serializer_->Run(
+        [resolver = Ref()]() mutable {
```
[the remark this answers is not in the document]
eugeneo: Hope I understood the issue correctly.

eugeneo (commented):
Thank you for the review. Hope I managed to address all the comments...

eugeneo (commented):
Thank you for the review. Hope I managed to address all the comments...

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -267,6 +316,140 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
   EXPECT_EQ(2, backends_[2]->backend_service()->request_count());
   ResetBackendCounters();
 }
+
+TEST_P(OverrideHostTest, HostOverridden) {
+  CreateAndStartBackends(4);
```
markdroth, line 321: We probably only need 3 backends for this test, 1 in the first cluster and 2 in the second.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -267,6 +316,140 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
+  EXPECT_EQ(backends_[3]->backend_service()->request_count(), 0);
+}
+
+TEST_P(OverrideHostTest, HostGone) {
+  CreateAndStartBackends(4);
+  const char* kNewCluster1Name = "new_cluster_1";
+  const char* kNewEdsService1Name = "new_eds_service_name_1";
+  const char* kNewCluster2Name = "new_cluster_2";
+  const char* kNewEdsService2Name = "new_eds_service_name_2";
+  const uint32_t kWeight1 = std::numeric_limits<uint32_t>::max() / 4;
```
markdroth, line 375: Same as above: The weights can be 1 and 3.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -267,6 +316,140 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
+  SetListenerAndRouteConfiguration(
+      balancer_.get(), BuildListenerWithStatefulSessionFilter(),
+      BuildRouteConfiguration(
+          {{kNewCluster1Name, kWeight1}, {kNewCluster2Name, kWeight2}}));
+  WaitForAllBackends(DEBUG_LOCATION, 0, 4);
+  auto session_cookie =
+      GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 3, 10);
+  ASSERT_FALSE(session_cookie.empty());
+  // Remove backends[1] from cluster 2
+  SetEdsCdsResources(kNewCluster2Name, kNewEdsService2Name, 1, 3);
```
markdroth, line 395: This does not need to change the CDS resource, only the EDS resource.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -267,6 +316,140 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
+  WaitForAllBackends(DEBUG_LOCATION, 0, 3);
+  CheckRpcSendOk(DEBUG_LOCATION, kNumEchoRpcs,
+                 RpcOptions().set_metadata(session_cookie));
+  // Traffic goes to a second cluster, where it is equally distributed between
```
markdroth, line 399: I don't think this test needs to care about the exact distribution, since we're not testing round_robin here.  I think that we should just check that the sum of requests seen by the two backends add up to `kNumEchoRpcs`.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -267,6 +316,140 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
+  // All requests go to the backend we requested.
+  CheckRpcSendOk(DEBUG_LOCATION, kNumEchoRpcs,
+                 RpcOptions().set_metadata(session_cookie));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 0);
+  EXPECT_EQ(backends_[1]->backend_service()->request_count(), 0);
+  EXPECT_EQ(backends_[2]->backend_service()->request_count(), kNumEchoRpcs);
+  EXPECT_EQ(backends_[3]->backend_service()->request_count(), 0);
+}
+
+TEST_P(OverrideHostTest, HostGone) {
```
markdroth, line 369: Please call this something like `ClusterOverrideHonoredButHostGone`.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -150,6 +155,50 @@ class OverrideHostTest : public XdsEnd2endTest {
...
+    for (const auto& cluster : clusters) {
+      auto* weighted_cluster =
+          route1->mutable_route()->mutable_weighted_clusters()->add_clusters();
+      weighted_cluster->set_name(cluster.clusterName);
+      weighted_cluster->mutable_weight()->set_value(cluster.weight);
+    }
+    return new_route_config;
+  }
+
+  void SetEdsCdsResources(const char* cluster_name,
```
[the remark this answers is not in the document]

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -150,6 +155,50 @@ class OverrideHostTest : public XdsEnd2endTest {
...
+  struct ClusterData {
+    const char* clusterName;
+    uint32_t weight;
+  };
+
+  RouteConfiguration BuildRouteConfiguration(
+      const std::vector<ClusterData>& clusters) {
+    RouteConfiguration new_route_config = default_route_config_;
+    auto* route1 = new_route_config.mutable_virtual_hosts(0)->mutable_routes(0);
+    route1->mutable_match()->set_prefix("");
```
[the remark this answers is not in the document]

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -150,6 +155,50 @@ class OverrideHostTest : public XdsEnd2endTest {
...
+    cluster.mutable_eds_cluster_config()->set_service_name(service_name);
+    balancer_->ads_service()->SetCdsResource(cluster);
+  }
+
+  struct ClusterData {
+    const char* clusterName;
+    uint32_t weight;
+  };
+
+  RouteConfiguration BuildRouteConfiguration(
```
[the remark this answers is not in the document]

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -121,13 +122,17 @@ class OverrideHostTest : public XdsEnd2endTest {
   std::vector<std::pair<std::string, std::string>>
   GetAffinityCookieHeaderForBackend(grpc_core::DebugLocation debug_location,
                                     size_t backend_index,
+                                    size_t max_requests = 0,
```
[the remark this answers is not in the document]
eugeneo: Removed. It was simply to prevent hangs. Limited number of retries to number of backends * 10. Slightly worried about intermittent failures.

eugeneo (commented):
@markdroth Thank you for the throughout review! I hope I addressed all the comments.

eugeneo (commented):
@markdroth Thank you for the throughout review! I hope I addressed all the comments.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -317,137 +306,131 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
+      BuildRouteConfigurationWithWeightedClusters(
           {{kNewCluster1Name, kWeight1}, {kNewCluster2Name, kWeight2}}));
   WaitForAllBackends(DEBUG_LOCATION, 0, 2);
-  auto session_cookie =
-      GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 1, 10);
-  ASSERT_FALSE(session_cookie.empty());
-  SetListenerAndRouteConfiguration(
-      balancer_.get(), BuildListenerWithStatefulSessionFilter(),
-      BuildRouteConfiguration({{kNewCluster1Name, kWeight1}}));
+  auto host2_in_cluster2_cookie =
```
markdroth, line 405: Please call this `backend1_cluster2_cookie`.  The backend indexes are 0-based, not 1-based.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -317,137 +306,131 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
-  SetEdsCdsResources(kNewCluster1Name, kNewEdsService1Name, 0, 2);
-  WaitForAllBackends(DEBUG_LOCATION, 0, 2);
+  SetCdsAndEdsResources(kNewCluster1Name, kNewEdsService1Name, 0, 2);
+  SetCdsAndEdsResources(kNewCluster3Name, kNewEdsService3Name, 2, 3);
+  SetListenerAndRouteConfiguration(
+      balancer_.get(), BuildListenerWithStatefulSessionFilter(),
+      BuildRouteConfigurationWithWeightedClusters(
+          {{kNewCluster1Name, kWeight1}, {kNewCluster3Name, kWeight2}}));
+  WaitForAllBackends(DEBUG_LOCATION, 0, 3);
+  auto host3_in_cluster3_cookie =
```
markdroth, line 419: I don't think we need this check.  The goal here is to check what we do with the original cookie, not to get a new cookie.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -317,137 +306,131 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
-      GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 1, 10);
-  ASSERT_FALSE(session_cookie.empty());
-  SetListenerAndRouteConfiguration(
-      balancer_.get(), BuildListenerWithStatefulSessionFilter(),
-      BuildRouteConfiguration({{kNewCluster1Name, kWeight1}}));
+  auto host2_in_cluster2_cookie =
+      GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 1);
+  ASSERT_FALSE(host2_in_cluster2_cookie.empty());
   // Cluster is gone
   balancer_->ads_service()->UnsetResource(kEdsTypeUrl, kNewEdsService1Name);
```
markdroth, line 409: I don't think we need to delete these EDS resources; we just need to update them.  And we don't need to update the CDS or EDS resources for cluster1, since we're not changing that at all; we're just changing the route config to stop using that cluster.

I think the code here should be:

```
// Create a new cluster, cluster 3, containing a new backend, backend 2.
SetCdsAndEdsResources(kNewCluster3Name, kNewEdsService3Name, 2, 3);
// Send an EDS update for cluster 1 that adds backend 1. (Now cluster 1 has backends 0 and 1.)
balancer_->ads_service()->SetEdsResource(BuildEdsResource(
    EdsResourceArgs({{"locality0", CreateEndpointsForBackends(0, 2)}}),
    kNewEdsService1Name));
```

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -317,137 +306,131 @@ TEST_P(OverrideHostTest, DrainingExcludedFromOverrideSet) {
...
   // Both backends are in the same cluster
-  SetEdsCdsResources(kNewCluster1Name, kNewEdsService1Name, 0, 2);
-  WaitForAllBackends(DEBUG_LOCATION, 0, 2);
+  SetCdsAndEdsResources(kNewCluster1Name, kNewEdsService1Name, 0, 2);
+  SetCdsAndEdsResources(kNewCluster3Name, kNewEdsService3Name, 2, 3);
+  SetListenerAndRouteConfiguration(
+      balancer_.get(), BuildListenerWithStatefulSessionFilter(),
+      BuildRouteConfigurationWithWeightedClusters(
+          {{kNewCluster1Name, kWeight1}, {kNewCluster3Name, kWeight2}}));
+  WaitForAllBackends(DEBUG_LOCATION, 0, 3);
```
markdroth, line 418: This should be:

```
WaitForBackend(DEBUG_LOCATION, 2);
```

markdroth (commented):
This is getting closer!

eugeneo (commented):
Thank you for the review.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -402,25 +408,24 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
...
   SetListenerAndRouteConfiguration(
       balancer_.get(), BuildListenerWithStatefulSessionFilter(),
       BuildRouteConfigurationWithWeightedClusters(
           {{kNewCluster1Name, kWeight1}, {kNewCluster3Name, kWeight2}}));
-  WaitForAllBackends(DEBUG_LOCATION, 0, 3);
-  auto host3_in_cluster3_cookie =
-      GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 2);
-  ASSERT_FALSE(host3_in_cluster3_cookie.empty());
+  WaitForAllBackends(DEBUG_LOCATION, 2);
+  GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 2, 5);
```
markdroth, line 426: What's the purpose of this line?  Seems like we don't actually need it.

markdroth (commented):
This is getting close!

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -162,6 +255,31 @@ class XdsResolver : public Resolver {
...
+    auto it = cluster_state_map_.find(cluster_name);
+    if (it == cluster_state_map_.end()) {
+      auto cluster = MakeRefCounted<ClusterRef>(Ref(), cluster_name);
+      cluster_state_map_.emplace(cluster->cluster_name(), cluster->WeakRef());
+      return cluster;
+    }
+    return it->second->Ref();
+  }
+
+  absl::StatusOr<RefCountedPtr<ServiceConfig>> CreateMethodConfig(
```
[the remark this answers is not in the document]
eugeneo: Did that. Note that resolver argument had to be added.

eugeneo (commented):
Thank you for your comments. New revision uploaded.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -147,14 +146,14 @@ absl::string_view GetClusterToUse(
       service_config_call_data->GetCallAttribute<XdsClusterAttribute>();
   GPR_ASSERT(cluster_attribute != nullptr);
   auto cluster_to_use = cluster_attribute->cluster();
-  constexpr absl::string_view kClusterPrefix = "cluster:";
+  constexpr static absl::string_view kClusterPrefix = "cluster:";
```
markdroth, line 149: Nit: Prefer `static constexpr` to `constexpr static`.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -147,14 +146,14 @@ absl::string_view GetClusterToUse(
       service_config_call_data->GetCallAttribute<XdsClusterAttribute>();
   GPR_ASSERT(cluster_attribute != nullptr);
   auto cluster_to_use = cluster_attribute->cluster();
```
markdroth, line 148: Since you've changed the logic here to no longer actually use this variable as the cluster to use in all cases, I suggest renaming this to `current_cluster`.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -83,35 +85,140 @@ StatefulSessionFilter::StatefulSessionFilter(ChannelFilter::Args filter_args)
...
 namespace {
 
+absl::string_view AllocateStringOnArena(
+    absl::string_view src1, absl::string_view src2 = absl::string_view()) {
+  if (src1.empty() && src2.empty()) {
+    return absl::string_view();
+  }
+  char* arena_allocated_value =
+      static_cast<char*>(GetContext<Arena>()->Alloc(src1.size() + src2.size()));
+  if (!src1.empty()) {
```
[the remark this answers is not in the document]
eugeneo: UBSan got angry:
```
I0607 00:14:59.863605298      28 xds_server.h:443]                     ADS[0x712000000300]: client does not need update for type=type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment name=eds_service_name
src/core/ext/filters/stateful_session/stateful_session_filter.cc:96:47: runtime error: null pointer passed as argument 2, which is declared to never be null
```

src/core/ext/filters/stateful_session/stateful_session_filter.cc
```diff
@@ -129,32 +131,42 @@ void MaybeUpdateServerInitialMetadata(
...
-    } else {
-      cluster_to_use = ToArena(cluster);
     }
   }
-  cluster_attribute->set_cluster(cluster_to_use);
-  return cluster_to_use;
+  // If we chose the cluster from the cookie above, then update
+  // the cluster name attribute as well.  Note that the attribute will
+  // point to the arena-allocated string, which will live long enough.
+  cluster_attribute->set_cluster(
```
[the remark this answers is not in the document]
markdroth: Thanks, this looks good now.

I'm not sure where the gap in your understanding is, but if you can help me understand which part you're not understanding, I can try to help you understand it. :)

markdroth (approved):
This looks great!  All remaining comments are minor, so feel free to merge after addressing.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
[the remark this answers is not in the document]
markdroth: Oh, interesting.  Well, in that case, I guess go ahead and put the conditionals back.

src/core/ext/filters/client_channel/resolver/xds/xds_resolver.cc
```diff
@@ -306,6 +306,11 @@ class XdsResolver : public Resolver {
         const std::vector<XdsRouteConfigResource::Route>& routes,
         const Duration& default_max_stream_duration);
 
+    static absl::StatusOr<RefCountedPtr<ServiceConfig>> CreateMethodConfig(
```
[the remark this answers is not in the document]
eugeneo: Done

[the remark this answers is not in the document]
eugeneo: Done. Also moved the definition.

src/core/ext/filters/stateful_session/stateful_session_filter.cc
eugeneo: Done

PR #32974, opened 2023-04-29
[python O11Y] Initial Implementation

XuanWang-Amos:
Testing Command: `bazel test --cache_test_results=no --test_output=streamed --runs_per_test=1 --test_timeout=10 "//src/python/grpcio_tests/tests/observability:_observability_test"`

### TODO:
 * Better error handling.


gnossen:
Looks like there are [build issues in CI](https://source.cloud.google.com/results/invocations/76456fd2-db6f-4666-a1e0-6b4b3f035061):

```
ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:client_call_tracer.h'
ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:python_census_context.h'
ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:sampler.h'
ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:server_call_tracer.h'
ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: 4 input file(s) do not exist
ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11 Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: 4 input file(s) do not exist
INFO: Elapsed time: 283.638s, Critical Path: 66.59s
INFO: 3692 processes: 1387 internal, 2305 processwrapper-sandbox.
FAILED: Build did NOT complete successfully
```

gnossen (commented):
Just a first batch. Sorry for the delay.

XuanWang-Amos:
> Looks like there are [build issues in CI](https://source.cloud.google.com/results/invocations/76456fd2-db6f-4666-a1e0-6b4b3f035061):
> 
> ```
> ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:client_call_tracer.h'
> ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:python_census_context.h'
> ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:sampler.h'
> ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: missing input file '//src/python/grpcio_observability/grpc_observability:server_call_tracer.h'
> ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11: Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: 4 input file(s) do not exist
> ERROR: /var/local/jenkins/grpc/src/python/grpcio_observability/grpc_observability/BUILD.bazel:19:11 Middleman _middlemen/_S_Ssrc_Spython_Sgrpcio_Uobservability_Sgrpc_Uobservability_Cobservability-cc_library-compile failed: 4 input file(s) do not exist
> INFO: Elapsed time: 283.638s, Critical Path: 66.59s
> INFO: 3692 processes: 1387 internal, 2305 processwrapper-sandbox.
> FAILED: Build did NOT complete successfully
> ```

I didn't include all files in this PR (the server and client tracer), will probably include them later or in another PR.

gnossen:
> or in another PR.

We cannot break CI in this PR, so they'll need to be included in this PR. Generally speaking, any PR put up for review should be complete, in the sense that it is buildable and would cause no issues if merged to the master branch.

This can also cause issues during review if there's a reference to code that is not included in the PR. I see a reference, go looking for it, and can't find it. If the issue is that the client and server tracers are still being worked on or not currently fit for inclusion on the master branch, you can just include a stub for the moment -- a no-op implementation of the interface.

XuanWang-Amos:
> > or in another PR.
> 
> We cannot break CI in this PR, so they'll need to be included in this PR. Generally speaking, any PR put up for review should be complete, in the sense that it is buildable and would cause no issues if merged to the master branch.
> 
> This can also cause issues during review if there's a reference to code that is not included in the PR. I see a reference, go looking for it, and can't find it. If the issue is that the client and server tracers are still being worked on or not currently fit for inclusion on the master branch, you can just include a stub for the moment -- a no-op implementation of the interface.

Makes sense. My initial thought was that this PR was already huge, and since the tracer changes are somewhat independent from the rest of the changes, I excluded them for now. I'll add them soon.

BUILD
```diff
@@ -2231,6 +2231,28 @@ grpc_cc_library(
...
+        "absl/base",
+        "absl/base:core_headers",
+        "absl/meta:type_traits",
+        "absl/status",
+        "absl/strings",
+        "absl/time",
+    ],
+    language = "c++",
+    tags = ["nofixdeps"],
+    visibility = ["//visibility:public"],
```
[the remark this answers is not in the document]
gnossen: `@grpc:public` is still too broad. This needs to be for _internal_ use only. I didn't realize before that there's an issue here that this visibility needs to work across both google-internal and github. This will likely require making some changes to the `grpc_cc_library` macro definition.

You'll need to add something like `@grpc:grpc_python_observability` which is included [here](https://github.com/grpc/grpc/blob/7b74b078854e8e683f5a441ab7e41c5c6388e1c0/bazel/grpc_build_system.bzl#L87) as well as the internal version of this file.

The reason here is that if we leave this public then it is a signal to all Bazel users that this is a _public API_ and we would be on the hook to maintain it as such going forward, including writing gRFCs for all additions to it.

src/cpp/ext/gcp/BUILD
```diff
@@ -81,7 +81,7 @@ grpc_cc_library(
         "absl/types:optional",
     ],
     language = "c++",
-    visibility = ["//test:__subpackages__"],
+    # visibility = ["//test:__subpackages__"],
```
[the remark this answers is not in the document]
gnossen: Same comment as above. Public is not sufficient.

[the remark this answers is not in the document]
XuanWang-Amos: ACK, will change it to `@grpc:grpc_python_observability`

BUILD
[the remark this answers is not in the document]
XuanWang-Amos: ACK, will change it to `@grpc:grpc_python_observability`

src/python/grpcio_observability/grpc_observability/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
...
+    deps = [
+        ":observability",
+    ],
+)
+
+py_library(
+    name = "pyobservability",
+    srcs = [
+        "__init__.py",
+        # "_measures.py",
```
gnossen, line 59: Please remove or add a comment explaining the presence of these commented lines.

gnossen (commented):
Another batch.

src/python/grpcio_observability/grpc_observability/_observability.py
```diff
@@ -0,0 +1,179 @@
...
+            _LOGGER.exception("grpc_observability init failed with: %s", e)
+
+        # 5. Init grpc.
+        # 5.1 Refister grpc_observability
+        # 5.2 set_server_call_tracer_factory
+        grpc._observability_init(self)
+
+    def exit(self) -> None:
+        # Sleep for 0.5s so all data can be flushed.
+        time.sleep(0.5)
```
gnossen, line 131: This constant is suspicious. Is there any synchronization we can do so that we don't have to sleep for an arbitrary amount of time?

src/python/grpcio/grpc/_observability.py
```diff
@@ -0,0 +1,117 @@
...
+    def create_server_call_tracer_factory(
+            self) -> ServerCallTracerFactoryCapsule:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def record_rpc_latency(self, method: str, rpc_latency: float,
+                           status_code: Any) -> None:
+        raise NotImplementedError()
+
+    def enable_tracing(self, enable: bool) -> None:
```
gnossen, line 70: The name `enable_tracing` implies that this unconditionally enables tracing and therefore takes no arguments. Instead, this should probably be named `set_tracing`. Same for `enable_stats` below.

src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -0,0 +1,353 @@
...
+        self.test_exporter = TestExporter(self.all_metric, self.all_span)
+
+    def tearDown(self):
+        os.environ['GRPC_GCP_OBSERVABILITY_CONFIG'] = ''
+
+    def testRecordUnaryUnary(self):
+        os.environ[
+            'GRPC_GCP_OBSERVABILITY_CONFIG'] = _VALID_CONFIG_TRACING_STATS
+        with grpc_observability.GCPOpenCensusObservability() as o11y:
+            o11y.init(exporter=self.test_exporter)
```
gnossen, line 136: Is there any reason not to initialize with the context manager. That is, to call `init` as part of `grpc_observability.GCPOpenCensusObservability().__enter__`?

src/python/grpcio_observability/grpc_observability/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
...
+        "client_call_tracer.h",
+        "constants.h",
+        "observability_main.h",
+        "python_census_context.h",
+        "sampler.h",
+        "server_call_tracer.h",
+    ],
+    includes = ["."],
+    deps = [
+        "//:grpc",
```
gnossen, line 38: When we get to the setuptools build, we're going to need to be very careful to ensure that the build only includes the code that is actually referenced from our o11y module otherwise the shared object library is going to be huge.

src/python/grpcio_observability/grpc_observability/observability_main.cc
```diff
@@ -0,0 +1,189 @@
...
+void AddCensusDataToBuffer(CensusData data) {
+  std::unique_lock<std::mutex> lk(kCensusDataBufferMutex);
+  kCensusDataBuffer->push(data);
+  if (kCensusDataBuffer->size() >= kExportThreshold) {
+      CensusDataBufferCV.notify_all();
+  }
+}
+
+
+GcpObservabilityConfig ReadObservabilityConfig() {
```
gnossen, line 90: This name implies that this is a [pure function](https://en.wikipedia.org/wiki/Pure_function), but it's modifying global mutable state. How about the name `ReadAndActivateObservabilityConfig`?

gnossen (commented):
One more batch.

src/python/grpcio_observability/grpc_observability/_observability.py
XuanWang-Amos: This is the time we wait for next batch in `AwaitNextBatchLocked`. And C++ is [doing similar things too](http://shortn/_huZUevSe7d). We might need change this interval to a more appropriate number later (already added TODO).

src/python/grpcio_tests/tests/observability/_observability_test.py
XuanWang-Amos: I do remember I have some reasons to separate them, but I don't recall :(, I'll move it to `__enter__`.

src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -0,0 +1,353 @@
...
+
+            port = self._start_server()
```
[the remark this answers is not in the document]
XuanWang-Amos: I'll move stop server to `tearDown`,  but we need inject server tracer with different configuration before server start so it's better to keep it out of `setUp`.

src/python/grpcio/grpc/_observability.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def _observability_enabled(self) -> bool:
+        return self._tracing_enabled() or self._stats_enabled()
+
+
+def _observability_init(grpc_observability: GrpcObservability) -> None:
+    try:
+        grpc._observability._grpc_observability_stub = grpc_observability
+        _cygrpc.set_server_call_tracer_factory(grpc_observability)
+    except Exception as e:  # pylint:disable=broad-except
```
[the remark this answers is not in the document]
XuanWang-Amos: I changed it a bit and right now it only catches error from `set_server_call_tracer_factory`.

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
```diff
@@ -0,0 +1,249 @@
...
+
+def _start_exporting_thread(object exporter) -> None:
+  global global_export_thread
+  global GLOBAL_SHUTDOWN_EXPORT_THREAD
+  GLOBAL_SHUTDOWN_EXPORT_THREAD = False
+  global_export_thread = Thread(target=_export_census_data, args=(exporter,))
+  global_export_thread.start()
+
+
+def set_gcp_observability_config(object py_config) -> bool:
```
gnossen, line 71: Please add a comment indicating what the return value represents.

src/python/grpcio_observability/grpc_observability/_observability.py
gnossen: If these two constants need to agree, then they should be pulled from a common constant so that we can't accidentally change one value but not the other.

This method isn't fool-proof. A sleep guarantees that a thread will take _at least_ that long, not exactly that long. If the thread running `AwaitNextBatchLocked` gets swapped out for some reason, then it's totally possible for the `exit` call to complete before `AwaitNextBatchLocked` finishes. The question is how big an issue that would be.

@yashykt Do you have any thoughts on this?

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
```diff
@@ -0,0 +1,249 @@
...
+  SERVER_RECEIVED_BYTES_PER_RPC = kRpcServerReceivedBytesPerRpcMeasureName
+  SERVER_SERVER_LATENCY = kRpcServerServerLatencyMeasureName
+  SERVER_STARTED_RPCS = kRpcServerStartedRpcsMeasureName
+
+# Delay map creation due to circular dependencies
+_CY_METRICS_NAME_TO_PY_METRICS_NAME_MAPPING = {}
+
+def cyobservability_init(object exporter) -> None:
+  global _CY_METRICS_NAME_TO_PY_METRICS_NAME_MAPPING
+  _CY_METRICS_NAME_TO_PY_METRICS_NAME_MAPPING = {x.value: x for x in grpc_observability.MetricsName}
```
gnossen, line 58: This could be done at module initialization time instead of during this function call by just moving it up a few lines. This would avoid the runtime modification of global mutable state (signalled by the `global` keyword), which should be done sparingly.

src/python/grpcio/grpc/_observability.py
```diff
@@ -0,0 +1,117 @@
...
+from grpc._cython import cygrpc as _cygrpc
+
+_LOGGER = logging.getLogger(__name__)
+
+_channel = Any  # _channel.py imports this module.
+ClientCallTracerCapsule = TypeVar('ClientCallTracerCapsule')
+ServerCallTracerFactoryCapsule = TypeVar('ServerCallTracerFactoryCapsule')
+
+_lock: threading.RLock = threading.RLock()
+_grpc_observability_stub: Optional[ObservabilityPlugin] = None  # pylint: disable=used-before-assignment
```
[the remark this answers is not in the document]
XuanWang-Amos: It's still complaining: http://shortn/_C9qcLmg3PE

src/python/grpcio_observability/grpc_observability/observability_util.cc
```diff
@@ -0,0 +1,189 @@
...
+
+#include "src/python/grpcio_observability/grpc_observability/observability_util.h"
+#include "src/python/grpcio_observability/grpc_observability/server_call_tracer.h"
+#include "src/python/grpcio_observability/grpc_observability/client_call_tracer.h"
+
+#include "absl/strings/string_view.h"
+
+namespace grpc_observability {
+
+std::queue<CensusData>* kCensusDataBuffer;
```
gnossen, line 23: [The `k` prefix indicates that this is a constant](https://google.github.io/styleguide/cppguide.html#Constant_Names), which it is not (otherwise, the `const` keyword would show up somewhere). I think you want a [`g` prefix to indicate that it is a global](https://github.com/grpc/grpc/blob/eecc8978b653f426eebda3d2a40871302108ebeb/src/core/ext/transport/chttp2/transport/chttp2_transport.cc#L122). Also, [variable names in our version of C++ are generally in `snake_case`](https://google.github.io/styleguide/cppguide.html#Variable_Names). So, for example, this variable should be named `g_census_data_buffer`. Same for the accompanying mutex and condition variable.

I know these rules probably seem gratuitous, but they make the job of future readers of the code much easier. We don't have to pay much attention to variables that start with `k` (because they never change), but variables starting with `g` are suspicious because they can change over time and are often the root causes of bugs.

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
```diff
@@ -0,0 +1,249 @@
...
+  NativeObservabilityInit()
+  _start_exporting_thread(exporter)
+
+
+def _start_exporting_thread(object exporter) -> None:
+  global global_export_thread
+  global GLOBAL_SHUTDOWN_EXPORT_THREAD
+  GLOBAL_SHUTDOWN_EXPORT_THREAD = False
```
gnossen, line 66: Is it necessary to set this here? This is already the default value.

src/python/grpcio/grpc/_observability.py
```diff
@@ -0,0 +1,117 @@
...
+    @property
+    def observability_enabled(self) -> bool:
+        return self.tracing_enabled or self.stats_enabled
+
+
+def _observability_init(observability_plugin: ObservabilityPlugin) -> None:
+    global _grpc_observability_stub  # pylint: disable=global-statement
+    try:
+        with _lock:
+            _grpc_observability_stub = observability_plugin
```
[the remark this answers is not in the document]
gnossen: Okay, I wasn't sure. Let me take another look through this code with that in mind.

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
XuanWang-Amos: That's what's I did at first, but we'll have an circular import issue: This map needs `grpc_observability.MetricsName` and `grpc_observability.MetricsName`  requires `Cython.MetricsName`.

I moved `grpc_observability.MetricsName` to Cython file so we don't have circular import anymore, but the downside is `MetricsName` which should be a public class is now imported from a Cython file.

src/python/grpcio_observability/grpc_observability/observability_util.cc
XuanWang-Amos: Agree that we should follow naming conventions, and I'm trying to do so across all of the files. I must've missed those while checking...

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
gnossen: > but we'll have an circular import issue

That makes sense. Please add a comment here explaining why this is being done.

> but the downside is MetricsName which should be a public class is now imported from a Cython file

Besides this hack to get around the circular import, that doesn't seem like a big deal to me. Could you maybe instead move the definition of `grpc_observability.MetricsName` to this file (or some new third file) and re-import it in `__init__.py`?

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
XuanWang-Amos: I think it's better to keep this line, for the use case that user tries to enable observability again after exit (also in testing where we import `grpc_observability` once and need reinitiate it multiple times across different test cases).

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
XuanWang-Amos: That's what I did now, both `MetricsName` and `CyMetricsName` are in `.pyx` file and `MetricsName` was imported to `__init__.py`.

src/python/grpcio_observability/grpc_observability/client_call_tracer.cc
```diff
@@ -0,0 +1,242 @@
...
+    context_.EndSpan();
+    if (IsSampled()) {
+      RecordSpan(context_.Span().ToCensusData());
+    }
+  }
+}
+
+
+PythonCensusContext PythonOpenCensusCallTracer::CreateCensusContextForCallAttempt() {
+  auto context = PythonCensusContext(absl::StrCat("Attempt.", method_), &(context_.Span()), context_.Labels());
```
gnossen, line 71: What distinguishes individual attempts from each other? It looks like both the name and the span ID will be the same for all attemps.

gnossen, line 71: Why "." as a separator between "Attempt" and the method? Seems like " " might be a better delimiter since " " is not a legal character in method names.

src/python/grpcio_observability/grpc_observability/client_call_tracer.cc
```diff
@@ -0,0 +1,242 @@
...
+    if (status_code_ != absl::StatusCode::kOk) {
+      context_.Span().SetStatus(StatusCodeToString(status_code_));
+    }
+    context_.EndSpan();
+    if (IsSampled()) {
+      RecordSpan(context_.Span().ToCensusData());
+    }
+  }
+
+  delete this;
```
gnossen, line 232: Please add the comment "After RecordEnd, Core will make no further usage of this CallAttemptTracer, so we are free it here."

src/python/grpcio_observability/grpc_observability/_observability.py
```diff
@@ -0,0 +1,180 @@
...
+                          stats_enabled: bool = False) -> None:
+        self.project_id = project_id
+        self.stats_enabled = stats_enabled
+        self.tracing_enabled = tracing_enabled
+        self.labels = labels
+        self.sampling_rate = sampling_rate
+
+
+# pylint: disable=no-self-use
+class GCPOpenCensusObservability(grpc._ObservabilityPlugin):
```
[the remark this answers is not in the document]
gnossen: Just realized the submission thread implementation actually _is_ specific to OpenCensus, so it's probably fine to leave this the way it ls.

src/python/grpcio_observability/grpc_observability/client_call_tracer.cc
```diff
@@ -0,0 +1,242 @@
...
+
+
+void PythonOpenCensusCallTracer::PythonOpenCensusCallAttemptTracer::RecordCancel(
+    absl::Status /*cancel_error*/) {}
+
+
+void PythonOpenCensusCallTracer::PythonOpenCensusCallAttemptTracer::RecordEnd(
+    const gpr_timespec& /*latency*/) {
+  if (PythonOpenCensusStatsEnabled()) {
+    std::vector<Label> labels = context_.Labels();
```
gnossen, line 210: Instead of copying to a local variable, can you just use `context_.Labels()` in this function where you're currently using `labels`? You're about to delete the context anyway, so it seems like adding to the labels shouldn't hurt anything.

src/python/grpcio_observability/grpc_observability/client_call_tracer.cc
gnossen: > we'll create a new spanId for each individual attempts

Where does this happen? I don't see a new span ID being generated in this function and the result of this functions seems to be used as the context in the CallAttemptTracer constructor.

src/python/grpcio_observability/grpc_observability/python_census_context.cc
```diff
@@ -0,0 +1,268 @@
...
+  return std::string(hex_string.str());
+}
+
+} // namespace
+
+//
+// Span
+//
+
+Span Span::StartSpan(absl::string_view name, Span* parent) {
```
gnossen, line 190: You'll have to change around the `Span` APIs to make this work, but this should be `const Span*` since we're not mutating `parent` here.

To make this work, you'll need to add a `ConstContext()` method on `Span` or you could just get rid of the method-based accessor, make `Span::context_` public, and access it directly here.
 
More generally, if a method/function does not mutate one of its pointer/reference arguments (non-const value parameters are fine), then that parameter should be marked `const` unless we have a strong reason not to do so.

src/python/grpcio_observability/grpc_observability/python_census_context.cc
```diff
@@ -0,0 +1,268 @@
...
+  if (parent != nullptr) {
+    parent_span_id = parent->Context().SpanId();
+    trace_id = parent->Context().TraceId();
+    should_sample = parent->Context().IsSampled();
+  } else {
+    trace_id = generateTraceId();
+    should_sample = ShouldSample(trace_id);
+  }
+
+  context = SpanContext(trace_id, span_id, should_sample);
```
gnossen, line 207: Nit: You could just declare `context` here instead of default constructing then overwriting.

src/python/grpcio_observability/grpc_observability/python_census_context.h
```diff
@@ -0,0 +1,327 @@
...
+ private:
+  std::string trace_id_;
+  std::string span_id_;
+  bool should_sample_;
+  bool is_valid_;
+};
+
+class Span final {
+ public:
+  explicit Span(std::string name, std::string parent_span_id, absl::Time start_time, SpanContext context)
```
gnossen, line 159: Const issues:

```c++
explicit Span(const std::string& name, const std::string& parent_span_id, absl::Time start_time, const SpanContext& context)
```

(`absl::Time` is a pass-by-value type based on its size)

gnossen (commented):
I think I've looked over every file now.

src/python/grpcio/grpc/_observability.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class GrpcObservability(Generic[PyCapsule], metaclass=abc.ABCMeta):
+    # we need to add hooks so that the GCP observability package can register functions with
+    # the grpcio module and so can any other observability module conforming to the interface.
+    _TRACING_ENABLED: bool = False
+    _STATS_ENABLED: bool = False
+
+    @abc.abstractmethod
+    def create_client_call_tracer_capsule(self, method_name: bytes) -> PyCapsule:
```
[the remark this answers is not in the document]
gnossen: Leaving this unresolved for now as a reminder.

src/python/grpcio/grpc/_observability.py
```diff
@@ -0,0 +1,125 @@
...
+
+def set_plugin(observability_plugin: Optional[_ObservabilityPlugin]) -> None:
+    global _OBSERVABILITY_PLUGIN  # pylint: disable=global-statement
+    with _plugin_lock:
+        if observability_plugin and _OBSERVABILITY_PLUGIN:
+            raise ValueError("observability_plugin was already set!")
+        _OBSERVABILITY_PLUGIN = observability_plugin
+
+
+def _observability_init(observability_plugin: _ObservabilityPlugin) -> None:
```
gnossen, line 104: What I'm currently seeing is:

- When entering the o11y context manager, so the `observability` module calls `_observability_init`
  - `_observability_init` calls `set_plugin`
  - `_observability_init` calls `set_server_call_tracer_factory`
- When exiting the o11y context manager, the `observability` module calls `set_plugin(None)`

Some thoughts about this.
- I don't see parallelism between the calls that the o11y context manager makes. For example, `set_plugin(foo)` and `set_plugin(None)` or `observability_init(foo)` and `observability_deinit()`
- In the current setup, I don't see us ever cleaning up the server call tracer factory.
- The name scheme here is confusing. `set_plugin` and `_observability_init` are both protected APIs -- ones that should only be used by packages maintained by us but I'm seeing a mixture of underscore patterns. I would suggest letting the underscore of the `_observability` module indicate that people should not touch this and removing the underscores from the functions in this module.

src/python/grpcio/grpc/_observability.py
```diff
@@ -0,0 +1,125 @@
...
+    set_plugin(observability_plugin)
```
gnossen, line 105: We don't want to allow nested context managers, so we should raise an exception if we try to set a plugin when one has already been installed.

src/python/grpcio_observability/grpc_observability/_observability.py
gnossen: I just got some clarity on this from @ejona86 . The sleep is considered a kludge. We can add some synchronization with the export thread to exit as soon as all data has been exported.

src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -0,0 +1,341 @@
...
+            return _MethodHandler(False, True)
+        elif handler_call_details.method == _STREAM_UNARY:
+            return _MethodHandler(True, False)
+        elif handler_call_details.method == _STREAM_STREAM:
+            return _MethodHandler(True, True)
+        else:
+            return None
+
+
+class ObservabilityTest(unittest.TestCase):
```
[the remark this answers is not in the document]
XuanWang-Amos: I checked existing interop tests and looks like we already have most of them covered with following exceptions:
* Test trace sampling rate.
  * Will add this.
* Test use config file over env var.
  * Will add this.
* Test empty config.
  * WiIl add this.
* Logging test.
  * Logging not implemented yet.
* Test custom labels.
  * This will be added together with `OpenCensusExporter` (The plan is add it in next PR).
* Test integration with GCP.
  * I don't think we need this since this will be handled by OC StackDriver extension, which can also be test in interop tests.
* Test resource type and labels.
  * Those will be added by OC StackDriver extension, I need to check if we can verify them before sending to GCP, otherwise I suggest test it in interop tests.

src/python/grpcio_observability/grpc_observability/server_call_tracer.h
```diff
@@ -0,0 +1,44 @@
...
+#include "src/python/grpcio_observability/grpc_observability/python_census_context.h"
+#include "src/python/grpcio_observability/grpc_observability/observability_util.h"
+
+namespace grpc_observability {
+
+class PythonOpenCensusServerCallTracerFactory
+    : public grpc_core::ServerCallTracerFactory {
+ public:
+  grpc_core::ServerCallTracer* CreateNewServerCallTracer(
+      grpc_core::Arena* arena) override;
```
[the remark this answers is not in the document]
XuanWang-Amos: They're not public API but he said we can use them in Python.

src/python/grpcio_observability/grpc_observability/client_call_tracer.cc
```diff
@@ -0,0 +1,244 @@
...
+        double latency_ms = absl::ToDoubleMilliseconds(absl::Microseconds(
+            gpr_timespec_to_micros(transport_stream_stats->latency)));
+        RecordDoubleMetric(kRpcClientTransportLatencyMeasureName, latency_ms, labels);
+      }
+    }
+  }
+}
+
+
+void PythonOpenCensusCallTracer::PythonOpenCensusCallAttemptTracer::RecordCancel(
```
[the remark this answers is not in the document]
XuanWang-Amos: Checked with Yash, the answer is yes.

src/python/grpcio_observability/grpc_observability/_observability.py
```diff
@@ -0,0 +1,172 @@
...
+
+        if self.config.tracing_enabled:
+            self.set_tracing(True)
+        if self.config.stats_enabled:
+            self.set_stats(True)
+
+    def exit(self) -> None:
+        # Sleep for 0.5s so all data can be flushed.
+        # The time equals to the time in AwaitNextBatchLocked.
+        time.sleep(_cyobservability.CENSUS_EXPORT_BATCH_INTERVAL)
```
XuanWang-Amos, line 119: > The sleep is considered a kludge. We can add some synchronization with the export thread to exit as soon as all data has been exported.
* It's not enough to exit as soon as all data has been exported, If we shutdown export thread immediately after exit, it's possible that core didn't call RecordEnd in callTracer, and all data recorded by calling RecordEnd will be lost.
* It's also possible to ignore those metrics since we have already exited the observability context, so anything recorded after this can be ignored. However, I think we should keep it because RecordEnd is part of RPC, and we should at least try our best (sleep one interval) to make sure it's called before exit.

src/python/grpcio_observability/grpc_observability/server_call_tracer.h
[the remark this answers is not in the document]
gnossen: If non-public Core APIs are a part of our API surface for the `ObservabilityPlugin` interface, then we cannot allow _any_ alternative implementations for the `ObservabilityPlugin` interface besides our `grpc_observability` package. This should be display very prominently in the docstrings.

Since this API also seems like it might not work cross-DLL on Windows, I think it may be better to provide an API that does not require use of the Arena for memory allocation.

CC @yashykt

src/python/grpcio_observability/grpc_observability/python_census_context.cc
```diff
@@ -0,0 +1,268 @@
...
+  if (tracing_buf_size < kGrpcTraceBinHeaderLen) {
+    return 0;
+  }
+  ToGrpcTraceBinHeader(context, reinterpret_cast<uint8_t*>(tracing_buf));
+  return kGrpcTraceBinHeaderLen;
+}
+
+
+size_t StatsContextSerialize(size_t /*max_tags_len*/, grpc_slice* /*tags*/) {
+  // TODO(unknown): Add implementation. Waiting on stats tagging to be added.
```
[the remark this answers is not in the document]
gnossen: Do we need the `TODO` then?

src/python/grpcio_tests/tests/observability/_observability_test.py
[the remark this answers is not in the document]
gnossen: Sounds great!

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
```diff
@@ -0,0 +1,238 @@
...
+  void RecordSendInitialMetadata(
+      grpc_metadata_batch* /*send_initial_metadata*/) override {}
+
+  void RecordSendTrailingMetadata(
+      grpc_metadata_batch* send_trailing_metadata) override;
+
+  void RecordSendMessage(const grpc_core::SliceBuffer& send_message) override {
+    RecordAnnotation(
+        absl::StrFormat("Send message: %ld bytes", send_message.Length()));
+    ++sent_message_count_;
```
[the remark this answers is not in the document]
gnossen: >  if we are following the same model that the C++ call tracer follows.

Can you be a bit more clear on this. What is the model exactly?

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
```diff
@@ -0,0 +1,331 @@
...
+  _start_exporting_thread(exporter)
+
+
+def _start_exporting_thread(object exporter) -> None:
+  exporter: grpc_observability.Exporter
+
+  global GLOBAL_EXPORT_THREAD
+  global GLOBAL_SHUTDOWN_EXPORT_THREAD
+  GLOBAL_SHUTDOWN_EXPORT_THREAD = False
+  GLOBAL_EXPORT_THREAD = Thread(target=_export_census_data, args=(exporter,))
```
gnossen, line 94: Should this be a daemon thread? If not, the thread hanging could prevent the process from exiting?

src/python/grpcio_observability/grpc_observability/_observability.py
gnossen: > It's also possible to ignore those metrics since we have already exited the observability context, so anything recorded after this can be ignored. However, I think we should keep it because RecordEnd is part of RPC, and we should at least try our best 

This seems reasonable. The guarantee would be that "any RPC started while the context manager is active would be recorded".

So the concern is a situation like this?

- RPC A starts
- `GCPOpenCensusObservability.exit()` is called
  - A flush of the buffer is initiated
  - The export of the census data finishes
- `GCPOpenCensusObservability.exit()` returns based on its synchronization with the export thread
- RPC A finishes and `StartEnd` is called, but o11y has finished

I can see this being an issue. Could you just keep track of the number of outstanding RPCs and keep stats collection open until either all of the RPCs have finished or a timeout has elapsed? (You don't want to just block indefinitely for RPCs to finish)

> If we shutdown export thread immediately after exit, it's possible that core didn't call RecordEnd in callTracer

My concern that led to the request for synchronization was actually the opposite -- that by waiting for too short a period of time, the stats would not be exported in time, the process would exit, and the buffered data would be lost.

This relates to the lifecycle of the stats exporting thread. I just realized that this is not currently a daemon thread, so if the process tries to exit, it will block waiting for the stats thread to exit. This could happen if, for example, there is a bug in the opencensus library, so I _do_ think we should switch to a daemon thread.

But that introduces the problem with the process exiting (and killing the exporting thread) before the stats have been exported. My thought was that the `exit()` method here would block until either exporting has finished or we hit a user-configurable timeout.

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
```diff
@@ -0,0 +1,238 @@
...
+}
+
+//
+// PythonOpenCensusServerCallTracerFactory
+//
+
+grpc_core::ServerCallTracer*
+PythonOpenCensusServerCallTracerFactory::CreateNewServerCallTracer(
+    grpc_core::Arena* arena) {
+  return arena->ManagedNew<PythonOpenCensusServerCallTracer>();
```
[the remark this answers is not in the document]
gnossen: Discussed with @yashykt offline. To avoid this issue, we'll ignore the provided arena, call `new` / `delete` ourselves, and destroy the object at the end of the `RecordEnd` method.

src/python/grpcio_observability/grpc_observability/python_census_context.cc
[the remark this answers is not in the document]
XuanWang-Amos: Not really, removed.

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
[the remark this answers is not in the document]
XuanWang-Amos: Confirmed with Yash, same model here means:
* Once a call tracer factory is created and passed to core, core will control when to create call tracer and when to call `RecordSendMessage` and `RecordReceivedMessage`.

The same logic applies to why we don't need synchronization on client CallAttemptTracer.

src/python/grpcio/grpc/_observability.py
XuanWang-Amos: Discussed a little bit on this offline, it's more about the naming, I'll change the name to `observability_init` and `observability_deinit` on both `grpc_observability` and `grpc` module.

src/python/grpcio_observability/grpc_observability/server_call_tracer.h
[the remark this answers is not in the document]
XuanWang-Amos: I added comment to mention it was only mean to be used by gRPC team.

As for the DDL issue, we decided not to used the provided arena.

src/python/grpcio/grpc/_observability.py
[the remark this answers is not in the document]
gnossen: Ping.

bazel/grpc_build_system.bzl
```diff
@@ -113,6 +113,7 @@ def _update_visibility(visibility):
         "tsi": PRIVATE,
         "xds": PRIVATE,
         "xds_client_core": PRIVATE,
+        "grpc_python_observability": PRIVATE,
```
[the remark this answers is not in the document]
XuanWang-Amos: 👌

src/python/grpcio_observability/grpc_observability/_observability.py
```diff
@@ -0,0 +1,175 @@
...
+    exporter: "grpc_observability.Exporter"
+
+    def __init__(self, exporter: "grpc_observability.Exporter" = None):
+        self.exporter = None
+        self.config = GcpObservabilityPythonConfig.get()
+        if exporter:
+            self.exporter = exporter
+        else:
+            pass
+            # Actual implementation of OC exporter
```
gnossen, line 98: So there isn't a default exporter?

src/python/grpcio_observability/grpc_observability/_observability.py
```diff
@@ -0,0 +1,175 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        self.exit()
+
+    def exit(self) -> None:
+        # Sleep so we don't loss any data. If we shutdown export thread
+        # immediately after exit, it's possible that core didn't call RecordEnd
+        # in callTracer, and all data recorded by calling RecordEnd will be
+        # lost.
+        # The time equals to the time in AwaitNextBatchLocked.
+        time.sleep(_cyobservability.CENSUS_EXPORT_BATCH_INTERVAL)
```
gnossen, line 132: Please add a TODO linking to the issue tracking explicit synchronization.

src/python/grpcio_observability/grpc_observability/_cyobservability.pyx
```diff
@@ -0,0 +1,336 @@
...
+import logging
+import os
+from threading import Thread
+from typing import List, Mapping, Tuple, Union
+
+import grpc_observability
+
+# Time we wait for batch exporting census data
+# TODO(xuanwn): change interval to a more appropriate number
+CENSUS_EXPORT_BATCH_INTERVAL = float(os.environ.get('GRPC_PYTHON_CENSUS_EXPORT_BATCH_INTERVAL', 0.5))
```
gnossen, line 29: Please include a time unit in the name of this env var. I'm assuming seconds, so `GRPC_PYTHON_CENSUS_EXPORT_BATCH_INTERVAL_SECS`

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
```diff
@@ -0,0 +1,247 @@
...
+
+//
+// PythonOpenCensusServerCallTracerFactory
+//
+
+grpc_core::ServerCallTracer*
+PythonOpenCensusServerCallTracerFactory::CreateNewServerCallTracer(
+    grpc_core::Arena* arena) {
+  // We don't use arena here to to ensure that memory is allocated and freed in
+  // the same DDL in Windows.
```
gnossen, line 243: `DDL` -> `DLL`

src/python/grpcio_observability/grpc_observability/observability_util.h
```diff
@@ -0,0 +1,114 @@
...
+namespace grpc_observability {
+
+struct CensusData {
+  DataType type;
+  std::vector<Label> labels;
+  // TODO(xuanwn): We can use union here
+  SpanCensusData span_data;
+  Measurement measurement_data;
+  CensusData() {}
+  CensusData(const Measurement& mm, std::vector<Label>& labels)
```
gnossen, line 45: `const std::vector<Label>&`

src/python/grpcio_observability/grpc_observability/observability_util.h
```diff
@@ -0,0 +1,114 @@
...
+  CloudTrace cloud_trace;
+  CloudLogging cloud_logging;
+  std::string project_id;
+  std::vector<Label> labels;
+  bool is_valid;
+  GcpObservabilityConfig() : is_valid(false) {}
+  GcpObservabilityConfig(bool valid) : is_valid(true) {}
+  GcpObservabilityConfig(CloudMonitoring cloud_monitoring,
+                         CloudTrace cloud_trace, CloudLogging cloud_logging,
+                         std::string project_id, std::vector<Label> labels)
```
gnossen, line 75: `const std::string&`
`const std::vector<Label>&`

In general, any type that is large or may be backed by heap allocations should be passed by const ref. This includes `std::string`, `std::vector`, and all other containers. It also includes all classes/structs that include these as fields. Please do a a scrub through the PR for this.

src/python/grpcio_observability/grpc_observability/observability_util.h
```diff
@@ -0,0 +1,114 @@
...
+void* CreateServerCallTracerFactory();
+
+void NativeObservabilityInit();
+
+void AwaitNextBatchLocked(std::unique_lock<std::mutex>& lock, int timeout_ms);
+
+void AddCensusDataToBuffer(CensusData buffer);
+
+void RecordIntMetric(MetricsName name, int64_t value,
+                     std::vector<Label>& labels);
```
gnossen, line 101: `const std::vector<Label>&`

src/python/grpcio_observability/grpc_observability/observability_util.h
```diff
@@ -0,0 +1,114 @@
...
+
+void RecordDoubleMetric(MetricsName name, double value,
+                        std::vector<Label>& labels);
```
gnossen, line 104: `const std::vector<Label>&`

src/python/grpcio_observability/grpc_observability/observability_util.cc
```diff
@@ -0,0 +1,225 @@
...
+  return server_call_tracer_factory;
+}
+
+void AwaitNextBatchLocked(std::unique_lock<std::mutex>& lock, int timeout_ms) {
+  auto now = std::chrono::system_clock::now();
+  g_census_data_buffer_cv.wait_until(
+      lock, now + std::chrono::milliseconds(timeout_ms));
+}
+
+void AddCensusDataToBuffer(CensusData data) {
```
gnossen, line 117: `const CensusData&`

gnossen (commented):
Getting very close now. Good job pushing this huge change through!

src/python/grpcio_observability/grpc_observability/_observability.py
gnossen: No need to implement a mock now. Just raise a [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError) here and add a TODO pointing to a Github issue to take care of this.

src/python/grpcio_observability/grpc_observability/_observability.py
```diff
@@ -94,12 +95,10 @@ def __init__(self, exporter: "grpc_observability.Exporter" = None):
         if exporter:
             self.exporter = exporter
         else:
-            pass
-            # Actual implementation of OC exporter
-            # open_census = importlib.import_module(
-            #     "grpc_observability._open_census")
-            # self.exporter = open_census.OpenCensusExporter(
-            #     self.config.get().labels)
+            open_census = importlib.import_module(
```
gnossen, line 98: The `import_module` function should only be used when the module is not known ahead of time. Any reason you can't just import this at the top of the file with everything else?

src/python/grpcio_observability/grpc_observability/_observability.py
```diff
@@ -94,12 +95,10 @@ def __init__(self, exporter: "grpc_observability.Exporter" = None):
...
+                "grpc_observability._open_census_exporter")
+            self.exporter = open_census.OpenCensusExporter(
```
gnossen, line 100: We probably want to fail fast in this scenario so that people don't wonder why this is silently not exporting anything. Let's just `raise NotImplementedError()` for now.

src/python/grpcio_observability/grpc_observability/_observability.py
XuanWang-Amos: Circular import issue:
`grpc_observability.__init__ -> _observability.GCPOpenCensusObservability -> _open_census_exporter.OpenCensusExporter -> grpc_observability.__init__(Exporter/StatsData/TracingData)`

Any suggestions to to refactor some of them?

gnossen: Could you move `Exporter`, `StatsData`, and `TracingData` to a new file and reimport them in `__init__.py`?

src/python/grpcio_observability/grpc_observability/observability_util.h
```diff
@@ -0,0 +1,115 @@
...
+#include <vector>
+
+#include "absl/strings/string_view.h"
+
+#include <grpc/status.h>
+
+#include "src/python/grpcio_observability/grpc_observability/constants.h"
+#include "src/python/grpcio_observability/grpc_observability/python_census_context.h"
+
+#ifndef OBSERVABILITY_MAIN_H
```
gnossen, line 33: Include guards need to go before any actual content. After the copyright notice, but before includes. Here and in all header files.

src/python/grpcio_observability/grpc_observability/sampler.cc
```diff
@@ -0,0 +1,69 @@
...
+  // UINT64_MAX is (2^64)-1, but as a double rounds up to 2^64.
+  // For probabilities >= 1-(2^-54), the product wraps to zero!
+  // Instead, calculate the high and low 32 bits separately.
+  const double product = UINT32_MAX * probability;
+  double hi_bits, lo_bits = ldexp(modf(product, &hi_bits), 32) + product;
+  return (static_cast<uint64_t>(hi_bits) << 32) +
+         static_cast<uint64_t>(lo_bits);
+}
+
+uint64_t CalculateThresholdFromBuffer(const std::string trace_id) {
```
gnossen, line 41: `const std::string&`

src/python/grpcio_observability/grpc_observability/sampler.cc
```diff
@@ -0,0 +1,69 @@
...
+  static ProbabilitySampler* sampler = new ProbabilitySampler;
+  return *sampler;
+}
+
+void ProbabilitySampler::SetThreshold(double probability) {
+  uint64_t threshold = CalculateThreshold(probability);
+  threshold_ = threshold;
+}
+
+bool ProbabilitySampler::ShouldSample(const std::string trace_id) {
```
gnossen, line 62: `const std::string&`

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
```diff
@@ -0,0 +1,247 @@
...
+  delete this;
+}
+
+//
+// PythonOpenCensusServerCallTracerFactory
+//
+
+grpc_core::ServerCallTracer*
+PythonOpenCensusServerCallTracerFactory::CreateNewServerCallTracer(
+    grpc_core::Arena* arena) {
```
gnossen, line 241: Probably want an explicit `(void)arena;` to indicate that we're intentionally ignoring this argument.

src/python/grpcio_observability/grpc_observability/observability_util.cc
```diff
@@ -0,0 +1,225 @@
...
+  const char* value = std::getenv("GRPC_PYTHON_CENSUS_EXPORT_THRESHOLD");
+  if (value != nullptr) {
+    return std::stof(value);
+  }
+  return kExportThreshold;
+}
+
+int GetMaxExportBufferSize() {
+  const char* value =
+      std::getenv("GRPC_PYTHON_CENSUS_MAX_EXPORT_BUFFER_SIZE_BYTES");
```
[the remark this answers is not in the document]
gnossen: > Do we need count the current size of CensusData? My understanding is we just need a reasonable maximum size for the buffer.

Right. We don't actually need to calculate the byte footprint. Just limiting the number of `CensusData` instances should be enough. Also, let's pump this up to 10000 as the default starting number. That should be on the order of megabytes. If we go too low, then we'll probably start silently dropping a lot of data when the client starts going at speed.

src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -0,0 +1,399 @@
...
+        self._validate_metrics(self.all_metric)
+        self._validate_spans(self.all_span)
+
+        self.unary_unary_call()
+        self.assertEqual(len(self.all_metric), current_metric_len)
+        self.assertEqual(len(self.all_span), current_spans_len)
+
+    def testTraceSamplingRate(self):
+        # Make 20 UnaryCall's
+        # With 50% sampling rate, we should get 5-15 traces with 98.8% probability
```
gnossen, line 297: I've confirmed this by running the tests manually. Exactly 12 failures out of 1000 runs, all caused by this test. Nicely calculated.

But 98.8% is too low. Since this is within our control, let's aim for 99.9% or higher. We can do that by either pumping up the number of calls made (which increases the test length) and/or by widening the bounds (which could increase the number of false green runs).

If the system is working (with 50% sampling), when we make $`n`$ calls, the probability that $`k`$ calls will be traced, with $`0 \leq k \leq n`$, is $`\frac{n \choose k}{2^n}`$. We can write some code to calculate this for us:

```python
import functools

@functools.cache
def factorial(n):
    if n == 0:
        return 1
    return n * factorial(n - 1)

@functools.cache
def choose(n, k):
    return factorial(n) / (factorial(k) * factorial(n - k))

def probabilty_excluding_tails(variables, tail_length):
    assert tail_length * 2 <= variables
    n = variables
    prob_failure = 0.0
    # Only sum the left tail
    for k in range(tail_length):
        prob_failure += (choose(n, k)) / (2.0 ** n)

    # Right tail is equal in probability
    prob_failure *= 2.0

    return 1.0 - prob_failure
```

This reproduces your 98.8% figure:

```python
>>> prob.probabilty_excluding_tails(20, 5)
0.9881820678710938
```

It looks like doubling the number of RPCs and keeping the bounds the same should be more than sufficient:

```python
>>> prob.probabilty_excluding_tails(40, 5)
0.9999998142975528
```

If you're worried about test duration, then you can run multiple iterations, breaking if the assertions are met. If they're not met, then continue with more RPCs but the same bounds. This will very quickly converge on 100% passing.

src/python/grpcio_observability/grpc_observability/observability_util.cc
```diff
@@ -0,0 +1,224 @@
...
+namespace grpc_observability {
+
+std::queue<CensusData>* g_census_data_buffer;
+std::mutex g_census_data_buffer_mutex;
+std::condition_variable g_census_data_buffer_cv;
+// TODO(xuanwn): Change below to a more appropriate number.
+// Assume buffer will store 100 CensusData and start export when buffer is 70%
+// full.
+constexpr float kExportThreshold = 0.7;
+constexpr int kMaxExportBufferSizeBytes = 100;
```
gnossen, line 44: NIt: Still has "Bytes" in the name.

gnossen:
[The only remaining item](https://github.com/grpc/grpc/pull/32974#discussion_r1207234146) is the docstrings for the protected interface added to the `grpc` module.

XuanWang-Amos:
> [The only remaining item](https://github.com/grpc/grpc/pull/32974#discussion_r1207234146) is the docstrings for the protected interface added to the `grpc` module.

@gnossen Added comments and docstrings, PTALA.

gnossen (approved):
Great work! Remember that this will require a cherry-pick.

PR #32981, opened 2023-05-02
Include Rust in known implementations of server reflection

l-monninger:
This may well be omitted because it's third-party, but hyperium/tonic implements server-reflection--available as the [tonic-reflection](https://crates.io/crates/tonic-reflection) crate.

### DIFF
--- a/doc/server-reflection.md
+++ b/doc/server-reflection.md
@@ -171,5 +171,6 @@ each language:
 - [Go](https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md#enable-server-reflection)
 - [C++](https://grpc.io/grpc/cpp/md_doc_server_reflection_tutorial.html)
 - [Python](https://github.com/grpc/grpc/blob/master/doc/python/server_reflection.md)
+- [Rust](https://github.com/hyperium/tonic/tree/master/tonic-reflection)
 - Ruby: not yet implemented [#2567](https://github.com/grpc/grpc/issues/2567)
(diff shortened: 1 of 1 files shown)


doc/server-reflection.md
```diff
@@ -171,5 +171,6 @@ each language:
 - [Go](https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md#enable-server-reflection)
 - [C++](https://grpc.io/grpc/cpp/md_doc_server_reflection_tutorial.html)
 - [Python](https://github.com/grpc/grpc/blob/master/doc/python/server_reflection.md)
+- [Rust](https://github.com/hyperium/tonic/tree/master/tonic-reflection)
```
yashykt, line 174: This Rust library is not maintained in the grpc repo, so I'm not sure if we should be including it here.

l-monninger: Yeah, on second thought, I agree. It would also be inconsistent as other docs don't refer to `hyperium/grpc`. 

Maybe it's worth noting in the [contribution guidelines](https://github.com/grpc/grpc/blob/master/CONTRIBUTING.md) that only implementations in the `grpc` org should be referenced in the docs, so that nobody else does the same.

yashykt: I gave it a brief thought and it wasn't immediately clear what language to use. I'm not even 100% sold that we should *NOT* mention other implementations. Please feel free to suggest changes to the guidelines, and we would discuss it.

PR #32998, opened 2023-05-03
GenericSerialize avoid byte size recalculation

kitaisreal:
Avoid byte size recalculation in `GenericSerialize` method using `SerializeWithCachedSizesToZeroCopyStream` method instead of `SerializeToZeroCopyStream`.
Method `SerializeToZeroCopyStream`:
```
bool MessageLite::SerializeToZeroCopyStream(
    io::ZeroCopyOutputStream* output) const {
  ABSL_DCHECK(IsInitialized())
      << InitializationErrorMessage("serialize", *this);
  return SerializePartialToZeroCopyStream(output);
}

bool MessageLite::SerializePartialToZeroCopyStream(
    io::ZeroCopyOutputStream* output) const {
  const size_t size = ByteSizeLong();  // Force size to be cached.
  if (size > INT_MAX) {
    ABSL_LOG(ERROR) << GetTypeName()
                    << " exceeded maximum protobuf size of 2GB: " << size;
    return false;
  }
```
But in method `GenericSerialize` we already calculated byte size, so we can add special method `SerializeToZeroCopyStream` to use `GetCachedSize` method.

### DIFF
--- a/include/grpcpp/impl/proto_utils.h
+++ b/include/grpcpp/impl/proto_utils.h
@@ -59,7 +59,7 @@ Status GenericSerialize(const grpc::protobuf::MessageLite& msg, ByteBuffer* bb,
     return grpc::Status::OK;
   }
   ProtoBufferWriter writer(bb, kProtoBufferWriterMaxBufferLength, byte_size);
-  return msg.SerializeToZeroCopyStream(&writer)
+  return msg.SerializeWithCachedSizesToZeroCopyStream(&writer)
              ? grpc::Status::OK
              : Status(StatusCode::INTERNAL, "Failed to serialize message");
 }


kitaisreal:
Requires https://github.com/protocolbuffers/protobuf/pull/12644 to be accepted and merged first.

kitaisreal:
In YDB https://github.com/ydb-platform/ydb, during extensive data load, this recalculation can cost 3 to 10 percents.
Flamegraph before patch:
upload_data_protobuf_before_important
Part with `GenericSerialize`:
upload_data_protobuf_before_generic_serialize
Flamegraph after patch:
upload_data_protobuf_after_2_important
Part with `GenericSerialize`:
upload_data_protobuf_after_generic_serialize

veblush:
Thanks for the suggestion. As it looks like it requires a change on the Protobuf side, I'd like to close it and please feel free to reopen it once Protobuf had the change you suggested.

PR #32999, opened 2023-05-03
[chaotic-good] New frame serialization/deserialization in chaotic-good transport.

nanahpang:
Implement the frame serialization/deserialization method in chaotic-good transport.

### DIFF
--- a/test/core/transport/binder/end2end/end2end_binder_transport_test.cc
+++ b/test/core/transport/binder/end2end/end2end_binder_transport_test.cc
@@ -50,6 +50,7 @@ class End2EndBinderTransportTest
   ~End2EndBinderTransportTest() override {
     server_->Shutdown();
     service_.reset();
+    exec_ctx.Flush();
     delete end2end_testing::g_transaction_processor;
   }
 


src/core/ext/transport/chaotic_good/frame.cc
```diff
@@ -86,32 +115,35 @@ class FrameSerializer {
...
       : header_(header), input_(input) {}
   const FrameHeader& header() const { return header_; }
-  // If called, must be called before ReceiveMessage, ReceiveTrailers
+  // If called, must be called before ReceiveMessage, ReceivePadding,
+  // ReceiveTrailers, Finish.
   absl::StatusOr<SliceBuffer> ReceiveHeaders() {
     return Take(header_.header_length);
   }
-  // If called, must be called before ReceiveTrailers
+  // If called, must be called before ReceivePadding, ReceiveTrailers, Finish.
```
ctiller, line 138: Since messages are not part of the framing system anymore, I think we should remove ReceiveMessage (and therefore ReceivePadding) from this type.

ctiller: (instead we should add some helper functions to get the message lengths)

src/core/ext/transport/chaotic_good/frame_header.h
```diff
@@ -45,11 +47,11 @@ struct FrameSizes {
 };
 
 struct FrameHeader {
-  FrameType type;
-  BitSet<3> flags;
+  uint32_t type_and_flags;  // lower 8 bits specify the frame type.
```
ctiller, line 50: This approach will cause all frame manipulation code to know about this serialization detail, rather than just the code that's serializing it - I think it would be better to keep the type, flags separation (even if we need to change the flags representation)

ctiller (commented):
Some quick notes... I'll go through it more thoroughly soon.

nanahpang:
Oh, it seems like force push on my local branch will close the PR, :) I will open another PR with your comments. Sorry about that.

PR #33002, opened 2023-05-03
[client channel] use ChunkedVector for call attributes

markdroth:
Change call attributes to be stored in a `ChunkedVector` instead of `std::map<>`, so that the storage can be allocated on the arena.  This means that we're now doing a linear search instead of a map lookup, but the total number of attributes is expected to be low enough that that should be okay.

Also, we now hide the actual data structure inside of the `ServiceConfigCallData` object, which required some changes to the `ConfigSelector` API.  Previously, the `ConfigSelector` would return a `CallConfig` struct, and the client channel would then use the data in that struct to populate the `ServiceConfigCallData`.  This PR changes that such that the client channel creates the `ServiceConfigCallData` before invoking the `ConfigSelector`, and it passes the `ServiceConfigCallData` into the `ConfigSelector` so that the `ConfigSelector` can populate it directly.


src/core/lib/service_config/service_config_call_data.h
```diff
@@ -44,45 +49,61 @@ class ServiceConfigCallData {
...
+  static void Destroy(void* ptr) {
+    auto* self = static_cast<ServiceConfigCallData*>(ptr);
+    self->~ServiceConfigCallData();
+  }
+
   RefCountedPtr<ServiceConfig> service_config_;
-  const ServiceConfigParser::ParsedConfigVector* method_configs_;
-  CallAttributes call_attributes_;
+  const ServiceConfigParser::ParsedConfigVector* method_configs_ = nullptr;
+  ChunkedVector<CallAttributeInterface*, 10> call_attributes_;
```
ctiller, line 106: 10 means if we have one attribute we allocate 80 bytes here... given that we don't have a lot of these, suggest maybe 3 or 4?

ctiller (approved):

PR #33004, opened 2023-05-03
[xDS] Protect RBAC audit logging options field with environment variable

rockspore:
The protection is added at `xds_http_rbac_filter.cc` where we read the new field. With this disabling the feature, nothing from things like `xds_audit_logger_registry.cc` shall be invoked.


test/core/xds/xds_http_filters_test.cc
```diff
@@ -896,6 +897,7 @@ TEST_P(XdsRbacFilterConfigTest, AllPrincipalTypes) {
 }
 
 TEST_P(XdsRbacFilterConfigTest, AuditLoggingOptions) {
```
markdroth, line 899: Please also add a test that if the env var is *not* set, we do not use the new fields even if they are set.

rockspore: Done. Thanks!

markdroth (approved):
Looks great!  Thanks for doing that so quickly!

PR #33006, opened 2023-05-03
[Build] Removed Makefile

veblush:
Let's test this.


veblush:
Ruby and PHP still need `Makefile` to generate artifacts.

PR #33026, opened 2023-05-05
[Audit Logging] Stdout logger implementation

rockspore:
The logger uses `absl::FPrintF` to write to stdout. After reading a number of sources online, I got the impression that `std::fwrite` which is used by `absl::FPrintF` is atomic so there is no locking required here.


src/core/lib/security/authorization/stdout_logger.cc
```diff
@@ -0,0 +1,75 @@
...
+
+#include "src/core/lib/json/json.h"
+
+namespace grpc_core {
+namespace experimental {
+
+constexpr absl::string_view kName = "stdout_logger";
+const char kLogFormat[] =
+    "{\"grpc_audit_log\":{\"timestamp\":%d,\"rpc_method\":\"%s\",\"principal\":"
+    "\"%s\",\"policy_name\":\"%s\",\"matched_rule\":\"%s\",\"authorized\":%s}}";
```
gtcooke94, line 42: Will this have implicit newlines, and if so do we want a multi-line log?

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +82,82 @@ class AuditLoggingTest : public ::testing::Test {
...
-TEST_F(AuditLoggingTest, SuccessfulLoggerCreation) {
+//
+// AuditLoggerRegistryTest
+//
+
+TEST_F(AuditLoggerRegistryTest, SuccessfulLoggerCreation) {
   auto result = AuditLoggerRegistry::ParseConfig(kName, Json());
   ASSERT_TRUE(result.ok());
   ASSERT_NE(AuditLoggerRegistry::CreateAuditLogger(std::move(result.value())),
             nullptr);
```
gtcooke94, line 93: I know this test wasn't added in this PR, but I noticed that this functionally tests
1. `ParseConfig` returns something with `result.ok()`
2. `CreateAuditLogger` returns something that is not `nullptr`

Are there additional checks that would be worthwhile here? I know other tests cover more detail, but if they do is this really even worth having?

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +82,82 @@ class AuditLoggingTest : public ::testing::Test {
...
 }
 
-TEST_F(AuditLoggingTest, AuditLoggerFactoryExistenceChecks) {
+TEST_F(AuditLoggerRegistryTest, LoggerFactoryExistenceChecks) {
   EXPECT_TRUE(AuditLoggerRegistry::FactoryExists(kName));
   EXPECT_FALSE(AuditLoggerRegistry::FactoryExists("unknown_logger"));
 }
 
+//
+//  StdoutLoggerTest
```
gtcooke94, line 110: Would it be worth having StdOutLoggerTest in it's own file?

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +82,82 @@ class AuditLoggingTest : public ::testing::Test {
...
+//
+
+TEST(StdoutLoggerTest, LoggerFactoryExistenceChecks) {
+  EXPECT_TRUE(AuditLoggerRegistry::FactoryExists("stdout_logger"));
+}
+
+TEST(StdoutLoggerTest, BadLoggerConfig) {
+  auto result =
+      AuditLoggerRegistry::ParseConfig("stdout_logger", Json::FromBool(false));
```
gtcooke94, line 119: I think there's some value in making this more explicit
```suggestion
auto invalid_value = Json::FromBool(false)
auto result = AuditLoggerRegistry::ParseConfig("stdout_logger", invalid_value);
```

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +82,82 @@ class AuditLoggingTest : public ::testing::Test {
...
+  auto result =
+      AuditLoggerRegistry::ParseConfig("stdout_logger", Json::FromObject({}));
+  ASSERT_TRUE(result.ok());
+  auto logger =
+      AuditLoggerRegistry::CreateAuditLogger(std::move(result.value()));
+  AuditContext context("method", "spiffe", "policy", "rule", true);
+  ::testing::internal::CaptureStdout();
+  logger->Log(context);
+  auto output = ::testing::internal::GetCapturedStdout();
+  std::cout << output << '\n';
```
gtcooke94, line 135: What's the purpose of this line? It seems like we are capturing the stdout and keeping it in `output`, but then writing it back out?

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +82,82 @@ class AuditLoggingTest : public ::testing::Test {
...
+  auto log_or = JsonParse(output);
+  ASSERT_TRUE(log_or.ok());
+  ASSERT_EQ(log_or.value().type(), Json::Type::kObject);
+  auto it = log_or.value().object().find("grpc_audit_log");
+  ASSERT_NE(it, log_or.value().object().end());
+  ASSERT_EQ(it->second.type(), Json::Type::kObject);
+  auto object = it->second.object();
+  ASSERT_NE(object.find("timestamp"), object.end());
+  EXPECT_EQ(object.find("timestamp")->second.type(), Json::Type::kNumber);
```
gtcooke94, line 144: Is there a way we could factor the code such that the timestamp is also an input somewhere and we could test that we are correctly outputting it with the expected format, etc?

gtcooke94 (commented):
Looks good, a few comments

src/core/lib/security/authorization/stdout_logger.cc
rockspore: Good catch! I meant to add a newline to flush the buffer but forgot. Added.

test/core/security/grpc_audit_logging_test.cc
rockspore: This test is to check third-party logger factory registration and logger creation. It's somewhat different from stdout logger that we register in the ctor of the registry. So I think it's still needed. I also don't think there is anything else needed since we don't test the `TestAuditLoggerFactory` itself.

test/core/security/grpc_audit_logging_test.cc
rockspore: The stdout logger is added to the same library target in the BUILD file. I think it's easier to have things in the same test target.

test/core/security/grpc_audit_logging_test.cc
rockspore: Sure. Done.

test/core/security/grpc_audit_logging_test.cc
rockspore: Sorry. This was useless stuff that I forgot to revert after using it locally. Thanks for the catch.

test/core/security/grpc_audit_logging_test.cc
rockspore: Added a check to verify the timestamp is very close to now. pTAL.

src/core/lib/security/authorization/stdout_logger.cc
gtcooke94: I actually meant the other way around, do we want _any_ newlines here at all, or one big single line that is the audit log?

gtcooke94 (commented):
This LGTM other than the question about multi-line vs. single-line logs

src/core/lib/security/authorization/stdout_logger.cc
rockspore: As you also saw in the Go PR, we want one log entry to be one-liner.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +85,87 @@ class AuditLoggingTest : public ::testing::Test {
...
+  ASSERT_EQ(it->second.type(), Json::Type::kObject);
+  auto object = it->second.object();
+  // Check if the recorded timestamp is close to now.
+  absl::Time t1 = absl::Now();
+  ASSERT_NE(object.find("timestamp"), object.end());
+  EXPECT_EQ(object.find("timestamp")->second.type(), Json::Type::kNumber);
+  uint64_t ts;
+  ASSERT_TRUE(absl::SimpleAtoi(object.find("timestamp")->second.string(), &ts));
+  absl::Time t2 = absl::FromUnixSeconds(ts);
+  EXPECT_TRUE(t1 - t2 < absl::Seconds(10));
```
markdroth, line 152: Rather than hard-coding this 10-second tolerance, I suggest storing the time before we call `logger->Log()`, then storing the time after that call returns, and then verifying that the logged timestamp is between those two values.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +85,87 @@ class AuditLoggingTest : public ::testing::Test {
...
+  // Check exact values of everything else.
```
markdroth, line 153: This check might be a bit more readable if you construct a new JSON object without the timestamp field and then write it to a string, which you can then compare to an expected value.

markdroth (commented):
This looks good overall!  Comments are relatively minor.

src/core/lib/security/authorization/stdout_logger.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "absl/time/time.h"
+
+#include <grpc/grpc_audit_logging.h>
+#include <grpc/support/json.h>
+#include <grpc/support/log.h>
+
+namespace grpc_core {
+namespace experimental {
+
+constexpr absl::string_view kName = "stdout_logger";
```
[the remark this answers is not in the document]
rockspore: Fixed.

src/core/lib/security/authorization/stdout_logger.cc
```diff
@@ -0,0 +1,75 @@
...
+constexpr char kLogFormat[] =
+    "{\"grpc_audit_log\":{\"timestamp\":%d,\"rpc_method\":\"%s\",\"principal\":"
+    "\"%s\",\"policy_name\":\"%s\",\"matched_rule\":\"%s\",\"authorized\":%s}}"
+    "\n";
+
+void StdoutAuditLogger::Log(const AuditContext& context) {
+  absl::FPrintF(stdout, kLogFormat, absl::ToUnixSeconds(absl::Now()),
```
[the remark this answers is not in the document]
rockspore: With some offline discussion, I tentatively changed the timestamp format to RFC3339 (i.e., `2023-05-12T22:02:07.642583807+00:00`). I will confirm with @ejona86 to finalize the accepted format.

For now, given the cost from parsing and formatting the time, I feel `absl::Now` is acceptable. Let me know if that's not correct.

src/core/lib/security/authorization/stdout_logger.cc
```diff
@@ -0,0 +1,75 @@
...
+  return kName;
+}
+
+std::string StdoutAuditLoggerFactory::Config::ToString() const { return "{}"; }
+
+absl::string_view StdoutAuditLoggerFactory::name() const { return kName; }
+
+absl::StatusOr<std::unique_ptr<AuditLoggerFactory::Config>>
+StdoutAuditLoggerFactory::ParseAuditLoggerConfig(const Json& json) {
+  if (json.type() != Json::Type::kObject) {
```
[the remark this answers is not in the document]
rockspore: It is already enforced (https://github.com/grpc/grpc/blob/ee0cf2fada30dc40ac56f41d69e01fac031ea1ee/src/core/lib/security/authorization/rbac_translator.cc#L402). Likewise, we ensure the proto from xDS gets converted appropriately. In this case, I've removed this check here as well as the unit test for it.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +85,87 @@ class AuditLoggingTest : public ::testing::Test {
...
+      AuditLoggerRegistry::ParseConfig("stdout_logger", Json::FromObject({}));
+  ASSERT_TRUE(result.ok());
+  auto logger =
+      AuditLoggerRegistry::CreateAuditLogger(std::move(result.value()));
+  AuditContext context("method", "spiffe", "policy", "rule", true);
+  ::testing::internal::CaptureStdout();
+  logger->Log(context);
+  auto log_or = JsonParse(::testing::internal::GetCapturedStdout());
+  ASSERT_TRUE(log_or.ok());
+  ASSERT_EQ(log_or.value().type(), Json::Type::kObject);
```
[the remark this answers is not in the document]
rockspore: Fixed. Thanks for the tip!

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -81,26 +85,87 @@ class AuditLoggingTest : public ::testing::Test {
...
+  auto it = log_or.value().object().find("grpc_audit_log");
+  ASSERT_NE(it, log_or.value().object().end());
+  ASSERT_EQ(it->second.type(), Json::Type::kObject);
+  auto object = it->second.object();
```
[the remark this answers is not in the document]
rockspore: Done.

test/core/security/grpc_audit_logging_test.cc
rockspore: Done.

test/core/security/grpc_audit_logging_test.cc
rockspore: Done.

rockspore (commented):
Thanks for the review. I addressed the comments as I can for now. Will ask for another look once the timestamp format is finalized.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -134,36 +126,38 @@ TEST(StdoutLoggerTest, StdoutLoggerCreationAndLogInvocation) {
...
-  EXPECT_TRUE(t1 - t2 < absl::Seconds(10));
+  EXPECT_EQ(object.find("timestamp")->second.type(), Json::Type::kString);
+  absl::Time log_time;
+  ASSERT_TRUE(absl::ParseTime(absl::RFC3339_sec,
+                              object.find("timestamp")->second.string(),
+                              &log_time, nullptr));
+  int64_t time_at_log = absl::ToUnixSeconds(log_time);
+  // Check if the recorded timestamp is in between the recorded interval with
+  // the precision of one second.
+  EXPECT_TRUE(time_at_log >= time_before_log && time_at_log <= time_after_log);
```
markdroth, line 148: If this assertion fails, the message will just say "false != true", which isn't very easy to debug.  Instead, I suggest writing this in a way that will provide much clearer output when the test fails:

```
EXPECT_GT(time_at_log, time_before_log);
EXPECT_LT(time_at_log, time_after_log);
```

Note that we probably want to use `>` and `<` instead of `>=` and `<=`, since it should be strictly between those two times.

test/core/security/grpc_audit_logging_test.cc
```diff
@@ -134,36 +126,38 @@ TEST(StdoutLoggerTest, StdoutLoggerCreationAndLogInvocation) {
...
   // Check exact values of everything else.
-  std::map<std::string, std::string> fields = {{"rpc_method", "method"},
-                                               {"principal", "spiffe"},
-                                               {"policy_name", "policy"},
-                                               {"matched_rule", "rule"}};
-  for (const auto& p : fields) {
-    auto it = object.find(p.first);
+  std::vector<std::string> fields = {"rpc_method", "principal", "policy_name",
```
markdroth, line 150: This should use `absl::string_view` instead of `std::string` to avoid unnecessary copying.

But actually, better yet would be to just exclude the one field that we *don't* want to check, so that we don't need to remember to add field names here if we ever add new fields to the output format:

```
Json::Object json_object = object;
json_object.erase("timestamp");
```

markdroth (approved):
Looks great!  Just a couple of minor nits remaining.  Feel free to merge after addressing.

test/core/security/grpc_audit_logging_test.cc
rockspore: Thanks for the tip! Will change it. We do need `>=` and `<=` since I am only taking seconds and everything happens so fast that they are often in the same second. Let me know if you prefer nanosecond.

markdroth: I think we definitely need more than second-level precision here.  And the output format needs to specify exactly what the precision should be, since it needs to be consistent cross-language.

markdroth (approved):
Looks great!

rockspore:
In one recent [build](https://source.cloud.google.com/results/invocations/f40b20c6-dcce-4b56-ab16-a70342b9110a/targets/%2F%2Ftest%2Fcore%2Fsecurity:grpc_audit_logging_test/tests), the invocation of `Log` happened so soon that `time_before_log` was actually equal to `time_at_log` even with nanosecond:
```
test/core/security/grpc_audit_logging_test.cc:146
Expected: (time_at_log) > (time_before_log), actual: 2023-05-16T00:45:04.663233+00:00 vs 2023-05-16T00:45:04.663233+00:00
```

I changed it back to `EXPECT_GE` and `EXPECT_LE` to avoid flakiness in the future.

PR #32924, opened 2023-04-22
[ObjC] CF Stream Event Engine Client

HannahShiSFB:
bazel build --config=macos --genrule_strategy=local --copt="-DGRPC_CFSTREAM=1" //test/cpp/end2end:cfstream_test
succeeds

Fixing failure described here: https://github.com/grpc/grpc/pull/32882#issuecomment-1512210309

PR #33027, opened 2023-05-05
Revert "[ObjC] CF Stream Event Engine Client"

drfloob:
Reverts grpc/grpc#32924. This breaks the build again, unfortunately.

From `test/core/event_engine/cf:cf_engine_test`:

```
error: module .../grpc/test/core/event_engine/cf:cf_engine_test does not depend on a module exporting 'grpc/support/port_platform.h'
```

@sampajano I recommend looking into CI tests to catch iOS problems before merging. We can enable EventEngine experiments in the CI generally once this PR lands, but this broken test is not one of those experiments. A normal build should have caught this.

cc @HannahShiSFB


yijiem (approved):

sampajano:
> Reverts #32924. This breaks the build again, unfortunately.
> 
> From `test/core/event_engine/cf:cf_engine_test`:
> 
> ```
> error: module .../grpc/test/core/event_engine/cf:cf_engine_test does not depend on a module exporting 'grpc/support/port_platform.h'
> ```
> 
> @sampajano I recommend looking into CI tests to catch iOS problems before merging. We can enable EventEngine experiments in the CI generally once this PR lands, but this broken test is not one of those experiments. A normal build should have caught this.
> 
> cc @HannahShiSFB

Oh sorry that this has failed again.. yes.. will look into CI issue before merging again!!

PR #33030, opened 2023-05-05
[EventEngine] Change TXT lookup result type to std::vector<std::string>

yijiem:
One TXT lookup query can return multiple TXT records (see the following example). `EventEngine::DNSResolver` should return all of them to let the caller (e.g. `event_engine_client_channel_resolver`) decide which one they would use.

```
$ dig TXT wikipedia.org

; <<>> DiG 9.18.12-1+build1-Debian <<>> TXT wikipedia.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 49626
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;wikipedia.org.                 IN      TXT

;; ANSWER SECTION:
wikipedia.org.          600     IN      TXT     "google-site-verification=AMHkgs-4ViEvIJf5znZle-BSE2EPNFqM1nDJGRyn2qk"
wikipedia.org.          600     IN      TXT     "yandex-verification: 35c08d23099dc863"
wikipedia.org.          600     IN      TXT     "v=spf1 include:wikimedia.org ~all"
```

Note that this change also deviates us from the iomgr's DNSResolver API which uses std::string as the result type.

### DIFF
--- a/src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
+++ b/src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
@@ -31,6 +31,7 @@
 #include "absl/container/flat_hash_set.h"
 #include "absl/status/status.h"
 #include "absl/status/statusor.h"
+#include "absl/strings/match.h"
 #include "absl/strings/str_cat.h"
 #include "absl/strings/strip.h"
 #include "absl/types/optional.h"
@@ -128,7 +129,7 @@ class EventEngineClientChannelDNSResolver : public PollingResolver {
     void OnBalancerHostnamesResolved(
         std::string authority,
         absl::StatusOr<std::vector<EventEngine::ResolvedAddress>> addresses);
-    void OnTXTResolved(absl::StatusOr<std::string> service_config);
+    void OnTXTResolved(absl::StatusOr<std::vector<std::string>> service_config);
     // Returns a Result if resolution is complete.
     // callers must release the lock and call OnRequestComplete if a Result is
     // returned. This is because OnRequestComplete may Orphan the resolver,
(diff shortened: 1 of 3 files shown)


yijiem:
This would need be cherrypicked.

drfloob (approved):

src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
```diff
@@ -400,7 +401,21 @@ void EventEngineClientChannelDNSResolver::EventEngineDNSRequestWrapper::
...
       service_config_json_ = service_config.status();
     } else {
-      service_config_json_ = absl::StrCat("grpc_config=", *service_config);
+      constexpr char kServiceConfigAttributePrefix[] = "grpc_config=";
+      auto result = std::find_if(service_config->begin(), service_config->end(),
+                                 [&](absl::string_view s) {
+                                   return absl::StartsWith(
+                                       s, kServiceConfigAttributePrefix);
+                                 });
+      if (result != service_config->end()) {
```
Vignesh2208, line 410: This appears to differ from previous logic where we simply concatenate the grpc_config prefix to all returned values. Instead now it searches for a result with the given prefix. Is there a specific reason ?

drfloob:
Thanks for finding & fixing this API problem, Yijie! Please make sure it gets added to the release notes, since it's a public API.

src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
yijiem: yeah, I think the previous logic is not correct (this code is still under development and not used until EventEngine::DNSResolver is ready) since gRPC's service config data starts with `grpc_config=` followed by the JSON-formatted data. See: https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md#encoding-in-dns-txt-records.

PR #32701, opened 2023-03-24
[EventEngine] PosixEventEngine DNS Resolver

yijiem:
This PR implements a c-ares based DNS resolver for EventEngine with the reference from the original [grpc_ares_wrapper.h](../blob/master/src/core/ext/filters/client_channel/resolver/dns/c_ares/grpc_ares_wrapper.h). The PosixEventEngine DNSResolver is implemented on top of that. Tests which use the client channel resolver API ([resolver.h](../blob/master/src/core/lib/resolver/resolver.h#L54)) are ported, namely the [resolver_component_test.cc](../blob/master/test/cpp/naming/resolver_component_test.cc) and the [cancel_ares_query_test.cc](../blob/master/test/cpp/naming/cancel_ares_query_test.cc). The WindowsEventEngine DNSResolver will use the same EventEngine's grpc_ares_wrapper and will be worked on next.

The [resolve_address_test.cc](https://github.com/grpc/grpc/blob/master/test/core/iomgr/resolve_address_test.cc) which uses the iomgr [DNSResolver](../blob/master/src/core/lib/iomgr/resolve_address.h#L44) API has been ported to EventEngine's dns_test.cc. That leaves only 2 tests which use iomgr's API, notably the [dns_resolver_cooldown_test.cc](../blob/master/test/core/client_channel/resolvers/dns_resolver_cooldown_test.cc) and the [goaway_server_test.cc](../blob/master/test/core/end2end/goaway_server_test.cc) which probably need to be restructured to use EventEngine DNSResolver (for one thing they override the original grpc_ares_wrapper's free functions). I will try to tackle these in the next step.

PR #33050, opened 2023-05-09
[Tools] Account for header files from the srcs attribute of cc_library in fix_build_deps.py

yijiem:
Spin off from https://github.com/grpc/grpc/pull/32701. Header files can come from the `srcs` attribute of `cc_library` which makes them private to the library in terms of bazel dependency check. That seems to confuse fix_build_deps.py before this change.

### DIFF
--- a/tools/distrib/fix_build_deps.py
+++ b/tools/distrib/fix_build_deps.py
@@ -343,6 +343,10 @@ def grpc_cc_library(name,
 
     for hdr in hdrs + public_hdrs:
         vendors[_get_filename(hdr, parsing_path)].append(name)
+    for src in srcs:
+        if src.endswith('.h'):
+            # headers can come from srcs as well
+            vendors[_get_filename(src, parsing_path)].append(name)
     inc = set()
     original_deps[name] = frozenset(deps)
     original_external_deps[name] = frozenset(external_deps)


drfloob:
Interesting. Are there many targets using this private header pattern? And what problems were you seeing before this fix, what does a confused fix_build_deps look like?

yijiem:
> Interesting. Are there many targets using this private header pattern? And what problems were you seeing before this fix, what does a confused fix_build_deps look like?

I'm using this pattern in https://github.com/grpc/grpc/pull/32701 for a header file (`src/core/lib/event_engine/nameser.h`) since it's not used in other dependent libs. fix_build_deps would err on this line:
https://github.com/grpc/grpc/blob/master/tools/distrib/fix_build_deps.py#L670-L671, e.g.
```
# ERROR: can't categorize header: src/core/lib/event_engine/nameser.h used by //src/core:grpc_ares_wrapper
```

drfloob:
That is interesting. It seems like a fine pattern to separate concerns. As an alternative, since the code is only used in one place, you could move the nameser.h code into an anonymous namespace inside the CC file.

@ctiller given your interest in bazel target organization, do you have any thoughts or guiding principles against this usage pattern?

yijiem:
> That is interesting. It seems like a fine pattern to separate concerns. As an alternative, since the code is only used in one place, you could move the nameser.h code into an anonymous namespace inside the CC file.
> 
> @ctiller given your interest in bazel target organization, do you have any thoughts or guiding principles against this usage pattern?

FWIW https://bazel.build/reference/be/c-cpp#hdrs

ctiller (approved):

PR #33078, opened 2023-05-11
[Misc] Add necessary things for privacy

ananda1066:


include/grpc/impl/grpc_types.h
```diff
@@ -467,6 +467,8 @@ typedef struct {
 /** If set, uses a local subchannel pool within the channel. Otherwise, uses the
  * global subchannel pool. */
 #define GRPC_ARG_USE_LOCAL_SUBCHANNEL_POOL "grpc.use_local_subchannel_pool"
+/** If set, privacy. */
```
ctiller, line 470: Let's omit from the public API and just include it in a header we can visibility protect?

src/core/lib/experiments/experiments.yaml
```diff
@@ -96,13 +96,13 @@
     If set, use the new gRPC promise based call code when it's appropriate
     (ie when all filters in a stack are promise based)
   default: false
-  expiry: 2023/03/01
+  expiry: 2023/06/01
```
ctiller, line 99: Thankyou

include/grpc/impl/grpc_types.h
ananda1066: removed arg.

ctiller (approved):

PR #33089, opened 2023-05-11
Revert "Revert "[reflection] python: reflection response returns dependencies""

XuanWang-Amos:
Reverts grpc/grpc#33088

The internal test is failing because after the change, transitive dependencies will be included in response too, thus we need add `empty2_pb2` to response since it's a dependency of `empty2_extensions_pb2`.

We already did that in OSS test, we need do the same for internal test case.


tomsoftware:
Is there something I can do?

XuanWang-Amos:
> Is there something I can do?

No need, we'll take it from here :)

gnossen (approved):

PR #33090, opened 2023-05-12
[PSM Interop] URL Map graceful teardown

sergiitk:
Similar to what we already do in other test suites:

- Try cleaning up resources three times.
- If unsuccessful, don't fail the test and just log the error. The cleanup script should be the one to deal with this.

ref b/282081851


sergiitk:
Test run:
- [ ] [grpc/core/v1.54.x/branch/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/ffeb770b-7281-4c4e-803d-7a77765a4001)

murgatroid99 (approved):

PR #33091, opened 2023-05-12
[ObjC] Upgrade podspec xx.deployment_target

HannahShiSFB:
Upgrade apple platform deployment_target versions to fix the cocoapods push of BoringSSL-GRPC about the following error:
```
ld: file not found: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_macosx.a
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
ref: https://developer.apple.com/forums/thread/725300

This also aligns with the versions required by [protobuf](https://github.com/protocolbuffers/protobuf/pull/10652)

```
ios.deployment_target = '10.0'
osx.deployment_target = '10.12'
tvos.deployment_target = '12.0'
watchos.deployment_target = '6.0'
```

### DIFF
--- a/examples/cpp/helloworld/cocoapods/Podfile
+++ b/examples/cpp/helloworld/cocoapods/Podfile
@@ -1,5 +1,5 @@
 source 'https://github.com/CocoaPods/Specs.git'
-platform :ios, '8.0'
+platform :ios, '10.0'
 
 install! 'cocoapods', :deterministic_uuids => false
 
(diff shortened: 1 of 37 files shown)


sampajano (approved):
Wow.. that's a lot of podspecs that we have..... 😂 

Thanks so much for figuring this out, Hannah.. Great investigative work... 😄

@mikefan1991 FYI

sampajano:
@HannahShiSFB can you fix the "Mergeable" CI failure too? I think you can modify some labels and it should work :)

PR #33098, opened 2023-05-12
[Testing] increase orca oob test timeout

YifeiZhuang:
each oob test runs at least 2s (2 verifications with each server sending report each 1s), current configuration allows at most 3 clients run in parallel. In reality, probably only the two clients against one server.  See past results successful cases:
https://testgrid.corp.google.com/Interop%20Tests#Interop%20Tests%20cloud-to-cloud%20Linux

Currently 8 jobs are running each batch, and it happens that each batch has at most 2 oob clients against one server.
Since 3 languages support orca, so we should allow 3 clients running against one server so the timeout is at least 4s. We give some buffers to allow tests running more reliably.

10s in theory support 5-6 clients running against one server.

cc. @markb74 @dfawley @ejona86

### DIFF
--- a/doc/interop-test-descriptions.md
+++ b/doc/interop-test-descriptions.md
@@ -1152,7 +1152,8 @@ Procedures:
       }
     }
     ```
-2. After getting a response, client waits up to 5 seconds to receive a OOB load 
+2. After getting a response, client waits up to 10 seconds (or a total of 30s for the
+entire test case) to receive an OOB load
 report that matches the requested load report in step 1. To wait for load 
 report, client may inject a callback to the custom LB policy, or poll the result
 by doing empty unary call that carries a reference, e.g. using 
@@ -1173,7 +1174,8 @@ CallOptions, that will be filled in by the custom LB policy as part of the
       }
     }
     ```
-4. After getting a response, client waits up to 5 seconds to receive a OOB load
+4. After getting a response, client waits up to 10 seconds (or a total of 30s for the entire
+test case) to receive an OOB load
 report that matches the requested load report in step 3. Similar to step 2.
 5. Client half closes the stream, and asserts the streaming call is successful. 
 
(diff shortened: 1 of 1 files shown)


doc/interop-test-descriptions.md
```diff
@@ -1152,7 +1152,7 @@ Procedures:
       }
     }
     ```
-2. After getting a response, client waits up to 5 seconds to receive a OOB load 
+2. After getting a response, client waits up to 10 seconds to receive a OOB load
```
dfawley, line 1155: How do you feel about the Go implementation using a 30 second timeout for the entire process?

YifeiZhuang: what does entire process mean?
how about this single verification timeout?

YifeiZhuang: okay i see, that sounds fine.

dfawley: This is for all of this procedure (steps 1-5).  The verification of any step can take any or all of that total 30 seconds.

doc/interop-test-descriptions.md
```diff
@@ -1152,7 +1152,8 @@ Procedures:
       }
     }
     ```
-2. After getting a response, client waits up to 5 seconds to receive a OOB load 
+2. After getting a response, client waits up to 10 seconds (or a total of 30s for the
```
dfawley, line 1155: Text LGTM; please rewrap the lines though.

doc/interop-test-descriptions.md
YifeiZhuang: Using a total timeout might make the debug harder if test failed, while a separate timeout can pin point the exact failed step. Java does not have other tests using overall timeout as Go does. Since it is clear enough that we are pulling for certain condition to be true, either is acceptable for me.

doc/interop-test-descriptions.md
YifeiZhuang: what is the line length requirement? yapf is not complaining.

dfawley (approved):

ejona86 (commented):
LGTM

PR #33124, opened 2023-05-15
[PHP] Fix path munging when package is absent

### LINKED ISSUE #33114: grpc_php_plugin does not respect CodeGeneratorResponse specification when no package is specified in a Protobuf file

gRPC version: v1.54.2

MacOS Ventura 13.3.1

Not using any particular runtime - this is an issue with the `grpc_php_plugin` itself.

### What did you do?

The `grpc_php_plugin` produces invalid values for `CodeGeneratorResponse.File.name` when a `.proto` file has no package specified. The spec https://github.com/protocolbuffers/protobuf/blob/95e6c5b4746dd7474d540ce4fb375e3f79a086f8/src/google/protobuf/compiler/plugin.proto#L122 defines that a path in `name` must be relative and not contain any `.` or `..` paths. When there is a package in the `.proto` file, i.e. `package foo.bar;`, this is respected, and `grpc_php_plugin` produces i.e. `Foo/Bar/GreeterClient.php`, however when there is no package (as in the quick start), the `grpc_php_plugin` produces `/GreeterClient.php`, which is invalid per the `CodeGeneratorResponse` spec, as it is an absolute path.

A reproduction:

1. Make a temporary directory and `cd` into it for future steps:

```bash
mkdir tmp
cd ./tmp
```

2. Create the file from https://grpc.io/docs/languages/php/quickstart:

```bash
cat <<EOF >helloworld.proto
syntax = "proto3";

// note no package

service Greeter {
  rpc SayHello (HelloRequest) returns (HelloReply) {}
}

message HelloRequest {
  string name = 1;
}

message HelloReply {
  string message = 1;
}
EOF
```

3. Create a quick proxy plugin that prints out the file names from the `CodeGeneratorResponse`:

```bash
go mod init repro
mkdir -p cmd/protoc-gen-repro
cat <<EOF >cmd/protoc-gen-repro/main.go
package main

import (
	"bytes"
	"fmt"
	"io/ioutil"
	"os"
	"os/exec"

	"google.golang.org/protobuf/proto"
	"google.golang.org/protobuf/types/pluginpb"
)

func main() {
	if err := run(); err != nil {
		fmt.Fprintln(os.Stderr, err.Error())
		os.Exit(1)
	}
}

func run() error {
	stdinData, err := ioutil.ReadAll(os.Stdin)
	if err != nil {
		return err
	}
	stdinBuffer := bytes.NewBuffer(stdinData)
	stdoutBuffer := bytes.NewBuffer(nil)
	cmd := exec.Command("grpc_php_plugin")
	cmd.Stdin = stdinBuffer
	cmd.Stdout = stdoutBuffer
	if err := cmd.Run(); err != nil {
		return err
	}
	stdoutData := stdoutBuffer.Bytes()
	codeGeneratorResponse := &pluginpb.CodeGeneratorResponse{}
	if err := proto.Unmarshal(stdoutData, codeGeneratorResponse); err != nil {
		return err
	}
	for _, file := range codeGeneratorResponse.File {
		fmt.Fprintf(os.Stderr, "FILE: %q\n", file.GetName())
	}
	_, err = os.Stdout.Write(stdoutData)
	return err
}
EOF
go mod tidy
go install ./cmd/protoc-gen-repro
```

4. Run the proxy plugin:

```bash
mkdir -p gen
protoc -I . --repro_out=gen helloworld.proto 
```

This should result in the following  being printed to stderr:

```
FILE: "GreeterClient.php"
```

### What did you see instead?

```
FILE: "/GreeterClient.php"
```

Note, however, if I add a package to `helloworld.proto`, then this is fine. If `package foo.bar;` is added, you get:

```
FILE: "Foo/Bar/GreeterClient.php"
```

That should be all you need. This should be an easy fix.

fowles:
fixes #33114

### DIFF
--- a/src/compiler/php_generator_helpers.h
+++ b/src/compiler/php_generator_helpers.h
@@ -60,8 +60,10 @@ inline std::string GetPHPServiceFilename(
           << grpc_generator::CapitalizeFirstLetter(tokens[i]);
     }
   }
-  return oss.str() + "/" +
-         GetPHPServiceClassname(service, class_suffix, is_server) + ".php";
+  std::string path = oss.str();
+  if (!path.empty()) path += "/";
+  path += GetPHPServiceClassname(service, class_suffix, is_server) + ".php";
+  return path;
 }
 
 // Get leading or trailing comments in a string. Comment lines start with "// ".


fowles:
this should be for PHP but I can't modify the label

alto-php (approved):

PR #33126, opened 2023-05-15
[PHP] Fix path munging when package is absent

veblush:
Cloning https://github.com/grpc/grpc/pull/33124 to workaround a CLA bot issue.


stanley-cheung (approved):
LGTM.

PR #33129, opened 2023-05-15
[Example] Adding error_handling & error_details example

veblush:
Adding a new example showing how to propagate error status from the server to the client with optional detail information.

### DIFF
--- a/examples/cpp/error_handling/greeter_error_client.cc
+++ b/examples/cpp/error_handling/greeter_error_client.cc
@@ -30,12 +30,13 @@
 
 #ifdef BAZEL_BUILD
 #include "examples/protos/helloworld.grpc.pb.h"
-#include "src/proto/grpc/status/status.pb.h"
 #include "google/rpc/error_details.pb.h"
+
+#include "src/proto/grpc/status/status.pb.h"
 #else
+#include "error_details.pb.h"
 #include "helloworld.grpc.pb.h"
 #include "status.pb.h"
-#include "error_details.pb.h"
 #endif
 
 ABSL_FLAG(std::string, target, "localhost:50051", "Server address");
@@ -98,12 +99,14 @@ class GreeterClient {
     if (error_details.empty()) {
       return;
     }
+    // If error_details are present in the status, this tries to deserialize
+    // those assuming they're proto messages.
     google::rpc::Status s;
     if (!s.ParseFromString(error_details)) {
       std::cout << "Failed to deserialize `error_details`" << std::endl;
       return;
     }
-    std::cout << absl::StrFormat("Details: ") << std::endl;
+    std::cout << absl::StrFormat("Details:") << std::endl;
     for (auto& detail : s.details()) {
       google::rpc::QuotaFailure quota_failure;
       if (detail.UnpackTo(&quota_failure)) {
--- a/examples/cpp/error_handling/greeter_error_client.cc
+++ b/examples/cpp/error_handling/greeter_error_client.cc
@@ -1,20 +1,16 @@
-/*
- *
- * Copyright 2021 gRPC authors.
- *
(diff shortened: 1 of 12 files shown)


examples/cpp/error_handling/greeter_error_client.cc
```diff
@@ -0,0 +1,144 @@
...
+              << std::endl;
+    stub_->async()->SayHello(&context, &request, &reply,
+                             [&mu, &cv, &done, &status](Status s) {
+                               status = std::move(s);
+                               std::lock_guard<std::mutex> lock(mu);
+                               done = true;
+                               cv.notify_one();
+                             });
+    std::unique_lock<std::mutex> lock(mu);
+    while (!done) {
```
eugeneo, line 78: Suggestion: `mu.LockWhen` would remove the need for a separate condition variable, streamlining the code.

examples/cpp/error_handling/greeter_error_client.cc
```diff
@@ -0,0 +1,144 @@
...
+ABSL_FLAG(std::string, target, "localhost:50051", "Server address");
+
+using grpc::Channel;
+using grpc::ClientContext;
+using grpc::Status;
+using helloworld::Greeter;
+using helloworld::HelloReply;
+using helloworld::HelloRequest;
+
+class GreeterClient {
```
eugeneo, line 47: Suggestion: change this to a SayHello function that takes a stub as an argument.

eugeneo (approved):
The comments are suggestions, mostly trying to make the example shorter so it is easier to follow.

examples/cpp/error_handling/greeter_error_client.cc
eugeneo: Ignore this comment, this is `std` and not `ABSL`

examples/cpp/error_handling/greeter_error_client.cc
```diff
@@ -0,0 +1,144 @@
...
+    ClientContext context;
+    // The actual RPC.
+    std::mutex mu;
+    std::condition_variable cv;
+    bool done = false;
+    Status status;
+    std::cout << absl::StrFormat("### Send: SayHello(name=%s)", user)
+              << std::endl;
+    stub_->async()->SayHello(&context, &request, &reply,
+                             [&mu, &cv, &done, &status](Status s) {
```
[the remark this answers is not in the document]
veblush: Done.

Code style guide says that `Use default capture by reference ([&]) only when the lifetime of the lambda is obviously shorter than any potential captures.`. In this case, `[&]` can be used here.

examples/cpp/error_handling/greeter_error_server.cc
```diff
@@ -0,0 +1,123 @@
...
+  ServerUnaryReactor* SayHello(CallbackServerContext* context,
+                               const HelloRequest* request,
+                               HelloReply* reply) override {
+    ServerUnaryReactor* reactor = context->DefaultReactor();
+    // Checks the length of name
+    if (request->name().empty() || request->name().length() > 10) {
+      // Returns an error status with error code and message.
+      reactor->Finish(Status(StatusCode::INVALID_ARGUMENT,
+                             "Length of `Name` should be between 1 and 10"));
+      return reactor;
```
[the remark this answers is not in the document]
veblush: Done.

examples/cpp/error_handling/greeter_error_client.cc
veblush: I see where this is coming from but I've decided to keep this structure to be consistent with other examples.

veblush:
I just realized that I need to split this one example into two examples; one for a simple error handling and the other for a complicate one demonstrating how to use error details.

PR #33131, opened 2023-05-15
[EE client channel resolver] Use std::strlen instead of sizeof in TXT record parsing

yijiem:
This is a mistake made in https://github.com/grpc/grpc/pull/33030. `sizeof()` would count the null byte terminated the C string and would cause us to skip a byte if it is used as the index to `result->substr()`. This would also crash if `result` only contains `grpc_config=` as @drfloob pointed out.

### DIFF
--- a/src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
+++ b/src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
@@ -20,7 +20,6 @@
 
 #include <algorithm>
 #include <chrono>
-#include <cstring>
 #include <memory>
 #include <string>
 #include <type_traits>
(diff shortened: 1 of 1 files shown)


src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
```diff
@@ -409,7 +410,7 @@ void EventEngineClientChannelDNSResolver::EventEngineDNSRequestWrapper::
                                  });
       if (result != service_config->end()) {
         service_config_json_ =
-            result->substr(sizeof(kServiceConfigAttributePrefix));
+            result->substr(std::strlen(kServiceConfigAttributePrefix));
```
markdroth, line 413: Instead of using `std::strlen()`, I suggest changing line 405 to define `kServiceConfigAttributePrefix` as type `absl::string_view` instead of `char[]`, and make it `static`.  Then you can change this to use `kServiceConfigAttributePrefix.size()`, and the code will not need to count the number of characters more than once per process.  (The compiler may even be smart enough to do this at compile time; I'm not sure.)

markdroth (approved):

drfloob (approved):

src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
yijiem: Great point! And looks like both MSVC and GCC can compute strlen at compile time: https://github.com/abseil/abseil-cpp/blob/b971ac5250ea8de900eae9f95e06548d14cd95fe/absl/strings/string_view.h#L607-L623, except clang though.

yijiem: Oh wrong, clang seems to have it too: https://godbolt.org/z/ac5n8d5v3.

PR #33133, opened 2023-05-16
[C++ Admin] Fix ODR violation and update bazel dependencies

yashykt:
Fix ODR violation in `grpcpp_admin` library rule.

Earlier, this target was depending on "//src/proto/grpc/testing/xds/v3:csds_proto" which is a test-only proto target in the gRPC repo which copies contents from the original csds.proto from the envoy repo for convenience purposes. Exposing this in public targets can cause ODR violation since now you could potentially have the same proto types being built through different targets.

In the effort of trying to use the proto from the envoy repo directly, I needed to update some of the additional dependencies that the envoy repo requires -
* `io_bazel_rules_go` is updated to https://github.com/bazelbuild/rules_go/releases/tag/v0.39.0
* `bazel_gazelle` is updated to https://github.com/bazelbuild/bazel-gazelle/releases/tag/v0.30.0
* `com_envoyproxy_protoc_gen_validate` is updated to https://github.com/bufbuild/protoc-gen-validate/releases/tag/v1.0.1 (Note the change in repo)

In a subsequent PR, I'll delete the envoy protos from the src/proto repo as well. (I want to make sure that I don't break anything else, trying to keep the change as small as possible.)


test/cpp/end2end/xds/xds_csds_end2end_test.cc
```diff
@@ -22,25 +22,26 @@
...
 #include "src/cpp/client/secure_credentials.h"
-#include "src/proto/grpc/testing/xds/v3/cluster.grpc.pb.h"
-#include "src/proto/grpc/testing/xds/v3/endpoint.grpc.pb.h"
-#include "src/proto/grpc/testing/xds/v3/http_connection_manager.grpc.pb.h"
-#include "src/proto/grpc/testing/xds/v3/listener.grpc.pb.h"
-#include "src/proto/grpc/testing/xds/v3/route.grpc.pb.h"
 #include "test/core/util/test_config.h"
 #include "test/cpp/end2end/xds/xds_end2end_test_lib.h"
 
 #ifndef DISABLED_XDS_PROTO_IN_CC
```
markdroth, line 40: I suspect that we no longer need this guard.  IIRC, this was a hack that Lidi added to handle CSDS being disabled internally.  But hopefully, with this PR, it will no longer be necessary to do that.

Please do a test import to make sure everything works properly before merging this.

markdroth (commented):
This is awesome!

Can you please also remove our local copies of the xDS protos in src/proto/grpc/testing/xds?

Thanks!

markdroth (approved):
Just noticed that you said you were going to remove the copies of the xDS protos in a subsequent PR.  That's fine too.

yashykt:
This PR is currently blocked due to an internal build system mess.

PR #33137, opened 2023-05-16
[PSM Interop] URL Map: register the cleanup hook early

sergiitk:
I've noticed we add the cleanup hook after setting up the infrastructure. Thus, if infra setup failed, the cleanup won't work.
This fixes it, and adds extra checks to not call `cls.test_client_runner` if it's not set.


sergiitk:
Test run:
- [ ] [grpc/core/v1.55.x/branch/linux/grpc_xds_url_map](https://source.cloud.google.com/results/invocations/142a76e0-7b91-4efa-8924-e44e3f85d858)

PR #33138, opened 2023-05-16
[Python] Migrate from yapf to black

sergiitk:
- Switched  from yapf to black
- Reconfigure isort for black
- Resolve black/pylint idiosyncrasies 

Note: I used `--experimental-string-processing` because black was producing "implicit string concatenation", similar to what described here: https://github.com/psf/black/issues/1837. While currently this feature is experimental, it will be enabled by default: https://github.com/psf/black/issues/2188. After running black with the new string processing so that the generated code merges these `"hello" " world"` strings concatenations, then I removed `--experimental-string-processing` for stability, and regenerated the code again.

To the reviewer: don't even try to open "Files Changed" tab 😄 It's better to review commit-by-commit, and ignore `run black and isort`.


sergiitk:
Sorry, had to force-push to preserve the commit structure.

sergiitk:
Yea, sorry, force pushes will have to happen in this PR.

sergiitk:
In both cases I added more autogenerated files to the exclude list, re-ran black, and replace the `run black and isort` commit. Otherwise, I'd have to manually find and restore each autogenerated file, which is prone to errors.

sergiitk:
Per @ctiller's recommendation, we're postponing this merge until after the end of the fixit. Then I'll sync it with master, and redo the code generation.

tools/run_tests/xds_k8s_test_driver/bin/black.sh
```diff
@@ -48,11 +50,11 @@ source "${XDS_K8S_DRIVER_DIR}/bin/ensure_venv.sh"
 
 if [[ "$1" == "--diff" ]]; then
   readonly MODE="--diff"
+elif [[ "$1" == "--check" ]]; then
+  readonly MODE="--check"
 else
-  readonly MODE="--in-place"
-  readonly VERBOSE="--verbose" # print out file names while processing
+  readonly MODE=""
```
XuanWang-Amos, line 56: What will this default mode do?

pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.black]
+line-length = 80
+target-version = ['py39']
```
gnossen: Based on [this](https://github.com/psf/black/issues/751#issuecomment-473066811), I think this should actually be our full supported version range, which is currently 3.7, 3.8, 3.9, 3.10, and 3.11.

pyproject.toml
```diff
@@ -17,3 +17,29 @@ extend-exclude = '''
...
+[tool.isort]
+profile = "black"
+line_length = 80
+src_paths = [
+    "examples/python/data_transmission",
+    "examples/python/async_streaming",
+    "tools/run_tests/xds_k8s_test_driver",
+    "src/python/grpcio_tests",
+    "tools/run_tests",
+]
```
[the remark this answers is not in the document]
gnossen: For the examples, I think the answer is basically "Lidi wrote these examples." I'm surprised that the implementation directory is not included here, but if it never was to begin with, I'm fine with leaving this the way it is.

gnossen (commented):
Huge thanks for this. We've been wanting this for literally years.

tools/run_tests/xds_k8s_test_driver/bin/black.sh
sergiitk: Black in the default mode (no args) will rewrite the code in place. `yapf` needed `--in-place` to do that, otherwise it'd just print the formatted file contents to the stdout.

pyproject.toml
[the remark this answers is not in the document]
sergiitk: Yea, that's weird. I just translated all the inline arguments to isort into a nice and readable config as is. The only change to isort settings is to use `black` profile, for obvious reasons.

pyproject.toml
sergiitk: Good eyes. Will update this.

sergiitk:
Getting ready to rebase and force-push.

pyproject.toml
sergiitk: Just a note: after updating the branch and regenerating the code, I redid the whole process of code regeneration again with updated target-version. It had no effect: https://github.com/grpc/grpc/compare/1b68c728ece6b6f3da5e9fcee22de051df396282..75a1b832e4660ae7f51633bc79d552840b02e55b

sergiitk:
Just a note: The last force push was regenerating the code after with target-version updated to py37...py311. It had no effect on the generated code (and that's OK).

sergiitk:
@XuanWang-Amos Note I made [a small change](https://github.com/grpc/grpc/pull/33138/commits/f1ed6949cdb9d52e726dfb0214f71ff3860ed60b#diff-af2a2d5b2f2145bbd27b4b6181347756ab692a0cd55d474ba480d54a23504ddf) to `src/python/grpcio/grpc/_observability.py` - after reformatting pylint was complaining on
```
src/python/grpcio/grpc/_observability.py:33:4: E0601: Using variable 'ObservabilityPlugin' before assignment (used-before-assignment)
```

I just used the typehint notation made for this case - putting the class declared later in the same file in quotes.

gnossen (approved):

jtattermusch:
@sergiitk  It would be good if the PR description contained info on what is the right command to run now (instead of `tools/distrib/yapf_format.sh`)

sergiitk:
@jtattermusch Sorry, this was a large PR, easy to miss stuff.

`tools/distrib/yapf_code.sh` was replaced by `tools/distrib/black_code.sh`. They are the scripts themselves are compatible API- (argument-) wise. See [a7447dc](https://github.com/grpc/grpc/pull/33138/commits/a7447dc9be8260a582ff944699114cf34c2c6642).

Issue #33080, opened 2023-05-11
[C#] AmbiguousMatchException: Multiple custom attributes of the same type found


alexeygritsenko:
### What version of gRPC and what language are you using?
C# projects:

- GrpcClients (netstandard2.0)
- Net7WebApi (net 7 on IIS)
- Net48WebApi (net 4.8 on IIS)

```
<PropertyGroup>
	<TargetFramework>netstandard2.0</TargetFramework>
</PropertyGroup>

<ItemGroup>
	<PackageReference Include="Google.Protobuf" Version="3.23.0" />
	<PackageReference Include="Grpc.Core" Version="2.46.6" />
	<PackageReference Include="Grpc.Tools" Version="2.54.0">
		<PrivateAssets>all</PrivateAssets>
		<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
	</PackageReference>
	<PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="7.0.0" />
</ItemGroup>
```

### What operating system (Linux, Windows,...) and version?
Win Server 2019

### What runtime / compiler are you using (e.g. python version or version of gcc)
.Net Framework v4.8, .Net 7, VS 2022 latest

### What did you do?
`var channel = new Grpc.Core.Channel(config.Host, config.Port, CreateSslCredentials(cert));`

### What did you see instead?

```
[AmbiguousMatchException: Multiple custom attributes of the same type found.]
   System.Attribute.GetCustomAttribute(Assembly element, Type attributeType, Boolean inherit) +110
   System.Runtime.InteropServices.RuntimeInformation.get_FrameworkDescription() +88

[TargetInvocationException: Exception has been thrown by the target of an invocation.]
   System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor) +0
   System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments) +92
   System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture) +190
   System.Reflection.RuntimePropertyInfo.GetValue(Object obj, BindingFlags invokeAttr, Binder binder, Object[] index, CultureInfo culture) +57
   System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index) +21
[...]

PR #33141, opened 2023-05-16
Handle FrameworkDescription exceptions

tonydnewell:
See https://github.com/grpc/grpc/issues/33080

In some rare circumstances with IIS and .NET Framework the call to `System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription` fails with `System.Reflection.AmbiguousMatchException`. The failure is internal to the `FrameworkDescription` implementation.

The reason for this happening is still unknown.  This fix catches exceptions from `FrameworkDescription` and returns `null` . This will just affect the value we put in the UserAgent string.

### DIFF
--- a/src/csharp/Grpc.Core/Internal/PlatformApis.cs
+++ b/src/csharp/Grpc.Core/Internal/PlatformApis.cs
@@ -206,16 +206,26 @@ namespace Grpc.Core.Internal
         /// </summary>
         static string TryGetFrameworkDescription()
         {
+            try
+            {
 #if NETSTANDARD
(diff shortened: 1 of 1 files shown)


jtattermusch:
Since it seems we don't fully understand the circumstances under which this behavior could happen and Grpc.Core is now very close to being declared as deprecated (it was in maintenance mode until May 2023, we're currently looking into making the deprecation official and based on the data I've seen it's unlikely we'd extend the maintenance period further), I think it's safer to close this. We don't want to risk causing an issue in the last version of Grpc.Core we plan to ever release.

PR #33144, opened 2023-05-16
[infra] Fix absl::Mutex check and remove all uses

drfloob:
`tools/run_tests/sanity/check_absl_mutex.sh` was broken, a missing paren crashed the script if run locally. It's unclear yet how our sanity checks were not complaining about this, `run_tests.py` does not save the log.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -2971,7 +2971,6 @@ grpc_cc_library(
         "absl/status:statusor",
         "absl/strings",
         "absl/strings:cord",
-        "absl/synchronization",
         "absl/types:optional",
         "absl/types:variant",
         "upb_collections_lib",
--- a/src/core/BUILD
+++ b/src/core/BUILD
@@ -1066,7 +1066,6 @@ grpc_cc_library(
         "absl/container:flat_hash_set",
         "absl/status",
         "absl/strings",
-        "absl/synchronization",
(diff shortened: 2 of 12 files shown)


drfloob:
Lol.

```➜  grpc git:(tmptmptmp) tools/run_tests/sanity/check_absl_mutex.sh
find: invalid expression; I was expecting to find a ')' somewhere but did not see one.
➜  grpc git:(tmptmptmp) ✗ echo $?
0
```

drfloob:
This is an odd one, I'm not sure where the bug is here exactly. I figured `set -e` would cause the script to exit when the find command fails, but it doesn't. And the diff at the end succeeds, so the script returns 0.

```
➜  grpc git:(no-absl-mutex) find . \( \( -name '*foo' \) -print0 | xargs -0 grep -n 'Mutex'; echo $? 
find: invalid expression; I was expecting to find a ')' somewhere but did not see one.
123
➜  grpc git:(no-absl-mutex) find . \( \( -name '*foo' \) -print0 | xargs -0 grep -n 'Mutex' | diff - /dev/null; echo $?
find: invalid expression; I was expecting to find a ')' somewhere but did not see one.
0
```

yashykt (approved):

PR #33158, opened 2023-05-17
[JSON] fix bug that incorrectly allowed trailing commas after an empty container

markdroth:


markdroth:
I believe the intent here is that `container_just_begun_` indicates when the previous character is a [`{`](https://github.com/grpc/grpc/blob/55a8302969c32534627e87eb2bd11f6a865f01b9/src/core/lib/json/json_reader.cc#L610) or [`[`](https://github.com/grpc/grpc/blob/55a8302969c32534627e87eb2bd11f6a865f01b9/src/core/lib/json/json_reader.cc#L618).  We're using that to determine if it's okay to accept [a `}` character when we're expecting an object key](https://github.com/grpc/grpc/blob/55a8302969c32534627e87eb2bd11f6a865f01b9/src/core/lib/json/json_reader.cc#L468) or [a `]` character when we're expecting an array element](https://github.com/grpc/grpc/blob/55a8302969c32534627e87eb2bd11f6a865f01b9/src/core/lib/json/json_reader.cc#L475).  However, the only place that we were setting it back to false was [here](https://github.com/grpc/grpc/blob/55a8302969c32534627e87eb2bd11f6a865f01b9/src/core/lib/json/json_reader.cc#L524), and that code-path was not used when we saw a `}` or `]` character, so we were never resetting it to false in the case that the container was empty.

PR #33161, opened 2023-05-17
[fuzzing] Extract and modernize ChannelArgs fuzzer configuration

drfloob:
ChannelArgs fuzz configuration is expected to be used in other fuzzing targets as well. This PR extracts the common code from the API fuzzer and converts to use the C++ types.

This also fixes a minor bug in the API fuzzer where quotas were resized even when they were unused.


test/core/end2end/fuzzers/api_fuzzer.proto
```diff
@@ -95,12 +85,12 @@ message Metadatum {
 
 message CreateChannel {
     string target = 1;
-    repeated ChannelArg channel_args = 2;
+    grpc.testing.FuzzingChannelArgs channel_args = 2;
```
markdroth, line 88: To avoid breaking existing corpora entries, we could leave the existing field in place but add the new field as well, and then have the API fuzzer code ignore the old field if the new field is set.

This may or may not be worth doing, but I figured I'd offer the suggestion.

test/core/end2end/fuzzers/api_fuzzer.cc
```diff
@@ -272,7 +274,7 @@ class ApiFuzzer : public BasicFuzzer {
   grpc_completion_queue* cq_ = nullptr;
   grpc_server* server_ = nullptr;
   grpc_channel* channel_ = nullptr;
-  grpc_resource_quota* resource_quota_;
+  grpc_core::RefCountedPtr<grpc_core::ResourceQuota> resource_quota_ = nullptr;
```
markdroth, line 277: No need for `= nullptr`; that's the default.

test/core/util/fuzzing_channel_args.proto
```diff
@@ -0,0 +1,32 @@
...
+
+package grpc.testing;
+
+message ResourceQuota {}
+
+message FuzzingChannelArg {
+    string key = 1;
+    oneof value {
+        string str = 2;
+        int64 i = 3;
```
markdroth, line 25: Suggest calling this `int`.

test/core/util/fuzzing_channel_args.h
```diff
@@ -0,0 +1,40 @@
...
+namespace grpc_core {
+namespace testing {
+
+// Create ChannelArgs from a fuzzer configuration.
+// resource_quota is an optional input/output argument. If it's not a nullptr,
+// it will be set on the ChannelArgs object. If it is a nullptr, then it will be
+// set to a ResourceQuota object iff the fuzzer configuration has requested that
+// a ResourceQuota be created.
+ChannelArgs CreateFuzzingChannelArgs(
+    absl::string_view prefix_identifier,
```
markdroth, line 33: Suggest moving this to right before the `resource_quota` arg, since it's only used for RQ creation.  Also, please document.

test/core/end2end/fuzzers/api_fuzzer.proto
drfloob: Given that the api fuzzer corpus only has 2 items with channel_args specified, I'd be more inclined to change those two to match the new format than to keep the tech debt of both ChannelArgs proto specifications. And I also trust Craig that the entries would be regenerated pretty quickly if regressions occurred.

We could aim for backwards compatibility with the text-protos that we keep in the corpus, but I think managing the collection would be a bit messier - each fuzzer would have a collection of `FuzzingChannelArg`, instead of a single `FuzzingChannelArgs`, and we might need a templated creation method.

test/core/util/fuzzing_channel_args.h
drfloob: Done. It could be used for more than a ResourceQuota eventually, but I only documented it as that.

test/core/util/fuzzing_channel_args.proto
drfloob: C++ does not like `fuzz_arg.int()`. I considered `integer`, but that seems unnecessary.

```
test/core/util/fuzzing_channel_args.cc:41:66: error: expected unqualified-id                                                                                                                                                                                                                                                                                                                
        channel_args = channel_args.Set(fuzz_arg.key(), fuzz_arg.int()); 
```

test/core/end2end/fuzzers/api_fuzzer.cc
drfloob: Good catch, fixed.

markdroth (approved):
Craig should probably have a look as well, but this looks good to me!

test/core/util/fuzzing_channel_args.h
```diff
@@ -0,0 +1,44 @@
...
+// Create ChannelArgs from a fuzzer configuration.
+// resource_quota is an optional input/output argument. If it's not a nullptr,
+// it will be set on the ChannelArgs object. If it is a nullptr, then it will be
+// set to a ResourceQuota object iff the fuzzer configuration has requested that
+// a ResourceQuota be created.
+// The prefix_identifier is a name given to the ResourceQuota if one is created.
+ChannelArgs CreateFuzzingChannelArgs(
+    const grpc::testing::FuzzingChannelArgs& fuzzing_channel_args,
+    absl::string_view prefix_identifier,
+    RefCountedPtr<ResourceQuota>* resource_quota = nullptr);
```
ctiller, line 39: let's change this now to be something like:
```
ChannelArgs CreateFuzzingChannelArgs(
    const grpc::testing::FuzzingChannelArgs& fuzzing_channel_args,
    absl::string_view prefix_identifier,
    FuzzingEnvironment fuzzing_environment);
 ```
 with
```
struct FuzzingEnvironment {
  optional<RefCountedPtr<ResourceQuota>> resource_quota;
};
```

we'll want to add more types in the future that get bound into the fuzzers, may as well set things up so it's an O(1) operation when it happens.

**Review by ctiller — APPROVED** [review]
One small API change to make things easier in the future.

File: test/core/util/fuzzing_channel_args.h (hunk 5 above)
Comment (drfloob) [reply test/core/util/fuzzing_channel_args.h]: Good idea, I added something similar. It's fairly clean and has maybe one unnecessary allocation in some cases (probably none when we get designated initializers with C++20 ... someday).

New thought for another improvement: the API fuzzer currently uses the same resource quota in both the client & server channels. It seems valuable to split them up to exercise separate reclamation actions for clients & servers. The logic gets a bit nicer if we make a proto change, but the following could work if we don't want to change the proto.

```
message Action {
    reserved 2;
    oneof type {
        ...
        // resize both client and server quotas.
        uint32 resize_resource_quota = 22;
        // use separate quotas. These are ignored if resize_resource_quota is specified.
        uint32 resize_client_resource_quota = 23;
        uint32 resize_server_resource_quota = 23;
    }
}
```

PR #33162, opened 2023-05-17
[example] Server reflection

drfloob:
This is nearly identical to the helloworld example's greeter callback server, but without the health check service enabled. Having a separate example for reflection may help folks find this code.

### DIFF
--- a/examples/cpp/reflection/BUILD
+++ b/examples/cpp/reflection/BUILD
@@ -1,10 +1,10 @@
-# Copyright 2023 The gRPC Authors
+# Copyright 2023 gRPC authors.
 #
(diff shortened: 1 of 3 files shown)


eugeneo (approved):
Should there also be an example for how clients can use the reflection? Would be really helpful to tools developers.

drfloob:
> Should there also be an example for how clients can use the reflection? Would be really helpful to tools developers.

For consistency with the other gRPC implementations, only server-side reflection examples have been created. See [go's](https://github.com/grpc/grpc-go/tree/master/examples/features/reflection) and [java's](https://github.com/grpc/grpc-java/tree/master/examples/example-reflection) examples for comparison.

eugeneo (approved):

Issue #33157, opened 2023-05-17
Some false sharing problems in struct grpc_completion_queue.


LeiW000:
Seeing some false sharing problems in "struct grpc_completion_queue". 

The following is how I identified the problems:
1. sudo -E ./run_performance_tests.py -l c++ --category scalable -r cpp_protobuf_async_streaming_qps_unconstrained_1cq_secure
2. Run "perf c2c record  -a -u --ldlat 50 -p $PID -F 2000 -- sleep 10" while the command above is running.
3. sudo perf c2c report -NN -g --call-graph --full-symbols -c pid,iaddr --stdio > perf_report.txt

Below is the output of "perf c2c report". There are HITMs (Hit modified cache line) at offset 0x0, 0x8, 0x10 and 0x18 of the same cacheline 0x7fff50011d40. It means that there are some false sharings in "struct grpc_completion_queue". I guess we can see some performance improvement if adding some padding in that struct.
<img width="945" alt="image" src="https://github.com/grpc/grpc/assets/51774176/a9095301-dd38-4483-9622-35af652b8703">
<img width="943" alt="image" src="https://github.com/grpc/grpc/assets/51774176/ae937fd8-44ca-4354-9d8b-6355063c632d">

veblush:
@LeiW000 Is this now ready to be closed as PRs got merged?

LeiW000:
Yes, close it as the PR has been merged. Thanks.

PR #33173, opened 2023-05-18
[Core] Fix the false sharing in struct grpc_completion_queue

LeiW000:
For the details of the issue, please see #33157.

Add several paddings in struct grpc_comletion_quete to avoid the false sharing.

I see performance improvement with the QPS tests.

cpp_protobuf_async_streaming_qps_unconstrained_1cq_insecure: ~2.9%
cpp_protobuf_async_unary_qps_unconstrained_1cq_insecure: ~7%

### DIFF
--- a/src/core/lib/surface/completion_queue.cc
+++ b/src/core/lib/surface/completion_queue.cc
@@ -350,9 +350,13 @@ struct grpc_completion_queue {
   /// Once owning_refs drops to zero, we will destroy the cq
   grpc_core::RefCount owning_refs;
 
+  char padding_1[GPR_CACHELINE_SIZE];
   gpr_mu* mu;
 
+  char padding_2[GPR_CACHELINE_SIZE];
   const cq_vtable* vtable;
+
+  char padding_3[GPR_CACHELINE_SIZE];
   const cq_poller_vtable* poller_vtable;
 
 #ifndef NDEBUG
--- a/src/core/lib/surface/completion_queue.cc
+++ b/src/core/lib/surface/completion_queue.cc
@@ -349,7 +349,7 @@ struct cq_callback_data {
 struct grpc_completion_queue {
   /// Once owning_refs drops to zero, we will destroy the cq
   grpc_core::RefCount owning_refs;
-
+  /// Add the paddings to fix the false sharing
   char padding_1[GPR_CACHELINE_SIZE];
   gpr_mu* mu;
 


ctiller:
This looks promising for single CQ workloads. Generally I'd recommend sharding across CQs for folks seeing contention, but this might make things better for misconfigurations.

I have some concerns over synchronous API use, since this is adding about 200 bytes per CQ, and in that world it's about 200 bytes per call. We have some relatively large sync API usage around and this will be negatively impactful for those users.

Ultimately we're going to prefer callback APIs to either the old async or sync APIs (and hopefully this year divorce the CQ concept from that API entirely)... in the meantime, could I get you to make the same comparison for some callback usecases too?

LeiW000:
For concerns over synchronous API use, I tested with the QPS tests. It looks like there is no negative impact on it. Actually, the QPS increases a little bit. :)
cpp_protobuf_sync_streaming_qps_unconstrained_insecure: +0.39%
cpp_protobuf_sync_unary_qps_unconstrained_insecure: +1.15%

>in the meantime, could I get you to make the same comparison for some callback usecases too?
Yes, I will do it. Is there any existing test case I can use directly? Or, do I have to write them myself?

ctiller:
I did some math and satisfied myself that the additional memory bloat is survivable for the performance increase.

LeiW000:
I made some changes to scenario_config.py and run the callback usecases. It looks like the testing results are positive for callback cases.

cpp_protobuf_callback_streaming_qps_unconstrained_insecure: +~1.0%
cpp_protobuf_callback_unary_qps_unconstrained_insecure: +~5.87%

src/core/lib/surface/completion_queue.cc
```diff
@@ -350,9 +350,13 @@ struct grpc_completion_queue {
   /// Once owning_refs drops to zero, we will destroy the cq
   grpc_core::RefCount owning_refs;
 
+  char padding_1[GPR_CACHELINE_SIZE];
```
veblush, line 353: How about having some comments here to explain the intention as these code can be puzzling to some poeple?

LeiW000: I just updated the PR to add the comment.

LeiW000: @veblush , could you take a look at the PR again? I have added the comment as you suggested. Thanks.

veblush: lgtm. thanks!

veblush (approved):

Issue #33032, opened 2023-05-06
ENH: Support shared library builds on windows


h-vetinari:
Hey, I help maintain grpc (and many downstream users of it) in conda-forge, where we're extremely strongly inclined to build shared libraries for everything, because we can handle the ABI impacts (by recompiling where necessary; package metadata is good enough to ensure no published artefacts will collide or crash), and because static libs bloat up all consumers.

Grpc is one of the few libraries where for reasons of pragmatism, we've so far stayed on static builds for windows, because the shared ones just don't compile. This is because MSVC requires all the horrible, _horrible_, HORRIBLE `__declspec(dllimport)` / `__declspec(dllexport)` notations, which _need_ to be macros that switch based on whether you're building or consuming the library (for _each_ shared library individually).

In most cases, there's a magic CMake variable called [CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html), however, this doesn't work for for global _data_ symbols, where it still _is_ necessary to use the declspec stuff.

Abseil has an [implementation](https://github.com/abseil/abseil-cpp/blob/20230125.3/absl/base/config.h#L743-L760) of the required switch and annotated all its functions/symbols accordingly. I have a prototype implementation based on that switch that _only_ annotates the global data symbols (relying on the above-mentioned CMake magic otherwise) and at least passes the compilation phase, see https://github.com/conda-forge/grpc-cpp-feedstock/pull/243. It's not a pretty sight, as it needs 4 different switches (one each for GPR, gRPC, gRPC++ & UBP).

For easier inspection, I've pushed a [branch](https://github.com/h-vetinari/grpc/tree/win_shared_1.54) with the relevant commits for that (the PR on the feedstock is mixed with fixing https://github.com/grpc/grpc/issues/30838). The branch doesn't cleanly rebase on master, so it's based on the `v1.54.x` branch. Key commit is this one: https://github.com/h-vetinari/grpc/commit/de89137c824e9838099dc94e2657b785ef99d66e.

[...]

PR #33175, opened 2023-05-18
[Build] WIP: Shared builds on windows

h-vetinari:
Towards https://github.com/grpc/grpc/issues/33032, more details in that issue.

I don't know the CI setup here, but this needs `BUILD_SHARED_LIBS=ON` somewhere, as well as  `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON`. Developed in collaboration with @isuruf in the [context](https://github.com/conda-forge/grpc-cpp-feedstock/pull/243) of redistribution in conda-forge (where we strongly prefer shared libraries; that PR also tries to solve https://github.com/grpc/grpc/issues/30838 at the same time for our purposes).

In many ways, this is a minimal effort to get shared builds to compile _at all_. Following the compilation warnings, there'd be many more things to consider adding to the DLL interface, but for now this is good enough to run grpcio against a shared libgrpc.

Note that this is almost certainly incomplete, but meant to serve as a starting point for more concrete discussions.

CC @veblush @coryan

### DIFF
--- a/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
+++ b/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
@@ -130,10 +130,10 @@ configLangArgs8core+=( -l c++ )
 configLangArgs32core+=( -l c++ )
 runnerLangArgs+=( -l "cxx:${GRPC_CORE_REPO}:${GRPC_CORE_COMMIT}" )
 
-# # dotnet
-# configLangArgs8core+=( -l dotnet )
-# configLangArgs32core+=( -l dotnet )
-# runnerLangArgs+=( -l "dotnet:${GRPC_DOTNET_REPO}:${GRPC_DOTNET_COMMIT}" )
+# dotnet
+configLangArgs8core+=( -l dotnet )
+configLangArgs32core+=( -l dotnet )
+runnerLangArgs+=( -l "dotnet:${GRPC_DOTNET_REPO}:${GRPC_DOTNET_COMMIT}" )
 
 # # go
 # configLangArgs8core+=( -l go )
--- a/tools/internal_ci/linux/grpc_e2e_performance_gke_experiment.sh
+++ b/tools/internal_ci/linux/grpc_e2e_performance_gke_experiment.sh
@@ -124,10 +124,10 @@ configLangArgs8core+=( -l c++ )
 configLangArgs32core+=( -l c++ )
 runnerLangArgs+=( -l "cxx:${GRPC_CORE_REPO}:${GRPC_CORE_COMMIT}" )
 
-# # dotnet
-# configLangArgs8core+=( -l dotnet )
-# configLangArgs32core+=( -l dotnet )
-# runnerLangArgs+=( -l "dotnet:${GRPC_DOTNET_REPO}:${GRPC_DOTNET_COMMIT}" )
+# dotnet
+configLangArgs8core+=( -l dotnet )
+configLangArgs32core+=( -l dotnet )
+runnerLangArgs+=( -l "dotnet:${GRPC_DOTNET_REPO}:${GRPC_DOTNET_COMMIT}" )
 
 # # go
 # configLangArgs8core+=( -l go )


veblush:
Thank you for the PR. I have two comments;
- Are you adding `GRPC_DLL` to all public classes? But you did it to a few classes. Would you explain how you chose those? 
- What `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` does in this case? I thought that it would make all classes exported automatically.

h-vetinari:
Just stumbled over some relevant [comments](https://github.com/protocolbuffers/protobuf/blob/v23.2/cmake/README.md#notes-on-compiler-warnings) in protobuf about the warning (C4251) I mentioned above:
> C4251 is of particular note, if you are compiling the Protocol Buffer library as a DLL (see previous section). The protocol buffer library uses templates in its public interfaces. MSVC does not provide any reasonable way to export template classes from a DLL. However, in practice, it appears that exporting templates is not necessary anyway. Since the complete definition of any template is available in the header files, anyone importing the DLL will just end up compiling instances of the templates into their own binary. The Protocol Buffer implementation does not rely on static template members being unique, so there should be no problem with this, but MSVC prints warning nevertheless. So, we disable it. Unfortunately, this warning will also be produced when compiling code which merely uses protocol buffers, meaning you may have to disable it in your code too.

So not all of those warnings are necessarily something we need to follow up on (though OTOH, abseil builds with the equivalent of `-Werror` on windows and will fail if this warning is encountered; [example](https://github.com/abseil/abseil-cpp/pull/1445))

dawidcha:
Hi h-vetinari,

I work for a commercial organization with an interest in getting gRPC working as a shared library. Esun (veblush) pointed me to this PR following a conference call where we communicated that this was something we were pursuing. In our case, the pressure is because we were experiencing stack corruption caused by an apparent MSVC compiler bug encountered due to extensive use of C++14 features in the gRPC code, and can resolve that by compiling gRPC into a DLL using Visual Studio 2022 in which the bug is fixed - and then linking that with our main branch code compiling on 2017

I'd like to help out in this endeavour to build DLL versions of these libraries and ultimately get them merged into the master branch.

I've cloned the repo and applied the changes in this MR.

I'm seeing a link error at the end of the build:
binder_transport.obj : error LNK2001: unresolved external symbol "class grpc_core::DebugOnlyTraceFlag grpc_slice_refcount_trace" ... [....\build\grpc++.vcxproj] - could that be because I rebased?

h-vetinari:
Based on the name of the missing symbol (`grpc_core::DebugOnlyTraceFlag`), it could be because you're building the debug version. We've only been building the release version based on these patches. I rebased them for 1.57 already, and things look [good](https://github.com/conda-forge/grpc-cpp-feedstock/pull/307) from our POV. I can update the PR here as well.

h-vetinari:
I also put up a [branch](https://github.com/h-vetinari/grpc/tree/cf_patches_1.57) with the full patches for the `v1.57.x` branch. On windows especially you need to be careful which C++ version you're compiling with, because abseil's ABI changes based on that (causing linker errors etc.).

dawidcha:
@h-vetinari I had to add GRPC_DLL to the definition of grpc_slice_refcount_trace in core/lib/slice/slice_refcount.h - maybe a variation in the way the repos were rebased?

I think it's going to be very difficult to merge any PR that annotates _all_ necessary symbols with __declspec(dllim/export), because it will be very broad and susceptible to interference from other unrelated merges

@veblush - this PR seems pretty self-contained, and while it is by no means a complete solution, it does lay the groundwork. , is unlikely to impact existing functionality. Does this approach seem correct? What additional work (testing, etc) is missing to get to a level of confidence that this PR can be accepted.

include/grpc/support/port_platform.h
```diff
@@ -53,6 +53,42 @@
...
+#if defined(gRPC_DLL_EXPORTS)
+#define GRPC_DLL __declspec(dllexport)
+#elif defined(gRPC_DLL_IMPORTS)
+#define GRPC_DLL __declspec(dllimport)
+#else
+#define GRPC_DLL
+#endif // defined(gRPC_DLL_EXPORTS)
+
+// same for gRPC++
+#if defined(gRPCXX_DLL_EXPORTS)
```
veblush, line 75: Can we use `GRPCPP` instead of `gRPCXX`?

include/grpc/support/port_platform.h
```diff
@@ -53,6 +53,42 @@
...
+// inspired by https://github.com/abseil/abseil-cpp/blob/20220623.1/absl/base/config.h#L730-L747
+//
+// When building gRPC as a DLL, this macro expands to `__declspec(dllexport)`
+// so we can annotate symbols appropriately as being exported. When used in
+// headers consuming a DLL, this macro expands to `__declspec(dllimport)` so
+// that consumers know the symbol is defined inside the DLL. In all other cases,
+// the macro expands to nothing.
+// Note: gRPC_DLL_EXPORTS is set in CMakeLists.txt when building shared grpc{,_unsecure}
+//       gRPC_DLL_IMPORTS is set by us as part of the interface for consumers of the DLL
+#if defined(gRPC_DLL_EXPORTS)
```
veblush, line 66: Can we use all capital letters for macros like `GRPC_DLL_EXPORTS`?

include/grpc/support/port_platform.h
```diff
@@ -53,6 +53,42 @@
 #define WIN32_LEAN_AND_MEAN
 #endif /* WIN32_LEAN_AND_MEAN */
 
+// GPRC_DLL
+// inspired by https://github.com/abseil/abseil-cpp/blob/20220623.1/absl/base/config.h#L730-L747
+//
+// When building gRPC as a DLL, this macro expands to `__declspec(dllexport)`
```
veblush, line 59: I'd like to include comments like "This DLL usage is experimental" to give the right expectation.

include/grpc/support/port_platform.h
```diff
@@ -53,6 +53,42 @@
...
+// same for gRPC++
+#if defined(gRPCXX_DLL_EXPORTS)
+#define GRPCXX_DLL __declspec(dllexport)
+#elif defined(gRPCXX_DLL_IMPORTS)
+#define GRPCXX_DLL __declspec(dllimport)
+#else
+#define GRPCXX_DLL
+#endif // defined(GRPCXX_DLL_EXPORTS)
+
+// same for GPR
```
veblush, line 83: GPR is now private so we don't want to expose a control knob to have a dedicated DLL for GPR. All GPR symbols should be able to use `GRPC` macros.

third_party/upb/upb/port/def.inc
```diff
@@ -327,3 +327,30 @@ void __asan_unpoison_memory_region(void const volatile *addr, size_t size);
 #else
 #define UPB_DESC(sym) google_protobuf_##sym
 #endif
+
```
veblush, line 330: All files under `third_party/upb' and `src/core/ext/upb-generated' are not allowed to be modified as they belong to other project. Would you undo this change?

ctiller:
What's the strategy for testing this for regressions?

include/grpc/support/port_platform.h
h-vetinari: Yeah, no problem.

include/grpc/support/port_platform.h
h-vetinari: No problem.

include/grpc/support/port_platform.h
h-vetinari: OK, I'll probably have to try this in our infra first. I didn't introduce these for fun, I'm pretty sure they were necessary at the time. But if it works with one less macro now, all the better!

third_party/upb/upb/port/def.inc
h-vetinari: I can do this for this PR, but then the shared build is not going to work anymore. So I guess we need to raise this with protobuf?

h-vetinari:
> What's the strategy for testing this for regressions?

I think the most reasonable would be a CI job that tries to do a shared build on windows.

CMakeLists.txt
dawidcha, line 1: I think this file is generated by a script from CMakeLists.txt.template, no? I can make that change if it helps.

dawidcha: I updated CMakeLists.txt.template to add the DLL annotations as per the hand-edited CMakeLists.txt, however:
* Not sure how to add additional dependencies - looks like these get generated from Bazel BUILD files.
* The existing template seems to be some way behind the existing CMakeLists.txt file - has the generation idea fallen by the wayside?

dawidcha:
> > What's the strategy for testing this for regressions?
> 
> We can do a nifty trick - build grpc DLLs then build a dummy C++ file that links it. That tests both that the DLLs can be built and that they can be linked.

I've done this. Both the shared and static library builds compile and can be linked to an executable

third_party/upb/upb/port/def.inc
veblush: For `third_party/upb/*`, https://github.com/protocolbuffers/upb is the right place to make this change. We're going to use git-submodule instead of git-subtree for upb soon. Then all changes made here will be discarded.

For `src/core/ext/upb-generated/*`, why do we need to export those symbols? Those upb generated files are supposed to be private so I don't understand why this is required. If there is some case where those symbol is part of gRPC public API, we can change it not to do that.

CMakeLists.txt
veblush: We might be able to create a new test based on https://github.com/grpc/grpc/blob/master/test/distrib/cpp/run_distrib_test_cmake.bat

dawidcha:
I'm summarizing what I believe to be the tasks to be addressed in order to get this PR approved:

* Identifier name changes e.g. gRPC_DLL_EXPORTS -> GRPC_DLL_EXPORTS
* Comments to the effect that shared build is experimental
* Determine if upb changes and symbol publication is necessary
* Assuming they are, either find a way of avoiding that, or make the change in protobuf/upb
* Changes to CMakeLists.txt.template, if needed.
* Write test to do a shared build and link the resultant DLL to a minimal executable.
* Fix android build
* Fix Basic Tests C/C++ Linux [Build Only]
* Fix Basic Tests C/C++ MacOS
* Fix Basic Tests ObjC Examples (tests are failing because XXX_DLL identifiers are not being set)

Open questions:
* Are changes to expose symbols from upb necessary?
* Should we be editing CMakeLists.txt directly or generating from CMakeLists.txt.template?

@h-vetinari @veblush Is there anything I've missed here? What needs to happen to answer the open questions?

I have some capacity and I can take some of these so I'm going to look at answering the upb dependency question. Are we duplicating work here? Should I look at something else?

third_party/upb/upb/port/def.inc
h-vetinari: > For `src/core/ext/upb-generated/*`, why do we need to export those symbols? Those upb generated files are supposed to be private so I don't understand why this is required. If there is some case where those symbol is part of gRPC public API, we can change it not to do that.

This must be used somewhere; I only added the minimal things to avoid build failure.

h-vetinari:
> I have some capacity and I can take some of these

I invited you to collaborate on my fork, feel free to push into this PR or open a new one. From my side, this process will take longer than you seem to be willing to wait.

dawidcha:
So, first piece of news: compiling upb as a static library works just fine.

Upb is linked into grpc and grpc_unsecure (and a few tests), so I don’t think there’ll be a problem with duplication.

I’m going to create a branch with suggested changes

dawidcha:
I think it would be helpful to set up a biweekly(?) sync-up call to discuss progress, do you agree?

From: Chamberlin, David [Engineering]
Sent: 14 August 2023 14:06
To: grpc/grpc ***@***.***>; grpc/grpc ***@***.***>
Cc: Comment ***@***.***>; Houshmand, Mehdi [Engineering] ***@***.***>
Subject: RE: [grpc/grpc] [Build] WIP: Shared builds on windows (PR #33175)

So, first piece of news: compiling upb as a static library works just fine.

Upb is linked into grpc and grpc_unsecure (and a few tests), so I don’t think there’ll be a problem with duplication.

I’m going to create a branch with suggested changes

h-vetinari:
> I think it would be helpful to set up a biweekly(?) sync-up call to discuss progress, do you agree?

No, I don't agree, sorry. I don't have capacity to promise any timeline here, much less have regular syncs on this. Also, I do not decide when this is mergeable, for that you need to appeal to the grpc maintainers (who are also very busy). At least yu now have (an open invitation for) commit rights into this PR, so feel free to work through the list you gathered.

PS. You are replying via mail to github, and the result is pretty mangled. 
PPS. Please stop sending email privately.

dawidcha:
I've committed a change to build upb as a static library which removes the necessity to edit def.inc and annotate functions in upb.

For the upb generated code, it's a bit more complicated. The problem occurs because some of the generated symbols like grpc_health_v1_HealthCheckRequest_msg_init are compiled into grpc, and used by grpc++. A workaround for this would be to include those compilation units with the symbols that need resolving into grpc++ as well. Obviously not a great solution as it's very ad hoc and we end up with duplicate code in grpc++.

I'd suggest that the *right* way to do this is to generate the code as part of the build, and have it include the necessary annotations. That's certainly possible with protoc using --cpp_out=dllexport_decl=MY_EXPORT_MACRO:output/directory as described [__here__](--cpp_out=dllexport_decl=MY_EXPORT_MACRO:output/directory), but this code is generated with upbc.

Do we know if it's possible to get upbc to emit the necessary annotations?

dawidcha:
I have a build working without any annotations to the generated upb code - using the strategy of including extra (only 2 to 4) compilation units into the grpc++, grpc++_unsecure and grpc++_alts builds.

This is feeling like the most practical way forward at the moment.

Any objections to my committing this against the PR? Any guidance on how we might get the generated upb code to optionally include __declspec annotations?

veblush:
> I've committed a change to build upb as a static library which removes the necessity to edit def.inc and annotate functions in upb.
> 
> For the upb generated code, it's a bit more complicated. The problem occurs because some of the generated symbols like grpc_health_v1_HealthCheckRequest_msg_init are compiled into grpc, and used by grpc++. A workaround for this would be to include those compilation units with the symbols that need resolving into grpc++ as well. Obviously not a great solution as it's very ad hoc and we end up with duplicate code in grpc++.
> 
> I'd suggest that the _right_ way to do this is to generate the code as part of the build, and have it include the necessary annotations. That's certainly possible with protoc using --cpp_out=dllexport_decl=MY_EXPORT_MACRO:output/directory as described **here**, but this code is generated with upbc.
> 
> Do we know if it's possible to get upbc to emit the necessary annotations?

If there is an use-case where both gRPC Core and gRPC C++ depend on upb-generated code, we should make upb support DLL build as well. Otherwise, we cannot get it. I'll follow up on this.

veblush:
@drfloob AJ, would you mind reviewing changes around Folk and IOManager? I'd like to check whether those would a performance implication.

veblush:
> * Should we be editing CMakeLists.txt directly or generating from CMakeLists.txt.template?

Yes, the template file is what you need to change. Running `tools/buildgen/generate_projects.sh` will update `CMakeLists.txt` based on `CMakeLists.txt.template`

> * The above list of failing tests was probably incomplete - I didn't see the scrollbar on the right. Looks like the tests haven't run since pushing changes, is there a way to force them to run immediately?

This is because only gRPC maintainers can run the test. I just triggered it and please let me know if you want to run the test again.

dawidcha:
@veblush Thanks for rerunning tests - as I suspected there are many more than the ones highlighted in the task list. However, it looks like most of the issues are down to the GRPC_DLL, GPR_DLL and GRPCXX_DLL symbols not being defined in these tests. It's difficult to see why that might be happening - these symbols are always defined to something (in these cases it should probably be an empty string) in grpc/support/port_platform.h which is included in the headers that are failing with an error.

As far as I can see, there must be some way that the wrong port_platform.h is being included. Is there a way I can run these tests locally to investigate?

In the meantime, I'm going to code around this by adding defensive code like:
#ifndef GRPC_DLL
#define GRPC_DLL
#endif

dawidcha:
Update:
- Changes to CMakeLists.txt.template, if needed. _This is almost ready for review. The sticking point is that one edit to CMakeLists.txt added a dependency on grpc++ to grpc++_unsecure. This feels wrong, if the whole point of grpc++_unsecure is to offer a cut-down version of grpc++ for use in testing... (is it?) Also, the list of dependencies is divined from the bazel build files in a prior 'code generation' step, so hacking in another dependency that doesn't align with that feels a bit kludgy._
- Write test to do a shared build and link the resultant DLL to a minimal executable. _I've not managed to get the original test (run_distrib_test_cmake.bat) passing. I think I've pulled in all the third party dependencies, but it's failing saying that address_sorting.lib was not built correctly._

I'll merge the changes that should address the main issue behind the pre-review check failures tomorrow and keep plugging at resolving the missing dependencies in grpc++_unsecure some other way and getting that test to pass.

veblush:
> In the meantime, I'm going to code around this by adding defensive code like: #ifndef GRPC_DLL #define GRPC_DLL #endif

This is because these new symbols are defined only on Windows. Then all non-Windows build cannot find it. So you may need to move `#define *_DLL` outside of the windows if-block so that non-WIndows build can have the definition of those macros.

> Changes to CMakeLists.txt.template, if needed

Yes, it's needed. 

> Write test to do a shared build and link the resultant DLL to a minimal executable

I think we can do it in a separate PR if you don't want to be blocked on.

dawidcha:
@veblush I've fixed the definitions in port_platform.h. Can you kick off the validation tests to see which, if any, are fixed as a result.

I've also updated CMakeLists.txt to remove the direct dependency of grpc++_unsecure on grpc++. To do this I had to include some additional source files in the source and lib list - I've done this in such a way that it shouldn't impact the non-Windows dynamic build.

I'm nearly there with the CMakeLists.txt.template - I'll merge that soon, and update CMakeLists.txt with the generated version.

Remaining work to do is ensuring all the tests pass - I haven't been building the tests (with cmake .... -DgRPC_BUILD_TESTS=ON). Will this happen automatically, as part of the PR validation?

h-vetinari:
@jtattermusch @veblush, could you reopen this PR? I have no idea how github considers this related to https://github.com/grpc/grpc/commit/3b4b69a509a182fbd5b92f27ad39bbdd80ed85b2 (no "Fixes #xxx" or similar either), but from the content of the commit it's clear that it has nothing to do with it.

h-vetinari:
> @dawidcha force-pushed the `win_shared` branch from https://github.com/grpc/grpc/commit/1f21ab09e7c93b0233162a680f72a8dc49ea8ace to https://github.com/grpc/grpc/commit/3b4b69a509a182fbd5b92f27ad39bbdd80ed85b2 1 hour ago

Ah well, that answers the relation to https://github.com/grpc/grpc/commit/3b4b69a509a182fbd5b92f27ad39bbdd80ed85b2

dawidcha:
> > @dawidcha force-pushed the `win_shared` branch from [1f21ab0](https://github.com/grpc/grpc/commit/1f21ab09e7c93b0233162a680f72a8dc49ea8ace) to [3b4b69a](https://github.com/grpc/grpc/commit/3b4b69a509a182fbd5b92f27ad39bbdd80ed85b2) 1 hour ago
> 
> Ah well, that answers the relation to [3b4b69a](https://github.com/grpc/grpc/commit/3b4b69a509a182fbd5b92f27ad39bbdd80ed85b2)

I'm rebasing...

h-vetinari:
Yes well, don't push while you're in the middle of a rebase. I did the rebase in #34103 now

PR #18156, opened 2019-02-25
Let the c++ interop_client send additional metadata, controlled by a flag.

mkbehr:
Add the --additional_metadata flag to the c++ interop client. The flag specifies extra metadata that the client will add to each request.

This lets the client work with server setups that need to see specific HTTP headers, e.g. in order to route the request in a specific way.

PR #33180, opened 2023-05-18
[interop test docs] document --additional_metadata flag

apolcyn:
The flag was introduced a while back for C++: https://github.com/grpc/grpc/pull/18156

https://github.com/grpc/grpc-go/pull/6295 and https://github.com/grpc/grpc-java/pull/10201 for the same flag in Java and Go (need this for RLS integration tests).

### DIFF
--- a/doc/interop-test-descriptions.md
+++ b/doc/interop-test-descriptions.md
@@ -39,6 +39,9 @@ Clients should accept these arguments:
 * --service_config_json=SERVICE_CONFIG_JSON
     * Disables service config lookups and sets the provided string as the
       default service config.
+* --additional_metadata=ADDITIONAL_METADATA
+    * Additional metadata to send in each request, as a semicolon-separated list
+      of key:value pairs. All character but semicolons are permitted in values.
 
 Clients must support TLS with ALPN. Clients must not disable certificate
 checking.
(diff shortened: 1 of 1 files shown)


doc/interop-test-descriptions.md
```diff
@@ -39,6 +39,9 @@ Clients should accept these arguments:
 * --service_config_json=SERVICE_CONFIG_JSON
     * Disables service config lookups and sets the provided string as the
       default service config.
+* --additional_metadata=ADDITIONAL_METADATA
+    * Additional metadata to send in each request, as a semicolon-separated list
```
ejona86, line 43: Have some mention that the keys must be ASCII-only. No `-bin` headers allowed?

Also make clear the first colon separates the key and value.

ejona86:
Note: if this was sent out when originally added, I'd have probably argued to name the argument "metadata". I do actually use these tools directly from the command line for testing. Oh well.

doc/interop-test-descriptions.md
apolcyn: Done

apolcyn:
> 
Note: if this was sent out when originally added, I'd have probably argued to name the argument "metadata". I do actually use these tools directly from the command line for testing. Oh well.

FWIW the original C++ [PR](https://github.com/grpc/grpc/pull/18156) was done years ago by an external contributor. At first, I was working on a similar change, but then realized this old PR already did what I needed.

ejona86 (approved):

PR #33188, opened 2023-05-18
[test] Remove no-op artifacts builds

drfloob:


jtattermusch:
See discussion in cl/533274447.
I think once we remove the grpc_build_artifact configs, we should remove the grpc_build_packages.cfg and grpc_distribtests.cfg jobs as well?

PR #33190, opened 2023-05-19
Update reference and commit variable names.

paulosjca:


tools/internal_ci/linux/grpc_e2e_performance_gke_experiment.sh
```diff
@@ -109,12 +124,11 @@ trap deleteImages EXIT
 
 # Build and push prebuilt images for running tests.
 time ../test-infra/bin/prepare_prebuilt_workers \
-    -l "cxx:${GRPC_CORE_GITREF}" \
-    -l "dotnet:${GRPC_DOTNET_GITREF}" \
-    -l "go:${GRPC_GO_GITREF}" \
-    -l "java:${GRPC_JAVA_GITREF}" \
-    -l "python:${GRPC_CORE_GITREF}" \
```
wanlin31, line 116: Is there a reason why we removed python prebuild image? Seems like we are going to run python test.

paulosjca: This is just an example I may have based it on a branch that was not current (we had removed python before for a breakage).

PR #33196, opened 2023-05-19
[Audit Logging] End2end test for audit logging in authorization policy

rockspore:
I generated a new client key and cert where a Spiffe ID is added as the URI SAN. As such, we are able to test the audit log contains the principal correctly.

As for the test strategy, a test logger is registered to check the log count only and the log content is verified in details via the built-in stdout logger.

Update: I switched to use the test logger to verify the log content and removed stdout logger here because one the failure of [RBE Windows Debug C/C++](https://source.cloud.google.com/results/invocations/c3187f41-bb1f-44b3-b2b1-23f38e47386d).

Update again: Refactored the test logger in a util such that the authz engine test also uses the same logger. Subsequently, xDS e2e test will also use it.


src/core/tsi/test_creds/README
```diff
@@ -62,6 +62,19 @@ common name which is set to testclient2.
 $ openssl x509 -req -CA ca.pem -CAkey ca.key -CAcreateserial -in client2.csr \
   -out client2.pem -days 3650
 
+client3 is issued by CA:
```
gtcooke94, line 65: Let's name is something more clear that this one is a valid SPIFFE ID
client_with_spiffe or something like that.
That way in the future we don't have to read through the `cnf` file to see what the distinguishing feature of this set of credentials is

test/cpp/end2end/BUILD
```diff
@@ -943,8 +943,8 @@ grpc_cc_test(
     srcs = ["grpc_authz_end2end_test.cc"],
     data = [
         "//src/core/tsi/test_creds:ca.pem",
-        "//src/core/tsi/test_creds:client.key",
-        "//src/core/tsi/test_creds:client.pem",
+        "//src/core/tsi/test_creds:client3.key",
```
gtcooke94, line 946: I see we've totally replaced the `client1` here, just want to verify that we aren't losing any coverage because of this, and `client3` is strictly a superset of what `client1` was testing?
Otherwise we can keep the old test with the old certs as well

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -54,6 +69,40 @@ std::string ReadFile(const char* file_path) {
...
+
+ private:
+  int* log_count_;
+};
+
+class TestAuditLoggerFactory : public AuditLoggerFactory {
+ public:
+  class Config : public AuditLoggerFactory::Config {
+    absl::string_view name() const override { return kLoggerName; }
+    std::string ToString() const override { return ""; }
```
gtcooke94, line 87: I'd be a fan of at least returning a sensible name here in case this gets put into some kind of format string

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -415,6 +485,253 @@ TEST_F(GrpcAuthzEnd2EndTest,
...
+  EXPECT_TRUE(status.ok());
+  EXPECT_EQ(log_count_, 0);
+
+  ClientContext context2;
+  context2.AddMetadata("key-foo", "bar");
+  ::testing::internal::CaptureStdout();
+  status = SendRpc(channel, &context2, &resp);
+  CompareAuditLogEntry(
+      ::testing::internal::GetCapturedStdout(),
+      "{\"authorized\":false,\"matched_rule\":\"\",\"policy_name\":\"authz\","
```
gtcooke94, line 547: Why do we expect `matched_rule` to be empty here?

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -415,6 +485,253 @@ TEST_F(GrpcAuthzEnd2EndTest,
...
+  grpc::Status status;
+
+  ClientContext context1;
+  context1.AddMetadata("key-foo", "foo");
+  status = SendRpc(channel, &context1, &resp);
+  EXPECT_TRUE(status.ok());
+  EXPECT_EQ(log_count_, 0);
+
+  ClientContext context2;
+  context2.AddMetadata("key-foo", "bar");
```
gtcooke94, line 542: Add a comment saying what this combo should cause

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -415,6 +485,253 @@ TEST_F(GrpcAuthzEnd2EndTest,
...
+  CompareAuditLogEntry(
+      ::testing::internal::GetCapturedStdout(),
+      "{\"authorized\":false,\"matched_rule\":\"\",\"policy_name\":\"authz\","
+      "\"principal\":\"spiffe://foo.com/bar/baz\",\"rpc_method\":\"/"
+      "grpc.testing.EchoTestService/Echo\"}");
+  EXPECT_EQ(status.error_code(), grpc::StatusCode::PERMISSION_DENIED);
+  EXPECT_EQ(status.error_message(), "Unauthorized RPC request rejected.");
+  EXPECT_EQ(log_count_, 1);
+
+  ClientContext context3;
```
gtcooke94, line 554: Would it be problematic to split this into 3 different tests? That way you don't have to chain failures or spend time figuring out which part failed

gtcooke94 (commented):
A few initial comments

test/cpp/end2end/BUILD
rockspore: `client.pem` is simply a cert signed by the same CA without any x509 extensions. The new cert is a strict superset with the URI SAN added. The existing test doesn't check anything inside the cert other than verifying it. We don't need the old test.

src/core/tsi/test_creds/README
rockspore: Renamed.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: My understanding is this is supposed to return the config. I changed it to return `{}`. I am thinking about making it crash if called.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: This is denied because it doesn't match any allow rules. So there is no matched rule.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: I don't see any benefit from splitting them. When it fails somewhere, the test log will clearly show which one it is including the line number. I haven't had much trouble navigating that.

test/cpp/end2end/grpc_authz_end2end_test.cc
gtcooke94: I like either of those

rockspore:
[Bazel RBE Windows Debug C/C++](https://source.cloud.google.com/results/invocations/c3187f41-bb1f-44b3-b2b1-23f38e47386d) has been consistently failing and it seems like the captured stdout failed to be unmarshaled into a valid json. This is difficult to troubleshoot as I don't have a windows machine (and running RBE Windows locally needs it as well).

I think I will switch to use the test logger to verify the log content and have to ditch stdout logger here.

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -533,38 +529,36 @@ TEST_F(GrpcAuthzEnd2EndTest, StaticInitWithAuditLoggingOnDeny) {
...
-      "{\"authorized\":false,\"matched_rule\":\"deny_bar\",\"policy_name\":"
-      "\"authz\","
-      "\"principal\":\"spiffe://foo.com/bar/baz\",\"rpc_method\":\"/"
-      "grpc.testing.EchoTestService/Echo\"}");
   EXPECT_EQ(status.error_code(), grpc::StatusCode::PERMISSION_DENIED);
   EXPECT_EQ(status.error_message(), "Unauthorized RPC request rejected.");
-  EXPECT_EQ(log_count_, 2);
+  EXPECT_EQ(audit_logs_.size(), 2);
+  EXPECT_THAT(
+      audit_logs_,
```
gtcooke94, line 554: I think it would be a little clearer to check the contents of the first log after its associated request, then the last log after the last request rather than all of them at the end

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -615,32 +606,33 @@ TEST_F(GrpcAuthzEnd2EndTest, StaticInitWithAuditLoggingOnAllow) {
...
+  EXPECT_EQ(audit_logs_.size(), 1);
+  EXPECT_THAT(
+      audit_logs_,
+      ::testing::ElementsAre(
+          "{\"authorized\":true,\"matched_rule\":\"allow_foo\",\"policy_"
+          "name\":\"authz\",\"principal\":\"spiffe://foo.com/bar/"
+          "baz\",\"rpc_method\":\"/grpc.testing.EchoTestService/Echo\"}"));
 
   ClientContext context2;
+  // Does not match the allow rule.
```
gtcooke94, line 622:
```suggestion
  // Does not match the allow rule. Does not emit an audit log.
```

gtcooke94: Please add to each spot that is designed not to emit an audit log.
I think the readability is significantly better with the slight comment addition.

My main nit here is that you have to read the whole test case to get the context for the checks - for example, here I only know that this is _not_ supposed to emit a log because the previous RPC emitted one, so the count was already `1` by the time the check on line 627 is done.
Adding a few comments will balance that enough for a test

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -691,45 +680,39 @@ TEST_F(GrpcAuthzEnd2EndTest, StaticInitWithAuditLoggingOnDenyAndAllow) {
...
-      "\"principal\":\"spiffe://foo.com/bar/baz\",\"rpc_method\":\"/"
-      "grpc.testing.EchoTestService/Echo\"}");
   EXPECT_EQ(status.error_code(), grpc::StatusCode::PERMISSION_DENIED);
   EXPECT_EQ(status.error_message(), "Unauthorized RPC request rejected.");
-  EXPECT_EQ(log_count_, 3);
+  EXPECT_EQ(audit_logs_.size(), 3);
+  EXPECT_THAT(
+      audit_logs_,
+      ::testing::ElementsAre(
+          "{\"authorized\":true,\"matched_rule\":\"allow_foo\",\"policy_"
```
gtcooke94, line 707: Same comment about checking the content after the associated RPCs, feels cleaner that way

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -1220,20 +1185,26 @@ TEST_F(GrpcAuthzEnd2EndTest, FileWatcherWithAuditLoggingRecoversFromFailure) {
...
-      "\"authz\",\"principal\":\"spiffe://foo.com/bar/baz\",\"rpc_method\":\"/"
-      "grpc.testing.EchoTestService/Echo\"}");
+  EXPECT_EQ(audit_logs_.size(), 3);
   dynamic_cast<grpc_core::FileWatcherAuthorizationPolicyProvider*>(
       provider->c_provider())
       ->SetCallbackForTesting(nullptr);
+  EXPECT_THAT(
+      audit_logs_,
+      ::testing::ElementsAre(
+          "{\"authorized\":true,\"matched_rule\":\"allow_echo\",\"policy_"
```
gtcooke94, line 1199: Same comment about checking log content

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Probably because I wrote this, either way seems equally clear to me. But I changed it so that more people are happy :)

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Added the comment.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Done.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Done.

gtcooke94 (approved):
LGTM (assuming all tests are good, etc)

test/cpp/end2end/grpc_authz_end2end_test.cc
erm-g, line 1: Looks good! I think the only minor suggestion I have is to formally test `"audit_condition": "NONE"`

erm-g (changes requested):

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Thanks. I added the test. PTAL.

erm-g (approved):
LGTM

rockspore:
> Would it make sense to name it `authz_audit_logging_utils`?

We have public headers simply named `grpc_audit_logging.h` or `audit_logging.h` so I don' think it's strictly necessary. But let me know if you feel strongly about making this clear and I am open to changing it accordingly. @erm-g

erm-g:
> > Would it make sense to name it `authz_audit_logging_utils`?
> 
> We have public headers simply named `grpc_audit_logging.h` or `audit_logging.h` so I don' think it's strictly necessary. But let me know if you feel strongly about making this clear and I am open to changing it accordingly. @erm-g

Yeah in this case I think I'm fine

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -82,9 +93,15 @@ class GrpcAuthzEnd2EndTest : public ::testing::Test {
     channel_options.watch_identity_key_cert_pairs();
     channel_options.watch_root_certs();
     channel_creds_ = grpc::experimental::TlsCredentials(channel_options);
+
```
markdroth, line 96: Nit: Please remove unnecessary blank lines within functions.

Same thing throughout.

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -415,6 +433,302 @@ TEST_F(GrpcAuthzEnd2EndTest,
...
+      "    ]"
+      "  }"
+      "}";
+  InitServer(CreateStaticAuthzPolicyProvider(policy));
+  auto channel = BuildChannel();
+  grpc::testing::EchoResponse resp;
+  grpc::Status status;
+
+  ClientContext context1;
+  // Matches the allow rule.
```
markdroth, line 481: Please split each of these cases into their own `TEST_F()` blocks, as per go/tott/649.

Same for all of these tests.

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -798,6 +1120,11 @@ TEST_F(GrpcAuthzEnd2EndTest, FileWatcherRecoversFromFailure) {
   grpc::Status status = SendRpc(channel, &context1, &resp1);
   EXPECT_TRUE(status.ok());
   EXPECT_EQ(resp1.message(), kMessage);
+  ASSERT_EQ(audit_logs_.size(), 1);
```
markdroth, line 1123: Suggest using `EXPECT_THAT(audit_logs_, ::testing::ElementsAre(...))` here.  That way, if the check fails, we won't bail out early.

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -823,6 +1150,11 @@ TEST_F(GrpcAuthzEnd2EndTest, FileWatcherRecoversFromFailure) {
   status = SendRpc(channel, &context2, &resp2);
   EXPECT_TRUE(status.ok());
   EXPECT_EQ(resp2.message(), kMessage);
+  EXPECT_EQ(audit_logs_.size(), 2);
```
markdroth, line 1153: Suggest using `ElementsAre()` here as well.  Note that you can say `audit_logs_.clear()` after the previous pass to clear out the entry you've already checked.

test/cpp/end2end/grpc_authz_end2end_test.cc
```diff
@@ -871,6 +1212,11 @@ TEST_F(GrpcAuthzEnd2EndTest, FileWatcherRecoversFromFailure) {
   EXPECT_EQ(status.error_code(), grpc::StatusCode::PERMISSION_DENIED);
   EXPECT_EQ(status.error_message(), "Unauthorized RPC request rejected.");
   EXPECT_TRUE(resp3.message().empty());
+  EXPECT_EQ(audit_logs_.size(), 3);
```
markdroth, line 1215: Same here.

markdroth (approved):
This looks good!  My comments are all fairly minor, so feel free to merge after addressing.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Done.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Done. Thanks for the link!

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Done.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Done.

test/cpp/end2end/grpc_authz_end2end_test.cc
rockspore: Done.

rockspore (commented):
Thanks for the review, Mark!

PR #33201, opened 2023-05-19
[ring hash] revert 64-to-32-bit changes for objc

markdroth:
This reverts the changes made to ring_hash in #29872.  The comment in the picker code specifically says not to change these variables to an unsigned type, but that's exactly what that PR did.  I don't know if this has actually been causing any problems, but given that we duplicated this algorithm (and that comment) from elsewhere without doing a detailed analysis of it, it seems prudent to stick with the types that the original code suggested were important.

To avoid causing problems for ObjC, I have changed this such that ring_hash is not built unless we are building with xDS support, which we exclude on mobile.  Currently, there is no way to use ring_hash without xDS, although that might change in the future; if it does, we can deal with any problems that arise at that point.

CC @HannahShiSFB


src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
```diff
@@ -360,12 +360,12 @@ RingHash::PickResult RingHash::Picker::Pick(PickArgs args) {
   // Ported from https://github.com/RJ/ketama/blob/master/libketama/ketama.c
   // (ketama_get_server) NOTE: The algorithm depends on using signed integers
   // for lowp, highp, and first_index. Do not change them!
-  size_t lowp = 0;
-  size_t highp = ring.size();
-  size_t first_index = 0;
+  int64_t lowp = 0;
+  int64_t highp = ring.size();
```
yashykt, line 364: you might need a static_cast<int64_t> here to avoid warnings

yashykt (approved):

src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
markdroth: That wasn't there prior to #29872, so I suspect it's not needed.

Issue #33204, opened 2023-05-20
Suggest to add a GPR_LIKELY in cq_pluck().


LeiW000:
According to the assembly code, "if (c->tag == tag) {" is true and jump to 330 in most cases in cq_pluck(). It looks that we can add a GPR_LIKELY around "c->tag == tag" to improve branch prediction. 

By the way, I found it with "cpp_protobuf_sync_streaming_qps_unconstrained_10mps_secure" test.

<img width="647" alt="image" src="https://github.com/grpc/grpc/assets/51774176/8156ba78-a0ec-4aa3-aece-65e83fc711cf">

PR #33207, opened 2023-05-22
[Core] Add a GPR_LIKELY in cq_pluck()

LeiW000:
Make the change to improve branch prediction.

For the details, please see #33204.

With the change, the QPS increases by ~0.5% in "cpp_protobuf_sync_streaming_qps_unconstrained_10mps_secure" test.
cpp_protobuf_sync_streaming_qps_unconstrained_10mps_secure: +0.515%


ctiller:
While this looks fine, we expect to regress sync & async api performance going forward by moving completion queues out of being first class citizens and being only secondary objects (with callbacks from core being the primary thing - supporting the callback api more closely).

We can accept some tinkering now, but with no guarantee that it will hold in the future.

LeiW000:
Thanks for that info. I will put more focus on callbacks.

ctiller (approved):

PR #32782, opened 2023-04-03
[EventEngine] Simple iOS EventEngine experiment configuration.

drfloob:
cc @sampajano

PR #33224, opened 2023-05-23
[CF stream] Add changes to test e2e tests in PR.

mikefan1991:


HannahShiSFB:
It needs to be defined `GRPC_IOS_EVENT_ENGINE_CLIENT` to enable event engine. 
ref: https://github.com/grpc/grpc/pull/32782

@mikefan1991

third_party/cares/cares
eugeneo, line 1: Is this intentional? I believe it needs to be reverted.

third_party/xds
eugeneo, line 1: Is this intentional? I believe it needs to be reverted.

eugeneo (changes requested):

third_party/xds
mikefan1991: This is just a test PR and it will be removed later

PR #33226, opened 2023-05-23
[xDS] clean up cert provider factory and registry APIs

markdroth:
- switch to json_object_loader for config parsing
- use `absl::string_view` instead of `const char*` for cert provider names
- change cert provider registry to use a map instead of a vector
- remove unused mesh_ca cert provider factory


src/core/ext/xds/file_watcher_certificate_provider_factory.cc
```diff
@@ -71,66 +71,54 @@ std::string FileWatcherCertificateProviderFactory::Config::ToString() const {
...
+          .OptionalField("certificate_file", &Config::identity_cert_file_)
+          .OptionalField("private_key_file", &Config::private_key_file_)
+          .OptionalField("ca_certificate_file", &Config::root_cert_file_)
+          .OptionalField("refresh_interval", &Config::refresh_interval_)
+          .Finish();
+  return loader;
+}
+
+void FileWatcherCertificateProviderFactory::Config::JsonPostLoad(
+    const Json& json, const JsonArgs& /*args*/, ValidationErrors* errors) {
```
yashykt, line 87: side non-blocking question on the JSON API - I wonder why we look at the JSON instead of the parsed object in `PostLoad`

yashykt (approved):

src/core/ext/xds/file_watcher_certificate_provider_factory.cc
markdroth: Because the field in the parsed object will be an empty string in both of the following cases:
1. The field was not present or null in the JSON.
2. The field was present in the JSON but had an invalid value.

The logic here wants to add errors in case 1.  In case 2, we will already have added an error during the normal validation for the field, so we don't want to add a second error here.

PR #33228, opened 2023-05-24
[experiments] Split experiments into two separate experiment definition and rollout definition files

Vignesh2208:
The PR does the following:

* Splits the single experiments.yaml file into two files: experiments.yaml and rollouts.yaml.
* The experiments.yaml will now only include experiment definitions. The default values of the experiments must now be specified in rollouts.yaml
* Removes the 'release' default value because it is not used.
* Adds an additional_constraints character string to ExperimentMetadata.
* Introduces a hook in src/core/lib/experiments/config.h to allow registering arbitrary experiment constraint validation callbacks. These callbacks would take an ExperimentMetadata object as input and return the correct value to use for an experiment subject to additional constraints.


src/core/lib/experiments/config.cc
```diff
@@ -50,13 +50,21 @@ ForcedExperiment g_forced_experiments[kNumExperiments];
 
 std::atomic<bool> g_loaded(false);
 
+absl::AnyInvocable<bool(struct ExperimentMetadata)> g_check_constraints_cb =
```
ctiller, line 53: `AnyInvocable` is not trivially destructable, so we'll need to choose something else.
Probably in this case a pointer would be fine (we could even delete it once we've checked the constraints!)... so `absl::AnyInvocable<bool(struct ExperimentMetadata)>* g_check_constraints_cb`?

src/core/lib/experiments/experiments.yaml
```diff
@@ -15,14 +15,6 @@
 # Format of each entry:
```
ctiller, line 15: Probably worth a few lines explaining this defines the experiments, and to refer to rollouts.yaml for the rollout state?

ctiller (approved):

src/core/lib/experiments/config.cc
Vignesh2208: Fixed.

src/core/lib/experiments/experiments.yaml
Vignesh2208: Fixed.

PR #33233, opened 2023-05-24
[ObjC] dns service resolver for cf event engine

HannahShiSFB:
Implement DNS using dns service for iOS.

Current limitation:
1. Using a custom name server is not supported.
2. Only supports `LookupHostname`. `LookupSRV` and `LookupTXT` are not implemented.
3. Not tested with single stack (ipv4 or ipv6) environment
4. ~Not tested with multiple ip records per stack~ manually tested with wsj.com
5. Not tested with multiple interface environment


src/core/lib/event_engine/cf_engine/dns_service_resolver.h
```diff
@@ -0,0 +1,112 @@
...
+#include "src/core/lib/gprpp/ref_counted.h"
+#include "src/core/lib/gprpp/ref_counted_ptr.h"
+
+namespace grpc_event_engine {
+namespace experimental {
+
+class DNSServiceResolverImpl
+    : public grpc_core::RefCounted<DNSServiceResolverImpl> {
+  struct DNSServiceRequest {
+    EventEngine::DNSResolver::LookupHostnameCallback on_resolve_;
```
drfloob, line 37: style guide https://google.github.io/styleguide/cppguide.html#Variable_Names: struct variable naming should not end in `_`.

src/core/lib/event_engine/cf_engine/dns_service_resolver.h
```diff
@@ -0,0 +1,112 @@
...
+    uint16_t port_;
+    std::vector<EventEngine::ResolvedAddress> result_;
+    bool has_ipv4_response_ = false;
+    bool has_ipv6_response_ = false;
+  };
+
+ public:
+  DNSServiceResolverImpl(std::shared_ptr<CFEventEngine> engine)
```
drfloob, line 45: single-argument constructors should be `explicit`. See style guide conventions: https://google.github.io/styleguide/cppguide.html#Implicit_Conversions

src/core/lib/event_engine/cf_engine/dns_service_resolver.h
```diff
@@ -0,0 +1,112 @@
...
+  dispatch_queue_t queue_ =
+      dispatch_queue_create("dns_service_resolver", nullptr);
+  grpc_core::Mutex request_mu_;
+  absl::flat_hash_map<DNSServiceRef, DNSServiceRequest> requests_
+      ABSL_GUARDED_BY(request_mu_);
+};
+
+class DNSServiceResolver : public EventEngine::DNSResolver {
+ public:
+  DNSServiceResolver(std::shared_ptr<CFEventEngine> engine) {
```
drfloob, line 78: single-argument constructors should be `explicit`. See style guide conventions: https://google.github.io/styleguide/cppguide.html#Implicit_Conversions

src/core/lib/event_engine/cf_engine/dns_service_resolver.h
```diff
@@ -0,0 +1,112 @@
...
+    impl_ =
```
drfloob, line 79: This can be done as a member initializer. https://en.cppreference.com/w/cpp/language/constructor

src/core/lib/event_engine/cf_engine/dns_service_resolver.h
```diff
@@ -0,0 +1,112 @@
...
+
+  void LookupHostname(
+      EventEngine::DNSResolver::LookupHostnameCallback on_resolve,
+      absl::string_view name, absl::string_view default_port) override {
+    impl_->LookupHostname(std::move(on_resolve), name, default_port);
+  };
+
+  void LookupSRV(EventEngine::DNSResolver::LookupSRVCallback on_resolve,
+                 absl::string_view /* name */) override {
+    on_resolve(absl::UnimplementedError(
```
drfloob, line 93: It could be problematic do to this inline. Imagine a mutex being held while calling LookupSRV, and the on_resolve callback trying to lock that same mutex.

src/core/lib/event_engine/cf_engine/dns_service_resolver.h
```diff
@@ -0,0 +1,112 @@
...
+        "The DNS Service resolver does not support looking up SRV records"));
+  }
+
+  void LookupTXT(EventEngine::DNSResolver::LookupTXTCallback on_resolve,
+                 absl::string_view /* name */) override {
+    on_resolve(absl::UnimplementedError(
```
drfloob, line 99: Same as above. It could be problematic do to this inline. Imagine a mutex being held while calling LookupSRV, and the on_resolve callback trying to lock that same mutex.

src/objective-c/tests/BUILD
```diff
@@ -341,6 +341,7 @@ grpc_objc_testing_library(
         "//src/core:cf_event_engine",
         "//test/core/event_engine/test_suite/posix:oracle_event_engine_posix",
         "//test/core/event_engine/test_suite/tests:client",
+        "//test/core/event_engine/test_suite/tests:dns",
```
drfloob, line 344: I'm not sure the DNS tests will be able to run on iOS, they are going to rely on a twisted-based Python server running on the same platform. There may be no harm in including this for now, but likely the DNS tests will be disabled for this implementation.

test/core/event_engine/cf/cf_engine_test.cc
```diff
@@ -80,6 +82,211 @@ TEST(CFEventEngineTest, TestConnectionCancelled) {
...
+  std::vector<std::string> ip_strings;
+  std::transform(addresses.cbegin(), addresses.cend(),
+                 std::back_inserter(ip_strings), [](auto const& address) {
+                   return ResolvedAddressToString(address).value_or("ERROR");
+                 });
+  return ip_strings;
+}
+}  // namespace
+
+TEST(CFEventEngineTest, CFEventEngineTest_TestCreateDNSResolver_Test) {
```
drfloob, line 97: `CFEventEngineTest_` in the name is redundant, it can just be `TEST(CFEventEngineTest, CreateDNSResolver)`. I believe this will show up in bazel test logs as `CFEventEngineTest.CreateDNSResolver`.

test/core/event_engine/cf/cf_engine_test.cc
```diff
@@ -80,6 +82,211 @@ TEST(CFEventEngineTest, TestConnectionCancelled) {
...
+  grpc_core::Notification resolve_signal;
+
+  auto cf_engine = std::make_shared<CFEventEngine>();
+  auto dns_resolver = cf_engine->GetDNSResolver({});
+
+  dns_resolver.value()->LookupHostname(
+      [&resolve_signal](auto result) {
+        EXPECT_TRUE(result.status().ok());
+        EXPECT_THAT(ResolvedAddressesToStrings(result.value()),
+                    testing::UnorderedElementsAre("127.0.0.1:80", "[::1]:80"));
```
drfloob, line 124: Mostly out of curiosity, I'm not an iOS dev: are all iOS test environments guaranteed to support ipv4 and ipv6?

test/core/event_engine/cf/cf_engine_test.cc
```diff
@@ -80,6 +82,211 @@ TEST(CFEventEngineTest, TestConnectionCancelled) {
...
+
+        resolve_signal.Notify();
+      },
+      "localtest.me:80", "443");
```
drfloob, line 147: Word of caution that any test that relies on an external service may A) become flaky even though the implementation is fine, and B) may be pretty unfriendly if we run `bazel test --runs_per_test=1000 ...`.

test/core/event_engine/cf/cf_engine_test.cc
```diff
@@ -80,6 +82,211 @@ TEST(CFEventEngineTest, TestConnectionCancelled) {
...
+  dns_resolver.value()->LookupHostname(
+      [&resolve_signal](auto result) {
+        EXPECT_TRUE(result.status().ok());
+        EXPECT_THAT(ResolvedAddressesToStrings(result.value()),
+                    testing::IsSubsetOf(
+                        {"1.2.3.4:80", "[64:ff9b::102:304]:80" /*NAT64*/}));
+
+        resolve_signal.Notify();
+      },
+      "1.2.3.4.nip.io:80", "");
```
drfloob, line 167: Same as above. Word of caution that any test that relies on an external service may A) become flaky even though the implementation is fine, and B) may be pretty unfriendly if we run `bazel test --runs_per_test=1000 ....`

test/core/event_engine/cf/cf_engine_test.cc
```diff
@@ -80,6 +82,211 @@ TEST(CFEventEngineTest, TestConnectionCancelled) {
...
+  dns_resolver.value()->LookupHostname(
+      [&resolve_signal](auto result) {
+        EXPECT_TRUE(result.status().ok());
+        EXPECT_THAT(
+            ResolvedAddressesToStrings(result.value()),
+            testing::UnorderedElementsAre("[2607:f8b0:400a:801::1002]:80"));
+
+        resolve_signal.Notify();
+      },
+      "2607-f8b0-400a-801--1002.sslip.io.", "80");
```
drfloob, line 187: Same as above. Word of caution that any test that relies on an external service may A) become flaky even though the implementation is fine, and B) may be pretty unfriendly if we run `bazel test --runs_per_test=1000 `

True below as well, I won't comment on it.

test/core/event_engine/cf/cf_engine_test.cc
```diff
@@ -80,6 +82,211 @@ TEST(CFEventEngineTest, TestConnectionCancelled) {
...
+  auto cf_engine = std::make_shared<CFEventEngine>();
+  auto dns_resolver = std::move(cf_engine->GetDNSResolver({})).value();
+
+  dns_resolver->LookupHostname(
+      [&resolve_signal](auto result) {
+        EXPECT_EQ(result.status().code(), absl::StatusCode::kCancelled);
+
+        resolve_signal.Notify();
+      },
+      "dont-care-since-wont-be-resolved.test.com", "443");
```
drfloob, line 258: This can race, right? Are you not seeing any failures?

test/core/event_engine/test_suite/tests/dns_test.cc
```diff
@@ -201,6 +201,8 @@ class EventEngineDNSTest : public EventEngineTest {
 EventEngineDNSTest::DNSServer EventEngineDNSTest::dns_server_;
 bool EventEngineDNSTest::skip_end2end_tests_ = false;
 
+#ifndef GRPC_IOS_EVENT_ENGINE_CLIENT
```
drfloob, line 204: Please add `// TODO(hork): implement XFAIL for resolvers that don't support TXT or SRV`.

Context: third-party EventEngine implementations might want to only support A/AAAA records as well, and we can't add ifdefs throughout the test suite for every unknown implementation. I have a spec for a system that will allow declarative support (or lack of) in each implementation's build configuration.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+    port = 80;
+  } else if (port_string == "https") {
+    port = 443;
+  } else if (!absl::SimpleAtoi(port_string, &port)) {
+    on_resolve(absl::InvalidArgumentError(
+        absl::StrCat("Failed to parse port in name: ", name)));
+    return;
+  }
+
+  // TODO(yijiem): Change this when refactoring code in
```
drfloob, line 68: FYI @yijiem

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+  // src/core/lib/address_utils to use EventEngine::ResolvedAddress.
+  grpc_resolved_address addr;
+  const std::string hostport = grpc_core::JoinHostPort(host, port);
+  if (grpc_parse_ipv4_hostport(hostport.c_str(), &addr,
+                               true /* log errors */) ||
```
drfloob, line 73: The C++ linter will check names if you use `/*log_errors=*/true`.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+      grpc_parse_ipv6_hostport(hostport.c_str(), &addr,
+                               true /* log errors */)) {
+    // Early out if the target is an ipv4 or ipv6 literal.
+    std::vector<EventEngine::ResolvedAddress> result;
+    result.emplace_back(reinterpret_cast<sockaddr*>(addr.addr), addr.len);
+    on_resolve(std::move(result));
```
drfloob, line 79: Calling this inline could be problematic. See my other comments on that.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+
+  DNSServiceRef sdRef;
+  auto host_string = std::string{host};
+  auto error = DNSServiceGetAddrInfo(
+      &sdRef, kDNSServiceFlagsTimeout | kDNSServiceFlagsReturnIntermediates, 0,
+      kDNSServiceProtocol_IPv4 | kDNSServiceProtocol_IPv6, host_string.c_str(),
+      &DNSServiceResolverImpl::ResolveCallback, this /* do not Ref */);
+
+  if (error != kDNSServiceErr_NoError) {
+    on_resolve(absl::UnknownError(
```
drfloob, line 91: Calling this inline could be problematic. See my other comments on that.

I won't comment on them further, please locate and fix all instances.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+    port_string = default_port;
+  }
+
+  int port = 0;
+  if (port_string == "http") {
+    port = 80;
+  } else if (port_string == "https") {
+    port = 443;
+  } else if (!absl::SimpleAtoi(port_string, &port)) {
+    on_resolve(absl::InvalidArgumentError(
```
drfloob, line 63: Calling this inline could be problematic. See my other comments on that.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+  absl::string_view port_string;
+  if (!grpc_core::SplitHostPort(name, &host, &port_string)) {
+    on_resolve(
+        absl::InvalidArgumentError(absl::StrCat("Unparseable name: ", name)));
+    return;
+  }
+  GPR_ASSERT(!host.empty());
+  if (port_string.empty()) {
+    if (default_port.empty()) {
+      on_resolve(absl::InvalidArgumentError(absl::StrFormat(
```
drfloob, line 50: Calling this inline could be problematic. See my other comments on that.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+  GRPC_EVENT_ENGINE_DNS_TRACE(
+      "DNSServiceResolverImpl::LookupHostname: name: %.*s, default_port: %.*s, "
+      "this: %p",
+      static_cast<int>(name.length()), name.data(),
+      static_cast<int>(default_port.length()), default_port.data(), this);
+
+  absl::string_view host;
+  absl::string_view port_string;
+  if (!grpc_core::SplitHostPort(name, &host, &port_string)) {
+    on_resolve(
```
drfloob, line 43: Calling this inline could be problematic. See my other comments on that.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+        "DNSServiceSetDispatchQueue failed with error:%d", error)));
+    return;
+  }
+
+  requests_.try_emplace(
+      sdRef, DNSServiceRequest{
+                 std::move(on_resolve), static_cast<uint16_t>(port), {}});
+}
+
+/* static */
```
drfloob, line 110: Is this a functional comment somehow?

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,205 @@
...
+  } else if (address->sa_family == AF_INET6) {
+    request.has_ipv6_response_ = true;
+  }
+
+  // collect results if there is no error (not kDNSServiceErr_NoSuchRecord)
+  if (errorCode == kDNSServiceErr_NoError) {
+    request.result_.emplace_back(address, address->sa_len);
+    auto& resolved_address = request.result_.back();
+    if (address->sa_family == AF_INET) {
+      ((struct sockaddr_in*)resolved_address.address())->sin_port =
```
drfloob, line 154: Please use C++-style casts, the C-style cases have mostly been removed from the codebase.

drfloob (commented):
Great start!

test/core/event_engine/cf/cf_engine_test.cc
HannahShiSFB: Yeah, but specifying a local dns server is not supported for this implementation.
I guess it's less likely having a DNS issue comparing to other external services (homebrew for instance) we are using to setup the environment.
We used it for [iOS PerfTest](https://github.com/grpc/grpc/blob/v1.56.0/src/objective-c/tests/PerfTests/PerfTests.m#L363) for a while which was pretty stable.
So I suppose it's better have test coverage? Or are there any alternative Google services?

Stress runs should be fine as I imagine it'll hit the cache.

test/core/event_engine/cf/cf_engine_test.cc
HannahShiSFB: Technically yes, but very unlikely as it needs to connect to the server (dnssd), send a request and wait for at least two responses (ipv4 and v6) for the lookup to complete.
I didn't see it failed in my tests.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
HannahShiSFB: No, only cosmetic.

test/core/event_engine/cf/cf_engine_test.cc
HannahShiSFB: I don't know. I don't have a ipv4 or ipv6 only network environment and somehow have the simulator or an iOS device to connect to.

src/objective-c/tests/BUILD
HannahShiSFB: I added #ifndef of `GRPC_IOS_EVENT_ENGINE_CLIENT` around to disable the new tests that depending the custom server as well.
The tests moved from `test/core/iomgr/resolve_address_test.cc` are passing.

test/core/event_engine/cf/cf_engine_test.cc
HannahShiSFB: PS: I tested with domains with multiple records (e.g. wsj.com) and it works. But I don't know a service that can always response the same set of IPs, so I don't have a test case about it.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
HannahShiSFB: Also changed to false as I saw ares resolver is passing false now.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
HannahShiSFB: Updated all via `engine_->Run` except those from the callback as they are called from the dispatch queue thread.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
HannahShiSFB: Is it OK to const cast the return of `ResolvedAddress::address()`, otherwise it's going to be considerably more lines to create the addresses separately.

test/core/event_engine/cf/cf_engine_test.cc
drfloob: I don't have a better solution for iOS, I'm not sure what options you have in your test environments. Maybe it will be fine. I'm simply noting that if there is noise in the tests because of services you don't control, those tests might end up not being very valuable.

> I tested with domains with multiple records (e.g. wsj.com) and it works (got all 4 v4 IPs). 

Great to hear!

test/core/event_engine/cf/cf_engine_test.cc
drfloob: I know you can mark the entire bazel test target as flaky, but not sure if you can do that for an individual test.

test/core/event_engine/test_suite/tests/dns_test.cc
HannahShiSFB: Sure.
BTW, this disables the LookupHostname cases as well as the new cases need a custom name server.

test/core/event_engine/cf/cf_engine_test.cc
HannahShiSFB: It does fails with --runs_per_test=1000.
I"m only verifying the code if status is not ok.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
drfloob: The cast you're talking about is pretty common already.

https://github.com/grpc/grpc/blob/498fc99479a01177aa2ad257f1390521495b0788/src/core/lib/event_engine/tcp_socket_utils.cc#L223-L224

There is a clang tidy check enabled to try to prevent C-style casts from ending up in the codebase. I'm guessing tidy does not run on iOS code, but we should try to avoid them regardless. https://clang.llvm.org/extra/clang-tidy/checks/google/readability-casting.html

See ec0b88a5d73445ce5e6ff63c4691956bfe37e57f  and 2ff84ed8c2cb272b26766039d5bb07b009df2617 and d2bff16c25d7c207ef0afb6844fc50f567a2b36d and e0d8c498a60be0b5b5891e280bd441ec4a9d0785 for some of the work done here.

src/objective-c/tests/BUILD
drfloob: Glad to know the twisted server works in the test environment. We can work with the ifdefs for now.

Unfortunately, because you force-pushed to the branch, I can't see if the DNS tests are actually running in the CI environment because the results were cached from a previous run that I don't know how to find now. The links disappear when you rewrite history.

test/core/event_engine/cf/cf_engine_test.cc
drfloob: That's an improvement. I still expect that, given time, this test will also see Unknown and NotFound errors. Marking it flaky is probably sufficient for that as well.

HannahShiSFB:
All comments addressed except for
1. testing with remote domain names. I guess we can disable them if we end up find them failing.
2. is it possible that iOS simulator or device have ipv4 or ipv6 stack only.

test/core/event_engine/cf/cf_engine_test.cc
HannahShiSFB: I changed it to localtest.me as well, so there should be no NotFound errors. For Unknown errors it can happen to other tests as well and I guess we'll need to investigate.

src/objective-c/tests/BUILD
HannahShiSFB: Oh I didn't know it. I only rewrote the address comments one, though the commit history would be cleaner.

HannahShiSFB: I found it in my browser history:
https://source.cloud.google.com/results/invocations/474fcf95-0375-43e1-8825-cec3b6b80c24/targets

PS, the new tests in `test/core/event_engine/cf/cf_engine_test.cc` are only run on Mac for now in CI.
I have manually tested them on iOS locally, I'll need sometime to restructure the test code before I can submit it (a new commit or separate PR)

HannahShiSFB: For the `the twisted server`, it doesn't work on iOS or Mac, the tests depend on it didn't run.

drfloob: I see. Looks like the tests are indeed running, that's great.

drfloob (approved):
LGTM from an EventEngine-API perspective. @sampajano please review as well.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,222 @@
...
+        sdRef, hostname,
+        ResolvedAddressToString(resolved_address).value_or("ERROR").c_str(),
+        context);
+  }
+
+  if (!(flags & kDNSServiceFlagsMoreComing) && request.has_ipv4_response &&
+      request.has_ipv6_response) {
+    if (request.result.empty()) {
+      request.on_resolve(absl::NotFoundError(absl::StrFormat(
+          "address lookup failed for %s: Domain name not found", hostname)));
```
sampajano, line 190: Would it be cleaner to condition on `kDNSServiceErr_NoSuchRecord rather than request.result.empty()?

sampajano (approved):
Thanks so much for the change, Hannah! Looks fantastic!! 😃

Note I didn't take a look line by line but mostly went over the main logic. I trust your coding & testing here 😃

Thanks so much again!

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
HannahShiSFB: This callback will be called at least two times which can be:
1. callback with an IPv4 address
2. callback with kDNSServiceErr_NoSuchRecord to indicate no IPv6 address

So in the second call of callback, although I got no records error, still need to return the IPv4 address.
This can also happen in reversed order where NoSuchRecord comes in the first call of the callback, I need to keep waiting for the result to come from the other stack.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
```diff
@@ -0,0 +1,222 @@
...
+
+  // TODO(yijiem): Change this when refactoring code in
+  // src/core/lib/address_utils to use EventEngine::ResolvedAddress.
+  grpc_resolved_address addr;
+  const std::string hostport = grpc_core::JoinHostPort(host, port);
+  if (grpc_parse_ipv4_hostport(hostport.c_str(), &addr,
+                               /*log_errors=*/false) ||
+      grpc_parse_ipv6_hostport(hostport.c_str(), &addr,
+                               /*log_errors=*/false)) {
+    // Early out if the target is an ipv4 or ipv6 literal.
```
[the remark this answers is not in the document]
sampajano: aha ok thanks for explaining!

In that case, if there's a way for us to detect whether the passed param is a "ip literal" v.s. not, do we still need the former method?

I'm not suggesting that we need to do that, but just curious about whether it would be possible to reduce dependency on 2 sets of implementations, and what is the cost of that (since logic to detect whether something is a "ip literal" can be as simple as a regex match :))

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
sampajano: oh ok! thanks for explaining! (btw if you think this logic is not obvious to understand, please consider putting some comments for future understandings :))

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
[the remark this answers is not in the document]
HannahShiSFB: That's c-core util functions, I expect depending on them should be fine, no?
It's more than regex as we are also validating an ip address, need to consider the invalid values, broad cast addresses, etc. IPv6 can be even more complex.

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
HannahShiSFB: Yeah, it also took me a while to understand how it works as there isn't much document about it.
Let me see how I can better describe it for the resolving flow, which maybe in a follow up PR

sampajano: It's ok. No big deal.. Just do it only if you see fit.. Thanks for explaining! :)

src/core/lib/event_engine/cf_engine/dns_service_resolver.cc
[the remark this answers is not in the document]
sampajano: It is fine to depend on it! I was just trying to understand the reasons and necessity for it. If it saves meaningful complexity to reuse their code, then it's totally valid to do so!

Maybe you could add a comment explaining what case are you relying on them v.s. the iOS solutions? So at least we have awareness on how to simplify if there's ever a need in the future. 😃

HannahShiSFB:
Will merge this PR tomorrow morning in case it fails some internal checks.

sampajano:
Hi Hannah, the PR had to be reverted because it's causing CI failures.

Here's some logs from failed tests:

https://source.cloud.google.com/results/invocations/248656d3-15ac-4d42-a2f8-6bec5c411e67/targets/%2F%2Fsrc%2Fobjective-c%2Ftests:EventEngineTests/log

Do you have an idea how they escaped the CI? 😃

HannahShiSFB:
A bad merge with #33882
Will fix with a new PR.

PR #33234, opened 2023-05-24
[xDS] support "tls" channel creds in bootstrap file

### LINKED ISSUE #32977: Add file-based credentials support for on-prem gRPC clients to talk TLS with xDS Server
We are evaluating gRPC proxyless service mesh, and we have our own xDS server implementation that uses standard file-based m-TLS for Authn.

One potential issue is that c-core only allows 3 default credentials when talking with xDS Server: Fake, Insecure, and GoogleDefault  (https://github.com/grpc/grpc/blob/master/src/core/lib/security/credentials/channel_creds_registry_init.cc#L71). Although c-core, Java, Go all provide ways to implement additional ChannelCredsFactory, libraries that use C-binding such as Python doesn't seem to support the API to register additional ChannelCredsFactory. So it seems that only Insecure and GoogleDefault would be supported in these libraries, and therefore, we cannot use the TLS certs in our on-prem datacenter for them.

Add a new ChannelCredsFactory in c-core so that file-based TLS certs will be supported by default in c-core, and Python / other language bindings that use c-core can directly use this factory as well by configuring it in the bootstrap.json file.

We've also considered extending the Cython library to add the hook for registering additional ChannelCredsFactory, yet it would be nicer if c-core can directly support it by default.

### Additional context
N/A.

markdroth:
Implements [gRFC A65](https://github.com/grpc/proposal/blob/master/A65-xds-mtls-creds-in-bootstrap.md).

Fixes #32977.


yashykt (approved):
Do you think it's worth adding a PSM interop test or a local xds_end2end_test for this?

markdroth:
We can't write a PSM interop test for this, since TD requires GoogleDefaultCreds instead of mTLS.

We can write an e2e test for it, but we'd have to add a bunch of plumbing to the test xDS server to support mTLS.  I suspect that would be more effort than it's worth.

PR #33254, opened 2023-05-25
[lb pick_first] Enable random shuffling of address list

eugeneo:
Part of [gRFC A62](https://github.com/grpc/proposal/blob/master/A62-pick-first.md).


test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -66,15 +129,28 @@ TEST_F(PickFirstTest, Basic) {
...
+  // Without shuffling, the same address will always be returned
+  EXPECT_EQ(ResetPickedAddress(kAddressUris, kPolicyUpdates, false).size(), 1);
+}
+
 }  // namespace
 }  // namespace testing
 }  // namespace grpc_core
 
 int main(int argc, char** argv) {
+  grpc_core::testing::ScopedExperimentalEnvVar env_var(
```
markdroth, line 152: I think we should set this inside of the `WithShuffle` test rather than doing it globally.  In particular, note that we need a test that shows that if the env var is not set, then the pick_first policy does not actually look at the config field, and it always tries the addresses in the order in which they were provided to it.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -37,19 +39,80 @@ class PickFirstTest : public LoadBalancingPolicyTest {
...
+      bool shuffleAddressList) {
+    return MakeConfig(Json::FromArray({Json::FromObject({{
+        "pick_first",
+        Json::FromObject(
+            {{"shuffleAddressList", Json::FromBool(shuffleAddressList)}}),
+    }})}));
+  }
+
+  template <typename C>
+  absl::string_view GetConnectionRequestedAddress(
```
markdroth, line 52: I don't think this is the right approach here.  What this seems to be doing is checking to see which address pick_first starts connecting to first, and then sending a new address list before that connection attempt even finishes.  I think all this really checks is that pick_first will pick *one* of the addresses first, but it does not actually check the order in which it attempts the other addresses in the list if the connection attempt on the first address fails.  And since the tests don't care *which* address the policy chooses first each time, I think these tests would actually pass even if the shuffling functionality was not implemented.

I've sent you #33255 to add a couple of tests showing how pick_first handles multiple addresses, which I hope will be useful examples for you to work from.  That will hopefully give you some good examples to work from here.

I think what we want to test here is the following:
- Send only a single address list update at the start of the test.  It should not be necessary to update the address list after that.
- We want to do multiple connection attempt passes.  Each pass, we should record the order in which pick_first tries the addresses.  The connection attempt for every address except the last one should fail, so that we can record the full order in which it is trying the addresses.  The last address it tries should succeed.  Then we have that subchannel indicate the connection fails by reporting IDLE, which will cause pick_first to report IDLE.
- Just like in the `GoesIdleWhenConnectionFailsThenCanReconnect` test that I added in #33255, calling `ExpectStateAndQueuingPicker(GRPC_CHANNEL_IDLE)` will trigger the pick_first policy to try to reconnect, thus starting a new connection attempt pass.  As per my comment elsewhere, it should re-shuffle the address list right before it starts to connect again.
- We can go through 3 connection attempt passes.  Then we can compare the order in which pick_first tried the addresses in each pass.  If it is shuffling the order, then those lists should be in a different order in each pass; if it's not, then they should be identical in each pass.

markdroth (commented):
This looks like a good start!

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -281,6 +322,13 @@ absl::Status PickFirst::UpdateLocked(UpdateArgs args) {
   }
   // Update latest_update_args_.
   latest_update_args_ = std::move(args);
+  if (latest_update_args_.config != nullptr) {
+    RefCountedPtr<PickFirstConfig> config = latest_update_args_.config;
+    if (config->shuffle_addresses() && ShufflePickFirstEnabled()) {
+      static absl::BitGen bit_gen;
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -281,6 +322,13 @@ absl::Status PickFirst::UpdateLocked(UpdateArgs args) {
   }
   // Update latest_update_args_.
   latest_update_args_ = std::move(args);
+  if (latest_update_args_.config != nullptr) {
+    RefCountedPtr<PickFirstConfig> config = latest_update_args_.config;
+    if (config->shuffle_addresses() && ShufflePickFirstEnabled()) {
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -281,6 +322,13 @@ absl::Status PickFirst::UpdateLocked(UpdateArgs args) {
   }
   // Update latest_update_args_.
   latest_update_args_ = std::move(args);
+  if (latest_update_args_.config != nullptr) {
+    RefCountedPtr<PickFirstConfig> config = latest_update_args_.config;
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -281,6 +322,13 @@ absl::Status PickFirst::UpdateLocked(UpdateArgs args) {
   }
   // Update latest_update_args_.
   latest_update_args_ = std::move(args);
+  if (latest_update_args_.config != nullptr) {
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -62,6 +68,41 @@ namespace {
...
+  static const JsonLoaderInterface* JsonLoader(const JsonArgs&) {
+    static const auto kJsonLoader =
+        JsonObjectLoader<PickFirstConfig>()
+            .OptionalField("shuffleAddressList",
+                           &PickFirstConfig::shuffle_addresses_)
+            .Finish();
+    return kJsonLoader;
+  }
+
+  explicit PickFirstConfig() {}
```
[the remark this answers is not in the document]
eugeneo: Thank you. Probably had it for something like logging during the debug 😀

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -62,6 +68,41 @@ namespace {
...
+  auto value = GetEnv("GRPC_EXPERIMENTAL_PICKFIRST_LB_CONFIG");
+  if (!value.has_value()) return false;
+  bool parsed_value;
+  bool parse_succeeded = gpr_parse_bool_value(value->c_str(), &parsed_value);
+  return parse_succeeded && parsed_value;
+}
+
+class PickFirstConfig : public LoadBalancingPolicy::Config {
+ public:
+  static const JsonLoaderInterface* JsonLoader(const JsonArgs&) {
```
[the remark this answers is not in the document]
eugeneo: Done

eugeneo (commented):
Thank you for the reviews!

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -39,6 +47,78 @@ class PickFirstTest : public LoadBalancingPolicyTest {
...
+    subchannel->SetConnectivityState(GRPC_CHANNEL_READY);
+    auto picker = WaitForConnected();
+    ASSERT_NE(picker, nullptr);
+    EXPECT_EQ(ExpectPickComplete(picker.get()), address);
+    subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
+    ExpectReresolutionRequest();
+    ExpectStateAndQueuingPicker(GRPC_CHANNEL_IDLE);
+  }
+
+  static std::vector<absl::string_view> ExcludeAddress(
```
markdroth, line 112: Doesn't look like this method is needed anymore.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -177,6 +257,76 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+  constexpr std::array<absl::string_view, 6> kAddresses = {
+      "ipv4:127.0.0.1:443", "ipv4:127.0.0.1:444", "ipv4:127.0.0.1:445",
+      "ipv4:127.0.0.1:446", "ipv4:127.0.0.1:447", "ipv4:127.0.0.1:448"};
+  absl::Status status = ApplyUpdate(
+      BuildUpdate(kAddresses, MakePickFirstConfig(true)), lb_policy_.get());
+  EXPECT_TRUE(status.ok()) << status;
+  std::vector<absl::string_view> prev_attempt_connect_order;
+  GetOrderAddressesArePicked(kAddresses, &prev_attempt_connect_order);
+  size_t first_shuffle = 0;
+  for (size_t i = 0; i < kAddresses.size(); ++i) {
```
markdroth, line 274: I don't think we need to iterate over the individual elements; instead, we can just compare the vectors as a whole.  We also don't need to handle the initial attempt differently from the subsequent attempts.  I think we can just do this:

```
std::vector<absl::string_view> prev_order(kAddresses.begin(), kAddresses.end());
for (size_t attempt = 0; attempt < 3; ++attempt) {
  std::vector<absl::string_view> new_order;
  GetOrderAddressesArePicked(kAddresses, &new_order);
  EXPECT_NE(new_order, prev_order);
  prev_order = std::move(new_order);
}
```

Same thing for the next test, except that it will use `EXPECT_EQ()` instead of `EXPECT_NE()`.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -177,6 +257,76 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+      if (address_order[i] != prev_attempt_connect_order[i]) {
+        shuffle_count += 1;
+      }
+    }
+    std::swap(prev_attempt_connect_order, address_order);
+  }
+  ASSERT_GT(shuffle_count, 0) << "Addresses are not reshuffled";
+}
+
+TEST_F(PickFirstTest, DisabledExperiment) {
```
markdroth, line 297: Please call this test `ShufflingDisabled`.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -39,6 +47,78 @@ class PickFirstTest : public LoadBalancingPolicyTest {
...
+
+  // Gets order the addresses are being picked. Return type is void so
+  // assertions can be used
+  void GetOrderAddressesArePicked(
+      absl::Span<const absl::string_view> addresses,
+      std::vector<absl::string_view>* out_address_order) {
+    absl::string_view address;
+    SubchannelState* subchannel = nullptr;
+    out_address_order->clear();
+    while (addresses.size() > out_address_order->size()) {
```
markdroth, line 84: The logic here is written in a way that is a little hard to follow.  For example, reporting TRANSIENT_FAILURE is done at the start of the loop for the subchannel from the previous loop iteration, so it's hard to understand the order in which the actions occur for each subchannel.  Also, we are winding up calling `FindSubchannel()` twice, once in `GetConnectingSubchannel()` and then again here, and we're doing this for every iteration of the loop.

Instead, I suggest constructing a list of subchannels at the start of the function, and removing elements from it as you iterate.  I think this approach results in more readable code:

```
// Construct a map of subchannel to address.
// We will remove entries as each subchannel starts to connect.
std::map<SubchannelState*, absl::string_view> subchannels;
for (auto address : addresses) {
  auto* subchannel = FindSubchannel(
      address, ChannelArgs().Set(GRPC_ARG_INHIBIT_HEALTH_CHECKING, true));
  subchannels.emplace_back(subchannel, address);
}
// Now process each subchannel in the order in which pick_first tries it.
while (!subchannels.empty()) {
  // Find the subchannel that is being attempted.
  SubchannelState* subchannel = nullptr;
  for (const auto& p : subchannels) {
    if (p.first->ConnectionRequested()) {
      out_address_order->push_back(p.second);
      subchannel = p.first;
      break;
    }
  }
  ASSERT_NE(subchannel, "");
  // The subchannel reports CONNECTING.
  // If this is the first subchannel being attempted, expect a CONNECTING update.
  if (subchannels.size() == kAddresses.size()) {
    ExpectConnectingUpdate();
  }
  if (subchannels.size() > 1) {
    // Not the last subchannel in the list.  Connection attempt should fail.
    subchannel->SetConnectivityState(
        GRPC_CHANNEL_TRANSIENT_FAILURE,
        absl::UnavailableError("failed to connect"));
    subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
  } else {
    // Last subchannel in the list.  Connection attempt should succeed.
    auto picker = WaitForConnected();
    ASSERT_NE(picker, nullptr);
    EXPECT_EQ(ExpectPickComplete(picker.get()), out_address_order->back());
    // Then it should become disconnected.
    subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
    ExpectReresolutionRequest();
    ExpectStateAndQueuingPicker(GRPC_CHANNEL_IDLE);
  }
  // Remove the subchannel from the map.
  subchannels.erase(subchannel);
}
```

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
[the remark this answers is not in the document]
markdroth: Looks like we're having some internal discussions about whether we actually want to shuffle the address list when we get the updated address list or when we start connecting.  If it turns out that we actually want to do it when we get the address list, we'll have to make some changes here.  But for now, this looks good!

eugeneo:
Unit test should be fixed now. This PR is ready for review.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -39,6 +50,87 @@ class PickFirstTest : public LoadBalancingPolicyTest {
...
+  static RefCountedPtr<LoadBalancingPolicy::Config> MakePickFirstConfig(
+      bool shuffle_address_list) {
+    return MakeConfig(Json::FromArray({Json::FromObject({{
+        "pick_first",
+        Json::FromObject(
+            {{"shuffleAddressList", Json::FromBool(shuffle_address_list)}}),
+    }})}));
+  }
+
+  absl::string_view GetConnectingSubchannel(
```
markdroth, line 62: This method is no longer needed.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -269,32 +281,26 @@ TEST_F(PickFirstTest, WithShuffle) {
...
-      first_shuffle += 1;
-    }
-  }
   // There is 0.1% chance this check fails by design. Not an assert to prevent
   // flake
-  EXPECT_GT(first_shuffle, 0) << "Addresses were not shuffled";
-  size_t shuffle_count = 0;
-  // 3 attempts to reduce flakes. Probability of a flake is 0.001%
-  for (size_t attempt = 0; attempt < 3; ++attempt) {
+  if (absl::MakeConstSpan(prev_attempt_connect_order) ==
```
markdroth, line 286: This block doesn't seem useful.  Please remove.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -177,6 +269,61 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+      "GRPC_EXPERIMENTAL_PICKFIRST_LB_CONFIG");
+  // 6 addresses have 6! = 720 permutations or 0.1% chance that the shuffle
+  // returns initial sequence
+  constexpr std::array<absl::string_view, 6> kAddresses = {
+      "ipv4:127.0.0.1:443", "ipv4:127.0.0.1:444", "ipv4:127.0.0.1:445",
+      "ipv4:127.0.0.1:446", "ipv4:127.0.0.1:447", "ipv4:127.0.0.1:448"};
+  absl::Status status = ApplyUpdate(
+      BuildUpdate(kAddresses, MakePickFirstConfig(true)), lb_policy_.get());
+  EXPECT_TRUE(status.ok()) << status;
+  std::vector<absl::string_view> prev_attempt_connect_order;
```
markdroth, line 283: I think we need to initialize this to the order from `kAddresses`.  Otherwise, the condition on line 295 below will always evaluate to true on the very first attempt, which means that this test will pass even if shuffling is not working properly.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -177,6 +269,61 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+  }
+  constexpr size_t kMaxAttempts = 5;
+  bool shuffled = false;
+  for (size_t attempt = 0; attempt < kMaxAttempts && !shuffled; ++attempt) {
+    std::vector<absl::string_view> address_order;
+    GetOrderAddressesArePicked(kAddresses, &address_order);
+    if (address_order != prev_attempt_connect_order) {
+      shuffled = true;
+    }
+    std::swap(prev_attempt_connect_order, address_order);
```
markdroth, line 298: This can just be:

```
prev_attempt_connect_order = std::move(address_order);
```

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -177,6 +269,61 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+  std::vector<absl::string_view> prev_attempt_connect_order;
+  // There is 0.1% chance this check fails by design. Not an assert to prevent
+  // flake
+  if (absl::MakeConstSpan(prev_attempt_connect_order) ==
+      absl::MakeConstSpan(kAddresses)) {
+    gpr_log(GPR_INFO, "Address order did not change");
+  }
+  constexpr size_t kMaxAttempts = 5;
+  bool shuffled = false;
+  for (size_t attempt = 0; attempt < kMaxAttempts && !shuffled; ++attempt) {
```
markdroth, line 292: Instead of checking `!shuffled` here, please add a `break` on line 297 below.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -177,6 +269,61 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+}
+
+TEST_F(PickFirstTest, ShufflingDisabled) {
+  constexpr std::array<absl::string_view, 6> kAddresses = {
+      "ipv4:127.0.0.1:443", "ipv4:127.0.0.1:444", "ipv4:127.0.0.1:445",
+      "ipv4:127.0.0.1:446", "ipv4:127.0.0.1:447", "ipv4:127.0.0.1:448"};
+  absl::Status status = ApplyUpdate(
+      BuildUpdate(kAddresses, MakePickFirstConfig(true)), lb_policy_.get());
+  EXPECT_TRUE(status.ok()) << status;
+  std::vector<absl::string_view> prev_attempt_connect_order;
```
markdroth, line 310: I don't think we need `prev_attempt_connect_order` at all in this test.  Instead, we can just check that the connection order in every iteration of the loop is the same as `kAddresses`:

```
for (size_t attempt = 0; attempts < 5; ++attempts) {
  std::vector<absl::string_vewi> address_order;
  GetOrderAddressesArePicked(kAddresses, &address_order);
  EXPECT_THAT(address_order, ::testing::ElementsAreArray(kAddresses));
}
```

test/core/util/test_lb_policies.cc
```diff
@@ -78,6 +78,15 @@ class ForwardingLoadBalancingPolicy : public LoadBalancingPolicy {
   ~ForwardingLoadBalancingPolicy() override = default;
 
   absl::Status UpdateLocked(UpdateArgs args) override {
+    // Use correct config for the delegate load balancing policy
+    auto config =
+        grpc_core::CoreConfiguration::Get()
+            .lb_policy_registry()
+            .ParseLoadBalancingConfig(Json::FromArray({Json::FromObject(
+                {{std::string(delegate_->name()), Json::FromObject({})}})}));
+    if (config.ok()) {
```
markdroth, line 87: This should be a `GPR_ASSERT()`.  If it doesn't parse successfully, then we still don't want to pass the parent's config down; we should crash in that case.

eugeneo: Done

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done. Pre-rval ref habits 😀

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Oh! There was a bug where I did not fetch address order on first try. I believe this test needs to separately check initial shuffle and subsequent shuffles.

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: It will log if initial shuffle did not work.

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done

eugeneo (commented):
Thank you for the review! Updated.

markdroth:
In the future, olease don't force-push to a PR once a review has started.  That makes it extremely difficult for the reviewer to tell what has changed since the last review pass.  Thanks!

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -218,6 +261,13 @@ void PickFirst::AttemptToConnectUsingLatestUpdateArgsLocked() {
   if (latest_update_args_.addresses.ok()) {
     addresses = *latest_update_args_.addresses;
   }
+  if (latest_update_args_.config != nullptr) {
```
markdroth, line 264: As per out internal conversation, it looks like we actually want to move this back to `UpdateLocked()`.  Looks like this can go in an `else` block on line 277.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -177,6 +251,54 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
 
+TEST_F(PickFirstTest, WithShuffle) {
+  testing::ScopedExperimentalEnvVar env_var(
+      "GRPC_EXPERIMENTAL_PICKFIRST_LB_CONFIG");
+  // 6 addresses have 6! = 720 permutations or 0.1% chance that the shuffle
+  // returns initial sequence
+  constexpr std::array<absl::string_view, 6> kAddresses = {
+      "ipv4:127.0.0.1:443", "ipv4:127.0.0.1:444", "ipv4:127.0.0.1:445",
+      "ipv4:127.0.0.1:446", "ipv4:127.0.0.1:447", "ipv4:127.0.0.1:448"};
+  absl::Status status = ApplyUpdate(
```
markdroth, line 262: Now that we're going to do the address shuffling in `UpdateLocked()` instead of in `AttemptToConnectUsingLatestUpdateArgsLocked()`, I think we can just move this `ApplyUpdate()` call inside of `GetOrderAddressesArePicked()`, so that we trigger a new shuffle each time we call it.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -39,6 +49,70 @@ class PickFirstTest : public LoadBalancingPolicyTest {
...
+      if (subchannels.size() == addresses.size()) {
+        ExpectConnectingUpdate();
+      }
+      if (subchannels.size() > 1) {
+        // Not the last subchannel in the list.  Connection attempt should fail.
+        subchannel->SetConnectivityState(
+            GRPC_CHANNEL_TRANSIENT_FAILURE,
+            absl::UnavailableError("failed to connect"));
+        subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
+      } else {
```
markdroth, line 100: I think we should remove this `else` block.  We can just leave the last subchannel trying to connect, since the policy will stop using it as soon as we deliver a new address list.

eugeneo: I left it so we test that the addresses are not reshuffled.

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: I did it slightly different so GetOrderAddressesArePicked can be used to verify addresses are not reshuffled if all subchannels fail

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: This is all changed now.

eugeneo (commented):
Thank you for the review. Please review new changes.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -254,45 +255,45 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
-    std::vector<absl::string_view> address_order;
-    GetOrderAddressesArePicked(kAddresses, &address_order);
-    if (address_order != prev_attempt_connect_order) {
-      shuffled = true;
+  // 6 addresses have 6! = 720 permutations or roughly 0.14% chance that
+  // the shuffle returns same permutation. We allow for several tries to
+  // prevent flake test.
+  constexpr size_t kMaxTries = 10;
+  std::vector<absl::string_view> addresses_after_update;
+  for (size_t i = 0; i < kMaxTries; i++) {
```
markdroth, line 266: Prefer preincrement over postincrement.

https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -336,6 +329,12 @@ absl::Status PickFirst::UpdateLocked(UpdateArgs args) {
           DisableOutlierDetectionAttribute::kName,
           std::make_unique<DisableOutlierDetectionAttribute>()));
     }
+    if (args.config != nullptr) {
+      auto config = static_cast<PickFirstConfig*>(args.config.get());
```
markdroth, line 333: I'd rather not put this code in this block, because this block is going to go away in the not-too-distant future, as per the TODO comment on line 321.  Instead, please put this code in an `else` clause on line 320.  (Sorry, looks like I wrote the wrong line number in my previous comment.)

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -254,45 +255,45 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+    GetOrderAddressesArePicked(kAddresses, &addresses_after_update);
+    if (absl::MakeConstSpan(addresses_after_update) !=
+        absl::MakeConstSpan(kAddresses)) {
       break;
     }
-    prev_attempt_connect_order = std::move(address_order);
   }
-  ASSERT_TRUE(shuffled) << "Addresses are not reshuffled";
+  // Same elements, different order
+  ASSERT_THAT(addresses_after_update,
```
markdroth, line 277: These two checks seem unnecessary.  Instead, why not just set a bool on line 273 before we break out of the loop, and then check the bool here?

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -254,45 +255,45 @@ TEST_F(PickFirstTest, GoesIdleWhenConnectionFailsThenCanReconnect) {
...
+              ::testing::Not(::testing::ElementsAreArray(kAddresses)));
+  ASSERT_THAT(addresses_after_update,
+              ::testing::UnorderedElementsAreArray(kAddresses));
+  // Address order should be stable between updates
+  std::vector<absl::string_view> addresses_on_another_try;
+  GetOrderAddressesArePicked(kAddresses, &addresses_on_another_try);
+  ASSERT_THAT(addresses_on_another_try,
```
markdroth, line 284: This should be an `EXEPECT_` instead of an `ASSERT_`.  And it can use `EXPECT_EQ()` instead of using `ContainerEq()`.

test/core/client_channel/lb_policy/pick_first_test.cc
markdroth: I don't think this is going to work reliably.  The test is going to be flaky.

The problem is that when we call `ExpectStateAndQueuingPicker(GRPC_CHANNEL_IDLE)` for the last subchannel, doing the pick will trigger the LB policy to exit idle state and start trying to connect again.  It will start with the first subchannel in the *current* order of the addresses, so `ConnectionRequested()` will return true for that subchannel even before we send a new address list update to the LB policy, which will cause it to shuffle again.  So the next time we call `GetOrderAddressesArePicked()`, we may see the connection attempts in a different order than the new shuffled order.

For example, consider the following sequence of events:
1. We send the LB policy an address update containing [A, B, C].  The LB policy shuffles to the order [C, B, A].
1. On the first call to `GetOrderAddressesArePicked()`, we see the shuffled order [C, B, A].  We fail the connection attempts to C and B, and then we succeed connecting to A.  Then we have A report IDLE, which causes the policy to report IDLE.  Now we request a pick, which triggers the policy to start trying to connect again.  It will request a connection on C, since that's the first address in the current order.  At this point, the subchannel for C will return true for `ConnectionRequested()`, but we haven't yet checked it.
2. Now we send the LB policy a new address list update.  This time, it shuffles the addresses into order [B, A, C].  It will now immediately drop the old subchannel list and switch to the new one, and it will start connecting to address B.  This means the subchannel for B will return true for `ConnectionRequested()`.
6. In the test, we now call `GetOrderAddressesArePicked()` again.  At this point, the subchannels for both addresses B and C will return true for `ConnectionRequested()`.  We want the test to see B first, since that's the first entry in the shuffled address list, but it may happen to see C first instead, thus returning a different order than what the LB policy is actually using.  (The order in which we check the addresses here depends on the addresses that happen to be used to allocate the `SubchannelState` objects, since that address is the key of the map on line 69.)

To solve this, I think you need to add a parameter to this function indicating whether we want to finish the connection attempt pass.  In the case where we are going to send another address update after this one, we should avoid doing a pick after the LB policy reports IDLE, so that it does not start trying to connect again.  In other words, we want something like this on line 110:

```
if (exit_idle_after_connection_fails) {
  ExpectStateAndQueuingPicker(GRPC_CHANNEL_IDLE);
} else {
  ExpectState(GRPC_CHANNEL_IDLE);
}
```

But this also means that we need to trigger the LB policy to exit idle *after* we deliver the subsequent address list update to it.  So we'll probably need to add code at the top of this function to hop into the `WorkSerializer` and call the LB policy's `ExitIdleLocked()` method.

eugeneo: Done

eugeneo (commented):
Thank you for the reviews!

test/core/client_channel/lb_policy/pick_first_test.cc
markdroth: I think that `GetOrderAddressesArePicked()` already guarantees that all addresses will be present, because it removes the subchannel for each address from its map as it sees that subchannel connect, and it keeps going until the map is empty.  So I think this check is redundant.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -292,10 +308,12 @@ TEST_F(PickFirstTest, ShufflingDisabled) {
   constexpr static size_t kMaxAttempts = 5;
   for (size_t attempt = 0; attempt < kMaxAttempts; ++attempt) {
     absl::Status status = ApplyUpdate(
-        BuildUpdate(kAddresses, MakePickFirstConfig(true)), lb_policy_.get());
+        BuildUpdate(kAddresses, MakePickFirstConfig(false)), lb_policy_.get());
```
markdroth, line 311: This needs to pass `true` to `MakePickFirstConfig()`.  The goal of this test is to show that if the env var is not set, then the config parameter is ignored.

Please rename the test to `ShufflingDisabledViaEnvVar` to make this clear.  Also, please add a TODO indicating that this test should be removed when the env var protection is removed.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -292,10 +308,12 @@ TEST_F(PickFirstTest, ShufflingDisabled) {
...
+    work_serializer_->Run([&]() { lb_policy_->ExitIdleLocked(); },
+                          DEBUG_LOCATION);
     EXPECT_TRUE(status.ok()) << status;
     std::vector<absl::string_view> address_order;
-    GetOrderAddressesArePicked(kAddresses, &address_order);
+    GetOrderAddressesArePicked(kAddresses, true, &address_order);
```
markdroth, line 316: We need to set `exit_idle_after_connection_fails` to false here.  It always needs to be set to false if we're going to do another pass after the current one.

If this is going to be too error prone, maybe we should just remove the `exit_idle_after_connection_fails` parameter and always use `ExpectState(GRPC_CHANNEL_IDLE)` instead of `ExpectStateAndQueueingPicker(GRPC_CHANNEL_IDLE)`.  That will skip checking the picker, but maybe that's okay in this case, since we already cover that case in the `GoesIdleWhenConnectionFailsThenCanReconnect` test.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -263,26 +275,30 @@ TEST_F(PickFirstTest, WithShuffle) {
   // prevent flake test.
   constexpr size_t kMaxTries = 10;
   std::vector<absl::string_view> addresses_after_update;
-  for (size_t i = 0; i < kMaxTries; i++) {
+  bool shuffled = false;
+  for (size_t i = 0; i < kMaxTries; ++i) {
     absl::Status status = ApplyUpdate(
         BuildUpdate(kAddresses, MakePickFirstConfig(true)), lb_policy_.get());
+    work_serializer_->Run([&]() { lb_policy_->ExitIdleLocked(); },
```
markdroth, line 282: Suggest moving this inside of `GetOrderAddressesArePicked()`, so that we don't need to duplicate it in 3 places.

markdroth (commented):
Just a couple of remaining issues here.

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done

eugeneo: (tried multiple runs, seems perfectly stable)

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: There's now only one check here... I would like to have an explicit check, mostly to protect from an error when writing the test.

test/core/client_channel/lb_policy/pick_first_test.cc
markdroth: The parameter still needs to be set to false.

The problem here is not instability, it's that the test is not actually covering what it should.  The reason that you're not seeing any flakiness is that the first subchannel that the LB policy tries to connect when we exit idle is the same one that it will always try first after the next address list update.  But if the LB policy failed to try that particular address first after the update, the test could still pass, and in that case we would not want it to.

At this point, I think it's clear that we should remove the `exit_idle_after_connection_fails` parameter and always use `ExpectState(GRPC_CHANNEL_IDLE)` instead of `ExpectStateAndQueueingPicker(GRPC_CHANNEL_IDLE)`.  And we should add a comment indicating that we're explicitly *not* using the latter, because we don't want to trigger exiting idle until after the next address list update, if any.  Otherwise, the subtleties of the behavior here are going to be very unclear to future engineers who look at this code.

test/core/client_channel/lb_policy/pick_first_test.cc
markdroth: I don't see how `GetOrderAddressesArePicked()` could produce any other result and still do its job, so I'm not sure what kind of error we're protecting against here.  And if we are worried about that, why check this only after the last iteration of the loop?  Wouldn't you want to check this inside the loop, so that we know that `GetOrderAddressesArePicked()` returns the full set every time?

This seems odd and not useful to me, but I'll leave it up to you.

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Done.

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Removed.

eugeneo (commented):
Thank you for the reviews. Please take another look.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -114,11 +112,8 @@ class PickFirstTest : public LoadBalancingPolicyTest {
         // Then it should become disconnected.
         subchannel->SetConnectivityState(GRPC_CHANNEL_IDLE);
         ExpectReresolutionRequest();
-        if (exit_idle_after_connection_fails) {
-          ExpectStateAndQueuingPicker(GRPC_CHANNEL_IDLE);
-        } else {
-          ExpectState(GRPC_CHANNEL_IDLE);
-        }
+        // Do not trigger exit idle.
```
markdroth, line 115: This comment isn't sufficiently clear.  I suggest the following wording:

"""
We would normally call ExpectStateAndQueueingPicker() here instead of just ExpectState().  However, calling the picker would also trigger exiting IDLE, which we don't want here, because if the test is going to send an address list update and call GetOrderAddressesArePicked() again, we don't want to trigger a connection attempt on any subchannel until after that next address list update is processed.
"""

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -313,8 +305,7 @@ TEST_F(PickFirstTest, ShufflingDisabled) {
         BuildUpdate(kAddresses, MakePickFirstConfig(false)), lb_policy_.get());
     EXPECT_TRUE(status.ok()) << status;
     std::vector<absl::string_view> address_order;
-    GetOrderAddressesArePicked(kAddresses, true, &address_order);
-    EXPECT_THAT(address_order, ::testing::ElementsAreArray(kAddresses));
```
markdroth: This expectation is still needed.  The whole point of this test is to verify that the address list was *not* shuffled.  Without this expectation, the test is not doing that.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -329,8 +320,7 @@ TEST_F(PickFirstTest, ShufflingDisabledViaEnvVar) {
         BuildUpdate(kAddresses, MakePickFirstConfig(true)), lb_policy_.get());
     EXPECT_TRUE(status.ok()) << status;
     std::vector<absl::string_view> address_order;
-    GetOrderAddressesArePicked(kAddresses, false, &address_order);
-    EXPECT_THAT(address_order, ::testing::ElementsAreArray(kAddresses));
```
markdroth: This expectation is also still needed, for the same reason.

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Reverted

test/core/client_channel/lb_policy/pick_first_test.cc
eugeneo: Reverted

eugeneo (commented):
Thank you! Updated

markdroth (approved):
Looks great!

PR #33257, opened 2023-05-26
[chaotic-good] Implement a promise-based endpoint for chaotic-good transport to read & write to EventEngine::Endpoint.

nanahpang:
This PR is continuing the work of prototyping in https://github.com/grpc/grpc/pull/31592, and the design doc is at [link](https://docs.google.com/document/d/1vRy0yse-d1heLQRmLPo_0figsTPXJAnNN84tBCAne_s/edit?pli=1&resourcekey=0-JvUPdq0LaZq8gMkgT9Pzlw#heading=h.qgvc5vr55ytg).


src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+    };
+  }
+
+  // Returns a promise that resolves to a byte with type `uint8_t`.
+  auto ReadByte() {
+    ReleasableMutexLock lock(&read_mutex_);
+
+    // Previous read result has not been polled.
+    GPR_ASSERT(!read_result_.has_value());
+
```
ctiller, line 217: remove blank line

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+    // Should not have pending reads.
+    GPR_ASSERT(pending_read_buffer_.Count() == 0u);
+
```
ctiller, line 220: remove blank line

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+    if (read_buffer_.Length() == 0u) {
+      lock.Release();
+
```
ctiller, line 223: remove blank line

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+                          &pending_read_buffer_, nullptr)) {
+        ReadByteCallback(absl::OkStatus());
+      }
+    } else {
+      read_result_ = absl::OkStatus();
+    }
+
+    return [this]() -> Poll<absl::StatusOr<uint8_t>> {
+      MutexLock lock(&read_mutex_);
+      if (!read_result_.has_value()) {
```
ctiller, line 238: comment these if/then cases

src/core/lib/transport/promise_endpoint.cc
```diff
@@ -0,0 +1,132 @@
...
+#include "src/core/lib/slice/slice_buffer.h"
+
+namespace grpc_core {
+
+PromiseEndpoint::PromiseEndpoint(
+    std::unique_ptr<grpc_event_engine::experimental::EventEngine::Endpoint>
+        endpoint,
+    SliceBuffer already_received)
+    : endpoint_(std::move(endpoint)),
+      write_buffer_(),
```
ctiller, line 41: no need for these default initializers

src/core/lib/transport/promise_endpoint.cc
```diff
@@ -0,0 +1,132 @@
...
+      write_result_(),
+      pending_read_buffer_(),
+      read_result_() {
+  GPR_ASSERT(endpoint_ != nullptr);
+
```
ctiller, line 46: remove blank line

src/core/lib/transport/promise_endpoint.cc
```diff
@@ -0,0 +1,132 @@
...
+void PromiseEndpoint::ReadCallback(
+    absl::Status status, size_t num_bytes_requested,
+    absl::optional<
+        struct grpc_event_engine::experimental::EventEngine::Endpoint::ReadArgs>
+        requested_read_args) {
+  if (!status.ok()) {
+    // Invalidates all previous reads.
+    pending_read_buffer_.Clear();
+    read_buffer_.Clear();
+
```
ctiller, line 85: remove blank line

test/core/transport/promise_endpoint_test.cc
```diff
@@ -0,0 +1,1389 @@
...
+#include "src/core/lib/slice/slice_internal.h"
+#include "test/core/promise/test_wakeup_schedulers.h"
+
+using testing::MockFunction;
+using testing::StrictMock;
+
+namespace grpc_core {
+namespace testing {
+
+class MockEndpoint
```
ctiller, line 52: It's unclear to me what we're gaining by this being a mock.

At present (though I could be convinced otherwise) I think we'd be better off having a `FakeEndpoint` - still deriving from `grpc_event_engine::experimental::EventEngine::Endpoint` but directly implementing Read, Write, etc.

The rest of the classes working would be the same, but we'd drop a bunch of the expect calls in tests and end up with a more readable thing.

ctiller (changes requested):
This is looking very good!

Lots of comments, but mostly nits.

It'd be good to convince ourselves that the MockEndpoint is the right testing strategy prior to submission, and prepare for an iteration later improving the locking story.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+
+#include "src/core/lib/gprpp/sync.h"
+#include "src/core/lib/promise/activity.h"
+#include "src/core/lib/promise/poll.h"
+#include "src/core/lib/slice/slice.h"
+#include "src/core/lib/slice/slice_buffer.h"
+
+namespace grpc_core {
+
+class PromiseEndpoint {
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+
+  // Returns a promise that resolves to `Slice` with at least
+  // `num_bytes` bytes which should be less than INT64_MAX bytes.
+  //
+  // Concurrent reads are not supported, which means callers should not call
+  // `ReadSlice()` before the previous read finishes. Doing that results in
+  // undefined behavior.
+  auto ReadSlice(size_t num_bytes) {
+    ReleasableMutexLock lock(&read_mutex_);
+    if (num_bytes >= static_cast<size_t>(std::numeric_limits<int64_t>::max())) {
```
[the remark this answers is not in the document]
nanahpang: Removed, thanks for catching this.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+  // Concurrent writes are not supported, which means callers should not call
+  // `Write()` before the previous write finishes. Doing that results in
+  // undefined behavior.
+  auto Write(SliceBuffer data) {
+    {
+      MutexLock lock(&write_mutex_);
+
+      // Previous write result has not been polled.
+      GPR_ASSERT(!write_result_.has_value());
+
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+  // Returns a promise that resolves to `SliceBuffer` with
+  // `num_bytes` bytes.
+  //
+  // Concurrent reads are not supported, which means callers should not call
+  // `Read()` before the previous read finishes. Doing that results in
+  // undefined behavior.
+  auto Read(size_t num_bytes) {
+    ReleasableMutexLock lock(&read_mutex_);
+
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+    // Previous read result has not been polled.
+    GPR_ASSERT(!read_result_.has_value());
+
+    // Should not have pending reads.
+    GPR_ASSERT(pending_read_buffer_.Count() == 0u);
+
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+                                    std::placeholders::_1, num_bytes,
+                                    absl::nullopt /* uses default arguments */),
+                          &pending_read_buffer_,
+                          nullptr /* uses default arguments */)) {
+        ReadCallback(absl::OkStatus(), num_bytes, absl::nullopt);
+      }
+    } else {
+      read_result_ = absl::OkStatus();
+    }
+
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+    if (num_bytes >= static_cast<size_t>(std::numeric_limits<int64_t>::max())) {
+      read_result_ = absl::Status(
+          absl::StatusCode::kInvalidArgument,
+          absl::StrFormat(
+              "Requested size is bigger than the maximum supported size %lld.",
+              std::numeric_limits<int64_t>::max()));
+    } else {
+      // Previous read result has not been polled.
+      GPR_ASSERT(!read_result_.has_value());
+
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+      // Should not have pending reads.
+      GPR_ASSERT(pending_read_buffer_.Count() == 0u);
+
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+      if (read_buffer_.Length() < num_bytes) {
+        lock.Release();
+        const struct grpc_event_engine::experimental::EventEngine::Endpoint::
+            ReadArgs read_args = {static_cast<int64_t>(num_bytes)};
+
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.cc
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
nanahpang: Done.

src/core/lib/transport/promise_endpoint.cc
nanahpang: Removed, thanks for catching this.

test/core/transport/promise_endpoint_test.cc
nanahpang: I see, by less expect calls you mean remove the ones that related to the `EventEngine::Endpoint` mocks or in general, the expect calls that are redundant?

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,282 @@
...
+    if (read_buffer_.Length() < num_bytes) {
+      lock.Release();
+      // If `Read()` returns true immediately, the callback will not be
+      // called. We still need to call our callback to pick up the result and
+      // maybe do further reads.
+      if (endpoint_->Read(std::bind(&PromiseEndpoint::ReadCallback, this,
+                                    std::placeholders::_1, num_bytes,
+                                    absl::nullopt /* uses default arguments */),
+                          &pending_read_buffer_,
+                          nullptr /* uses default arguments */)) {
```
Vignesh2208, line 113: Probably can be improved by providing the EventEngine::Endpoint::ReadArgs optional argument with num_bytes as hint. Some EE implementations (e.g Posix) will ensure that the Read doesn't complete internally without reading the  minimum required number of bytes.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,282 @@
...
+      } else if (!read_result_->ok()) {
+        const auto ret = *read_result_;
+        read_result_.reset();
+        return ret;
+      } else if (read_buffer_.RefSlice(0).size() == num_bytes) {
+        read_result_.reset();
+        return Slice(read_buffer_.TakeFirst().TakeCSlice());
+      } else {
+        MutableSlice ret = MutableSlice::CreateUninitialized(num_bytes);
+        read_buffer_.MoveFirstNBytesIntoBuffer(num_bytes, ret.data());
```
Vignesh2208, line 181: Probably for the future: This will do a malloc and memcpy to copy the `num_bytes` into a slice. Depending on where this ReadSlice method is going to be invoked (for e.g in the critical path), we should think about some optimization to avoid the memcpy

src/core/lib/transport/promise_endpoint.h
nanahpang: Done.

test/core/transport/promise_endpoint_test.cc
nanahpang: As discuss offline, I update the test with mock methods only, and remove the fake endpoint logics.

src/core/lib/transport/promise_endpoint.h
nanahpang: Done, thanks for the suggestion.

src/core/lib/transport/promise_endpoint.h
nanahpang: Thanks for catching this, I added a TODO comment.
So If the `read_buffer_ `slice size is equal to the requested `num_bytes`, then there won't be a memcpy. But if the `read_buffer_` slice size is not equal to the requested `num_bytes`, it needs to copy the buffer to a slice that has size equal to `num_bytes`. 

Is there an existing function that could return any size Slice from the SliceBuffer based on the start location? If not, I think we will need to add one in SliceBuffer to avoid this memcpy.

Vignesh2208: I dont think there is an easy way to do this based on the current grpc_slice_buffer and grpc_slice implementations. Lets try to avoid using this method in the critical sections and if that is not possible, we could discuss some alternatives.

Also nit: I think you also need to handle the case where read_buffer_.Size() < num_bytes. Its not always the case that EventEngine endpoints will return at least num_bytes.

nanahpang: I see, we could use `Read(num_bytes)` in the transport implement to avoid memcpy. 

>I think you also need to handle the case where read_buffer_.Size() < num_bytes. Its not always the case that EventEngine endpoints will return at least num_bytes.

Yes, there's a `GPR_ASSERT(src->length >= n);` check in `MoveFirstNBytesIntoBuffer` method. Or should I handle `read_buffer_.Size() < num_bytes` case in Promise Endpoint separately? I was expecting it to throw an assert failed error with `read_buffer_.Size() < num_bytes`, but if it isn't the best practice, I could try to return a `absl::Status(absl::StatusCode::kInternal, absl::StrFormat( "EventEngine read buffer size %d is less than the request bytes %d.", read_buffer_.Size(), num_bytes))`

Vignesh2208: I think you'd need to handle the read_buffer_.Size() < num_bytes case similar to how it is handled in the Read(num_bytes) method. It is not guaranteed that the event engine endpoint will always return at least num_bytes amount of data. Some event engines may ignore the ReadArgs passed to the endpoint (e.g, windows, ios).

nanahpang: I see, actually `ReadSlice(num_bytes)` is handling `read_buffer_.Size() < num_bytes` same as the `Read(num_bytes)`. It first checks if `read_buffer_.Size() < num_bytes`, if true then do event engine endpoint::Read, and then schedule the `ReadCallback`, and continuously check if `read_buffer_.Size() < num_bytes` in `ReadCallback`.  So at the end `read_buffer_.Size()` is expected to be greater or equal to num_bytes.

Vignesh2208: Oh I missed it. thanks for the clarification.

Vignesh2208 (approved):
Thanks for working on this!

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+          ReadCallback(absl::OkStatus(), num_bytes, read_args);
+        }
+      } else {
+        read_result_ = absl::OkStatus();
+      }
+    }
+
+    return [this, num_bytes]() -> Poll<absl::StatusOr<Slice>> {
+      MutexLock lock(&read_mutex_);
+      if (!read_result_.has_value()) {
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,305 @@
...
+                          nullptr /* uses default arguments */)) {
+        ReadCallback(absl::OkStatus(), num_bytes, absl::nullopt);
+      }
+    } else {
+      read_result_ = absl::OkStatus();
+    }
+
+    return [this, num_bytes]() -> Poll<absl::StatusOr<SliceBuffer>> {
+      MutexLock lock(&read_mutex_);
+      if (!read_result_.has_value()) {
```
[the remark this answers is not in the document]
nanahpang: Done.

ctiller (approved):

ctiller:
Looks like a clang-tidy lint to fix before submission:

```
test/core/transport/promise_endpoint_test.cc:41:16: error: using decl '_' is unused [misc-unused-using-decls,-warnings-as-errors]
using testing::_;
               ^
/root/.cache/bazel/_bazel_root/d2dc70c3d9da3fab488ba0dcbbd35051/execroot/com_github_grpc_grpc/test/core/transport/promise_endpoint_test.cc:41:16: note: remove the using
using testing::_;
~~~~~~~~~~~~~~~^~
```

nanahpang:
Thanks for all the reviews!

src/core/lib/transport/promise_endpoint.h
```diff
@@ -0,0 +1,297 @@
...
+      // Set read args with hinted bytes.
+      grpc_event_engine::experimental::EventEngine::Endpoint::ReadArgs
+          read_args;
+      read_args.read_hint_bytes = num_bytes;
+      // If `Read()` returns true immediately, the callback will not be
+      // called. We still need to call our callback to pick up the result and
+      // maybe do further reads.
+      if (endpoint_->Read(std::bind(&PromiseEndpoint::ReadCallback, this,
+                                    std::placeholders::_1, num_bytes,
+                                    absl::nullopt /* uses default arguments */),
```
ralphchung, line 117: The current implementation makes `requested_read_args` in `PromiseEndpoint::ReadCallback()` useless. We probably should consider clean it up.

The role of `requested_read_args` in the prototype PR is based on the assumption that the `num_byte` in this function could be large so that a single call to `endpoint_->Read()` will not be enough in most cases. Smaller `num_byte` should call `ReadSlice()` instead. The argument `requested_read_args` is the only difference for `Read()` and `ReadSlice()` when creating the lambda in the prototype PR.

Seeing the implementation here, I am a little confused of the difference between `Read()` and `ReadSlice()`.

nanahpang: @Vignesh2208 correct me if I am wrong, from my understanding, read_hint_bytes is a good to have setting, but not guarantee that EventEngine will read all the bytes at once? (Also, I should use `read_args` instead of `absl::nullopt` in the bind function as well, I will send a following PR to fix this.)

I think the difference between Read() and ReadSlice() is mostly the return type and use scenarios. And since ReadSlice() will do a copy from SliceBuffer to Slice, we will probably avoid using it for now until we find a better solution to remove the memcpy issue.

ralphchung: If we always have `read_hint_bytes`, we don't need to have the argument in the callback. You may clean that up in the following-up PR as well.

nanahpang: Oh, I see what you meant now. Yeah, the parameter of `requested_read_args` in ReadCallback isn't useful anymore, I will clean it up.

PR #32847, opened 2023-04-11
[core] Add support for vsock transport

YadongQi:
This is another attempt to add support for vsock in grpc since previous
PRs(#24551, #21745) all closed without merging.

The VSOCK address family facilitates communication between
virtual machines and the host they are running on.
This patch will introduce new scheme: [vsock:cid:port] to
support VSOCK address family.

Fixes #32738.

PR #33276, opened 2023-05-30
Revert "[core] Add support for vsock transport"

drfloob:
Reverts grpc/grpc#32847. Breaks the Python build.


drfloob:
CC @YadongQi

PR #33278, opened 2023-05-30
[interop] fix C++ server to add mutex for orca_oob test

eugeneo:
Following improvements were made to `orca_oob` interop test to increase compatibility and stability:

1. Timeout was increased to 10s (see #33098)
2. Server will block clients trying to run this test concurrently.
3. Data is cleared in the beginning of the call.

### DIFF
--- a/test/cpp/interop/interop_client.cc
+++ b/test/cpp/interop/interop_client.cc
@@ -1021,6 +1021,7 @@ bool InteropClient::DoOrcaPerRpc() {
 }
 
 bool InteropClient::DoOrcaOob() {
+  static constexpr auto TIMEOUT = absl::Seconds(10);
   gpr_log(GPR_DEBUG, "testing orca oob");
   load_report_tracker_.ResetCollectedLoadReports();
   grpc_core::CoreConfiguration::RegisterBuilder(RegisterBackendMetricsLbPolicy);
@@ -1059,7 +1060,7 @@ bool InteropClient::DoOrcaOob() {
                          }
                          return true;
                        },
-                       absl::Seconds(5), 10)
+                       TIMEOUT, 10)
                    .has_value());
   }
   {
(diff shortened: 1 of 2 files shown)


test/cpp/interop/interop_client.cc
```diff
@@ -1021,6 +1021,7 @@ bool InteropClient::DoOrcaPerRpc() {
 }
 
 bool InteropClient::DoOrcaOob() {
+  static constexpr auto TIMEOUT = absl::Seconds(5);
```
markdroth, line 1024: This should be named `kTimeout`, as per https://google.github.io/styleguide/cppguide.html#Constant_Names.

markdroth, line 1024: #33098 says this should be 10s, not 5s.

test/cpp/interop/interop_server.cc
```diff
@@ -296,6 +296,7 @@ class TestServiceImpl : public TestService::Service {
     StreamingOutputCallRequest request;
     StreamingOutputCallResponse response;
     bool write_success = true;
+    std::unique_ptr<grpc_core::MutexLock> orca_oob_lock;
```
markdroth, line 299: Suggest using `absl::optional<>` instead of `std::unique_ptr<>`.  There's no need for dynamic allocation here.

markdroth (approved):
All comments are minor; feel free to merge after addressing.  Thanks!

test/cpp/interop/interop_server.cc
eugeneo: Compiler outputs a warning about releasing the mutex that was not held if `absl::optional` is used here.

test/cpp/interop/interop_client.cc
eugeneo: Done

eugeneo: Done

Issue #33218, opened 2023-05-23
Waiting for thread pool to idle before forking


hyerra:
### What version of gRPC and what language are you using?
gRPC: 1.55.0
Python: 3.10.0

### What operating system (Linux, Windows,...) and version?
macOS 13.2.1 (22D68)

### What runtime / compiler are you using (e.g. python version or version of gcc)
Python: 3.10.0

### What did you do?
I'm running into an issue where my application doesn't run and I only see this in the console:
```
E0523 12:35:55.355078000 8402944320 thread_pool.cc:230]                Waiting for thread pool to idle before forking
E0523 12:35:58.360160000 8402944320 thread_pool.cc:230]                Waiting for thread pool to idle before forking
E0523 12:36:01.362086000 8402944320 thread_pool.cc:230]                Waiting for thread pool to idle before forking
```

I saw this issue was raised before, and you can follow the same steps outlined in this issue to reproduce the problem: https://github.com/grpc/grpc/issues/31885.

### What did you expect to see?
My application should run without encountering this issue.

### What did you see instead?
I got the log messages above.

### Anything else we should know about your project / environment?
I think the latest release published last night (1.55.0) located [here](https://pypi.org/project/grpcio/#history) broke my project. Downgrading to version 1.54.2 fixed everything.

XuanWang-Amos:
Sorry about the breakage, t There are still some fork-related issues that we need to resolve (mentioned in #31885), but they should also be present in 1.54.2. Could you try to reproduce the issue in 1.55.0rc1?

Additionally, could you please share more logs? You can add those environment variables so that we can have more detailed logs:
`GRPC_VERBOSITY=debug GRPC_TRACE=all,-timer,-timer_check`

hyerra:
The issue isn't present in 1.55.0rc1 for me, and works fine on that version. I have attached the logs from 1.55.0 though, but it's kinda long.

[test.log](https://github.com/grpc/grpc/files/11546749/test.log)

eapolinario:
Also seeing this on 1.55.0 on a Mac M2. Here's the [logs](https://gist.github.com/eapolinario/17a9d966aec1be6932573996bc209c81).

joerunde:
+1 also seeing this issue on grpcio `1.55.0` and `1.55.0-rc1` but not `1.54.2`

XuanWang-Amos:
We just yanked gRPC Python v1.55.0, sorry for the inconvenience it caused.

[...]

PR #33279, opened 2023-05-30
[EventEngine] Disable EventEngine polling in gRPC Python

drfloob:
This is a hack to get around an issue on Apple devices caused by the PosixEventEngine's `poll` poller not supporting fork. This PR disables the EventEngine poller entirely in Python builds. It will therefore prevent the release of the EventEngine generally, and prevent any testing of EventEngine integration with gRPC Python, until the `poll` poller is fixed.

### DIFF
--- a/bazel/cython_library.bzl
+++ b/bazel/cython_library.bzl
@@ -65,6 +65,7 @@ def pyx_library(name, deps = [], py_deps = [], srcs = [], **kwargs):
         )
 
     shared_objects = []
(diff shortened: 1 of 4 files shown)


gnossen (approved):

gnossen (approved):

gnossen:
This will temporarily mitigate https://github.com/grpc/grpc/issues/33218

gnossen (approved):
Thanks for the fix!

PR #33282, opened 2023-05-31
[experiments] Fix bug in codegen for debug experiments

Vignesh2208:
Somehow this was missed earlier because none of the experiments were enabled in debug builds.


ctiller:
We should probably have unit tests 😅

ctiller (approved):

PR #33286, opened 2023-05-31
[core] checking the status of thread creation

chansen3:
It is possible for allocation of a system resource to fail. While the Thread abstraction in the core allows for a 'success' parameter, it is optional and not generally checked where used. This can lead to degenerate behavior which is hard to diagnose. In particular, the Executors assign Closures to their queues regardless of whether the thread handling to queue is actually running, leading those async requests to hang.

This change makes the 'success' parameter required and tries to make the correct decision on how to proceed in the code where those Threads are started.  For the Executor, we will attempt to carry on with the number of threads we were able to create.  In other areas, we assert, assuming that whatever function is being performed by the allocated thread is critical to the overall function of the library.

For the tests, we ignore the value of 'success' as we assume the tests will fail if thread creation failed. This is no different than the existing behavior.

### DIFF
--- a/src/core/lib/gprpp/thd.h
+++ b/src/core/lib/gprpp/thd.h
@@ -85,14 +85,14 @@ class Thread {
   /// Normal constructor to create a thread with name \a thd_name,
   /// which will execute a thread based on function \a thd_body
   /// with argument \a arg once it is started.
-  /// The optional \a success argument indicates whether the thread
+  /// The \a success argument indicates whether the thread
   /// is successfully created.
   /// The optional \a options can be used to set the thread detachable.
   Thread(const char* thd_name, void (*thd_body)(void* arg), void* arg,
-         bool* success = nullptr, const Options& options = Options());
+         bool* success, const Options& options = Options());
 
-  Thread(const char* thd_name, absl::AnyInvocable<void()> fn,
-         bool* success = nullptr, const Options& options = Options())
+  Thread(const char* thd_name, absl::AnyInvocable<void()> fn, bool* success,
+         const Options& options = Options())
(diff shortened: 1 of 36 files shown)


ctiller:
Looks like this is still needing some call sites updated?

chansen3:
> Looks like this is still needing some call sites updated?

Indeed. I believe I have addressed these now.

chansen3:
I believe the latest test failures are addressed.

chansen3:
> There are still test failures here

I believe I have addressed the failure.

chansen3:
Anything I can do to facilitate the review?

veblush (approved):

veblush:
@ctiller Would you do one more review on this PR?

chansen3:
Is there anything I can do to facilitate this review?

chansen3:
Hello? Anyone home?

yashykt:
@ctiller ping

ctiller (approved):

veblush:
We found a bunch of build error when having this change internally because there are internal use-cases of `Thread` which are not available here. If we want to have this, we need to keep the old ctor without success so that internal code can continue to work.

chansen3:
> We found a bunch of build error when having this change internally because there are internal use-cases of `Thread` which are not available here. If we want to have this, we need to keep the old ctor without success so that internal code can continue to work.

I find it difficult to address this comment as the point of this change is to require that the status of thread creation is checked. If I were to create a new constructor that allowed one to ignore the status, then the problem hasn't been addressed properly. I would suggest instead that we fix these internal use-cases that are not visible to me.

PR #33288, opened 2023-05-31
Update gRPC to RE2 release `2023-06-01`.

junyer:
Ping, @veblush? :)


junyer:
Thanks for taking over! Please let me know if you run into problems; I can hopefully help find solutions. (RE2 **_still_** manages to support GNU make, CMake and Bazel across Linux, macOS and Windows, so...)

PR #33290, opened 2023-05-31
[Rls] de-experimentalize RLS in XDS

apolcyn:
Integration tests are passing, so we should be ready to de-experimentalize.

Related: internal bug b/265209578

### DIFF
--- a/src/core/ext/xds/xds_route_config.cc
+++ b/src/core/ext/xds/xds_route_config.cc
@@ -80,10 +80,9 @@
 
 namespace grpc_core {
 
-// TODO(donnadionne): Remove once RLS is no longer experimental
 bool XdsRlsEnabled() {
   auto value = GetEnv("GRPC_EXPERIMENTAL_XDS_RLS_LB");
-  if (!value.has_value()) return false;
+  if (!value.has_value()) return true;
   bool parsed_value;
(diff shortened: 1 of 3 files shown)


src/core/ext/xds/xds_route_config.cc
```diff
@@ -80,10 +80,9 @@
 
 namespace grpc_core {
 
-// TODO(donnadionne): Remove once RLS is no longer experimental
```
markdroth, line 83: This TODO needs to stay, although you can change the assignee.  We still want to get rid of this env var guard before too long.

markdroth (approved):
In C-core, we normally just remove the env var guard once interop tests pass rather than just flipping the default, so that we don't have to remember to come back later and remove the env var guard completely.

If you want to keep the ability to disable this feature at run-time for a short period of time just to be safe, I'm okay with that, as long as you commit to coming back and removing the env var guard after that -- let's say no longer than 1-2 releases.  I don't want this cruft sitting in our code forever.

src/core/ext/xds/xds_route_config.cc
apolcyn: SG, I suspect this env var will be useful at least for upcoming weeks while we iron out possible RLS issues beyond the client.

Updated the comment to remove by 1.58

apolcyn:
bazel c/c++ tool failure

PR #33309, opened 2023-06-01
[core] Add support for vsock transport

YadongQi:
Revert "Revert "[core] Add support for vsock transport" (https://github.com/grpc/grpc/pull/33276)"
This reverts commit https://github.com/grpc/grpc/commit/c5ade3011af41c16e32c4c4006a9c31c433f36e4.

And fix the issue which broke the python build.

@markdroth @drfloob please review this PR. Thank you very much.


src/core/lib/iomgr/port.h
```diff
@@ -38,9 +38,9 @@
 #define GRPC_HAVE_MSG_NOSIGNAL 1
 #define GRPC_HAVE_UNIX_SOCKET 1
 #ifdef LINUX_VERSION_CODE
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 9, 0)
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(3, 9, 0)) && defined AF_VSOCK
```
drfloob, line 41: I don't think `AF_VSOCK` will be defined anywhere until `socket.h` is imported, we have a chicken and egg problem. Alternatively, we could enable VSOCK for new-enough glibc versions using `#ifdef __GLIBC__` and `#if __GLIBC_PREREQ(2, 18)`.

YadongQi: I was to use `__GLIBC_PREREQ` at first, but later I found that the python build will also use musl-libc(`python_musllinux_1_1_x64_cpxx`), so I changed to use `AF_VSOCK` since no method to check musl-libc's version. But just as you commented, this is incorrect here since `socket.h` is not imported.
Updated patch to use `__GLIBC_PREREQ`, please review.

drfloob: I think that's probably fine for now, VSOCK will only be supported when using glibc. If a request comes in to enable this feature for other libc implementations, we can revisit the guards.

src/core/lib/iomgr/port.h
```diff
@@ -37,6 +37,15 @@
 #define GRPC_HAVE_IP_PKTINFO 1
 #define GRPC_HAVE_MSG_NOSIGNAL 1
 #define GRPC_HAVE_UNIX_SOCKET 1
+#ifdef LINUX_VERSION_CODE
```
drfloob, line 40: nit: I think short-circuit macro evaluation would work here, since we only care about the inner-most scope.

```
#if defined(LINUX_VERSION_CODE) && LINUX_VERSION_CODE >= KERNEL_VERSION(3, 9, 0) && \
    defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 18)
#define GRPC_HAVE_VSOCK 1
#endif

File: src/core/lib/iomgr/port.h (hunk 1 above)
Comment (YadongQi) [reply src/core/lib/iomgr/port.h]: agree.

File: src/core/lib/iomgr/port.h (hunk 2 above)
Comment (YadongQi) [reply src/core/lib/iomgr/port.h]: Updated, please review.

**Comment by drfloob** [comment]
I took the liberty of pushing a formatting fix to your branch (please pull again before you attempt to push another commit, if necessary).

**Comment by YadongQi** [comment]
@drfloob the preprocessor is reporting error
```
      ./src/core/lib/iomgr/port.h:65:41: error: missing binary operator before token "("
         65 |     LINUX_VERSION_CODE >= KERNEL_VERSION(3, 9, 0) && \
```
preprocessor will interpret `KERNEL_VERSION` as `0` by default when it is not defined . It will not be ignored when combined in one line with `#if defined(LINUX_VERSION_CODE)`.

I submitted new commit to fix this, please review it.

drfloob (approved):
This looks good to me. There is an unrelated breakage in the python distribution tests, I've pinged the team about it. @gnossen @XuanWang-Amos: I believe we've solved the issue in this PR that broke the python distribution tests last time we merged. Should we merge this PR, or wait for the CI job to get fixed?

markdroth (approved):

drfloob:
The Python distribtests fix was merged, and the test is passing on this PR. Merging. Thanks for the great work, @YadongQi.

PR #33310, opened 2023-06-01
Revert "Revert "[CSHARP] Add base_namespace experimental option to C#…

tonydnewell:
Retry PR https://github.com/grpc/grpc/pull/32636 which was reverted because of an internal build failure.


tonydnewell:
@apolcyn here is the "revert revert" for the protocol buffers grpc csharp plugin for you to run additional tests on

PR #33329, opened 2023-06-02
[Bazel Python3.11] Update Bazel dependencies for Python 3.11 (v1.56.x backport)

XuanWang-Amos:
Backport of #33318 to v1.56.x.
---
We just found out that our current Bazel setup does not support Python 3.11.

This PR updates some dependencies to allow using Bazel in Python 3.11.

Cython:
 * Cython [backported Python 3.11 support change to 0.29x](https://github.com/cython/cython/issues/4500), but it appears that the Cython version we are using in Bazel does not include the fix, so we're using the latest stable version instead.

Gevent:
* The first version of gevent that supports [Python 3.11 is 22.08.0](https://github.com/gevent/gevent/issues/1903#issuecomment-1303227507).

#### Testing
* Tested locally using Python 3.11 virtual environment, was able to reproduce the issue and verified that those changes were able to fix it.

### DIFF
--- a/bazel/grpc_python_deps.bzl
+++ b/bazel/grpc_python_deps.bzl
@@ -39,9 +39,9 @@ def grpc_python_deps():
         http_archive(
             name = "cython",
             build_file = "@com_github_grpc_grpc//third_party:cython.BUILD",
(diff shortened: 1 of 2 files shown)


gnossen (approved):

gnossen:
CC @yijiem

PR #33337, opened 2023-06-03
[EventEngine] Fix use-after-free in WindowsEventEngine endpoint shutdown

drfloob:
On endpoint shutdown, the `AsyncIOState` object was sometimes freed before the last async read callback execution had run. This resulted in callback handlers attempting to access a deleted `io_state_` object. This PR removes the AsyncIOState dependency from the user callback execution step, since only the callback and its status are needed at that point.

Previous crashes were non-deterministic, but would reliably pop up once on most runs of:

```
bazel test --cxxopt='/DEBUG=full' --config=windows_dbg --test_filter='CoreEnd2endTest*' \
  --flaky_test_attempts=1 --cache_test_results=no \
  --test_env=GRPC_VERBOSITY=debug --test_env=GRPC_TRACE=event_engine \
  //test/core/end2end:core_end2end_tests@experiment=event_engine_client```

### DIFF
--- a/src/core/lib/event_engine/windows/windows_endpoint.cc
+++ b/src/core/lib/event_engine/windows/windows_endpoint.cc
@@ -347,7 +347,6 @@ void WindowsEndpoint::HandleWriteClosure::Run() {
   auto io_state = std::move(io_state_);
   GRPC_EVENT_ENGINE_ENDPOINT_TRACE("WindowsEndpoint::%p Handling Write Event",
                                    io_state->endpoint);
-  auto cb = std::move(cb_);
   const auto result = io_state->socket->write_info()->result();
   absl::Status status;
   if (result.wsa_error != 0) {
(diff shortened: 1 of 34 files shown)

### CODE REVIEW

File: src/core/lib/event_engine/windows/windows_endpoint.cc [hunk 1]
```diff
@@ -356,7 +355,7 @@ void WindowsEndpoint::HandleWriteClosure::Run() {
   } else {
     GPR_ASSERT(result.bytes_transferred == buffer_->Length());
   }
-  Reset();
+  std::ignore = ResetAndReturnCallback();
```
yijiem, line 358: Not really familiar with this code, would removing line 350 and 359 and just use `return ResetAndReturnCallback()(status);` here have the same effect?

yijiem (approved):

drfloob:
Confirmed this passes the additional Windows builds.

PR #33351, opened 2023-06-05
[core/security] Add Custom Token Lifetime - Service Acc Impersonation

mario-vimal:
Adds access token lifetime configuration for workload identity federation with service account impersonation for both explicit and implicit flows.

Changes: 
1. Adds a new member "service_account_impersonation" to the ExternalAccountCredentials class. "token_lifetime_seconds" is a member of "service_account_impersonation".
2. Adds validation checks, like token_lifetime_seconds should be between the minimum and maximum accepted value, during the creation of an ExternalAccountCredentials object. 
3. Appends "lifetime" to the body of the service account impersonation request.

Tests:
1. Modifies a test to check if the default value is passed when "service_account_impersonation" is empty.
2. Adds tests to check if the token_lifetime_seconds value is propagated to the request body.
3. Adds tests to verify that an error is thrown when token_lifetime_seconds is invalid.


mario-vimal:
@aeitzman Please review the PR.

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -197,6 +199,30 @@ RefCountedPtr<ExternalAccountCredentials> ExternalAccountCredentials::Create(
...
+    if (service_acc_imp_obj_it != it->second.object().end()) {
+      if (service_acc_imp_obj_it->second.type() != Json::Type::kNumber) {
+        *error = GRPC_ERROR_CREATE("token_lifetime_seconds must be a number.");
+        return nullptr;
+      }
+      lifetime =
+          strtol(service_acc_imp_obj_it->second.string().c_str(), nullptr, 10);
+      if (lifetime > TWELVE_HOURS_IN_SECONDS) {
+        *error = GRPC_ERROR_CREATE(
+            "token_lifetime_seconds must be less than 43200s.");
```
aeitzman, line 216: Lets also check the minimum value (600 seconds)

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -197,6 +199,30 @@ RefCountedPtr<ExternalAccountCredentials> ExternalAccountCredentials::Create(
...
+        return nullptr;
+      }
+    }
+  }
+  Json::Object service_account_impersonation;
```
aeitzman, line 221: Can we not just do
options.service_account_impersonation = json.object().find("service_account_impersonation") instead of reading in the json and recreating it? And then use the default in the call to iamcredentials endpoint instead of inserting it here.

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -430,8 +456,16 @@ void ExternalAccountCredentials::ImpersenateServiceAccount() {
   headers[1].key = gpr_strdup("Authorization");
   headers[1].value = gpr_strdup(str.c_str());
   request.hdrs = headers;
+  std::vector<std::string> body_members;
   std::string scope = absl::StrJoin(scopes_, " ");
-  std::string body = absl::StrFormat("scope=%s", scope);
+  body_members.push_back(absl::StrFormat("scope=%s", UrlEncode(scope).c_str()));
```
aeitzman, line 461: why was the urlEncode added here?

test/core/security/credentials_test.cc
```diff
@@ -2177,6 +2180,45 @@ void validate_external_account_creds_service_account_impersonation_request(
                     "Bearer token_exchange_access_token") == 0);
 }
 
+void validate_external_account_creds_serv_acc_imp_custom_lifetime_request(
+    const grpc_http_request* request, const char* host, const char* path,
+    const char* body, size_t body_size, bool /*expect_actor_token*/) {
+  // Check that the body is constructed properly.
+  GPR_ASSERT(body != nullptr);
+  GPR_ASSERT(body_size != 0);
+  GPR_ASSERT(strcmp(body, "scope=scope_1%20scope_2&lifetime=1800s") == 0);
```
aeitzman, line 2189: Is there not a way to pass the expected lifetime? Having the validate function only work with the hardcoded 1800s is a little confusing to me.

aeitzman (changes requested):

test/core/security/credentials_test.cc
```diff
@@ -2451,21 +2500,61 @@ TEST(CredentialsTest,
   HttpRequest::SetOverride(nullptr, nullptr, nullptr);
 }
 
+TEST(
+    CredentialsTest,
+    TestExternalAccountCredsSuccessWithServiceAccountImpersonationAndCustomTokenLifetime) {
```
aeitzman, line 2505: Can you add a test that shows the error is thrown when the lifetime value provided is over 43200s? And that we use the default value of 3600s when no lifetime is provided.

aeitzman:
@lsirac

src/core/lib/security/credentials/external/external_account_credentials.cc
mario-vimal: Added UrlEncode as the content-type for the request is x-www-url-formencoded.

test/core/security/credentials_test.cc
mario-vimal: Unfortunately no, as the method is being called implicitly.

BigTailWolf:
Please also add (release notes: no)

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -63,6 +63,9 @@
   "urn:ietf:params:oauth:token-type:access_token"
 #define GOOGLE_CLOUD_PLATFORM_DEFAULT_SCOPE \
   "https://www.googleapis.com/auth/cloud-platform"
+#define IMPERSONATED_CRED_DEFAULT_LIFETIME_IN_SECONDS 3600
+#define IMPERSONATED_CRED_MIN_LIFETIME_IN_SECONDS 10 * 60
+#define IMPERSONATED_CRED_MAX_LIFETIME_IN_SECONDS 12 * 60 * 60
```
BigTailWolf, line 68: We don't have consistency on `3600` and `10 * 60` for expression.

aeitzman (approved):

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -55,14 +55,18 @@
...
   "urn:ietf:params:oauth:grant-type:token-exchange"
 #define EXTERNAL_ACCOUNT_CREDENTIALS_REQUESTED_TOKEN_TYPE \
   "urn:ietf:params:oauth:token-type:access_token"
 #define GOOGLE_CLOUD_PLATFORM_DEFAULT_SCOPE \
   "https://www.googleapis.com/auth/cloud-platform"
+#define IMPERSONATED_CRED_DEFAULT_LIFETIME_IN_SECONDS \
+  1 * 60 * 60                                              // 1 hour - 3600s
+#define IMPERSONATED_CRED_MIN_LIFETIME_IN_SECONDS 10 * 60  // 10 mins - 600s
+#define IMPERSONATED_CRED_MAX_LIFETIME_IN_SECONDS \
+  12 * 60 * 60  // 12 hours - 43200s
```
BigTailWolf, line 69: Personally I think calculated value make more sense here if we already have comments following that.

lsirac (approved):

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -55,14 +55,16 @@
...
 
 #define EXTERNAL_ACCOUNT_CREDENTIALS_GRANT_TYPE \
   "urn:ietf:params:oauth:grant-type:token-exchange"
 #define EXTERNAL_ACCOUNT_CREDENTIALS_REQUESTED_TOKEN_TYPE \
   "urn:ietf:params:oauth:token-type:access_token"
 #define GOOGLE_CLOUD_PLATFORM_DEFAULT_SCOPE \
   "https://www.googleapis.com/auth/cloud-platform"
+#define IMPERSONATED_CRED_DEFAULT_LIFETIME_IN_SECONDS 3600  // 1 hour
+#define IMPERSONATED_CRED_MIN_LIFETIME_IN_SECONDS 600       // 10 mins
+#define IMPERSONATED_CRED_MAX_LIFETIME_IN_SECONDS 43200     // 12 hours
```
erm-g, line 67: I like the idea of secure defaults and related validation. Would it also make sense to expose it by env variables?

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -197,6 +199,31 @@ RefCountedPtr<ExternalAccountCredentials> ExternalAccountCredentials::Create(
...
+      int32_t lifetime =
+          strtol(service_acc_imp_obj_it->second.string().c_str(), nullptr, 10);
+      if (lifetime > IMPERSONATED_CRED_MAX_LIFETIME_IN_SECONDS) {
+        *error = GRPC_ERROR_CREATE(
+            "token_lifetime_seconds must be less than 43200s");
+        return nullptr;
+      }
+      if (lifetime < IMPERSONATED_CRED_MIN_LIFETIME_IN_SECONDS) {
+        *error =
+            GRPC_ERROR_CREATE("token_lifetime_seconds must be more than 600s");
```
erm-g, line 222: `600s` and '43200s' are hardcoded in the message - is there a way to externalize it?

test/core/security/credentials_test.cc
```diff
@@ -2451,21 +2499,110 @@ TEST(CredentialsTest,
...
+  ExecCtx exec_ctx;
+  Json credential_source = Json::FromString("");
+  auto service_account_impersonation =
+      JsonParse("{\"token_lifetime_seconds\":1800}");
```
erm-g, line 2508: Could you also add tests where expected names contain typos to make sure code correctly handles such cases? Ie `"service_account_impersonationZ"` and `"token_lifetime_secondsZ"`.

erm-g (changes requested):

src/core/lib/security/credentials/external/external_account_credentials.cc
aeitzman: The implementation for this feature is already completed in other libraries, so doesn't make sense to add env variable support for just the gRPC implementation.

src/core/lib/security/credentials/external/external_account_credentials.h
```diff
@@ -51,6 +51,7 @@ class ExternalAccountCredentials
     std::string audience;
     std::string subject_token_type;
     std::string service_account_impersonation_url;
+    Json service_account_impersonation;
```
rockspore, line 54: Any specific reason to store this as a Json object? It'd be more efficient to have parsed content here so that subsequent calls (ImpersonateServiceAccount) don't have to parse the Json again.

test/core/security/credentials_test.cc
mario-vimal: We are creating the "options" object manually and using it to create a "Test External Acc Credential". So there is no json to credential creation happening here to test the misspelled keys. Also this type of test for misspelled key has not been done for any other credential types.

src/core/lib/security/credentials/external/external_account_credentials.h
mario-vimal: Modified this to store parsed content here. Thank you.

rockspore (approved):
@erm-g is OOO till next week but he said we should take care of it. LGTM.

rockspore:
Taking a more careful look, the failure in the [RBE Windows](https://source.cloud.google.com/results/invocations/a5c52924-55bc-42cd-83f2-46240f7e0348/targets/%2F%2Ftest%2Fcore%2Fsecurity:credentials_test/tests) build seems to be caused by this PR. Please fix.

test/core/security/credentials_test.cc
```diff
@@ -2451,21 +2502,116 @@ TEST(CredentialsTest,
...
+      "\"credential_source\":{\"url\":\"https://foo.com:5555/"
+      "generate_subject_token_format_json\","
+      "\"headers\":{\"Metadata-Flavor\":\"Google\"},"
+      "\"format\":{\"type\":\"json\",\"subject_token_field_name\":\"access_"
+      "token\"}},\"quota_project_id\":\"quota_project_id\","
+      "\"client_id\":\"client_id\",\"client_secret\":\"client_secret\"}";
+  grpc_error_handle error1, error2;
+  auto json1 = JsonParse(options_string1);
+  std::vector<std::string> scopes1 = {"scope1", "scope2"};
+  gpr_log(GPR_INFO, "Create Credentials!\n");
```
rockspore, line 2564: I suppose these gpr_logs are not needed? I don't see them in other tests.

test/core/security/credentials_test.cc
```diff
@@ -2451,21 +2502,113 @@ TEST(CredentialsTest,
...
+      "\"token_url\":\"https://foo.com:5555/token\","
+      "\"token_info_url\":\"https://foo.com:5555/token_info\","
+      "\"credential_source\":{\"url\":\"https://foo.com:5555/"
+      "generate_subject_token_format_json\","
+      "\"headers\":{\"Metadata-Flavor\":\"Google\"},"
+      "\"format\":{\"type\":\"json\",\"subject_token_field_name\":\"access_"
+      "token\"}},\"quota_project_id\":\"quota_project_id\","
+      "\"client_id\":\"client_id\",\"client_secret\":\"client_secret\"}";
+  grpc_error_handle error1, error2;
+  auto json1 = JsonParse(options_string1);
```
rockspore, line 2562: I just realized you are directly using the value of this StatusOr. I think you want to assert json1.ok(). Same for json2.

test/core/security/credentials_test.cc
```diff
@@ -2451,21 +2502,113 @@ TEST(CredentialsTest,
...
+  std::vector<std::string> scopes1 = {"scope1", "scope2"};
+  ExternalAccountCredentials::Create(*json1, std::move(scopes1), &error1);
```
rockspore, line 2564: nit: Why not just put `{"scope1", "scope2"}` here instead of having a temp variable and moving it? Same for scopes2.

rockspore (changes requested):

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -197,6 +201,36 @@ RefCountedPtr<ExternalAccountCredentials> ExternalAccountCredentials::Create(
...
     }
   }
+  it = json.object().find("service_account_impersonation");
+  options.service_account_impersonation.token_lifetime_seconds =
+      IMPERSONATED_CRED_DEFAULT_LIFETIME_IN_SECONDS;
+  if (it != json.object().end() && it->second.type() == Json::Type::kObject) {
+    auto service_acc_imp_json = it->second;
+    auto service_acc_imp_obj_it =
+        service_acc_imp_json.object().find("token_lifetime_seconds");
+    if (service_acc_imp_obj_it != it->second.object().end()) {
```
rockspore, line 211: It should be `service_acc_imp_json.object().end()` right?

rockspore: Nvm, it should be the same. I was just spectulating why the windows RBE build keeps failing.

mario-vimal: Yes, it should be. But "it->second" was assigned to "service_acc_imp_json", so they should be the same.
I'll update it for better code clarity.

rockspore: Not sure why but it looks like this change indeed fixed the windows build.

rockspore (approved):
@erm-g As we discussed offline, I'll let you handle the rest.

PR #33365, opened 2023-06-07
[Python o11y] Fix python distribtests

XuanWang-Amos:
Fix for Cython build issue in aarch64.

We're seeing this error in aarch64 distribution test:
```
In file included from ./src/core/lib/slice/slice.h:36,
                 from ./src/core/lib/slice/slice_buffer.h:29,
                 from ./src/core/lib/transport/transport.h:60,
                 from ./src/core/lib/channel/channel_stack.h:75,
                 from ./src/core/lib/channel/call_tracer.h:32,
                 from src/python/grpcio/grpc/_cython/cygrpc.cpp:2230:
./src/core/lib/slice/slice_refcount.h: In member function 'void grpc_slice_refcount::Ref(grpc_core::DebugLocation)':
./src/core/lib/slice/slice_refcount.h:55:25: error: expected ')' before 'PRIdPTR'
               "REF %p %" PRIdPTR "->%" PRIdPTR, this, prev_refs, prev_refs + 1);
                         ^~~~~~~~
                         )
```

Based on [this post](https://stackoverflow.com/questions/26182336/priuptr-preprocessor-bug-in-gcc), it's caused by including `<inttypes.h>` before define `__STDC_FORMAT_MACROS` marco.

`<inttypes.h>` was included in `core/lib/channel/call_tracer.h` and this macro should already be defined in `grpc/grpc.h` through `port_platform.h`, but we're still having issue in aarch64, so we manually define the macro in this PR.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -415,7 +415,7 @@ else:
 
 # Fix for multiprocessing support on Apple devices.
 # TODO(vigneshbabu): Remove this once the poll poller gets fork support.
-DEFINE_MACROS += (('GRPC_DO_NOT_INSTANTIATE_POSIX_POLLER', 1),)
+DEFINE_MACROS += (('GRPC_DO_NOT_INSTANTIATE_POSIX_POLLER', 1), ('__STDC_FORMAT_MACROS', None))
 
 LDFLAGS = tuple(EXTRA_LINK_ARGS)
 CFLAGS = tuple(EXTRA_COMPILE_ARGS)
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -417,6 +417,13 @@ def _quote_build_define(argument):
 # TODO(vigneshbabu): Remove this once the poll poller gets fork support.
 DEFINE_MACROS += (('GRPC_DO_NOT_INSTANTIATE_POSIX_POLLER', 1),)
 
+# Fix for Cython build issue in aarch64.
```
gnossen, line 420: Please add an issue and a TODO to figure out what's going on in the aarch64 build so that we can supports gcc + Bazel.

gnossen (approved):

PR #33368, opened 2023-06-07
[Ruby] Fix memory leak in grpc_rb_call_run_batch

peterzhu2118:
The function grpc_rb_call_run_batch has many places that could raise errors, including in child functions. Since a raised error will longjump out of the function, it will cause memory leaks since the function cannot perform any clean up. This commit fixes the issue by wrapping the whole function in an rb_ensure, which will ensure that a cleanup function is ran before the error is propagated upwards.


apolcyn:
@peterzhu2118 FYI there is no need to force push - all commits will be squashed down at the time of merging.

apolcyn (approved):

veblush (approved):

PR #33369, opened 2023-06-07
[cherrypick][v1.49] ring_hash: fix subchannel list to not shutdown until picker is destro…

eugeneo:
…yed (#30714)

* ring_hash: fix subchannel list to not shutdown until picker is destroyed

* hop into WorkSerializer before unreffing subchannels

* use a weak ref for subchannel connectivity state watches

* Automated change: Fix sanity tests

* fix memory leak

* clang-format

* fix circular reference problem by moving ring into subchannel list

* Automated change: Fix sanity tests


markdroth:
Please include https://github.com/grpc/grpc/pull/32012 in this back-port.

markdroth (approved):

PR #33375, opened 2023-06-08
[resource quota] Fix bugs in iomgr and event engine endpoint interactions with resource quota

Vignesh2208:
The following bugs are fixed:

* Missing ExecCtx in event engine endpoints and listeners
* Ref counting issue with iomgr endpoint which causes crashes in overloaded situations

The PR includes a test which triggers these bugs by simulating an overloaded system.


src/core/lib/event_engine/posix_engine/posix_endpoint.cc
```diff
@@ -549,29 +551,52 @@ void PosixEndpointImpl::MaybeMakeReadSlices() {
...
-  absl::AnyInvocable<void(absl::Status)> cb = std::move(read_cb_);
-  read_cb_ = nullptr;
-  incoming_buffer_ = nullptr;
-  read_mu_.Unlock();
-  cb(status);
+  return true;
+}
+
+void PosixEndpointImpl::HandleRead(absl::Status status) {
+  if (grpc_core::ExecCtx::Get() == nullptr) {
```
ctiller, line 573: This pattern repeats a few times. I think we should eliminate the duplication it incurs, either by locally adding extra methods, or maybe:

```
template <typename F>
void EnsureRunInExecCtx(F f) {
  if (grpc_core::ExecCtx::Get() == nullptr) {
    grpc_core::ApplicationCallbackExecCtx app_ctx;
    grpc_core::ExecCtx exec_ctx;
    f();
  } else {
    f();
  }
}
```

this could be added to exec_ctx.h and used wherever needed during this transition:

```
EnsureRunInExecCtx([]{
      grpc_core::ReleasableMutexLock lock(&read_mu_);
    if (!HandleReadLocked(status)) {
      lock.Release();
      handle_->NotifyOnRead(on_read_);
      return;
    }
    absl::AnyInvocable<void(absl::Status)> cb = std::move(read_cb_);
    read_cb_ = nullptr;
    incoming_buffer_ = nullptr;
    lock.Release();
    cb(status);
  });
```

ctiller (approved):

PR #33386, opened 2023-06-09
[EventEngine] Eliminate busy loop in the work stealing lifeguard's shutdown

drfloob:


src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -268,15 +268,15 @@ void WorkStealingThreadPool::WorkStealingThreadPoolImpl::Lifeguard::
...
+  lifeguard_is_shut_down_cv_.Signal();
 }
 
 void WorkStealingThreadPool::WorkStealingThreadPoolImpl::Lifeguard::
     BlockUntilShutdown() {
   grpc_core::MutexLock lock(&lifeguard_shutdown_mu_);
   while (lifeguard_running_) {
-    lifeguard_shutdown_cv_.Signal();
-    lifeguard_shutdown_cv_.WaitWithTimeout(
+    lifeguard_should_shut_down_cv_.Signal();
```
ctiller, line 278: We should be able to move this signal to be just when the shutdown bit changes, right?

(doing so would help catch bugs sooner by not spinning in a wakeup loop)

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.h
```diff
@@ -215,7 +215,9 @@ class WorkStealingThreadPool final : public ThreadPool {
       // Used for signaling that the lifeguard thread has stopped running.
       grpc_core::Mutex lifeguard_shutdown_mu_;
       bool lifeguard_running_ ABSL_GUARDED_BY(lifeguard_shutdown_mu_) = false;
-      grpc_core::CondVar lifeguard_shutdown_cv_
+      grpc_core::CondVar lifeguard_should_shut_down_cv_
+          ABSL_GUARDED_BY(lifeguard_shutdown_mu_);
```
ctiller, line 219: Why are we guarding the cv with a mutex (they need not be)

drfloob: Oddly enough, performance has reliably suffered when we CVs are used without locking their mutex. Counter-intuitive, but reproducible.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
drfloob: If we only signal once, I think it's possible for the lifeguard main loop to miss it and wait longer than necessary (whatever the current backoff time is). This repeated signalling was a benchmark-driven optimization as well.

ctiller: I think this bit needs to be fixed prior to de-experimentalizing: this wakeup ought to be possible with a single signal when the flag changes.

That we need a loop here will likely lead to some wildly varying and load-dependent shutdown times.

(it may be that we need to accept a small perf regression here to achieve that, and I think the robustness improvement will be worth it)

ctiller: (happy for this to be a followup PR however)

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.h
drfloob: CV removed.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
drfloob: Replaced the CV+Mutex with Notifications, which permits a single signal.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -216,6 +217,7 @@ void WorkStealingThreadPool::WorkStealingThreadPoolImpl::PrepareFork() {
   thread_count()->BlockUntilThreadCount(CounterType::kLivingThreadCount, 0,
                                         "forking", &work_signal_);
   lifeguard_.BlockUntilShutdown();
+  lifeguard_.Reset();
```
ctiller, line 220: Let's merge Reset and BlockUntilShutdown? - given it's not safe to call Reset until immediately after BlockUntilShutdown

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.h
```diff
@@ -213,10 +216,9 @@ class WorkStealingThreadPool final : public ThreadPool {
       WorkStealingThreadPoolImpl* pool_;
       grpc_core::BackOff backoff_;
       // Used for signaling that the lifeguard thread has stopped running.
-      grpc_core::Mutex lifeguard_shutdown_mu_;
-      bool lifeguard_running_ ABSL_GUARDED_BY(lifeguard_shutdown_mu_) = false;
-      grpc_core::CondVar lifeguard_shutdown_cv_
-          ABSL_GUARDED_BY(lifeguard_shutdown_mu_);
+      std::unique_ptr<grpc_core::Notification> lifeguard_should_shut_down_;
+      std::unique_ptr<grpc_core::Notification> lifeguard_is_shut_down_;
```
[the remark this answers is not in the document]
drfloob: I have seen Start and Quiesce race, leading to premature destruction and use-after-free. We'd need to synchronize around thread assignment/access in that case, and block Quiesce on a successful lifeguard startup.

I can't reproduce that race in the thread pool tests or benchmarks today using the thread-join approach (ASAN & TSAN, 1k runs). I've made the change, and we'll see if *SAN reproduces it in one of the other tests. I can't convince myself it isn't possible.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
drfloob: Done

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.h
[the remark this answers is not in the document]
drfloob: I added a stress test that races Quiesce and Start, which is important for fork support. I made an attempt this afternoon at getting the test to pass with thread joins instead of these atomics, and it ultimately required the same synchronization that is already here - there's potential to try to Join a thread that isn't assigned yet if (for instance) an EventEngine is destroyed in some post-fork handler.

I'm going to pass on this option unless you have some suggestions about how it can be done cleanly. It would be neat, if possible.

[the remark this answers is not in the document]
drfloob: As discussed, I added a half page of doc comments explaining why something like this is necessary (best we can tell).

ctiller (approved):

PR #33428, opened 2023-06-13
[HTTP2] Fix inconsistencies in keepalive configuration

yashykt:
Noticed some inconsistencies in our keepalive configuration - 

* Earlier, even if keepalive pings were disabled, we would be scheduling keepalive pings at an interval of INT_MAX ms.
* We were not using `g_default_client_keepalive_permit_without_calls` / `g_default_server_keepalive_permit_without_calls`. They are both false by default but they can be overridden in `grpc_chttp2_config_default_keepalive_args`.


ctiller:
This seems like an ideal time to add tests around the expected configuration.

yashykt:
what kind of tests are you imagining?

ctiller:
a set of asserts that would trigger if, say, we disabled keep alives but scheduled at INT_MAX seconds regardless?
or something that verifies that g_default_client_keepalive_permit_without_calls is actually used?

test/core/transport/chttp2/configuration_test.cc
```diff
@@ -0,0 +1,174 @@
...
+  EXPECT_EQ(t->keepalive_time, Duration::Seconds(20));
+  EXPECT_EQ(t->keepalive_timeout, Duration::Seconds(10));
+  EXPECT_EQ(t->keepalive_permit_without_calls, true);
+  EXPECT_EQ(t->ping_policy.max_pings_without_data, 3);
+  grpc_transport_destroy(&t->base);
+}
+
+// This test modifies the defaults of the client side settings, so it would
+// affect any test that is run after this.
+// TODO(yashykt): If adding more server side tests after this, add a rest to
```
ctiller, line 138: s/rest/reset

ctiller (approved):

PR #33433, opened 2023-06-14
[Release] Add grpc-java 1.56.0 to client_matrix.py

sergiitk:


sergiitk:
Adhoc run: https://source.cloud.google.com/results/invocations/7590dbde-b066-4d7e-8f72-e62ad2b53c61

tools/interop_matrix/client_matrix.py
```diff
@@ -418,6 +418,7 @@ def __init__(self, patch=[], runtimes=[], testcases_file=None):
             ("v1.51.1", ReleaseInfo()),
             ("v1.52.0", ReleaseInfo()),
             ("v1.54.0", ReleaseInfo()),
+            ("v1.56.0", ReleaseInfo()),
```
larry-safran, line 421: You should probably add 1.55.  Sometime we should get rid of 1.2x and 1.3x

larry-safran (approved):

tools/interop_matrix/client_matrix.py
sergiitk: I will do this separately.

gnossen (approved):

PR #33436, opened 2023-06-14
Update googleauth to ~> 1.0

### LINKED ISSUE #33435: Update googleauth ruby gem to 1.x release

grpc rubygem

debian gnu/linux
### What did you do?
In debian, we want to update googleauth gem to 1.3 but grpc gemspec has a stricter requirement   s.add_development_dependency 'googleauth',         '>= 0.5.1', '< 0.10'

We prefer to use a single version of googleauth across debian so this is blocking our googleauth update.

This dependency should be updated with a less strict requirement (assuming semver.org compliance since it is 1.0 release) to '~> 1.0'

### What did you see instead?

stricter requirement for an older version as mentioned earlier.

In debian, we prefer to use a single version of a library or a tool (in exceptional cases we do provide multiple versions for example multiple versions of gcc or python but we try to minimize this) to reduce the overall maintenance since we support this version for 5 years.

pravi:
Update googleauth ruby gem to ~> 1.0. This fixes #33435

### DIFF
--- a/grpc.gemspec
+++ b/grpc.gemspec
@@ -43,7 +43,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec',              '~> 3.6'
   s.add_development_dependency 'rubocop',            '~> 1.41.0'
   s.add_development_dependency 'signet',             '~> 0.7'
-  s.add_development_dependency 'googleauth',         '>= 0.5.1', '< 0.10'
+  s.add_development_dependency 'googleauth',         '~> 1.0'
 
   s.extensions = %w(src/ruby/ext/grpc/extconf.rb)
 


grpc.gemspec
```diff
@@ -43,7 +43,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'rspec',              '~> 3.6'
   s.add_development_dependency 'rubocop',            '~> 1.41.0'
   s.add_development_dependency 'signet',             '~> 0.7'
-  s.add_development_dependency 'googleauth',         '>= 0.5.1', '< 0.10'
+  s.add_development_dependency 'googleauth',         '~> 1.0'
```
apolcyn, line 46: this is actually generated from https://github.com/grpc/grpc/blob/133640507bc6191ee74bf76602b64ee5631c96b6/templates/grpc.gemspec.template#L48

Because we're just changing one line, you can just make the exact same change in that line of the template file -- can you please do that?

PR #33437, opened 2023-06-14
[xDS] Don't rely on move to clear a map

yousukseung:
Map is in an undefined state after moved.

### DIFF
--- a/test/cpp/end2end/xds/xds_cluster_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_cluster_end2end_test.cc
@@ -1494,65 +1494,73 @@ TEST_P(ClientLoadReportingTest, SendAllClusters) {
   balancer_->ads_service()->SetEdsResource(BuildEdsResource(args));
   // Wait until all backends are ready.
   size_t num_warmup_rpcs = WaitForAllBackends(DEBUG_LOCATION);
-  // Send kNumRpcsPerAddress RPCs per server.
-  xds::data::orca::v3::OrcaLoadReport backend_metrics;
-  auto& named_metrics = (*backend_metrics.mutable_named_metrics());
-  named_metrics["foo"] = 1.0;
-  named_metrics["bar"] = 2.0;
-  CheckRpcSendOk(DEBUG_LOCATION, kNumRpcsPerAddress * backends_.size(),
-                 RpcOptions().set_backend_metrics(backend_metrics));
-  named_metrics["foo"] = 0.3;
-  named_metrics["bar"] = 0.4;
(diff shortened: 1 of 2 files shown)


test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -1494,65 +1494,73 @@ TEST_P(ClientLoadReportingTest, SendAllClusters) {
...
-  for (size_t i = 0; i < kNumFailuresPerAddress * backends_.size(); ++i) {
-    CheckRpcSendFailure(DEBUG_LOCATION, StatusCode::FAILED_PRECONDITION, "",
-                        RpcOptions().set_server_fail(true).set_backend_metrics(
-                            backend_metrics));
-  }
-  // Check that each backend got the right number of requests.
-  for (size_t i = 0; i < backends_.size(); ++i) {
-    EXPECT_EQ(kNumRpcsPerAddress + kNumFailuresPerAddress,
-              backends_[i]->backend_service()->request_count());
+  for (int report = 0; report < 3; ++report) {
```
markdroth, line 1497: This change should be made in the `Vanilla` test above, not here in the `SendAllClusters` test.

Also, please add a comment explaining why we are doing this several times.

markdroth (commented):
What state is the map left in if not empty?

yousukseung:
> What state is the map left in if not empty?

It's almost always empty from what I tested but I missed that std::move() leaves the moved object in an unspecified state and we should not count on it (go/totw/77#beware-zombies)

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
yousukseung: Updated.

markdroth:
Having chatted about this with @ctiller, our consensus is that it isn't worth doing this.  While it's technically true that `std::move()` does not guarantee any particular state for the object, almost any reasonable `std::map<>` implementation would have this property.  And I'm almost certain that there are many other places in our code where we're making this same assumption, so it doesn't seem useful to check it in this one particular spot without doing a more systemic effort to eliminate all such occurrances and impose some sort of check to prevent them from recurring.

yousukseung:
Systemic effort sounds good but I don't see a reason to abandon this...? There's no hit to the performance, it's more "correct", with better test coverage now.

markdroth:
The point is that we're *not* undertaking such a systemic effort, because there's not enough benefit to be worth the investment.  The code is almost certainly fine the way it is.

PR #33447, opened 2023-06-14
[EventEngine] Fix use-after-free in the IOCP read callback

drfloob:
In rare cases across gRPC's end2end tests (only in Windows RBE with debug builds, interestingly enough), the endpoint may be destroyed before the final IOCP read callback has run. This edge case is only triggered when all the following are true:

* the previous read operation received data (not an error, not 0-length), so more data is expected in the stream.
* the socket has not yet shut down
* the application destroyed its endpoint before (or during) the IOCP callback execution.
* the Read operation has not yet called the client's on_read callback.

This is a valid scenario, and it is expected that the engine implementation should call the application callbacks with error statuses when this occurs. This PR fixes two associated bugs.

TODO: attempt to make this ordering part of the test suite, possibly as a stress test.

### DIFF
--- a/src/core/lib/event_engine/windows/windows_endpoint.cc
+++ b/src/core/lib/event_engine/windows/windows_endpoint.cc
@@ -55,9 +55,8 @@ WindowsEndpoint::WindowsEndpoint(
     std::shared_ptr<EventEngine> engine)
     : peer_address_(peer_address),
       allocator_(std::move(allocator)),
-      thread_pool_(thread_pool),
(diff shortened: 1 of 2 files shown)


drfloob:
I confirmed that the `event_engine_client` experiments in the Windows RBE debug tests are now passing with this PR patched in: https://source.cloud.google.com/results/invocations/9f62b2c0-3101-4d23-905e-a54c9bc997c7/targets

drfloob:
Confirmed that the extra Windows build checks all pass.

PR #33453, opened 2023-06-15
[Build] Add llvm-mingw support to CMake

hongnod:
support llvm-mingw from: https://github.com/mstorsjo/llvm-mingw


veblush:
Thank you for the PR. would you run `generate_project.sh`?

PR #33458, opened 2023-06-15
[PollingResolver] Allow subclasses to utilize the work serializer

anicr7:
Child classes of polling resolvers might have use cases where they would want to request a resolution explicitly which honor the min time between requests. As RequestReresolutionLocked needs to be called in from work_serializer which is usually owned by polling resolver completely, allow passing the work serializer back to the child class.

cc: @markdroth

### DIFF
--- a/src/core/ext/filters/client_channel/resolver/polling_resolver.h
+++ b/src/core/ext/filters/client_channel/resolver/polling_resolver.h
@@ -64,11 +64,6 @@ class PollingResolver : public Resolver {
   // To be invoked by the subclass when a request is complete.
   void OnRequestComplete(Result result);
 
-  // Can be used by the subclass to request a re-resolution.
-  // Subclasses might not be able to RequestReresolution if they don't have
-  // access to the work_serializer, so this provides a clean alternative.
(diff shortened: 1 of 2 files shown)


src/core/ext/filters/client_channel/resolver/polling_resolver.h
```diff
@@ -69,6 +69,7 @@ class PollingResolver : public Resolver {
   const std::string& name_to_resolve() const { return name_to_resolve_; }
   grpc_pollset_set* interested_parties() const { return interested_parties_; }
   const ChannelArgs& channel_args() const { return channel_args_; }
+  std::shared_ptr<WorkSerializer> work_serializer() { return work_serializer_; }
```
markdroth, line 72: I think this should return a raw pointer, not a `shared_ptr<>`.  The subclass is not going to need to take its own ref.

anicr7: Sounds good, the raw pointer works for my use case.

markdroth (approved):

PR #33459, opened 2023-06-16
[EventEngine] Simplify `EventEngine::DNSResolver` API

yijiem:
This change simplifies `EventEngine::DNSResolver`'s API based on the proposal: [go/event-engine-dns-resolver-api-changes](http://go/event-engine-dns-resolver-api-changes). Note that this API change + the implementation described in [go/event-engine-dns-resolver-implementation](http://go/event-engine-dns-resolver-implementation) has already been tested against our main test suites and are passing them.


include/grpc/event_engine/event_engine.h
```diff
@@ -356,44 +361,35 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
     using LookupTXTCallback =
         absl::AnyInvocable<void(absl::StatusOr<std::vector<std::string>>)>;
 
+    /// Implementations should make sure that all pending requests are cancelled
```
markdroth, line 364: Suggest just removing this comment; I think this information is already covered in the comment above.

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
```diff
@@ -149,74 +151,39 @@ class FuzzingResolverEventEngine
...
       if (!engine_->has_been_orphaned_) {
-        engine_->runner_.Run(
-            [this, cb = std::move(on_resolve), handle]() mutable {
-              if (!HandleExists(handle)) return;
-              DeleteHandle(handle);
-              cb(engine_->hostname_responses_);
-              CheckAndSetOrphan(ExecutionStep::AFTER_LOOKUP_HOSTNAME_CALLBACK);
-            });
+        engine_->runner_.Run([this, cb = std::move(on_resolve)]() mutable {
+          CheckAndSetOrphan(ExecutionStep::AFTER_LOOKUP_HOSTNAME_CALLBACK);
```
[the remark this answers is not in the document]
yijiem: Yes, I think what happened is that calling the `on_resolve` callback can trigger `EventEngineClientChannelDNSResolver::EventEngineDNSRequestWrapper::Orphan`. Previously, we only call `event_engine_resolver_->CancelLookup` in `Orphan`, but now we will destroy `event_engine_resolver_` in `Orphan` and will cause a crash if we access anything within the object after the callback returns.

include/grpc/event_engine/event_engine.h
yijiem: Done.

include/grpc/event_engine/event_engine.h
```diff
@@ -321,6 +321,11 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
   /// called.
   virtual bool CancelConnect(ConnectionHandle handle) = 0;
   /// Provides asynchronous resolution.
+  ///
+  /// gRPC requires this object to have a destruction-is-cancellation semantic.
```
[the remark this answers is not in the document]
yijiem: Done.

include/grpc/event_engine/event_engine.h
```diff
@@ -321,6 +321,11 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
...
+  /// Implementations should make sure that all pending requests are cancelled
+  /// when the object is destroyed and all pending callbacks will eventually be
```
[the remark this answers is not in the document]
yijiem: Done. Good point!

include/grpc/event_engine/event_engine.h
```diff
@@ -363,37 +368,26 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
...
-    /// Implementations should pass the appropriate statuses to the callback.
-    /// For example, callbacks might expect to receive DEADLINE_EXCEEDED or
+    /// When the lookup is complete or cancelled, the \a on_resolve callback
+    /// will be invoked with a status indicating the success or failure of the
+    /// lookup. Implementations should pass the appropriate statuses to the
+    /// callback. For example, callbacks might expect to receive CANCELLED or
     /// NOT_FOUND.
-    ///
-    /// If cancelled, \a on_resolve will not be executed.
-    virtual LookupTaskHandle LookupHostname(LookupHostnameCallback on_resolve,
```
drfloob, line 373: Github UI won't let me comment on the relevant lines, but the `LookupTaskHandle` type is now unused and should be deleted. Same for the `LookupTaskHandleSet` and the overloads for handle-to-string conversions. Do you plan to do that cleanup in a separate step, and is so, what's the reasoning?

include/grpc/event_engine/event_engine.h
```diff
@@ -363,37 +368,26 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
     /// \a default_port may be a non-numeric named service port, and will only
     /// be used if \a address does not already contain a port component.
     ///
-    /// When the lookup is complete, the \a on_resolve callback will be invoked
-    /// with a status indicating the success or failure of the lookup.
-    /// Implementations should pass the appropriate statuses to the callback.
-    /// For example, callbacks might expect to receive DEADLINE_EXCEEDED or
+    /// When the lookup is complete or cancelled, the \a on_resolve callback
```
apolcyn, line 371: Not sure if it's assumed already, but maybe add a note that cancellation is "best effort", in the sense that a lookup will not necessarily fail if it's cancelled (for example a cancellation might race with completion of a successful lookup, in which case the lookup succeeds).

src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
```diff
@@ -272,24 +272,27 @@ void EventEngineClientChannelDNSResolver::EventEngineDNSRequestWrapper::
...
-    for (const auto& handle : balancer_hostname_handles_) {
-      event_engine_resolver_->CancelLookup(handle);
-    }
-    if (txt_handle_.has_value()) {
-      event_engine_resolver_->CancelLookup(*txt_handle_);
+    if (timeout_handle_.has_value()) {
+      resolver_->event_engine_->Cancel(*timeout_handle_);
+      timeout_handle_.reset();
     }
+    // Even if cancellation fails here, OnResolvedLocked will return early, and
```
apolcyn, line 279: since the callback is now always invoked, does this callback still apply?

apolcyn (approved):

include/grpc/event_engine/event_engine.h
yijiem: The `LookupTaskHandle` type is still used in the iomgr DNSResolver interface: https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/resolve_address.h#L46-L47

The `LookupTaskHandleSet` type is used in the ares implementation of that interface (aka `AresDNSResolver`): https://github.com/grpc/grpc/blob/master/src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc#L709

Looks like `HandleToString` is also part of the iomgr DNSResolver interface. I think we can remove these when we remove the iomgr DNSResolver?

drfloob: Can we move the `LookupTaskHandle` type to `grpc_core::DNSResolver`? Doing it now allows us to make fewer, more self-contained public API changes, and we get to make them earlier, which are both wins for engine implementers.

src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
yijiem: Yes, I think it still apply. This part of the code hasn't been changed: https://github.com/grpc/grpc/blob/master/src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc#L475

And I think it's still the case that the resolver will not see a complete result after ee_client_channel_resolver is Orphaned.

apolcyn: Got it , thanks (I was thinking about the DNSResolver cancellation rather than "client channel DNS resolver cancellation"

include/grpc/event_engine/event_engine.h
yijiem: I'm on the fence on this. Since we don't have an explicit Cancel API now, adding the note here seems a bit out of context. I added the following note to the DNSResolver class declaration:

```
  /// Note: in the gRPC implementation, cancellation (through destruction) is
  /// "best effort" as if the request has completed during the cancellation, the
  /// result might still be reported through its callback.
```

PTAL. Thanks!

include/grpc/event_engine/event_engine.h
yijiem: Makes sense. I moved `LookupTaskHandle` and `LookupTaskHandleSet` to `grpc_core::DNSResolver`. PTAL. Thanks!

include/grpc/event_engine/event_engine.h
```diff
@@ -321,17 +321,17 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
   /// called.
   virtual bool CancelConnect(ConnectionHandle handle) = 0;
   /// Provides asynchronous resolution.
+  ///
+  /// This object has a destruction-is-cancellation semantic.
+  /// Implementations should make sure that all pending requests are cancelled
+  /// when the object is destroyed and all pending callbacks will be called
+  /// shortly.
+  ///
+  /// Note: in the gRPC's async implementation, cancellation is "best effort" in
```
drfloob, line 330: Keep in mind that this is a public contract, and as such, changing it as little as possible will be the best thing for our customers. It's not appropriate to suggest internal implementation details in this contract, since they may change.

You can say something like "If cancellation races with request completion, implementations may choose to either cancel or satisfy the request."

include/grpc/event_engine/event_engine.h
drfloob: I've provided a suggestion for this above.

include/grpc/event_engine/event_engine.h
yijiem: Good point. Done.

src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
```diff
@@ -272,24 +270,27 @@ void EventEngineClientChannelDNSResolver::EventEngineDNSRequestWrapper::
...
   }
   Unref(DEBUG_LOCATION, "Orphan");
 }
 
+void EventEngineClientChannelDNSResolver::EventEngineDNSRequestWrapper::
+    OnTimeout() {
+  MutexLock lock(&on_resolved_mu_);
+  GRPC_EVENT_ENGINE_RESOLVER_TRACE("DNSResolver::%p OnTimeout",
+                                   resolver_.get());
+  timed_out_ = true;
```
drfloob, line 289: I believe this bool is redundant, since `timeout_handle_.has_value()` gives the same information (conflated with orphaned). Since `timed_out_` is only used to determine whether to do the subsequent SRV->A record lookups, and we don't want to do those lookups when either orphaned or timed out, I think the bool is unnecessary.

include/grpc/event_engine/event_engine.h
drfloob: Looks great, thanks!

src/core/ext/filters/client_channel/resolver/dns/event_engine/event_engine_client_channel_resolver.cc
yijiem: Great catch! Yes indeed. We don't want to do the balancer lookups when the SRV lookup completion races with cancellation due to timeout. Here we can check for `!timeout_handle_.has_value()`.

drfloob (approved):
Looks great!

PR #33465, opened 2023-06-16
[grpclb] Use a separate channel arg to enable load reporting filter

markdroth:
This avoids overloading the GRPC_ARG_LB_POLICY_NAME channel arg, which is really intended for passing from the application to the client channel.

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
@@ -153,6 +153,10 @@
 #define GRPC_GRPCLB_DEFAULT_FALLBACK_TIMEOUT_MS 10000
 #define GRPC_GRPCLB_DEFAULT_SUBCHANNEL_DELETION_DELAY_MS 10000
 
+// Channel arg used to enable load reporting filter.
+#define GRPC_ARG_GRPCLB_ENABLE_LOAD_REPORTING_FILTER \
+  "grpc.internal.grpclb_enable_load_reporting_filter"
+
 namespace grpc_core {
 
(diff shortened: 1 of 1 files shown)


src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
```diff
@@ -1728,8 +1730,10 @@ void GrpcLb::OnFallbackTimerLocked() {
 ChannelArgs GrpcLb::CreateChildPolicyArgsLocked(
     bool is_backend_from_grpclb_load_balancer) {
   ChannelArgs r =
-      args_.Set(GRPC_ARG_ADDRESS_IS_BACKEND_FROM_GRPCLB_LOAD_BALANCER,
-                is_backend_from_grpclb_load_balancer);
+      args_
+          .Set(GRPC_ARG_ADDRESS_IS_BACKEND_FROM_GRPCLB_LOAD_BALANCER,
+               is_backend_from_grpclb_load_balancer)
+          .Set(GRPC_ARG_GRPCLB_ENABLE_LOAD_REPORTING_FILTER, 1);
```
apolcyn, line 1736: if `is_backend_from_grpclb_load_balancer == false`, i.e. if we're using fallback addresses, is it still correct to enable the load reporting filter?

apolcyn (approved):

src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
markdroth: Arguably no, but that's the existing behavior, so I don't want to change it.

PR #33477, opened 2023-06-17
Windows: fix error on 2nd configure

Jan-E:
If you run the configure command on Windows multiple times without clearing the the build directory the second and following fail with a 
>Microsoft JScript runtime error: File already exists

This PR adds a JScript function to check if the build directories have already been created  before performing `FSO.CreateFolder()`
I ran into this when building php_grpc.dll for PHP 8.2. See https://github.com/phalcon/cphalcon/issues/16318#issuecomment-1592534432 for the request and the result.

Note: I created this PR for the current git head. If other PR's are merged that change config.w32 there is a chance that merging fails. Please edit the PR if that happens.


Jan-E:
@drfloob As you seem to be the last one to have done something Windows-specific: could you review and merge this PR

PR #33486, opened 2023-06-19
[logging] Add tests for cases where we don't send any metadata and improve debuggability

yashykt:


ctiller:
This isn't passing tests, and also the error message is near useless to figure out what's going on:

image

yashykt:
Ah I forgot to fix this test.

The trick to debugging these is to look at the matchers and see which one didn't match, and then compare what the expectation was and what was actually sent.

yashykt:
Fixed it, and improving the error message as well.

ctiller (approved):
Approving with reservations: I think we're leaning on testing frameworks too heavily here, and we'd have a better test of we wrote some of these expectations directly.

Imagine:
EXPECT_EQ(entries.size(), 12);
auto* client_header_from_client = GetEntry(entries, kClientHeader, Logger::kClient);
EXPECT_EQ(client_header_from_client->authority, server_address_);
// etc...

the main test would be about as long (possibly shorter!), and the output would be crystal clear as to what the problem we're seeing is.

yashykt:
> Approving with reservations: I think we're leaning on testing frameworks too heavily here, and we'd have a better test of we wrote some of these expectations directly.

Yeah, I think you are right. I'll switch these over.

PR #33494, opened 2023-06-20
[alts] Remove ExecCtx check from alts_zero_copy_grpc_protector_create function.

matthewstevenson88:
I'm fixing the ALTS/Envoy transport socket extension (which is currently broken). Along the way, I'm trying to remove as many uses of gRPC internals as possible (with the eventual goal of only relying on public gRPC APIs and the alts_zero_copy_grpc_protector). To this end, I need to remove the ExecCtx check in the alts_zero_copy_grpc_protector_create function, so that Envoy can call into this function without needing to create an ExecCtx.

### DIFF
--- a/src/core/tsi/alts/zero_copy_frame_protector/alts_grpc_record_protocol_common.cc
+++ b/src/core/tsi/alts/zero_copy_frame_protector/alts_grpc_record_protocol_common.cc
@@ -124,9 +124,8 @@ tsi_result alts_grpc_record_protocol_init(alts_grpc_record_protocol* rp,
 tsi_result alts_grpc_record_protocol_protect(
     alts_grpc_record_protocol* self, grpc_slice_buffer* unprotected_slices,
     grpc_slice_buffer* protected_slices) {
-  if (grpc_core::ExecCtx::Get() == nullptr || self == nullptr ||
-      self->vtable == nullptr || unprotected_slices == nullptr ||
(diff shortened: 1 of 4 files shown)


markdroth:
Matt, can you confirm that the frame protector API is only used synchronously?  If so, then I agree that this is safe, but if not, we may need to discuss this a bit further.

markdroth:
I don't think it matters if one thread calls `protect()` and the other thread calls `unprotect()` at the same time.  What matters is whether there is a way to use either `protect()` or `unprotect()` asynchronously -- in other words, is there a case where a thread can call either of those APIs and get a result delivered asynchronously via a callback, where the callback can be invoked in some other arbitrary thread?

If the answer to that is no, then I think this should be fine.

I'd also like @apolcyn to review as well.

apolcyn (approved):
> If the answer to that is no, then I think this should be fine.

> I'd also like @apolcyn to review as well.

I'm not super familiar with the encryption/decryption code paths here, but overall +1 that we need to guarantee that nothing is or will be done asynchronously from within these APIs.

From my look at the code, I believe that's true, so I'll approve, but @matthewstevenson88 please confirm.

matthewstevenson88:
Thanks Mark and Alex. Confirmed that there is no async use of these APIs, in the sense that one thread calls the API and another thread gets the result via a callback.

markdroth (approved):

PR #33495, opened 2023-06-20
[fork] simplify Fork::SetResetChildPollingEngineFunc to fix nested forking

apolcyn:
Noticed this in a nested forking test in https://github.com/grpc/grpc/pull/33430 (by nested fork - forked child process forks again).

Before this change, the EventEngine and non-EventEngine pollers were competing with each other in their calls to `Fork::SetResetChildPollingEngineFunc`. In the first child's after-fork handler, the EE engine would [clear out](https://github.com/grpc/grpc/blob/123da4a866d278b898516c27fb44962f3d7bae32/src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc#L260) the post-fork handlers of the non-EE poller, leaving the grandchild without the right post-fork cleanup method.

### DIFF
--- a/src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
+++ b/src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
@@ -255,10 +255,6 @@ void ResetEventManagerOnFork() {
     poller->Close();
   }
   gpr_mu_unlock(&fork_fd_list_mu);
-  if (grpc_core::Fork::Enabled()) {
-    gpr_mu_destroy(&fork_fd_list_mu);
-    grpc_core::Fork::SetResetChildPollingEngineFunc(nullptr);
-  }
   InitEpoll1PollerLinux();
 }
 
@@ -273,8 +269,10 @@ bool InitEpoll1PollerLinux() {
     return false;
   }
   if (grpc_core::Fork::Enabled()) {
-    gpr_mu_init(&fork_fd_list_mu);
-    grpc_core::Fork::SetResetChildPollingEngineFunc(ResetEventManagerOnFork);
+    if (grpc_core::Fork::RegisterResetChildPollingEngineFunc(
+            ResetEventManagerOnFork)) {
+      gpr_mu_init(&fork_fd_list_mu);
+    }
   }
   close(fd);
   return true;
--- a/src/core/lib/gprpp/fork.cc
+++ b/src/core/lib/gprpp/fork.cc
@@ -189,19 +189,16 @@ void Fork::DoDecExecCtxCount() {
   NoDestructSingleton<ExecCtxState>::Get()->DecExecCtxCount();
 }
 
-void Fork::SetResetChildPollingEngineFunc(
+bool Fork::RegisterResetChildPollingEngineFunc(
     Fork::child_postfork_func reset_child_polling_engine) {
   if (reset_child_polling_engine_ == nullptr) {
-    reset_child_polling_engine_ = new std::vector<Fork::child_postfork_func>();
+    reset_child_polling_engine_ = new std::set<Fork::child_postfork_func>();
   }
-  if (reset_child_polling_engine == nullptr) {
-    reset_child_polling_engine_->clear();
-  } else {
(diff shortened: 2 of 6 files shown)


src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
```diff
@@ -273,8 +269,10 @@ bool InitEpoll1PollerLinux() {
     return false;
   }
   if (grpc_core::Fork::Enabled()) {
-    gpr_mu_init(&fork_fd_list_mu);
-    grpc_core::Fork::SetResetChildPollingEngineFunc(ResetEventManagerOnFork);
+    if (grpc_core::Fork::RegisterResetChildPollingEngineFunc(
+            ResetEventManagerOnFork)) {
+      gpr_mu_init(&fork_fd_list_mu);
```
drfloob, line 274: Is it possible for the child fork handler to run before the mutex is initialized?

src/core/lib/gprpp/fork.cc
```diff
@@ -189,19 +191,16 @@ void Fork::DoDecExecCtxCount() {
...
-  } else {
-    reset_child_polling_engine_->emplace_back(reset_child_polling_engine);
+    reset_child_polling_engine_ = new std::set<Fork::child_postfork_func>();
   }
+  auto ret = reset_child_polling_engine_->insert(reset_child_polling_engine);
+  return ret.second;
 }
 
-const std::vector<Fork::child_postfork_func>&
+const std::set<Fork::child_postfork_func>&
```
drfloob, line 203: I think this is mostly a good thing, but we're adding non-determinism to callback execution ordering. This _may_ unearth bugs based on implicit ordering requirements. I hope there aren't such dependencies between polling engines, but we should be aware of the possibility.

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
apolcyn: It's not: the first time RegisterResetChildPollingEngineFunc returns true here, we will initialize the mutex, and we'll leave it initialized thereafter.

So the only way we could possibly race here is if we forked while this thread was running, which would put us in a broken/undefined state anyways.

src/core/lib/gprpp/fork.cc
apolcyn: ack

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
drfloob: > the only way we could possibly race here is if we forked while this thread was running

That's what I'm thinking. It is possible, probably rare, but having the mutex initialization before the fork-handler registration would protect it a bit. If the handler registration is the last thing in this init method, I'm not quite sure how that leaves the poller in a broken/undefned state.

apolcyn: If we fork while this thread is running, that means we've just chopped this thread. That leaves us in undefined behavior land (e.g. locks could easily be held elsewhere in this thread that still need to be released), so I don't think it's worth making any changes to support that.

drfloob: I'm not going to block this PR on a rare edge case, it's fine, especially if our fork tests are passing. And if we see new fork problems in the wild where a null mutex is accessed, we might have a suspect change. Threads aren't exactly "chopped" when a fork occurs and fork support is engaged, the pre-fork handlers do some work in letting threads come to a resting spot (in some cases shutting them down) before allowing the fork to proceed.

drfloob (approved):

src/core/lib/event_engine/posix_engine/ev_epoll1_linux.cc
drfloob: Ah right, we are registering a post-fork handler here, not a pre-fork handler. The pre-fork handlers will not allow a fork to occur between these statements. That should be fine then.

PR #33505, opened 2023-06-21
[ObjC] generate Package.swift from template

HannahShiSFB:
This is to help simplify the grpc iOS release flow.

Package.swift was manually written and out of sync for a long time in this repo.
They are currently being updated updated for releases in [grpc-ios](https://github.com/grpc/grpc-ios) repo, which includes a lot of manual work to include new files and exclude tests.

This diff generate Package.swift from a template like other languages to avoid the tedious work. Currently only gRPC-Core is generated form the template, gRPC-cpp remains the same for now as they don't change often.

CC @sampajano


sampajano (approved):
thank you so much for optimizing the release flow, Hannah!

PR #33507, opened 2023-06-21
[C#] Proposal for C# null ref types support in code generation

tonydnewell:
This document is a proposal for what to do about adding C# null ref types support in the code generated by the protocol buffers compiler and plugin.

- I'm not sure if this is the correct location for this document
- The proposal is to do nothing - but I'm happy to hear counterarguments, and even spend time prototyping an implementation.


tonydnewell:
@jtattermusch @jskeet this is a document that we can point people to _if_ we decide to "do nothing" for supporting null reference types in the generated code.

jskeet:
Will have a look next week - I'm OOO until Monday.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,170 @@
...
+See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version 
+
+Note: It is possible to edit the project file and explicitly set the language 
+version to a later version, but other language features that require runtime 
+changes in the CLR will not work. C# 7.3 is the latest language version that is 
+fully supported with *.NET Framework* projects.
+
+### 3. Building gRPC projects with nullable enabled works today
+
+There does not appear to be a problem building *.NET Core* projects that have
```
jskeet, line 78: I don't know what the preferred way of saying ".NET Core and .NET" is these days. Possibly just that. But given that all .NET Core platforms are now out of MS support (IIRC) - I think it would at least be better to drop the Core. Maybe we need an introductory paragraph with terminology - I'd be fine with "Within this document, references to '.NET Core' are intended to include .NET 6.0+" or similar.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,170 @@
...
+Changes to `Google.Protobuf` library and other libraries 
+(such as `Google.Api.CommonProtos`) are outside the scope of this proposal.
+
+### Exposing protocol buffers optional fields as nullable properties
+
+See https://github.com/protocolbuffers/protobuf/issues/9083
+
+This has been rejected by the protocol buffers team.
+
+> "We get requests in some other languages that have null support to allow
```
jskeet, line 127: This also be massively breaking...

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,170 @@
...
+- Specifying a non-null message or stream is returned from a gRPC service call.
+
+By not supporting nullable reference types in the generated code these compile
+time checks cannot be done.
+
+## Outside the scope of this proposal
+
+### Changes to related packages
+
+Changes to `Google.Protobuf` library and other libraries
```
jskeet, line 118: Possibly explicitly say that there is a long-standing (slow-moving) effort to enable NRT for Google.Protobuf?

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,170 @@
...
+Options for enabling this feature:
+
+- Generate conditional code and the user has to define a constant in the project files to enable it
+  - The generated conditional code might be messy
+  - Could change `Grpc.Tools` to define the constant if it detects project has nullable enabled
+
+- Generate code that relies on definitions of nullable attributes for older framework versions
+  - Like those in https://github.com/grpc/grpc-dotnet/blob/master/src/Shared/NullableAttributes.cs
+  - Where would these definitions of nullable attributes be defined?
+  Don’t want to define them in every generated C# file. Is there a single
```
jskeet, line 153: Putting the annotations in any shared library would mean they'd have to be public rather than internal, which then causes issues with potential collisions IME.

jskeet (approved):
This looks generally good to me.

I think the "this will never play nicely with .NET Framework" is the most compelling reason.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,170 @@
...
+- Generate code that relies on definitions of nullable attributes for older framework versions
+  - Like those in https://github.com/grpc/grpc-dotnet/blob/master/src/Shared/NullableAttributes.cs
+  - Where would these definitions of nullable attributes be defined?
+  Don’t want to define them in every generated C# file. Is there a single 
+  library where they could live?
+
+- Add command line options to `protoc` and `grpc_csharp_plugin` 
+  - Option would have to be passed to both `protoc` and `grpc_csharp_plugin`
+  - Could change `Grpc.Tools` to pass the option if it detects project has nullable enabled
+  - Cannot generate the code once and use it in different projects (e.g. a *.NET Framework* client and a *.NET Core* server project)
```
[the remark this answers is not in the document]
tonydnewell: I've made updates based on your comments

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+## Background
+
+A note on the terminology used in this document for the various *.NET* versions:
+- *.NET Framework* - the implementation of *.NET* that runs only on Windows, such 
+  as *.NET Framework 4.8.1*
+- *.NET Core* - within this document, references to *.NET Core* are intended to include
+ *.NET Core 3.1*, *.NET 5*, and later. Strictly Microsoft refers to all versions
+ from *.NET 5* and later as just *.NET*. 
+
+*Nullable reference types* refers to a group of optional features introduced
```
jtattermusch, line 26: nit: add a link to the official nullable reference types documentation?

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+This proposal discusses why there are no plans to explicitly handle nullable 
+reference types in the generated code.
+
+## Background
+
+A note on the terminology used in this document for the various *.NET* versions:
+- *.NET Framework* - the implementation of *.NET* that runs only on Windows, such 
+  as *.NET Framework 4.8.1*
+- *.NET Core* - within this document, references to *.NET Core* are intended to include
+ *.NET Core 3.1*, *.NET 5*, and later. Strictly Microsoft refers to all versions
```
jtattermusch, line 23: nit: rephrase sentence "Strictly Microsoft refers ...", it doesn't read well?

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
+ Handling of null reference types in C# code generation
+----
+* Author(s): Tony Newell
+* Approver: 
+* Status: Draft
+* Last updated: June 2023
```
jtattermusch, line 6: According to the guidelines in the proposal process, for each proposal we need to open a discussion thread in the grpc.io forum so that community has a chance to voice their opinions.
The process: https://github.com/grpc/proposal/blob/master/README.md#process (see bullet 5)
Example: https://github.com/grpc/proposal/blob/master/L103-core-move-insecure-creds-declaration.md

jtattermusch:
CC @JamesNK

jtattermusch (commented):
Overall this looks pretty good, thanks!

doc/csharp/csharp-null-ref-types-proposal.md is not the right place for the proposal.

Since the proposal already has the format required by gRFC proposals in the https://github.com/grpc/proposal repository, I think we can make this a formal proposal that follows the proposal process: https://github.com/grpc/proposal/blob/master/README.md

Can you please open a PR against the https://github.com/grpc/proposal and add this as a "L"anguage-specific proposal (L105 is the next available proposal number I think).
I'll be the approver.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+See https://github.com/protocolbuffers/protobuf/issues/9083
+
+This has been rejected by the protocol buffers team.
+
+> "We get requests in some other languages that have null support to allow 
+optional fields to have nullable properties. The main problem is that we like to 
+keep languages consistent and other languages don't have those same options. 
+Additionally, it's not clear with some primitive fields what a null value would 
+be (is it the default? What if users meant to set it to that option?). For those 
+reasons, we will not support this in the near future.”
```
jskeet, line 157: Nit: opening quote is straight; closing quote is curly. Perhaps remove them entirely given that you're quoting anyway?

jskeet (approved):

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+depending on some configuration, or to generate conditional code containing `#if` 
+everywhere.
+
+Generating different code depending on configuration could lead to confusion or errors
+if code generated in one project with nullable enabled was consumed in another project 
+that did not have nullable enabled. It would also have an impact on the level of 
+testing that would need to be done.
+
+Generating conditional code would produce messy and potentially convoluted code, again
+having an impact on testing.
```
JamesNK, line 121: Yes, but I think we should figure out how big the difference will be between nullable enabled generated code and nullable disabled generated code.

I don't think there are many places in generated source that would be annotated as nullable. I took a quick look through some generated source code:

* Message fields can be null.
* `Equals(obj)` and `Equals(T)` would have nullable parameters.
* `MergeFrom(T)` would have a nullable parameter.
* `BindService` would have a nullable parameter (we could skip annotating this one since the only caller that would pass null is `Grpc.AspNetCore.Server`).

I doubt that is everything, but I don't think there is a lot more.

doc/csharp/csharp-null-ref-types-proposal.md
JamesNK: I think it is worth mentioning that well known types are in Google.Protobuf. Annotating them is valuable to the user experience.

It would be odd if user gRPC types were annotated, but `Goole.Protobuf.Any` wasn't.

doc/csharp/csharp-null-ref-types-proposal.md
JamesNK, line 157: I don't understand this bit. Nullable reference types is only for reference types. Primitive fields (int, double, etc) won't change.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+the generated code. These are given later.
+
+The above reasons are explained in more detail:
+
+### 1. The generated code is not in a nullable context
+
+The generated code is not in a nullable context. The `// <auto-generated>` 
+comment in the code disables it. No nullable checks are made and no compiler 
+warnings are generated for this code, even if the code is compiled in a project 
+with nullable checks enabled.
```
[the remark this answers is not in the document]
tonydnewell: Yes I know. The point was that the code generated today is compiler with nullable context disabled thus it compiles OK with no warnings, as in nothing needs to be "fixed" to get it compiling cleanly today.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+- In *Advanced Build Settings | Language version* the drop down to select the 
+language version is greyed out and says “*Automatically selected based on 
+framework version*”
+
+See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version 
+
+Note: It is possible to edit the project file and explicitly set the language 
+version to a later version, but other language features that require runtime 
+changes in the CLR will not work. C# 7.3 is the latest language version that is 
+fully supported with *.NET Framework* projects.
```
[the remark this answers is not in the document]
tonydnewell: I know. Again the point is that by default it is C# 7.3 and Visual Studio doesn't encourage you or help you to change the compiler version. Changing the compiler version is possible via editing the project file but the user has to be aware that while nullable reference types feature is supported by going to a later compiler, other language feature they think they might be getting may not be supported, and this is not obvious.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+### 3. Building gRPC projects with nullable enabled works today
```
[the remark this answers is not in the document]
tonydnewell: The point is that the status quo works. I'm yet to see a compelling reason to make any changes.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+class may cause errors (not warnings) in user code that has already been derived 
+from the base class:
+
+> *error CS8765: Nullability of type of parameter 'request' doesn't match 
+overridden member (possibly because of nullability attributes).*
+
+Note: the advantages of adding these attributes to the service methods is 
+questionable as the only caller of these methods is the server hosting the 
+services, not user code. Therefore this may not be an issue if it is not needed 
+to be done.
```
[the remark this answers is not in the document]
tonydnewell: E.g. the server side base class for a service could be changed to:
```C#
 public abstract partial class PersonLookupBase
    {
#if GRPC_ENABLE_NULLABLE_ATTRIBUTES
            [return: global::System.Diagnostics.CodeAnalysis.NotNull]
#endif
      public virtual global::System.Threading.Tasks.Task<global::Attrtest.Person> FindPerson(
#if GRPC_ENABLE_NULLABLE_ATTRIBUTES
          [global::System.Diagnostics.CodeAnalysis.NotNull]
#endif
          global::Attrtest.SearchCriteria request,
#if GRPC_ENABLE_NULLABLE_ATTRIBUTES
          [global::System.Diagnostics.CodeAnalysis.NotNull]
#endif
          grpc::ServerCallContext context)
      {
        throw new grpc::RpcException(new grpc::Status(grpc::StatusCode.Unimplemented, ""));
      }
```
i.e. it doesn't return null, or accept null for any of the parameters.

But is there any advantage is doing this as the only caller is the web server hosting the service and not user code.

doc/csharp/csharp-null-ref-types-proposal.md
tonydnewell: There are a few other places in the generated code that break when nullable enabled, e.g.
```
public virtual global::Attrtest.Person FindPerson(
   global::Attrtest.SearchCriteria request,
   grpc::Metadata headers = null,
   global::System.DateTime? deadline = null,
   global::System.Threading.CancellationToken cancellationToken = default(global::System.Threading.CancellationToken)
   )
      {
        return FindPerson(request, new grpc::CallOptions(headers, deadline, cancellationToken));
      }
```
the `grpc::Metadata headers = null` is one such place above that needs changing to make it compile without warnings.

It just gets messy.

doc/csharp/csharp-null-ref-types-proposal.md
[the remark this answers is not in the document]
JamesNK: If someone cares about generated gRPC code having nullable type annotations then the rest of their code has nullable type annotations enabled already. If they're using .NET Framework then they've already figured this out.

doc/csharp/csharp-null-ref-types-proposal.md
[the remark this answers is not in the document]
JamesNK: `NotNull` doesn't need to be used like that. There are some situations where it is useful, but it is mostly when working with a generic type argument that might or might not be a reference type. Or for informing static analysis that if this method didn't throw an error when a value was passed in, the value must not be null. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#postconditions-maybenull-and-notnull

It should look like this:
```cs
#nullable enable
public abstract partial class PersonLookupBase
{
    public virtual global::System.Threading.Tasks.Task<global::Attrtest.Person> FindPerson(
        global::Attrtest.SearchCriteria request,
        grpc::ServerCallContext context)
    {
        throw new grpc::RpcException(new grpc::Status(grpc::StatusCode.Unimplemented, ""));
    }
}
```
Reference types without `?` on them are inferred not to allow null.

doc/csharp/csharp-null-ref-types-proposal.md
```diff
@@ -0,0 +1,197 @@
...
+
+### 5. Keeping the generated code simple and consistent
+
+It would be possible to generate different code for nullable and non-nullable contexts
+depending on some configuration, or to generate conditional code containing `#if` 
+everywhere.
```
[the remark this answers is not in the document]
tonydnewell: I've been advised that the protocol buffers team discourage adding more command line options as it increases testing. We might also get inconsistent code before protoc and the gRPC plugin as both would need to be passed the same command line options. Grpc.Tools can do that but if someone is using the compiler without Grpc.Tools then they might pass incosistent arguments. Thus if we were to do anything the preference might be to generate conditional code with `#if` throughout the code - (before you comment - yes, that increases testing too). 

The question is how is the user experience impacted by:
- not doing anything (compiles OK today)
- versus the advantages the user gains
- versus additional complexity of the right configuration or messiness of the generated code?

doc/csharp/csharp-null-ref-types-proposal.md
JamesNK: I updated my list with `headers`.

doc/csharp/csharp-null-ref-types-proposal.md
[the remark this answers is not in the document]
JamesNK: The goal is to give gRPC users information about null references in their code. For example:

```proto
message User {
  string name = 1;
  Address address = 2;
}

message Address {
  string street = 1;
  string city = 2;
}
```

```cs
public override Task<Empty> AddUser(User user, ServerCallContext context)
{
    AddressEntity entity = new
    {
        Street = user.Address.Street, // Possible null reference exception
        City = user.Address.City,
    };
    _addressRepository.Add(entity);

    // rest of the method
}
```

Right now this will happily compile without warnings. If gRPC code gen had annotations then they would get a warning and would know that `User.Address` could be null.

```cs
public override Task<Empty> AddUser(User user, ServerCallContext context)
{
    if (user.Address != null)
    {
        var entity = new AddressEntity
        {
            Street = user.Address.Street,
            City = user.Address.City,
        };
        _addressRepository.Add(entity);
    }

    // rest of the method
}
```

tonydnewell:
@JamesNK @jskeet @jtattermusch I know this is the wrong place for this proposal but I don't want to move it at the moment and loose this discussion.

 If it make the discussion easier I'm happy to prototype generating conditional code that supports nullable so we can see the real life impact. I don't know if this will be a worthwhile exercise. (We can later decide if conditional code is the right way to do this, or whether to have a protoc option).

What I'm still not clear on is what advantages are we gaining. I keep going back to the fact that the grpc-dotnet examples all have nullable enabled and are using generated code without nullable enabled and there doesn't seem to be an issue.

doc/csharp/csharp-null-ref-types-proposal.md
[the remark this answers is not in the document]
tonydnewell: Update: @jskeet has spoken to the protocol buffers team and they are OK with having a command line option.

PR #33464, opened 2023-06-16
[Test] Add v1.56.0 release of grpc-go to interop matrix

arvindbr8:
Add grpc-go v1.56.0 to interop matrix

PR #33511, opened 2023-06-21
Add gRPC-Go version 1.56.1 to interop matrix

zasweq:
We had to do a patch release. Thus, we can close #33464.


dfawley:
:cry: replacing with another PR soon

Issue #33483, opened 2023-06-19
gem install error: Undefined symbols for architecture arm64


andrisll:
### What version of gRPC and what language are you using?
grpc: 1.56.0
language: ruby 3.2.2

### What operating system (Linux, Windows,...) and version?
macOS 13.4
Apple M1 chip

### What runtime / compiler are you using (e.g. python version or version of gcc)
Apple clang version 14.0.3 (clang-1403.0.22.14.1)

### What did you do?
`gem install grpc -v 1.56.0`

### What did you expect to see?
Gem successfully installed.

### What did you see instead?
`
40 warnings generated.
compiling rb_enable_cpp.cc
linking shared-object grpc/grpc_c.bundle
Undefined symbols for architecture arm64:
  "_CFRelease", referenced from:
      absl::lts_20230125::time_internal::cctz::local_time_zone() in libgrpc.a(time_zone_lookup.o)
  "_CFStringGetCString", referenced from:
      absl::lts_20230125::time_internal::cctz::local_time_zone() in libgrpc.a(time_zone_lookup.o)
  "_CFStringGetLength", referenced from:
      absl::lts_20230125::time_internal::cctz::local_time_zone() in libgrpc.a(time_zone_lookup.o)
  "_CFStringGetMaximumSizeForEncoding", referenced from:
      absl::lts_20230125::time_internal::cctz::local_time_zone() in libgrpc.a(time_zone_lookup.o)
  "_CFTimeZoneCopyDefault", referenced from:
      absl::lts_20230125::time_internal::cctz::local_time_zone() in libgrpc.a(time_zone_lookup.o)
  "_CFTimeZoneGetName", referenced from:
      absl::lts_20230125::time_internal::cctz::local_time_zone() in libgrpc.a(time_zone_lookup.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [grpc_c.bundle] Error 1

make failed, exit code 2
`

tdyas:
I am encountering this issue as well. The error seems like the CoreFoundation framework is not being linked in somehow?

tdyas:
I was able to work around the issue with `bundle config build.grpc --with-ldflags="-Wl,-undefined,dynamic_lookup"`.

michael-cybrid:
See this issue as well; moved past it with similar steps to @tdyas. Running with:
```
grpc: 1.56.0
ruby 3.2.2
[...]

PR #33513, opened 2023-06-21
[Ruby] add -weak_framework CoreFoundation to ruby extension

alto-ruby:
Fixes Undefined symbols of _CFXxxx on arm MacOS.

Fixes https://github.com/grpc/grpc/issues/33483

### DIFF
--- a/src/ruby/ext/grpc/extconf.rb
+++ b/src/ruby/ext/grpc/extconf.rb
@@ -141,7 +141,7 @@ end
 
 ext_export_file = File.join(grpc_root, 'src', 'ruby', 'ext', 'grpc', ext_export_filename())
 $LDFLAGS << ' -Wl,--version-script="' + ext_export_file + '.gcc"' if linux
-$LDFLAGS << ' -Wl,-exported_symbols_list,"' + ext_export_file + '.clang"' if apple_toolchain
+$LDFLAGS << ' -weak_framework CoreFoundation -Wl,-exported_symbols_list,"' + ext_export_file + '.clang"' if apple_toolchain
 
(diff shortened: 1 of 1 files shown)


src/ruby/ext/grpc/extconf.rb
```diff
@@ -141,7 +141,7 @@ def ext_export_filename()
 
 ext_export_file = File.join(grpc_root, 'src', 'ruby', 'ext', 'grpc', ext_export_filename())
 $LDFLAGS << ' -Wl,--version-script="' + ext_export_file + '.gcc"' if linux
-$LDFLAGS << ' -Wl,-exported_symbols_list,"' + ext_export_file + '.clang"' if apple_toolchain
+$LDFLAGS << ' -weak_framework CoreFoundation -Wl,-exported_symbols_list,"' + ext_export_file + '.clang"' if apple_toolchain
```
apolcyn, line 144: can we add this flag only when we're building apple *and* arm?

flavorjones: See https://github.com/grpc/grpc/pull/33538 for this change

alto-ruby: Sorry for late response, I was have access to x86 Macs to verify this doesn't happen with x86 when building from source. Now I have verified this only happens to ARM Macs.

Thanks for making the change, closing this PR.

stanhu: I had someone on my team report the same issue on `x86-darwin`, but we used the precompiled gem as a workaround.

PR #33514, opened 2023-06-21
[resolver API] remove address attribute interface

markdroth:
The address attribute interface was intended to provide a mechanism to pass attributes separately from channel args, for values that do not affect subchannel behavior and therefore do not need to be present in the subchannel key, which does include channel args.  However, the mechanism as currently designed is fairly clunky and is probably not the direction we will want to go in the long term.

Eventually, we will want some mechanism for registering channel args, which would provide a cleaner way to indicate that a given channel arg should not be used in the subchannel key, so that we don't need a completely different mechanism.  For now, this PR is just doing an interim step, which is to establish a special channel arg key prefix to indicate that an arg is not needed in the subchannel key.


src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
```diff
@@ -470,16 +470,14 @@ RingHash::RingHashSubchannelList::Ring::Ring(
...
-        ServerAddressWeightAttribute::kServerAddressWeightAttributeKey));
+    auto weight_arg = sd->address().args().GetInt(GRPC_ARG_ADDRESS_WEIGHT);
     AddressWeight address_weight;
     address_weight.address =
         grpc_sockaddr_to_string(&sd->address().address(), false).value();
     // Weight should never be zero, but ignore it just in case, since
     // that value would screw up the ring-building algorithm.
-    if (weight_attribute != nullptr && weight_attribute->weight() > 0) {
-      address_weight.weight = weight_attribute->weight();
+    if (weight_arg.has_value() && *weight_arg > 0) {
```
eugeneo, line 479: This is a subjective suggestion, but does have less characters:
```suggestion
    if (weight_arg.value_or(0) > 0) {
```

eugeneo (approved):
No comments of substance. Thank you!

src/core/ext/filters/client_channel/lb_policy/ring_hash/ring_hash.cc
markdroth: Good idea, done.

src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -1173,7 +1173,9 @@ ChannelArgs ClientChannel::MakeSubchannelArgs(
       // uniqueness.
       .Remove(GRPC_ARG_HEALTH_CHECK_SERVICE_NAME)
       .Remove(GRPC_ARG_INHIBIT_HEALTH_CHECKING)
-      .Remove(GRPC_ARG_CHANNELZ_CHANNEL_NODE);
+      .Remove(GRPC_ARG_CHANNELZ_CHANNEL_NODE)
```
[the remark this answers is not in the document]
markdroth: No, because the application may pass arbitrary channel args that might affect subchannel behavior.

PR #33519, opened 2023-06-22
[interop] Fix crash in pick_first LB policy

eugeneo:
Fixes http://b/288420022

### DIFF
--- a/test/cpp/interop/rpc_behavior_lb_policy.cc
+++ b/test/cpp/interop/rpc_behavior_lb_policy.cc
@@ -93,7 +93,7 @@ class RpcBehaviorLbPolicy : public LoadBalancingPolicy {
                 {{std::string(delegate_->name()),
                   grpc_core::Json::FromObject({})}})}));
     GPR_ASSERT(delegate_config.ok());
-    args.config = *delegate_config;
+    args.config = std::move(*delegate_config);
     return delegate_->UpdateLocked(std::move(args));
   }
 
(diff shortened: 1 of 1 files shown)


eugeneo:
I ran the interop test locally and can confirm this fixes the issue.

test/cpp/interop/rpc_behavior_lb_policy.cc
```diff
@@ -86,6 +86,14 @@ class RpcBehaviorLbPolicy : public LoadBalancingPolicy {
...
     RefCountedPtr<RpcBehaviorLbPolicyConfig> config = std::move(args.config);
     rpc_behavior_ = std::string(config->rpc_behavior());
+    // Use correct config for the delegate load balancing policy
+    auto delegate_config =
+        CoreConfiguration::Get().lb_policy_registry().ParseLoadBalancingConfig(
+            grpc_core::Json::FromArray({grpc_core::Json::FromObject(
+                {{std::string(delegate_->name()),
+                  grpc_core::Json::FromObject({})}})}));
+    GPR_ASSERT(delegate_config.ok());
+    args.config = *delegate_config;
```
markdroth, line 96: This can use `std::move()`.

markdroth (approved):

test/cpp/interop/rpc_behavior_lb_policy.cc
eugeneo: Done

PR #33534, opened 2023-06-23
[ObjC] run cf event engine test suit on ios

HannahShiSFB:
Run cf event engine test suit on ios

CC: @sampajano  @mikefan1991

### DIFF
--- a/src/objective-c/tests/BUILD
+++ b/src/objective-c/tests/BUILD
@@ -333,6 +333,23 @@ grpc_objc_ios_unit_test(
     deps = [":CFStreamTests-lib"],
 )
 
+grpc_objc_testing_library(
+    name = "EventEngineTests-lib",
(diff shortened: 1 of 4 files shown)


sampajano (approved):
Thanks Hannah!!

Do you mean that before this PR, the tests were not run? Could you clarify what exactly was covered by test before and after this change (maybe in PR description)? Thanks!!

HannahShiSFB:
These tests were only run on Mac, so do all the tests in provided by c-core (`/test/` directory) except  the old iomgr cfstream tests (`test/core/iomgr/ios/CFStreamTests/`).

All other iOS tests are under `src/objective-c/tests/`. They are currently run with iomgr but tested with event engine in https://github.com/grpc/grpc/pull/32791. We need a new CI job to run them with event engine enabled.

PR #33535, opened 2023-06-23
[csharp] reintroduce base_namespace experimental option to C# (with a patch)

jtattermusch:
Reintroduces patched version of https://github.com/grpc/grpc/pull/32636 (which was reverted in https://github.com/grpc/grpc/pull/32957).

Together with cl/542843305, the internal build should work fine.

Supersedes https://github.com/grpc/grpc/pull/33310 (since a patch is also needed).


jtattermusch:
@tonydnewell

jtattermusch:
The internal build is passing, so I'm going the assume that my fix worked and this PR is ready to merge.

src/compiler/csharp_generator_helpers.h
```diff
@@ -26,9 +26,25 @@ namespace grpc_csharp_generator {
...
+  // Otherwise the behavior remains the same as before.
+  if (base_namespace.empty()) {
+    out_file = grpc_generator::FileNameInUpperCamel(file, false) + file_suffix;
+  } else {
+    out_file = GRPC_CUSTOM_CSHARP_GETOUTPUTFILE(file, file_suffix, true,
+                                                base_namespace, error);
+    if (out_file.empty()) {
+      return false;
+    }
+  }
```
felschr, line 47: I'm not entirely sure what happens here, but with `protoc`'s `base_namespace` option I can specify `base_namespace=` (yes, without a value after the `=`) and it'll generate the code into a folder structure instead of putting everything into the root directory.
I expected the same behavior for grpc, but it still tries to put all files into the root directory, which in my case causes issues with duplicate file names.

Since `GRPC_CUSTOM_CSHARP_GETOUTPUTFILE()` already handles the `base_namespace.empty()` case perhaps the check in grpc just needs to be removed?:
https://github.com/protocolbuffers/protobuf/blob/1bfcedaf1280923f36e52a6fa0a5645ab63568f0/src/google/protobuf/compiler/csharp/names.cc#L137-L153

felschr: Just found the empty string behavior in protoc documented here:
https://github.com/protocolbuffers/protobuf/blob/1bfcedaf1280923f36e52a6fa0a5645ab63568f0/src/google/protobuf/compiler/csharp/csharp_options.h#L63-L67

So, essentially when `base_namespace` is explicitly set via the CLI the variable `base_namespace_specified` is set to `true` which in turn causes `GetOutputFile()` to be invoked with `generate_directories` set to `true`.

To match the behavior, grpc should also check if `base_namespace` was explicitly set and use that in the conditional, e.g.:
```suggestion
  if (!base_namespace_specified) {
    out_file = grpc_generator::FileNameInUpperCamel(file, false) + file_suffix;
  } else {
    out_file = GRPC_CUSTOM_CSHARP_GETOUTPUTFILE(file, file_suffix, true,
                                                base_namespace, error);
    if (out_file.empty()) {
      return false;
    }
  }
```

No need to removing this check entirely to retain backwards compatibility. See https://github.com/grpc/grpc/pull/32636/files#r1148360058 for some notes on that.

felschr: I've created a new issue for this: https://github.com/grpc/grpc/issues/34113

PR #33540, opened 2023-06-23
[xDS LB] xDS pick first support

eugeneo:


test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
```diff
@@ -0,0 +1,170 @@
...
+      logical_dns_cluster_resolver_response_generator_;
+};
+
+// Run both with and without load reporting, just for test coverage.
+INSTANTIATE_TEST_SUITE_P(
+    XdsTest, PickFirstTest,
+    ::testing::Values(XdsTestType(), XdsTestType().set_enable_load_reporting()),
+    &XdsTestType::Name);
+
+TEST_P(PickFirstTest, PickFirstConfigurationIsPropagated) {
```
markdroth, line 112: I think this test is way more complicated than it needs to be.  I think that what we actually want here is something similar to the logic that we have in the [existing unit test](https://github.com/grpc/grpc/blob/21f2eba14389ddb8346dfd16ac2c306037d3c7e9/test/core/client_channel/lb_policy/pick_first_test.cc#L261), with just a few differences.  In particular, note that the goal here is just to verify that the shuffle option is being passed in to the pick_first policy; we don't need to actually verify the behavior of that option, since we're already doing that in the unit test.  So I think that all we really need to do is to check that pick_first chooses a backend randomly each time we send it an update.

Also, note that we don't need multiple CDS resource names, nor do we need multiple EDS resource names.  We can update the default CDS resource just once at the start of the test to configure it to use pick_first, and then we don't need to modify it again.  And we can populate the default EDS resource with the set of backends we want, and then we can update that EDS resource whenever we want -- updating the EDS resource will send an updated address list to the LB policy, even if the new list is identical to the old list.

I suggest something like the following:

```
CreateAndStartBackends(3);
auto eds_resource = BuildEdsResource(EdsResourceArgs(
    {{"locality0", CreateEndpointsForBackends()}}));
// Change cluster to use pick_first with shuffle option.
auto cluster = default_cluster_;
PickFirst pick_first;
pick_first.set_shuffle_address_list(true);
cluster.mutable_load_balancing_policy()
    ->add_policies()
    ->mutable_typed_extension_config()
    ->mutable_typed_config()
    ->PackFrom(pick_first);
balancer_->ads_service()->SetCdsResource(cluster);
// Give 10 tries to get a different backend.
size_t last_seen_backend = backends_.size();
bool saw_different_backend = false;
for (size_t i = 0; i < 10; ++i) {
  // Update EDS resource.  This will send a new address list update to the LB policy.
  balancer_->ads_service()->SetEdsResource(eds_resource);
  // Send an RPC and figure out which backend it went to.
  CheckRpcSendOk(DEBUG_LOCATION);
  size_t backend_index = 0;
  for (; backend_index < backends_.size(); ++backend_index) {
    if (backends_[backend_index]->backend_service()->request_count() > 0) {
      break;
    }
  }
  ASSERT_LT(backend_index, backends_.size());
  // If we found a different backend than last time, then the test succeeded.
  if (last_seen_backend < backends_.size() &&
      backend_index != last_seen_backend) {
    saw_different_backend = true;
    break;
  }
  last_seen_backend = backend_index;
  // Restart the backend to force the client to go IDLE before the
  // next iteration.
  ShutdownBackend(backend_index);
  EXPECT_TRUE(channel_->WaitForStateChange(
      GRPC_CHANNEL_READY, grpc_timeout_seconds_to_deadline(5)));
  EXPECT_THAT(channel_->GetState(false), GRPC_CHANNEL_IDLE);
  StartBackend(backend_index);
}
EXPECT_TRUE(saw_different_backend);
```

markdroth (commented):
The overall structure looks good here!  Comments are mostly minor -- the only noteworthy issues are the one about the env var protection and the one about simplifying the e2e test.

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
eugeneo: I arrived to something halfway between my previous version and your proposal. I am toggling between endpoints to ensure EDS update was propagated. Also this assertion - `EXPECT_THAT(channel_->GetState(false), GRPC_CHANNEL_IDLE);` was always failing as it would go to READY. The code seems to work without this now.

eugeneo (commented):
Thank you for the review!

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
```diff
@@ -0,0 +1,174 @@
...
+        ->mutable_typed_extension_config()
+        ->mutable_typed_config()
+        ->PackFrom(pick_first);
+
+    // Set ordered cluster configuration
+    balancer_->ads_service()->SetCdsResource(cluster);
+  }
+
+  grpc_core::RefCountedPtr<grpc_core::FakeResolverResponseGenerator>
+      logical_dns_cluster_resolver_response_generator_;
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/xds/xds_lb_policy_registry_test.cc
```diff
@@ -591,6 +593,44 @@ TEST(XdsLbPolicyRegistryTest, MaxRecursion) {
               ::testing::EndsWith("error:exceeded max recursion depth of 16]"));
 }
 
+//
+// PickFirst
```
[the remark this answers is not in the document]
eugeneo: Missed the comment, addressing now

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
```diff
@@ -0,0 +1,136 @@
...
+  bool saw_different_backend = false;
+  size_t start_index = 0;
+  for (size_t i = 0; i < 100 && !saw_different_backend; ++i) {
+    // Update EDS resource.  This will send a new address list update to the LB
+    // policy.
+    balancer_->ads_service()->SetEdsResource(BuildEdsResource(
+        EdsResourceArgs({{"locality0", CreateEndpointsForBackends(
+                                           start_index, start_index + 3)}})));
+    auto result = WaitForAnyBackendHit(start_index, start_index + 3);
+    size_t backend_index = absl::get<size_t>(result);
```
markdroth, line 106: If the variant holds a status, then this will crash.  I think you need to check if it failed before extracting the value.

The code here will become much cleaner once you switch to using `StatusOr<>`:

```
auto backend_index = WaitForAnyBackendHit(start_index, start_index + 3);
ASSERT_TRUE(backend_index.ok()) << backend_index.status();
if (*backend_index != start_index) return;
```

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
markdroth: This approach looks good.

The channel won't go IDLE in this approach, because there's no point at which one of the existing connections fails.  We simply switch from one connection to another due to the address update.

markdroth (commented):
This test approach looks good!  Remaining comments are relatively minor.

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
```diff
@@ -0,0 +1,136 @@
...
+
+namespace grpc {
+namespace testing {
+namespace {
+
+using ::envoy::extensions::load_balancing_policies::pick_first::v3::PickFirst;
+
+class PickFirstTest : public XdsEnd2endTest {
+ protected:
+  absl::variant<Status, size_t> WaitForAnyBackendHit(size_t start, size_t end) {
```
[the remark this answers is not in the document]
eugeneo: Done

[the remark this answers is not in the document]
eugeneo: Done

eugeneo (commented):

I ran the tests 2000 times and I verified that internal import works.

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
```diff
@@ -94,23 +93,23 @@ TEST_P(PickFirstTest, PickFirstConfigurationIsPropagated) {
...
-  bool saw_different_backend = false;
   size_t start_index = 0;
-  for (size_t i = 0; i < 100 && !saw_different_backend; ++i) {
+  for (size_t i = 0; i < 100; ++i) {
     // Update EDS resource.  This will send a new address list update to the LB
     // policy.
     balancer_->ads_service()->SetEdsResource(BuildEdsResource(
         EdsResourceArgs({{"locality0", CreateEndpointsForBackends(
                                            start_index, start_index + 3)}})));
     auto result = WaitForAnyBackendHit(start_index, start_index + 3);
```
markdroth, line 103: Suggestions to improve readability:
- Instead of taking the value out of the `StatusOr<>` and putting it in another variable, just access it directly using `*backend_index`.
- Instead of writing `(backend_index - start_index) > 0`, it seems clearer to say `*backend_index != start_index`.
- If the body of an `if` statement is just a single statement like `return`, it can usually fit on the same line without braces.

I think this reduces 6 lines to just 3 and is actually more readable:

```
auto backend_index = WaitForAnyBackendHit(start_index, start_index + 3);
ASSERT_TRUE(backend_index.ok()) << backend_index.status();
if (*backend_index != start_index) return;
```

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
```diff
@@ -94,23 +93,23 @@ TEST_P(PickFirstTest, PickFirstConfigurationIsPropagated) {
...
+    ASSERT_TRUE(result.ok()) << result.status();
+    size_t backend_index = *result;
+    if ((backend_index - start_index) > 0) {
+      return;
+    }
     // Toggle between backends 0-2 and 3-5
     start_index = 3 - start_index;
   }
-  EXPECT_TRUE(saw_different_backend);
+  FAIL() << "did not choose a different backend after 100 tries";
```
markdroth, line 112: Suggest removing the "after 100 tries" part, because it's likely to get out of sync with the actual number used in the loop above.

markdroth (approved):
This looks great!  Remaining comments are minor.

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
eugeneo: Done

test/cpp/end2end/xds/xds_pick_first_end2end_test.cc
eugeneo: Done

PR #33543, opened 2023-06-24
[experiments] Add specification of different default values for different platforms.

Vignesh2208:
It introduces the following syntax:

The following would mark the experiment as broken on ios, false on windows and debug on posix. If a platform is un-specified, the default for that platform will be set to false. Refer to test/core/experiments/fixtures/test_experiments_rollout.yaml for examples which are tested.

- name: experiment_1
   default:
      &nbsp; ios: broken
      &nbsp; windows: false
      &nbsp; posix: debug
      
  It also supports the already existing syntax and interprets it as just specifying one default for all platforms.
    
Supported platform tags: ios, windows, posix


tools/codegen/core/experiments_compiler.py
```diff
@@ -394,6 +419,62 @@ def GenerateExperimentsHdr(self, output_file, mode):
             print(file=H)
             print(f"#endif  // {include_guard}", file=H)
 
+    def _GenerateExperimentsSrcForPlatform(self, platform, mode, file_desc):
+        print("namespace {", file=file_desc)
+        have_defaults = set()
+        for _, exp in self._experiment_definitions.items():
+            print(
+                "const char* const description_%s = %s;"
+                % (exp.name, ToCStr(exp.description)),
```
drfloob, line 428: For another cleanup, maybe: this file uses both `"%s" % var` formatting and `f"{var}"` formatting styles. This PR introduces more of both. One or the other should suffice, and consistency would be good.

drfloob (approved):

tools/codegen/core/experiments_compiler.py
Vignesh2208: Thanks for approving. I'll make the cleanup in another PR incase we need to rollback

PR #33548, opened 2023-06-26
[otel] Add bazel dependency

yashykt:
Add bazel dependency on opentelemetry-cpp.


test/cpp/ext/filters/otel/BUILD
```diff
@@ -0,0 +1,39 @@
...
+    srcs = [
+        "otel_plugin_test.cc",
+    ],
+    external_deps = [
+        "gtest",
+        "otel/api",
+        "otel/sdk/src/metrics",
+    ],
+    language = "C++",
+    linkstatic = True,
```
veblush, line 35: Why is this required?

veblush (approved):

test/cpp/ext/filters/otel/BUILD
yashykt: I dunno, I just copied it over. I'll remove it

PR #33553, opened 2023-06-26
[Fuzzing] Fix use-after-free in resolver_fuzzer

yijiem:
In FuzzingDNSResolver, capturing the engine as raw pointers in the lambda functions instead of capturing the `this` pointer. By the time the lambda is ran, the FuzzingDNSResolver might already be destroyed but the engine should still be alive.

### DIFF
--- a/test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
+++ b/test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
@@ -154,50 +154,49 @@ class FuzzingResolverEventEngine
     void LookupHostname(LookupHostnameCallback on_resolve,
                         absl::string_view /* name */,
                         absl::string_view /* default_port */) override {
-      CheckAndSetOrphan(ExecutionStep::DURING_LOOKUP_HOSTNAME, engine_);
+      engine_->CheckAndSetOrphan(ExecutionStep::DURING_LOOKUP_HOSTNAME);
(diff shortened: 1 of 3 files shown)


test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
```diff
@@ -154,40 +154,45 @@ class FuzzingResolverEventEngine
...
         });
       }
     }
 
    private:
-    void CheckAndSetOrphan(ExecutionStep current_execution_step) {
-      if (engine_->should_orphan_at_step_ == current_execution_step) {
-        *engine_->done_resolving_ = true;
-        engine_->has_been_orphaned_ = true;
+    static void CheckAndSetOrphan(ExecutionStep current_execution_step,
```
drfloob, line 191: Since this doesn't touch any `FuzzingDNSResolver` internals, and since it might be called after the `FuzzingDNSResolver` object is destroyed, I suggest moving this method into `FuzzingResolverEventEngine`. The signature can remain the as it was before.

drfloob (commented):
Good catch!

test/core/ext/filters/event_engine_client_channel_resolver/resolver_fuzzer.cc
yijiem: Good point! Done.

drfloob (approved):

PR #33558, opened 2023-06-27
Remove use of SSL_CTX_set_client_CA_list in ssl_transport_security.cc.

matthewstevenson88:


include/grpc/grpc_security.h
```diff
@@ -564,6 +566,9 @@ GRPCAPI grpc_server_credentials*
 grpc_ssl_server_credentials_create_with_options(
     grpc_ssl_server_credentials_options* options);
 
+GRPCAPI void grpc_ssl_server_credentials_set_send_client_ca_list(
```
markdroth, line 569: I think this method should operate on `grpc_ssl_server_credentials_options`, not directly on `grpc_server_credentials`.

include/grpc/grpc_security.h
```diff
@@ -526,7 +528,7 @@ GRPCAPI grpc_server_credentials* grpc_ssl_server_credentials_create(
    authenticate client certificates.*/
 GRPCAPI grpc_server_credentials* grpc_ssl_server_credentials_create_ex(
     const char* pem_root_certs, grpc_ssl_pem_key_cert_pair* pem_key_cert_pairs,
-    size_t num_key_cert_pairs,
+    size_t num_key_cert_pair,
```
markdroth, line 531: Why rename this parameter?  The plural seems more grammatically correct to me.

src/core/lib/security/credentials/ssl/ssl_credentials.h
```diff
@@ -120,6 +125,7 @@ class grpc_ssl_server_credentials final : public grpc_server_credentials {
       grpc_ssl_pem_key_cert_pair* pem_key_cert_pairs, size_t num_key_cert_pairs,
       grpc_ssl_client_certificate_request_type client_certificate_request);
 
+  bool send_client_ca_list_ = true;
```
markdroth, line 128: I think this field belongs in `grpc_ssl_server_credentials_options` instead of here.

markdroth (commented):
Overall, looks good, but needs a bit of cleanup in terms of how the SslCreds option is plumbed through.

include/grpc/grpc_security.h
```diff
@@ -896,6 +901,21 @@ GRPCAPI void grpc_tls_credentials_options_set_crl_directory(
...
+/**
+ * EXPERIMENTAL API - Subject to change
+ *
+ * Sets whether or not a TLS server should send a list of CA names in the
+ * ServerHello. This list of CA names is read from the server's trust bundle, so
+ * that the client can use this list as a hint to know which certificate it
+ * should send to the server.
+ *
+ * WARNING: This API is extremely dangerous and should not be used. If the
+ * server's trust bundle is too large, then the TLS server will be unable to
```
erm-g, line 913: Does it make sense to add smth like '"too large" definition depends on OpenSSL lib being used'?

matthewstevenson88: Good call, done.

include/grpc/grpc_security.h
matthewstevenson88: Done.

include/grpc/grpc_security.h
matthewstevenson88: Not sure how/why I changed this, reverted.

src/core/lib/security/credentials/ssl/ssl_credentials.h
matthewstevenson88: Done.

matthewstevenson88:
Thanks for the reviews @markdroth and @erm-g, I think the PR is ready for another pass.

include/grpc/grpc_security.h
```diff
@@ -528,9 +528,9 @@ GRPCAPI grpc_server_credentials* grpc_ssl_server_credentials_create(
    authenticate client certificates.*/
 GRPCAPI grpc_server_credentials* grpc_ssl_server_credentials_create_ex(
     const char* pem_root_certs, grpc_ssl_pem_key_cert_pair* pem_key_cert_pairs,
-    size_t num_key_cert_pair,
+    size_t num_key_cert_pairs,
     grpc_ssl_client_certificate_request_type client_certificate_request,
-    void* reserved);
+    bool send_client_ca_list, void* reserved);
```
markdroth, line 533: In this particular case, I don't think we want to add an argument to this existing method, for several reasons.

The back-story here is that when we first adding some of these options to SslCreds, there was some confusion about our policy with regard to backward-incompatible changes in the C-core API.  Our actual policy is that non-backward-compatible changes are fine, and the only requirement is that we need to publish a gRFC, as per [gRFC P3](https://github.com/grpc/proposal/blob/master/P3-grfcs-for-core-api-changes.md).  However, when these changes were added, for some reason we were thinking that we couldn't change the original `grpc_ssl_server_credentials_create()` API, so we added this `grpc_ssl_server_credentials_create_ex()` API with an alternative set of arguments.  And then later, we tried to generalize it more by adding `grpc_ssl_server_credentials_create_with_options()` (which is a good approach, and you'll notice that's the only interface we provide in the newer TlsCreds API).  But the result of all of this is that we now have 3 different methods for creating `grpc_ssl_server_credentials`, which is awful -- and if it wasn't for the fact that our long-term goal is to deprecate SslCreds in favor of TlsCreds, I would be pushing to clean this up.

The bottom line here is that we can make breaking changes, but it will require a gRFC, which has an inherent 10-day waiting period, which I don't think is what you want here.  (Technically, we should require a gRFC even for a non-breaking pure addition to the API, such as adding the `grpc_ssl_server_credentials_options_set_send_client_ca_list()` method.  But I'm willing to ignore that in this case, since (a) it's really unlikely that this addition will prove problematic or controversial, and (b) I don't care so much about long-term consequences, since we want to eventually deprecate SslCreds anyway.  But I am not comfortable making a breaking change to an existing API without a gRFC.)

Also, since we're going to want this PR back-ported to 1.56, we can't have any breaking changes here anyway, since that would require bumping the major version of C-core.  We generally only bump the major version of C-core in between minor versions, not in between two patch versions of the same minor version.

So, please don't add this parameter to this method.  Instead, just change any caller using `grpc_ssl_server_credentials_create_ex()` that needs to set this option to instead use `grpc_ssl_server_credentials_create_with_options()`.

matthewstevenson88: Thanks for the explanation, that makes sense. For the sake of getting this patch out quickly, I'm going to remove the changes to SslCreds from this PR (we can add this knob in a future PR) and turn the focus to TlsCreds only.

matthewstevenson88: Thanks @markdroth, the PR is ready for another pass. :)

include/grpc/grpc_security.h
```diff
@@ -21,6 +21,8 @@
 
 #include <grpc/support/port_platform.h>
 
+#include <stdbool.h>
```
markdroth, line 24: Note: This header was added in C99, and historically the C-core headers have been C89, which precludes use of `bool`.  But since we're already in the process of moving from C to C++ for these headers, I think this is fine.

markdroth (approved):

include/grpc/grpc_security.h
matthewstevenson88: Ack SG. If it ends up causing problems and needs to be removed, please let me know and I can refactor to use `int` instead of `bool`.

matthewstevenson88:
Thanks for the review Mark!

PR #33561, opened 2023-06-27
[python O11Y] Refactor census propagation flow.

XuanWang-Amos:
Refactored OpenCensus context propagation flow, now propagation happens for each call and context will be automatically propagated from gRPC server to gRPC client.

 We're using `execution_context` in OpenCensus since the context is related to OpenCensus and it helps wrap `contextVar` for us.

### Testing
* Added a new Bazel test case for context propagation.

### DIFF
--- a/src/python/grpcio_observability/grpc_observability/_gcp_observability.py
+++ b/src/python/grpcio_observability/grpc_observability/_gcp_observability.py
@@ -56,6 +56,8 @@ GRPC_STATUS_CODE_TO_STRING = {
     grpc.StatusCode.DATA_LOSS: "DATA_LOSS",
 }
 
+GRPC_SPAN_CONTEXT = "grpc_span_context"
+
 
 @dataclass
 class GcpObservabilityPythonConfig:
@@ -165,11 +167,12 @@ class GCPOpenCensusObservability(grpc._observability.ObservabilityPlugin):
     def create_client_call_tracer(
         self, method_name: bytes
     ) -> ClientCallTracerCapsule:
-        current_span = execution_context.get_current_span()
-        if current_span:
-            # Propagate existing OC context
-            trace_id = current_span.context_tracer.trace_id.encode("utf8")
-            parent_span_id = current_span.span_id.encode("utf8")
+        grpc_span_context = execution_context.get_opencensus_attr(
+            GRPC_SPAN_CONTEXT
+        )
+        if grpc_span_context:
+            trace_id = grpc_span_context.trace_id.encode("utf8")
+            parent_span_id = grpc_span_context.span_id.encode("utf8")
             capsule = _cyobservability.create_client_call_tracer(
                 method_name, trace_id, parent_span_id
             )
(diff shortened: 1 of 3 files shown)


src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -35,6 +35,9 @@
 _STREAM_UNARY = "/test/StreamUnary"
 _STREAM_STREAM = "/test/StreamStream"
 STREAM_LENGTH = 5
+PORT = 50051
```
gnossen, line 38: We cannot introduce a fixed port. This makes it impossible to run multiple instances of the test in parallel.

src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -35,6 +35,9 @@
...
+TRIGGER_RPC_METADATA = ("control", "triger_rpc")
```
gnossen, line 39: Nit: `s/triger/trigger/`

src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -157,13 +162,29 @@ def testRecordUnaryUnary(self):
...
+        # Verify that the trace_id was propagated to the 2nd RPC.
+        self._set_config_file(_VALID_CONFIG_TRACING_ONLY)
+        with grpc_observability.GCPOpenCensusObservability(
+            exporter=self.test_exporter
+        ):
+            self._start_server(port=PORT)
+            unary_unary_call(port=PORT, metadata=TRIGGER_RPC_METADATA_TUPLE)
+
+        # 2 of each for ["Recv", "Sent", "Attempt"]
+        self.assertEqual(len(self.all_span), 6)
```
gnossen, line 183: I think it would be useful to have a test that records one more RPC in the client -> server -> server chain. The tricky part is making sure that the context from an incoming request propagates to an outgoing request. While this _mostly_ does that, it's possible for an implementation to record an outgoing RPC as belonging to a trace (since the first server is responsible for sending this info out) but for the RPC to not actually carry the trace information, resulting in the second server not receiving this trace.

The most straightforward way to to do this would be to have the server be sort of "recursive" by sending a request to itself, but this runs the risk of causing false positive tests; if the server thread that receives the response is the same as the server thread that sends the request, then the test can pass trivially without any trace context propagation functionality happening at all. I would suggest either going multiprocess or having multiple servers in the same process, each with their own thread pool.

src/python/grpcio_observability/grpc_observability/_gcp_observability.py
```diff
@@ -197,10 +200,12 @@ def save_trace_context(
         trace_options = trace_options_module.TraceOptions(0)
         trace_options.set_enabled(is_sampled)
         span_context = span_context_module.SpanContext(
-            trace_id=trace_id, span_id=span_id, trace_options=trace_options
+            trace_id=trace_id,
+            span_id=span_id,
+            trace_options=trace_options,
+            from_header=True,
```
gnossen, line 206: https://opencensus.io/api/python/_modules/opencensus/trace/span_context.html#SpanContextWhat does this do? A quick scan of [the file]() doesn't turn up anything obvious.

gnossen: Can you please provide a reference?

XuanWang-Amos: I'm taking a similar approach to other propagators (like [BinaryFormatPropagator](https://opencensus.io/api/python/_modules/opencensus/trace/propagation/binary_format.html#BinaryFormatPropagator.from_header)), but since it's an optional field, I'm fine to remove it if you have specific concerns.

XuanWang-Amos:
@gnossen PTALA

src/python/grpcio_tests/tests/observability/_observability_test.py
```diff
@@ -413,6 +445,41 @@ def _validate_spans(
             self.assertTrue(prefix_exist)
 
 
+def unary_unary_call(port, metadata=None):
+    with grpc.insecure_channel(f"localhost:{port}") as channel:
+        multi_callable = channel.unary_unary(_UNARY_UNARY)
+        if metadata:
```
gnossen, line 451: Nit: You don't need the conditional. You can just do `metadata = metadata or ()`

gnossen (approved):

PR #33563, opened 2023-06-27
[git attributes] add Package.swift to list of generated files

markdroth:
This will prevent this auto-generated file from showing up in github diffs by default.


sampajano (approved):
lgtm thanks! :)

PR #33571, opened 2023-06-28
[resolvers] use proper %-encoding of authority by default

markdroth:
- Change the `ResolverFactory::GetDefaultAuthority()` method to %-encode the authority by default, so individual resolver impls don't need to remember to do this.
- Remove the hack in the xds resolver for setting the authority to everything after the last `/` character.
- Change the `unix`, `unix-abstract`, and `vsock` resolvers to use a real authority instead of hard-coding to "localhost".

### DIFF
--- a/src/core/lib/resolver/resolver_factory.h
+++ b/src/core/lib/resolver/resolver_factory.h
@@ -68,7 +68,7 @@ class ResolverFactory {
   /// Returns a string representing the default authority to use for this
   /// scheme.
   virtual std::string GetDefaultAuthority(const URI& uri) const {
-    return std::string(absl::StripPrefix(uri.path(), "/"));
+    return URI::PercentEncodeAuthority(absl::StripPrefix(uri.path(), "/"));
   }
 };
 
--- a/src/core/lib/resolver/resolver_factory.h
+++ b/src/core/lib/resolver/resolver_factory.h
@@ -66,7 +66,8 @@ class ResolverFactory {
   virtual OrphanablePtr<Resolver> CreateResolver(ResolverArgs args) const = 0;
 
   /// Returns a string representing the default authority to use for this
-  /// scheme.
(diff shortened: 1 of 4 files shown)


src/core/lib/resolver/resolver_factory.h
```diff
@@ -68,7 +68,7 @@ class ResolverFactory {
   /// Returns a string representing the default authority to use for this
   /// scheme.
```
anicr7, line 69: Can we add a comment here that indicates that by default we URL encode the path?

anicr7 (approved):

src/core/lib/resolver/resolver_factory.h
markdroth: Done.

XuanWang-Amos:
@markdroth Is there a spec for why we're making this change? Looks like this change is not expected from Kestrel server.

markdroth:
@XuanWang-Amos We had a cross-language discussion about this, and the argument was that because of the way that browsers encode URIs, the authority would never wind up having certain characters, and there is some existing library code out in the wild that doesn't handle this well.  @ejona86 @dfawley 

What is the exact case that is causing a problem here?  What resolver is being used, what character is being %-encoded, and what server is unhappy with that?

markdroth:
Oh, looks like the issue is #34305.

If the C# server is not liking the %-encoded character, then the client can override the authority by setting the `GRPC_ARG_DEFAULT_AUTHORITY` channel arg when creating the channel.

XuanWang-Amos:
I see, thanks for the info, looks like this is expected then.

markdroth:
I don't know of any reason why using a % there is illegal.  I suspect the C# server is just being too picky.  But feel free to follow up on that in #34305.

dfawley:
I thought this was a gRPC-C# server, but it looks like it's a third party server implementation that we can't do anything about directly.

PR #33582, opened 2023-06-30
[EventEngine] Only use fork handlers when enabled via an environment variable

drfloob:


src/core/lib/event_engine/forkable.cc
```diff
@@ -35,50 +36,67 @@ bool g_registered ABSL_GUARDED_BY(g_mu){false};
...
 Forkable::~Forkable() { StopManagingForkable(this); }
 
 void RegisterForkHandlers() {
-  grpc_core::MutexLock lock(g_mu.get());
-  if (!std::exchange(g_registered, true)) {
-    pthread_atfork(PrepareFork, PostforkParent, PostforkChild);
+  if (IsForkEnabled()) {
+    grpc_core::MutexLock lock(g_mu.get());
+    if (!std::exchange(g_registered, true)) {
+      pthread_atfork(PrepareFork, PostforkParent, PostforkChild);
```
apolcyn, line 54: optional: now that we have an env var guard, we might be able to remove some code here by narrowing the scope of `GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK` to specifically the `pthread_atfork` calls.

For example:

```
void RegisterForkHandlers() {
  if (IsForkEnabled()) {
    grpc_core::MutexLock lock(g_mu.get());
    if (!std::exchange(g_registered, true)) {
#ifdef GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK
      pthread_atfork(PrepareFork, PostforkParent, PostforkChild);
#endif
    }
  }
```

apolcyn (approved):

src/core/lib/event_engine/forkable.cc
drfloob: Good idea, done.

I also improved the tests so that forkable behavior is exercised whether or not `GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK` is defined.

src/core/lib/event_engine/forkable.cc
```diff
@@ -17,11 +17,12 @@
 #include "src/core/lib/event_engine/forkable.h"
 
 #ifdef GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK
```
apolcyn, line 19: nit: initial GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK not needed now

test/core/event_engine/forkable_test.cc
```diff
@@ -92,7 +100,44 @@ TEST_F(ForkableTest, BasicPthreadAtForkOperations) {
     }
   }
 }
-#endif  // GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK
+
+TEST_F(ForkableTest, NonPthreadManualForkOperations) {
+  // Manually simulates a fork event for non-pthread-enabled environments
+#ifdef GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK
+  return;
```
apolcyn, line 107: nit: I think you can use GTEST_SKIP here

test/core/event_engine/forkable_test.cc
```diff
@@ -92,7 +100,44 @@ TEST_F(ForkableTest, BasicPthreadAtForkOperations) {
...
+#endif
+
+  class SomeForkable : public Forkable {
+   public:
+    void PrepareFork() override { prepare_called_ = true; }
+    void PostforkParent() override { parent_called_ = true; }
+    void PostforkChild() override { child_called_ = true; }
+
+    // bool prepare_called() { return prepare_called_; }
```
apolcyn, line 116: nit: remove commented-out code

test/core/event_engine/forkable_test.cc
```diff
@@ -92,7 +100,44 @@ TEST_F(ForkableTest, BasicPthreadAtForkOperations) {
...
+      EXPECT_EQ(child_called_, child);
+    }
+
+   private:
+    bool prepare_called_ = false;
+    bool parent_called_ = false;
+    bool child_called_ = false;
+  };
+
+  SomeForkable forkable;
```
apolcyn, line 132: does this test need the env var set?

apolcyn (approved):

src/core/lib/event_engine/forkable.cc
drfloob: That protects the pthread include. gRPC is compiled in places that don't have pthread support (e.g, Windows).

test/core/event_engine/forkable_test.cc
drfloob: Oh good point. These tests now force-enable fork support, but exercise the Forkable with and without pthread support enabled. I think that's a fair compromise. By default, all other core & C++ tests exercise the forkable code when fork support is disable, so this is decent coverage.

test/core/event_engine/forkable_test.cc
drfloob: Yep, thanks. Done

test/core/event_engine/forkable_test.cc
drfloob: Done

PR #33589, opened 2023-07-01
[PSM Interop] Update Kokoro to Ubuntu 22.04

sergiitk:
Makes the changes necessary to run the new PSM interop framework on Ubuntu 22.04:

- Install dependencies via apt: `kubectl`, `google-cloud-sdk-gke-gcloud-auth-plugin` (previously were pre-provisioned or available as part of gcloud distribution)
- Use venv instead of pyenv
- Use python 3.10 instead of python3 .9

Other changes:
- Do not update gcloud components - the one provisioned is relatively recent, and expected to be updated as new base images are released
- Unpin pip from `21.0.1`. Not sure if we're OK about using the latest one `venv --upgrade-deps`, or if we should just pin it to something more recent (currently it resolves to `pip 22.0.2` and `setuptools 68.0.0`)

ref b/274944592, cl/547690787

### DIFF
--- a/tools/internal_ci/linux/grpc_xds_k8s_install_test_driver.sh
+++ b/tools/internal_ci/linux/grpc_xds_k8s_install_test_driver.sh
@@ -16,7 +16,7 @@
 set -eo pipefail
 
 # Constants
-readonly PYTHON_VERSION="${PYTHON_VERSION:-3.9}"
+readonly PYTHON_VERSION="${PYTHON_VERSION:-3.10}"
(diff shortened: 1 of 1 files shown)


sergiitk:
Tests:

- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/43dda248-57e0-49a0-9c84-6fcfcf077d8c)

gnossen (approved):

PR #33590, opened 2023-07-01
[ObjC] run ios tests with event engine enabled

HannahShiSFB:

### DIFF
--- a/tools/internal_ci/macos/grpc_objc_bazel_test.sh
+++ b/tools/internal_ci/macos/grpc_objc_bazel_test.sh
@@ -119,3 +119,33 @@ objc_bazel_tests/bazel_wrapper \
   -- \
   "${EXAMPLE_TARGETS[@]}" \
   "${TEST_TARGETS[@]}"
+
+
+# Enable event engine and run tests again.
+TEST_TARGETS=(
+  //src/objective-c/tests:InteropTestsLocalCleartext
+  //src/objective-c/tests:InteropTestsLocalSSL
+  //src/objective-c/tests:InteropTestsRemote
+  //src/objective-c/tests:MacTests
+  //src/objective-c/tests:UnitTests
+  //src/objective-c/tests:tvtests_build_test
+  # codegen plugin tests
+  //src/objective-c/tests:objc_codegen_plugin_test
+  //src/objective-c/tests:objc_codegen_plugin_option_test
+)
+
+python3 tools/run_tests/python_utils/bazel_report_helper.py --report_path objc_event_engine_bazel_tests
+
+objc_event_engine_bazel_tests/bazel_wrapper \
+  --bazelrc=tools/remote_build/include/test_locally_with_resultstore_results.bazelrc \
+  test \
+  --google_credentials="${KOKORO_GFILE_DIR}/GrpcTesting-d0eeee2db331.json" \
+  "${BAZEL_REMOTE_CACHE_ARGS[@]}" \
+  $BAZEL_FLAGS \
+  --cxxopt=-DGRPC_IOS_EVENT_ENGINE_CLIENT=1 \
+  --test_env=GRPC_EXPERIMENTS=event_engine_client \
+  --test_env=GRPC_VERBOSITY=debug --test_env=GRPC_TRACE=event_engine,api \
+  "${OBJC_TEST_ENV_ARGS[@]}" \
+  -- \
+  "${EXAMPLE_TARGETS[@]}" \
+  "${TEST_TARGETS[@]}"


HannahShiSFB:
@jtattermusch I was trying to add a new job that runs objc tests with event engine enabled, however I couldn't find the new job in this PR.
Do you by chance know whether there are additional configs need to be updated at the kokoro side? Can you help on this?

--
CC: @sampajano

jtattermusch:
@HannahShiSFB yes, in order to add a new kokoro job, the job needs to be created internally first. But at the same time, the job for PRs can't be created unless the configuration for it already exists (and is tested), since otherwise a failure would start showing up on everyone else's PRs.

So what you probably need to do:
- test the setup you already have in this PR by hijacking an existing test job. This can be done by redirecting the existing .cfg file to your new .sh test script (I usually do that in a separate  "DO NOT MERGE" commit): https://github.com/grpc/grpc/pull/33590/files#diff-0a198360ef70347906ee1fc68fd1c64e20fc86c671a726f19f0ce5393202f718R18

- is it possible to run the new tests under the existing test job? we don't want to create more test jobs that necessary since they're hard to maintain (and as you saw, you also did a lot of copy and pasting of the test scripts which is a maintenance burden).

- we can add a new kokoro job if needed but it needs to be clear that we've investigated the alternatives first and we should make effort to minimize future maintenance costs.

HannahShiSFB:
@jtattermusch Thanks for the info.
1. I have another PR #33591 that modifies the existing job to run the same tests and it was successful.
2. I tried to update the existing job `tools/internal_ci/macos/grpc_objc_bazel_test.sh` that run the targets with a different configuration (event engine enabled) however I couldn't find them in the job invocation logs, not sure if it is because they are same targets or it is log report wrapper related. Do you have any suggestions?

jtattermusch:
> @jtattermusch Thanks for the info.
> 
> 1. I have another PR [[ObjC] enable event engine in objc tests #33591](https://github.com/grpc/grpc/pull/33591) that modifies the existing job to run the same tests and it was successful.
> 2. I tried to update the existing job `tools/internal_ci/macos/grpc_objc_bazel_test.sh` that run the targets with a different configuration (event engine enabled) however I couldn't find them in the job invocation logs, not sure if it is because they are same targets or it is log report wrapper related. Do you have any suggestions?

It should be possible to run 2 bazel invocations (or more) in the same job and get a separate result link for each of them.

Basically, it requires a separate invocation of bazel_report_helper.py  for each bazel invocation.  The objc. bazel test already has 2 bazel invocations (1. build interop server 2. run objc tests), it shouldn't be hard to add 3. objc test with event engine enabled. If you show me what exactly you tried, I might be able to point you in the right direction.

HannahShiSFB:
Thanks @jtattermusch I've update the existing objc bazel test with a new objc_event_engine_bazel_tests target which enabled event engine and the tests passed.
This PR is ready to review.

tools/internal_ci/macos/grpc_objc_bazel_test.sh
```diff
@@ -119,3 +119,33 @@ objc_bazel_tests/bazel_wrapper \
   -- \
   "${EXAMPLE_TARGETS[@]}" \
   "${TEST_TARGETS[@]}"
+
+
+# Enable event engine and run tests again.
+TEST_TARGETS=(
```
jtattermusch, line 125: nit: let's call this EVENT_ENGINE_TEST_TARGETS  instead and update the command below instead of redefining TEST_TARGETS?

tools/internal_ci/macos/grpc_objc_bazel_test.sh
```diff
@@ -119,3 +119,33 @@ objc_bazel_tests/bazel_wrapper \
...
+  //src/objective-c/tests:InteropTestsLocalCleartext
+  //src/objective-c/tests:InteropTestsLocalSSL
+  //src/objective-c/tests:InteropTestsRemote
+  //src/objective-c/tests:MacTests
+  //src/objective-c/tests:UnitTests
+  //src/objective-c/tests:tvtests_build_test
+  # codegen plugin tests
+  //src/objective-c/tests:objc_codegen_plugin_test
```
jtattermusch, line 133: sounds like the plugin_test and plugin_option test  won't be any different under event engine (they only run the protoc code AFAICT), so you might as well skip them here?

jtattermusch (approved):
This seems fair. Even though running multiple different test suites one by one isn't exactly the best practice, it's preferrable compared to creating new kokoro jobs (with a lot of copy and paste of not-so-simple objc bazel test config) and then needing to maintain them. 

LGTM after addressing the comments.

jtattermusch:
Ok, looks like this was automerged. Please address the comments as a followup PR.

PR #33611, opened 2023-07-05
[ssl] Temporarily increase default BIO buffer size from 17kb to 30kb.

matthewstevenson88:
There is a bug in the SSL stack that was only partially fixed in #29176: if more than 17kb is written to the BIO buffer, then everything over 17kb will be discarded, and the SSL handshake will fail with a bad record mac error.

It's relatively uncommon to hit this bug, because the TLS handshake messages need to be much larger than normal for you to have a chance of hitting this bug. However, there was a separate bug in the SSL stack (recently fixed in #33558) that causes the ServerHello produced by a gRPC-C++ TLS server to grow linearly in size with the size of the trust bundle. 

The correct fix to this bug is to (a) keep the default BIO buffer size at 17kb, and (b) add logic in `ssl_transport_security.cc` to never write more than 17kb at one time to the BIO. This is a bit delicate to do (specifically, setting up a test to reproduce the problem) and will happen in a follow-up PR, but this PR is being sent as a temporary mitigation for clients built from g3 that are often failing because of the combination of these 2 bugs.

To emphasize: this PR is temporary and will be replaced with a long-term fix when it is ready, likely early next week.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -309,8 +309,6 @@ let package = Package(
         "src/core/ext/transport/chttp2/transport/internal.h",
         "src/core/ext/transport/chttp2/transport/parsing.cc",
         "src/core/ext/transport/chttp2/transport/stream_lists.cc",
-        "src/core/ext/transport/chttp2/transport/stream_map.cc",
-        "src/core/ext/transport/chttp2/transport/stream_map.h",
         "src/core/ext/transport/chttp2/transport/varint.cc",
         "src/core/ext/transport/chttp2/transport/varint.h",
         "src/core/ext/transport/chttp2/transport/writing.cc",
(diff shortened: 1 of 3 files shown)


ctiller:
Does this affect memory consumed post-handshake?

matthewstevenson88:
> Does this affect memory consumed post-handshake?

Yes. Some details:
- When you create the SSL handshaker, you also create the "BIO" object which handles I/O between the application and OpenSSL/BoringSSL. There is a buffer that underlies the "BIO" object whose default length we are increasing.
- This "BIO" object is moved to the SSL frame protector after the handshake completes, and it is destroyed when the SSL frame protector is destroyed.

I added some more context to the PR description on why we want this change. Please note that this change is temporary only while we put together a more robust fix.

WDYT? Do you think this temporary change is likely to create other problems?

matthewstevenson88:
Closing, in favor of waiting for the more robust fix instead.

Issue #33615, opened 2023-07-06
[ruby] Unable to build grpc gem if a more recent version of re2 is installed in include path of Ruby's CPPFLAGS


stanhu:
Currently the `grpc` gem cannot be compiled if:

1. A recent version of re2 is installed with a custom prefix.
2. The Ruby interpreter is compiled with `CPPFLAGS` with that prefix.

For example, we have:

1. re2 `2023-03-01` compiled with a `prefix` of `/opt/gitlab/embedded`.
1. The Ruby interpreter is built with `CPPFLAGS` with `-I/opt/gitlab/embedded/include`.

When compiling the grpc gem, the grpc-core `Makefile` will attempt to include `/opt/gitlab/embedded/include/re2/re2.h` instead of the local `third_party/re2/re2.h` file, which uses re2 `2022-04-01`. The method definitions are different from re2 `2023-03-01`, so the compilation fails. 
  
As discussed in https://github.com/grpc/grpc/pull/32580#issuecomment-1462231263, the resulting behavior of `extconf.rb` changed over grpc versions:

1. Prior to v1.48.0, the Ruby build process did not include any external `CPPFLAGS` while building grpc-core.
5. Starting with v1.48.0, https://github.com/grpc/grpc/pull/27660 passed along `ENV['CPPFLAGS']` while building grpc-core.
6. v1.56.0 attempted to fix this via https://github.com/grpc/grpc/pull/32580 by passing in `RbConfig::CONFIG['CPPFLAGS']` instead of `ENV['CPPFLAGS']`. However, I overlooked that the Ruby compiler flags may also have `CPPFLAGS` that give precedence to the newer re2 path:

```ruby
irb(main):001:0> RbConfig::CONFIG['CPPFLAGS']
=> "-I/opt/gitlab/embedded/include -O3 -D_FORTIFY_SOURCE=2 -fstack-protector -O3 -g -pipe -fno-omit-frame-pointer -I/opt/gitlab/embedded/include  "
```

The key problem here is that no matter what flags are supplied by the Ruby `extconf.rb`, I believe grpc-core should always give precedence to the the `third_party` directory and ensure `re2` headers are included during the build process.

I attempted to do this in https://github.com/grpc/grpc/commit/37a6ff0ee53b175a6d0950e6a99c3f7277db57d2, but according to https://github.com/grpc/grpc/pull/32580#issuecomment-1468748087 building with the `Makefile` is deprecated.

### Describe the solution you'd like

I think we have a few options:

[...]

PR #33616, opened 2023-07-06
[ruby] Fix re2 compilation when different system version installed

stanhu:
re2 or abseil previously failed to compile if:

1. A different and incompatible version were installed with a non-standard system prefix, such as `/opt/local`.
2. The Ruby interpreter has CPPFLAGS to include the prefix, such as `CPPFLAGS=-I/opt/local/include`.

Running `make` would result in method prototype mismatches because the Makefile would previously attempt to use the headers from `/opt/local/include/re2` before the `third_party/re2/re2` directory.

To ensure the current directories receive priority, include the relevant re2 and abseil directories in the include path.

Relates to https://github.com/grpc/grpc/issues/33615


stanhu:
I'm going to close this even though https://github.com/grpc/grpc/issues/33615 is still an issue. a8508ec46573cbf009a41c867f74af976aa5514e dropped `libre2` flags from the `Makefile`.

We've resolved our issue internally by dropping the system libre2 and using precompiled, native gems in the `re2` gem.

PR #33625, opened 2023-07-06
[protobuf] Update third_party/protobuf to 3.23.3

drfloob:


veblush:
Having a lot of files under `tools/distrib/python/grpcio_tools/third_party` doesn't make sense. I guess one of our scripts needs a change to deal with this. And we did 22.x and 23.x protobuf upgrade manually so I suspect this problem was there but we didn't have a chance to notice it. (Also upb should be upgraded at the same time, which AJ is already aware of)

drfloob:
Superseded by https://github.com/grpc/grpc/pull/33695, done manually. I also added a fix for `make_grpcio_tools.py`, the upgrade script should not have the same problem next time.

PR #33638, opened 2023-07-07
[ssl] Fix SSL stack to handle large handshake messages whose length exceeds the BIO buffer size.

matthewstevenson88:
There is a bug in the SSL stack that was only partially fixed in #29176: if more than 17kb is written to the BIO buffer, then everything over 17kb will be discarded, and the SSL handshake will fail with a bad record mac error or hang if not enough bytes have arrived yet.

It's relatively uncommon to hit this bug, because the TLS handshake messages need to be much larger than normal for you to have a chance of hitting this bug. However, there was a separate bug in the SSL stack (recently fixed in #33558) that causes the ServerHello produced by a gRPC-C++ TLS server to grow linearly in size with the size of the trust bundle; these 2 bugs combined to cause a large number of TLS handshake failures for gRPC-C++ clients talking to gRPC-C++ servers when the server had a large trust bundle.

This PR fixes the bug by ensuring that all bytes are successfully written to the BIO buffer. An initial quick fix for this bug was planned in #33611, but abandoned because we were worried about temporarily doubling the memory footprint of all SSL channels.

The complexity in this PR is mostly in the test: it is fairly tricky to force gRPC-C++'s SSL stack to generate a sufficiently large ServerHello to trigger this bug.

### DIFF
--- a/src/core/tsi/ssl_transport_security.cc
+++ b/src/core/tsi/ssl_transport_security.cc
@@ -65,6 +65,7 @@
 
 // --- Constants. ---
 
+#define TSI_SSL_MAX_BIO_WRITE_ATTEMPTS 100
 #define TSI_SSL_MAX_PROTECTED_FRAME_SIZE_UPPER_BOUND 16384
 #define TSI_SSL_MAX_PROTECTED_FRAME_SIZE_LOWER_BOUND 1024
 #define TSI_SSL_HANDSHAKER_OUTGOING_BUFFER_INITIAL_SIZE 1024
@@ -1538,15 +1539,42 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
   // If there are received bytes, process them first.
   tsi_ssl_handshaker* impl = reinterpret_cast<tsi_ssl_handshaker*>(self);
   tsi_result status = TSI_OK;
-  size_t bytes_consumed = received_bytes_size;
   size_t bytes_written = 0;
   if (received_bytes_size > 0) {
-    status = ssl_handshaker_process_bytes_from_peer(impl, received_bytes,
-                                                    &bytes_consumed, error);
-    while (status == TSI_DRAIN_BUFFER) {
-      status = ssl_handshaker_write_output_buffer(self, &bytes_written, error);
-      if (status != TSI_OK) return status;
-      status = ssl_handshaker_do_handshake(impl, error);
+    unsigned char* remaining_bytes_to_write_to_network_io =
+        const_cast<unsigned char*>(received_bytes);
+    std::size_t remaining_bytes_to_write_to_network_io_size =
+        received_bytes_size;
+    std::size_t number_bio_write_attempts = 0;
+    while (remaining_bytes_to_write_to_network_io_size > 0 &&
+           (status == TSI_OK || status == TSI_INCOMPLETE_DATA) &&
+           number_bio_write_attempts < TSI_SSL_MAX_BIO_WRITE_ATTEMPTS) {
+      number_bio_write_attempts++;
+
+      // Try to write all of the remaining bytes to the BIO.
+      std::size_t bytes_written_to_network_io =
+          remaining_bytes_to_write_to_network_io_size;
+      status = ssl_handshaker_process_bytes_from_peer(
+          impl, remaining_bytes_to_write_to_network_io,
+          &bytes_written_to_network_io, error);
+
+      // As long as the BIO is full, drive the SSL handshake to consume bytes
+      // from the BIO. If the SSL handshake returns any bytes, write them to the
+      // peer.
+      while (status == TSI_DRAIN_BUFFER) {
+        status =
+            ssl_handshaker_write_output_buffer(self, &bytes_written, error);
+        if (status != TSI_OK) return status;
+        status = ssl_handshaker_do_handshake(impl, error);
+      }
+
+      // Move the pointer to the first byte not yet successfully written to the
+      // BIO.
+      remaining_bytes_to_write_to_network_io_size -=
+          bytes_written_to_network_io;
+      if (remaining_bytes_to_write_to_network_io_size > 0) {
+        remaining_bytes_to_write_to_network_io += bytes_written_to_network_io;
+      }
     }
   }
   if (status != TSI_OK) return status;
--- a/src/core/tsi/ssl_transport_security.cc
+++ b/src/core/tsi/ssl_transport_security.cc
@@ -1543,21 +1543,18 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
   if (received_bytes_size > 0) {
     unsigned char* remaining_bytes_to_write_to_network_io =
         const_cast<unsigned char*>(received_bytes);
-    std::size_t remaining_bytes_to_write_to_network_io_size =
-        received_bytes_size;
-    std::size_t number_bio_write_attempts = 0;
+    size_t remaining_bytes_to_write_to_network_io_size = received_bytes_size;
+    size_t number_bio_write_attempts = 0;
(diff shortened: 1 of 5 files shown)


matthewstevenson88:
@zeromath Could you also take a quick look at this PR?

test/core/tsi/transport_security_test_lib.cc
```diff
@@ -664,3 +680,71 @@ void tsi_test_frame_protector_fixture_destroy(
...
+  ASN1_GENERALIZEDTIME* infinite_future = ASN1_GENERALIZEDTIME_new();
+  GPR_ASSERT(
+      ASN1_GENERALIZEDTIME_set_string(infinite_future, "99991231235959Z"));
+  GPR_ASSERT(X509_set1_notAfter(x509, infinite_future));
+  ASN1_GENERALIZEDTIME_free(infinite_future);
+
+  // Set the subject DN.
+  X509_NAME* subject_name = X509_NAME_new();
+  GPR_ASSERT(X509_NAME_add_entry_by_txt(
+      subject_name, "CN", MBSTRING_ASC,
```
zeromath, line 714: micro-nit: /*field=*/"CN", similar below.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1538,15 +1539,42 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
...
+            ssl_handshaker_write_output_buffer(self, &bytes_written, error);
+        if (status != TSI_OK) return status;
+        status = ssl_handshaker_do_handshake(impl, error);
+      }
+
+      // Move the pointer to the first byte not yet successfully written to the
+      // BIO.
+      remaining_bytes_to_write_to_network_io_size -=
+          bytes_written_to_network_io;
+      if (remaining_bytes_to_write_to_network_io_size > 0) {
```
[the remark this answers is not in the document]
matthewstevenson88: If we do that, then on the last write we will push the `remaining_bytes_to_write_to_network_io` out of the allowed range `[received_bytes, received_bytes + received_bytes_size]`. If the pointer was used after that, then it would be undefined behavior.

test/core/tsi/transport_security_test_lib.cc
matthewstevenson88: Done.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -65,6 +65,7 @@
 
 // --- Constants. ---
 
+#define TSI_SSL_MAX_BIO_WRITE_ATTEMPTS 100
```
erm-g, line 68: As we discussed offline, let's do a follow up PR to expose `100` as another parameter.

erm-g (approved):

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: SG, filed and a bug for tracking and CC-ed you.

src/core/tsi/ssl_transport_security.cc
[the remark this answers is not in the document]
matthewstevenson88: Yup fair point, updated.

zeromath:
LGTM!

(I didn't find a button for this, maybe it's because I don't have the permission to do so.)

matthewstevenson88:
@markdroth Friendly ping. :)

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1538,15 +1539,40 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
...
-  size_t bytes_consumed = received_bytes_size;
   size_t bytes_written = 0;
   if (received_bytes_size > 0) {
-    status = ssl_handshaker_process_bytes_from_peer(impl, received_bytes,
-                                                    &bytes_consumed, error);
-    while (status == TSI_DRAIN_BUFFER) {
-      status = ssl_handshaker_write_output_buffer(self, &bytes_written, error);
-      if (status != TSI_OK) return status;
-      status = ssl_handshaker_do_handshake(impl, error);
+    unsigned char* remaining_bytes_to_write_to_network_io =
```
markdroth, line 1544: The names of these two variables are confusing.  If I'm understanding the code correctly (but please correct me if I'm not), these aren't the bytes being written to the network; they're the bytes from the peer that we're reading from the network.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1538,15 +1539,40 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
...
+        const_cast<unsigned char*>(received_bytes);
+    std::size_t remaining_bytes_to_write_to_network_io_size =
```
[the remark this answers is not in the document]
matthewstevenson88: Done.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1538,15 +1539,40 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
...
+        received_bytes_size;
+    std::size_t number_bio_write_attempts = 0;
+    while (remaining_bytes_to_write_to_network_io_size > 0 &&
+           (status == TSI_OK || status == TSI_INCOMPLETE_DATA) &&
+           number_bio_write_attempts < TSI_SSL_MAX_BIO_WRITE_ATTEMPTS) {
+      number_bio_write_attempts++;
```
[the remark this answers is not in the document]
matthewstevenson88: Done.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1538,15 +1539,40 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
...
+
```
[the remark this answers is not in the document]
matthewstevenson88: Done.

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Agreed that the wording is a bit confusing. Let me clarify what's happening first: we're getting bytes from the peer, and writing these bytes to OpenSSL via the `BIO` object, and the name of this `BIO` object variable is `network_io`.

I'm happy to send a follow-up PR to rename the `BIO` object variable (because it causes issues like this throughout this file), but I'd prefer to keep that change out-of-scope for this PR if it's OK.

In the meantime, I've changed the variable name to something else to avoid confusion.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1541,23 +1541,20 @@ static tsi_result ssl_handshaker_next(tsi_handshaker* self,
   tsi_result status = TSI_OK;
   size_t bytes_written = 0;
   if (received_bytes_size > 0) {
-    unsigned char* remaining_bytes_to_write_to_network_io =
+    unsigned char* remaining_bytes_to_write_to_openssl =
```
markdroth, line 1544: Do we actually need these two variables in the first place?  Can't we just modify the `received_bytes` and `received_bytes_size` variables as we consume the data?

markdroth (approved):

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: `received_bytes_size` is needed later on when calculating the unused bytes, so we would at least need to copy the initial value of this argument. Rather than doing that, I'm leaning towards keeping the current version since I think it's more readable. More than happy to change it if you disagree though.

markdroth: Okay, this is fine, then.

markdroth (approved):

matthewstevenson88:
Thanks all for the reviews!

PR #33639, opened 2023-07-08
[tracing] Fix flakiness in tcp_posix_test

Vignesh2208:
tcp_posix_test is incorrectly assuming that all endpoint_writes with timestamps enabled will be successfully traced. Remove the timestamps checking related tests to prevent flakes when the test is enabled internally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -793,9 +793,6 @@ if(gRPC_BUILD_TESTS)
   if(_gRPC_PLATFORM_LINUX OR _gRPC_PLATFORM_POSIX)
     add_dependencies(buildtests_c static_stride_scheduler_benchmark)
   endif()
-  if(_gRPC_PLATFORM_LINUX OR _gRPC_PLATFORM_POSIX)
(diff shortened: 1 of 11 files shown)


yousukseung:
I think it's preferred to give up and miss that timestamp in such a case (which is the case already?) and only fix if we currently end up leaking or crashing in such cases.

On one hand, the way Fathom timestamp handling is implemented with iomgr today is somewhat limited to only take the first timestamp for each kind (we need all to debug rtx, qdisc delay etc) so I guess missing the timestamp won't be a big deal. But more importantly AFAICT this effectively serializes all writes to the socket which I assume will have a much larger impact.

ctiller:
Writes are already always serialized (hard to make concurrent writes to the same socket make sense).

yousukseung:
Fine to disable the test (for now). 

For context, the way traced buffers are added today depends on sendmsg() to accept the whole range, which holding the lock longer only slightly helps but won't guarantee and Fathom may still miss timestamps, so we decided the fix is not worth it.

yousukseung (approved):

PR #33649, opened 2023-07-10
Add annotation with metadata limits and sizes 

ananda1066:


src/core/ext/transport/chttp2/transport/hpack_parser.cc
```diff
@@ -1104,20 +1130,31 @@ grpc_error_handle HPackParser::Parse(const grpc_slice& slice, bool is_last) {
...
     }
     global_stats().IncrementHttp2MetadataSize(state_.frame_length);
+    if (call_tracer != nullptr && metadata_buffer_ != nullptr) {
+      std::string metadata_annotation = absl::StrCat(
+          "gRPC metadata soft_limit:",
+          state_.metadata_early_detection.soft_limit(),
+          ",hard_limit:", state_.metadata_early_detection.hard_limit(), ",");
+      MetadataSizeEncoder encoder(metadata_annotation);
+      metadata_buffer_->Encode(&encoder);
+      call_tracer->RecordAnnotation(metadata_annotation);
```
ctiller, line 1156: @yashykt @markdroth for general input here too: my sense here is that we should add a custom method (or enum AnnotationType) to the `CallTracer` interface for something like this, so that `CallTracer` implementations can choose whether they want to trace this data or not -- the set of things that might be traced will certainly grow, and different customers are going to want different things -- knowledge of which is best gathered in one place rather than scattered throughout the codebase (where possible)

markdroth: Yeah, I agree that that would be good.  This seems like exactly the kind of annotation that we would want to be opt-in for OSS.

For now, an `enum AnnotationType` is probably fine.  In the long run, I suspect that we'll need something more sophisticated, since we'll need a way to support opt-in metrics.  But I don't think we should add anything like that until we have a concrete use-case.

ananda1066: Hm how granular would this type be? I think Yash had the idea previously of something like: 
```
enum class AnnotationType {
  kProto,
  kJson,
  kString, // no specific format
  kUnknown
};
```
but if we want to opt-in for specific annotations,  it might look something more like: `{kDefault, kMetadataSizes}`. 

Also, in this case, we want to avoid encoding metadata sizes on every call for performance reasons, instead using `s->traced` to limit encoding to only those calls that are Dapper traced, which means OSS will never see the annotation. @yashykt would the `IsSampled()` function in calltracer work instead to rate-limit how often we construct the annotation? This would allow us to opt-in to the annotation in OSS as well.

ctiller: Chatted with @markdroth a little this morning... and we were converging on something that I think looks like the following:

```
class CallTracer {
 public:
  // enum associated with the concrete type of Annotation (see below)
  // use an enum here for any types that need to be down-casted to - and specifically
  // an enum so we can get good code generation for switch statements in call tracer
  // implementations
  // (this means that things aren't completely open extensibility: if a call tracer needs
  // to be able to down cast to a type then we need to add a name here - but no other
  // details - this is an explicit trade-off to ensure reasonable code generation - which
  // is a requirement in this space)
  enum AnnotationType {
    kUnknown,
    kMetadataSizes, // annotation type must be grpc_core::MetadataSizesAnnotation
    kDoNotUse_MustBeLast // force default: clause
  };
  // base class defines how to pull out a string to report
  class Annotation {
   public:
    explicit Annotation(AnnotationType type) : type_(type) {}
    AnnotationType type() { return type_; }
    virtual std::string ToString() const = 0;

   private:
    const AnnotationType type_;
  };
  // record some arbitrary annotation
  void RecordAnnotation(const Annotation& annotation)
};

// with elsewhere defined:

class MetadataSizesAnnotation : public CallTracer::Annotation {
 public:
  MetadataSizesAnnotation(uint64_t soft, uint64_t hard) : Annotation(AnnotationType::kMetadataSizes), soft_(soft), hard_(hard) {}

  uint64_t soft() const { return soft_; }
  uint64_t hard() const { return hard_; }
  std::string ToString() const { return absl::StrCat("soft:", soft_, " hard:", hard_); }
 
 private:
  uint64_t soft_;
  uint64_t hard_;
};
```

markdroth: Yeah, what Craig wrote looks like the right approach.  My only suggestion is that it might be a good idea to use [`UniqueTypeName`](https://github.com/grpc/grpc/blob/master/src/core/lib/gprpp/unique_type_name.h) here instead of an enum, since that way we don't need a central registry of all the types.

ctiller: Just to close the loop here - pretty adamant that it not be something like `UniqueTypeName` -- doing so would create big if/else chains in CallTracer implementations that can't be resolved at compile time (`UniqueTypeName` doesn't get its bit representation until program initialization).

I want to be able to write:

```
void MyCallTracer::RecordAnnotation(const Annotation& annotation) {
  switch (annotation.type()) {
    case AnnotationType::kGeneric: puts(annotation.ToString().c_str()); break;
    case AnnotationType::kMetadataSizes: /* do something cool and unique */ break;
    // etc...
  }
}
```

Understood that this necessitates some lightweight central registry, but this stuff is going to be called with very high frequency and so we need to reduce the cost per call as far as we reasonably can.

markdroth: Okay, that makes sense.  I am concerned that it may become a maintenance burden in the long run, but I'm fine with starting that way and reevaluating if it becomes a problem.

PR #33656, opened 2023-07-11
[iomgr][EventEngine] Improve server handling of file descriptor exhaustion

drfloob:


src/core/lib/gprpp/time.h
```diff
@@ -37,8 +37,8 @@
     uint64_t now = grpc_core::Timestamp::FromTimespecRoundDown( \
                        gpr_now(GPR_CLOCK_MONOTONIC))            \
                        .milliseconds_after_process_epoch();     \
-    uint64_t prev_tsamp = prev.exchange(now);                   \
-    if (prev_tsamp == 0 || now - prev_tsamp > (n)*1000) {       \
+    if (prev == 0 || now - prev > (n)*1000) {                   \
```
Vignesh2208, line 40: why are we not updating prev all the time ? This could have an un-desired effect of some logs not getting printed at the correct frequency

src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
```diff
@@ -138,6 +138,22 @@ void PosixEngineListenerImpl::AsyncConnectionAcceptor::NotifyOnAccept(
...
+          // either a read event triggers the on_read callback, or time has
+          // passed and the accept should be re-tried regardless. This callback
+          // is not cancelled, so a spurious wakeup may occur even when there's
+          // nothing to accept. This is not a performant code path, but if an fd
+          // limit has been reached, the system is likely in an unhappy state
+          // regardless.
+          GRPC_LOG_EVERY_N_SEC(1, GPR_ERROR, "%s",
+                               "File descriptor limit reached. Retrying.");
+          handle_->NotifyOnRead(notify_on_accept_);
+          if (retry_timer_armed_.exchange(false)) return;
```
Vignesh2208, line 153: should this be retry_timer_armed_.exchange(true) ?

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
```diff
@@ -121,6 +126,8 @@ class PosixEngineListenerImpl
     ListenerSocketsContainer::ListenerSocket socket_;
     EventHandle* handle_;
     PosixEngineClosure* notify_on_accept_;
+    PosixEngineClosure* retry_closure_;
```
Vignesh2208, line 129: retry_closure also needs to be deleted explicitly since its a permanent closure. Otherwise it'll be a memory leak. Maybe it can be a self deleting closure instead of a permanent closure ?  Or maybe the timer needs to be cancelled and closure needs to be deleted in the destructor

src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
drfloob: This has been replaced with a mutex and a TaskHandle (to support the required cancellation before engine destruction).

src/core/lib/event_engine/posix_engine/posix_engine_listener.h
drfloob: As discussed, I replaced it with an AnyInvocable.

src/core/lib/gprpp/time.h
drfloob: `now - prev_tsamp` is forever a small single-digit number of millis if you always update `prev` to `now`, and call this macro in a tight loop. When have you seen this working?

drfloob: I've reverted this fix since it's unrelated to the PR. I'll land it separately.

Vignesh2208 (approved):

glaubitz:
Is there any chance this could be backported to the 1.2x series?

drfloob:
@glaubitz No, the 1.2x releases are 3 to 4 years old, it would be far more effort than it's worth for us to try to provide support going that far back. I highly recommend staying up to date with the [most recent releases](https://github.com/grpc/grpc/releases); numerous bug fixes and various improvements have landed in the past few years.

glaubitz:
OK, thanks. I will have to update grpc itself then which requires some effort due to the target being an LTS distribution.

Mingli-001:
@drfloob 
Could you help to confirm these below questions?
1, Is this the fix for CVE-2023-4785?
2, For some reason, I need to still use 1.36.4, could you help to backport the fix for 1.36.4?
Thanks very much!

PR #33671, opened 2023-07-12
[LB policies] use helper GetAuthority() instead of GRPC_ARG_SERVER_URI channel arg

markdroth:


src/core/ext/filters/client_channel/lb_policy/grpclb/grpclb.cc
```diff
@@ -1372,10 +1366,6 @@ ChannelArgs BuildBalancerChannelArgs(
             // Strip out the service config, since we don't want the LB policy
             // config specified for the parent channel to affect the LB channel.
             .Remove(GRPC_ARG_SERVICE_CONFIG)
-            // The channel arg for the server URI, since that will be different
-            // for the LB channel than for the parent channel.  The client
-            // channel factory will re-add this arg with the right value.
-            .Remove(GRPC_ARG_SERVER_URI)
```
apolcyn, line 1378: Why do we need to filter out this channel arg here? Is this an extra cleanup, or is it necessary for the PR to work? (from my reading, I can't see why this change is needed)

markdroth: This is a bit tangential to the PR, but it seemed reasonable to clean it up, since I was looking at all occurences of this channel arg anyway.

Basically, we used to use the C representation of channel args throughout C-core.  That representation essentially stored the args as a vector, so adding an arg did not replace any previous value of the arg; instead, it added a duplicate element to the vector, and the duplicate entry would almost certainly not actually be used, since code using the args would search through the vector from the start until they found the key they want, which would result in returning the original value, not the duplicate value.  So the code here needed to make sure that it didn't pass down this arg, lest the arg have the wrong value in the balancer channel.

We have subsequently switched to a new internal representation of channel args that is essentially a map, so setting an arg will automatically replace any previous value of the arg.  And since the balancer channel will set the arg itself, that value will replace any value that we might pass in here, so it no longer matters whether we remove the arg here.

apolcyn (approved):
lgtm

PR #33674, opened 2023-07-12
[ruby] simplify shutdown; remove unnecessary attempts at grpc_shutdown

apolcyn:
So far, we've structured global C-core init/shutdown as follows:
1) every grpc-ruby object calls `grpc_init` when allocated, and `grpc_shutdown` when finalized
2) grpc-ruby background threads are each wrapped in a `grpc_init/shutdown` pair - for example see https://github.com/grpc/grpc/blob/b32d94de05550ed15f4394b9e4b33bee733dfd89/src/ruby/ext/grpc/rb_event_thread.c#L122 and https://github.com/grpc/grpc/blob/b32d94de05550ed15f4394b9e4b33bee733dfd89/src/ruby/ext/grpc/rb_event_thread.c#L136

But because ruby doesn't join ruby threads when the process is terminating, the `init/shutdown` pairs in 2) are always left open. I.e., thus far we have never been invoking the final call to `grpc_shutdown` which actually does C-core global shutdown. Thus our calls to `grpc_shutdown` are useless.

So we might as well keep things simple and not even attempt to call `grpc_shutdown`. Now we just call `grpc_init` before using C-core, and then never even attempt global shutdown (in non-forking situations).

As a bonus, this fixes the issue with the event thread's `grpc_ruby_init` racing with `prefork` that is mentioned in https://github.com/grpc/grpc/pull/33666

### DIFF
--- a/src/ruby/ext/grpc/rb_call_credentials.c
+++ b/src/ruby/ext/grpc/rb_call_credentials.c
@@ -193,7 +193,6 @@ static void grpc_rb_call_credentials_free_internal(void* p) {
 /* Destroys the credentials instances. */
 static void grpc_rb_call_credentials_free(void* p) {
   grpc_rb_call_credentials_free_internal(p);
-  grpc_ruby_shutdown();
 }
 
 /* Protects the mark object from GC */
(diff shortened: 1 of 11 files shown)


apolcyn:
cc @HannahShiSFB @alto-ruby FYI

stanley-cheung (approved):
I wonder if we should visit this in PHP @HannahShiSFB

HannahShiSFB:
@stanley-cheung Did you see any issues related in PHP or have specific concerns?

PR #33688, opened 2023-07-13
Release callback function before unreferencing call

ywelsch:
The lifetime of the callback function exceeded that of the call reference.


eugeneo:
I do not see any difference in behavior...

PR #33696, opened 2023-07-13
[tracing] Add new RecordAnnotation method to CallTracer API

ananda1066:
Adding a new method for custom Annotation types (per design suggested in #33649) to unblock metadata annotation.


src/core/lib/channel/call_tracer.h
```diff
@@ -49,11 +49,30 @@ namespace grpc_core {
 // The base class for all tracer implementations.
 class CallTracerAnnotationInterface {
  public:
+  // Enum associated with types of Annotations.
+  enum AnnotationType {
```
ctiller, line 53: let's make this an `enum class`

src/core/lib/channel/call_tracer.h
```diff
@@ -49,11 +49,30 @@ namespace grpc_core {
...
+    kDoNotUse_MustBeLast,
+  };
+
+  // Base class to define a new type of annotation.
+  class Annotation {
+   public:
+    explicit Annotation(AnnotationType type) : type_(type) {}
+    AnnotationType type() const { return type_; }
+    virtual std::string ToString() const = 0;
+    virtual ~Annotation() {}
```
ctiller, line 64: we don't need a virtual destructor here... instead make a protected non-virtual destructor.

reasoning: we never want to destroy from the base pointer, we always destroy the most derived type at the callsite, so there's no reason to add bloat (extra vtable entry, extra virtual calls)

src/core/lib/channel/call_tracer.h
```diff
@@ -49,11 +49,30 @@ namespace grpc_core {
...
+
+   private:
+    AnnotationType type_;
```
ctiller, line 67: This could be const

src/cpp/ext/filters/census/client_filter.cc
```diff
@@ -288,6 +288,15 @@ void OpenCensusCallTracer::OpenCensusCallAttemptTracer::RecordAnnotation(
   context_.AddSpanAnnotation(annotation, {});
 }
 
+void OpenCensusCallTracer::OpenCensusCallAttemptTracer::RecordAnnotation(
+    const Annotation& annotation) {
+  switch (annotation.type()) {
+    default:
+      // If tracing is disabled, the following will be a no-op.
+      context_.AddSpanAnnotation(annotation.ToString(), {});
```
ctiller, line 296: hmmm there will be some cost here (the cost of calling `ToString` and then throwing it away) - maybe we want to lift that check to the top of this function?

src/cpp/ext/filters/census/client_filter.cc
```diff
@@ -359,6 +368,14 @@ void OpenCensusCallTracer::RecordAnnotation(absl::string_view annotation) {
   context_.AddSpanAnnotation(annotation, {});
 }
 
+void OpenCensusCallTracer::RecordAnnotation(const Annotation& annotation) {
+  switch (annotation.type()) {
+    default:
+      // If tracing is disabled, the following will be a no-op.
+      context_.AddSpanAnnotation(annotation.ToString(), {});
```
ctiller, line 375: Same cost analysis as earlier

src/cpp/ext/filters/census/server_call_tracer.cc
```diff
@@ -160,6 +160,13 @@ class OpenCensusServerCallTracer : public grpc_core::ServerCallTracer {
     context_.AddSpanAnnotation(annotation, {});
   }
 
+  void RecordAnnotation(const Annotation& annotation) override {
+    switch (annotation.type()) {
+      default:
+        context_.AddSpanAnnotation(annotation.ToString(), {});
```
ctiller, line 166: Same cost analysis... maybe move the check in here?

ctiller (changes requested):
Really good start, some minor feedback and I think this is gtg

src/core/lib/channel/call_tracer.h
yousukseung, line 53: Since it's nested anyway, should we just name it "Type" to reduce clutter?

src/core/lib/channel/call_tracer.h
```diff
@@ -49,11 +49,30 @@ namespace grpc_core {
...
+  };
+
   virtual ~CallTracerAnnotationInterface() {}
   // Records an annotation on the call attempt.
   // TODO(yashykt): If needed, extend this to attach attributes with
   // annotations.
   virtual void RecordAnnotation(absl::string_view annotation) = 0;
+  virtual void RecordAnnotation(const Annotation& annotation) = 0;
```
yousukseung, line 75: No need to pass an enum by reference.

ctiller: @yousukseung this is not the enum, it's the enclosing class/interface

yousukseung: Oh sorry, closing.

src/cpp/ext/filters/census/client_filter.cc
ananda1066: Yeah I debated doing that but had left it as is since that was what the original RecordAnnotation function did. I changed both now.

src/cpp/ext/filters/census/server_call_tracer.cc
ananda1066: Updated.

src/core/lib/channel/call_tracer.h
ananda1066: Done

src/core/lib/channel/call_tracer.h
ananda1066: Done

src/core/lib/channel/call_tracer.h
ananda1066: Done

ctiller (approved):

XuanWang-Amos:
This change [broke python o11y](https://source.cloud.google.com/results/invocations/86dbcf6e-c908-4ab7-97c2-0467b741a91a/targets/%2F%2Fsrc%2Fpython%2Fgrpcio_observability%2Fgrpc_observability:cyobservability/log) since it also uses CallTracer API, can you please take a look?

ananda1066:
Fixed with #33899.

PR #31747, opened 2022-11-22
Build native MacOS arm64 artifacts (universal2)

gnossen:
This PR builds all MacOS artifacts as [`universal2`](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) artifacts by default. This builds the `.whl` with a shared object library supporting both the `x86_64` and `arm64` platforms:

```
rbellevi-macbookpro:tmp.u9QVKZQH rbellevi$ file ./grpc/_cython/cygrpc.cpython-311-darwin.so
./grpc/_cython/cygrpc.cpython-311-darwin.so: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit bundle x86_64Mach-O 64-bit bundle x86_64] [arm64]
./grpc/_cython/cygrpc.cpython-311-darwin.so (for architecture x86_64):	Mach-O 64-bit bundle x86_64
./grpc/_cython/cygrpc.cpython-311-darwin.so (for architecture arm64):	Mach-O 64-bit bundle arm64
```

This also removes the hack introduced by https://github.com/grpc/grpc/pull/29857 since we are now truly building universal artifacts. Fixes #29262

PR #33700, opened 2023-07-13
[boringssl] Simplify BoringSSL assembly build

davidben:
~~NB: I haven't tested this at all and am hoping the CI will tell me where I've (undoubtedly) messed something up.~~ Edit: looks like CI is now clear!

BoringSSL's gas-compatible assembly files, like its C files, are now wrapped with preprocessor ifdefs to capture which platforms each file should be enabled on. This means that, provided the platform can process .S files it all (i.e. not Windows), we no longer need to detect the exact CPU architecture in the build.

Switch gRPC's build to take advantage of this. I've retained BUILD_OVERRIDE_BORING_SSL_ASM_PLATFORM, on the off chance anyone is using it to cross-compile between Windows and non-Windows, though I doubt that works particularly well.

As part of this, restore assembly optimizations in a few places where they were seemingly disabled for issues relating to this:

- https://github.com/grpc/grpc/pull/31747 had to disable the assembly, because at the time assembly required the library be built differently for each architecture and then stitched back together. This should now work.

- tools/run_tests/run_tests.py disabled x86 assembly due to some issues with CMAKE_SYSTEM_PROCESSOR in a Docker image. This too should now be moot.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -417,7 +417,7 @@ if BUILD_WITH_BORING_SSL_ASM and not BUILD_WITH_SYSTEM_OPENSSL:
     # by the preprocessor. Provided the platform has a gas-compatible assembler
     # (i.e. not Windows), we can include the assembly files and let BoringSSL
     # decide which ones should and shouldn't be used for the build.
-    if "win32" not in boringssl_asm_platform:
+    if not boringssl_asm_platform.startswith("win"):
         asm_key = "crypto_asm"
     else:
         print(
--- a/setup.py
+++ b/setup.py
@@ -413,19 +413,12 @@ if BUILD_WITH_BORING_SSL_ASM and not BUILD_WITH_SYSTEM_OPENSSL:
         if BUILD_OVERRIDE_BORING_SSL_ASM_PLATFORM
         else util.get_platform()
     )
-    LINUX_X86_64 = "linux-x86_64"
-    LINUX_ARM = "linux-arm"
-    LINUX_AARCH64 = "linux-aarch64"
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -413,19 +413,12 @@ def _quote_build_define(argument):
...
-        asm_key = "crypto_linux_aarch64"
-    elif "mac" in boringssl_asm_platform and "x86_64" in boringssl_asm_platform:
-        asm_key = "crypto_apple_x86_64"
-    elif "mac" in boringssl_asm_platform and "arm64" in boringssl_asm_platform:
-        asm_key = "crypto_apple_aarch64"
+    # BoringSSL's gas-compatible assembly files are all internally conditioned
+    # by the preprocessor. Provided the platform has a gas-compatible assembler
+    # (i.e. not Windows), we can include the assembly files and let BoringSSL
+    # decide which ones should and shouldn't be used for the build.
+    if "win32" not in boringssl_asm_platform:
```
davidben, line 420: Looks like the problem is Python's docs are false (sigh) and it doesn't return `sys.platform` on Windows. Will try to fix.

davidben:
Okay, ready for another attempt through CI.

davidben:
Looks like CI is clear now!

drfloob:
CC @jtattermusch @veblush

gnossen (approved):
Really happy Apple silicon users will be able to take advantage of assembly optimizations now. Thanks!

jtattermusch (approved):
LGTM, thanks for the improvement!

PR #33702, opened 2023-07-14
[fix] Remove GRPC_MUST_USE_RESULT from class definition

drfloob:
Problem introduced in https://github.com/grpc/grpc/pull/31676. The PHP distribtests have been failing for 8 months with:

```
/tmp/pear/temp/grpc/src/core/ext/transport/chttp2/transport/flow_control.h:78:28: warning: 'warn_unused_result' attribute only applies to function types [-Wattributes]
 class GRPC_MUST_USE_RESULT FlowControlAction {
                            ^~~~~~~~~~~~~~~~~
```

This only fails on the following distribtest, which is not built every time. So the master build only fails sporadically.

```
  python_linux_x64_buster, labels ['distribtest', 'python', 'linux', 'x64', 'buster', 'presubmit']
```

### DIFF
--- a/src/core/ext/transport/chttp2/transport/flow_control.h
+++ b/src/core/ext/transport/chttp2/transport/flow_control.h
@@ -75,7 +75,7 @@ enum class StallEdge { kNoChange, kStalled, kUnstalled };
 // Encapsulates a collections of actions the transport needs to take with
 // regard to flow control. Each action comes with urgencies that tell the
 // transport how quickly the action must take place.
-class GRPC_MUST_USE_RESULT FlowControlAction {
+class FlowControlAction {
  public:
   enum class Urgency : uint8_t {
     // Nothing to be done.


ctiller:
I don't think we should remove this attribute.

Can we introduce a new macro that is compatible here?
Or remove the antique build configuration that's failing?

drfloob:
> I don't think we should remove this attribute.
>
> Can we introduce a new macro that is compatible here? Or remove the antique build configuration that's failing?

Maybe so. I don't think this usage is safe across compilers. gcc claims `warn_unused_result` is a function attribute only ([docs](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-warn_005funused_005fresult-function-attribute)), whereas clang loosely claims (best I can find) that it should work on the type as well ([docs](https://clang.llvm.org/docs/AttributeReference.html#nodiscard-warn-unused-result)). I'm not interested in maintaining another compatibility macro for this, keeping track as compiler version support changes and whatnot, especially given there are other well-defined patterns for using `warn_unused_result` across compilers.

This attribute should be removed _to fix the breakage that it introduced_, and we can move forward with some more appropriate fix. The most common alternative in this repo is to annotate the functions that return this type, which does appear to be defined consistently across the major compilers.

ctiller:
We need this for types, and I'd like to introduce more types that use it.

Introduced https://github.com/grpc/grpc/pull/33742 to do this portably moving forward. I expect it doesn't fix the older PHP compiler bug - for that specific compiler we should ensure that we `#define GRPC_MUST_USE_RESULT /* empty */`

drfloob:
Alright. Given that this approach is rejected, and I'm not a huge fan of the tech-debt-incurring alternative, I'll hand this off to you. Upgrading gcc on that test might be another way forward.

PR #33695, opened 2023-07-13
[Protobuf] Upgrade third_party/protobuf to 23.4

drfloob:
This was done manually due to a problem with `tools/distrib/python/make_grpcio_tools.py`. ~I fixed it in this PR (depends on cl/547979185), so there is a fair chance this upgrade will work normally for the next release.~ The fix may be problematic for upgrading protobuf on older release branches, so the improvement will be worked on separately. CC @jtattermusch 

This also updates the UPB dep to the latest commit on the 23.x branch.

PR #33711, opened 2023-07-14
[release] Teach make_grpcio_tools script to delete temporary files

drfloob:
This (or something like it) is necessary for the automated protobuf upgrade script, to ensure we don't commit a redundant third_party tree to the repo. Here is an example of the protobuf upgrade without this fix: https://github.com/grpc/grpc/pull/33625

One alternative could be adding the filepaths to `.gitignore`.

Spun off from https://github.com/grpc/grpc/pull/33695

### DIFF
--- a/tools/distrib/python/make_grpcio_tools.py
+++ b/tools/distrib/python/make_grpcio_tools.py
@@ -16,6 +16,7 @@
 
 from __future__ import print_function
 
+import argparse
 import errno
 import filecmp
 import glob
@@ -89,6 +90,8 @@ COPY_FILES_SOURCE_TARGET_PAIRS = [
     ("third_party/protobuf/src", "third_party/protobuf/src"),
     ("third_party/utf8_range", "third_party/utf8_range"),
 ]
+
+DELETE_TARGETS_ON_CLEANUP = ["third_party"]
 
 # grpc repo root
 GRPC_ROOT = os.path.abspath(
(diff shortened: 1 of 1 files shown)


gnossen:
WANT_LGTM=gnossen

gnossen (approved):

jtattermusch:
I'm ok with this if @gnossen is happy. Also, now this is a separate PR, we can always backport it to older release branches should we need to upgrade protobuf on them.

jtattermusch (approved):
I'd appreciate a few sentences that describe where the "temporary third_party" folder comes from FTR. I guess it has to do with some of the make_grpcio_tools.py hackery, but that one's rather cryptic so clarifying the situation would help IMHO.

Also, I'd love us to clean up the make_grpcio_tools.py script at some point, it has caused a lot of confusion over the years.

tools/distrib/python/make_grpcio_tools.py
```diff
@@ -228,7 +231,19 @@ def _copy_source_tree(source, target):
...
 
+def _delete_source_tree(target):
+    """Deletes the copied target directory."""
+    target = GRPCIO_TOOLS_ROOT_PREFIX + target
+    target_abs = os.path.join(*target.split("/"))
+    print("Deleting copied folder %s" % (target_abs))
+    shutil.rmtree(target_abs, ignore_errors=True)
+
 def main():
+    parser = argparse.ArgumentParser()
```
jtattermusch, line 242: actually, looks like that since we weren't parsing arguments of the script before (at all), on older version of make_grpcio_tools.py, the option `--cleanup-third-party` would be silently ignored?

Whether that's something we want or not I'm not sure, but I wanted to point this behavior out.

Also, can we simply make the "--cleanup-third-party" behavior the default (and not need a new cmdline flag at all)?

tools/distrib/python/make_grpcio_tools.py
```diff
@@ -228,7 +231,19 @@ def _copy_source_tree(source, target):
...
+    parser.add_argument(
+        "--cleanup-third-party", default=False, action="store_true", help="Delete the temporary third_party folder"
```
jtattermusch, line 244: nit: naming:  throughout the grpc repo ,we use underscores in flag names AFAICT. So `--cleanup_third_party` would be more consistent?

jtattermusch:
Added a few more comments.

PR #33721, opened 2023-07-14
[EventEngine] fix bug in unix-abstract socket URI processing

Vignesh2208:
URI should be resolved to unix abstract only if the first byte is NULL but the path length is > 1 (i.e there is more than one byte in the path). Otherwise treat is as unix URI.

Moreover, when the AsyncConnectionAcceptor is destroyed, if it was managing a unix domain socket, then the socket should be unlinked to ensure we dont leave the underlying file hanging around forever.

### DIFF
--- a/src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
+++ b/src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
@@ -153,11 +153,11 @@ void PosixEngineListenerImpl::AsyncConnectionAcceptor::NotifyOnAccept(
     }
 
     // For UNIX sockets, the accept call might not fill up the member
-    // sun_path of sockaddr_un, so explicitly call getsockname to get it.
+    // sun_path of sockaddr_un, so explicitly call getpeername to get it.
     if (addr.address()->sa_family == AF_UNIX) {
       socklen_t len = EventEngine::ResolvedAddress::MAX_SIZE_BYTES;
-      if (getsockname(fd, const_cast<sockaddr*>(addr.address()), &len) < 0) {
-        gpr_log(GPR_ERROR, "Closing acceptor. Failed getsockname: %s",
+      if (getpeername(fd, const_cast<sockaddr*>(addr.address()), &len) < 0) {
+        gpr_log(GPR_ERROR, "Closing acceptor. Failed getpeername: %s",
                 strerror(errno));
         close(fd);
         // Shutting down the acceptor. Unref the ref grabbed in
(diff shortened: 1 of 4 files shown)


src/core/lib/event_engine/posix_engine/posix_engine_listener.cc
```diff
@@ -153,11 +153,11 @@ void PosixEngineListenerImpl::AsyncConnectionAcceptor::NotifyOnAccept(
     }
 
     // For UNIX sockets, the accept call might not fill up the member
-    // sun_path of sockaddr_un, so explicitly call getsockname to get it.
+    // sun_path of sockaddr_un, so explicitly call getpeername to get it.
     if (addr.address()->sa_family == AF_UNIX) {
       socklen_t len = EventEngine::ResolvedAddress::MAX_SIZE_BYTES;
-      if (getsockname(fd, const_cast<sockaddr*>(addr.address()), &len) < 0) {
-        gpr_log(GPR_ERROR, "Closing acceptor. Failed getsockname: %s",
+      if (getpeername(fd, const_cast<sockaddr*>(addr.address()), &len) < 0) {
```
drfloob, line 159: This reverts the fix in https://github.com/grpc/grpc/pull/31108

drfloob: Ooh! It's the same fix. My mistake.

src/core/lib/event_engine/tcp_socket_utils.cc
```diff
@@ -123,9 +123,9 @@ absl::StatusOr<std::string> ResolvedAddrToUriUnixIfPossible(
   GRPC_RETURN_IF_ERROR(path.status());
   std::string scheme;
   std::string path_string;
-  if (path->at(0) == '\0') {
+  if (path->at(0) == '\0' && path->length() > 1) {
     scheme = "unix-abstract";
-    path_string = path->length() > 1 ? path->substr(1, std::string::npos) : "";
+    path_string = path->substr(1, std::string::npos);
   } else {
     scheme = "unix";
```
drfloob, line 130: Should `\0` be treated as a valid unix socket? We could return an error on that edge case.

drfloob (commented):
Nice catch!

src/core/lib/event_engine/tcp_socket_utils.cc
Vignesh2208: Fixed now

drfloob (approved):

PR #33744, opened 2023-07-17
[EventEngine] Change `GetDNSResolver` to return `absl::StatusOr<std::unique_ptr<DNSResolver>>`

yijiem:
Based on the discussion at: https://github.com/grpc/grpc/pull/32701/files/595a75cc5d56f8d7341a9533a24428ad37810269..e3b402a8fa27fc8787a73775068721bf9f322640#r1244325752


include/grpc/event_engine/event_engine.h
```diff
@@ -397,8 +397,9 @@ class EventEngine : public std::enable_shared_from_this<EventEngine> {
   virtual bool IsWorkerThread() = 0;
 
   /// Creates and returns an instance of a DNSResolver, optionally configured by
-  /// the \a options struct.
-  virtual std::unique_ptr<DNSResolver> GetDNSResolver(
+  /// the \a options struct. This method may return a non-OK status if an error
+  /// occurred when creating the DNSResolver.
+  virtual absl::StatusOr<std::unique_ptr<DNSResolver>> GetDNSResolver(
```
drfloob, line 402: Do we still need a unique_ptr here? The original reason behind having it return a unique_ptr was to provide some small signal if the resolver could not be created (i.e., it returns a nullptr). Now that it's a StatusOr, I think we could just do `virtual absl::StatusOr<DNSResolver> GetDNSResolver(...)`

markdroth: We can't return by value here, because `DNSResolver` is a base class, and we are going to be returning an implementation of a subclass, which will almost certainly have a different storage size.

markdroth (approved):

include/grpc/event_engine/event_engine.h
drfloob: Ah, that's right. I need more coffee.

drfloob (approved):

include/grpc/event_engine/event_engine.h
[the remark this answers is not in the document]
yijiem: Done.

PR #32941, opened 2023-04-26
[Python] Support musl-linux binary wheels on arm64

merlimat:
Add Musl based build for arm64

PR #33760, opened 2023-07-19
[dockerfiles] Improvements and fixes to push_testing_images.sh script

jtattermusch:
- add check that there are no unwanted .current_version files left in the repo after foobar/Dockerfile gets deleted.
- revisit the logic for repo digest vs id digest of docker images.

### DIFF
--- a/tools/dockerfile/push_testing_images.sh
+++ b/tools/dockerfile/push_testing_images.sh
@@ -147,11 +147,23 @@ do
   # if the .current_version file doesn't exist or it doesn't contain the right SHA checksum,
   # it is out of date and we will need to rebuild the docker image locally.
   LOCAL_BUILD_REQUIRED=""
-  grep "^${ARTIFACT_REGISTRY_PREFIX}/${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}@sha256:.*" ${DOCKERFILE_DIR}.current_version >/dev/null || LOCAL_BUILD_REQUIRED=true
+  grep "^${ARTIFACT_REGISTRY_PREFIX}/${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}@sha256:.*$" ${DOCKERFILE_DIR}.current_version >/dev/null || LOCAL_BUILD_REQUIRED=true
+
+  # If the current version file has contains SHA checksum, but not the remote image digest,
+  # it means the locally-built image hasn't been pushed to artifact registry yet.
+  DIGEST_MISSING_IN_CURRENT_VERSION_FILE=""
+  grep "^${ARTIFACT_REGISTRY_PREFIX}/${DOCKER_IMAGE_NAME}:${DOCKER_IMAGE_TAG}$" ${DOCKERFILE_DIR}.current_version >/dev/null && DIGEST_MISSING_IN_CURRENT_VERSION_FILE=true
 
   if [ "${LOCAL_BUILD_REQUIRED}" == "" ]
   then
     echo "Dockerfile for ${DOCKER_IMAGE_NAME} hasn't changed. Will skip 'docker build'."
+    continue
+  fi
+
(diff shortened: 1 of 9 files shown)


tools/dockerfile/push_testing_images.sh
```diff
@@ -67,6 +67,22 @@ ALL_DOCKERFILE_DIRS=(
 
 CHECK_FAILED=""
 
+if [ "${CHECK_MODE}" != "" ]
+then
+  # Check that there are no stale .current_version files (for which the corresponding
+  # dockerfile_dir doesn't exist anymore).
+  for CURRENTVERSION_FILE in $(find tools/ third_party/rake-compiler-dock -name '*.current_version')
+  do
+    DOCKERFILE_DIR="$(echo ${CURRENTVERSION_FILE} | sed 's/.current_version$//')"
```
drfloob, line 76: I think you need to escape the `.`, or it'll match any char. `sed 's/\.current_version$//'`

tools/dockerfile/push_testing_images.sh
```diff
@@ -67,6 +67,22 @@ ALL_DOCKERFILE_DIRS=(
...
+    if [ ! -e "${DOCKERFILE_DIR}/Dockerfile" ]
+    then
+       echo "Found that ${DOCKERFILE_DIR} has '.current_version' file but there is no corresponding Dockerfile."
+       echo "Should the ${CURRENTVERSION_FILE} file be deleted?"
+       CHECK_FAILED=true
```
drfloob, line 81: Why not `read` a prompt and `rm` here? You could also collect the set of unmatched current_version files and do a single read/rm prompt afterwards, I'm not sure when I'd want to keep them.

drfloob (commented):
Looks pretty good, just a few comment left inline.

To future-proof this a bit, I suggest adding a check to `tools/dockerfile/push_testing_images.sh` that ensures we don't commit any current_version files that are missing a digest. If `docker image inspect ...` fails for example, the version won't be updated, but the image will have been pushed anyway.

tools/dockerfile/push_testing_images.sh
drfloob: Actually, I'm not sure how you'd detect if it's running interactively. You can disregard this idea.

tools/dockerfile/push_testing_images.sh
jtattermusch: Done.

jtattermusch:
> Looks pretty good, just a few comment left inline.
> 
> To future-proof this a bit, I suggest adding a check to `tools/dockerfile/push_testing_images.sh` that ensures we don't commit any current_version files that are missing a digest. If `docker image inspect ...` fails for example, the version won't be updated, but the image will have been pushed anyway.

Good catch. I originally thought this is ok, assuming a sanity check would prevent you from submitting such a PR.
But then I realized that the sanity check actually runs in as LOCAL_MODE=true, and that wouldn't reveal the issue since it wouldn't query the artifact registry. So I'll have to add an extra check for this.

https://github.com/grpc/grpc/blob/19414da96d6235226c41063d607dfef2c7210df1/tools/run_tests/sanity/check_testing_docker_images.sh#L23

jtattermusch:
@drfloob @gnossen  PTAL, I added an explicit check for locally-built images that weren't pushed (see https://github.com/grpc/grpc/pull/33760#issuecomment-1645698061 for context)

drfloob (approved):
Looks good. In the future, please try not to force-push the branch when I am a reviewer, it makes reviews take longer. Given that PRs are squashed into a single commit when merged, even a string of non-atomic "fix" commits would be better, since it allows reviewers to compare changes that happened since their last review.

gnossen:
@jtattermusch 

FYI, this appears not be working as intended:
- [PR](https://github.com/grpc/grpc/pull/32941)
- [Failing test with unknown docker image hash](https://source.cloud.google.com/results/invocations/2dbbd1f4-1ee3-43a3-a0a6-3c89fae9fd5e/log)

PR #33770, opened 2023-07-19
[Backport] [dependency] Restrict cython to less than 3.X

gnossen:
*Beep boop. This is an automatically generated backport of #33738.*

This should resolve breakage on master caused by the jump to Cython 3.0.0 this morning.


apolcyn (approved):

gnossen:
- EventEngine Client Tests C++ iOS - Logs unvailable but test failing on all branches, so assuming this is a benign failure.
- Distribution Tests Python Windows - Looks like a race conditoin: `ERROR: Could not install packages due to an EnvironmentError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Python37\\Lib\\site-packages\\setuptools\\extension.py'`
- Distribution Tests Python MacOS -  PyPi flake:  `WARNING: Skipping page https://pypi.org/simple/cython/ because the GET request got Content-Type: Unknown. The only supported Content-Types are application/vnd.pypi.simple.v1+json, application/vnd.pypi.simple.v1+html, and text/html`
- Bazel RBE Fuzztest C/C++  - Logs unavailable but failing on all branches
- Bazel RBE Debug C/C++ - This change should not affect this test
- Basic Tests Ruby MacOS  - This change should not affect this test
- Basic Tests Ruby Linux - this change should not affect this test
- Basic Tests ObjC Examples  - This change should not affect this test.

Issue #33787, opened 2023-07-20
[Ruby] v1.57.0-alpha SEGV when calling `GRPC.postfork_child`


casperisfine:
### What version of gRPC and what language are you using?

Ruby 3.2.2
grpc 1.57.0-alpha (exact ref is https://github.com/Shopify/grpc/tree/v1.57.0-alpha2-keep-symbol to have symbols for debugging)

### What operating system (Linux, Windows,...) and version?

Linux / Ubuntu 22.04

### What runtime / compiler are you using (e.g. python version or version of gcc)

gcc.

### What did you do?

I'll try to reduce this to a simpler repro, but our application is huge so it may take a while to isolate what may trigger this, so I'm reporting what I see first in case the bug might be evident with the backtrace.

To make sure I never forget to call the hooks, I use the Ruby 3.1+ fork hook interface:

```ruby
module GRPCForkHook
  def _fork
    PanamaClient.before_fork
    if (pid = super)
      PanamaClient.after_fork_parent
    else
      PanamaClient.after_fork_child
    end
    pid
  end
end

Process.singleton_class.prepend(GRPCForkHook)
```

Here `PanamaClient` calls are thing wrapper around GRPC to do some extra things before or after the hooks:

```ruby
        def before_fork
          INSTANCES.each_value(&:before_fork)
          GRPC.prefork if GRPC_FORK_SUPPORT
        end

        def after_fork_child
          GRPC.postfork_child if GRPC_FORK_SUPPORT
          INSTANCES.each_value(&:after_fork)
        end

        def after_fork_parent
          GRPC.postfork_parent if GRPC_FORK_SUPPORT
          INSTANCES.each_value(&:after_fork)
        end
```

### What did you expect to see?

I'd expect the program to continue

### What did you see instead?

```
/tmp/bundle/ruby/3.2.0/gems/panama_client-2.4.0/lib/panama_client/client/fork_safety.rb:24: [BUG] Segmentation fault at 0x0000000000000038
ruby 3.2.2 (2023-07-19 revision 36486ba854) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0016 p:---- s:0083 e:000082 CFUNC  :postfork_child
c:0015 p:0008 s:0079 e:000078 METHOD /tmp/bundle/ruby/3.2.0/gems/panama_client-2.4.0/lib/panama_client/client/fork_safety.rb:24
[...]

PR #33788, opened 2023-07-20
[ruby] fix crash when prefork/postfork is used without previously using grpc

apolcyn:
Should fix https://github.com/grpc/grpc/issues/33787

### DIFF
--- a/src/ruby/ext/grpc/rb_grpc.c
+++ b/src/ruby/ext/grpc/rb_grpc.c
@@ -362,9 +362,12 @@ void grpc_ruby_init() {
 // into the gRPC library during or after prefork has been called, until
 // the corresponding postfork_{parent,child} APIs have been called.
 static VALUE grpc_rb_prefork(VALUE self) {
-  gpr_once_init(
-      &g_once_init,
-      grpc_ruby_basic_init);  // maybe be the first time called into gRPC
+  // This might be the first time we've called into the grpc library, so make
+  // sure basic one-time initialization is taken care of. Note that if this is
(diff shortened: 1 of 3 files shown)


src/ruby/ext/grpc/rb_grpc.c
```diff
@@ -362,9 +362,12 @@ void grpc_ruby_init() {
...
 // the corresponding postfork_{parent,child} APIs have been called.
 static VALUE grpc_rb_prefork(VALUE self) {
-  gpr_once_init(
-      &g_once_init,
-      grpc_ruby_basic_init);  // maybe be the first time called into gRPC
+  // This might be the first time we've called into the grpc library, so make
+  // sure basic one-time initialization is taken care of. Note that if this is
+  // the case, then grpc_init() will start up c-core threads; that's OK since
+  // they will be shut down in C-core's pthread_atfork handler.
+  gpr_once_init(&g_once_init, grpc_ruby_basic_init);
```
casperisfine, line 369: I had a different solution that didn't involve initializing the threads: https://github.com/Shopify/grpc/commit/627707501374c88ab665214192a97ef45763b105

But I guess this is OK.

drfloob (approved):

src/ruby/ext/grpc/rb_grpc.c
apolcyn: Your fix also looks OK, but I'll favor the fix here for simplicity - I want to minimize the fork related state we track for maintainability.

casperisfine: No problem.

PR #33791, opened 2023-07-20
[ruby] improve possible error message in postfork parent

apolcyn:
The case of `!grpc_ruby_initial_pid()` can be a *cause* of the second case, `!grpc_ruby_initial_thread`, so check the pid first.


sampajano (approved):
lgtm thanks Alex!

PR #33793, opened 2023-07-20
[xDS] check logical_dns hostname target validity at CDS validation time

markdroth:
Also change DNS resolver factories to reject paths that contain slashes.


src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
```diff
@@ -347,10 +347,15 @@ class AresClientChannelDNSResolverFactory : public ResolverFactory {
   absl::string_view scheme() const override { return "dns"; }
 
   bool IsValidUri(const URI& uri) const override {
-    if (absl::StripPrefix(uri.path(), "/").empty()) {
+    absl::string_view path = absl::StripPrefix(uri.path(), "/");
```
eugeneo, line 350: Q: Is there some sort of "utility" package we could put code like this? 3x duplication pretty much guarantees the code will diverge. This code only depends on STL and ABSL so can go anywhere...

eugeneo (approved):
No comments, just a single question.

src/core/ext/filters/client_channel/resolver/dns/c_ares/dns_resolver_ares.cc
markdroth: When the EventEngine migration finishes, two of the three copies will go away, so I'm not too worried about it.

PR #33803, opened 2023-07-21
[build metadata] Bazel to "other build systems" improvements

jtattermusch:
- Extract build metadata for some external dependencies from bazel build. This is achieved by letting extract_metadata_from_bazel_xml.py analyze some external libraries and sources. The logic is basically the same as for internal libraries, I only needed to teach extract_metadata_from_bazel_xml.py which external libraries it is allowed to analyze.
     * currently, the list of source files is automatically determined for `z`, `upb`, `re2` and `gtest` dependencies (at least for the case where we're building in "embedded" mode - e.g. mostly native extensions for python, php, ruby etc. - cmake has the ability to replace some of these dependencies by actual cmake dependency.)
- Eliminate the need for manually written gen_build_yaml.py for some dependencies. 
- Make the info on target dependencies in build_autogenerated.yaml more accurate and complete. Until now, there were some depdendencies that were allowed to show up in  build_autogenerated.yaml and some that were being skipped. This made generating the CMakeLists.txt and Makefile quite confusing (since some dependencies are being explicitly mentioned and some had to be assumed by the build system).
- Overhaul the Makefile
     * the Makefile is currently only used internally (e.g. for ruby and PHP builds)
     * until now, the makefile wasn't really using the info about which targets depend on what libraries, but it was effectively hardcoding the depedendency data (by magically "knowing" what is the list of all the stuff that e.g. "grpc" depends on).
     * After the overhaul, the Makefile.template now actually looks at the library dependencies and uses them when generating the makefile. This gives a more correct and easier to maintain makefile.
     * since csharp is no longer on the master branch, remove all mentions of "csharp" targets in the Makefile.

Other notable changes:
- make extract_metadata_from_bazel_xml.py capable of resolving workspace bind() rules (so that it knows the real name of the target that is referred to as e.g. `//external:xyz`)

TODO:
- [DONE] ~~pkgconfig C++ distribtest~~
- [DONE} ~~update third_party/README to reflect changes in how some deps get updated now.~~

Planned followups:
- cleanup naming of some targets in build metadata and buildsystem templates: libssl vs boringssl, ares vs cares etc.
- further cleanup of Makefile
- further cleanup of CMakeLists.txt
- remote the need from manually hardcoding extra metadata for targets in build_autogenerated.yaml. Either add logic that determines the properties of targets automatically, or use metadata from bazel BUILD.


jtattermusch:
FTR: https://source.cloud.google.com/results/invocations/2ef86b7a-66bd-4555-8d07-48f40ea4cdd9/targets;collapsed=/github%2Fgrpc%2Fbuild_artifacts/tests;passed=true;group=tasks;test=build_artifact.ruby_native_gem_linux_x86_64-darwin;row=7

https://github.com/tpoechtrager/osxcross/issues/222
https://github.com/tpoechtrager/cctools-port/commit/fe91a8e6a2e67af37359d5bcef15aa827d0627c5#diff-29672385ad34e6caa3dda5151d05ef6c4baebdb79e3ba1d2a6e6a2bf53b193f1R3097

jtattermusch:
@drfloob @gnossen @veblush  this is the first step in (hopefully many) improvements to our bazel -> "other build systems" conversion logic.  There's still a few things to address (see TODOs), but overall this is ready for reviews.

I realize that the logic in extract_metadata_from_bazel.xml is currently complex (so changes to it seem esoteric), but I hope to simplify it in the future. Some simplifications first require overhaul of some other stuff (e.g. the Makefile.template) and it would be wrong to make too many changes at once. IMHO this PR has about the right number of changes in it, but if needed, I can try to split it.

gnossen (approved):

drfloob (approved):

veblush (approved):

Issue #33802, opened 2023-07-21
[Ruby] v1.57.0-alpha race condition in `GRPC.postfork_parent` leading to an assertion crash


casperisfine:
### What version of gRPC and what language are you using?

Ruby 3.2.2
grpc 1.57.0-alpha (exact ref is https://github.com/Shopify/grpc/tree/v1.57.0-alpha2-keep-symbol to have symbols for debugging)

### What operating system (Linux, Windows,...) and version?

Linux / Ubuntu 22.04

### What runtime / compiler are you using (e.g. python version or version of gcc)

gcc.

### What did you do?

I managed to craft a minimal repro for it:

```ruby
ENV["GRPC_ENABLE_FORK_SUPPORT"] = "1"
require 'grpc'
10_000.times do
  GRPC.prefork
  GRPC.postfork_parent
end
```

### What did you expect to see?

I'd expect the program to work fine

### What did you see instead?

```
E0721 10:34:23.323129764      91 rb_channel.c:681]                     ASSERTION FAILED: !g_abort_channel_polling
shopify-build/bin/entrypoint: line 181:    86 Aborted                 (core dumped) bundle exec ruby test/repro.rb
```

### Early investigation

`rb_channel.c:681` is `run_poll_channels_loop_no_gil`, that function is called from `run_poll_channels_loop` which itself is started as:

```c
g_channel_polling_thread = rb_thread_create(run_poll_channels_loop, NULL);
```

So what I think is happening is:

  - `GRPC.prefork` calls `grpc_rb_channel_polling_thread_stop()`, which sets `g_abort_channel_polling = 1` and then back to `false` once the threads are actually shutdown.
  - `GRPC.postfork_parent` calls `grpc_ruby_init_threads();` -> `grpc_rb_channel_polling_thread_start()` which spawns a thread calling `run_poll_channels_loop`
  - Then, before `run_poll_channels_loop` has reached `GPR_ASSERT(!g_abort_channel_polling)` on `rb_channel.c:681`, `GRPC.prefork` is called again, and sets `g_abort_channel_polling = 1`, causing the race condition.

So I suppose we're missing some synchronization here, we shouldn't be able to enter `grpc_rb_channel_polling_thread_stop` if the threads spawned by `grpc_ruby_init_threads()` aren't ready.

I'll see if I can work on a patch, but I think that bug is a bit too subtle for me to fix in a timely manner.

PR #33804, opened 2023-07-21
[Ruby] grpc_ruby_init_threads wait for polling thread to be started before returning

casperisfine:
Fix: https://github.com/grpc/grpc/issues/33802

Otherwise if `GRPC.prefork` is called too soon after `GRPC.postfork_parent` `GPR_ASSERT(!g_abort_channel_polling)` may be executed while the main thread just have set `g_abort_channel_polling = 1` causing a crash.

cc @apolcyn

### DIFF
--- a/src/ruby/end2end/prefork_loop_test.rb
+++ b/src/ruby/end2end/prefork_loop_test.rb
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+#
+# Copyright 2016 gRPC authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ENV['GRPC_ENABLE_FORK_SUPPORT'] = "1"
+fail "forking only supported on linux" unless RUBY_PLATFORM =~ /linux/
+
+this_dir = File.expand_path(File.dirname(__FILE__))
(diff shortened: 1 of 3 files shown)


apolcyn:
Thanks for finding this! The fix makes sense, but there is a bit of cleanup I think we should do if we take this approach.

Main thing is, if we are adding "wait for thread startup" logic to `grpc_rb_channel_polling_thread_start`, then we should be able to remove it from everywhere else:

- https://github.com/grpc/grpc/blob/19414da96d6235226c41063d607dfef2c7210df1/src/ruby/ext/grpc/rb_channel.c#L229
- https://github.com/grpc/grpc/blob/19414da96d6235226c41063d607dfef2c7210df1/src/ruby/ext/grpc/rb_channel.c#L431

I *think* we can just remove those two calls to `wait_until_channel_polling_thread_started_no_gil` after putting one in `grpc_rb_channel_polling_thread_start` - I'm forgetting why we didn't do this earlier, though (perhaps we should have).

-------

Another approach might be to simply remove `g_channel_polling_thread_started` and related logic to wait for that thread to start up. And then change code in `https://github.com/grpc/grpc/blob/19414da96d6235226c41063d607dfef2c7210df1/src/ruby/ext/grpc/rb_channel.c#L681C15-L681C38` to something like:

```
if (g_abort_channel_polling) {
  gpr_mu_unlock(&global_connection_polling_mu);
  gpr_log(GPR_DEBUG, "GRPC_RUBY: run_poll_channels_loop_no_gil - early exit");
  return;
}
```

... and remove the assertion.

That might be a nice simplification, and I'm failing to see why we can't do that.

apolcyn:
Actually, further on my last comment, we shouldn't even need that early exit code.

We should be able to rely on https://github.com/grpc/grpc/blob/19414da96d6235226c41063d607dfef2c7210df1/src/ruby/ext/grpc/rb_channel.c#L690 happening, because the unblocking func will [do this](https://github.com/grpc/grpc/blob/19414da96d6235226c41063d607dfef2c7210df1/src/ruby/ext/grpc/rb_channel.c#L745).

In that case I think we can just remove the assertion on `g_abort_channel_polling`.

PR #33814, opened 2023-07-21
[docker] Use the remote server as authoritative for image shas

gnossen:
In general, remote docker images shas are _not_ the same as local docker image shas. Since https://github.com/grpc/grpc/pull/31591, I have faced "non-existent" image issues in CI after every run of this script, which I have had to fix by manually updating the shas in the `current_version` file. This change has had 100% success for me.


jtattermusch:
https://github.com/grpc/grpc/pull/33760 seems to be a better version of the same thing?

PR #33820, opened 2023-07-22
[python] use f-string for string formatting

antonkesy:
Are there any reasons not to use new _new_ f-string or ```.format()``` for string formatting?

``` print(f"Unknown event {event}.\n") ```
is much more readable and less error-prone than 
``` print("Unknown event {}.\n".format(event)) ```

Those changes can be easily automated using [flynt](https://github.com/ikamensh/flynt) or [pyupgrade](https://github.com/asottile/pyupgrade)

_DISCLAIMER: The current changes only cover the basic cases ..._

Any suggestions? Adding the missing cases or not changing it at all?


gnossen:
Thank you for the interest!

Here's my take on f-strings versus other forms of string interpolation, such as `.format()` and the `%` operator. You might get a slight bump in readability, but other than that it really seems to come down to personal preference. If we had a way to guarantee that there were no issues with a migration from other string interpolations to f-strings, I would be all for it, but very often, we'll run into corner cases where a log happens and only in that very exceptional case do we run into a simple mistake/typo with the interpolation. If f-strings provided some sort of "compile-time guarantee" as in other languages, that would be a distinct advantage of f-strings, but we don't have that.

I'm fairly confident that the _implementation_ of the `grpcio` package has good enough test coverage that a large-scale switch to f-strings would be just fine. For the rest of the Python code in the repo, however, I would take some more convincing.

gnossen:
It looks like there are functional errors with this PR even in the portions that are well tested. Since there are large portions of the tooling that are not themselves tested, I do not think we can justify the risk of migrating all string interpolation to f-strings until there's a much clearer benefit to it.

PR #33823, opened 2023-07-24
[Test] Update benchmarks job configuration to match experiment.

paulosjca:
This change updates the kokoro job configuration to match the changes in https://github.com/grpc/grpc/pull/33238.

### DIFF
--- a/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
+++ b/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
@@ -19,6 +19,20 @@ cd "$(dirname "$0")/../../.."
 
 source tools/internal_ci/helper_scripts/prepare_build_linux_rc
 
+
+# Environment variables to select repos and branches for various repos.
+# You can edit these lines if you want to run from a fork.
+GRPC_CORE_REPO=grpc/grpc
+GRPC_CORE_GITREF=master
+GRPC_DOTNET_REPO=grpc/grpc-dotnet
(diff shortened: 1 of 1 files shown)


tools/internal_ci/linux/grpc_e2e_performance_gke.sh
```diff
@@ -19,6 +19,20 @@ cd "$(dirname "$0")/../../.."
 
 source tools/internal_ci/helper_scripts/prepare_build_linux_rc
 
+
+# Environment variables to select repos and branches for various repos.
+# You can edit these lines if you want to run from a fork.
+GRPC_CORE_REPO=grpc/grpc
```
drfloob, line 25: If you want devs to be able to customize this script more easily without having to edit it, you can let the shell's environment variables override the script variables with something like `GRPC_CORE_REPO="${GRPC_CORE_REPO:-grpc/grpc}"`

Then you can invoke it using `GRPC_CORE_REPO=drfloob/grpc grpc_e2e_performance_gke.sh`

drfloob (approved):

tools/internal_ci/linux/grpc_e2e_performance_gke.sh
paulosjca: Thank you. That requires other changes too, because the script right now assumes it is invoked by kokoro, not from the command line (it looks for variables like `KOKORO_GITHUB_COMMIT_URL` and others).

PR #33827, opened 2023-07-24
[import] Fix missing dependency in experiments_tag_test

Vignesh2208:


drfloob (approved):

test/core/experiments/BUILD
```diff
@@ -53,6 +53,7 @@ grpc_cc_test(
         "gtest",
         "absl/strings",
         "absl/status",
+        "absl/status:statusor",
```
drfloob, line 56: I suggest adding "test/core/experiments" to this list, to enable build dependency checking https://github.com/grpc/grpc/blob/457fd65b93cd0c2932c24b0b0f2699e854b68fa7/tools/distrib/fix_build_deps.py#L378-L394

Vignesh2208: Done.

PR #33869, opened 2023-07-25
[PSM Interop] Legacy tests: delete PHP and Ruby

sergiitk:
ref b/257007535, cl/550946605


ejona86 (commented):
Hmmm... I'm surprised to see Ruby and PHP listed at https://github.com/grpc/grpc/blob/master/doc/grpc_xds_features.md . Seems something needs to be fixed (the doc or the tests), but I'm unsure which.

ejona86 (approved):

PR #33870, opened 2023-07-25
[core/gpr] move subprocess to gpr

mario-vimal:
Move subprocess util to gpr.
Add support for communication with the subprocess.
This is required to support authentication using an executable.


ctiller:
If possible it would be better to make this its own bazel build target (`name = 'subprocess'`)

ctiller (approved):

PR #33876, opened 2023-07-26
[chaotic-good] Client transport write path

nanahpang:
This is the initial implementation of the chaotic-good client transport write path. There will be a follow-up PR to fulfill the read path.


test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -0,0 +1,179 @@
...
+  MockEndpoint& control_endpoint_;
+  MockEndpoint& data_endpoint_;
+  std::shared_ptr<grpc_event_engine::experimental::FuzzingEventEngine>
+      event_engine_;
+  ClientTransport client_transport_;
+  ScopedArenaPtr arena_;
+  Pipe<MessageHandle> pipe_client_to_server_messages_;
+};
+
+TEST_F(ClientTransportTest, AddOneStream) {
```
Vignesh2208, line 131: Is it possible to generalize this test to add N streams and send K messages per stream ?

Vignesh2208 (commented):
Thanks for working on this Nana! Looks like a great start! Left some suggestions for improvement

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -0,0 +1,121 @@
...
+      control_endpoint_(std::move(control_endpoint)),
+      data_endpoint_(std::move(data_endpoint)),
+      event_engine_(event_engine) {
+  auto hpack_compressor = std::make_shared<HPackCompressor>();
+  auto write_loop = Loop(Seq(
+      // Get next outgoing frame.
+      this->outgoing_frames_.Next(),
+      // Construct data buffers that will be sent to the endpoints.
+      [hpack_compressor](ClientFrame client_frame) {
+        SliceBuffer control_endpoint_buffer;
```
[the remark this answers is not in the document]
nanahpang: Sounds good, I will change them to be shared buffer in ClientTransport then.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -0,0 +1,121 @@
...
+                    frame->message->payload()->Length(), data_endpoint_buffer);
+              }
+            },
+            [hpack_compressor,
+             &control_endpoint_buffer](CancelFrame* frame) mutable {
+              control_endpoint_buffer.Append(
+                  frame->Serialize(hpack_compressor.get()));
+            });
+        return std::make_tuple<SliceBuffer, SliceBuffer>(
+            std::move(control_endpoint_buffer),
```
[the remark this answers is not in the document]
nanahpang: Yes, thanks for catching this!

src/core/ext/transport/chaotic_good/client_transport.h
```diff
@@ -0,0 +1,114 @@
...
+    uint64_t stream_id;
+    {
+      MutexLock lock(&mu_);
+      stream_id = next_stream_id_++;
+    }
+    bool initial_frame = true;
+    return Seq(
+        // Continuously send data frame with client to server messages.
+        ForEach(std::move(*call_args.client_to_server_messages),
+                [stream_id, &initial_frame,
```
ctiller, line 70: initial_frame = false here -- and remove the variable on the stack

nanahpang: Done, thanks for catching this.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -0,0 +1,121 @@
...
+        return TryJoin(
+            this->control_endpoint_->Write(std::move(std::get<0>(ret))),
+            this->data_endpoint_->Write(std::move(std::get<1>(ret))));
+      },
+      // Finish writes and return status.
+      [](absl::StatusOr<std::tuple<Empty, Empty>> ret)
+          -> LoopCtl<absl::Status> {
+        // If writes failed, return failure status.
+        if (!ret.ok()) {
+          return ret.status();
```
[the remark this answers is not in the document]
nanahpang: Yes, as discussed with Craig offline, the writer_ activity should close the transport when there's error return in line 116. And the AddStream promise will fail when the read path is added. So the workflow will be: AddStream(send frames to writer) -> writer activity fails and closes mpsc_pipe/transport -> AddStream(receive frames from reader) fails -> return error. 

I added a TODO comment for now, and will add the close transport logics in the following read path PR.

src/core/ext/transport/chaotic_good/client_transport.h
```diff
@@ -0,0 +1,113 @@
...
+          frame.stream_id = stream_id;
+          frame.headers = std::move(client_initial_metadata);
+          frame.end_of_stream = false;
+          return outgoing_frames.Send(ClientFrame(std::move(frame)));
+        },
+        // Continuously send data frame with client to server messages.
+        ForEach(std::move(*call_args.client_to_server_messages),
+                [stream_id, outgoing_frames = outgoing_frames_.MakeSender()](
+                    MessageHandle result) mutable {
+                  ClientFragmentFrame frame;
```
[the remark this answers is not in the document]
nanahpang: As we discussed offline, the CancelFrame will be sent in a different API that is called by the upper caller promise. This API is not yet implemented, so I will wait until the caller promise is done before implementing CancelFrame in the client transport.

test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -0,0 +1,179 @@
...
+  SliceBuffer buffer;
+  buffer.Append(Slice::FromCopiedString("test add stream."));
+  auto message = arena_->MakePooled<Message>(std::move(buffer), 0);
+  ClientMetadataHandle md;
+  auto args = CallArgs{
+      std::move(md), ClientInitialMetadataOutstandingToken::Empty(), nullptr,
+      nullptr,       &pipe_client_to_server_messages_.receiver,      nullptr};
+  StrictMock<MockFunction<void(absl::Status)>> on_done;
+  EXPECT_CALL(on_done, Call(absl::OkStatus()));
+  EXPECT_CALL(control_endpoint_, Write).WillOnce(Return(true));
```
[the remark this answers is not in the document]
nanahpang: Sure, I've done that. The AddStream promise currently does not expect to fail when EE write fails, because the error relies on the read path mentioned above. I have added a TODO comment for now to track this issue. Please let me know if this makes sense.

test/core/transport/chaotic_good/client_transport_test.cc
nanahpang: Yes, I've added the tests. Please let me know if there's other test coverage needed.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -0,0 +1,120 @@
...
+    std::unique_ptr<PromiseEndpoint> control_endpoint,
+    std::unique_ptr<PromiseEndpoint> data_endpoint,
+    std::shared_ptr<grpc_event_engine::experimental::EventEngine> event_engine)
+    : outgoing_frames_(MpscReceiver<ClientFrame>(4)),
+      control_endpoint_(std::move(control_endpoint)),
+      data_endpoint_(std::move(data_endpoint)),
+      control_endpoint_write_buffer_(SliceBuffer()),
+      data_endpoint_write_buffer_(SliceBuffer()),
+      event_engine_(event_engine) {
+  auto hpack_compressor = std::make_shared<HPackCompressor>();
```
Vignesh2208, line 58: Lets make hpack_compressor also as a member of the ClientTransport class ? Is there a reason not to do so ?

test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -0,0 +1,390 @@
...
+  StrictMock<MockFunction<void(absl::Status)>> on_done;
+  EXPECT_CALL(on_done, Call(absl::OkStatus()));
+  EXPECT_CALL(control_endpoint_, Write).Times(3).WillRepeatedly(Return(true));
+  EXPECT_CALL(data_endpoint_, Write).Times(3).WillRepeatedly(Return(true));
+  auto activity = MakeActivity(
+      Seq(
+          // Concurrently: send messages into the pipe, and receive from the
+          // pipe.
+          Join(Seq(pipe_client_to_server_messages_.sender.Push(
+                       std::move(message)),
```
Vignesh2208, line 240: Could this cause a use-after free error ? You seem to be std::move the same message 3 times. Maybe you meant to allocate a different message each time

test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -0,0 +1,390 @@
...
+  EXPECT_CALL(control_endpoint_, Write).Times(2).WillRepeatedly(Return(true));
+  EXPECT_CALL(data_endpoint_, Write).Times(2).WillRepeatedly(Return(true));
+  auto activity = MakeActivity(
+      Seq(
+          // Concurrently: send messages into the pipe, and receive from the
+          // pipe.
+          Join(
+              // Send message to first stream pipe.
+              Seq(pipe_client_to_server_messages_.sender.Push(
+                      std::move(message)),
```
Vignesh2208, line 285: Same comment as above. I think a new message should be allocated for each std::move

test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -0,0 +1,390 @@
...
+                  pipe_client_to_server_messages_.sender.Push(
+                      std::move(message)),
+                  pipe_client_to_server_messages_.sender.Push(
+                      std::move(message)),
+                  [this] {
+                    pipe_client_to_server_messages_.sender.Close();
+                    return absl::OkStatus();
+                  }),
+              // Send messages to second stream pipe.
+              Seq(pipe_client_to_server_messages_second_.sender.Push(
```
Vignesh2208, line 349: Same comment as above

Vignesh2208 (commented):
Left a few more comments

src/core/ext/transport/chaotic_good/client_transport.cc
nanahpang: Not really, I think we were using it as a local variable in Constructor because there's no other place need it. But moving to the class member should good, I've updated the code.

test/core/transport/chaotic_good/client_transport_test.cc
nanahpang: Done, thanks for catching this.

test/core/transport/chaotic_good/client_transport_test.cc
nanahpang: Done, thanks for catching this.

test/core/transport/chaotic_good/client_transport_test.cc
nanahpang: I didn't see the use-after free error, but I think you are right about moving the same message 3 times will cause problem. The latter 2 moves will point to nullptr after the first move. I have changed it to allocate different messages. 

There's an issue in test build which cause the checks are failed after sync to the master branch. I am debugging the root cause.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -0,0 +1,124 @@
...
+              &client_frame,
+              [this](ClientFragmentFrame* frame) mutable {
+                control_endpoint_write_buffer_.Append(
+                    frame->Serialize(hpack_compressor_.get()));
+                if (frame->message != nullptr) {
+                  char* header_string = grpc_slice_to_c_string(
+                      control_endpoint_write_buffer_.c_slice_buffer()
+                          ->slices[0]);
+                  auto frame_header =
+                      FrameHeader::Parse(
```
Vignesh2208, line 74: I think you can avoid the need to create and free header_string by changing this line to

```
auto frame_header = FrameHeader::Parse(
  reinterpret_cast<const uint8_t*(
    GRPC_SLICE_START_PTR(
       control_endpoint_write_buffer_.c_slice_buffer()
                          ->slices[0]));
```

Vignesh2208 (commented):
Thanks. Its mostly looking good. Just left one more comment

src/core/ext/transport/chaotic_good/client_transport.cc
nanahpang: Great, done, thanks for the suggestion.

Vignesh2208 (approved):

ctiller (approved):

Issue #20729, opened 2019-10-22
Support nullable type references in Grpc.Tools


JamesNK:
C# 8 adds nullable type references - https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

Grpc.Tools should be updated to support adding nullable type metadata to generated messages and services. It could be added as an option that is passed to the protoc compiler. MSBuild integration could then detect whether Grpc.Tools is being used in a project with nullable reference types enabled and automatically tell Grpc.Tools to include nullable reference type metadata in generated code.

jtattermusch:
Also see https://github.com/protocolbuffers/protobuf/issues/6632

jtattermusch:
In general I'm not opposed to this, but to add nullable type references support in the entire protobuf C# and the codegen is quite a lot of work. Also we would need to figure out how to make this in a backwards-compatible way (the generated code cannot require C# 8)

Falco20019:
Maybe adding a grpc-gen option for protoc? Then we could have it in Grpc.Tools per item or per property for the whole project.

jtattermusch:
@Falco20019  protoc options are a pain to maintain and test, they make codegen harder to use and can cause backward incompatibilities, so we don't want to introduce them unless it's absolutely necessary.

Falco20019:
In case this gets added at some point with .NET 5 being around the corner, we also might want to have another look at #7874 whenever we would be introducing breaking changes. Just adding this here, since it might get attention when this is resolved.

mlankamp:
Any updates about this feature?

Falco20019:
@mlankamp I did give https://github.com/protocolbuffers/protobuf/issues/6632 another visit today to see, if we are getting warnings (now with nullability being the default in the templates for .NET 6). I'm also patiently waiting for this to happen at some point. But until the aforementioned issue on the ProtoBuf repo is not addressed, there is nothing that we can do support it on the `Grpc.Tools` side.

jtattermusch:
@tonydnewell has looked into the question of nullable reference types in our codegen more in detail, as has come up with a proposal on how to handle this. The proposal is currently in a draft state here https://github.com/grpc/grpc/pull/33507, but we will make it more official soon.

tonydnewell:
See 
- https://github.com/grpc/grpc/pull/33878
- https://github.com/grpc/grpc/pull/33879

tonydnewell:
A gRFC has been published to discuss:
[...]

PR #33878, opened 2023-07-26
[C#] Add null reference types support to gRPC C# plugin

tonydnewell:
Add support for null reference types to gRPC C# plugin.

To enable, add the option "nullable_reference_types" to the plugin, e.g.
```bash
protoc --plugin=protoc-gen-grpc=grpc_csharp_plugin \
    --csharp_out=. --grpc-csharp_out=. \
    --grpc-csharp_opt=nullable_reference_types \
    myproto.proto
```

This PR is a prerequisite for adding support to Grpc.Tools - see https://github.com/grpc/grpc/issues/20729

Note: this is the gRPC plugin only - adding support to `protoc` code generation is separate (see https://github.com/protocolbuffers/protobuf/pull/13218 for one PR doing this)

### DIFF
--- a/src/compiler/csharp_generator.cc
+++ b/src/compiler/csharp_generator.cc
@@ -482,7 +482,8 @@ void GenerateServerClass(Printer* out, const ServiceDescriptor* service) {
   out->Print("\n");
 }
 
-void GenerateClientStub(Printer* out, const ServiceDescriptor* service) {
+void GenerateClientStub(Printer* out, const ServiceDescriptor* service,
+                        bool enable_nrt) {
   out->Print("/// <summary>Client for $servicename$</summary>\n", "servicename",
              GetServiceClassName(service));
   GenerateObsoleteAttribute(out, service->options().deprecated());
@@ -544,14 +545,15 @@ void GenerateClientStub(Printer* out, const ServiceDescriptor* service) {
       GenerateGeneratedCodeAttribute(out);
       out->Print(
           "public virtual $response$ $methodname$($request$ request, "
-          "grpc::Metadata "
+          "grpc::Metadata$qmark$ "
           "headers = null, global::System.DateTime? deadline = null, "
           "global::System.Threading.CancellationToken "
           "cancellationToken = "
           "default(global::System.Threading.CancellationToken))\n",
           "methodname", method->name(), "request",
           GRPC_CUSTOM_CSHARP_GETCLASSNAME(method->input_type()), "response",
-          GRPC_CUSTOM_CSHARP_GETCLASSNAME(method->output_type()));
+          GRPC_CUSTOM_CSHARP_GETCLASSNAME(method->output_type()), "qmark",
+          enable_nrt ? "?" : "");
       out->Print("{\n");
       out->Indent();
       out->Print(
@@ -591,14 +593,14 @@ void GenerateClientStub(Printer* out, const ServiceDescriptor* service) {
     GenerateGeneratedCodeAttribute(out);
     out->Print(
         "public virtual $returntype$ "
-        "$methodname$($request_maybe$grpc::Metadata "
+        "$methodname$($request_maybe$grpc::Metadata$qmark$ "
         "headers = null, global::System.DateTime? deadline = null, "
         "global::System.Threading.CancellationToken "
         "cancellationToken = "
         "default(global::System.Threading.CancellationToken))\n",
         "methodname", method_name, "request_maybe",
         GetMethodRequestParamMaybe(method), "returntype",
-        GetMethodReturnTypeClient(method));
+        GetMethodReturnTypeClient(method), "qmark", enable_nrt ? "?" : "");
     out->Print("{\n");
     out->Indent();
 
@@ -708,7 +710,8 @@ void GenerateBindServiceMethod(Printer* out, const ServiceDescriptor* service) {
 }
 
 void GenerateBindServiceWithBinderMethod(Printer* out,
-                                         const ServiceDescriptor* service) {
+                                         const ServiceDescriptor* service,
+                                         bool enable_nrt) {
   out->Print(
       "/// <summary>Register service method with a service "
       "binder with or without implementation. Useful when customizing the "
@@ -735,14 +738,15 @@ void GenerateBindServiceWithBinderMethod(Printer* out,
   for (int i = 0; i < service->method_count(); i++) {
     const MethodDescriptor* method = service->method(i);
     out->Print(
-        "serviceBinder.AddMethod($methodfield$, serviceImpl == null ? null : "
+        "serviceBinder.AddMethod($methodfield$, serviceImpl == null ? "
+        "null$nullforgiving$ : "
         "new $servermethodtype$<$inputtype$, $outputtype$>("
         "serviceImpl.$methodname$));\n",
         "methodfield", GetMethodFieldName(method), "servermethodtype",
         GetCSharpServerMethodType(method), "inputtype",
         GRPC_CUSTOM_CSHARP_GETCLASSNAME(method->input_type()), "outputtype",
         GRPC_CUSTOM_CSHARP_GETCLASSNAME(method->output_type()), "methodname",
-        method->name());
+        method->name(), "nullforgiving", enable_nrt ? "!" : "");
   }
 
   out->Outdent();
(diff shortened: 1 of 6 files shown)


tonydnewell:
@jtattermusch @jskeet for review

jskeet:
Will look at this next week. (I'm off on Friday.)

src/compiler/csharp_generator.cc
```diff
@@ -735,14 +738,15 @@ void GenerateBindServiceWithBinderMethod(Printer* out,
   for (int i = 0; i < service->method_count(); i++) {
     const MethodDescriptor* method = service->method(i);
     out->Print(
-        "serviceBinder.AddMethod($methodfield$, serviceImpl == null ? null : "
+        "serviceBinder.AddMethod($methodfield$, serviceImpl == null ? "
+        "null$nullforgiving$ : "
```
jskeet, line 742: It may be worth adding a comment for why this is necessary. (I'm about to look to work it out...)

jskeet: Having looked at the code, I still don't understand why this is necessary. Grpc.Core.Api doesn't have nullable annotations yet anyway, does it? So I'd expect the generated code to be aware of that, and not warn - and when Grpc.Core.Api *does* add nullable annotations, I'd expect ServiceBinderMase.AddMethod to make the second parameter nullable. But I may be missing something...

jskeet (commented):
Generally fine, with one nit... but it's hard to know what I might have missed.

src/compiler/csharp_generator.cc
tonydnewell: If I remove the null forgiving operator the warning is:
`CS8604 Possible null reference argument for parameter 'handler' in 'void ServiceBinderBase.AddMethod<Point, Feature>(Method<Point, Feature> method, UnaryServerMethod<Point, Feature> handler)'.`
so it does appear to be necessary.

jskeet: We should work out why that is, then. Because it clearly doesn't actually need to be non-null - or we shouldn't be passing in a null reference.

tonydnewell: The generated code says it is part of an experimental API used in customising service bind logic - not sure if/when it is used:
```C#
 /// <summary>Register service method with a service binder with or without implementation. Useful when customizing the service binding logic.
 /// Note: this method is part of an experimental API that can change or be removed without any prior notice.</summary>
 /// <param name="serviceBinder">Service methods will be bound by calling <c>AddMethod</c> on this object.</param>
 /// <param name="serviceImpl">An object implementing the server-side handling logic.</param>
 public static void BindService(grpc::ServiceBinderBase serviceBinder, RouteGuideBase serviceImpl)
 {
   // NOTE - the "null" here needs the null forgiving operation
   serviceBinder.AddMethod(__Method_GetFeature, serviceImpl == null ? null : new grpc::UnaryServerMethod<global::Routeguide.Point, global::Routeguide.Feature>(serviceImpl.GetFeature));
   // ... etc
 }
```

This isn't the `BindService` method that is normally used in Grpc.Core examples - not sure about Grpc.AspNetCore.

tonydnewell: FYI - it is called from Grpc.AspNetCore via reflection:
https://github.com/grpc/grpc-dotnet/blob/2b1da779bf45a0c55006bc3d36d1e54485c8e834/src/Grpc.AspNetCore.Server/Model/Internal/BinderServiceModelProvider.cs#L54C23-L54C23
```C#
 bindMethodInfo.Invoke(null, new object?[] { binder, null });
```

So I think keeping my fix of adding null forgiving operator is correct.

jskeet: I'm still confused... fundamentally, it sounds like the method really *should* have a nullable parameter, given that we're intentionally calling it with a null argument. So can we fix that? Or have I completely missed something?

(I'm somewhat unnerved by generating calls to an experimental API within a non-experimental generator, admittedly...)

tonydnewell: The `AddMethod` definitions are in `ServiceBinderBase.cs` in `Grpc.Core.Api`.

Maybe we should add nullable annotations to `Grpc.Core.Api` - but then the generated code is dependent on a particular version of the Grpc.Core.Api package if you don't want to see warnings.

tonydnewell: I think the safest thing (for now) is to add the annotations to the generated code, then update the code in `Grpc.Core.Api` and  `Grpc.AspNetCore.Server` later.

- If we update `Grpc.Core.Api` and add annotations to `ServiceBinderBase.cs` to say the `handler` parameter may be null then we also need to add annotations to `ProviderServiceBinder.cs` in `Grpc.AspNetCore.Server` as the base class definition will have changed and warnings are treated as errors in that project

- If we don't add the null forgiving operator annotations to the generated code then projects using new code generation but using older `Grpc.Core.Api` without the changes will have warnings.

JamesNK: The `!` is required because the `ServiceBinderBase.AddMethod` overloads aren't annotated to allow a nullable handler: https://github.com/grpc/grpc-dotnet/blob/3f908444494b01eed9ffa9f141bbe9f170aee7f6/src/Grpc.Core.Api/ServiceBinderBase.cs#L43

This should be fixed in Grpc.Core.Api. However, even if fixed there, nullable generated code would require a fixed version of Grpc.Core.Api. If an older Grpc.Core.Api version is used, then there would be a warning that null can't be passed to the method.

What I think we should do:
1. Fix `ServiceBinderBase.AddMethod` to make handler nullable.
1. Leave the `!` here but add a comment explaining why it is needed and add a TODO that can eventually be removed. Once the updated version of Grpc.Core.Api has been public for some time, maybe in a years time, then the workaround can be removed.

This would reduce the chance of people seeing warnings from generated code and give a path to a clean state.

JamesNK (approved):
Looks good, just add a comment around the `ServiceBinderBase.AddMethod` code.

src/compiler/csharp_generator.cc
JamesNK: https://github.com/grpc/grpc-dotnet/pull/2270

jtattermusch: > The `!` is required because the `ServiceBinderBase.AddMethod` overloads aren't annotated to allow a nullable handler: https://github.com/grpc/grpc-dotnet/blob/3f908444494b01eed9ffa9f141bbe9f170aee7f6/src/Grpc.Core.Api/ServiceBinderBase.cs#L43
> 
> This should be fixed in Grpc.Core.Api. However, even if fixed there, nullable generated code would require a fixed version of Grpc.Core.Api. If an older Grpc.Core.Api version is used, then there would be a warning that null can't be passed to the method.
> 
> What I think we should do:
> 
> 1. Fix `ServiceBinderBase.AddMethod` to make handler nullable in grpc/grpc-dotnet.
> 2. Leave the `!` here, but add a comment explaining why it is needed and add a TODO that it should be removed in the future. Once the updated version of Grpc.Core.Api has been public for some time, maybe in a years time, then the workaround can be removed.
> 
> This would reduce the chance of people seeing warnings from generated code and give a path to a clean state.

I agree with @JamesNK here. Also this is basically what @tonydnewell  has already proposed so we seem to be in agreement.

@tonydnewell  can you please add the comment?

src/compiler/csharp_plugin.cc
```diff
@@ -62,14 +63,17 @@ class CSharpGrpcGenerator : public grpc::protobuf::compiler::CodeGenerator {
         // The option may be removed or file names generated may change
         // in the future.
         base_namespace = options[i].second;
+      } else if (options[i].first == "enable_nrt") {
```
jtattermusch, line 66: Also, I'd say let's add a note saying that the new "enable_nrt" (or whatever name we end up choosing) option is experimental (same as "base_namespace")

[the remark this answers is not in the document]
tonydnewell: There was a discussion here about the naming on the option in protoc: https://github.com/protocolbuffers/protobuf/pull/13218#discussion_r1276246494

lmorvan:
I'm eager to see this implemented into gRPC. I never contributed to this amazing library, but can I help you in any way?

anslmorvan:
@gumbarros Same here 😭

LucioPelinson:
> Any chance to merge this? Is really important to have nullable support, does not make any sense the generated optional fields don't be nullable and causes a lot of issues migrating from another communication style.

Not supporting or properly handling nullable types—especially in C#—is highly limiting. We had planned a migration to gRPC but had to abandon it due to this limitation, as the required changes would have introduced significant breaking changes to our codebase.

PR #33879, opened 2023-07-26
[C#] Grpc.Tools changes to support nullable reference types

tonydnewell:
A support to Grpc.Tools to support nullable reference types.

This **should NOT BE MERGED** until nullable reference types are supported in the protoc and gRPC C# plugin code generators.
See:

- https://github.com/grpc/grpc/pull/33878
- https://github.com/protocolbuffers/protobuf/pull/13218
- https://github.com/protocolbuffers/protobuf/issues/6632
- https://github.com/grpc/grpc/issues/20729

Grpc.Tools detects of `<Nullable>enable<\Nullable>` is in the project file and adds the appropriate options to `protoc` and the plugin.

These can be overridden by setting the following properties to `enable` or `disable`:
- `gRPC_NullableReferenceTypes` for the plugin
- `Protobuf_NullableReferenceTypes` for protoc

Maybe just one property for both?

TODO:
- adding tests
- updating documentation (to do after NuGet package is published)

### DIFF
--- a/src/csharp/Grpc.Tools/build/_grpc/_Grpc.Tools.targets
+++ b/src/csharp/Grpc.Tools/build/_grpc/_Grpc.Tools.targets
@@ -32,6 +32,13 @@
   </Target>
 
   <Target Name="_gRPC_PrepareCompileOptions" AfterTargets="Protobuf_PrepareCompileOptions">
+
+    <PropertyGroup>
+      <!-- Enable nullabe reference types if enabled in the project and not been overridden
+           by setting gRPC_NullableReferenceTypes -->
+      <gRPC_NullableReferenceTypes Condition=" '$(gRPC_NullableReferenceTypes)' == '' and '$(Nullable)' == 'Enable' ">Enable</gRPC_NullableReferenceTypes>
+    </PropertyGroup>
+
     <ItemGroup Condition=" '$(Language)' == 'C#' ">
       <Protobuf_Compile Condition=" %(Protobuf_Compile.GrpcServices) != 'None' ">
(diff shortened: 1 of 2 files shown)


tonydnewell:
@jtattermusch @JamesNK for review
@jskeet FYI

src/csharp/Grpc.Tools/build/_grpc/_Grpc.Tools.targets
```diff
@@ -32,6 +32,13 @@
   </Target>
 
   <Target Name="_gRPC_PrepareCompileOptions" AfterTargets="Protobuf_PrepareCompileOptions">
+
+    <PropertyGroup>
+      <!-- Enable nullabe reference types if enabled in the project and not been overridden
+           by setting gRPC_NullableReferenceTypes -->
+      <gRPC_NullableReferenceTypes Condition=" '$(gRPC_NullableReferenceTypes)' == '' and '$(Nullable)' == 'Enable' ">Enable</gRPC_NullableReferenceTypes>
```
jtattermusch, line 39: I think besides being able to set the NRT handling with gRPC_NullableReferenceTypes and Protobuf_NullableReferenceTypes separately (which seems useful), we'd also need a property that enables/disables both at the same time (since likely that's what users will want if they care about NRT and if both protoc and grpc_csharp_plugin support NRT option in the future).

Or should we rely on most users simply having `Nullable == Enable` and therefore the NRT codegen being used by default (and the gRPC_NullableReferenceTypes and Protobuf_NullableReferenceTypes will only be there as an option to disable it if needed?).

stebet:
I'm really looking forward to this being merged! Awesome to see :)

Issue #33643, opened 2023-07-10
`pickle.loads` from pickled `AioRpcError` error


hzqmwne:
### What version of gRPC and what language are you using?

1.56.0

Python

### What operating system (Linux, Windows,...) and version?

Linux CentOS8

### What runtime / compiler are you using (e.g. python version or version of gcc)

Python3.10

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.

```python
import pickle

import grpc

rpc_error = grpc.aio.AioRpcError(None, None, None)

s = pickle.dumps(rpc_error)
pickle.loads(s)                         
```

### What did you expect to see?

The `pickle.loads` runs successfully.

### What did you see instead?

```
Traceback (most recent call last):
  File "/root/test.py", line 8, in <module>
    pickle.loads(s)
TypeError: AioRpcError.__init__() missing 3 required positional arguments: 'code', 'initial_metadata', and 'trailing_metadata'
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

### Anything else we should know about your project / environment?

`AioRpcError` is subclass of `Exception`, and `AioRpcError.__init__` needs three necessary arguments.
Because `BaseException` has a default `__reduce__` which does not handle extra args for `__init__`, this cause error when unpickle from pickled `AioRpcError` instance. 

Currently I use a monkey patch for `AioRpcError.__reduce__`。It works but not good. The best solution is to do this in grpcio lib. 
```python
grpc.aio.AioRpcError.__reduce__ = lambda s: (
    grpc.aio.AioRpcError,
    (s._code, s._details, s._initial_metadata, s._trailing_metadata, s._debug_error_string),
)
```

---

Why do I must pickle and unpickle an `AioRpcError` ?  In my project, I need to do a grpc aio client call in **mulprocessing** ,  and if the subprocess raise an `AioRpcError`, the mulprocessing module will transfer the exception to main process through pickle then unpickle, but unpickle error, this will cause subprocess die suddenly. Here is an example:
```python
import multiprocessing

if __name__ == "__main__":
    multiprocessing.set_start_method("forkserver")


import asyncio
[...]

PR #33891, opened 2023-07-26
[Aio AioRpcError] Allow pickle AioRpcError

XuanWang-Amos:
Fix: #33643.

This change adds `__reduce__` to `AioRpcError` to allow pickle.

### Testing

Added bazel unit test, without this change, test will fail with error:
```
TypeError: AioRpcError.__init__() missing 3 required positional arguments: 'code', 'initial_metadata', and 'trailing_metadata'
```

### DIFF
--- a/src/python/grpcio/grpc/aio/_call.py
+++ b/src/python/grpcio/grpc/aio/_call.py
@@ -153,6 +153,17 @@ class AioRpcError(grpc.RpcError):
     def __str__(self) -> str:
         return self._repr()
 
+    def __reduce__(self):
+        return (
+            type(self),
+            (
+                self._code,
+                self._initial_metadata,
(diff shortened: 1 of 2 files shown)


src/python/grpcio/grpc/aio/_call.py
```diff
@@ -153,6 +153,18 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         return self._repr()
 
+    def __reduce__(self):
```
gnossen, line 156: Can we add an accompanying test?

XuanWang-Amos: Sure, added test case in `aio_rpc_error_test.py`.

gnossen (approved):

PR #33892, opened 2023-07-26
[PSM Interop] Update and improve local-dev.cfg example

sergiitk:
The most important change here is to setting `resource_suffix` and `server_xds_port` flags to "generate randomly" by default. Previously we were suggesting static values, and devs ended up with resource conflict errors.


sergiitk:
FYI @zasweq @ginayeh

PR #33894, opened 2023-07-27
[EventEngine][Windows] Temporary changes for rare-flake debugging

drfloob:
CNR a WindowsEventEngine listener flake in:
 * 10k local Windows development machine runs
 * 50k Windows RBE runs
 * 10k Windows VM runs

It fails ~5 times per day on the master CI jobs.

This PR adds some logging to try to see if an edge is missed, and switches the thread pool implementation to see if that makes the flake go away. If the flakes disappear, I'll try removing one or the other to see if either independently fix the problem (hopefully not logging).


test/core/end2end/tests/no_logging.cc
```diff
@@ -139,6 +139,12 @@ void SimpleRequest(CoreEnd2endTest& test) {
 }
 
 CORE_END2END_TEST(NoLoggingTest, NoLoggingTest) {
+// TODO(hork): remove when the listener flake is identified
+#ifdef GPR_WINDOWS
+  if (IsEventEngineListenerEnabled()) {
+    return;
```
ctiller, line 145: GTEST_SKIP() << "not for windows + event engine listener";

would probably be preferred

ctiller (approved):

test/core/end2end/tests/no_logging.cc
drfloob: Fixed.

PR #33907, opened 2023-07-27
[benchmark] Add "dashboard" category to scenario_config.py

drfloob:
This PR covers C++ only. Other language owners: feel free to ping me if this would be useful for you.

This allows scenario_config to produce a small superset of tests required to generate the performance dashboard https://grafana-dot-grpc-testing.appspot.com/. This only adds a category to some existing scenarios, and to my knowledge, should not affect any current automation that uses scenario_config.

I plan to use this category to run gRPC-core experiments and produce equivalent dashboards. It seems worth landing this independently of those job configurations, but I'm flexible.


paulosjca (approved):
By my count this is 18 C++ scenarios (no other languages).

```shell
./tools/run_tests/performance/scenario_config_exporter.py --count_scenarios --category=dashboard
```
Assuming this is intended, LGTM.

drfloob:
> By my count this is 18 C++ scenarios (no other languages).

That is intended. I'm not sure this would be useful for other languages, but if there's a need, it would not be hard to extract the other scenarios from the grafana config.

CC @ejona86 @dfawley

PR #33909, opened 2023-07-28
[benchmark] Add gRPC experiment support to the OSS benchmarks

drfloob:
This adds a new GKE benchmark job, which runs the set of "dashboard" scenarios for every gRPC experiment configured in the script. Results are published to BigQuery at `e2e_benchmarks.ci_cxx_experiment_results_${N}core.${experiment}`

See https://github.com/grpc/grpc/pull/33907 for the scenario config.


paulosjca (approved):

tools/run_tests/performance/loadtest_config.py
```diff
@@ -529,6 +537,7 @@ def main() -> None:
         "DRIVER_PORT": "${DRIVER_PORT}",
         "KILL_AFTER": "${KILL_AFTER}",
         "POD_TIMEOUT": "${POD_TIMEOUT}",
+        "grpc_experiment": "",
```
paulosjca, line 540: On second thoughts, I would remove this line and leave the template for experiments separate from the main template.

The other keys in this dictionary are variables set by the load test controller (so part of the machinery of runnign a load test), and not just substitution keys. The key `grpc_experiment` is related to the particular template, so it should just be included as a key to be substituted (as already done in `grpc_e2e_performance_gke_cxx_experiment_framework.sh`), without making changes to the core `loadtest_config.py` script.

drfloob: This message didn't show up on my UI until after I submitted. I can make the change in another PR.

paulosjca: Sure. Review appears to be optional too.

drfloob: Followup PR: https://github.com/grpc/grpc/pull/33924

PR #33910, opened 2023-07-28
[tracing] Add annotation with metadata sizes and limits

ananda1066:
Only create annotation when call is sampled for cost reasons.


src/core/lib/channel/promise_based_filter.cc
```diff
@@ -302,6 +302,9 @@ BaseCallData::Flusher::~Flusher() {
     gpr_log(GPR_INFO, "FLUSHER:forward batch: %s",
             grpc_transport_stream_op_batch_string(release_[0], false).c_str());
   }
+  if (call_->call_context_ != nullptr && call_->call_context_->traced()) {
```
ctiller, line 305: @yousukseung the bug we talked about last week

ctiller (approved):

PR #33911, opened 2023-07-28
[EventEngine] Add TODOs to re-enable EventEngine end2end tests

drfloob:
These tests should be re-enabled before we claim confidence in the engine implementations. It seems these tests are still being run, not sure if that's true in all cases ([example](https://source.cloud.google.com/results/invocations/be524340-c98f-4915-a833-192047ae9925/targets/%2F%2Ftest%2Fcore%2Fend2end:call_creds_test@experiment%3Devent_engine_client/log)).

Alternatively, we can scrap this PR and enable all tests now if you feel you're ready to start looking at PosixEventEngine test failures.

CC @yijiem @ctiller


drfloob:
@yijiem would you prefer these be assigned to you?

Vignesh2208 (approved):

PR #33915, opened 2023-07-28
[xDS] Remove filter name from GenerateServiceConfig

rockspore:
We decided to not populate `policy_name` with the HTTP filter name in xDS case. So removing it from `GenerateServiceConfig`. This will be consistent across languages. The gRFC [PR](https://github.com/grpc/proposal/pull/346) has been updated.

### DIFF
--- a/src/core/ext/filters/rbac/rbac_service_config_parser.cc
+++ b/src/core/ext/filters/rbac/rbac_service_config_parser.cc
@@ -774,7 +774,6 @@ void RbacConfig::RbacPolicy::Rules::AuditLogger::JsonPostLoad(
 
 Rbac RbacConfig::RbacPolicy::Rules::TakeAsRbac() {
   Rbac rbac;
-  rbac.name = "";
   rbac.action = static_cast<Rbac::Action>(action);
   rbac.audit_condition = audit_condition;
   for (auto& p : policies) {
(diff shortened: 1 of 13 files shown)


rockspore:
@yashykt Would you mind taking a look at this since @markdroth is OOO? This PR is just to revert the changes regarding HTTP filter name in audit logging and I think you may have some context. Thanks.

src/core/ext/filters/rbac/rbac_service_config_parser.cc
```diff
@@ -773,9 +772,9 @@ void RbacConfig::RbacPolicy::Rules::AuditLogger::JsonPostLoad(
 // RbacConfig::RbacPolicy::Rules
 //
 
-Rbac RbacConfig::RbacPolicy::Rules::TakeAsRbac(std::string name) {
+Rbac RbacConfig::RbacPolicy::Rules::TakeAsRbac() {
   Rbac rbac;
-  rbac.name = std::move(name);
+  rbac.name = "";
```
yashykt, line 777: Looks like a no-op?

yashykt (changes requested):
Can we also update the comment on `name` in `rbac_policy.h` please

src/core/ext/filters/rbac/rbac_service_config_parser.cc
rockspore: Updated the comment. Also removed this line. I thought the absence of it would cause msan failure but now I realize std::string has default ctor.

yashykt (approved):

PR #33916, opened 2023-07-28
[Fork] Kick the EE PollPoller in prefork

yijiem:
And add some trace points. This does not solve [go/event-engine-forkable-prefork-deadlock](http://go/event-engine-forkable-prefork-deadlock), but is a necessary step.
So ¯\\_(ツ)_/¯.


drfloob (commented):
LGTM. This needs to be tested with python multiprocessing, preferably on Mac (with the `GRPC_DO_NOT_INSTANTIATE_POSIX_POLLER` macro disabled in the build). @gnossen do you have any recommendations on how to best test this?

gnossen:
The easiest way would be to change [this line](https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests/fork/_fork_interop_test.py#L57) to `poll` instead of `epoll1` and then run [this job](https://fusion2.corp.google.com/ci;ids=1923284992/kokoro/prod:grpc%2Fcore%2Fmaster%2Flinux%2Fgrpc_python_fork_bazel/activity/1b701d33-d97b-4fc6-b7f4-57b8f54b224a/summary). Ultimately, we'll want to parameterize this test so that it runs both polling engines, but that can be done as a follow-on.

yijiem:
> The easiest way would be to change [this line](https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests/fork/_fork_interop_test.py#L57) to `poll` instead of `epoll1` and then run [this job](https://fusion2.corp.google.com/ci;ids=1923284992/kokoro/prod:grpc%2Fcore%2Fmaster%2Flinux%2Fgrpc_python_fork_bazel/activity/1b701d33-d97b-4fc6-b7f4-57b8f54b224a/summary). Ultimately, we'll want to parameterize this test so that it runs both polling engines, but that can be done as a follow-on.

Thanks! It seems like there are some issues (`SIGSEGV`) when enabling the poll poller for the test. So probably need to more fixing/tweaking. Converted this to draft.

yijiem:
This is sufficient to make the python tests pass: https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fcore%2Fmaster%2Flinux%2Fa40e3141-dc83-41c5-960f-39b8b0f97ec8/activity/606faed1-d098-4eb4-af5c-378d7f10836c/log

Document the findings:
1. Note that the EE poller (epoll1 or poll) does not bear any real loads. I don't think it will work if they are. They are just created with the client channel and the EE. The current state is just to make sure that the tests which use iomgr polling do not deadlock or crash across fork.
2. The current sequence of events seems to be (to the best of my knowledge):
    2a. **Prefork**: the poller is kicked (without this a deadlock may occur) and all threads from the thread pool are shutdown.
    2b. **Postfork child**: Python will [close](https://github.com/grpc/grpc/blob/48f455a9a0c79ed7dc5f2fb2ecce20d48341ee57/src/python/grpcio/grpc/_cython/_cygrpc/fork_posix.pyx.pxi#L75) the channel in postfork child. The polling part of the EE post fork is pretty much dead since no more [Work](https://github.com/grpc/grpc/blob/48f455a9a0c79ed7dc5f2fb2ecce20d48341ee57/src/core/lib/event_engine/poller.h#L53) call will be scheduled. The EE will be shutdown and destroyed due to the close of the channel. During that process it will try to kick the poller which is prevented by this change. (this change also prevent a crash in this case)
    2c. **Postfork parent**: note that the polling part of the EE is dead. This is why I think if the EE poller bear any real loads the test wouldn't work. Though it's future work and out-of-scope for this change.
3. To conclude, the change in this PR prevents a deadlock in the EE poll poller during prefork and a crash in the postfork child.

setup.py
```diff
@@ -454,10 +454,6 @@ def _quote_build_define(argument):
         ("GRPC_ENABLE_FORK_SUPPORT", 1),
     )
 
-# Fix for multiprocessing support on Apple devices.
```
drfloob, line 457: My mistake, @yijiem. This will bring back some Python multiprocessing issues that our own tests are not yet sensitive to. Please leave the poller disabled.

yijiem: Done.

Vignesh2208 (approved):

PR #33928, opened 2023-07-31
[CallTracer] Allow registering multiple call tracers on a call context

yashykt:
This PR adds in delegating call tracers that work like so -
If this is the first call tracer that is being added onto the call context, just add it as earlier.
If this is the second call tracer that is being added onto the call context, create a delegating call tracer that contains a list of call tracers (including the first call tracer).
Any more call tracers added, just get added to the list of tracers in the delegating call tracer.

(This is not yet used other than through tests.)


src/core/lib/channel/call_tracer.cc
```diff
@@ -48,4 +57,272 @@ absl::string_view ServerCallTracerFactory::ChannelArgName() {
...
+  }
+  std::string TraceId() override { return tracers_[0]->TraceId(); }
+  std::string SpanId() override { return tracers_[0]->SpanId(); }
+  bool IsSampled() override { return tracers_[0]->IsSampled(); }
+  bool IsDelegatingTracer() override { return true; }
+
+  void AddTracer(ClientCallTracer* tracer) { tracers_.push_back(tracer); }
+
+ private:
+  std::vector<ClientCallTracer*> tracers_;
```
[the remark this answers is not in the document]
yashykt: Done

src/core/lib/channel/call_tracer.cc
```diff
@@ -48,4 +57,272 @@ absl::string_view ServerCallTracerFactory::ChannelArgName() {
...
+  }
+  std::string TraceId() override { return tracers_[0]->TraceId(); }
+  std::string SpanId() override { return tracers_[0]->SpanId(); }
+  bool IsSampled() override { return tracers_[0]->IsSampled(); }
+  bool IsDelegatingTracer() override { return true; }
+
+  void AddTracer(ServerCallTracer* tracer) { tracers_.push_back(tracer); }
+
+ private:
+  std::vector<ServerCallTracer*> tracers_;
```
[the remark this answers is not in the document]
yashykt: Done

veblush (approved):

src/core/lib/channel/call_tracer.cc
```diff
@@ -48,4 +57,284 @@ absl::string_view ServerCallTracerFactory::ChannelArgName() {
...
+    for (auto* tracer : tracers_) {
+      tracer->RecordAnnotation(annotation);
+    }
+  }
+  std::string TraceId() override { return tracers_[0]->TraceId(); }
+  std::string SpanId() override { return tracers_[0]->SpanId(); }
+  bool IsSampled() override { return tracers_[0]->IsSampled(); }
+  bool IsDelegatingTracer() override { return true; }
+
+  void AddTracer(ServerCallTracer* tracer) { tracers_.push_back(tracer); }
```
ctiller, line 267: Needs something like 'AddTracer cannot be called after the tracer has been passed to the transport' or some such thing

yashykt: Added comments in call_tracer.h. For the client-side, it is straightforward. For the server-side, we really only want the call tracers to be inserted when all the server tracer factories are invoked. I'm still thinking where to put the list of factories (one solution is to use ServerBuilderPlugins that will put in the factories in a list in a new channel arg, but it makes me wonder whether the ServerBuilderPlugin infrastructure is adding any value here). Anyway I would want that in a separate PR so that eyes can specifically look at that.

ctiller: Let's chat soon... I'm beginning to be in favor of having channel tracers be a channel/server level configuration that's independent of filters/interceptors.

ctiller (approved):

src/core/lib/channel/call_tracer.cc
yashykt: alright, i'll schedule something for next week

PR #33932, opened 2023-07-31
[Deps] Update pyyaml to 6.0.1 for bazel build system

yijiem:
The previous version (`3.12`) is 7 years old and does not support the newest Python 3 versions. This causes issues to move certain test targets (which depends on `pyyaml`) to Python 3 when some CI environment (e.g. `arm64v8/debian:11`) does not have Python 2 installed. And in general, we should move away from Python 2. Thus, updated `pyyaml` to the latest version.

This hopefully should also fix the `prod:grpc/core/master/linux/arm64/grpc_bazel_test_c_cpp` job breakage.

### DIFF
--- a/bazel/grpc_deps.bzl
+++ b/bazel/grpc_deps.bzl
@@ -533,11 +533,11 @@ def grpc_test_only_deps():
     if "com_github_yaml_pyyaml" not in native.existing_rules():
         http_archive(
             name = "com_github_yaml_pyyaml",
-            sha256 = "6b4314b1b2051ddb9d4fcd1634e1fa9c1bb4012954273c9ff3ef689f6ec6c93e",
-            strip_prefix = "pyyaml-3.12",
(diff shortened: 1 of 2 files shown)


drfloob (commented):
LGTM, but I'll defer to @gnossen and @XuanWang-Amos. I don't know of any hard requirement on that older version.

yijiem:
> LGTM, but I'll defer to @gnossen and @XuanWang-Amos. I don't know of any hard requirement on that older version.

AFAICT, pyyaml is a test only dependency and it seems to be used only in one place through Bazel: https://github.com/search?q=repo%3Agrpc%2Fgrpc+%22%5C%22yaml%5C%22%2C%22&type=code

veblush:
As this is only for dns_server test binary, it looks good as long as it passes the test.

veblush (approved):

PR #33937, opened 2023-08-01
[php] Initialise fields of the structure

mikhainin:
Polaris by Sinopsis reported "Uninitialized Scalar Variable" here. I tend to agree that "gc" isn't initialised properly. It might be not a problem at the moment but if `zend_resource` changes at some point, we might get into a trouble.


Sannis:
@ctiller can you please review this?

PR #33938, opened 2023-08-01
[php] Initialise fields of the structure

mikhainin:
One more problem reported by Polaris

[php_grpc_read_args_array()](https://github.com/grpc/grpc/blob/9aca06d38a4c914a48d2ae697af5b4b7748472ab/src/php/ext/grpc/channel.c#L117) may return `FAILURE` before args->args is allocated with will cause `efree()` against an uninitialised pointer.


Sannis:
@ctiller can you please review this?

PR #33939, opened 2023-08-01
[PSM Interop] Report per-RPC metadata if requested.

eugeneo:


yashykt (approved):

src/proto/grpc/testing/messages.proto
```diff
@@ -103,9 +103,6 @@ message SimpleRequest {
 
   // If set the server should record this metrics report data for the current RPC.
   TestOrcaReport orca_per_query_report = 11;
-
```
sanjaypujare, line 106: So the source of this file is really https://github.com/grpc/grpc-proto/blob/master/grpc/testing/messages.proto no?

I see that changes from that file are merged in here and we are also making new changes here. Will they then be merged into the file in grpc-proto? What is the normal process?

src/proto/grpc/testing/messages.proto
```diff
@@ -210,9 +207,28 @@ message LoadBalancerStatsRequest {
...
+  }
+  message RpcMetadata {
+    // Metadata associated with the RPC call
+    repeated MetadataEntry metadata = 1;
+  }
+  message MetadataByPeer {
+    // metadata values in the response for each rpc for all keys
+    // specified in LoadBalancerStatsRequest.metadata_keys
+    // Binary values are base64 encoded
+    // note this is a list of maps: each element in the list
```
sanjaypujare, line 227: The comment doesn't match the definition. There is no map as far as I can see. This is really a list of lists. The outer list is list of RPCs and the inner list is the list of key+value pair. Is that right? If correct can you fix the comment so it correctly describes the proto?

test/cpp/interop/xds_stats_watcher.cc
```diff
@@ -50,19 +55,32 @@ void XdsStatsWatcher::RpcCompleted(const AsyncClientCallResult& call,
       ++num_rpcs;
       auto rpcs_started = method_stat.rpcs_started();
       method_stat.set_rpcs_started(++rpcs_started);
+      auto rpc_metadata = metadata_by_peer_[peer].add_rpc_metadata();
```
sanjaypujare, line 58: This code (lines 58 thru 67) gets executed even when peer is empty (line 38 is true). We should protect it when peer is empty, is it not?

test/cpp/interop/xds_interop_client.cc
```diff
@@ -265,20 +269,22 @@ class LoadBalancerStatsServiceImpl : public LoadBalancerStatsService::Service {
...
-      watcher = new XdsStatsWatcher(start_id, end_id);
-      stats_watchers_->watchers.insert(watcher);
+      watcher = std::make_unique<XdsStatsWatcher>(
+          start_id, end_id,
+          std::vector<std::string>(request->metadata_keys().begin(),
+                                   request->metadata_keys().end()));
+      stats_watchers_->watchers.insert(watcher.get());
     }
-    watcher->WaitForRpcStatsResponse(response, request->timeout_sec());
+    *response = watcher->WaitForRpcStatsResponse(request->timeout_sec());
```
sanjaypujare, line 283: Why was the `response` pointer changed to returned struct? Isn't this an unnecessary copy of the object?

src/proto/grpc/testing/messages.proto
eugeneo: Yes. I will make a PR in that repo if we agree on these changes. It was easier for me to do the changes in this repo first so I could see how it works with the actual code.

test/cpp/interop/xds_interop_client.cc
eugeneo: Just a cleaner API. No need to allocate object outside, no risk of `null` being passed by mistake, etc. Compiler will likely do copy elision so there will be no performance impact.

src/proto/grpc/testing/messages.proto
eugeneo: Modified. PTAL.

test/cpp/interop/xds_stats_watcher.cc
eugeneo: Done.

sanjaypujare (changes requested):
Left some comments and requesting a couple of changes

eugeneo (commented):
Thank you for the review. Please see my responses inline.

test/cpp/interop/xds_interop_client.cc
sanjaypujare: > ...No need to allocate object outside

Well, if you are doing

```
*response = ...
```

You are assuming `response` is pointing to an already allocated object. In the previous code the same pointer was being passed inside and now the returned value (an object) is being assigned to `*response`. But I agree compiler will (might?) do copy elision so perf impact might be zero.

sanjaypujare (approved):

PR #33941, opened 2023-08-01
[benchmark] Increase the C++ benchmarks' warmup times

drfloob:
The Core/C++ implementation needs more warmup time to scale the thread pool.


paulosjca (commented):
This may require an adjustment to test concurrency and node pool sizes used for tests in CI. We can examine that once the tests are running with the new settings.

PR #33951, opened 2023-08-02
[Aio] use timeout instead of deadline for intercepted call creation

SilverBzH:
Intercepted call class expect a timeout as parameters but a deadline is provided instead. Only the UnaryUnary class is correctly created with a timeout while the others remains with the deadlines. This commit fix the instanciation of the other ones.

### DIFF
--- a/src/python/grpcio/grpc/aio/_channel.py
+++ b/src/python/grpcio/grpc/aio/_channel.py
@@ -207,7 +207,7 @@ class UnaryStreamMultiCallable(
             call = InterceptedUnaryStreamCall(
                 self._interceptors,
                 request,
-                deadline,
+                timeout,
                 metadata,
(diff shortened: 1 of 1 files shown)


src/python/grpcio/grpc/aio/_channel.py
```diff
@@ -207,7 +207,7 @@ def __call__(
             call = InterceptedUnaryStreamCall(
                 self._interceptors,
                 request,
-                deadline,
+                timeout,
```
gnossen, line 210: Can we also delete the calculation of deadline above this, since it is now unused? Here and in the other two arities?

gnossen (commented):
Thanks for the change! LGTM. Just one nit.

gnossen (approved):

PR #33959, opened 2023-08-02
[bazel] Add bazelified run_tests.py tests for ruby and python

jtattermusch:
Context: https://github.com/grpc/grpc/pull/33707

Add python and ruby tests. Both required a few fixes so I wanted to add them separately from https://github.com/grpc/grpc/pull/33707

Manual run of all the tests: https://source.cloud.google.com/results/invocations/24c7b6ed-a8b3-4367-a4bf-8828919a9340


apolcyn (approved):
lgtm for ruby

tools/bazelify_tests/test/BUILD
```diff
@@ -59,29 +59,28 @@ grpc_run_tests_py_test(
...
+    size = "enormous",
+    args = [
+        "-l ruby -c dbg",
+    ],
+    docker_image_version = "tools/dockerfile/test/ruby_debian11_x64.current_version",
+    prepare_script = ":prepare_ruby.sh",
+    use_login_shell = True,  # ruby's docker image uses RVM which wierdly requires login shell
+)
+
+grpc_run_tests_py_test(
```
gnossen, line 74: i know this isn't specific to this PR (maybe an an out-of-band clean-up), but I'm realizing that the suffix "_py_test" implies that this is a test for python code. Compare `py_test`, `py_strict_test`, etc. Maybe we could find a different name. Regardless, not super high priority and does not block merge of this PR.

gnossen (approved):

tools/bazelify_tests/test/BUILD
jtattermusch: Good catch, I'll try to come up with something better as a followup.

jtattermusch:
The ruby basictests failure https://source.cloud.google.com/results/invocations/b279fbf0-2478-4252-94d0-5a353a4d6e44 needs to be fixed. It probably needs the same fix as this:
https://github.com/grpc/grpc/pull/33959/files#diff-8f78f00ec5f5704193843849693effa382305cdf5526e61af70b779da4a96e28R19

(looks like we are running into this because I re-built the ruby image)

jtattermusch:
Another manual run of bazelified tests: https://source.cloud.google.com/results/invocations/dc40b3e4-98b2-42db-88ae-5140554b9a0f/targets;collapsed=

PR #33970, opened 2023-08-03
[benchmark] Delete references to `node_purejs`.

paulosjca:
There is only one supported implementation for performance benchmarks, so the name is unified to `node` (previously `node_purejs`).

### DIFF
--- a/tools/run_tests/performance/scenario_config.py
+++ b/tools/run_tests/performance/scenario_config.py
@@ -1659,12 +1659,8 @@ class GoLanguage(Language):
 
 
 class NodeLanguage(Language):
-    def __init__(self, node_purejs=False):
-        super().__init__()
-        self.node_purejs = node_purejs
-
     def worker_cmdline(self):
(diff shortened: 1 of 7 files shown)


tools/run_tests/performance/scenario_config.py
```diff
@@ -1659,33 +1659,23 @@ def __str__(self):
 
 
 class NodeLanguage(Language):
-    def __init__(self, node_purejs=False):
-        super().__init__()
-        self.node_purejs = node_purejs
-
     def worker_cmdline(self):
-        fixture = "native_js" if self.node_purejs else "native_native"
+        fixture = "native_js"
```
murgatroid99, line 1663: This `fixture` should be set to `js_js`.

murgatroid99 (approved):

tools/run_tests/performance/scenario_config.py
paulosjca: Done.

PR #33971, opened 2023-08-03
[ObjC] dns service resolver for cf event engine

HannahShiSFB:
re-submit of #33233 with refactored tests

1. split ios event engine tests to client tests (require oracle engine) and unit test
2. disable dns server setup in [dns_test.cc](https://github.com/grpc/grpc/blob/master/test/core/event_engine/test_suite/tests/dns_test.cc#L127) for ios test, this is what's caused the revert.
3. disable dns_test in cf_event_engine_test for MacOS


sampajano (approved):
Thanks for the fix! 😃

PR #33975, opened 2023-08-03
[EventEngine][Windows] Fix trace log use-after-free

drfloob:
Identified as a low-rate flake via manual RBE tests with all EE logging enabled.


drfloob:
Internal Windows build passes.

ctiller (approved):

PR #33980, opened 2023-08-03
[DNS resolver] enable resolver component tests on windows

apolcyn:
I can't see a reason why windows is skipped with these now - there should be no test dependency issues on bazel RBE.


apolcyn:
superseded by https://github.com/grpc/grpc/pull/34122

PR #33983, opened 2023-08-03
[core/gpr] Move subprocess to gpr and add subprocess creation using execve 

mario-vimal:
Move subprocess util to gpr.
Add support for communication with the subprocess.
This is required to support authentication using an executable.


ctiller (approved):
Please document that you're adding new functions and what those changes involve in the final commit message.

PR #33999, opened 2023-08-07
Revert "[bazel] Add bazelified run_tests.py tests for ruby and python"

XuanWang-Amos:
Reverts grpc/grpc#33959


yashykt (approved):

jtattermusch:
FTR, the failure that the original PR was causing is https://source.cloud.google.com/results/invocations/f6ed7c9b-f8b1-4bc4-b2ec-57f1fde65c23/log

```
Cloning into '/var/local/git/grpc'...
done.
fatal: detected dubious ownership in repository at '/var/local/jenkins/grpc'
To add an exception for this directory, call:

	git config --global --add safe.directory /var/local/jenkins/grpc
+ DOCKER_EXIT_CODE=128
+ '[' /tmpfs/src/github/grpc '!=' '' ']'
+ mkdir -p /tmpfs/src/github/grpc
+ cp -r /tmp/tmp.bYdAWZXxUL/reports /tmpfs/src/github/grpc
+ '[' '' '!=' '' ']'
+ exit 128

=== run_tests.py DEBUG INFO ===
command: "tools/run_tests/run_tests.py --use_docker -t -j 16 -x run_tests/python_linux_opt_native_arm64_default/sponge_log.xml --report_suite_name python_linux_opt_native_arm64_default -l python -c opt --iomgr_platform native --arch arm64 --compiler default --bq_result_table aggregate_results --measure_cpu_costs --report_multi_target"
dockerfile: tools/dockerfile/test/python_debian11_default_arm64
kokoro job name: grpc/core/master/linux/arm64/046a330e-2a0f-4227-ae36-85c2507564fa
===============================

2023-08-07 17:36:06,768 FAILED: run_tests_python_linux_opt_native_arm64_default [ret=128, pid=2320, time=14.5sec]
```

I think the rootcause is that there was a security patch for git recently https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9

The updated docker images from https://github.com/grpc/grpc/pull/33959 must have gotten a newer version of git, which then triggers the above error.
Since when https://github.com/grpc/grpc/blob/master/tools/run_tests/dockerize/build_and_run_docker.sh mounts the local git workspace to the docker container,  the files are owned by the non-root user (kbuilder on kokoro) while git inside the docker container runs as root.
(and since dir owner doesn't match the current user, git fails).
I think we need to introduce a workaround for this behavior, since it basically means that as we updating the testing docker images over time, we'll start getting more and more errors of this kind.

CC @drfloob

jtattermusch:
Alright, I found that I'm not the first one to run into this: https://github.com/grpc/grpc/pull/32317. What I need is to include the "git-jenkins.include" in the affected dockerfiles.

PR #34001, opened 2023-08-07
[core/security] Adds support for Google External Account Auth using external processes

mario-vimal:
Extends the Google External Account Auth ADC spec to support non-standard flexible on-demand mechanisms for retrieving 3rd party tokens through 3rd party external processes (3rd party scripts/executables) which can be defined and managed by the customer independent of the Google Cloud Auth libraries.


src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
```diff
@@ -0,0 +1,401 @@
...
+  return it->second.string();
+}
+
+bool isExpired(int expiration_time) {
+  return gpr_time_cmp(
+             gpr_time_from_seconds(expiration_time, GPR_CLOCK_REALTIME),
+             gpr_now(GPR_CLOCK_REALTIME)) <= 0;
+}
+
+std::string get_impersonated_email(
```
lsirac, line 82: This should be defined in the base class

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
```diff
@@ -0,0 +1,401 @@
...
+                                  std::string key) {
+  auto it = json->object().find(key.c_str());
+  return it->second.string();
+}
+
+bool isExpired(int expiration_time) {
+  return gpr_time_cmp(
+             gpr_time_from_seconds(expiration_time, GPR_CLOCK_REALTIME),
+             gpr_now(GPR_CLOCK_REALTIME)) <= 0;
+}
```
lsirac, line 80: Maybe put in a central place?

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.h
```diff
@@ -0,0 +1,77 @@
...
+
+namespace grpc_core {
+
+class PluggableAuthExternalAccountCredentials final
+    : public ExternalAccountCredentials {
+ public:
+  struct ExecutableResponse {
+    bool success;
+    int version;
+    int expiration_time;
```
lsirac, line 41: expiration_time should be a long

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
```diff
@@ -0,0 +1,401 @@
...
+PluggableAuthExternalAccountCredentials::ExecutableResponse*
+PluggableAuthExternalAccountCredentials::CreateExecutableResponse(
+    std::string executable_output_string, grpc_error_handle* error) {
+  auto executable_output = JsonParse(executable_output_string);
+  ExecutableResponse* executable_response =
+      (ExecutableResponse*)gpr_malloc(sizeof(ExecutableResponse));
+  if (!executable_output.ok()) {
+    *error = GRPC_ERROR_CREATE(
+        absl::StrFormat("The response from the executable contains an invalid "
+                        "or malformed response: %s",
```
lsirac, line 158: nit: missing period

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
```diff
@@ -0,0 +1,401 @@
...
+          MIN_EXECUTABLE_TIMEOUT_MS, MAX_EXECUTABLE_TIMEOUT_MS));
+      return;
+    }
+  }
+  if (isKeyPresent(executable_json, "output_file"))
+    output_file_path_ = getStringValue(executable_json, "output_file");
+}
+
+PluggableAuthExternalAccountCredentials::ExecutableResponse*
+PluggableAuthExternalAccountCredentials::CreateExecutableResponse(
```
lsirac, line 150: nit: ParseExecutableResponse?

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
```diff
@@ -0,0 +1,401 @@
...
+          "`message` field when unsuccessful.");
+      return nullptr;
+    }
+    executable_response->error_message =
+        gpr_strdup(getStringValue(executable_output, "message").c_str());
+  }
+  return executable_response;
+}
+
+void PluggableAuthExternalAccountCredentials::RetrieveSubjectToken(
```
lsirac, line 239: This function is pretty long. I would split it up into ~3 (validate env, retrieve from outputfile, call executable)

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
```diff
@@ -0,0 +1,401 @@
...
+          executable_timeout_ms_ / 1000)) != std::future_status::timeout) {
+    thr.join();
+    if (!executable_output_future.get()) {
+      // An error must have occured in the executable.
+      FinishRetrieveSubjectToken(
+          "", GRPC_ERROR_CREATE(absl::StrFormat(
+                  "Executable failed with error: %s", error_string)));
+      return;
+    }
+    // If the output file is specified, then the executable has stored the
```
lsirac, line 335: IIRC even if output_file is specified, they are still expected to write to STDOUT

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
```diff
@@ -0,0 +1,401 @@
...
+        // Get the result of the executable from stderr and create an
+        // ExecutableResponse object from it.
+        executable_response_ = CreateExecutableResponse(stderr_data, &error);
+      }
+    }
+    if (executable_response_ == nullptr) {
+      FinishRetrieveSubjectToken("", error);
+      return;
+    }
+    if (!executable_response_->success) {
```
lsirac, line 362: If an error occurs in the executable and there is no `code` or `message`, we should still surface that output in the exception

lsirac (commented):
Did a first pass. Heads up -- nits may apply to other parts of the code but I didn't comment on all of them. Mostly looks good, thanks Mario!

test/core/security/credentials_test.cc
```diff
@@ -3799,6 +3813,895 @@ TEST(CredentialsTest,
...
+      get_pluggable_auth_cached_executable_contents(
+          gpr_time_to_millis(gpr_inf_future(GPR_CLOCK_REALTIME)));
+  char* cached_output_file_filename =
+      write_file(pluggable_auth_file_path_prefix, cached_response_contents);
+  auto credential_source =
+      get_valid_pluggable_auth_credential_source_with_output_file(
+          "invalid executable", cached_output_file_filename);
+  TestExternalAccountCredentials::ServiceAccountImpersonation
+      service_account_impersonation;
+  service_account_impersonation.token_lifetime_seconds = 3600;
```
lsirac, line 4079: Same here & in the other tests below. Should also have a constant options var that is re-used instead of re-declaring. Let's also use "real" values, not dummy values

test/core/security/credentials_test.cc
```diff
@@ -3799,6 +3813,895 @@ TEST(CredentialsTest,
...
+                           external_account_creds_httpcli_post_success,
+                           httpcli_put_should_not_be_called);
+  state->RunRequestMetadataTest(creds.get(), kTestUrlScheme, kTestAuthority,
+                                kTestPath);
+  ExecCtx::Get()->Flush();
+  HttpRequest::SetOverride(nullptr, nullptr, nullptr);
+  UnsetEnv("GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES");
+}
+
+TEST(CredentialsTest,
```
lsirac, line 4163: These tests are mostly copy and pasted many times over. Let's refactor them to avoid duplicating them a bunch (e.g. move common logical pieces into their own function, check if we parameterize tests in this repo)

test/core/security/credentials_test.cc
```diff
@@ -3799,6 +3813,895 @@ TEST(CredentialsTest,
...
+  std::string executable_file_contents = absl::StrFormat(
+      "#!/bin/bash\n\necho \"{\\\"version\\\": 1,\\\"success\\\": "
+      "\"true\",\\\"token_type\\\": "
+      "\\\"urn:ietf:params:oauth:token-type:saml2\\\",\\\"saml_response\\\":"
+      "\\\"test_subject_token\\\",\\\"expiration_time\\\":%d}\" > %s",
+      expiration_time, output_file_filename);
+  return gpr_strdup(executable_file_contents.c_str());
+}
+
+char* get_pluggable_auth_executable_file_contents(int32_t expiration_time) {
```
lsirac, line 3936: This should be something like get_pluggable_auth_executable_saml, and the other get_pluggable_auth_executable_oidc. Why not just create & return the file?

test/core/security/credentials_test.cc
lsirac, line 1: Some more tests:
- Test get_impersonated_email()
- Workforce + pluggable auth
- An error occurs in the executable (it should be surfaced, even if no code/message)
- Validate error for when executable field is not an object
- Validate error for when timeout_millis field is not a number
- I think you are setting GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES=1 in all the tests, but we need one that ensures an error occurs when it hasn't been set
- Validate that the GOOGLE_EXTERNAL_ACCOUNT_AUDIENCE, GOOGLE_EXTERNAL_ACCOUNT_TOKEN_TYPE, GOOGLE_EXTERNAL_ACCOUNT_INTERACTIVE, GOOGLE_EXTERNAL_ACCOUNT_IMPERSONATED_EMAIL have all been set
- Process interrupted
- Non-zero exit code

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
mario-vimal: isExpired method is specific to pluggable auth as it uses Realtime clock and checks if time in "seconds" has expired, so is it fine to leave this here?
I've used an already existing method for isKeyPresent and getStringValue.

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
mario-vimal: Did you mean throwing this error, "The executable response must contain the `code` field when unsuccessful."? If you meant this, it's already done.

test/core/security/credentials_test.cc
```diff
@@ -3799,6 +3813,895 @@ TEST(CredentialsTest,
...
+  ExecCtx exec_ctx;
+  char* executable_file_contents = get_pluggable_auth_executable_file_contents(
+      gpr_time_to_millis(gpr_inf_future(GPR_CLOCK_REALTIME)));
+  char* filename =
+      write_file(pluggable_auth_file_path_prefix, executable_file_contents);
+  auto credential_source = get_valid_pluggable_auth_credential_source(filename);
+  TestExternalAccountCredentials::ServiceAccountImpersonation
+      service_account_impersonation;
+  service_account_impersonation.token_lifetime_seconds = 3600;
+  ExternalAccountCredentials::Options options = {
```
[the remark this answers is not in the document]
mario-vimal: If service_account_impersonation_url is present then there is one more request that is being made to the service account impersonation endpoint and this requires additional things to be set up. So it is left blank to make the tests focus on pluggable auth features.

'options' is a struct which is being declared and initialized, so it requires initializing all of its members with values.

test/core/security/credentials_test.cc
```diff
@@ -3799,6 +3813,895 @@ TEST(CredentialsTest,
...
+char* get_pluggable_auth_executable_file_contents_for_id_token(
+    int32_t expiration_time) {
+  std::string executable_file_contents = absl::StrFormat(
+      "#!/bin/bash\n\necho \"{\\\"version\\\": 1,\\\"success\\\": "
+      "\"true\",\\\"token_type\\\": "
+      "\\\"urn:ietf:params:oauth:token-type:jwt\\\",\\\"id_token\\\":"
+      "\\\"test_subject_token\\\",\\\"expiration_time\\\":%d}\"",
+      expiration_time);
+  return gpr_strdup(executable_file_contents.c_str());
+}
```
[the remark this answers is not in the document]
mario-vimal: Yeah, you are right. A lot of the util functions are at the top. 
Will fix this.

test/core/security/credentials_test.cc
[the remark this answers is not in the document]
lsirac:
```
 TestExternalAccountCredentials::ServiceAccountImpersonation
      service_account_impersonation;
  service_account_impersonation.token_lifetime_seconds = 3600;
```

These are the options I am referencing. They are for configuring service account impersonation, but like you said, it is not being used. Anything that isn't used should also be blank/null, no?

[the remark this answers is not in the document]
mario-vimal: Oh yeah, I just set it to default rather than null. Will set it to null then.

src/core/lib/security/credentials/external/pluggable_auth_external_account_credentials.cc
lsirac: I think you handle this case actually. Just need to add a test. This is if the executable errors out and doesn't produce a valid response, that error should be surfaced.

markdroth:
There's been no action on this for a while, so I assume it's no longer needed.  If this is something you want to pursue, let us know, and we can reopen it.

PR #34009, opened 2023-08-08
[test] Reduce the set of available listening ports for isolated runtime environments

drfloob:
There's a chance we're seeing interaction effects between containers in remote test environments. If that's happening, this change should significantly increase the failure rate.


ctiller (approved):
challenge will be disambiguating real failures from extra infra induced failures if this guess is wrong

drfloob:
I updated the max port count to ~3000, and left justification for it in a comment.

PR #34010, opened 2023-08-08
[test] Allow set request/response size in interop soak test

mohanli-ml:
Internal bug: b/289109827

### DIFF
--- a/test/cpp/interop/client.cc
+++ b/test/cpp/interop/client.cc
@@ -120,6 +120,9 @@ ABSL_FLAG(int32_t, soak_min_time_ms_between_rpcs, 0,
 ABSL_FLAG(int32_t, iteration_interval, 10,
           "The interval in seconds between rpcs. This is used by "
           "long_connection test");
+ABSL_FLAG(int32_t, soak_request_size, 271828, "The request size in a soak RPC");
+ABSL_FLAG(int32_t, soak_response_size, 314159,
+          "The response size in a soak RPC");
 ABSL_FLAG(std::string, additional_metadata, "",
           "Additional metadata to send in each request, as a "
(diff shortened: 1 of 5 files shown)


test/cpp/interop/client.cc
```diff
@@ -120,6 +120,9 @@ ABSL_FLAG(int32_t, soak_min_time_ms_between_rpcs, 0,
 ABSL_FLAG(int32_t, iteration_interval, 10,
           "The interval in seconds between rpcs. This is used by "
           "long_connection test");
+ABSL_FLAG(int32_t, soak_request_size, 271828, "The request size in a soak RPC");
```
apolcyn, line 123: nit: let's mention in the comments that by default, sizes are configured as per the "large unary" spec

test/cpp/interop/interop_client.h
```diff
@@ -146,21 +151,26 @@ class InteropClient {
...
+                       const int32_t overall_timeout_seconds,
+                       const int32_t request_size, const int32_t response_size);
 
   ServiceStub serviceStub_;
   /// If true, abort() is not called for transient failures
   bool do_not_abort_on_transient_failures_;
   // Load Orca metrics captured by the custom LB policy.
   LoadReportTracker load_report_tracker_;
+
+  static constexpr int kLargeRequestSize = 271828;
```
apolcyn: Do these constants redefined here? If not can we leave them as they were next to the other constants?

mohanli-ml: Moved it back. PTAL.

test/cpp/interop/client.cc
mohanli-ml: Done.

apolcyn (approved):

PR #24911, opened 2020-12-05
Automate Protobuf Version Templating

gnossen:
~~Fixes #24897~~

~~Since both the `protobuf` and `grpcio-tools` packages include `libprotobuf` as a shared library, an application loading both of them has symbol conflicts that need to be resolved by the operating system's runtime linker. In most cases, the runtime linker uses the first loaded symbol, ignoring the second. In corner cases, such as in MacOS in #24897, some startup functions were run by the version from `grpcio-tools` while the version packaged with `protobuf` was called at runtime. This resulted in corruption from the perspective of the version of `libprotobuf` packaged by the  `protobuf` wheel because internal data structures had changed between the two versions.~~

~~By requiring an exact version, this is no longer an issue. This also brings Python into line with other wrapped languages, including Ruby, C#, and Objective-C.~~ Since updating the Protobuf version has been a tedious manual process in the past, I went ahead and pulled all languages into the templating system so in the future, only one instance of the protobuf version string will have to be changed -- in `build_handwritten.yaml`.

CC @jtattermusch since some of these changes will have to be reconciled with https://github.com/grpc/grpc/pull/24855. Once this is merged, I'll update the "Updating third_party/protobuf submodule in grpc" document accordingly.

CC @yulin-liang since this also touches Obj C files.

**Update:** @lidizheng made the good point that the protobuf Python package has much narrower platform support than we do, so pinning the version would be too restrictive. I've removed that portion of the PR so that this is now just an automation change for updaing the protobuf version. I'll address #24897 in a series of follow-up PRs.

PR #34021, opened 2023-08-09
[python] Update to correct major version of protobuf-python

ekzhang:
Ever since Protobuf version 21, the Python protobuf package has been on major version 4, not major version 3. So the current `PROTOBUF_VERSION` constant in `grpc_version.py` has been incorrect for the past 3 releases.

### DIFF
--- a/tools/distrib/python/grpcio_tools/grpc_version.py
+++ b/tools/distrib/python/grpcio_tools/grpc_version.py
@@ -15,4 +15,4 @@
 # AUTO-GENERATED FROM `$REPO_ROOT/templates/tools/distrib/python/grpcio_tools/grpc_version.py.template`!!!
 
 VERSION = '1.58.0.dev0'
-PROTOBUF_VERSION = '3.23.4'
+PROTOBUF_VERSION = '4.23.4'


gnossen:
Thanks for catching this! This seems like a minor issue since I don't think we're actually using this anywhere.

However, the file that you've modified here is actually auto-generated. The fix will have to be in [the code generator](https://github.com/grpc/grpc/blob/e9c2feb788a18621dc24a9480de44340edfc5b42/tools/buildgen/plugins/expand_version.py#L137), not this generated file.

This ultimately stems from the fact that protobuf has _changed their version schema_ since 21. They now have a per-language version that looks like `XX.YY.ZZ`. `YY` is the cross-language version of the release. `XX` is the major version, which indicates breaking changes. The major version can differ across languages. So for Python, the current major version is 4. We'll probably need to introduce a new template variable that's a per-language major version. Then we'll need to pull that into [the various templates that are currently referencing this variable](https://github.com/grpc/grpc/blob/e9c2feb788a18621dc24a9480de44340edfc5b42/templates/tools/distrib/python/grpcio_tools/grpc_version.py.template#L20C49-L20C49). The original implementation of all this can be found [here](https://github.com/grpc/grpc/pull/24911).

PR #34024, opened 2023-08-09
[EventEngine] Improve lock contention in WorkStealingThreadPool

drfloob:
Via CPU profiling, I found that AutoThreadCount mutex operations were taking up 33% of the time in the `bm_thread_pool` benchmarks. I replaced the thread counts with atomics (reducing the set of things done while the mutexes are held), which resulted in a 15% decrease in time spent changing and waiting on thread counts.

The `BM_ThreadPool_Closure_FanOut/2/70` benchmark went from ~8~ 12 million operations per second to ~10~ 16 million. (changed CPU governor to "performance")


drfloob:
Superseded by #34065

PR #33871, opened 2023-07-25
[c-ares DNS resolver] Fix file descriptor use-after-close bug when c-ares writes succeed but subsequent read fails

apolcyn:
Normally, c-ares related fds are destroyed after all DNS resolution is finished in [this code path](https://github.com/grpc/grpc/blob/c82d31677aeea66c128a5b912ad87efcd5f74d67/src/core/ext/filters/client_channel/resolver/dns/c_ares/grpc_ares_wrapper.cc#L210). Also there are some fds that c-ares may fail to open or write to initially, and c-ares will close them internally before grpc ever knows about them.

But if:
1) c-ares opens a socket and successfully writes a request on it
2) then a subsequent read fails

Then c-ares will close the fd in [this code path](https://github.com/c-ares/c-ares/blob/bad62225b7f6b278b92e8e85a255600b629ef517/src/lib/ares_process.c#L740), but gRPC will have a reference on the fd and will still use it afterwards.

Fix here is to leverage the c-ares socket-override API to properly track fd ownership between c-ares and grpc.

Related: internal issue b/292203138

PR #34025, opened 2023-08-09
[EventEngine] Port GrpcPolledFdFactoryPosix fix to EE

yijiem:
Port https://github.com/grpc/grpc/pull/33871 to EE's GrpcPolledFdFactoryPosix.

### DIFF
--- a/src/core/lib/event_engine/ares_resolver.cc
+++ b/src/core/lib/event_engine/ares_resolver.cc
@@ -197,7 +197,7 @@ void AresResolver::Orphan() {
     }
     for (const auto& fd_node : fd_node_list_) {
       if (!fd_node->already_shutdown) {
-        GRPC_ARES_RESOLVER_TRACE_LOG("request: %p shutdown fd: %s", this,
+        GRPC_ARES_RESOLVER_TRACE_LOG("resolver: %p shutdown fd: %s", this,
                                      fd_node->polled_fd->GetName());
         fd_node->polled_fd->ShutdownLocked(
             absl::CancelledError("AresResolver::Orphan"));
@@ -356,7 +356,8 @@ void AresResolver::CheckSocketsLocked() {
         if (iter == fd_node_list_.end()) {
           new_list.push_back(std::make_unique<FdNode>(
               socks[i], polled_fd_factory_->NewGrpcPolledFdLocked(socks[i])));
-          GRPC_ARES_RESOLVER_TRACE_LOG("request:%p new fd: %d", this, socks[i]);
+          GRPC_ARES_RESOLVER_TRACE_LOG("resolver:%p new fd: %d", this,
+                                       socks[i]);
         } else {
           new_list.splice(new_list.end(), fd_node_list_, iter);
         }
@@ -371,7 +372,7 @@ void AresResolver::CheckSocketsLocked() {
             // new data arrives and c-ares hasn't read all the data in the
             // previous ares_process_fd.
             GRPC_ARES_RESOLVER_TRACE_LOG(
-                "request:%p schedule read directly on: %d", this, fd_node->as);
+                "resolver:%p schedule read directly on: %d", this, fd_node->as);
             event_engine_->Run(
                 [self = Ref(DEBUG_LOCATION, "CheckSocketsLocked"),
                  fd_node]() mutable {
@@ -379,7 +380,7 @@ void AresResolver::CheckSocketsLocked() {
                 });
           } else {
             // Otherwise register with the poller for readable event.
-            GRPC_ARES_RESOLVER_TRACE_LOG("request:%p notify read on: %d", this,
+            GRPC_ARES_RESOLVER_TRACE_LOG("resolver:%p notify read on: %d", this,
                                          fd_node->as);
             fd_node->polled_fd->RegisterForOnReadableLocked(
                 [self = Ref(DEBUG_LOCATION, "CheckSocketsLocked"),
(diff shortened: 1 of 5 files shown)


src/core/lib/event_engine/ares_resolver.cc
```diff
@@ -702,4 +707,6 @@ void noop_inject_channel_config(ares_channel* /*channel*/) {}
 void (*event_engine_grpc_ares_test_only_inject_config)(ares_channel* channel) =
     noop_inject_channel_config;
 
+bool g_event_engine_grpc_ares_test_only_force_tcp = false;
```
apolcyn, line 710: will it ever be possible to compile this and the older grpc_ares_wrapper code into the same binary?

If so we would have an ODR violation here?

I suppose we have the same issue with `noop_inject_channel_config` ?

apolcyn: ah nevermind I see the name is different

test/cpp/naming/cancel_ares_query_test.cc
```diff
@@ -444,19 +445,32 @@ TEST_F(CancelDuringAresQuery, TestQueryFailsBecauseTcpServerClosesSocket) {
           CloseSocketUponReceivingBytesFromPeer);
   // TODO(yijiem): make this test work with the EE DNS resolver by supporting
```
apolcyn, line 446: this TODO can be removed

src/core/lib/event_engine/posix_engine/grpc_polled_fd_posix.h
```diff
@@ -93,16 +103,95 @@ class GrpcPolledFdFactoryPosix : public GrpcPolledFdFactory {
...
+      &GrpcPolledFdFactoryPosix::Close /* close */,
+      &GrpcPolledFdFactoryPosix::Connect /* connect */,
+      &GrpcPolledFdFactoryPosix::RecvFrom /* recvfrom */,
+      &GrpcPolledFdFactoryPosix::WriteV /* writev */,
+  };
+
   PosixEventPoller* poller_;
+  // fds that are used/owned by grpc - we (grpc) will close them rather than
+  // c-ares
+  std::unordered_set<ares_socket_t> owned_fds_;
```
apolcyn, line 194: One sharp edge on this API that I think becomes apparent with this change:

The `GrpcPolledFdFactoryPosix` will accumulate memory on each lookup, which won't be free'd until destroyed. So it can't be used as a long-lived object. For example we'd have a leak if the client channel resolver code was changed to create one resolver and keep using it.

To double check, no code will ever use this `GrpcPolledFdFactoryPosix` as a long-lived object, right? From my reading I think we are good as of today. Wondering if there's any better enforcement than a comment to prevent this kind of mistake in the future though.

test/cpp/naming/cancel_ares_query_test.cc
yijiem: Done.

src/core/lib/event_engine/posix_engine/grpc_polled_fd_posix.h
yijiem: Each client channel resolver indirectly creates a `GrpcPolledFdFactoryPosix`, so it will be destroyed when the request finishes.

Yeah, I think this is the overhead of this workaround as we shift the responsibility of closing sockets at the end of the request from c-ares to gRPC. And we are bookkeeping these sockets for this purpose. IMO, this has the benefit of simplicity with slight memory overhead, but I wouldn't worry too much about it since the number of sockets that c-ares created is probably low (e.g. maybe 1 udp and 1 tcp socket per-server).

So this might be the upper bound of the size of all sets:

    <# of client channels> * <# of servers-per-channel> * 2 sockets per-server

apolcyn: > Each client channel resolver indirectly creates a GrpcPolledFdFactoryPosix, so it will be destroyed when the request finishes.

apolcyn: Also note this isn't a new problem - i.e. this problem exists before this PR anyways because we use `ARES_FLAG_STAYOPEN`. So *most* file descriptors will not be closed until the c-ares channel is destroyed anyways.

I just meant that, this PR made the problem more obvious to me.

apolcyn (approved):

src/core/lib/event_engine/posix_engine/grpc_polled_fd_posix.h
```diff
@@ -93,16 +103,95 @@ class GrpcPolledFdFactoryPosix : public GrpcPolledFdFactory {
...
       : poller_(poller) {}
 
+  ~GrpcPolledFdFactoryPosix() override {
+    for (auto& fd : owned_fds_) {
+      close(fd);
+    }
+  }
+
   GrpcPolledFd* NewGrpcPolledFdLocked(ares_socket_t as) override {
+    GPR_ASSERT(owned_fds_.insert(as).second);
```
yijiem, line 113: Interesting NOTE: asserting `GPR_ASSERT(owned_fds_.insert(as).second);` here is not correct for EE's `AresResolver`. EE's `AresResolver` reuses the same channel for all 3 types of requests, so it will very likely use the same socket for different requests. If the previous `GrpcPolledFdPosix` has been destroyed, it would create a new one, but `GrpcPolledFdFactoryPosix` has already seen the socket.

src/core/lib/event_engine/posix_engine/grpc_polled_fd_posix.h
yijiem: Yeah, I was also slightly concerned that we keep sockets unclosed until when we are done with the request. Then I realized that it is the existing behavior.

src/core/lib/event_engine/posix_engine/grpc_polled_fd_posix.h
apolcyn: That is interesting. So if I understand we will create a `GrpcPolledFdPosix` potentially a twice over the same socket?

If so do we have a guarantee that the first `GrpcPolledFdPosix` will not mutate the socket in such a way that it becomes not reusable? E.g. do we have a guarantee that `OrphanHandle` and anything else will not call `shutdown` ?

yijiem: Yeah, we call `OrphanHandle` with a valid `release_fd` pointer and that will prevent the socket from being `shutdown` or `close.`

PR #34027, opened 2023-08-09
[PSM Interop] Add "hook service"

eugeneo:


yashykt (commented):
probably better to use callback api :D

eugeneo:
Done. PTAL.

test/cpp/interop/pre_stop_hook_server.h
```diff
@@ -0,0 +1,56 @@
...
+#include <grpcpp/server.h>
+
+#include "src/core/lib/config/core_configuration.h"
+
+namespace grpc {
+namespace testing {
+
+class PreStopHookServer;
+
+class PreStopHookServerManager {
```
yashykt, line 35: Can you add some comments here as to what this does please?

src/proto/grpc/testing/test.proto
```diff
@@ -89,10 +91,17 @@ service LoadBalancerStatsService {
       returns (LoadBalancerAccumulatedStatsResponse) {}
 }
 
+// Hook service that may be started on request by calling XdsUpdateHealthService
+// with HookRequestCommand::START
```
yashykt, line 95: Let's add a comment on the benefit of the service over here

test/cpp/interop/pre_stop_hook_server.cc
```diff
@@ -0,0 +1,209 @@
...
+      pending_status_ = status;
+    }
+    for (auto request : pending_requests_) {
+      request->Finish(status);
+    }
+    pending_requests_.clear();
+    request_var_.SignalAll();
+  }
+
+  bool ExpectRequests(size_t expected_requests_count, size_t timeout_s) {
```
yashykt, line 64: s/ExpectRequests/TestOnlyExpectRequests

test/cpp/interop/pre_stop_hook_server.cc
```diff
@@ -0,0 +1,209 @@
...
+    request_var_.SignalAll();
+    return reactor;
+  }
+
+  void SetReturnStatus(const Status& status) {
+    grpc_core::MutexLock lock(&mu_);
+    if (pending_requests_.empty()) {
+      pending_status_ = status;
+    }
+    for (auto request : pending_requests_) {
```
yashykt, line 57: as we discussed, 1 request - 1 status

yashykt (changes requested):

test/cpp/interop/pre_stop_hook_server.h
eugeneo: Did that.

gnossen:
Adding Sanjay as author of the test spec.

yashykt (approved):
Let's change the `ExpectRequests` to `TestOnlyExpectRequests`

I would've preferred to see some discussion on the spec and the need for this, but given that this entire thing is just for interop, I'm fine-ish with checking this in and testing it.

PR #34029, opened 2023-08-10
[PSM Interop] Change the log message from Client pods to Client container

XuanWang-Amos:


tools/run_tests/xds_k8s_test_driver/framework/xds_k8s_testcase.py
```diff
@@ -670,7 +670,7 @@ def tearDown(self):
                 client_restarts,
                 0,
                 msg=(
-                    "Client pods unexpectedly restarted"
+                    "Client container unexpectedly restarted"
```
sergiitk, line 673: Let's also make this change for the server

XuanWang-Amos: Sure, changed server message too.

sergiitk (approved):

PR #34033, opened 2023-08-10
[bazel] Reintroduce bazelified ruby and python tests

jtattermusch:
Reintroduce https://github.com/grpc/grpc/pull/33959.

I added a fix for the python arm64 build (which is the reason why the change has been reverted earlier).


jtattermusch:
Non bazel test run is green: https://source.cloud.google.com/results/invocations/90d99b8b-6496-4b99-8703-1fb6384d5761

XuanWang-Amos (approved):

PR #34035, opened 2023-08-10
[docker images] Consolidate git directory ownership workaround and upgrade ccache

jtattermusch:
Based on https://github.com/grpc/grpc/pull/34033

Bunch of cleanup and rebuilding many docker images from scratch
- consolidate the workaround for "dubious ownership" issue reported by git. Other team members have run into this recently and used similar but not identical workarounds so some cleanup is due.
- rebuilding many images increases the chance that we fix the "dubious ownership" git issue early on rather than later on in the one-at-a-time fashion in the future (and the former will prevent many teammembers from wasting time on this weird issue).
- Newer version of ccache is needed for some portability tests to be able to benefit from caching (e.g. the GCC 12 portability test to get benefits of local disk caching) - this is a prerequisite for reenabling the bazelified gcc12 portability test.
- upgrade node interop images to debian:11 (since debian jessie is long past EOL).


jtattermusch:
Adhoc run of interop tests (to test node interop docker image): https://source.cloud.google.com/results/invocations/e2413e40-0346-499e-b737-f03349ee838a

veblush (approved):

PR #34036, opened 2023-08-10
[bazel rbe] Remove the need to mirror RBE linux docker image in gcr.io

jtattermusch:
- Remove no-longer-needed workaround for b/275571385 (and switch back to using docker image from GAR)
- regenerate RBE linux toolchain for bazel 6.1.2 while at it.


jtattermusch:
Bazel python test failure seems unrelated: https://source.cloud.google.com/results/invocations/ce89e0d8-2898-4fe2-951f-ffbebbb43da7

veblush (approved):

PR #34040, opened 2023-08-10
[WIP][test] Add benchmark configuration to capture CPU profiles in C++ experiments

drfloob:


drfloob:
This is currently blocked on the inability to run the linux `perf` command inside a GKE container.  To the best of my knowledge, some security settings would need to change on the host and in the docker invocation itself. Here is the `perf` error for a basic test, run inside the cxx benchmark docker image (after installing perf tools):

```
perf_5.10 record -F 777 -o sleep.perf.data -g --event=cpu-cycles sleep 3                                                                                  
perf_event_open(..., PERF_FLAG_FD_CLOEXEC) failed with unexpected error 1 (Operation not permitted)
perf_event_open(..., 0) failed unexpectedly with error 1 (Operation not permitted)
Error:
No permission to enable cpu-cycles event.
```

And the kernel settings required to capture profiles cannot be changed.
```
echo 0 > /proc/sys/kernel/kptr_restrict; echo -1 > /proc/sys/kernel/perf_event_paranoid                                                                        
bash: /proc/sys/kernel/kptr_restrict: Read-only file system
bash: /proc/sys/kernel/perf_event_paranoid: Read-only file system
```

CC @paulosjca @gnossen

PR #34041, opened 2023-08-10
[EventEngine] Temporarily disable EventEngine experiments in end2end tests

drfloob:
We are not currently using the signal that these experiment tests provide, so we can speed up the CI by disabling them for now.


drfloob:
CC @Vignesh2208

ctiller (approved):

PR #34042, opened 2023-08-10
[xds] Apply the slowdown factor only once to LRS load reporting period

yousukseung:

### DIFF
--- a/test/cpp/end2end/xds/xds_server.h
+++ b/test/cpp/end2end/xds/xds_server.h
@@ -759,8 +759,7 @@ class LrsServiceImpl
         }
       }
       response.mutable_load_reporting_interval()->set_seconds(
-          client_load_reporting_interval_seconds_ *
-          grpc_test_slowdown_factor());
+          client_load_reporting_interval_seconds_);
       stream->Write(response);
       IncreaseResponseCount();
       // Wait for report.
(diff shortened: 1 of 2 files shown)


test/cpp/end2end/xds/xds_server.h
```diff
@@ -759,8 +759,7 @@ class LrsServiceImpl
         }
       }
       response.mutable_load_reporting_interval()->set_seconds(
-          client_load_reporting_interval_seconds_ *
-          grpc_test_slowdown_factor());
+          client_load_reporting_interval_seconds_);
```
drfloob, line 762: Previously, it was `20 * grpc_test_slowdown_factor() * grpc_test_slowdown_factor()`, and now it's just `20`. Did you mean to leave one of those in?

yousukseung: No, I think 20 seconds will do everywhere and we can make debug build tests shorter. All tests start blocking until the first event occurs, and when I checked this happens in <<100ms so I doubt 20s will cause flakiness.

@markdroth is certainly the better person to make the call so I'm fine to wait for his approval.

yousukseung: Updated to apply only once to unblock for now. PTAL.

ctiller (approved):

PR #34051, opened 2023-08-12
[GSM Observability] Metadata Exchange Implementation

yashykt:
A new metadata type `x-envoy-peer-metadata` is being introduced. We don't have a better way to do this at the moment compared to just adding it in `metadata_batch.h`. 

The GSM Observability plugin uses this metadata to send topology information to peers in the form of serialized and base64 encoded `google::protobuf::Struct`. The individual keys being used inside the struct are subject to change.


test/cpp/ext/otel/otel_test_library.h
```diff
@@ -0,0 +1,94 @@
...
+
+  bool OnShutDown(std::chrono::microseconds) noexcept override { return true; }
+
+  void OnInitialized() noexcept override {}
+};
+
+class OTelPluginEnd2EndTest : public ::testing::Test {
+ protected:
+  using ::testing::Test::SetUp;
+  void SetUp(
```
markdroth, line 59: If this is not actually intended to override the gtest `SetUp()` method, maybe rename it to avoid confusion?

test/cpp/ext/otel/otel_test_library.cc
```diff
@@ -0,0 +1,132 @@
...
+  if (global_meter_provider) {
+    opentelemetry::metrics::Provider::SetMeterProvider(
+        opentelemetry::nostd::shared_ptr<opentelemetry::metrics::MeterProvider>(
+            std::move(meter_provider)));
+  } else {
+    ot_builder.SetMeterProvider(std::move(meter_provider));
+  }
+  ot_builder.SetLabelsInjector(std::move(labels_injector));
+  ot_builder.BuildAndRegisterGlobal();
+  grpc_init();
```
markdroth, line 64: Why do we need to explicitly call `grpc_init()` and `grpc_shutdown()` in these tests?

test/cpp/ext/otel/otel_test_library.cc
```diff
@@ -0,0 +1,132 @@
...
+
+void OTelPluginEnd2EndTest::TearDown() {
+  server_->Shutdown();
+  grpc_shutdown_blocking();
+  delete grpc_core::ServerCallTracerFactory::Get(grpc_core::ChannelArgs());
+  grpc_core::ServerCallTracerFactory::RegisterGlobal(nullptr);
+}
+
+void OTelPluginEnd2EndTest::ResetStub(std::shared_ptr<Channel> channel) {
+  stub_ = EchoTestService::NewStub(channel);
```
markdroth, line 89: `std::move()`

test/cpp/ext/gsm/metadata_exchange_test.cc
```diff
@@ -0,0 +1,187 @@
...
+#include "test/cpp/end2end/test_service_impl.h"
+#include "test/cpp/ext/otel/otel_test_library.h"
+
+namespace grpc {
+namespace testing {
+namespace {
+
+class MetadataExchangeTest : public OTelPluginEnd2EndTest {
+ protected:
+  void SetUp(const absl::flat_hash_set<absl::string_view>& metric_names,
```
markdroth, line 46: Why is this needed?  Seems like it's just forwarding to the parent class, but if we didn't define this method in the first place, we'd just wind up directly calling the method from the parent class anyway.

markdroth (commented):
This looks good overall!  Comments are mostly minor, modulo a few questions about performance.

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -0,0 +1,243 @@
...
+                                 kMetadataExchangeCanonicalServiceNameKey,
+                                 canonical_service_value, arena.ptr());
+  size_t output_length;
+  char* output =
+      google_protobuf_Struct_serialize(metadata, arena.ptr(), &output_length);
+  serialized_labels_to_send_ = grpc_core::Slice::FromCopiedString(
+      absl::Base64Escape(absl::string_view(output, output_length)));
+  // Fill up local labels map. The rest we get from the detected Resource and
+  // from the peer.
+  local_labels_.emplace_back("canonical_service_name", canonical_service_value);
```
markdroth, line 151: I think we don't actually need to add canonical_service_name as a local label, because that will be part of the MonitoredResource, right?  I think we just need to record the remote canonical_service_name.

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -0,0 +1,243 @@
...
+void AddStringKeyValueToStructProto(google_protobuf_Struct* struct_pb,
+                                    absl::string_view key,
+                                    absl::string_view value, upb_Arena* arena) {
+  google_protobuf_Value* value_pb = google_protobuf_Value_new(arena);
+  google_protobuf_Value_set_string_value(value_pb, AbslStrToUpbStr(value));
+  google_protobuf_Struct_fields_set(struct_pb, AbslStrToUpbStr(key), value_pb,
+                                    arena);
+}
+
+std::string GetStringValueFromAttributeMap(
```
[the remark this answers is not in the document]
yashykt: Done

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -42,30 +42,35 @@ namespace internal {
...
-const char kPeerTypeAttribute[] = "peer_type";
-const char kPeerPodNameAttribute[] = "peer_pod_name";
-const char kPeerContainerNameAttribute[] = "peer_container_name";
-const char kPeerNamespaceNameAttribute[] = "peer_namespace_name";
-const char kPeerClusterNameAttribute[] = "peer_cluster_name";
-const char kPeerLocationAttribute[] = "peer_location";
-const char kPeerProjectIdAttribute[] = "peer_project_id";
-const char kPeerCanonicalServiceNameAttribute[] = "peer_canonical_service_name";
+// The keys that will be used for the peer attributes when recording metrics.
+const absl::string_view kPeerTypeAttribute = "gsm.remote_type";
```
markdroth, line 56: All of these strings except for canonical_service_name should be `gsm.remote_workload_*` instead of just `gsm.remote_*`.

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -42,30 +42,35 @@ namespace internal {
...
+const absl::string_view kPeerPodNameAttribute = "gsm.remote_pod_name";
+const absl::string_view kPeerContainerNameAttribute =
+    "gsm.remote_container_name";
+const absl::string_view kPeerNamespaceNameAttribute =
+    "gsm.remote_namespace_name";
+const absl::string_view kPeerClusterNameAttribute = "gsm.remote_cluster_name";
+const absl::string_view kPeerLocationAttribute = "gsm.remote_location";
+const absl::string_view kPeerProjectIdAttribute = "gsm.remote_project_id";
+const absl::string_view kPeerCanonicalServiceNameAttribute =
+    "gsm.remote_canonical_service_name";
```
markdroth, line 66: This one should not have the `_name` suffix, as per the way the spec is currently written (although we might change this later if we decide to add revision).

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -25,16 +25,37 @@
...
+
 namespace grpc {
 namespace internal {
 
+class LabelsInjector {
+ public:
+  virtual ~LabelsInjector() {}
+  // Read the incoming initial metadata to get the set of labels to be added to
+  // metrics. (Includes the local labels.)
+  virtual std::vector<std::pair<std::string, std::string>> GetPeerLabels(
```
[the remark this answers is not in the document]
yashykt: > Do we expect more than one value for the same label key? If not, why use a vector of pairs instead of a map?
Since the allocation of memory for a vector is contiguous, the iteration over a vector is faster than a map.

> The local labels (both keys and values) should be fixed for the entire process, so those should be able to be allocated only once, not once per RPC. Can we expose them as absl::string_view instead of std::string?

There are two kinds of local labels, 1) those that we need to send over to the peer and 2) those that we need to record along with our metrics (since they are not present in MonitoredResource)
For 1, yes, these are saved as a `grpc_core::Slice` and that's how we add them to the out going initial metadata
For 2, since OTel C++ accepts std::string and not absl::string_view, doing this, won't help us.

> The peer label keys and values could probably be exposed as Slice, which would take a ref instead of making a copy.
The same issue, right now, we don't have a way of passing a non std::string string type to OTel C++. 

> Maybe instead of using either a vector or a map, we can return an iterable interface?
I think we can pass an iterable interface to OTel C++, but we would first need the ability to pass in string_view.

test/cpp/ext/gsm/metadata_exchange_test.cc
yashykt: Ah, it was supposed to be a convenience thing, but I didn't actually use it. Used it now.

test/cpp/ext/gsm/metadata_exchange_test.cc
```diff
@@ -123,12 +122,8 @@ TEST_F(MetadataExchangeTest, ClientAttemptStarted) {
...
 
 TEST_F(MetadataExchangeTest, ClientAttemptDuration) {
-  Init(/*metric_names=*/{grpc::internal::
-                             OTelClientAttemptDurationInstrumentName()},
-       /*resource=*/TestGkeResource(),
-       /*labels_injector=*/
-       std::make_unique<grpc::internal::ServiceMeshLabelsInjector>(
-           TestGkeResource().GetAttributes()));
+  Init(/*metric_names=*/{
+      grpc::internal::OTelClientAttemptStartedInstrumentName()});
```
markdroth, line 126: Shouldn't this be `OTelClientAttemptDurationInstrumentName()`?

test/cpp/ext/gsm/metadata_exchange_test.cc
```diff
@@ -151,12 +146,8 @@ TEST_F(MetadataExchangeTest, ClientAttemptDuration) {
...
 
 TEST_F(MetadataExchangeTest, ServerCallDuration) {
-  Init(
-      /*metric_names=*/{grpc::internal::OTelServerCallDurationInstrumentName()},
-      /*resource=*/TestGkeResource(),
-      /*labels_injector=*/
-      std::make_unique<grpc::internal::ServiceMeshLabelsInjector>(
-          TestGkeResource().GetAttributes()));
+  Init(/*metric_names=*/{
+      grpc::internal::OTelClientAttemptStartedInstrumentName()});
```
markdroth, line 150: Shouldn't this be `OTelServerCallDurationInstrumentName()`?

src/cpp/ext/otel/otel_plugin.h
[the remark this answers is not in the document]
markdroth: > Since the allocation of memory for a vector is contiguous, the iteration over a vector is faster than a map.

I'm not sure how much that really matters, since the strings themselves may not be stored inline, so we could be reading from different locations in memory anyway.  I guess that if a `std::string` happens to be inlined, it might help, but certainly if we switch to `absl::string_view`, this would still be an issue.

> since OTel C++ accepts std::string and not absl::string_view, doing this, won't help us.

I argue that even if OT currently does not support this, it's better for us to have the right interface between two components in our own code, so that if OT fixes this later, we have less work to do.  In other words, even if we still need to construct a `std::string` here, I think it's better to do it in the `CallTracer` than to bake this assumption into the `LabelsInjector` API.

> I think we can pass an iterable interface to OTel C++, but we would first need the ability to pass in string_view.

Unless I'm missing something here (and please tell me if I am), I think that's actually an orthogonal issue.  Even if OT requires `std::string`, we can still construct the `std::string` as we return each element of the iterable interface.  Doing that would still avoid the allocation for the vector itself.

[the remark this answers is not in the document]
yashykt: > I'm not sure how much that really matters, since the strings themselves may not be stored inline, so we could be reading from different locations in memory anyway. I guess that if a std::string happens to be inlined, it might help, but certainly if we switch to absl::string_view, this would still be an issue.

The difference may not be much, but I don't see a reason to use a map given that a simple vector satisfies our purpose.

> I argue that even if OT currently does not support this, it's better for us to have the right interface between two components in our own code, so that if OT fixes this later, we have less work to do. In other words, even if we still need to construct a std::string here, I think it's better to do it in the CallTracer than to bake this assumption into the LabelsInjector API.

I was wrong. The OTel API actually does take in an `absl::string_view`. It's just that the SDK later converts it to a std::string. Irrespective, you are right, I shouldn't unnecessarily create a string when an `absl::string_view` does the job. I wonder why this wasn't working for me earlier.. maybe it's because they use their own bespoke `opentelemetry::nostd::string_view` instead of `absl::string_view`

> Unless I'm missing something here (and please tell me if I am), I think that's actually an orthogonal issue. Even if OT requires std::string, we can still construct the std::string as we return each element of the iterable interface. Doing that would still avoid the allocation for the vector itself.

Created an iterator. In the new logic, I use a std::string to store the labels that are read from the peer while we use absl::string_view for our local labels

test/cpp/ext/gsm/metadata_exchange_test.cc
yashykt: Fixed

test/cpp/ext/gsm/metadata_exchange_test.cc
yashykt: Fixed

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -0,0 +1,243 @@
...
+  return "unknown";
+}
+
+}  // namespace
+
+ServiceMeshLabelsInjector::ServiceMeshLabelsInjector(
+    const opentelemetry::sdk::common::AttributeMap& map) {
+  upb::Arena arena;
+  auto* metadata = google_protobuf_Struct_new(arena.ptr());
+  // Assume kubernetes for now
```
markdroth, line 129: Even if we're not supporting the labels for other environments right now, I think we should at least be smart enough not to try to populate the GKE labels if the environment is not GKE.  If we see a non-GKE environment, we should populate only the remote_workload_type and nothing else for now.  Then we can easily add support for other types later, but at least we won't be populating the wrong labels if someone runs on such an environment with this code.

src/cpp/ext/gsm/metadata_exchange.h
```diff
@@ -40,19 +41,20 @@ class ServiceMeshLabelsInjector : public LabelsInjector {
...
+  std::vector<std::pair<absl::string_view, absl::string_view>> GetLocalLabels()
+      override;
 
   // Modify the outgoing initial metadata with metadata information to be sent
   // to the peer.
   void AddLabels(grpc_metadata_batch* outgoing_initial_metadata) override;
 
  private:
-  std::vector<std::pair<std::string, std::string>> local_labels_;
+  std::vector<std::pair<absl::string_view, absl::string_view>> local_labels_;
```
markdroth, line 57: Is this actually going to be able to use `absl::string_view`?  Where is the string going to be owned?

src/cpp/ext/otel/otel_plugin.h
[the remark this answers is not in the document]
markdroth: The changes to use `absl::string_view` look good!  (As a side note, I realized that my earlier comment about using `Slice` doesn't make sense, since we're not encoding any of the label values as a single string by itself; we're always encoding them into a protobuf first.)

However, I think the way you're using the iterable here isn't sufficient to actually solve the problem that I was concerned about.  It looks like you're using the iterable between the `CallTracer` and OT, but you're still using a vector between the `LabelsInjector` and the `CallTracer`, so this isn't actually helping us reduce the unnecessary overhead between the components in our own code.

What I had intended here was to use an iterable interface to return the data from `LabelsInjector` to `CallTracer`, to avoid the need for allocating the vector.  For example, in `GetPeerLabels()`, the iterable could hold a slice ref to the metadata value.  On its first iteration, it could deserialize the protobuf.  It can have a fixed list of labels that it will look at and track the current index into that last, so that it knows which element to return on each iteration.  This would avoid the need to allocate a vector, and it would avoid the need to allocate a `std::string` for each label value (it could be returned as a `string_view`).

Note that once you have such an iterable interface between `LabelsInjector` and `CallTracer`, it should be trivial to wrap that iterable in the OT iterable interface that you need to pass the data from `CallTracer` to OT.

src/core/lib/transport/metadata_batch.h
```diff
@@ -305,6 +305,13 @@ struct GrpcTagsBinMetadata : public SimpleSliceBasedMetadata {
   static absl::string_view key() { return "grpc-tags-bin"; }
 };
 
+// XEnvoyPeerMetadata
+struct XEnvoyPeerMetadata : public SimpleSliceBasedMetadata {
```
ctiller, line 309: How frequently does this change? Is it expected to be constant over a single client<->server pair? Changing between a few different values?

ctiller: (also how big is this expected to be... min/median/max bytes? - a rough guess would suffice)

markdroth: If there are no proxies in the middle, it will always be the same for each client/server pair.  The problem is that we can't depend on that, because if there is a proxy in between, then the proxy could route each request from a given client to a different server (i.e., the client might see a different value for each request), and the proxy could send requests from different clients on the same connection to a server (i.e., the server might see a different value for each request).

FWIW, keep in mind that in the long run, we want to completely eliminate this mechanism in favor of some out-of-band lookup for the metadata, so hopefully this isn't going to stick around forever.  But there are obviously no guarantees.

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: Done

src/core/lib/transport/metadata_batch.h
yashykt: The metadata that we send remains the same for the lifetime of the binary.  The metadata that we receive as Mark mentioned is potentially changing.

ctiller: Ok... and the other question: how big is it?
(I'm trying to sanity check whether the compression policy makes sense)

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -120,6 +122,116 @@ absl::string_view GetStringValueFromUpbStruct(google_protobuf_Struct* struct_pb,
...
+
+ private:
+  size_t pos_ = 0;
+  const std::vector<std::pair<absl::string_view, std::string>>& labels_;
+};
+
+class PeerLabelsIterable : public LabelsIterable {
+ public:
+  explicit PeerLabelsIterable(grpc_core::Slice remote_metadata) {
+    std::string decoded_metadata;
```
markdroth, line 151: Instead of doing this decoding here, I suggest doing it in `Next()` when `pos_` is 0.  That way, we're not holding on to the memory for the deserialized proto until we actually need it.  Note that for a streaming call, there can be a long time between receiving the initial metadata and the end of the call, which is where we actually need to do the iteration.

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -120,6 +122,116 @@ absl::string_view GetStringValueFromUpbStruct(google_protobuf_Struct* struct_pb,
...
+      return absl::nullopt;
+    }
+    if (pos_++ == 0) {
+      return std::make_pair(kPeerTypeAttribute, type_);
+    }
+    // Only handle GKE type for now.
+    if (type_ != kGkeType) {
+      return absl::nullopt;
+    }
+    switch (pos_ - 1) {
```
markdroth, line 174: Why subtract 1 here?  Seems like the case statements could just be incremented by 1.

src/core/lib/transport/metadata_batch.h
yashykt: So, the format of the payload is a base64 encoded serialized proto struct.. For my test values, the size was 300 bytes.

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -120,6 +122,116 @@ absl::string_view GetStringValueFromUpbStruct(google_protobuf_Struct* struct_pb,
...
+    }
+    return 8;
+  }
+
+  void ResetIteratorPosition() override { pos_ = 0; }
+
+ private:
+  upb::Arena arena_;
+  google_protobuf_Struct* struct_pb_ = nullptr;
+  absl::string_view type_;
```
markdroth, line 231: To save a bit of memory, this could be an enum.  That would also make the conditional more efficient in the future when we add support for GCE and other environments.

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -178,59 +295,18 @@ ServiceMeshLabelsInjector::ServiceMeshLabelsInjector(
...
-  labels.emplace_back(
-      kPeerProjectIdAttribute,
-      GetStringValueFromUpbStruct(struct_pb, kMetadataExchangeProjectIdKey,
-                                  arena.ptr()));
-  labels.emplace_back(
-      kPeerCanonicalServiceAttribute,
-      GetStringValueFromUpbStruct(
-          struct_pb, kMetadataExchangeCanonicalServiceKey, arena.ptr()));
-  return labels;
+  return std::make_unique<PeerLabelsIterable>(std::move(peer_metadata).value());
```
markdroth, line 305: Suggest using `*std::move(peer_metadata)` instead of using the `value()` method.  See go/totw/181.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -202,5 +202,48 @@ absl::flat_hash_set<std::string> OpenTelemetryPluginBuilder::BaseMetrics() {
           OTelServerCallRcvdTotalCompressedMessageSizeInstrumentName())};
 }
 
+absl::string_view StatusCodeToString(grpc_status_code code) {
```
markdroth, line 205: We already have a function to do this:

https://github.com/grpc/grpc/blob/fb635a1959e8828a70194134c29a87c0c6b8c0fd/src/core/lib/channel/status_util.h#L37

src/cpp/ext/otel/otel_call_tracer.h
```diff
@@ -96,9 +95,9 @@ class OpenTelemetryCallTracer : public grpc_core::ClientCallTracer {
     const bool arena_allocated_;
     // Start time (for measuring latency).
     absl::Time start_time_;
-    std::vector<std::pair<absl::string_view,
-                          absl::variant<absl::string_view, std::string>>>
-        labels_;
+    std::string method_;
```
markdroth, line 98: Can this be stored as a slice to avoid an allocation?

src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -137,38 +131,27 @@ class OpenTelemetryServerCallTracer : public grpc_core::ServerCallTracer {
  private:
   absl::Time start_time_;
   absl::Duration elapsed_time_;
-  std::vector<std::pair<absl::string_view,
-                        absl::variant<absl::string_view, std::string>>>
-      labels_;
-  bool received_peer_labels_ = false;
+  std::string method_;
```
markdroth, line 134: Same question as on client side: Can we store this as a slice to avoid an allocation?

markdroth (commented):
The iteration interface looks nice!  Thanks for doing that.

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: We either hold on to the metadata or the deserialized proto. 
Note that for `Size()`, we need the deserialized proto anyway. We *could* do it lazily the first time we need it?

src/cpp/ext/otel/labels_iterable.h
```diff
@@ -0,0 +1,119 @@
...
+namespace grpc {
+namespace internal {
+
+// An iterable container interface that can be used as a return type for the
+// OTel plugin's label injector.
+class LabelsIterable {
+ public:
+  virtual ~LabelsIterable() = default;
+
+  virtual absl::optional<std::pair<absl::string_view, absl::string_view>>
```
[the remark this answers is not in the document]
yashykt: Done

src/cpp/ext/otel/labels_iterable.h
```diff
@@ -0,0 +1,119 @@
...
+
+inline opentelemetry::nostd::string_view AbslStrViewToOTelStrView(
+    absl::string_view str) {
+  return opentelemetry::nostd::string_view(str.data(), str.size());
+}
+
+// An iterable class based on opentelemetry::common::KeyValueIterable that
+// allows gRPC to iterate on its various sources of attributes and avoid an
+// allocation in cases wherever possible.
+template <typename T>
```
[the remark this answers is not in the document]
yashykt: That works as well

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: done

src/cpp/ext/otel/otel_plugin.cc
yashykt: ah Thanks!

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: Changed to do it lazily

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: done

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: ah, thanks! done

src/cpp/ext/otel/labels_iterable.h
```diff
@@ -0,0 +1,119 @@
...
+#include "opentelemetry/common/key_value_iterable.h"
+#include "opentelemetry/nostd/function_ref.h"
+#include "opentelemetry/nostd/string_view.h"
+
+namespace grpc {
+namespace internal {
+
+// An iterable container interface that can be used as a return type for the
+// OTel plugin's label injector.
+class LabelsIterable {
```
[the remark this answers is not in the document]
yashykt: done

src/cpp/ext/otel/otel_call_tracer.h
yashykt: uh, this should be unused actually.. removed

src/cpp/ext/otel/otel_server_call_tracer.cc
yashykt: done

src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -131,16 +134,17 @@ class OpenTelemetryServerCallTracer : public grpc_core::ServerCallTracer {
  private:
   absl::Time start_time_;
   absl::Duration elapsed_time_;
-  std::string method_;
+  grpc_core::Slice path_;
+  absl::string_view method_;
```
markdroth, line 138: Instead of storing this as a data member, we could just have a method that returns this value by calling `absl::StripPrefix(path_.as_string_view(), "/")`.  That would reduce per-call memory with very little additional CPU overhead.

Same thing in the client call tracer, I suspect.

test/cpp/ext/gsm/metadata_exchange_test.cc
```diff
@@ -41,50 +41,34 @@ namespace grpc {
...
-    OTelPluginEnd2EndTest::Init(
-        metric_names, /*resource=*/TestGkeResource(),
-        /*labels_injector=*/
-        std::make_unique<grpc::internal::ServiceMeshLabelsInjector>(
-            TestGkeResource().GetAttributes()));
+class TestScenario {
+ public:
+  enum class Type : std::uint8_t { kGke, kUnknown };
+
+  TestScenario(Type type) : type_(type) {}
```
markdroth, line 48: `explicit`

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -226,9 +240,35 @@ class PeerLabelsIterable : public LabelsIterable {
...
+    std::string decoded_metadata;
+    bool metadata_decoded =
+        absl::Base64Unescape(slice->as_string_view(), &decoded_metadata);
+    metadata_ = StructPb{};
+    auto& struct_pb = absl::get<StructPb>(metadata_);
+    if (metadata_decoded) {
+      struct_pb.struct_pb = google_protobuf_Struct_parse(
+          decoded_metadata.c_str(), decoded_metadata.size(),
+          struct_pb.arena.ptr());
+      type_ = StringToGcpResourceType(GetStringValueFromUpbStruct(
```
markdroth, line 262: Suggest not setting `type_` until line 170, where we return the type attribute.  That way, we don't need to call `GetStringValueFromUpbStruct()` on this field twice.

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -147,77 +157,81 @@ class LocalLabelsIterable : public LabelsIterable {
...
-      case 7:
-        return std::make_pair(
-            kPeerCanonicalServiceAttribute,
-            GetStringValueFromUpbStruct(struct_pb_,
-                                        kMetadataExchangeCanonicalServiceKey,
-                                        arena_.ptr()));
-      default:
+    switch (type_) {
+      case GcpResourceType::kGke:
+        switch (pos_) {
```
markdroth, line 178: Consider using a static array here:

```
struct GkeAttribute {
  absl::string_view otel_attribute;
  absl::string_view metadata_attribute;
};
static const std::array<GkeAttribute, 7> kGkeAttributeList = {
  { kPeerPodNameAttribute, kMetadataExchangePodNameKey },
  { kPeerContainerNameAttribute, kMetadataExchangeContainerNameKey },
  // ...
};
```

Then the code here becomes:

```
if (pos_ - 1 >= kGkeAttributeList.size()) return absl::nullopt;
const GkeAttribute& attr = kGkeAttributeList[pos_ - 1];
return std::make_pair(
    attr.otel_attribute,
    GetStringValueFromUpbStruct(
        struct_pb.struct_pb,
        attr.metadata_attribute,
        struct_pb.arena.ptr()));
```

Also, on line 237, you can change the hard-coded `8` to `kGkeAttributeList.size() + 1`.

With this approach, if we ever add new labels here, there will be only one place to update the code.

markdroth (commented):
This looks really good!  Just a few suggestions left.

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: Note that we also need this information in `Size()` and we don't have any guarantees on which will be called first. (At present, `size()` is not used by OTel, but that might change in the future)

markdroth: Good point.

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: done

test/cpp/ext/gsm/metadata_exchange_test.cc
yashykt: done

src/cpp/ext/otel/otel_server_call_tracer.cc
yashykt: done

src/cpp/ext/gsm/metadata_exchange.cc
```diff
@@ -0,0 +1,344 @@
...
+
+  size_t Size() const override {
+    auto& struct_pb = GetDecodedMetadata();
+    if (struct_pb.struct_pb == nullptr) {
+      return 0;
+    }
+    if (type_ != GcpResourceType::kGke) {
+      return 1;
+    }
+    return 8;
```
markdroth, line 201: Please change this to use `kGkeAttributeList.size() + 1`.

markdroth (approved):
Looks great!

src/cpp/ext/gsm/metadata_exchange.cc
yashykt: ofcourse, that's what we wanted to do in the first place!

yashykt:
Thanks for reviewing!

PR #34054, opened 2023-08-14
[pick_first] de-experiment pick first

mohanli-ml:
De-experiment pick first since we have both affinity and randomness E2E test running successfully.


ctiller:
What impact will this have on prod?

mohanli-ml:
> What impact will this have on prod?

I made this change for DirectPath cloud-to-prod. Currently when service owner specify `affinity` in their c2p_directpath config, gRPC will use ring hash instead of pick first as the LB policy. With this change, gRPC will use pick first as the default of affinity.

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -62,10 +62,9 @@ namespace grpc_core {
 
 TraceFlag grpc_lb_pick_first_trace(false, "pick_first");
 
-// TODO(eostroukhov): Remove once this feature is no longer experimental.
 bool ShufflePickFirstEnabled() {
```
markdroth, line 65: Let's just remove this completely.  If it's passing interop tests, we no longer need the knob.

test/core/xds/xds_lb_policy_registry_test.cc
```diff
@@ -486,13 +486,13 @@ TEST(PickFirst, ShuffleOmitted) {
   EXPECT_EQ(*result, "{\"pick_first\":{\"shuffleAddressList\":false}}");
 }
 
-TEST(PickFirst, EnvVarDisabled) {
+TEST(PickFirst, EnvVarEnabled) {
```
markdroth, line 489: This test no longer has value, so let's just remove it.

markdroth:
> What impact will this have on prod?

As per the [gRFC](https://github.com/grpc/proposal/blob/master/A62-pick-first.md#temporary-environment-variable-protection), the env var guards two things:
1. Looking at the `shuffleAddressList` field in the pick_first LB policy config.
2. Accepting the pick_first LB policy config via xDS.

I don't think either one affects prod.

test/core/xds/xds_lb_policy_registry_test.cc
mohanli-ml: Done!

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -94,11 +83,7 @@ class PickFirstConfig : public LoadBalancingPolicy::Config {
   }
 
   void JsonPostLoad(const Json& /* json */, const JsonArgs& /* args */,
```
markdroth, line 85: This method can be removed.  It doesn't need to be provided if it's a no-op.

markdroth (commented):
Please also remove all cases where we set the `GRPC_EXPERIMENTAL_PICKFIRST_LB_CONFIG` env var in tests.  Since the code no longer looks at that env var, there's no need for the tests to set it.

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
mohanli-ml: Done. Please take another look. Thanks!

markdroth (approved):
Looks great!

PR #34056, opened 2023-08-14
[EventEngine] C++ Alarm migration and PosixEventEngine performance enhancements

drfloob:
This PR is mainly a set of improvements that allow the C++ Alarm to be migrated away from legacy iomgr. It cannot be landed without significant speedup, due to third-parties relying on a fast path for immediate timer execution with deadlines <= now.

Previous EventEngine performance of bm_alarm, compared to baseline iomgr timers: *0.014%*
This PR: *2.5%*

Regarding previous failures to land this change: The cloud libraries team agreed to reduce the amount of stress in their alarm stress test https://github.com/googleapis/google-cloud-cpp/pull/12378


src/cpp/common/alarm.cc
```diff
@@ -52,78 +62,87 @@ class AlarmImpl : public grpc::internal::CompletionQueueTag {
...
+          OnCallbackAlarm(true);
+        });
   }
   void Cancel() {
-    grpc_core::ApplicationCallbackExecCtx callback_exec_ctx;
     grpc_core::ExecCtx exec_ctx;
-    grpc_timer_cancel(&timer_);
+    if (callback_armed_.load() &&
+        event_engine_->Cancel(callback_timer_handle_)) {
+      event_engine_->Run([this] { OnCallbackAlarm(/*is_ok=*/false); });
```
Vignesh2208, line 98: should we set callback_armed_.store(false) inside the if statement ?

src/cpp/common/alarm.cc
```diff
@@ -52,78 +62,87 @@ class AlarmImpl : public grpc::internal::CompletionQueueTag {
...
+    }
+    if (cq_armed_.load() && event_engine_->Cancel(cq_timer_handle_)) {
+      event_engine_->Run(
+          [this] { OnCQAlarm(absl::CancelledError("cancelled")); });
```
Vignesh2208, line 102: Same question as above w.r.t cq_armed_

src/cpp/common/alarm.cc
drfloob: Yes indeed, good catch. That means we don't have any tests that exercise a successful cancellation path with multi-set on the same Alarm. I'll add a test.

drfloob: Added some failing tests, then fixed the implementation.

src/cpp/common/alarm.cc
drfloob: Fixed

src/cpp/common/alarm.cc
```diff
@@ -52,78 +62,94 @@ class AlarmImpl : public grpc::internal::CompletionQueueTag {
...
-                    grpc_core::Timestamp::FromTimespecRoundUp(deadline),
-                    &on_alarm_);
+    Ref();
+    GPR_ASSERT(cq_armed_.exchange(true) == false);
+    GPR_ASSERT(!callback_armed_.load());
+    cq_timer_handle_ = event_engine_->RunAfter(
+        grpc_core::Timestamp::FromTimespecRoundUp(deadline) -
+            grpc_core::ExecCtx::Get()->Now(),
+        [this] {
+          cq_armed_.store(false);
```
yijiem, line 77: Is it possible to move `cq_armed_.store(false);` into `OnCQAlarm()`? Same for `callback_armed_.store(false);`.

src/cpp/common/alarm.cc
```diff
@@ -52,78 +62,94 @@ class AlarmImpl : public grpc::internal::CompletionQueueTag {
...
-                                       },
-                                       arg, nullptr),
-                                   error);
-        },
-        this, grpc_schedule_on_exec_ctx);
-    grpc_timer_init(&timer_,
-                    grpc_core::Timestamp::FromTimespecRoundUp(deadline),
-                    &on_alarm_);
+    GPR_ASSERT(callback_armed_.exchange(true) == false);
+    GPR_ASSERT(!cq_armed_.load());
```
yijiem, line 87: So an alarm can be either `callback_armed_` or `cq_armed_`, but not both? Is this semantic documented somewhere? What will happen if we allow both?

yijiem: Oh, just saw the original code use only one set of `grpc_timer timer_;` and `grpc_closure on_alarm_;`, nvm then.

src/cpp/common/alarm.cc
drfloob: Good idea, done.

src/cpp/common/alarm.cc
drfloob: I added documentation to the public API. The multi-set capability wasn't documented either, but we do test for it, so I added a note on that as well.

> What will happen if we allow both?

In the previous implementation, the timer_ and on_alarm_ objects were shared between the CQ and Callback implementations. They would have been overwritten (or some other UB, I think, since the iomgr timer API is not clear on re-use) if both `Set` methods were used concurrently.

drfloob: In this PR, if both are used concurrently, an assert will fail. I prefer this to UB (likely missing events).

yijiem (approved):
LGTM

PR #34058, opened 2023-08-14
[EventEngine] Enable core end2end tests for the event_engine_listener experiment

drfloob:
These may be needed again to help debug remote build environment problems.

### DIFF
--- a/bazel/experiments.bzl
+++ b/bazel/experiments.bzl
@@ -25,6 +25,7 @@ EXPERIMENTS = {
                 "transport_supplies_client_latency",
             ],
             "core_end2end_test": [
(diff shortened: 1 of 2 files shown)


ctiller (approved):
Probably worth leaving them enabled just to avoid backsliding on implementation progress.

drfloob:
> Probably worth leaving them enabled just to avoid backsliding on implementation progress.

Agreed. This one, and the `event_engine_client` experiment, were disabled last week because tests were timing out (collectively, due to accumulation of tests and some extra slow tests), and PRs couldn't be merged because of it. I think we're out of the woods now. I'll send a PR to enable the client tests as well.

PR #34064, opened 2023-08-15
[fix] don't leak when a range times out unacked

yousukseung:


ctiller (approved):

Vignesh2208:
Can you add this to the event engine buffer list as well ?  https://github.com/grpc/grpc/blob/master/src/core/lib/event_engine/posix_engine/traced_buffer_list.cc

yousukseung:
> Can you add this to the event engine buffer list as well ? https://github.com/grpc/grpc/blob/master/src/core/lib/event_engine/posix_engine/traced_buffer_list.cc

Done.

Vignesh2208 (approved):

PR #34065, opened 2023-08-15
[EventEngine] Improve lock contention in WorkStealingThreadPool (alternative)

drfloob:
Proposed alternative to https://github.com/grpc/grpc/pull/34024.

This version has a simpler, faster busy-count implementation based on a sharded set of atomic counts: fast increment/decrement operations, relatively slower summation of total counts (which need to happen much less frequently).


src/core/lib/event_engine/thread_pool/thread_count.h
```diff
@@ -0,0 +1,95 @@
...
+namespace grpc_event_engine {
+namespace experimental {
+
+// Tracks counts across some fixed number of shards.
+// It is intended for fast increment/decrement operations, but a slower overall
+// count operation.
+class BusyThreadCount {
+ public:
+  // Increments a per-shard counter on construction, decrements on destruction.
+  class AutoThreadCounter {
```
ctiller, line 34: We should implement move constructors or mark them deleted here - it's not immediately clear that MakeAutoThreadCounter is sound.

src/core/lib/event_engine/thread_pool/thread_count.h
```diff
@@ -0,0 +1,95 @@
...
+#include "src/core/lib/gprpp/sync.h"
+#include "src/core/lib/gprpp/time.h"
+
+namespace grpc_event_engine {
+namespace experimental {
+
+// Tracks counts across some fixed number of shards.
+// It is intended for fast increment/decrement operations, but a slower overall
+// count operation.
+class BusyThreadCount {
```
ctiller, line 31: Lets add a test for this. I'd suggest start some 10s of threads, increment/decrement a few 10's of thousands of times, and finally assert count == 0.

src/core/lib/event_engine/thread_pool/thread_count.h
```diff
@@ -0,0 +1,95 @@
...
+  std::vector<ShardedData> shards_;
+  std::atomic<int> next_idx_{0};
+};
+
+// Tracks the number of living threads. It is intended for a fast count
+// operation, with relatively slower increment/decrement operations.
+class LivingThreadCount {
+ public:
+  // Increments the global counter on construction, decrements on destruction.
+  class AutoThreadCounter {
```
ctiller, line 67: Same comment re constructors as above

src/core/lib/event_engine/thread_pool/thread_count.h
```diff
@@ -0,0 +1,95 @@
...
+    ~AutoThreadCounter();
+
+   private:
+    LivingThreadCount* counter_;
+  };
+
+  AutoThreadCounter MakeAutoThreadCounter();
+  void Increment() ABSL_LOCKS_EXCLUDED(mu_);
+  void Decrement() ABSL_LOCKS_EXCLUDED(mu_);
+  void BlockUntilThreadCount(size_t desired_threads,
```
ctiller, line 79: This seems prime for a unit test.

src/core/lib/event_engine/thread_pool/thread_count.h
```diff
@@ -0,0 +1,95 @@
...
+                             grpc_core::Duration timeout, const char* why)
```
ctiller, line 80: What is the timeout useful for?

src/core/lib/event_engine/thread_pool/thread_count.h
```diff
@@ -0,0 +1,95 @@
...
+  // call to distribute load and reduce contention.
+  size_t NextIndex();
+
+ private:
+  struct ShardedData {
+    std::atomic<int> busy_count{0};
+  } GPR_ALIGN_STRUCT(GPR_CACHELINE_SIZE);
+
+  std::vector<ShardedData> shards_;
+  std::atomic<int> next_idx_{0};
```
ctiller, line 59: Better to make this unsigned, for defined behavior on overflow

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: This is only used for periodic logging, "Waiting for thread pool to idle ...". An example where it's been useful: we've seen fork tests hang with a large list of these logs, and the long log list helps identify where and how things got stuck.

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: ooh I thought I had, it only needs to be size_t. Good catch.

src/core/lib/event_engine/thread_pool/thread_count.h
ctiller: so move only would be completely valid, as would unmovable/uncopyable (although that might make `MakeAutoThreadCounter` not work?).

Move only needs an `if` in the destructor to deal with destroying a now uncounted object, unmovable does not - the compiler ought to be able to optimize that away though (in most cases).

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: Done

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: I see what you're saying now, we're looking at a double-decrement if we don't fix the moved-from type.

drfloob: Fixed.

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: Added. I also started a thread to concurrently call `count` while the inc/dec operations were happening.

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: Done

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: ... and it wasn't actually doing that. I removed the timeout argument so it waits forever, and log every 3 seconds.

src/core/lib/event_engine/thread_pool/thread_count.h
drfloob: Added

ctiller (approved):
One nit to try out

src/core/lib/event_engine/thread_pool/thread_count.cc
```diff
@@ -0,0 +1,164 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#include <grpc/support/port_platform.h>
+
+#include "src/core/lib/event_engine/thread_pool/thread_count.h"
+
+#include <inttypes.h>
+
+#include <atomic>
```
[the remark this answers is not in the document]
drfloob: Done. Very likely improvements on a couple of benchmarks, CV is +/- 2% for all but the Lambda Fanout 2/70 and 4/8.
image

PR #34075, opened 2023-08-16
[Debuggability][C++] Making channel args more debuggable 

tanvi-jagtap:
CC : @yousukseung , @ctiller 

Part 1 of 5:
Making channel args more debuggable by being able to check their value at runtime.
Adding a function to check the current value of channel args.

Part 2 of 5 : (Next PR)
ChannelArgs class changes.
GetChannelArgsDebugInfo() will also give the source code details of where a channel arg was set.

release notes: no


src/core/lib/channel/channel_args.cc
```diff
@@ -313,6 +313,17 @@ std::string ChannelArgs::ToString() const {
   return absl::StrCat("{", absl::StrJoin(arg_strings, ", "), "}");
 }
 
+std::vector<std::vector<std::string>> ChannelArgs::GetChannelArgsDebugInfo() const {
```
ctiller, line 316: Where will we be using the return value (in production code) - I'm curious about the choice of data structure here.

tanvi-jagtap: Q - Where will we be using the return value (in production code).
A - Hopefully nowhere . Ideally nowhere. We just want to use this function to tie with the ChannelZ work that so that at any given time the contents of channel args can be queried

Q - Choice of Data Structure
A - While I am not 100% happy with it, because just looking at it does not tell you what is inside it. Here are the reasons I chose it
1. vector and string are both std . Available to all who use open source.
2. Yousuk said that the output has to be iteratable.
3.  Vector is an OK wait to represent the channel args.
4.  But representing each channel arg with vector < string > is something that looks a little clumsy . Better ideas are welcome here.

tanvi-jagtap: I thought of a proto , but that would not be suitable for Google.
I thought maybe a new struct or class. But then we would have to convert it to a string somewhere in the code...

tanvi-jagtap: Did 1. 
ToString is already getting used in 2 places. And they are expecting it to behave like a string. This is still returning vector<ChannelArgs::DebugEntry>

PR #34077, opened 2023-08-16
[WorkSerializer] add debug-only IsRunningInWorkSerializer() method and use it in client_channel

markdroth:


src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -513,6 +513,9 @@ class ClientChannel::SubchannelWrapper : public SubchannelInterface {
         ++it->second;
       }
     }
+#ifndef NDEBUG
+    GPR_ASSERT(chand_->work_serializer_->RunningInWorkSerializer());
```
yashykt, line 517: we've got `GPR_DEBUG_ASSERT`

src/core/ext/filters/client_channel/client_channel.cc
```diff
@@ -522,6 +525,9 @@ class ClientChannel::SubchannelWrapper : public SubchannelInterface {
               "chand=%p: destroying subchannel wrapper %p for subchannel %p",
               chand_, this, subchannel_.get());
     }
+#ifndef NDEBUG
+    GPR_ASSERT(chand_->work_serializer_->RunningInWorkSerializer());
```
yashykt, line 529: `GPR_DEBUG_ASSERT`

src/core/lib/gprpp/work_serializer.cc
```diff
@@ -102,10 +112,14 @@ void WorkSerializer::WorkSerializerImpl::Run(std::function<void()> callback,
...
   if (GetOwners(prev_ref_pair) == 0) {
     // We took ownership of the WorkSerializer. Invoke callback and drain queue.
+#ifndef NDEBUG
+    current_thread_ = std::this_thread::get_id();
+#endif
     if (GRPC_TRACE_FLAG_ENABLED(grpc_work_serializer_trace)) {
       gpr_log(GPR_INFO, "  Executing immediately");
     }
     callback();
+    callback = nullptr;  // Delete while still holding WorkSerializer.
```
yashykt, line 122: ah I see! Can you expand the comment to say something like - The callback might be holding refs that need to be given up from within the WorkSerializer

yashykt (approved):

src/core/ext/filters/client_channel/client_channel.cc
markdroth: Done.

src/core/ext/filters/client_channel/client_channel.cc
markdroth: Done.

src/core/lib/gprpp/work_serializer.cc
markdroth: Done.

PR #34080, opened 2023-08-16
[benchmarks] Disable dotnet and golang benchmarks

drfloob:
This should get the benchmarks running again. The dotnet benchmark is broken (unclear if it's still necessary), and the grpc-go benchmark build currently fails. The go benchmark should be re-enabled when the dockerfiles are fixed. The rest of the dotnet benchmark configuration / artifacts should be deleted or fixed as well. @jtattermusch


tools/internal_ci/linux/grpc_e2e_performance_gke.sh
```diff
@@ -130,16 +130,6 @@ configLangArgs8core+=( -l c++ )
 configLangArgs32core+=( -l c++ )
 runnerLangArgs+=( -l "cxx:${GRPC_CORE_REPO}:${GRPC_CORE_COMMIT}" )
 
-# dotnet
```
paulosjca, line 133: Please comment out instead of removing. That way we remember what has been disabled and could be reenabled or removed.

drfloob: Done

paulosjca (approved):

PR #34085, opened 2023-08-17
[clang-cl] Fix compile error found by clang-cl with strict warnings

yijiem:
This fix the following compile error using clang-cl with strict warnings on Windows:

```
test/core/util/build.cc(15,6): error: no previous prototype for function 'BuiltUnderValgrind' [-Werror,-Wmissing-prototypes]
bool BuiltUnderValgrind() {
     ^
test/core/util/build.cc(15,1): note: declare 'static' if the function is not intended to be used outside of this translation unit
bool BuiltUnderValgrind() {
^
static
test/core/util/build.cc(23,6): error: no previous prototype for function 'BuiltUnderTsan' [-Werror,-Wmissing-prototypes]
bool BuiltUnderTsan() {
     ^
test/core/util/build.cc(23,1): note: declare 'static' if the function is not intended to be used outside of this translation unit
bool BuiltUnderTsan() {
^
static
test/core/util/build.cc(39,6): error: no previous prototype for function 'BuiltUnderAsan' [-Werror,-Wmissing-prototypes]
bool BuiltUnderAsan() {
     ^
test/core/util/build.cc(39,1): note: declare 'static' if the function is not intended to be used outside of this translation unit
bool BuiltUnderAsan() {
^
static
test/core/util/build.cc(55,6): error: no previous prototype for function 'BuiltUnderMsan' [-Werror,-Wmissing-prototypes]
bool BuiltUnderMsan() {
     ^
test/core/util/build.cc(55,1): note: declare 'static' if the function is not intended to be used outside of this translation unit
bool BuiltUnderMsan() {
^
static
test/core/util/build.cc(71,6): error: no previous prototype for function 'BuiltUnderUbsan' [-Werror,-Wmissing-prototypes]
bool BuiltUnderUbsan() {
     ^
test/core/util/build.cc(71,1): note: declare 'static' if the function is not intended to be used outside of this translation unit
bool BuiltUnderUbsan() {
^
static
5 errors generated.
```

### DIFF
--- a/test/core/util/build.cc
+++ b/test/core/util/build.cc
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "test/core/util/build.h"
+
 bool BuiltUnderValgrind() {
 #ifdef RUNNING_ON_VALGRIND
   return true;


drfloob:
Usually, the path to fixing a new compiler warning involves adding it to the list of warnings that the build cares about, to avoid regressions https://github.com/grpc/grpc/blob/c21e3f18676f53e562ae28b31ba88e27bf0a79d1/bazel/copts.bzl#L19-L22. I'd suggest doing that here.

> It looks like the actual warning flags that the MSVC-compatible clang-cl.exe's -Wall and -Wextra enable are different

Without support for nested selects in starlark/bazel, I think we'd have to do some work on the build system to have an equivalent set of copts for posix & windows, but I only looked at it briefly before. It would be nice if the clang-cl flags could be a superset of the posix ones, possibly with things like `-Wno-missing-prototypes` for default checks we want to exclude.

drfloob (approved):
I'd suggest at least a TODO in copts.bzl to enable this check, but this improvement is good to land.

PR #34086, opened 2023-08-17
ChannelArgsSrc : Adding location to Value constructor

tanvi-jagtap:

### DIFF
--- a/src/core/lib/channel/channel_args.h
+++ b/src/core/lib/channel/channel_args.h
@@ -368,10 +368,12 @@ class ChannelArgs {
 
   class Value {
    public:
-    explicit Value(int n) : rep_(reinterpret_cast<void*>(n), &int_vtable_) {}
-    explicit Value(std::string s)
-        : rep_(RcString::Make(s).release(), &string_vtable_) {}
-    explicit Value(Pointer p) : rep_(std::move(p)) {}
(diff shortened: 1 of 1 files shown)


src/core/lib/channel/channel_args.h
```diff
@@ -401,10 +410,17 @@ class ChannelArgs {
...
+      std::string location_str(location.file());
+      location_str += ":";
+      location_str += std::to_string(location.line());
+      return location_str;
+    }
     static const grpc_arg_pointer_vtable int_vtable_;
     static const grpc_arg_pointer_vtable string_vtable_;
 
     Pointer rep_;
+    std::string location_;
```
ctiller, line 423: We should just store a SourceLocation here directly (it's much cheaper)

src/core/lib/channel/channel_args.h
```diff
@@ -389,6 +394,10 @@ class ChannelArgs {
 
     std::string ToString() const;
 
+    // TODO(tjagtap) : Make this available in the
+    // GetChannelArgsDebugInfo() function after rebasing.
+    std::string GetSourceLocationString() { return location_; };
```
ctiller, line 399: Just return SourceLocation here too... materialize the string only at the last moment

ctiller:
Approach lgtm, looks like there's a test failure left.

PR #34088, opened 2023-08-17
[bazel] Upgrade bazel repo-wide to 6.3.2

jtattermusch:
- Upgrade bazel
- Reduce the number of places where bazel version needs to be upgraded in future.
- also make sure the list of bazel versions to test by bazelified tests is loaded from supported_versions.txt (it was hardcoded before).
- Try upgrading windows RBE build to bazel 6.3.2 as well.

The core idea:
- the source of truth for supported bazel versions is in `bazel/supported_versions.txt`
- the first version listed in `bazel/supported_versions.txt` is considered to be the "primary" bazel version and is going to be used in most places thoroughout the repo.
- use templates to include the primary bazel version in testing dockerfiles and in a newly introduced `.bazelversion` files (which gets loaded by our existing `tools/bazel` wrapper).


jtattermusch:
Issue with windows RBE:
https://source.cloud.google.com/results/invocations/6a0bc4fb-be9d-4aff-b9c2-123893b95fdb
https://source.cloud.google.com/results/invocations/6a6f3201-c239-46f9-80b6-14d4ef5f6a1c

jtattermusch:
Since windows RBE still doesn't work after upgrading to bazel 6.3.2, I'm keeping the workaround for win RBE in place and I'll look into upgrading bazel for windows RBE separately.

veblush (approved):

davidben:
> Since windows RBE still doesn't work after upgrading to bazel 6.3.2, I'm keeping the workaround for win RBE in place and I'll look into upgrading bazel for windows RBE separately.

I'm not sure why the new vs. old bazel are different here, but that error reads as if you're not linking in the Windows socket libraries. Is it possible the issue is just that third_party/address_sorting/BUILD doesn't pull in ws2_32.lib? I noticed grpc_cc_library from bazel/grpc_build_system.bzl pulls it in, but that file's using something else.

jtattermusch:
> > Since windows RBE still doesn't work after upgrading to bazel 6.3.2, I'm keeping the workaround for win RBE in place and I'll look into upgrading bazel for windows RBE separately.
> 
> I'm not sure why the new vs. old bazel are different here, but that error reads as if you're not linking in the Windows socket libraries. Is it possible the issue is just that third_party/address_sorting/BUILD doesn't pull in ws2_32.lib? I noticed grpc_cc_library from bazel/grpc_build_system.bzl pulls it in, but that file's using something else.

Thanks! 
FYI @apolcyn 

@veblush  we can investigate this suspicion once we add the windows DLL support (since we're going to need to make sure we can dynamically anyway).

PR #34096, opened 2023-08-17
[PSM Interop] Return trailing metadata.

eugeneo:
1. Trailing metadata is now reported
2. messages.proto was synchronized.
3. Corrected order of arguments in EXPECT_EQ so the output makes sense now.


src/proto/grpc/testing/messages.proto
```diff
@@ -309,3 +317,23 @@ message TestOrcaReport {
   map<string, double> request_cost = 3;
   map<string, double> utilization = 4;
 }
+
+enum HookRequestCommand {
```
yashykt, line 321: I don't see these being used

eugeneo: They are from that other PR. This file source of truth is https://github.com/grpc/grpc-proto/blob/master/grpc/testing/messages.proto that already has the messages.

yashykt: can you merge master then please?

eugeneo: Rebased on top of "hook server" change, should be easier to follow now.

yashykt (approved):

PR #34098, opened 2023-08-17
[pick_first] fix test flake

markdroth:
CNR the flake, but I've changed the test (which is very old) to use some of our more modern helper functions that have saner timeouts.

Also re-add a `return` statement that was accidentally removed in #33753, which I noticed while working on this.  Its absence doesn't cause a real problem, but it does cause us to needlessly trigger a duplicate connection attempt or report a duplicate CONNECTING update in some cases.

### DIFF
--- a/test/cpp/end2end/client_lb_end2end_test.cc
+++ b/test/cpp/end2end/client_lb_end2end_test.cc
@@ -830,32 +830,38 @@ TEST_F(PickFirstTest, BackOffInitialReconnect) {
   args.SetInt(GRPC_ARG_INITIAL_RECONNECT_BACKOFF_MS,
               kInitialBackOffMs * grpc_test_slowdown_factor());
   const std::vector<int> ports = {grpc_pick_unused_port_or_die()};
-  const gpr_timespec t0 = gpr_now(GPR_CLOCK_MONOTONIC);
   auto response_generator = BuildResolverResponseGenerator();
   auto channel = BuildChannel("pick_first", response_generator, args);
   auto stub = BuildStub(channel);
   response_generator.SetNextResolution(ports);
-  // The channel won't become connected (there's no server).
-  ASSERT_FALSE(channel->WaitForConnected(
-      grpc_timeout_milliseconds_to_deadline(kInitialBackOffMs * 2)));
+  // Start trying to connect.  The channel will report
+  // TRANSIENT_FAILURE, because the server is not reachable.
+  const grpc_core::Timestamp t0 = grpc_core::Timestamp::Now();
+  ASSERT_TRUE(WaitForChannelState(
+      channel.get(),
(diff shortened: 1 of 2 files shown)


test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -830,32 +830,38 @@ TEST_F(PickFirstTest, BackOffInitialReconnect) {
...
-  ASSERT_TRUE(channel->WaitForConnected(
-      grpc_timeout_milliseconds_to_deadline(kInitialBackOffMs * 2)));
-  const gpr_timespec t1 = gpr_now(GPR_CLOCK_MONOTONIC);
-  const grpc_core::Duration waited =
-      grpc_core::Duration::FromTimespec(gpr_time_sub(t1, t0));
+  // Now the channel will become connected.
+  ASSERT_TRUE(WaitForChannelReady(channel.get()));
+  // Check how long it took.
+  const grpc_core::Timestamp t1 = grpc_core::Timestamp::Now();
+  const grpc_core::Duration waited = t1 - t0;
```
drfloob, line 856: nit: since t1 is only used once, you can delete the variable and add the `Now` call in the `waited` definition.

test/cpp/end2end/client_lb_end2end_test.cc
```diff
@@ -830,32 +830,38 @@ TEST_F(PickFirstTest, BackOffInitialReconnect) {
...
   // account for test and precision accuracy drift.
   EXPECT_GE(waited.millis(),
             (kInitialBackOffMs * grpc_test_slowdown_factor()) - 1);
   // But not much more.
-  EXPECT_GT(
-      gpr_time_cmp(
-          grpc_timeout_milliseconds_to_deadline(kInitialBackOffMs * 1.10), t1),
-      0);
+  EXPECT_LE(waited.millis(),
+            (kInitialBackOffMs * grpc_test_slowdown_factor()) * 1.1);
```
drfloob, line 864: nit: suggest declaring a `constexpr int` variable for the magic number here.

drfloob (approved):

test/cpp/end2end/client_lb_end2end_test.cc
markdroth: Done.

test/cpp/end2end/client_lb_end2end_test.cc
markdroth: I'd prefer to leave this as-is, since I want the tests to be clear about where they're using `grpc_test_slowdown_factor()`.  There are some places we plugin the raw timeout value because some other utiluty code is applying the slowdown factor, and there are other places where we apply the slowdown factor ourselves, and it can often be hard to tell where it's actually being applied, so I want to err on the side of clarity here.

PR #34102, opened 2023-08-18
[bazelified tests] Shard bazelified bazel distribtests to make them faster

jtattermusch:
Basically run each of the subtests (buildtest, distribtest_cpp, distribtest_python) as a separate bazel target.

- currently the bazel distribtest are the slowest targets in grpc_bazel_rbe_nonbazel
- the shards are basically independent tests anyway
- when split into multiple targets, they each get a separate target log so it's easier debug issues since there isn't multiple bazel invocations in each log.

### DIFF
--- a/test/distrib/bazel/test_single_bazel_version.sh
+++ b/test/distrib/bazel/test_single_bazel_version.sh
@@ -15,12 +15,26 @@
 
 set -ex
 
-if [ "$#" != "1" ] ; then
+if [ "$#" == "0" ] ; then
     echo "Must supply bazel version to be tested." >/dev/stderr
     exit 1
 fi
 
 VERSION="$1"
+shift 1
+
+# Accept the list of test shards to run on the commandline
+TEST_SHARDS=("$@")
+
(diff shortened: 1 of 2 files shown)


test/distrib/bazel/test_single_bazel_version.sh
```diff
@@ -15,12 +15,26 @@
...
+
+# Accept the list of test shards to run on the commandline
+TEST_SHARDS=("$@")
+
+# If list of shards to run hasn't been specified on the command line,
+# run all the known shards.
+if [ "${#TEST_SHARDS[@]}" == "0" ] ; then
+  TEST_SHARDS=(
+    buildtest
+    distribtest_cpp
```
gnossen, line 34: Seems like these defaults should somehow be kept synchronized with the `TEST_DIRECTORIES` variable below.

gnossen (approved):

test/distrib/bazel/test_single_bazel_version.sh
jtattermusch: done.

PR #34103, opened 2023-08-18
[Build] Shared builds on windows

h-vetinari:
Towards https://github.com/grpc/grpc/issues/33032,

Reopen after botched force-push in #33175 that then got "merged" and cannot be reopened anymore.

More context in that PR.


h-vetinari:
@dawidcha, please be more careful with force-pushes (and with adding spurious whitespace in your commits).

dawidcha:
> @dawidcha, please be more careful with force-pushes (and with adding spurious whitespace in your commits).

Sorry, bad Sync Fork operation. Thanks for fixing.

The whitespace you refer to - is that in CMakeLists.txt? It's there because this is generated from CMakeLists.txt.template - it was simpler/cleaner to write the template in a way that added a bit of trailing whitespace in a few places. I can change that if it's important. In any case, we should probably regenerate CMakeLists.txt, rather than merging changes.

@veblush can you kick the tests off, please.

h-vetinari:
Spurious whitespace should never be _necessary_. I fixed it while rebasing, but you should still see it if you look at the last commit of the pre-rebase state.

In any case, my rebase was pretty mechanical, so it's possible I messed something up (much less regenerate correctly from the template). Feel free to fix where you want, including fixing existing commits here. Just be careful to not reset to in-flight PRs, which is how we permanently closed the previous PR (I don't care about that per se, but it's annoying to split the context and discussion across several PRs, so best to avoid that).

dawidcha:
Update:
The build now seems pretty consistent:
- both shared and static builds are running to success - DLLs and import libraries are emitted for all libs
- no edits to code that is not part of the gRPC repo
- CMakeLists.txt.template has been updated so that it generates the CMakeLists.txt that is used in the build

I'm currently working on getting the tests to build (with gRPC_BUILD_TESTS=ON)

dawidcha:
@veblush @h-vetinari 

This code is now in a state where I've shown it will compile on Windows to both static and dynamic libraries. I saw that there were a few test failures - one was because I'd committed an old CMakeLists.txt.template - this is now fixed, but I didn't twig that committing would blow away the previous test failure details, so please kick off the tests once more so I can continue to fix tests.

The code as it stands at this point does not build in dynamic mode with gRPC_BUILD_TESTS=ON. I have some fixes for that in the pipeline, but they touch about 25 source files, so I think it would be simpler to review and rebase/merge in two stages.

wdyt?

h-vetinari:
Cool progress @dawidcha! 👍 

Can't help with the tests here (or the merge strategy), that's up to the grpc devs. Thanks for digging into this! 🙏

dawidcha:
rebased and ready to go pending passing tests @veblush can you kick off the tests?

h-vetinari:
This is a pretty bad rebase, it contains lots of unnecessary commits. I'll try to clean it up in a couple hours

h-vetinari:
OK, I rebased this twice. First to get rid of all the duplicate commits you somehow managed to add, and then a second time to rebase on the master branch. To ensure I didn't get the first rebase wrong, I pushed a branch `before_rebase` (on master) that has an empty [diff](https://github.com/h-vetinari/grpc/compare/c494c19b8389787adb09a48f383a7f7f987286c9..before_rebase) with your last commit (https://github.com/grpc/grpc/pull/34103/commits/c494c19b8389787adb09a48f383a7f7f987286c9) before you merged master.

Then I rebased (cleanly) on master. In general, before using words like "rebased", or attempting to do so, you should double-check before pushing (and then triple-check in the GH UI). I recommend trying to get familiar with `git rebase -i <hash>`, and how to do squashes/fix-ups/edits in that context.

dawidcha:
One test failure to go. If I've understood, this is a Linux test for memory leaks, and it's timing out after 5 mins. Log has some warnings, but no smoking gun afaics. @veblush Is this a known/transient failure? Does it make sense to increase the timeout, or is this a genuine catch - perhaps caused by moving some code out of headers and into .c/.cc files?

veblush:
@dawidcha You're right, Two tests are not relevant to this PR. I think this is now ready to be reviewed.

veblush:
While runnign the full benchmark suite internally, I checked the result of [micro benchmark](https://github.com/grpc/grpc/pull/34103/checks?check_run_id=16290211695) 

```
[microbenchmarks] Performance differences noted:
Benchmark                   cpu_time    real_time
--------------------------  ----------  -----------
BM_ClosureSched3OnCombiner  +8%         +8%
BM_PollAddFd                +9%         +9%
```

And +8% or +9% looks significant so it needs to be addressed. So we may want to do some hacky approach here to keep inline function work on non-Windows and support DLL on Windows at the same time.

dawidcha:
I've effectively reverted all changes that had a substantive effect on the code being compiled except when building a Windows DLL:
* Necessary adjustments to CMakeLists.txt only have effect when MSVC and BUILD_SHARED_LIBS are both set.
* GPR_DLL/GRPC_DLL/GRPCXX_DLL macros all define to empty string except for Windows DLL builds
* A couple of changes (to fork.cc/fork.h and tls_certificate_provider.cc/tls_certificate_provider.h were found to be unnecessary
* Changes to exec_ctx.cc/exec_ctx.h required because of limitations in the use of thread_local in DLLs are only applied (via conditional compilation) when both _WIN32 and _DLL macros are defined (i.e. when building a Windows DLL)

@veblush can you kick off tests and see if that has fixed the apparent performance degradation.

h-vetinari:
Rebased after merge of https://github.com/grpc/grpc/pull/34199 (removed all upb things from history). Also solved CLA issue by using consistent email address for @dawidcha

dawidcha:
@h-vetinari Thanks for your continued help with the git history tidy

veblush (approved):
Internal benchmark result doesn't appear to have any regression. So it's now to get merged! Thanks!

veblush:
@jtattermusch Would you do the second review?

dawidcha:
Bazel Basic Tests for Python failure:


```text
======================================================================
FAIL: test_cancelled_watch_removed_from_watch_list (__main__.HealthServicerTest)
----------------------------------------------------------------------
  File "/root/.cache/bazel/_bazel_root/954bb7512d44d20015390af6e76121c6/sandbox/processwrapper-sandbox/2484/execroot/com_github_grpc_grpc/bazel-out/k8-fastbuild/bin/src/python/grpcio_tests/tests_aio/health_check/health_servicer_test.runfiles/com_github_grpc_grpc/src/python/grpcio_tests/tests_aio/unit/_test_base.py", line 31, in wrapper
    return loop.run_until_complete(f(*args, **kwargs))
  File "/usr/local/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/root/.cache/bazel/_bazel_root/954bb7512d44d20015390af6e76121c6/sandbox/processwrapper-sandbox/2484/execroot/com_github_grpc_grpc/bazel-out/k8-fastbuild/bin/src/python/grpcio_tests/tests_aio/health_check/health_servicer_test.runfiles/com_github_grpc_grpc/src/python/grpcio_tests/tests_aio/health_check/health_servicer_test.py", line 246, in test_cancelled_watch_removed_from_watch_list
    self.assertTrue(queue.empty())
AssertionError: False is not true
```
Is this a transient failure?

veblush:
> Bazel Basic Tests for Python failure:
> ...
> Is this a transient failure?

It looks like an irrelevant failure. Let me rerun the test.

include/grpc/support/port_platform.h
```diff
@@ -53,6 +53,57 @@
...
+// Warning: shared library support for Windows (i.e. producing DLL plus import
+//   library instead of a static library) is experimental. Some symbols that can
+//   be linked using the static library may not be available when using the
+//   dynamically linked library.
+//
+// Note: GRPC_DLL_EXPORTS is set in CMakeLists.txt when building shared
+// grpc{,_unsecure}
+//       GRPC_DLL_IMPORTS is set by us as part of the interface for consumers of
+//       the DLL
+#if !defined(GRPC_DLL)
```
jtattermusch, line 75: note that the GRPC_DLL GRPCXX_DLL and GPR_DLL defines depend on the assumption that the only 3  "public" libraries known to us are gpr, grpc and grpc++. In the future, we are likely planning to make the layout of libraries in cmake more akin to what's in grpc's bazel BUILD (that has much more granular targets) at which point it won't be feasible to hand list all the known public libraries here?

I'll leave the decision whether the defines make sense as-is to @veblush  but I wanted to point out this potential problem.

CMakeLists.txt
```diff
@@ -421,8 +421,42 @@ endif()
...
+# are generated from protobuf definitions by upbc. This source code does not include
+# annotations needed to export these functions from gprc.lib so we have to
+# re-include a small subset of these.
+#
+# This is not an ideal situation because these functions will be unavailable
+# to clients of grpc and the libraries that need this (e.g. grpc++) will
+# include redundant duplicate code. Hence, the duplication is only activated
+# for DLL builds - and should be completely removed when source files are
+# generated with the necessary __declspec annotations.
+set(gRPC_UPB_GEN_DUPL_SRC
```
jtattermusch, line 439: ok, patching the build with gRPC_UPB_GEN_DUPL_SRC and gRPC_ADDITIONAL_WIN_DLL_SRC seems very dirty so we should really have a clear plan how to handle these in the future cleanly, otherwise we're introducing some serious and hard-to-understand technical debt.

If there's a clear (and commited) plan how to get rid of these hacks in the future I might be ok with having there temporarily, but otherwise this is difficult to accept.

templates/CMakeLists.txt.template
```diff
@@ -150,6 +150,20 @@
...
 
+  def lib_type_for_lib(lib_name):
+    """Returns STATIC/SHARED to force a static or shared lib build depending on the library."""
+    # grpc_csharp_ext is loaded by C# runtime and it
+    # only makes sense as a shared lib.
+    if lib_name in ['grpc_csharp_ext']:
+      return ' SHARED'
+
+    # upb always compiles as a static library on Windows
+    elif lib_name in ['upb','upb_collections_lib','upb_json_lib','upb_textformat_lib']:
```
jtattermusch, line 161: note that in the future we'll need a better way of telling which are the "upb" libraries than just listing them in multiple places in the CmakeLists.txt.tempate.
FWIW, there should be single place that defines sth like `is_upb_library` and use that from elsewhere. Unfortunately, since our upb cmake dependency is in flux, we're not there yet.

templates/CMakeLists.txt.template
```diff
@@ -759,6 +820,17 @@
     set_target_properties(${lib.name} PROPERTIES COMPILE_PDB_NAME "${lib.name}"
       COMPILE_PDB_OUTPUT_DIRECTORY <%text>"${CMAKE_BINARY_DIR}</%text>"
     )
+    % if lib.name == 'gpr':
```
jtattermusch, line 823: to improve our cmake build, we're trying to get fewer reference to libraries by their name, not more. So in my opinion this change introduces technical debt that will need to be dealt with later (so not an ideal outcome).

Basically deciding whether a library should have some special settings assigned to it should be based on library's metadata (e.g. is the a "base" grpc public library), not by listing libraries that should have these setting by their names.

Unfortunately, the refactoring of our cmake build to extrapolate metadata like this (instead of relying of knowing the names of individual targets) is only in progress and definitely needs much more work.  My worries is that making changes like the once suggested in this PR, we'll make the subsequent refactoring even harder.

include/grpc/support/port_platform.h
jtattermusch: at the same time, I realize that because of the import / export semantics in windows DLLs, there probably needs to be a separate define for each library, so this is tricky to resolve.
What does absl do here (since AFAIK absl consists of many shared libraries)?

jtattermusch (commented):
Thanks for the contribution.  Unfortunately I have very mixed feelings about this.
I generally feel like our cmake build isn't in the right shape to be able to accept this change:
- the upb inclusion is our build is in flux.
- ongoing effort to simplify our cmake build and the process of extracting build metadata from bazel build  (which collides with some of the changes made in this PR).

Also, there are some questionable hacks in this PR and I don't see a clear plan for addressing them in the future.

One more thing I don't understand is how you chose the symbols for which you needed to add the *_DLL annotation (trial and error?) and why is there seemingly so few places where it was needed (I thought all the public symbols need to be exported). How do we know that there are really all the symbols that needed to be annotated? How will we ensure that in the future we'll know that we need to add annotations to newly introduces symbols? (sounds like we need a distribtest for this?)

jtattermusch:
> Thanks for the contribution. Unfortunately I have very mixed feelings about this. I generally feel like our cmake build isn't in the right shape to be able to accept this change:
> 
> * the upb inclusion is our build is in flux.
> * ongoing effort to simplify our cmake build and the process of extracting build metadata from bazel build  (which collides with some of the changes made in this PR).
> 
> Also, there are some questionable hacks in this PR and I don't see a clear plan for addressing them in the future.
> 
> One more thing I don't understand is how you chose the symbols for which you needed to add the *_DLL annotation (trial and error?) and why is there seemingly so few places where it was needed (I thought all the public symbols need to be exported). How do we know that there are really all the symbols that needed to be annotated? How will we ensure that in the future we'll know that we need to add annotations to newly introduces symbols? (sounds like we need a distribtest for this?)

Maybe I missed it, but is the PR intended to be used after CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS is enabled? If so, where is that mentioned?

include/grpc/support/port_platform.h
dawidcha: > at the same time, I realize that because of the import / export semantics in windows DLLs, there probably needs to be a separate define for each library, so this is tricky to resolve. What does absl do here (since AFAIK absl consists of many shared libraries)?

Abseil uses a single pair of macros for import/export determination ABSL_BUILD_DLL & ABSL_CONSUME_DLL

The abseil shared build emits a single DLL: abseil_dll.dll with an import library of abseil_dll.lib. It also emits other more specific import libraries, but I believe that these still reference the same DLL

CMakeLists.txt
dawidcha: > ok, patching the build with gRPC_UPB_GEN_DUPL_SRC and gRPC_ADDITIONAL_WIN_DLL_SRC seems very dirty so we should really have a clear plan how to handle these in the future cleanly, otherwise we're introducing some serious and hard-to-understand technical debt.
> 
> If there's a clear (and commited) plan how to get rid of these hacks in the future I might be ok with having there temporarily, but otherwise this is difficult to accept.

Agreed - and it gets worse when we get the tests to build as well.

Question: is the upb-generated and upbdefs-generated code part of the public API?

As things stand, there's no way of annotating these generated files with dllimport/dllexport tags as the capability to do that in upb doesn't seem to exist (though it does in protobuf), so they are not currently exposed - hence the need to duplicate. 

I think the right solution is to factor out all this generated code from being included directly into grpc and into a separate always-static library, say upb_generated. The existing grpc library then gets renamed to something like grpc_base and the grpc definition simply includes grpc_base and the definition for grpc simply includes upb_generated and grpc_base - yielding a library with the exact same contents as the existing grpc, but allowing the duplicated source to be avoided by having those libs that need it because they depend on a grpc in which the symbols are not visible, to depend on grpc_base instead and mix in the symbols at the 'leaf' level.

I think that would work provided the upb-generated symbols are not part of the public API.

templates/CMakeLists.txt.template
dawidcha: I can add that function. Would the following predicate suffice?
```text
lib_name == 'upb' or lib_name.startswith('upb_')
```

templates/CMakeLists.txt.template
dawidcha: Yes, I get it - I did look at finding a way to add that to the metadata generated from bazel, but didn't want to move before agreeing a direction. Is there a branch with the cmake build refactoring in progress? I can take at contributing if it would help.

dawidcha:
> Maybe I missed it, but is the PR intended to be used after CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS is enabled? If so, where is that mentioned?

I added this setting on the cmake command line - it is true that it would probably be better to see if I can default it appropriately based on BUILD_SHARED_LIBS and MSVC settings - I'll do that.

dawidcha:
> Thanks for the contribution. Unfortunately I have very mixed feelings about this. I generally feel like our cmake build isn't in the right shape to be able to accept this change:
> 
> * the upb inclusion is our build is in flux.
> * ongoing effort to simplify our cmake build and the process of extracting build metadata from bazel build  (which collides with some of the changes made in this PR).
> 
> Also, there are some questionable hacks in this PR and I don't see a clear plan for addressing them in the future.
> 
> One more thing I don't understand is how you chose the symbols for which you needed to add the *_DLL annotation (trial and error?) and why is there seemingly so few places where it was needed (I thought all the public symbols need to be exported). How do we know that there are really all the symbols that needed to be annotated? How will we ensure that in the future we'll know that we need to add annotations to newly introduces symbols? (sounds like we need a distribtest for this?)

Thanks for the review and I do understand the mixed feelings, but hope I can address them satisfactorily.

The goal of this PR is to get the first stake in the ground - a build that is capable of emitting a shared library for Windows, while not affecting existing static Windows, Linux and other builds in any way.

In this capacity the goal was to establish the XXX_DLL macros which implement the capability for the generation of import/export symbols for a Windows shared library. So yes, in this first push the placement of the actual annotations was determined based on what was needed to get this first cut to compile.

I have some more changes in the pipeline which should annotation more functions and classes such that the tests all build correctly, and a third tranche which annotate some symbols that are needed by code we are working on and which prove that a functional DLL can be produced.

I expect further tranches will be necessary to expose an MVP set of functionality.

The thought process behind doing things in stages like this is simply that the character of these changes involves small changes to a lot of source files which is likely to be very difficult to merge given the likely time taken to rebase retest and approve.

I hope this approach makes sense and that we can make positive progress on all the concerns you raised.

include/grpc/support/port_platform.h
h-vetinari: > I realize that because of the import / export semantics in windows DLLs, there probably needs to be a separate define for each library, so this is tricky to resolve.
> What does absl do here (since AFAIK absl consists of many shared libraries)?

I believe the pain with having a separate symbol per library is why abseil essentially creates one big DLL (with consequently only one macro switching between dllimport & dllexport), rather than handling this on a per-library level.

grpc's current situation with 3 libs is perhaps still acceptable for a manual setup, but if there are going to be more, then the shared setup on windows will likely have to bunch them together.

Note that this is a "lesser-of-two-evils" situation, because the build divergence between `unix || static` and `win && shared` causes quite a bit of hassle, both for the build itself, but also for the pkgconfig & CMake metadata.

CC @derekmauro

dawidcha:
@jtattermusch I see that you've included additional 'hardcoded' metadata in 'extract_metadata_from_bazel_xml.py'. I could add some fields to that to avoid having to have logic in the template that checks for specific target names. However, I also read that the long term plan is to avoid hand-coded metadata: 'gradually get rid of the legacy "build.yaml"-specific fields one by one'. What is the plan on how to do that?

CMakeLists.txt
veblush: Future plan is to have a DLL suport from upb and I'm following up on this but it will take sometime as upb needs to create their cmakefile first. Once upb has a proper cmake with DLL support, we'll be able to remove this hack. Until we have it, I'm fine with this workaround.

dawidcha: > Future plan is to have a DLL suport from upb and I'm following up on this but it will take sometime as upb needs to create their cmakefile first. Once upb has a proper cmake with DLL support, we'll be able to remove this hack. Until we have it, I'm fine with this workaround.

@veblush @jtattermusch @h-vetinari Would it be possible to arrange a zoom call to discuss options on this? I can see that upb is in flux (I see it just got moved to be a direct part of protobuf from being a repo in its own right), but I don't believe that the proposed changes in this PR add significant extra work with respect to adapting to that change. I'm also not seeing a future in which it won't be necessary at some level to provide additional metadata above and beyond what can be gleaned from the Bazel BUILD file, though I totally agree that such knowledge does not belong in the CMakeLists template file.

include/grpc/support/port_platform.h
veblush: I think this is acceptable as we don't want to have many DLLs even after refactoring the cmake build to have 1:1 corresponding targets to bazel's.

CMakeLists.txt
veblush: Currently I'm chasing upb to have DLL support properly and desirable state is to have proper DLL support so that gRPC can get rid of this hack. Until then, I think we can live with this.

templates/CMakeLists.txt.template
veblush: I have a different opinion on this. As this is considered experimental, I'd like to keep this change within CMakeLists.txt file so that others are not changed. Then it's easy to understand to follow what hacks are in place.

templates/CMakeLists.txt.template
```diff
@@ -486,8 +507,42 @@
...
+  # for DLL builds - and should be completely removed when source files are
+  # generated with the necessary __declspec annotations.
+  set(gRPC_UPB_GEN_DUPL_SRC
+    src/core/ext/upb-generated/src/proto/grpc/gcp/altscontext.upb.c
+    src/core/ext/upb-generated/src/proto/grpc/health/v1/health.upb.c
+    src/core/ext/upb-generated/src/proto/grpc/gcp/transport_security_common.upb.c
+  )
+
+  # When building shared lib on Windows, additional missing symbols need to
+  # be resolved for grpc++_unsecure. Also that build needs to be able to resolve
```
veblush, line 532: Would you elaborate this problem? I think something is wrong if unsecure has to be linked with openssl.

veblush:
> 

Thank you for the summary. I'd rather want to do within `CMakeLists.txt` exclusively so that other BUILD pipeline isn't affected as a temporary approach. Once we have all the pieces such as upb DLL support and such, we can have a better idea on how to generalize this.

templates/CMakeLists.txt.template
h-vetinari: Even though @dawidcha has improved the implementation, I believe this goes back to something I did originally. As I tried to switch our arrow builds in conda-forge to depend on the shared builds of grpc, there was an experimental symbol (`grpc::experimental::StaticDataCertificateProvider`) that could not get found anymore.

When [trying](https://github.com/conda-forge/grpc-cpp-feedstock/pull/303) to add `GRPCXX_DLL` to that class, the build failed because it was relying on symbols that were only present in `grpc++` (rather than `grpc++_unsecure`). I didn't analyse this further as it didn't make a big difference on our side, but I guess the `StaticDataCertificateProvider` should be fully in `grpc++`?

dawidcha: The reason this is happening is that there is a dependency on StaticDataCertificateProvider which is ultimately resolved by including the file src\core\lib\security\credentials\tls\grpc_tls_certificate_provider.cc. This source file includes a function PrivateKeyAndCertificateMatch which has a dependency on openssl, hence the need to link it to resolve. For static libraries, I assume the PrivateKeyAndCertificateMatch function is being excluded by the linker because it isn't being used - the DLL can't do that because as a runtime artifact the linker can't know whether or not the function will be used, so has to include it.

I believe that it should be possible to get round this by moving the PrivateKeyAndCertificateMatch function (and maybe a few others) into a separate compilation unit that isn't included in the xx_unsecure libraries.

I can do that now, or in a follow up PR.

yashykt (approved):
My only request is to add TODOs on what all we want to clean-up (@veblush)

dawidcha:
@veblush I've made the changes you suggested yesterday: the grpc++_unsecure library is being skipped (or rather eviscerated) for msvc shared builds only.

That means that the hack of duplicate source includes and that library's need to depend on openssl has been removed.

I think this is the last outstanding barrier to merging.

As soon as this is done, I'll make the more robust fix we talked about - moving the openssl dependency from grpc_tls_certificate_provider to a separate source file - and re-enable grpc++_unsecure in CMakeLists.txt.

Subsequent to that, the next PR will fix build breaks with gRPC_BUILD_TESTS=ON.

Then another PR to expose functionality needed by our code from the generated DLLs.

dawidcha:
@veblush @yashykt Can you kick off the tests

veblush:
Thank you for your contribution!

h-vetinari:
> Thank you for your contribution!

Better late than never: I took the "you" to mean everyone involved, but explicitly also from my side: thanks to @dawidcha for shepherding this over the finish line (and of course to the grpc team for reviewing). 🙏

dawidcha:
> Better late than never: I took the "you" to mean everyone involved, but explicitly also from my side: thanks to @dawidcha for shepherding this over the finish line (and of course to the grpc team for reviewing). 🙏

fyi, there is a follow up PR to this which fixes an openssl dependency issue which was bypassed in this PR by disabling the grpc++_unsecure target [here](https://github.com/grpc/grpc/pull/34327) - just a code reorganization and already merged.

There is still work to be done here as currently not enough of the API is exposed to be actually useful. The first step to getting to that point is to get the tests compiling successfully (i.e. doing a build with gRPC_BUILD_TESTS=ON. That's the subject of [this PR](https://github.com/grpc/grpc/pull/34345), which is waiting for me to get a distrib test, that I added, running as part of the workflow checks.

Following on from that, I have some further changes, exposing more of the API to meet the use case that initially took us down this road.

I think we need to agree on a minimal viable API before we declare that we're done. @h-vetinari what does that look like from your point of view?

h-vetinari:
> I think we need to agree on a minimal viable API before we declare that we're done. @h-vetinari what does that look like from your point of view?

I already commented about this in #34345. From the POV of my use-case, long-term I'd like to be able to build with `-DgRPC_BUILD_TESTS=ON` and pass the test suite with the shared builds also on windows, but that's perhaps not exactly "minimal"? I don't know what more minimal, intermediate steps might still be possible, but perhaps it doesn't need all of the test suite?

FWIW, conda-forge has been shipping the patches that turned into this PR since 1.55, with grpcio built on top of that, without any major issues that I'm aware of. So if we're talking about "minimally viable", then that already has been achieved from my POV. But that doesn't meant that there isn't a lot left to improve, and I really appreciate that you've kept pushing on this! :)

Saleemfarheen1:
82501153837bac7ee6d1b69a20a81a573d8cd857
Hi please help 4
Saleemfarheen1234@gmail.com 
Saleemfarheen1786@gmail.com

PR #34104, opened 2023-08-18
[EventEngine] Track TCP global stats in the PosixEventEngine

drfloob:


ctiller (approved):

drfloob:
This can be merged when global stats no longer depends on ExecCtx. Alternatively, an ExecCtx can be created for every call to `grpc_core::global_stats()::Increment*`, but that seems unnecessary on the fast read/write paths. See https://github.com/grpc/grpc/pull/34124 for why it might be challenging to pepper ExecCtx more broadly through the PosixEndpoint code.

PR #34109, opened 2023-08-18
[ObjC] Enable DNS resolver with `GRPC_IOS_EVENT_ENGINE_CLIENT`

sampajano:
Now that iOS DNS resolver is ready, we can also enable it with `GRPC_IOS_EVENT_ENGINE_CLIENT`, so we can ask users to experiment with it.


drfloob (approved):

HannahShiSFB (approved):

sampajano:
Thanks so much for the reviews!

@HannahShiSFB Thanks for digging up the logs and confirming that the code is enabled! 😃

PR #34112, opened 2023-08-20
[Observability]Populate grpc_auth_context with local peer cert info

erm-g:
[Observability]Populate grpc_auth_context with local peer cert info.


src/core/tsi/transport_security_interface.h
```diff
@@ -258,6 +258,12 @@ typedef struct tsi_handshaker_result tsi_handshaker_result;
 tsi_result tsi_handshaker_result_extract_peer(const tsi_handshaker_result* self,
                                               tsi_peer* peer);
 
+// This method extracts tsi local peer. It returns TSI_OK assuming there is no fatal
+// error.
+// The caller is responsible for destructing the local peer.
+tsi_result tsi_handshaker_result_extract_local_peer(const tsi_handshaker_result* self,
```
matthewstevenson88, line 264: I think it may be overkill to add a new API to TSI - instead, can we just add the properties to the `tsi_peer` that we build during `tsi_handshaker_result_extract_peer`?

There is precedent for having local connection info in the `tsi_peer` (e.g. see the ALTS TSI handshaker) and it has several advantages:
1. Much less work in this PR and to import the code (there are many internal TSI implementations that would need updating).
2. Easier to maintain, because we don't have 2 parallel but nearly identical flows.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1254,6 +1255,57 @@ static tsi_result ssl_handshaker_result_extract_peer(
...
+  tsi_result result = TSI_OK;
+  const unsigned char* alpn_selected = nullptr;
+  unsigned int alpn_selected_len;
+  const tsi_ssl_handshaker_result* impl =
+      reinterpret_cast<const tsi_ssl_handshaker_result*>(self);
+  X509 *local_cert = SSL_get_certificate(impl->ssl);
+  if (local_cert != nullptr) {
+    result = peer_from_x509(local_cert, 1, local_peer);
+//    X509_free(local_cert);
+    if (result != TSI_OK) return result;
```
matthewstevenson88, line 1269: I do not think we should fail-closed here, can we add a debug-log instead of returning an error?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1254,6 +1255,57 @@ static tsi_result ssl_handshaker_result_extract_peer(
...
 }
 
+static tsi_result ssl_handshaker_result_extract_local_peer(
+    const tsi_handshaker_result* self, tsi_peer* local_peer) {
+  tsi_result result = TSI_OK;
+  const unsigned char* alpn_selected = nullptr;
+  unsigned int alpn_selected_len;
+  const tsi_ssl_handshaker_result* impl =
+      reinterpret_cast<const tsi_ssl_handshaker_result*>(self);
+  X509 *local_cert = SSL_get_certificate(impl->ssl);
```
matthewstevenson88, line 1265: Please check that `impl->ssl` is not nullptr before passing it into `SSL_get_certificate`. :)

matthewstevenson88 (commented):
Thanks for the PR! Left an initial high-level comment.

src/core/tsi/transport_security_interface.h
rockspore: Just to add an additional comment, I find this phrase `local_peer` confusing. IIUC, `peer` in this context usually means the other side.

PR #34116, opened 2023-08-21
[Python example] Add example for TLS

XuanWang-Amos:
Add example for TLS.

### DIFF
--- a/examples/python/auth/tls_client.py
+++ b/examples/python/auth/tls_client.py
@@ -1,0 +1,57 @@
+# Copyright 2023 The gRPC Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


examples/python/auth/tls_client.py
```diff
@@ -0,0 +1,57 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+_LOGGER.setLevel(logging.INFO)
+
+_PORT = 50051
+_SERVER_ADDR_TEMPLATE = "localhost:%d"
+
+
+def send_rpc(channel):
+    stub = helloworld_pb2_grpc.GreeterStub(channel)
```
gnossen, line 33: In general, we should encourage people to reuse stubs as well as channels. Let's construct the stub outside of the `send_rpc` function and pass it in instead of the channel.

gnossen (approved):
Looks great!

PR #34117, opened 2023-08-21
[benchmark] Local loadtest scenario runner

drfloob:
This helps developers run benchmark loadtests locally. See comments in scenario_runner.py for usage.


test/cpp/qps/scenario_runner.py
```diff
@@ -0,0 +1,138 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Local QPS benchmark runner for the OSS Benchmark loadtest configurations.
+
+This tool will run a scenario locally, either already extracted from
+scenario_config_extractor, or extracted from a benchmark loadtest config. The
```
paulosjca, line 20: Nit: Do you mean `scenario_config_exporter`?

test/cpp/qps/scenario_runner.py
```diff
@@ -0,0 +1,138 @@
...
+    loadtests = []
+    with open(
+        os.path.join(
+            os.path.dirname(os.path.abspath(__file__)), _LOADTEST_YAML.value
+        )
+    ) as f:
+        loadtests = list(yaml.safe_load_all(f))
+    if len(loadtests) > 1 and not (_RUN_FIRST.value or _RUN_ALL.value):
+        print(
+            "The loadtest configuration contains more than one scenario. Please specify --run_first or --run_all.",
```
paulosjca, line 111: Nit: The language being used here seems unclear. Each loadtest contains only one scenario. The "loadtest configuration file" may be a multipart yaml file containing multiple loadtests. So I would say "contains more than one loadtest".

It is also possible for a loadtest configuration file to contain multiple loadtests with the same scenario. This is the case for instance if they are generated with the `--runs_per_test` flag set.

https://github.com/grpc/grpc/blob/master/tools/run_tests/performance/README.md#generating-load-test-configurations

paulosjca (approved):

test/cpp/qps/scenario_runner.py
drfloob: Fixed, thanks.

test/cpp/qps/scenario_runner.py
drfloob: The language is fixed, thanks.

And that's a good point regarding the repeated scenarios. We have room for future improvements. This script ignores `--runs_per_test` in the loadtest configuration, and if folks invoke scenario_runner with `--run_all`, it will happily run the same scenario again if present.

PR #34122, opened 2023-08-21
[Windows] Make resolver_component_tests_runner_invoker run with Bazel on RBE

yijiem:
This makes the resolver component tests suite run on Window RBE by adding a flag in the test driver to further differentiate between Bazel local run and Bazel RBE run on Windows since they have different RUNFILES behavior.

Local Bazel run succeeds:
```
C:\Users\yijiem\projects\grpc>bazel --output_base=C:\bazel2 test --dynamic_mode=off --verbose_failures --test_arg=--running_locally=true //test/cpp/naming:resolver_component_tests_runner_invoker
INFO: Analyzed target //test/cpp/naming:resolver_component_tests_runner_invoker (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
Target //test/cpp/naming:resolver_component_tests_runner_invoker up-to-date:
  bazel-bin/test/cpp/naming/resolver_component_tests_runner_invoker.exe
INFO: Elapsed time: 196.080s, Critical Path: 193.21s
INFO: 2 processes: 1 internal, 1 local.
INFO: Build completed successfully, 2 total actions
//test/cpp/naming:resolver_component_tests_runner_invoker                PASSED in 193.1s

Executed 1 out of 1 test: 1 test passes.
```

RBE run succeeds:
```
C:\Users\yijiem\projects\grpc>bazel --bazelrc=tools/remote_build/windows.bazelrc test --config=windows_opt --dynamic_mode=off --verbose_failures --host_linkopt=/NODEFAULTLIB:libcmt.lib --host_linkopt=/DEFAULTLIB:msvcrt.lib --nocache_test_results //test/cpp/naming:resolver_component_tests_runner_invoker
INFO: Invocation ID: d467f2e3-7da6-4bb5-8b9b-84f1181ebc60
WARNING: --remote_upload_local_results is set, but the remote cache does not support uploading action results or the current account is not authorized to write local results to the remote cache.
INFO: Streaming build results to: https://source.cloud.google.com/results/invocations/d467f2e3-7da6-4bb5-8b9b-84f1181ebc60
INFO: Analyzed target //test/cpp/naming:resolver_component_tests_runner_invoker (0 packages loaded, 133 targets configured).
INFO: Found 1 test target...
Target //test/cpp/naming:resolver_component_tests_runner_invoker up-to-date:
  bazel-bin/test/cpp/naming/resolver_component_tests_runner_invoker.exe
INFO: Elapsed time: 41.627s, Critical Path: 39.42s
INFO: 2 processes: 1 internal, 1 remote.
//test/cpp/naming:resolver_component_tests_runner_invoker                PASSED in 33.0s

Executed 1 out of 1 test: 1 test passes.
INFO: Streaming build results to: https://source.cloud.google.com/results/invocations/d467f2e3-7da6-4bb5-8b9b-84f1181ebc60
INFO: Build completed successfully, 2 total actions
```


drfloob (approved):
CC @jtattermusch, in case you know of a way to automatically detect if an action is being run in the RBE environment.

hexiaokang (approved):

test/cpp/naming/resolver_component_tests_runner_invoker.cc
```diff
@@ -55,6 +55,16 @@ ABSL_FLAG(
     "False indicates that this test is running under run_tests.py. "
     "Child process test binaries are located differently based on this flag. ");
 
+#ifdef GPR_WINDOWS
+ABSL_FLAG(bool, running_locally, false,
```
jtattermusch, line 59: More options here:
- You could potentially use the "GRPC_PORT_ISOLATED_RUNTIME=1" define, which is only set when running on RBE (normally it's used to determine the port picker behavior, but with an explanatory comment you could potentially use it for detecting whether running on RBE or no).
https://github.com/grpc/grpc/blob/c867c418c13652b8ece1986cff6717541f89f2c8/tools/remote_build/include/rbe_remote_execution.bazelrc#L31
- alternatively, you could set an --action_env or --test_env variable and add it to rbe_remote_execution.bazelrc (since detection whether running on RBE or not is a reasonable thing to detect)

test/cpp/naming/resolver_component_tests_runner_invoker.cc
```diff
@@ -55,6 +55,16 @@ ABSL_FLAG(
...
+          "This flag only exists on Windows and is only effective when running "
+          "under Bazel. "
+          "Set this to true (e.g. --test_arg=--running_locally=true) when "
+          "running the test locally. "
+          "This is because the RUNFILES dir behavior is different between "
```
jtattermusch, line 64: What exactly is the difference? Is that documented somewhere? Have you investigated whether it's possible to make bazel runfiles behave the same when running locally vs remotely?

test/cpp/naming/resolver_component_tests_runner_invoker.cc
```diff
@@ -123,24 +133,47 @@ int main(int argc, char** argv) {
...
-        grpc::testing::NormalizeFilePath(manifest_file.Get(
-            "com_github_grpc_grpc/test/cpp/naming/utils/dns_server.exe")),
-        grpc::testing::NormalizeFilePath(
-            manifest_file.Get("com_github_grpc_grpc/test/cpp/naming/"
-                              "resolver_test_record_groups.yaml")),
-        grpc::testing::NormalizeFilePath(manifest_file.Get(
-            "com_github_grpc_grpc/test/cpp/naming/utils/dns_resolver.exe")),
-        grpc::testing::NormalizeFilePath(manifest_file.Get(
-            "com_github_grpc_grpc/test/cpp/naming/utils/tcp_connect.exe")));
+    if (absl::GetFlag(FLAGS_running_locally)) {
```
jtattermusch, line 136: there's a lot of copy/paste between the two possible option values. Sounds like you should have a utility function to locate a runfile and then you won't have the 2 invocations of grpc::testing::InvokeResolverComponentTestsRunner( with almost exactly the same params (and repeating each path twice).

Also, it sounds like accessing runfiles is something that we've must have done previously in other tests?

test/cpp/naming/resolver_component_tests_runner_invoker.cc
yijiem: good question. I found there is a `--enable_runfiles` option and is default off on Windows so Bazel will just create a MANIFEST file under the RUNFILES dir instead of creating the symlinks tree. When enabled, Bazel errored with:
```
C:\users\yijiem\_bazel_yijiem\install\1a52a8611778e900b15612375871f5a1\build-runfiles.exe --allow_relative bazel-out/x64_windows-fastbuild/bin/test/cpp/naming/utils/tcp_connect.exe.runfiles_manifest bazel-out/x64_windows-fastbuild/bin/test/cpp/naming/utils/tcp_connect.exe.runfiles: Process exited with status 1: Process exited with status 1
build-runfiles error: CreateSymbolicLinkW failed:
Bazel needs to create symlinks to build the runfiles tree.
Creating symlinks on Windows requires one of the following:
    1. Bazel is run with administrator privileges.
    2. The system version is Windows 10 Creators Update (1703) or later and developer mode is enabled.
```

After qualifying for these 2 conditions, Bazel runs the test just fine. I think most of the change in this file can be reverted and I have added logging to hint about this flag if Bazel is invoked locally (using the `GRPC_PORT_ISOLATED_RUNTIME` preprocessor symbol check).

test/cpp/naming/resolver_component_tests_runner_invoker.cc
yijiem: Done. not needed anymore.

test/cpp/naming/resolver_component_tests_runner_invoker.cc
yijiem: I have removed this flag since it's not needed anymore. We can just pass `--enable_runfiles=yes` to Bazel when run the test locally. See the below response.

test/cpp/naming/resolver_component_tests_runner_invoker.cc
```diff
@@ -123,24 +123,30 @@ int main(int argc, char** argv) {
         bin_dir + "/resolver_test_record_groups.yaml",
         bin_dir + "/utils/dns_resolver", bin_dir + "/utils/tcp_connect");
 #else
-    grpc::testing::ManifestFile manifest_file(
-        grpc::testing::NormalizeFilePath(test_srcdir.value()) + "\\MANIFEST");
-    grpc::testing::InvokeResolverComponentTestsRunner(
+#ifndef GRPC_PORT_ISOLATED_RUNTIME
+    gpr_log(GPR_ERROR,
```
jtattermusch, line 127: add a comment that you're knowingly misusing GRPC_PORT_ISOLATED_RUNTIME as an indication whether you're running on RBE or not (I think it's fine to use it, but the code should ack that this define is meant for something else and it just came handy in this situation). I think you can treat this as a TODO for finding a better way of detecting whether you're on RBE or not.

jtattermusch (approved):
This looks much better. LGTM after addressing the comment.

test/cpp/naming/resolver_component_tests_runner_invoker.cc
yijiem: Done.

PR #34120, opened 2023-08-21
[Deps] Update third_party/protobuf to 3.24.1

veblush:
Change was created by the release automation script. See go/grpc-release.

PR #34123, opened 2023-08-21
[Deps] Upgrade Protobuf and Upb to 24.x

veblush:
On top of https://github.com/grpc/grpc/pull/34120


veblush:
@jtattermusch Jan, would you take a look at this?

Upb's special target `generated_code_support__only_for_generated_code_do_not_use__i_give_permission_to_break_me` now has `upb/generated_code_support.h` header file that no other targets has. But our `extract_metadata_from_bazel_xml.py` doesn't seem to extract this header file.

jtattermusch (approved):
LGTM (once all the test are green), but the commit history is a mess and combined with updating lots of checked-in files, this is difficult to review. We should make it easier to review upb and protobuf update PRs in the future.

PR #34097, opened 2023-08-17
[PSM Interop] Cover TestConfig version parsing with unit tests

sergiitk:
This is to make sure upgrading packaging module won't break our logic on version-based version skipping.
This also fixes a small issue with `dev-` prefix - it should only be allowed on the left side of the comparison.

Context: packaging module needs to be upgraded to be compatible with `blackd`.

PR #34125, opened 2023-08-22
[PSM Interop] Add unittests CI with github actions

sergiitk:
- Add Github Action to conditionally run PSM Interop unit tests:
  - Only run when changes are detected in `tools/run_tests/xds_k8s_test_driver` or any of the proto files used by the driver
  - Only run against PRs and pushes to `master`, `v1.*.*` branches
  - Runs using `python3.9` and `python3.10`
  - Ready to be added to the list of required GitHub checks
- Add `tools/run_tests/xds_k8s_test_driver/tests/unit/__main__.py` test loader that recursively discovers all unit tests in `tools/run_tests/xds_k8s_test_driver/tests/unit`
- Add basic coverage for `XdsTestClient` and `XdsTestServer` to verify the test loader picks up all folders

Related:
- First unit tests without automated CI added in #34097

### DIFF
--- a/.github/workflows/psm-interop.yaml
+++ b/.github/workflows/psm-interop.yaml
@@ -1,0 +1,74 @@
+name: PSM Interop
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - 'v1.*'
+
+permissions:
+  contents: read
+
+jobs:
+  unittest:
+    # By default, only version is printed out in parens, e.g. "unittest (3.10)"
+    # This changes it to "unittest (python3.10)"
+    name: "unittest (python${{ matrix.python_version }})"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python_version: ["3.9", "3.10"]
+      fail-fast: false
+    permissions:
+      pull-requests: read  # Used by paths-filter to read the diff.
+    defaults:
+      run:
+        working-directory: 'tools/run_tests/xds_k8s_test_driver'
+
+    steps:
+      - uses: actions/checkout@v3
+
+      # To add this job to required GitHub checks, it's not enough to use
+      # the on.pull_request.paths filter. For required checks, the job needs to
+      # return the success status, and not be skipped.
+      # Using paths-filter action, we skip the setup/test steps when psm interop
+      # files are unchanged, and the job returns success.
+      - uses: dorny/paths-filter@v2
+        id: paths_filter
+        with:
+          filters:  |
+            psm_interop_src:
+              - 'tools/run_tests/xds_k8s_test_driver/**'
+              - 'src/proto/grpc/testing/empty.proto'
+              - 'src/proto/grpc/testing/messages.proto'
+              - 'src/proto/grpc/testing/test.proto'
+
+      - uses: actions/setup-python@v4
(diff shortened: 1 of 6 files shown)


sergiitk:
FYI @ejona86 @ctiller

sergiitk:
Test job run example already available in the PR checks: https://github.com/grpc/grpc/actions/runs/5933744203/job/16089599399?pr=34125

jtattermusch (approved):
LGTM from the test-infra standpoint. Someone should still review from the psm iterop test driver perspective, since I'm not familiar the logic that's being tested there.

A few general questions/pointers:
- Why does this test need to run under github actions? It's probably fine to run in as GA (since it's fast and seems to be stable), but the test could just as well run on kokoro.  So I'm trying to understand what lead you to setting this test up as a GA.
- I think this test harness could well run under bazel as a "bazelified test" (see go/grpc-bazelify-tests) in the future. I'm not going to enforce that now since the bazelified tests are basically still a prototype, but I think in the long run it's where this test should belong.

.github/workflows/psm-interop.yaml
```diff
@@ -0,0 +1,74 @@
...
+      # To add this job to required GitHub checks, it's not enough to use
+      # the on.pull_request.paths filter. For required checks, the job needs to
+      # return the success status, and not be skipped.
+      # Using paths-filter action, we skip the setup/test steps when psm interop
+      # files are unchanged, and the job returns success.
+      - uses: dorny/paths-filter@v2
+        id: paths_filter
+        with:
+          filters:  |
+            psm_interop_src:
```
jtattermusch, line 41: Fair enough, but filters like this tend to be hard to fragile and easily become out of date, so the question is whether the timesaving here is worth it (the test is currently very fast, right?).
If this was a bazel tests, you'd be able to make the test hermetic by only depending on the selected files, at which point you could be sure that the path filter always stays up-to-date.

Not blocking on this, but leaving up to you to consider.

.github/workflows/psm-interop.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  unittest:
+    # By default, only version is printed out in parens, e.g. "unittest (3.10)"
+    # This changes it to "unittest (python3.10)"
+    name: "unittest (python${{ matrix.python_version }})"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python_version: ["3.9", "3.10"]
```
gnossen, line 21: Not blocking, but since there are now multiple places where we're encoding the supported versions for the interop test framework, we might want to dedupe via (for example) templating.

gnossen (approved):

.github/workflows/psm-interop.yaml
sergiitk: Yes, good point. I didn't like it much too. For now, I think the benefits outweigh the risks. I _absolutely_ don't want to block anything in the repo because, f.e. newer pip breaks things.

For the long-term, I have plans to move the framework out to its own repo, but we should discuss it separately.

.github/workflows/psm-interop.yaml
sergiitk: Agree, but at the moment the manual burden of updating these once in a while is just cheaper/faster than templating it. Long term, I'll hoping to switch to bazel.

PR #34130, opened 2023-08-22
[PSM Interop] Don't initialize the secondary context when not needed

sergiitk:
Fixes an issue when an active context selected automatically picked up as context for `secondary_k8s_api_manager`.

This was introducing an error in GAMMA Baseline PoC

```
sys:1: ResourceWarning: unclosed <ssl.SSLSocket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('100.71.2.143', 56723), raddr=('35.199.174.232', 443)>
```

Here's how the secondary context is incorrectly falls back to the default context when `--secondary_kube_context` is not set:
```
k8s.py:142] Using kubernetes context "gke_grpc-testing_us-central1-a_psm-interop-security", active host: https://35.202.85.90
k8s.py:142] Using kubernetes context "None", active host: https://35.202.85.90
```


gnossen (approved):
Thanks for the fix!

PR #34131, opened 2023-08-22
[Testing] Disable failing OpenSSL Test

gtcooke94:
We enabled OpenSSL3 testing with #31256 and missed a failing test

It wasn't running before, so this isn't a regression - disabling it so master doesn't fail while we figure out how to fix it.


test/core/handshake/client_ssl.cc
```diff
@@ -399,6 +399,8 @@ static bool client_ssl_test(char* server_alpn_preferred) {
   return success;
 }
 
+// TODO(gtcooke94) Figure out why test is failing with OpenSSL and fix it.
+#ifdef OPENSSL_IS_BORING_SSL
```
matthewstevenson88, line 403: Instead of disabling the whole test, can we move the #ifdef's so that they are only around the final `client_ssl_test` call?

matthewstevenson88 (approved):

PR #34134, opened 2023-08-22
[ruby] Fix linking errors on x86-darwin

stanhu:
https://github.com/grpc/grpc/pull/33538 added `-weak_framework CoreFoundation` in `DLDFLAGS` for only `arm64-darwin` builds, but the issue reported in https://github.com/grpc/grpc/issues/33483 can also happen on `x86-darwin` builds. This can happen if:

1. The Ruby interpreter is compiled without `-Wl,-undefined,dynamic_lookup`.
2. This happens if the Ruby interpreter is built with XCode 14.0 to 14.2 (https://bugs.ruby-lang.org/issues/19005).

Simplify the logic and always include `-weak_framework CoreFoundation` for macOS builds.


alto-ruby:
@apolcyn I think this PR makes sense and I have approved it, can you also take a look?

alto-ruby (approved):

PR #34137, opened 2023-08-23
[csharp codegen] Handle empty base_namespace option value to fix #34113

### LINKED ISSUE #34113: setting C# option `base_namespace` to an empty string results in inconsistent behavior
gRPC 1.57.0, C#

NixOS 23.05 (Linux)

gcc (GCC) 12.3.0

### What did you do?
I invoked `protoc` with `grpc_csharp_plugin` & `--grpc_opt=base_namespace=` option:
```sh
protoc --plugin=protoc-gen-grpc=grpc_csharp_plugin \
    --grpc_out=OUT_DIR \
    --grpc_opt=base_namespace= \
    -I INCLUDE_DIR foo.proto
```

Generated C# files being written in a directory structure based on the namespace.
Which is the same behavior that protoc has for its equivalent `csharp_opt`:
https://github.com/protocolbuffers/protobuf/blob/1bfcedaf1280923f36e52a6fa0a5645ab63568f0/src/google/protobuf/compiler/csharp/csharp_options.h#L63-L67

### What did you see instead?
All generated C# files being written to the root directory, like `base_namespace` wasn't set at all.

I've originally reported this issue here: https://github.com/grpc/grpc/pull/33535#discussion_r1300188766

tonydnewell:
Fix for https://github.com/grpc/grpc/issues/34113 - setting C# option base_namespace to an empty string results in inconsistent behavior

Allow the `base_namespace` option in the protocol buffers C# gRPC plugin to be specified with an empty string to make sure the plugin has the same behaviour as the protocol buffers compiler.


tonydnewell:
@jtattermusch for review

jtattermusch (approved):

jtattermusch:
kick the mergeable status

PR #34170, opened 2023-08-25
Make the code generator pointer mutable in PluginMain

mkruskal-google:
This will allow us to pass some extra initialization information from the protobuf plugin code before the build actually starts


mkruskal-google:
@veblush

eugeneo (approved):

PR #34180, opened 2023-08-28
[Security] Move ownership of tsi_ssl_client_handshaker_factory to grpc_ssl_credentials.

gtcooke94:
Move the SSL_CTX to the level of the credentials rather than the subchannel.
The SSL_CTX should only get created once per credential rather than once per subchannel.

We should observe no behavior change with this PR, only efficiency gains.


src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,40 @@ grpc_ssl_credentials::create_security_connector(
   absl::optional<std::string> overridden_target_name =
       args->GetOwnedString(GRPC_SSL_TARGET_NAME_OVERRIDE_ARG);
   auto* ssl_session_cache = args->GetObject<tsi::SslSessionLRUCache>();
+  tsi_ssl_session_cache* tsi_ssl_session_cache =
+      ssl_session_cache == nullptr ? nullptr : ssl_session_cache->c_ptr();
+
+  const tsi_ssl_root_certs_store* root_store;
+  // Use default root certificates.
+  if (config_.pem_root_certs == nullptr) {
+    // TODO(gtcooke94) make sure to remove this memory
```
matthewstevenson88, line 77: Should we resolve or remove this TODO in this PR?

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,40 @@ grpc_ssl_credentials::create_security_connector(
   absl::optional<std::string> overridden_target_name =
       args->GetOwnedString(GRPC_SSL_TARGET_NAME_OVERRIDE_ARG);
   auto* ssl_session_cache = args->GetObject<tsi::SslSessionLRUCache>();
+  tsi_ssl_session_cache* tsi_ssl_session_cache =
```
matthewstevenson88, line 71: micro-nit: Please make the variable name different than the type name, e.g. maybe `session_cache`.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,40 @@ grpc_ssl_credentials::create_security_connector(
...
+    char* default_roots = strcpy(new char[root_len + 1], pem_root_certs);
+    config_.pem_root_certs = default_roots;
+    root_store = grpc_core::DefaultSslRootStore::GetRootStore();
+  } else {
+    config_.pem_root_certs = config_.pem_root_certs;
+    root_store = nullptr;
+  }
+
+  grpc_security_status status = initialize_client_handshaker_factory(
+      &config_, config_.pem_root_certs, root_store, tsi_ssl_session_cache);
```
matthewstevenson88, line 94: I don't think we can pass in the `tsi_ssl_session_cache` in this way because it is coming from the ChannelArgs, which are tied to the channel rather than to the SSL credentials.

To get around this, I was thinking that we could add a new `tsi_ssl_session_cache*` parameter to the `tsi_ssl_client_handshaker_factory_create_handshaker` API, and plumb the session cache pointer from here to where the security connector calls that API. WDYT?

If we can do that, then I think we can also move all of this logic to the constructor of the SSL credentials.

src/core/lib/security/credentials/ssl/ssl_credentials.h
```diff
@@ -70,6 +70,11 @@ class grpc_ssl_credentials : public grpc_channel_credentials {
                     const grpc_ssl_verify_peer_options* verify_options);
 
   grpc_ssl_config config_;
+  grpc_security_status initialize_client_handshaker_factory(
```
matthewstevenson88, line 73: nit: Please move this method to be above the member fields.

src/core/lib/security/credentials/ssl/ssl_credentials.h
```diff
@@ -70,6 +70,11 @@ class grpc_ssl_credentials : public grpc_channel_credentials {
...
+      const grpc_ssl_config* config, const char* pem_root_certs,
+      const tsi_ssl_root_certs_store* root_store,
+      tsi_ssl_session_cache* ssl_session_cache);
+  tsi_ssl_client_handshaker_factory* client_handshaker_factory_ = nullptr;
```
matthewstevenson88, line 77: Can we store this as a grpc_core::RefCounterPtr instead?

This also means that we will have to pass it around as a grpc_core::RefCounterPtr instead of as a raw pointer.

Rationale: We have to be very careful with lifetimes here, in case the user destroys the credentials object once the channel is created but before the TLS handshake is complete.

src/core/lib/security/credentials/ssl/ssl_credentials.h
matthewstevenson88, line 73: micro-nit: For C++, the function name should be capitalized and camel-case, e.g. `InitializeClientHandshakerFactory` rather than `initialize_client_handshaker_factory`. The rest of this file is following C98 style and we should clean it up at some point, but for new code we should definitely follow C++ style.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,40 @@ grpc_ssl_credentials::create_security_connector(
   absl::optional<std::string> overridden_target_name =
       args->GetOwnedString(GRPC_SSL_TARGET_NAME_OVERRIDE_ARG);
   auto* ssl_session_cache = args->GetObject<tsi::SslSessionLRUCache>();
+  tsi_ssl_session_cache* tsi_ssl_session_cache =
+      ssl_session_cache == nullptr ? nullptr : ssl_session_cache->c_ptr();
+
+  const tsi_ssl_root_certs_store* root_store;
+  // Use default root certificates.
```
matthewstevenson88, line 75: micro-nit: Maybe this should say "Use the default (e.g. OS) root certificates if the user did not pass in any root certificates."

matthewstevenson88 (commented):
Thanks for the PR Greg!

src/core/lib/security/credentials/ssl/ssl_credentials.cc
gtcooke94: I think that makes sense with the `tsi_ssl_session_cache`. `tsi_ssl_client_handshaker_factory_create_handshaker` ends up getting called here - https://github.com/grpc/grpc/blob/4295f0250c6344c1e492bdb45d4ad01ae28d9177/src/core/lib/security/security_connector/ssl/ssl_security_connector.cc#L140 and we have access to the channel args, so that should be easy to plumb.

I don't know if we can move all the logic to the SSL credentials - we're still making a lot of calls to OpenSSL that can fail when creating and populating the SSL_CTX, so we need to be able to do a status return

src/core/lib/security/credentials/ssl/ssl_credentials.h
gtcooke94: That makes a lot of sense, will make the change

src/core/lib/security/credentials/ssl/ssl_credentials.cc
gtcooke94: Yes, I will remove it

src/core/lib/security/credentials/ssl/ssl_credentials.h
gtcooke94: To make sure I'm clear, you want to go on and change this to `InitializeClientHandshakerFactory`? Or just note it and save it for a larger PR where we restyle everything

src/core/lib/security/credentials/ssl/ssl_credentials.h
gtcooke94: I realized the handshaker_factories are already doing ref counting c-style, added explicit code to make this clear.

See [here](https://github.com/grpc/grpc/blob/5c9e4050c1e8c66b1b8b4e93d48b8e23ff1ea818/src/core/tsi/ssl_transport_security.cc#L1759) for an example of where it is already being counted.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -119,6 +149,42 @@ void grpc_ssl_credentials::set_max_tls_version(
   config_.max_tls_version = max_tls_version;
 }
 
+grpc_security_status grpc_ssl_credentials::initialize_client_handshaker_factory(
+    const grpc_ssl_config* config, const char* pem_root_certs,
+    const tsi_ssl_root_certs_store* root_store,
+    tsi_ssl_session_cache* session_cache) {
+  if (client_handshaker_factory_ != nullptr) {
```
matthewstevenson88, line 156: I don't think the current strategy works because of the following example:
* User creates SSL credentials object and holds on to it for the lifetime of their application.
* User uses the SSL credentials object to create channel #1 without any session cache in the channel args.
* User uses the SSL credentials object to create channel #2 with a session cache in the channel args.

The user expects that both TLS handshakes are full handshakes, and the cache is populated with a session ticket after the creation of channel #2. However, this will not happen with the current strategy because the handshaker factory is created when we create channel #1 (when there is no session cache), and it is reused when we create channel #2.

To get around examples like this, I'm proposing we slightly tweak the logic as follows:
1. In the `grpc_ssl_credentials` constructor, call `initialize_client_handshaker_factory` with no session cache and store the resulting client handshaker factory as a member field of the `grpc_ssl_credentials` class.
2. In the `grpc_ssl_credentials::create_security_connector` method, try to read the session cache from the channel args, as you're doing now. If the session cache is nullptr, then using the client handshaker factory that is stored as a member field; otherwise, call `initialize_client_handshaker_factory` to build a new client handshaker factory (with the non-null session cache) and we will only use this client handshaker factory for this subchannel.

WDYT?

gtcooke94: That makes us call `SSL_CTX_new` and make a new `client_handshaker_factory` in the case of a session cache - we lose some efficiency but it would be simpler because then we're keeping this old behavior with a session cache, but if there is no session cache in the channel args then we take the new path tying SSL_CTX to the credential.
Does that make sense? I am good with that approach if so

matthewstevenson88: Correct: we maintain the status quo in terms of performance for applications that are specifying the session cache via the channel args, and we land the performance boost (by moving the `SSL_CTX`) for anyone else who is not specifying the session cache via the channel args.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,61 @@ grpc_ssl_credentials::create_security_connector(
...
+      return nullptr;
+    }
+    sc = grpc_ssl_channel_security_connector_create(
+        this->Ref(), std::move(call_creds), &config_, target,
+        overridden_target_name.has_value() ? overridden_target_name->c_str()
+                                           : nullptr,
+        factory_with_cache);
+    tsi_ssl_client_handshaker_factory_unref(factory_with_cache);
+  } else {
+    grpc_security_status status = initialize_client_handshaker_factory(
```
matthewstevenson88, line 113: Can we move this call to `initialize_client_handshaker_factory` to the constructor?

You can save the `status` as a member field if you want, and check the status here.

Otherwise, we're not getting the performance gain I think.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
matthewstevenson88: Friendly ping. :)

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,61 @@ grpc_ssl_credentials::create_security_connector(
...
-  grpc_core::RefCountedPtr<grpc_channel_security_connector> sc =
-      grpc_ssl_channel_security_connector_create(
-          this->Ref(), std::move(call_creds), &config_, target,
-          overridden_target_name.has_value() ? overridden_target_name->c_str()
-                                             : nullptr,
-          ssl_session_cache == nullptr ? nullptr : ssl_session_cache->c_ptr());
+  tsi_ssl_session_cache* tsi_ssl_session_cache =
+      ssl_session_cache == nullptr ? nullptr : ssl_session_cache->c_ptr();
+
+  const tsi_ssl_root_certs_store* root_store;
```
matthewstevenson88, line 74: Why do we need lines 74-91 on each call to `create_security_connector`?

Can we do this work in the constructor and cache the result?

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,61 @@ grpc_ssl_credentials::create_security_connector(
...
+    size_t root_len = strlen(pem_root_certs);
+    char* default_roots = strcpy(new char[root_len + 1], pem_root_certs);
+    config_.pem_root_certs = default_roots;
+    root_store = grpc_core::DefaultSslRootStore::GetRootStore();
+  } else {
+    config_.pem_root_certs = config_.pem_root_certs;
+    root_store = nullptr;
+  }
+
+  grpc_core::RefCountedPtr<grpc_channel_security_connector> sc = nullptr;
```
matthewstevenson88, line 93: nit: I know lots of the C-core does this, but the normal C++ style is to use underscore-separated non-abbreviated words for variable names instead of abbreviations. To that end, can we change this variable name from `sc` to `security_connector`?

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -66,12 +68,61 @@ grpc_ssl_credentials::create_security_connector(
...
+  // We need a separate factory and ctx if there's a cache in the channel
```
matthewstevenson88, line 94: Can we combine this comment and the one just inside the if statement? Also, can we explain why we're having to bifurcate in this way?

src/core/lib/security/credentials/ssl/ssl_credentials.h
matthewstevenson88: Correct, let's change the naming to `InitializeClientHandshakerFactory`. :)

src/core/lib/security/security_connector/ssl/ssl_security_connector.cc
```diff
@@ -204,8 +171,9 @@ class grpc_ssl_channel_security_connector final
                          overridden_target_name_.c_str(), auth_context));
   }
 
+  tsi_ssl_client_handshaker_factory* client_handshaker_factory_ = nullptr;
```
matthewstevenson88, line 174: Please make this a private member field.

src/core/tsi/ssl_transport_security.h
```diff
@@ -223,6 +223,10 @@ tsi_result tsi_ssl_client_handshaker_factory_create_handshaker(
     const char* server_name_indication, size_t network_bio_buf_size,
     size_t ssl_bio_buf_size, tsi_handshaker** handshaker);
 
+// Increments reference count of the client handshaker factory.
+tsi_ssl_client_handshaker_factory* tsi_ssl_client_handshaker_factory_ref(
```
matthewstevenson88, line 227: Please add unit tests for this function to make sure it plays well with unref-ing.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
gtcooke94: I think we still get the efficiency gain because `initialize_client_handshaker_factory` just returns the class `client_handshaker_factory_` if it already exists, so the credentials really just initiate it once the first time this is called.

Doing it in the constructor, storing the status, then checking the status at this location would just move the heavy lifting to construction time of the credentials from the initial establishing of the connection - all of this happens before the first connection regardless, so I don't think shifting it necessarily gets any notable gain, just displaces it.

WDYT?

src/core/lib/security/credentials/ssl/ssl_credentials.cc
gtcooke94: This was copy paste from elsewhere in the code - I agree that we should switch, will make the change

src/core/lib/security/credentials/ssl/ssl_credentials.cc
matthewstevenson88: We will get some efficiency gain, but the current code is not thread-safe: if you make multiple concurrent calls to `create_security_connector`, then there will be race conditions over populating the handshaker factory. To make it thread-safe, we would need to add a lock to guard the code block that populates and fetches the handshaker factory. However, I strongly suspect that contention over this lock would eat away at most of the performance gains that we want (at in the use cases where you're continuously creating a large volume of connections).

If we move everything to the constructor, then the cached handshaker factory is constant throughout the lifetime of the `grpc_ssl_credentials` object, and we avoid the lock contention issue.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
gtcooke94: Other than wanting to be able to return out, which we can still do if `pem_root_certs == nullptr`
Will move to constructor

src/core/lib/security/security_connector/ssl/ssl_security_connector.cc
gtcooke94: Made it private and then added it to the constructor of `grpc_ssl_channel_security_connector` PTAL

src/core/tsi/ssl_transport_security.h
gtcooke94: Added to a test in `ssl_transport_security_test.cc`

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -57,26 +79,62 @@ grpc_ssl_credentials::~grpc_ssl_credentials() {
...
   }
+  tsi_ssl_client_handshaker_factory_unref(client_handshaker_factory_);
 }
 
 grpc_core::RefCountedPtr<grpc_channel_security_connector>
 grpc_ssl_credentials::create_security_connector(
     grpc_core::RefCountedPtr<grpc_call_credentials> call_creds,
     const char* target, grpc_core::ChannelArgs* args) {
+  if (config_.pem_root_certs == nullptr) {
+    gpr_log(GPR_ERROR, "Could not get default pem root certs.");
```
matthewstevenson88, line 90: nit: I think this error message is not correct, since it should not say "default". Could we change it to e.g. "No root certs."?

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -57,26 +79,62 @@ grpc_ssl_credentials::~grpc_ssl_credentials() {
...
-          ssl_session_cache == nullptr ? nullptr : ssl_session_cache->c_ptr());
-  if (sc == nullptr) {
-    return sc;
+  tsi_ssl_session_cache* session_cache =
+      ssl_session_cache == nullptr ? nullptr : ssl_session_cache->c_ptr();
+
+  grpc_core::RefCountedPtr<grpc_channel_security_connector> security_connector =
+      nullptr;
+  if (session_cache != nullptr) {
+    // We need a separate factory and ctx if there's a cache in the channel
```
matthewstevenson88, line 102: micro-nit: s/ctx/SSL_CTX

For consistency throughout the paragraph.

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -57,26 +79,62 @@ grpc_ssl_credentials::~grpc_ssl_credentials() {
...
+    // channel args, so that prevents us from also keeping the session cache at
+    // the credentials level. In the case of a session cache, we still need to
+    // keep a separate factory and SSL_CTX at the subchannel/security_connector
+    // level.
+    tsi_ssl_client_handshaker_factory* factory_with_cache = nullptr;
+    grpc_security_status status = InitializeClientHandshakerFactory(
+        &config_, config_.pem_root_certs, root_store_, session_cache,
+        &factory_with_cache);
+    if (status != GRPC_SECURITY_OK) {
+      return nullptr;
```
matthewstevenson88, line 114: Can we have an error log that includes the status code in this case?

src/core/lib/security/credentials/ssl/ssl_credentials.cc
```diff
@@ -119,6 +177,45 @@ void grpc_ssl_credentials::set_max_tls_version(
   config_.max_tls_version = max_tls_version;
 }
 
+grpc_security_status grpc_ssl_credentials::InitializeClientHandshakerFactory(
+    const grpc_ssl_config* config, const char* pem_root_certs,
+    const tsi_ssl_root_certs_store* root_store,
+    tsi_ssl_session_cache* ssl_session_cache,
+    tsi_ssl_client_handshaker_factory** handshaker_factory) {
+  // This class level factory can't have a session cache by design. If we want
```
matthewstevenson88, line 185: Should this if statement be removed and we can make this into a static function?

src/core/lib/security/credentials/ssl/ssl_credentials.h
```diff
@@ -69,7 +69,16 @@ class grpc_ssl_credentials : public grpc_channel_credentials {
                     grpc_ssl_pem_key_cert_pair* pem_key_cert_pair,
                     const grpc_ssl_verify_peer_options* verify_options);
 
+  grpc_security_status InitializeClientHandshakerFactory(
```
matthewstevenson88, line 72: nit: Can we mark this as `static`?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1762,6 +1762,13 @@ void tsi_ssl_client_handshaker_factory_unref(
   tsi_ssl_handshaker_factory_unref(&factory->base);
 }
 
+tsi_ssl_client_handshaker_factory* tsi_ssl_client_handshaker_factory_ref(
+    tsi_ssl_client_handshaker_factory* client_factory) {
+  if (client_factory == nullptr) return nullptr;
+  return reinterpret_cast<tsi_ssl_client_handshaker_factory*>(
```
matthewstevenson88, line 1768: To make this code a bit more readable, could we instead do:
```
if (client_factory == nullptr) return nullptr;
tsi_ssl_handshaker_factory_ref(&client_factory->base);
return client_factory
```

?

src/core/lib/security/credentials/ssl/ssl_credentials.h
matthewstevenson88, line 72: Can we add a comment describing this function?

src/core/lib/security/credentials/ssl/ssl_credentials.cc
matthewstevenson88: Friendly ping :)

src/core/lib/security/credentials/ssl/ssl_credentials.cc
matthewstevenson88: Friendly ping :)

src/core/lib/security/credentials/ssl/ssl_credentials.cc
matthewstevenson88: Friendly ping :)

src/core/lib/security/credentials/ssl/ssl_credentials.cc
matthewstevenson88: Friendly ping :)

src/core/lib/security/credentials/ssl/ssl_credentials.h
matthewstevenson88: Friendly ping :)

matthewstevenson88: Friendly ping :)

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Friendly ping :)

src/core/lib/security/credentials/ssl/ssl_credentials.cc
gtcooke94: Changed to `No root certs in config. Client-side security connector must have root certs.`

src/core/lib/security/credentials/ssl/ssl_credentials.cc
gtcooke94: I don't think so - this is the bifurcation. The class level factory can't have a session cache, so if the `ssl_session_cache != nullptr` then we need to make a separate factory with a separate CTX that is not associated with the credentials class.

But in the case that there is _not_ a session cache, we do want to use a factory on the class instance that is only initialized once.

gtcooke94: So this if statement is essentially saying
"If we don't want to have a session cache, and we have an existing `client_handshaker_factory_` (i.e. `client_handshaker_factory_ != nullptr`), then everything is ok and we can use that factory that is already initialized.

src/core/lib/security/credentials/ssl/ssl_credentials.h
gtcooke94: See the thread above this - this initializes the `client_handshaker_factory_` on this class instance of credentials. But if there's a session cache we use it to make a new factory.
I don't think we want it to be static - it uses and modifies the instance `client_handshaker_factory_`

src/core/tsi/ssl_transport_security.cc
gtcooke94: I'm not 100% on this - the semantics of the `tsi_ssl_handshaker_factory_ref` that this calls is to return the factory, but's it's also just the same one that's in. I figured semantically since it returns it, there was a reason for that and it should be used?

gtcooke94: https://github.com/grpc/grpc/blob/1986007e1ee735ec3f977f31cc8569c5ad954c7e/src/core/tsi/ssl_transport_security.cc#L1132-L1137

src/core/lib/security/credentials/ssl/ssl_credentials.cc
matthewstevenson88: SGTM.

src/core/lib/security/credentials/ssl/ssl_credentials.h
gtcooke94: PTAL

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Up to you. I think they're equivalent, but I'm also totally fine leaving it as-is.

tools/run_tests/run_tests.py
```diff
@@ -56,7 +56,7 @@
 )
 sys.path.append(gcp_utils_dir)
 
-_ROOT = os.path.abspath(os.path.join(os.path.dirname(sys.argv[0]), "../.."))
+_ROOT = os.path.abspath(os.path.join(os.path.dirname(sys.argv[0]), "..", ".."))
```
matthewstevenson88, line 59: Do we need the changes in this file? Should they be reverted?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Let's leave as-is

tools/run_tests/run_tests.py
gtcooke94: These are actually related to the test-pr, this diff should go away when this gets refreshed to the new master

gtcooke94: Actually I'm wrong, changing this back

matthewstevenson88 (approved):
LGTM!

matthewstevenson88 (approved):

PR #34184, opened 2023-08-28
[XdsClient] don't unref watchers while holding mutex

markdroth:
This should fix the deadlock described in b/297601540.


src/core/ext/xds/xds_client.cc
```diff
@@ -1690,7 +1702,7 @@ void XdsClient::CancelResourceWatch(const XdsResourceType* type,
   if (resource_it == type_map.end()) return;
   ResourceState& resource_state = resource_it->second;
   // Remove watcher.
-  resource_state.watchers.erase(watcher);
+  delete_watcher_from(resource_state.watchers);
```
anicr7, line 1705: Discussed this offline but for posterity, the actual issue that we noticed is being caused by authority_state.channel_state.reset(line 1721) right now.

So we also need to find to prevent a possible deadlock due to it along with the watcher deletion.

PR #34185, opened 2023-08-28
[c-ares] fix spin loop bug when c-ares gives up on a socket that still has data left in its read buffer

apolcyn:
If we get a readable event on an fd and both the following happens:

- c-ares does *not* read all bytes off the fd

- c-ares removes the fd from the set ARES_GETSOCK_READABLE

... then we have a busy loop here, where we'd keep asking c-ares to process an fd that it no longer cares about.

This is indirectly related to a change in this code one month ago: https://github.com/grpc/grpc/pull/33942 - before that change, c-ares would close the socket when it called [handle_error](https://github.com/c-ares/c-ares/blob/7f3262312f246556d8c1bdd8ccc1844847f42787/src/lib/ares_process.c#L707) and so `IsFdStillReadableLocked` would start returning `false`, causing us to get away with [this loop](https://github.com/grpc/grpc/blob/f6a994229e72bc771963706de7a0cd8aa9150bb6/src/core/ext/filters/client_channel/resolver/dns/c_ares/grpc_ares_wrapper.cc#L371). Now, because `IsFdStillReadableLocked` will keep returning true (because of our overridden `close` API), we'll loop forever. 

The test illustrates one concrete example of how this bug can be hit.

Note that the EE version of this code already [gets this right](https://github.com/grpc/grpc/blob/f6fd5172added1ba92658a227302346540ad2022/src/core/lib/event_engine/ares_resolver.cc#L369).

Related: internal issue b/297538255

### DIFF
--- a/test/cpp/naming/cancel_ares_query_test.cc
+++ b/test/cpp/naming/cancel_ares_query_test.cc
@@ -286,10 +286,10 @@ void TestCancelDuringActiveQuery(
     grpc_status_code expected_status_code,
     absl::string_view expected_error_message_substring,
     gpr_timespec rpc_deadline, int dns_query_timeout_ms,
-    int /*fake_dns_server_port*/) {
+    int fake_dns_server_port) {
   // Create a call that will try to use the fake DNS server
   std::string client_target =
-      absl::StrFormat("dns:///%s", kFakeName);
+      absl::StrFormat("dns://[::1]:%d/%s", fake_dns_server_port, kFakeName);
   grpc_channel_args* client_args = nullptr;
   if (dns_query_timeout_ms >= 0) {
     grpc_arg arg;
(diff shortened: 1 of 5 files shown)


yijiem (approved):

test/cpp/naming/cancel_ares_query_test.cc
```diff
@@ -471,6 +471,71 @@ TEST_F(CancelDuringAresQuery, TestQueryFailsBecauseTcpServerClosesSocket) {
...
+//   1) We force c-ares to use TCP for its DNS queries
+//   2) We stand up a fake DNS server that, for each incoming connection, sends
+//      three all-zero bytes and then closes the socket.
+//   3) When the c-ares library receives the three-zero-byte response from the
+//      DNS server, it parses the first two-bytes as a length field:
+//      https://github.com/c-ares/c-ares/blob/6360e96b5cf8e5980c887ce58ef727e53d77243a/src/lib/ares_process.c#L410.
+//   4) Because the first two bytes were zero, c-ares attempts to malloc a
+//      zero-length buffer:
+//      https://github.com/c-ares/c-ares/blob/6360e96b5cf8e5980c887ce58ef727e53d77243a/src/lib/ares_process.c#L428.
+//   5) Because malloc(0) returns NULL, c-ares invokes handle_error and stops
```
yijiem, line 495: Just a small nit: maybe say c-ares' default_malloc(0) returns NULL instead? Since it seems like some systems may return a valid pointer on malloc(0): https://github.com/c-ares/c-ares/blob/main/src/lib/ares_library_init.c#L38-L42

yijiem (approved):
What a hack! Thanks Alex for adding the test to reproduce the issue!

test/cpp/naming/cancel_ares_query_test.cc
apolcyn: Good point, done

PR #34191, opened 2023-08-29
[chaotic-good] Initial change of chaotic-good client-read path.

nanahpang:
This is the initial change of chaotic-good client transport read path, which is a following PR of the client transport write path at #33876. There's a pending work of handling endpoint failures in the transport. It will be added after we have the inter-activity pipe with close function.


src/core/ext/transport/chaotic_good/client_transport.h
```diff
@@ -55,41 +66,101 @@ class ClientTransport {
...
+            ForEach(std::move(*call_args.client_to_server_messages),
+                    [stream_id, initial_frame = true,
+                     client_initial_metadata =
+                         std::move(call_args.client_initial_metadata),
+                     outgoing_frames = outgoing_frames_.MakeSender()](
+                        MessageHandle result) mutable {
+                      ClientFragmentFrame frame;
+                      frame.stream_id = stream_id;
+                      frame.message = std::move(result);
+                      if (initial_frame) {
```
Vignesh2208, line 101: I may have missed this in the previous review. According to the chaotic good specification, Isn't the first frame sent and received by the client expected to be Settings frame instead of a ClientFragmentFrame and a ServerFragmentFrame respectively ?

src/core/ext/transport/chaotic_good/frame.cc
```diff
@@ -115,7 +120,11 @@ absl::StatusOr<Arena::PoolPtr<Metadata>> ReadMetadata(
     uint32_t stream_id, bool is_header, bool is_client) {
   if (!maybe_slices.ok()) return maybe_slices.status();
   auto& slices = *maybe_slices;
-  Arena::PoolPtr<Metadata> metadata;
+  MemoryAllocator memory_allocator = MemoryAllocator(
+      ResourceQuota::Default()->memory_quota()->CreateMemoryAllocator(
+          "read_metadata"));
+  auto arena = MakeScopedArena(1024, &memory_allocator);
```
Vignesh2208, line 126: I dont think we should be allocating an arena each time a frame is parsed. We should change the ServerFragmentFrame::Deserialize and ReadMetadata methods to plumb the arena ptr from the transport down to this spot.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -113,6 +118,70 @@ ClientTransport::ClientTransport(
...
+              FrameHeader::Parse(
+                  reinterpret_cast<const uint8_t*>(GRPC_SLICE_START_PTR(
+                      read_buffer.c_slice_buffer()->slices[0])))
+                  .value());
+          // Read header and trailers from control endpoint.
+          // Read message padding and message from data endpoint.
+          return TryJoin(
+              control_endpoint_->Read(frame_header_->GetFrameLength()),
+              TrySeq(
+                  data_endpoint_->Read(frame_header_->message_padding), [this] {
```
Vignesh2208, line 141: Reading the padding first and then the message length may cause two separate syscalls which could be slow. Can we change this to read message_padding + message_length in one go and then later strip the padding away in the next promise possibly ?

src/core/lib/transport/promise_endpoint.cc
```diff
@@ -46,10 +46,10 @@ PromiseEndpoint::PromiseEndpoint(
 }
 
 PromiseEndpoint::~PromiseEndpoint() {
-  // Last write result has not been polled.
-  GPR_ASSERT(!write_result_.has_value());
-  // Last read result has not been polled.
-  GPR_ASSERT(!read_result_.has_value());
+  // Promise endpoint close when last write result has not been polled.
+  if (write_result_.has_value()) write_result_.reset();
```
Vignesh2208, line 50: You can just always call reset(). It will check internally if there is any value

test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -127,218 +244,447 @@ class ClientTransportTest : public ::testing::Test {
...
+                    return absl::OkStatus();
+                  }),
+              // Receive messages from control/data endpoints.
+              Seq(
+                  // Receive server initial metadata.
+                  pipe_server_intial_metadata_.receiver.Next(),
+                  // Receive server to client messages.
+                  pipe_server_to_client_messages_.receiver.Next(),
+                  [this] {
+                    // Close pipes after receive message.
```
Vignesh2208, line 347: Can you add expectations here and elsewhere to validate that the initial metadata and message values match the expected ones ?

Vignesh2208 (commented):
Hi Nana, Thanks for working on this. Left my initial round of comments

src/core/ext/transport/chaotic_good/client_transport.h
nanahpang: The setting frames are sent during the connection setup path (internal link: https://docs.google.com/document/d/1jQFfG1TGc6VUq54L0apQv2aP5JRbsVe_wMl06m4xG9w/edit?resourcekey=0-6ZkXEgvGvGKnQLpezGgDuw#bookmark=id.qe7uy5kftwqo), which is finished before the `AddStream` (request/response path). We will work on the connection setup step in following PRs.

src/core/ext/transport/chaotic_good/frame.cc
nanahpang: Yeah, thanks for catching this! I will create arena in one place of the transport and pass the ptr down.

src/core/ext/transport/chaotic_good/client_transport.h
```diff
@@ -55,41 +66,101 @@ class ClientTransport {
...
+                  server_frames]() mutable {
+              return TrySeq(
+                  // Receive incoming server frame.
+                  server_frames->Next(),
+                  // Save incomming frame results to call_args.
+                  [server_initial_metadata = std::move(server_initial_metadata),
+                   server_to_client_messages =
+                       std::move(server_to_client_messages)](
+                      ServerFrame server_frame) mutable {
+                    auto frame =
```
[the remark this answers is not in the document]
nanahpang: It doesn't need to be a shared_ptr, I have updated it to be local stack variable. Thanks for catching this.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -113,6 +118,70 @@ ClientTransport::ClientTransport(
...
+                    // The message_padding will be discarded here.
+                    return data_endpoint_->Read(frame_header_->message_length);
+                  }));
+        },
+        // Construct and send the server frame to corresponding stream.
+        [this](std::tuple<SliceBuffer, SliceBuffer> ret) mutable {
+          control_endpoint_read_buffer_ = std::move(std::get<0>(ret));
+          data_endpoint_read_buffer_ = std::move(std::get<1>(ret));
+          auto frame = std::make_shared<ServerFragmentFrame>();
```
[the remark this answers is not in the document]
nanahpang: No, it doesn't need to be a shared_ptr, I change it to local stack variable. Thanks for catching this.

src/core/lib/transport/promise_endpoint.cc
nanahpang: Done.

src/core/ext/transport/chaotic_good/client_transport.cc
nanahpang: Sounds good, done.

test/core/transport/chaotic_good/client_transport_test.cc
nanahpang: Sure, done.

test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -127,218 +232,524 @@ class ClientTransportTest : public ::testing::Test {
...
-      nullptr,       &pipe_client_to_server_messages_second_.receiver, nullptr};
+  auto first_stream_args =
+      CallArgs{std::move(md),
+               ClientInitialMetadataOutstandingToken::Empty(),
+               nullptr,
+               &pipe_server_intial_metadata_.sender,
+               &pipe_client_to_server_messages_.receiver,
+               &pipe_server_to_client_messages_.sender};
+  auto second_stream_args =
+      CallArgs{std::move(md),
```
Vignesh2208, line 576: md is a local variable. Calling std::move(md) two times may lead to crashes since the second move is pointing invalid memory.

test/core/transport/chaotic_good/client_transport_test.cc
```diff
@@ -127,218 +232,524 @@ class ClientTransportTest : public ::testing::Test {
...
```
Vignesh2208, line 736: Same comment as above w.r.t the local variable md

Vignesh2208 (commented):
Reviewed and it looks mostly good. Left a few clarifications. Thanks for making the changes!

test/core/transport/chaotic_good/client_transport_test.cc
nanahpang: Done, thanks for catching this!

test/core/transport/chaotic_good/client_transport_test.cc
nanahpang: Done.

src/core/ext/transport/chaotic_good/frame.h
```diff
@@ -82,6 +82,9 @@ struct ClientFragmentFrame final : public FrameInterface {
     return stream_id == other.stream_id && EqHdl(headers, other.headers) &&
            end_of_stream == other.end_of_stream;
   }
+  // Shared ownership with transport.
+  std::shared_ptr<Arena> arena_;
+  void SetArena(std::shared_ptr<Arena> arena) { arena_ = arena; }
```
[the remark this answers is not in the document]
nanahpang: Done.

src/core/ext/transport/chaotic_good/client_transport.h
```diff
@@ -55,56 +70,124 @@ class ClientTransport {
...
+                         std::move(call_args.client_initial_metadata),
+                     outgoing_frames = outgoing_frames_.MakeSender()](
+                        MessageHandle result) mutable {
+                      ClientFragmentFrame frame;
+                      frame.stream_id = stream_id;
+                      frame.message = std::move(result);
+                      if (initial_frame) {
+                        // Send initial frame with client intial metadata.
+                        frame.headers = std::move(client_initial_metadata);
+                        initial_frame = false;
```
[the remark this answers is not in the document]
nanahpang: The `initial_frame` will be false after the first frame is sent because the lambda function is mutable and the capture list value could be changed in runtime (I double checked this with printing out`std::cout<<initial_frame` in the lambda function).

src/core/ext/transport/chaotic_good/frame.cc
```diff
@@ -112,10 +113,14 @@ class FrameDeserializer {
 template <typename Metadata>
 absl::StatusOr<Arena::PoolPtr<Metadata>> ReadMetadata(
     HPackParser* parser, absl::StatusOr<SliceBuffer> maybe_slices,
-    uint32_t stream_id, bool is_header, bool is_client) {
+    uint32_t stream_id, bool is_header, bool is_client,
+    std::shared_ptr<Arena> arena) {
   if (!maybe_slices.ok()) return maybe_slices.status();
   auto& slices = *maybe_slices;
   Arena::PoolPtr<Metadata> metadata;
+  if (arena != nullptr) {
```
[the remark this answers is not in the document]
nanahpang: Done, thanks for the suggestion.

src/core/ext/transport/chaotic_good/frame.h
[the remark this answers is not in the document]
nanahpang: Actually I found that using `GetContext<Arena>()` is better here, so that we don't need to add an explicit constructor in the segment frames, and prevent distinguishing setting frames from segment frames during initialization.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -113,6 +123,73 @@ ClientTransport::ClientTransport(
...
+    return TrySeq(
+        // Read frame header from control endpoint.
+        this->control_endpoint_->Read(FrameHeader::frame_header_size_),
+        // Read different parts of the server frame from control/data endpoints
+        // based on frame header.
+        [this](SliceBuffer read_buffer) mutable {
+          frame_header_ = std::make_shared<FrameHeader>(
+              FrameHeader::Parse(
+                  reinterpret_cast<const uint8_t*>(GRPC_SLICE_START_PTR(
+                      read_buffer.c_slice_buffer()->slices[0])))
```
Vignesh2208, line 139: Why are we assuming here that the whole frame header will be in the first slice i.e slice at index 0 ? Is it possible for the promise endpoint to return the frame header split across many slices ?

src/core/ext/transport/chaotic_good/client_transport.h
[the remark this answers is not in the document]
Vignesh2208: Thanks for the clarification.

src/core/ext/transport/chaotic_good/client_transport.cc
```diff
@@ -113,6 +123,73 @@ ClientTransport::ClientTransport(
...
+        });
+  });
+  reader_ = MakeActivity(
+      // Continuously read next incoming frames from promise endpoints.
+      std::move(read_loop), EventEngineWakeupScheduler(event_engine_),
+      [](absl::Status status) {
+        GPR_ASSERT(status.code() == absl::StatusCode::kCancelled ||
+                   status.code() == absl::StatusCode::kInternal);
+        // TODO(ladynana): handle the promise endpoint read failures with
+        // iterating stream_map_ and close all the pipes once available.
```
Vignesh2208, line 192: I think you may also need to pass the arena_.get() variable to both the reader and writer activity like this: https://github.com/grpc/grpc/blob/master/test/core/promise/activity_test.cc#L266. Otherwise if GetContext<Arena> is executed in another thread, it will probably fail.

src/core/ext/transport/chaotic_good/frame.h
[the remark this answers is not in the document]
Vignesh2208: Thanks. Left a clarification question. I think you may need to pass the arena pointer to the activity.

Vignesh2208 (commented):
Thanks for making the changes Nana. Left a couple of final questions.

Vignesh2208 (approved):

gnossen:
[It looks like this may be the cause of a regression](https://fusion2.corp.google.com/ci;ids=1923284992/kokoro/prod:grpc%2Fcore%2Fmaster%2Flinux%2Fgrpc_basictests_c_cpp_dbg/activity/e056cddd-85cb-4351-9f08-f4dd8cd9e671/log)

PR #34192, opened 2023-08-29
[ObjC] Update swift package dependencies

HannahShiSFB:
This is to accommodate the change in #33803, adding missing dependencies of swift package.


sampajano (approved):
thanks! :)

PR #34196, opened 2023-08-29
[ObjC] remove libuv podspec and tests

HannahShiSFB:
libuv was previously considered as the base of event engine, it's not the case anymore.
It was removed from third_party in #33748, this PR remove it from objc


sampajano (approved):
Thanks for cleaning up! 😃

PR #34199, opened 2023-08-29
[Deps] Adding upb as a submodule

veblush:
Let's make it as a regular dependency like others. So gRPC is now unvendoring it first. Project generation part will be followed later.


yashykt (approved):

veblush:
Failing tests are caused by changing subtree to submodule. This won't happen once this is merged.

jtattermusch:
Hm, IMHO changing upb into a submodule isn't really possible without also switching our cmake build to using the upb's CMakeLists.txt (since for anything that isn't vendored in our repo directly, we need to depend on a properly installable library).

Looks like this PR is breaking the C++ distribtests precisely for this reason:
https://source.cloud.google.com/results/invocations/1cde23c8-bba2-4e69-9b63-0baac91d8a77/targets
https://source.cloud.google.com/results/invocations/12b587f0-ff07-4712-9136-ab82eb8d0259/targets/github%2Fgrpc/tests

So more followup work is needed.

matthewjmiller1:
> Hm, IMHO changing upb into a submodule isn't really possible without also switching our cmake build to using the upb's CMakeLists.txt (since for anything that isn't vendored in our repo directly, we need to depend on a properly installable library).
> 
> Looks like this PR is breaking the C++ distribtests precisely for this reason: https://source.cloud.google.com/results/invocations/1cde23c8-bba2-4e69-9b63-0baac91d8a77/targets https://source.cloud.google.com/results/invocations/12b587f0-ff07-4712-9136-ab82eb8d0259/targets/github%2Fgrpc/tests
> 
> So more followup work is needed.

Is there anything tracking this followup work?

As far as I understand it, this commit violated this invariant:

https://github.com/grpc/grpc/blob/master/third_party/README.md
> gRPC C++ needs to stay buildable/installable even if the submodules are not present (e.g. the tar.gz archive with gRPC doesn't contain the submodules), assuming that the dependencies are already installed. This is a requirement for being able to provide a reasonable install process (e.g. using cmake) and to support package managers for gRPC C++.

Since we can no longer build/install gRPC if the upb submodule is not present, even if it is already installed.

matthewjmiller1:
It also looks like upb itself may not support building via cmake:
https://github.com/protocolbuffers/protobuf/blob/main/upb/cmake/README.md
Which, even if gRPC were to support using upb as an external dependency instead of a submodule, still would make gRPC unusable in environments where all the dependencies are built from source using cmake.

PR #34211, opened 2023-08-30
[sanity] Fix tools/buildgen/generate_projects.sh

yijiem:


yijiem:
See https://github.com/grpc/grpc/pull/34212.

PR #33880, opened 2023-07-26
[bazel] Switching win RBE build to 6.x

jtattermusch:
Followup for https://github.com/grpc/grpc/pull/32910

PR #34216, opened 2023-08-31
[RBE windows] Use bazel 6.3.2 on windows RBE (after regenerating toolchain)

jtattermusch:
- Upgrade windows RBE builds to bazel 6.3.2 (supersedes https://github.com/grpc/grpc/pull/33880). To be able to do that, the RBE toolchain needed to be regenerated and
- Also added "--dynamic_mode=off" option for windows build since it seemed necessary for bazel 6.x builds to pass.
- Wrote instructions for generating windows RBE toolchain using the `rbe_configs_gen` tool (the original windows RBE toolchain was out of data and also it was generated by a custom script from go/rbe-windows-user-guide - using a standard tool is better)
- Wrote instructions for rebuilding the windows RBE docker image.

This addresses the problem where windows RBE is stuck on bazel 5.x (unlike the rest of the repository) and also documents the steps for making changes to the RBE docker image (e.g. upgrading the visual studio version used by RBE).

Since we were planning on adding testing for VS2022 (for which we don't have any test ATM), this will definitely come handy. With the documentation the process should now be relatively straightforward.


jtattermusch:
FYI @yijiem - for things like  adding a windows clang RBE build, you'd need a new (or updated) docker image and a new toolchain configuration (and this PR has instructions for doing both).

veblush (approved):

PR #34218, opened 2023-08-31
[pick_first] changes to support dualstack design

markdroth:
This rolls forward only the pick_first changes from #32692, which were rolled back in #33718.  Specifically:
- Changes PF to use its own subchannel list implementation instead of using the subchannel_list library, since the latter will be going away with the dualstack changes.
- As a result of no longer using the subchannel_list library, PF no longer needs to set the `GRPC_ARG_INHIBIT_HEALTH_CHECKING` channel arg.
- Adds an option to start a health watch on the chosen subchannel, to be used in the future when pick_first is the child of a petiole policy.  (Currently, this code is not actually called anywhere.)

### DIFF
--- a/src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
+++ b/src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
@@ -16,6 +16,8 @@
 
 #include <grpc/support/port_platform.h>
 
+#include "src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.h"
+
 #include <inttypes.h>
 #include <string.h>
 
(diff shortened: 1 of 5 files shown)


src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -418,46 +606,47 @@ void PickFirst::PickFirstSubchannelData::ProcessConnectivityChangeLocked(
...
   // with the first subchannel.
   if (!old_state.has_value()) {
-    subchannel_list()->subchannel(0)->ReactToConnectivityStateLocked();
+    subchannel_list_->subchannels_.front().ReactToConnectivityStateLocked();
     return;
   }
   // Ignore any other updates for subchannels we're not currently trying to
   // connect to.
-  if (Index() != subchannel_list()->attempting_index()) return;
+  if (Index() != subchannel_list_->attempting_index_) return;
```
eugeneo, line 626: Lack of encapsulation for `attempting_index_` seems unusual.

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -418,46 +606,47 @@ void PickFirst::PickFirstSubchannelData::ProcessConnectivityChangeLocked(
...
       // large recursion that could overflow the stack.
-      PickFirstSubchannelData* found_subchannel = nullptr;
+      SubchannelData* found_subchannel = nullptr;
       for (size_t next_index = Index() + 1;
-           next_index < subchannel_list()->num_subchannels(); ++next_index) {
-        PickFirstSubchannelData* sc = subchannel_list()->subchannel(next_index);
-        GPR_ASSERT(sc->connectivity_state().has_value());
-        if (sc->connectivity_state() != GRPC_CHANNEL_TRANSIENT_FAILURE) {
-          subchannel_list()->set_attempting_index(next_index);
+           next_index < subchannel_list_->size(); ++next_index) {
```
eugeneo, line 645: 645-667 seem to use more of `subchannel_list_` - so maybe there should be a "find_next_subchannel_not_transient_failure` method in subchannel list class?

eugeneo (approved):

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
markdroth: I'm not sure what you mean by "encapsulation".  If you mean the fact that we're directly accessing a data member, I think it's fine, because this access is coming from `SubchannelData`, which is nested inside of `SubchannelList` and can therefore access its data members by normal C++ scoping rules.  In effect, `SubchannelData` is part of the implementation of `SubchannelList`, so it's fine to access its private data members.  We use this pattern quite a bit.

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
markdroth: This isn't a method that would ever need to be called from anywhere else, so I don't see any real benefit in moving it to a separate function.

PR #34219, opened 2023-08-31
[call] Fix race in batch error processing

Vignesh2208:
Fixes b/285925908


ctiller (approved):

src/core/lib/surface/call.cc
```diff
@@ -1479,9 +1490,9 @@ void FilterStackCall::BatchControl::ReceivingTrailingMetadataReady(
   // If the call is cancelled, the ReceivingInitialMetadataReady would run
   // before this step and set the batch error to non OK status. We must use
   // as the final status of the rpc.
-  if (error.ok()) {
-    error = batch_error_.get();
-  }
+  // if (error.ok() && call_->is_client()) {
```
ctiller: Please remove commented out code

Vignesh2208: Its removed in the next commit

PR #34221, opened 2023-08-31
[HTTP Proxy] Rename source/header and move test

eugeneo:


yashykt (approved):

test/core/ext/filters/client_channel/http_proxy_mapper_test.cc
markdroth, line 1: This test should not have been moved.  We do not need to exactly mirror the directory structure from the `src` tree in the `test` tree -- and in the long term, we will probably remove the arbitrary division between the `lib` and `ext` directories in the `src` tree.

The client channel tests should stay in `test/core/client_channel`.

eugeneo: Sorry, my bad. My understanding was there was no need to move the test but it was ok to do that. PR to move it back: https://github.com/grpc/grpc/pull/34268

PR #34227, opened 2023-09-01
[Python] Add streaming interceptor example

ysk24ok:
Added the example because there is no Python streaming interceptor example.
This example doesn't include any decent processing and just prints messages. But I think it'd be helpful to tell where to put a certain processing based on the message.

Future work:
- Add server streaming (and bidirectional streaming) to [hellostreamingworld.proto](https://github.com/grpc/grpc/blob/master/examples/protos/hellostreamingworld.proto) and expand this example
- Add `async_streaming` to [examples/python/interceptors](https://github.com/grpc/grpc/tree/master/examples/python/interceptors)


examples/python/interceptors/streaming/hellostreamingworld_pb2.pyi
```diff
@@ -0,0 +1,67 @@
+"""
```
gnossen, line 1: These generated files should be added to [the Makefile](https://github.com/grpc/grpc/blob/master/examples/python/Makefile).

gnossen (approved):
Thanks for the contribution! I didn't realize this gap in the examples was there.

XuanWang-Amos:
Verified locally that it works as intended, we'll merge it once all tests passed.

ysk24ok:
@XuanWang-Amos It doesn't seem failed tests have anything to do with this PR, but could you tell me how to fix them?

XuanWang-Amos:
Just started another test run, can you help fix sanity tests? You can run scripts locally using command like `./tools/distrib/black_code.sh`

ysk24ok:
@XuanWang-Amos Thank you. Formatted the code.

PR #34230, opened 2023-09-01
[lang/c++] Make new LLVM version compile with unused variable

gaschler:
LLVM is making the unused variable warning a default: https://github.com/llvm/llvm-project/commit/92023b15099012a657da07ebf49dd7d94a260f84 This will make grpc compile after this LLVM update.

(release notes: no)


gaschler:
@drfloob please take a look (feel free to re-assign)

src/core/lib/channel/connected_channel.cc
```diff
@@ -548,7 +548,7 @@ ArenaPromise<ServerMetadataHandle> MakeClientCallPromise(
         }
         if (!done_recv_messages) {
           auto p = recv_messages();
-          if (auto* r = p.value_if_ready()) {
+          if ([[maybe_unused]] auto* r = p.value_if_ready()) {
```
yijiem, line 551: `[[maybe_unused]]` is a C++17 attribute. I don't think we have moved to C++17 yet. And if initialization is C++17 extension too, see: https://godbolt.org/z/Mc5TTMe8T. So I suggest changing this code to:
```
auto* r = p.value_if_ready();
if (r) {
...
} else {
...
}
```

Cc @ctiller

yashykt: probably just replace this with `p.value_if_ready() != nullptr`

gaschler: Done (first option), please take another look.

yijiem (approved):

src/core/lib/channel/connected_channel.cc
yijiem: Ah, sorry, I think @yashykt's suggestion is fair, i.e. change this to `if (p.value_if_ready() != nullptr) {`. At least we should not depend on implicit bool conversion.

gaschler: Done, please take another look

ctiller: This could just be spelled `if (p.ready()) {`

gaschler: Done, I now went with p.ready(), please take another look.
(Also, if you can trigger the remaining checks, that would be helpful, I'm unfamilar with how PRs are tested and merged in the grpc repo.)

veblush (approved):

PR #34239, opened 2023-09-01
[Example] Added gRPC C++ cancellation example

veblush:


examples/cpp/cancellation/client.cc
```diff
@@ -0,0 +1,122 @@
+/*
+ *
+ * Copyright 2018 gRPC authors.
```
yashykt, line 3: fix copyright year

examples/cpp/cancellation/server.cc
```diff
@@ -0,0 +1,108 @@
...
+using grpc::Server;
+using grpc::ServerBidiReactor;
+using grpc::ServerBuilder;
+using grpc::Status;
+using keyvaluestore::KeyValueStore;
+using keyvaluestore::Request;
+using keyvaluestore::Response;
+
+// Logic behind the server's behavior.
+class KeyValueStoreServiceImpl final : public KeyValueStore::CallbackService {
```
yashykt, line 48: This is not a key value store. I think you should add a method to helloworld.proto to have a bidi streaming RPC

yashykt (approved):

examples/cpp/cancellation/server.cc
```diff
@@ -0,0 +1,108 @@
+/*
+ *
+ * Copyright 2018 gRPC authors.
```
yashykt, line 3: fix copyright year here as well

examples/cpp/cancellation/client.cc
veblush: Done

examples/cpp/cancellation/server.cc
veblush: Done

examples/cpp/cancellation/server.cc
veblush: Done

Issue #34249, opened 2023-09-04
GetCpuStatsTest failed occasionally on arm


xiacunshun:
### What version of gRPC and what language are you using?
Version:        1.48.1

### What operating system (Linux, Windows,...) and version?
Linux VM-238-86-TS3 5.4.119-19-0011

### What runtime / compiler are you using (e.g. python version or version of gcc)
gcc-12.3.0-2.ocs23.aarch64 1692234622 81807597 02207b4a09e21f1f6a0e9954c1c6d084 installed
gcc-c++-12.3.0-2.ocs23.aarch64 1692234622 31436460 ed9d6c915e891be70378a65ffbf511ed installed

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.
compile the code and pack them into rpm.

### What did you expect to see?
success

### What did you see instead?
/builddir/build/BUILD/grpc-1.48.1/test/cpp/server/load_reporter/get_cpu_stats_test.cc:48: Failure
Expected: (prev.first) <= (cur.first), actual: 69428708 vs 69428707
[  FAILED  ] GetCpuStatsTest.Ascending (0 ms)
[----------] 3 tests from GetCpuStatsTest (0 ms total)
[----------] Global test environment tear-down
[==========] 3 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 2 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] GetCpuStatsTest.Ascending
 1 FAILED TEST
error: Bad exit status from /var/tmp/rpm-tmp.JDvo8k (%check)
    Bad exit status from /var/tmp/rpm-tmp.JDvo8k (%check)

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

xiacunshun:
I guess the compiler may have done some optimization that leads to this error?

yashykt:
My first reaction looking at this test is that I don't understand why this has condition has to be true.

xiacunshun:
> My first reaction looking at this test is that I don't understand why this has condition has to be true.
> 
[...]

PR #34264, opened 2023-09-06
[cpustat test] work around cpu stats may count in abnormal number

xiacunshun:
fix https://github.com/grpc/grpc/issues/34249


xiacunshun:
hello?

test/cpp/server/load_reporter/get_cpu_stats_test.cc
```diff
@@ -44,6 +44,9 @@ TEST(GetCpuStatsTest, Ascending) {
   const size_t kRuns = 100;
   auto prev = grpc::load_reporter::GetCpuStatsImpl();
   for (size_t i = 0; i < kRuns; ++i) {
+    // prev.first is larger than cur.first occasionally.
+    // Maybe a kernel or compiler problem.
+    sleep(1);
```
yashykt, line 49: No, I don't think we want to do this

PR #34266, opened 2023-09-06
[bazelified tests] Reenable runtests_php_linux_dbg after #34257

jtattermusch:
The valgrind based test (removed in https://github.com/grpc/grpc/pull/34257) was the reason why the dbg test was disabled in the past.


jtattermusch:
Manual run of the PHP bazelified tests with `--runs_per_test=10`:
https://source.cloud.google.com/results/invocations/5775ba24-6873-435c-b980-8b6c6b85b504/targets

stanley-cheung (approved):

PR #34270, opened 2023-09-06
[Python Misc] Remove fetch_build_eggs & Fix run_test

XuanWang-Amos:
* Remove `fetch_build_eggs` since they're deprecated and those deps will be installed by `setuptools`.
* Fix indentation on `run_test` so we don't miss `native` test cases.


gnossen (approved):
LGTM if the tests pass. Nice catch!

PR #34273, opened 2023-09-07
[OTel C++] Add ability to select channels for stats based on the target

yashykt:
@ctiller PTAL for core configuration changes. I converted the type from std::function to absl::AnyInvocable. Do you think the functor in RegisteredBuilder should be callable just once or multiple times?


ctiller:
I think this looks right... would be a boon if you could split the core config change from the otel change (especially to reduce churn in the event of rollbacks of one or the other)

yashykt:
Done https://github.com/grpc/grpc/pull/34282/files

veblush (approved):

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -132,6 +133,12 @@ class OpenTelemetryPluginBuilder {
   OpenTelemetryPluginBuilder& SetLabelsInjector(
       std::unique_ptr<LabelsInjector> labels_injector);
 
+  // If set, \a channel_selector is called once per channel to decide whether to
+  // collect metrics on the channel or not.
+  OpenTelemetryPluginBuilder& SetChannelSelector(
```
markdroth, line 138: Given our discussion about eventually wanting to separate the target and channel scopes, it seems like this is actually the target-level scope selector, not the channel-level scope selector.  We may eventually want a channel-level selector that takes channel args as input.  I don't want to actually add that right now, but let's leave ourselves room to do that -- can we rename this to `SetTargetSelector()`?

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -175,12 +186,21 @@ void OpenTelemetryPluginBuilder::BuildAndRegisterGlobal() {
...
         builder->channel_init()->RegisterStage(
             GRPC_CLIENT_CHANNEL, /*priority=*/INT_MAX,
-            [](grpc_core::ChannelStackBuilder* builder) {
-              builder->PrependFilter(
-                  &grpc::internal::OpenTelemetryClientFilter::kFilter);
+            [channel_selector = std::move(channel_selector)](
+                grpc_core::ChannelStackBuilder* builder) {
+              // Only register the filter if no channel selector has been set or
+              // the channel selector returns true for the target.
+              if (!channel_selector ||
```
markdroth, line 197: Prefer to write this as `channel_selector == nullptr`.

markdroth (approved):

src/cpp/ext/otel/otel_plugin.h
yashykt: yeah, i'm not gonna make this part of the API in the public header, but yeah I'll make the name change anyway.

src/cpp/ext/otel/otel_plugin.cc
yashykt: Done

Issue #33918, opened 2023-07-29
ENH: Compatibility with Cython 3.0


h-vetinari:
This was capped in #33738, but should be enabled again as soon as reasonable. Cython 0.29.x is deep in maintenance-only mode, now even more so that the release of Cython 3.0 has finally happened.

gnossen:
Thanks for the interest. We're planning on tackling this within the next month or two.

badshah400:
Since I package python-grpcio RPMs for openSUSE, and we have already moved to Cython 3 in our repositories, I was looking at what would be required to get grpc cython-ised builds succeeding again. Following the [cython3 migration guide](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html), I finally ended up with the following patch that does at least make openSUSE builds succeed:

```patch
Index: grpcio-1.57.0/src/python/grpcio/grpc/_cython/_cygrpc/vtable.pyx.pxi
===================================================================
--- grpcio-1.57.0.orig/src/python/grpcio/grpc/_cython/_cygrpc/vtable.pyx.pxi
+++ grpcio-1.57.0/src/python/grpcio/grpc/_cython/_cygrpc/vtable.pyx.pxi
@@ -13,16 +13,16 @@
 # limitations under the License.
 
 # TODO(https://github.com/grpc/grpc/issues/15662): Reform this.
-cdef void* _copy_pointer(void* pointer):
+cdef void* _copy_pointer(void* pointer) noexcept:
   return pointer
 
 
 # TODO(https://github.com/grpc/grpc/issues/15662): Reform this.
-cdef void _destroy_pointer(void* pointer):
+cdef void _destroy_pointer(void* pointer) noexcept:
   pass
 
 
-cdef int _compare_pointer(void* first_pointer, void* second_pointer):
+cdef int _compare_pointer(void* first_pointer, void* second_pointer) noexcept:
   if first_pointer < second_pointer:
     return -1
   elif first_pointer > second_pointer:
Index: grpcio-1.57.0/src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi
===================================================================
--- grpcio-1.57.0.orig/src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi
+++ grpcio-1.57.0/src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi
@@ -314,7 +314,7 @@ def server_credentials_ssl_dynamic_cert_
   return credentials
[...]

PR #34275, opened 2023-09-07
[Python] Specify noexcept for cdef functions.

badshah400:
To build against cython 3.0, cdef functions that do not raise exceptions need to be explicitly declared as noexcept. Fixes issue #33918.

### DIFF
--- a/src/python/grpcio/grpc/_cython/_cygrpc/aio/callback_common.pxd.pxi
+++ b/src/python/grpcio/grpc/_cython/_cygrpc/aio/callback_common.pxd.pxi
@@ -48,7 +48,7 @@ cdef class CallbackWrapper:
     @staticmethod
     cdef void functor_run(
             grpc_completion_queue_functor* functor,
-            int succeed)
+            int succeed) noexcept
 
     cdef grpc_completion_queue_functor *c_functor(self)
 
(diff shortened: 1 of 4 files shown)


XuanWang-Amos:
Thanks for the help! Can you also help try upgrade Cython to 3.0? (Totally optional).

badshah400:
> Thanks for the help! Can you also help try upgrade Cython to 3.0? (Totally optional).

Sure, I'll give it a shot, but I should probably do that as part of a separate PR, shouldn't I?

XuanWang-Amos:
> > Thanks for the help! Can you also help try upgrade Cython to 3.0? (Totally optional).
> 
> Sure, I'll give it a shot, but I should probably do that as part of a separate PR, shouldn't I?

If that requires some other work, then sure. The reason why I mention that is because we'd like to see update Cython to 3.0 works with those changes (basically a way of showing those are required for the upgrade).

badshah400:
> > > Thanks for the help! Can you also help try upgrade Cython to 3.0? (Totally optional).
> > 
> > 
> > Sure, I'll give it a shot, but I should probably do that as part of a separate PR, shouldn't I?
> 
> If that requires some other work, then sure. The reason why I mention that is because we'd like to see update Cython to 3.0 works with those changes (basically a way of showing those are required for the upgrade).

I see. I shall look at it this weekend and add commits to the PR as required. Thanks for the suggestion.

nanonyme:
> Sure, I'll give it a shot, but I should probably do that as part of a separate PR, shouldn't I?

To be fair, if this is not done, the PR most likely should *not* be marked to fix the issue. The issue is about Cython 3.0 support and updating to Cython 3.0 is part of that.

badshah400:
Agree, this will require more work. Thanks for the feedback.

PR #29957, opened 2022-06-08
MacOS system roots support

tdbhacks:
This change follows #29436 and includes:
* adding a cert file path for MacOS
* updating related test to run on MacOS too
* s/linux/supported/ since this now includes more platforms
* regenerating files affected by the name change

@ZhenLian

Issue #25533, opened 2021-02-23
Support loading system root certificates for Win32/OSX


ikstewa:
### Is your feature request related to a problem? Please describe.
When connecting to a TLS enabled GRPC endpoint the system root certificates are loaded on Linux/Android. When running on Win32/OSX the cert needs to be provided.

Related issue: https://github.com/grpc/grpc/issues/25167

### Describe the solution you'd like
Similar to the grpc impl for java/c# the client should load the trusted certs for operating systems beyond Linux/Android.

### Describe alternatives you've considered
The alternative is for each application to load the system certs and manually specify when creating the channel.

ikstewa:
Please don't close out! It would be really nice to have feature parity for Mac/Win/Linux.

RayKoopa:
I think this should be reopened because:

- The [example provided on the website](https://grpc.io/docs/guides/auth/#with-server-authentication-ssltls-2) doesn't work as provided and only has a rather easy-to-miss "Note" [at another place further up](https://grpc.io/docs/guides/auth/#using-client-side-ssltls) that it indeed won't work on non-POSIX systems.
- There is no example code found for Windows, which would be very helpful. The only hint I got is from a [comment](https://github.com/grpc/grpc/issues/25167#issuecomment-769605799) made on issue https://github.com/grpc/grpc/issues/25167. IMHO that issue should not be closed either, as the problem mentioned in the title is _not_ solved ("_Default_ SSL credentials don't work on Win32/OsX").
- Other implementations like grpc-dotnet don't have this issue with their default `SslCredentials` instances, which at least got me [quite confused](https://stackoverflow.com/questions/71580681/ssl-grpc-client-works-fine-in-c-but-fails-with-unavailable-empty-update-in-c). Once I figured out the WinAPI to manually set up the credentials I'll probably post it as an answer there.

(I originally mentioned running into this trouble at https://github.com/grpc/grpc/issues/25829 but noticed that issue was reported on Linux specifically, while I'm attempting to do this on Windows 10, so I'm not sure if that issue is related.)

ikstewa:
It looks like I don't have the ability to re-open this issue.

While that example code solves windows, we still have a similar issue on OSX as well.

RayKoopa:
Isn't macOS a POSIX OS? The warning in the authentication guide is slightly inaccurate then if it is not working with default SslCredentials.

ZhenLian:
[...]

PR #34276, opened 2023-09-07
Windows system roots support

tdbhacks:
This PR is similar to #29436 and #29957 in nature.

The code is based on the discussion in #25533, which has been confirmed to work both by other users and by some manual tests I've ran myself using the [Cloud KMS CNG provider](https://github.com/GoogleCloudPlatform/kms-integrations), which has been relying on manually setting the GRPC_DEFAULT_SSL_ROOTS_FILE_PATH environment variable to a previously downloaded PEM roots bundle. This change allows gRPC to try loading roots from the Windows root cert store too, and fallback to the existing loading methods if that fails.

@ZhenLian


jgeewax:
Hope this gets merged soon....

PR #34277, opened 2023-09-07
[Session Affinity] Update validation and add a test case

eugeneo:


test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -440,6 +449,36 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
       GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 1, kNumEchoRpcs / 3));
 }
 
+TEST_P(OverrideHostTest, DisabledPerPath) {
+  const std::string kCustomCookieName = "GSSA";
```
markdroth, line 453: Using a different cookie name doesn't seem to actually add any value for this test.  I think it can just use `kCookieName`.

If you do keep this for some reason, it can be `absl::string_view` instead of `std::string`.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -440,6 +449,36 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
       GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 1, kNumEchoRpcs / 3));
 }
 
+TEST_P(OverrideHostTest, DisabledPerPath) {
```
markdroth, line 452: This test should be called `EnablePerRoute`.

markdroth, line 452: In addition to this test, let's also test the case where the config is set at the top level and there is a per-route override that disables it.

markdroth (commented):
This looks good overall!  My main comment is that we also need to test the inverse case, and it looks like we have a bug to fix there as well.

src/core/ext/xds/xds_http_stateful_session_filter.cc
```diff
@@ -77,7 +77,6 @@ Json::Object ValidateStatefulSession(
       envoy_extensions_filters_http_stateful_session_v3_StatefulSession_session_state(
           stateful_session);
   if (session_state == nullptr) {
-    errors->AddError("field not present");
```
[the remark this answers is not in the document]
eugeneo: Will come in another PR

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -99,12 +102,16 @@ class OverrideHostTest : public XdsEnd2endTest {
   // Builds a Listener with Fault Injection filter config. If the http_fault
   // is nullptr, then assign an empty filter config. This filter config is
   // required to enable the fault injection features.
-  Listener BuildListenerWithStatefulSessionFilter() {
-    CookieBasedSessionState cookie_state;
-    cookie_state.mutable_cookie()->set_name(std::string(kCookieName));
+  Listener BuildListenerWithStatefulSessionFilter(
+      absl::string_view cookieName = kCookieName) {
```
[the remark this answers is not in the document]

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: I was working on replicating Richard's config as closely as possible. Would prefer to keep this.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc

eugeneo: I will add that in another PR

eugeneo (commented):
Please take another look

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
markdroth: As far as I can tell, the cookie name has absolutely no bearing on the bug we're fixing here, so I don't see why it's useful to add all of this additional plumbing for it in the test.  I won't insist on removing it, but IMHO this extra complexity does nothing except distract the reader from what's actually important here.

markdroth (approved):

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: Reverted to default cookie. Thank you for the review!

PR #34286, opened 2023-09-07
[server] Remove per-channel registered method map

ananda1066:


src/core/lib/surface/server.h
```diff
@@ -448,7 +434,10 @@ class Server : public InternallyRefCounted<Server>,
   bool starting_ ABSL_GUARDED_BY(mu_global_) = false;
   CondVar starting_cv_;
 
-  std::vector<std::unique_ptr<RegisteredMethod>> registered_methods_;
+  std::unordered_map<uint32_t /*hash of host and method*/,
+                     std::vector<std::unique_ptr<
```
ananda1066, line 438: @ctiller do we need a vector for each hash or can we assume there is only one registered method per host/method combo?

src/core/lib/surface/server.cc
```diff
@@ -1211,30 +1171,21 @@ void Server::ChannelData::InitTransport(RefCountedPtr<Server> server,
...
-    ChannelRegisteredMethod* rm =
-        &(*registered_methods_)[(hash + i) % registered_methods_->size()];
-    if (rm->server_registered_method == nullptr) break;
-    if (rm->has_host) continue;
-    if (rm->method != path) continue;
-    return rm;
+Server::RegisteredMethod* Server::ChannelData::GetRegisteredMethod(
+    const absl::string_view& host, const absl::string_view& path)
+    ABSL_LOCKS_EXCLUDED(registered_methods_mu_) {
+  MutexLock lock(&server_->registered_methods_mu_);
```
ctiller, line 1177: We'll need to figure out a way to avoid this lock: since it's on the per-request path it'll be a huge contention point.

One thing to consider is making `registered_methods_` a `shared_ptr<const flat_hash_map<...>>` - if we did that then when we register a new method we could acquire the lock, make a new `shared_ptr<const flat_hash_map<>>` and then install that into registered_methods_ -- and we could then have each channel pick up its own reference to the map.

The effect would be that newly registered methods *after* server instantiation would not be visible to channels, and maybe that's ok.

Another approach would be to crash on registering a method after the server has started -- if we did that, then we don't need a mutex to *read* from registered_methods_ because it's guaranteed immutable at that point.

src/core/lib/surface/server.cc
```diff
@@ -1211,30 +1171,21 @@ void Server::ChannelData::InitTransport(RefCountedPtr<Server> server,
...
+  if (server_->registered_methods_.empty()) return nullptr;
+  // check for exact match with host
+  auto key = MixHash32(absl::HashOf(host), absl::HashOf(path));
```
ctiller, line 1180: I don't think this is guaranteed unique.

We guarantee that a `pair<string,string>` of host & path is unique for a server, but the mix of the hash of the pair of strings is not.

We probably want to rephrase this so that registered_methods_ is a `flat_hash_map<pair<string,string>, ...>`

src/core/lib/surface/server.cc
```diff
@@ -1211,30 +1171,21 @@ void Server::ChannelData::InitTransport(RefCountedPtr<Server> server,
...
+  ;
```
ctiller, line 1181: nit: extra `;`

ctiller (changes requested):

src/core/lib/surface/server.cc
ananda1066: Took approach of crashing if we register after server started.

src/core/lib/surface/server.cc
ananda1066: Done.

ctiller:
Closing this because it's landed differently

PR #34290, opened 2023-09-08
[GSM Observability] Update opentelemetry cpp bazel deps

stanley-cheung:
This is needed for the upcoming work for GSM Observability integration testing.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -102,6 +102,14 @@ load("@com_github_google_benchmark//:bazel/benchmark_deps.bzl", "benchmark_deps"
 
 benchmark_deps()
 
+load("@io_opentelemetry_cpp//bazel:repository.bzl", "opentelemetry_cpp_deps")
+
+opentelemetry_cpp_deps()
+
+load("@io_opentelemetry_cpp//bazel:extra_deps.bzl", "opentelemetry_extra_deps")
+
+opentelemetry_extra_deps()
+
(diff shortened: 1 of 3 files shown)


stanley-cheung:
Looking into test failure

WORKSPACE
```diff
@@ -102,6 +102,14 @@ load("@com_github_google_benchmark//:bazel/benchmark_deps.bzl", "benchmark_deps"
 
 benchmark_deps()
 
+load("@io_opentelemetry_cpp//bazel:repository.bzl", "opentelemetry_cpp_deps")
```
yashykt, line 105: shouldn't this be in grpc_extra_deps?

stanley-cheung: hm, we tried that but it gives a cyclic dependency error.

stanley-cheung: I think that's because the opentelemetry repo calls `grpc_extra_deps()` - https://github.com/open-telemetry/opentelemetry-cpp/blob/main/WORKSPACE#L15

veblush: Not sure whether opentelemetry `WORKSPACE` file is evaluated when building gRPC targets from gRPC workspace. As long as application developer can use gRPC OpenTelemetry targets without calling additional functions other than grpc_deps` and `grpc_extra_deps`, this might be fine.

yashykt (approved):

PR #34292, opened 2023-09-08
[Python run_test] Fix run_test

XuanWang-Amos:
We're not running any test at all from `run_test.py` because of the way we filter test cases:
https://github.com/grpc/grpc/blob/1d136fd05fbd826024251b83bf8ba65c94963b1e/src/python/grpcio_tests/tests/_runner.py#L137

* `testcase_filter` is read from a json file (like [this one](https://github.com/grpc/grpc/blob/master/src/python/grpcio_tests/tests/tests.json)) and test name is similar to `unit._metadata_test.MetadataTest`.
* `case.id()` is loaded by `iterate_suite_cases` and will always have a prefix of `tests`, an example of case id will be: `tests.unit._metadata_test.MetadataTest`.

Because of the prefix, none of the test case will be matched thus we're not running any of the tests.

This PR fixes the prefix issue and all the regressions comes from not running tests using `run_test.py`.

#### Other Changes
* Added couple of `__init__.py` file since it's required to load tests.
* Added `py_status_code` to Aio rpc state.
  * `code()` is expecting to return a python gRPC code but current `status_code` is a Cython code.
* Added `libsqlite3-dev` to our dockers because it's required for `coverage==7.2.0`.
* Renamed csds and admin test because test case file have to end with `_test`: https://github.com/grpc/grpc/blob/1d136fd05fbd826024251b83bf8ba65c94963b1e/src/python/grpcio_tests/tests/_loader.py#L26
* Removed gevent test from `run_test.py` because Bazel gevent tests should be good enough for us.


gnossen:
> case.id() is loaded by iterate_suite_cases and will always have a prefix of tests, an example of case id will be: tests.unit._metadata_test.MetadataTest

Can we add an assertion to the sanity test that says something simple like `len(tests) > 1`? This is probably the lowest effort thing we can to help prevent future issues like this.

src/python/grpcio_tests/tests/_sanity/_sanity_test.py
```diff
@@ -39,9 +39,9 @@ def testTestsJsonUpToDate(self):
         )
 
         tests_json_string = pkgutil.get_data(self.TEST_PKG_PATH, "tests.json")
-        tests_json = tests_json_string.decode()
+        tests_json = json.loads(tests_json_string.decode())
 
-        self.assertSequenceEqual(tests_json, test_suite_names)
+        self.assertListEqual(tests_json, test_suite_names)
```
gnossen, line 44: Let me make sure I understand what the exact failure mode here was. This test looks like it _should_ fail. Is the issue that this test wasn't actually running at all due to the same test discovery issue?

src/python/grpcio_tests/tests_aio/unit/server_test.py
```diff
@@ -579,15 +582,18 @@ async def test_maximum_concurrent_rpcs(self):
         # Build the channel
         channel = aio.insecure_channel(bind_address)
         # Deplete the concurrent quota with 3 times of max RPCs
-        rpcs = []
+        rpc_tasks = []
         for _ in range(3 * _MAXIMUM_CONCURRENT_RPCS):
-            rpcs.append(channel.unary_unary(_BLOCK_BRIEFLY)(_REQUEST))
```
gnossen, line 584: Can you explain the change here in a bit more detail? `channel.unary_unary(_BLOCK_BRIEFLY)(_REQUEST)` should return a `UnaryUnaryCall`, [which is an Awaitable](https://github.com/grpc/grpc/blob/01907a776771eae2b2b5082fcf6d4ca2bd8386ce/src/python/grpcio/grpc/aio/_base_call.py#L138). So [`asyncio.wait`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait) should be able to accept these. Is the issue that the return type of `asyncio.wait` does not match up with the parameters expected by `self.loop.create_task`?

gnossen (commented):
Thanks for taking care of this Xuan! This is super important stuff.

src/python/grpcio_tests/tests_aio/unit/server_test.py
XuanWang-Amos: `asyncio.wait` does not accept `UnaryUnaryCall` anymore. This code works fine before 3.11 but will start getting this error in 3.11:
```
  File "/usr/local/google/home/xuanwn/.pyenv/versions/311env/lib/python3.11/site-packages/grpc/aio/_call.py", line 576, in _invoke
    serialized_response = await self._cython_call.unary_unary(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/call.pyx.pxi", line 331, in unary_unary
  File "src/python/grpcio/grpc/_cython/_cygrpc/aio/call.pyx.pxi", line 149, in grpc._cython.cygrpc._AioCall._set_status
  File "/usr/local/google/home/xuanwn/.pyenv/versions/3.11.0/lib/python3.11/asyncio/tasks.py", line 515, in _on_completion
    f.exception() is not None)):
    ^^^^^^^^^^^
AttributeError: 'UnaryUnaryCall' object has no attribute 'exception'
```

likely because [passing coroutine objects to asyncio.wait() directly is forbidden starting 3.11](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait).

In order to get it to work, we need create a task, and since `asyncio.create_task` only accept coroutine, we need a wrapper to wrap `UnaryUnaryCall`.

Reference:
* [Awaitable objects in Python after 3.10](https://stackoverflow.com/questions/76566490/awaitable-objects-in-python-after-3-10)

src/python/grpcio_tests/tests_aio/unit/metadata_test.py
```diff
@@ -183,9 +183,9 @@ async def _test_inspect_context(request, context):
         context.set_trailing_metadata(_TRAILING_METADATA)
 
         # ensure that we can read back the data we set on the context
-        assert context.get_code() == _NON_OK_CODE
```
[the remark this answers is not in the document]
gnossen: How did you discover this issue now then?

src/python/grpcio_tests/tests/_sanity/_sanity_test.py
gnossen: `assertListEqual` is fine. I just want to make sure I understand what was happening here.

src/python/grpcio_tests/tests_aio/unit/server_test.py
gnossen: I see. Python version-specific. Makes sense. Increases the urgency for multiversion Bazel tests.

src/python/grpcio_tests/tests_aio/unit/metadata_test.py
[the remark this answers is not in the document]
XuanWang-Amos: By accident, I'm fixing regression issues and noticed the error does not look right.

gnossen (approved):

PR #34293, opened 2023-09-08
[Test] Add concurrent test for session reuse

gtcooke94:
Add a test that runs concurrent requests using session caching.


test/core/end2end/h2_ssl_session_reuse_test.cc
```diff
@@ -263,9 +243,93 @@ TEST(H2SessionReuseTest, SingleReuse) {
...
+                               grpc_load_file(CLIENT_KEY_PATH, 1, &key_slice)));
+  const char* ca_cert =
+      reinterpret_cast<const char*> GRPC_SLICE_START_PTR(ca_slice);
+  const char* client_cert =
+      reinterpret_cast<const char*> GRPC_SLICE_START_PTR(cert_slice);
+  const char* client_key =
+      reinterpret_cast<const char*> GRPC_SLICE_START_PTR(key_slice);
+  grpc_ssl_pem_key_cert_pair signed_client_key_cert_pair = {client_key,
+                                                            client_cert};
+  grpc_channel_credentials* client_creds = grpc_ssl_credentials_create(
```
matthewstevenson88, line 262: Is it possible to move the creation of the `grpc_channel_credentials` to a helper function so we don't have to duplicate the code?

test/core/end2end/h2_ssl_session_reuse_test.cc
```diff
@@ -263,9 +243,93 @@ TEST(H2SessionReuseTest, SingleReuse) {
...
+      ca_cert, &signed_client_key_cert_pair, nullptr, nullptr);
+
+  do_round_trip(cq, server, server_addr.c_str(), cache, client_creds, false);
+  grpc_completion_queue* cq2 = grpc_completion_queue_create_for_next(nullptr);
+
+  std::thread* t1 = new std::thread([&]() -> void {
```
matthewstevenson88, line 325: It looks like you forgot to delete `t1`. You can make these local variables instead, e.g.

```
std::thread thread_1([&]() { do_round_trip(cq, server, server_addr.c_str(), cache, client_creds, true); });
```

matthewstevenson88, line 325: nit: Please use full names for variables, so e.g. replace `t1` with `thread_1` and similarly below.

matthewstevenson88 (commented):
Thanks for putting the PR together! Generally looks good, left a few comments.

test/cpp/end2end/concurrent_resumption_test.cc
```diff
@@ -0,0 +1,143 @@
...
+
+  std::shared_ptr<Channel> channel = grpc::CreateCustomChannel(
+      server_addr_, grpc::SslCredentials(sslOpts), channel_args);
+
+  auto stub = grpc::testing::EchoTestService::NewStub(channel);
+  grpc::testing::EchoRequest request;
+  grpc::testing::EchoResponse* response = nullptr;
+  request.set_message(kMessage);
+  ClientContext context;
+  context.set_deadline(grpc_timeout_milliseconds_to_deadline(100));
```
matthewstevenson88, line 114: To reduce some flakiness that I noticed while testing, I would bump this up to 1000.

test/cpp/end2end/concurrent_resumption_test.cc
```diff
@@ -0,0 +1,143 @@
...
+  ChannelArguments channel_args;
+  channel_args.SetPointer(std::string(GRPC_SSL_SESSION_CACHE_ARG), cache);
+  channel_args.SetSslTargetNameOverride("foo.test.google.fr");
+
+  std::shared_ptr<Channel> channel = grpc::CreateCustomChannel(
+      server_addr_, grpc::SslCredentials(sslOpts), channel_args);
+
+  auto stub = grpc::testing::EchoTestService::NewStub(channel);
+  grpc::testing::EchoRequest request;
+  grpc::testing::EchoResponse* response = nullptr;
```
matthewstevenson88, line 111: This turns out to be the bug. :)

Instead, you should do:
```
grpc::testing::EchoResponse response;
grpc::Status result = stub->Echo(&context, request, &response);
```

test/cpp/end2end/concurrent_resumption_test.cc
```diff
@@ -0,0 +1,143 @@
...
+  addr << "localhost:" << port;
+  server_addr_ = addr.str();
+  std::cout << "GREG: Serving on " << server_addr_ << "\n";
+  server_thread_ = new std::thread([&]() { RunServer(); });
+  sleep(1);
+
+  std::string root_cert = ReadFile(kCaCertPath);
+  std::string client_key = ReadFile(kClientKeyPath);
+  std::string client_cert = ReadFile(kClientCertPath);
+  grpc::SslCredentialsOptions sslOpts;
```
matthewstevenson88, line 96: nit: Per C++ style, this should be `ssl_options` instead.

test/cpp/end2end/concurrent_resumption_test.cc
```diff
@@ -0,0 +1,143 @@
...
+  //   }
+
+  TestServiceImpl service_;
+  std::unique_ptr<Server> server_;
+  std::thread* server_thread_;
+  std::string server_addr_;
+};
+
+TEST_F(GrpcResumptionTest, ConcurrentResumption) {
+  int port = grpc_pick_unused_port_or_die();
```
matthewstevenson88, line 85: Optional: You can make this more concise by rewriting these 4 lines as:
```
server_addr_ = absl::StrCat("localhost:", std::to_string(grpc_pick_unused_port_or_die()));
```

test/cpp/end2end/concurrent_resumption_test.cc
```diff
@@ -0,0 +1,143 @@
...
+  std::ostringstream addr;
+  addr << "localhost:" << port;
+  server_addr_ = addr.str();
+  std::cout << "GREG: Serving on " << server_addr_ << "\n";
+  server_thread_ = new std::thread([&]() { RunServer(); });
```
matthewstevenson88, line 90: To avoid some flakiness I noticed while playing with the test, you'll want to use an `absl::Notification` instead of `sleep(1)`. You can do this as follows:

```
#include "absl/synchronization/notification.h"

void RunServer(absl::Notification* notification) {
 ...
 server_ = builder.BuildAndStart();
 notification->Notify();
 server_->Wait();
}

TEST_F(GrpcResumptionTest, ConcurrentResumption) {
   ...
   absl::Notification notification;
   server_thread_ = new std::thread([this, &notification]() { RunServer(&notification); });
   notification.WaitForNotification();
   ...
}
```

You'll also need to add `"absl/synchronization"` to the `external_deps` of the `grpc_cc_test`.

test/cpp/end2end/concurrent_resumption_test.cc
```diff
@@ -0,0 +1,143 @@
+#include <memory>
```
matthewstevenson88, line 1: Could we make the name of this test file a bit more generic, so that we're free to use it for more e2e testing in the future? For example, `ssl_credentials_test.cc`?

matthewstevenson88 (commented):
Great work getting this PR together Greg! Left a few comments explaining where the bug was and a couple suggestions on how to reduce the flakiness of the test.

test/cpp/end2end/concurrent_resumption_test.cc
gtcooke94: Will definitely add that, the `sleep(1)` was my temporary solution while working on figuring on why the test wasn't working

test/cpp/end2end/concurrent_resumption_test.cc
gtcooke94: Thanks for catching this!

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+namespace {
+
+constexpr char kCaCertPath[] = "src/core/tsi/test_creds/ca.pem";
+constexpr char kServerCertPath[] = "src/core/tsi/test_creds/server1.pem";
+constexpr char kServerKeyPath[] = "src/core/tsi/test_creds/server1.key";
+constexpr char kClientCertPath[] = "src/core/tsi/test_creds/client.pem";
+constexpr char kClientKeyPath[] = "src/core/tsi/test_creds/client.key";
+constexpr char kMessage[] = "Hello";
+
+std::string ReadFile(const char* file_path) {
```
matthewstevenson88, line 31: micro-nit: Can we use `const std::string& file_path` for the parameter instead?

We can then pass `file_path.c_str()` to the `grpc_load_file` API.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+  grpc_slice slice;
+  GPR_ASSERT(
+      GRPC_LOG_IF_ERROR("load_file", grpc_load_file(file_path, 0, &slice)));
+  std::string file_contents(grpc_core::StringViewFromSlice(slice));
+  grpc_slice_unref(slice);
+  return file_contents;
+}
+
+class GrpcResumptionTest : public ::testing::Test {
```
matthewstevenson88, line 40: micro-nit: Should we maybe rename the test to something more generic, e.g. `SslCredentialsTest`, for future extensibility?

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+ protected:
+  void RunServer(absl::Notification* notification) {
+    std::string root_cert = ReadFile(kCaCertPath);
+    grpc::SslServerCredentialsOptions::PemKeyCertPair key_cert_pair = {
+        ReadFile(kServerKeyPath), ReadFile(kServerCertPath)};
+    grpc::SslServerCredentialsOptions ssl_options;
+    ssl_options.pem_key_cert_pairs.push_back(key_cert_pair);
+    ssl_options.pem_root_certs = root_cert;
+    ssl_options.force_client_auth = true;
```
matthewstevenson88, line 49: Can we leave a TODO (not in scope for this PR) to parametrize the test so that we test both the TLS and the mTLS cases?

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+        [&]() { DoRpc(server_addr_, ssl_options, cache, true); }));
+  }
+  for (int i = 0; i < 10; i++) {
+    threads[i]->join();
+  }
+  for (int i = 0; i < 10; i++) {
+    delete threads[i];
+  }
+
+  if (server_ != nullptr) {
```
matthewstevenson88, line 127: Can we move this if statement to a `TearDown` method? (This is the usual way this is done.)

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+    server_->Shutdown();
+    server_thread_->join();
+    delete server_thread_;
+  }
+  grpc_ssl_session_cache_destroy(cache);
+}  // namespace
```
matthewstevenson88, line 133: micro-nit: Please remove the "//namespace" comment here.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+
+  std::shared_ptr<Channel> channel = grpc::CreateCustomChannel(
+      server_addr, grpc::SslCredentials(ssl_options), channel_args);
+
+  auto stub = grpc::testing::EchoTestService::NewStub(channel);
+  grpc::testing::EchoRequest request;
+  grpc::testing::EchoResponse response;
+  request.set_message(kMessage);
+  ClientContext context;
+  context.set_deadline(grpc_timeout_milliseconds_to_deadline(1000));
```
matthewstevenson88, line 82: nit: Please add `/*time_ms=*/` in front of the `1000`.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+  grpc::Status result = stub->Echo(&context, request, &response);
+  EXPECT_TRUE(result.ok());
+  EXPECT_EQ(response.message(), kMessage);
+  std::shared_ptr<const AuthContext> auth_context = context.auth_context();
+  std::vector<grpc::string_ref> properties =
+      auth_context->FindPropertyValues(GRPC_SSL_SESSION_REUSED_PROPERTY);
+  ASSERT_EQ(1u, properties.size());
```
matthewstevenson88, line 89: nit: Prefer rewriting to avoid Yoda style, e.g. instead write `ASSERT_EQ(properties.size(), 1u);`. Similarly below.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+  std::string client_key = ReadFile(kClientKeyPath);
+  std::string client_cert = ReadFile(kClientCertPath);
+  grpc::SslCredentialsOptions ssl_options;
+  ssl_options.pem_root_certs = root_cert;
+  ssl_options.pem_private_key = client_key;
+  ssl_options.pem_cert_chain = client_cert;
+
+  grpc_ssl_session_cache* cache = grpc_ssl_session_cache_create_lru(16);
+
+  DoRpc(server_addr_, ssl_options, cache, false);
```
matthewstevenson88, line 114: nit: Please add `/*expect_session_reuse=*/` in front of `false`.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -0,0 +1,144 @@
...
+  std::vector<std::thread*> threads;
+  for (int i = 0; i < 10; i++) {
+    threads.push_back(new std::thread(
+        [&]() { DoRpc(server_addr_, ssl_options, cache, true); }));
```
matthewstevenson88, line 118: nit: Please add `/*expect_session_reuse=*/` in front of `true`.

matthewstevenson88 (approved):
Thanks for the PR!!

PR #34294, opened 2023-09-08
[OTel and Csm] Cosmetic API changes

yashykt:
Changes - 
* OTel API - Metrics enabling/disabling is updated to reflect design changes. 
* GSM is renamed to CSM


src/cpp/ext/gsm/gsm_observability.cc
markdroth, line 1: I think this file and the corresponding .h file need to be removed.

markdroth (approved):

src/cpp/ext/gsm/gsm_observability.cc
yashykt: Thanks!

PR #34310, opened 2023-09-11
[experiments] Bump expirations for some EE experiments

drfloob:
The work_stealing experiment may be deleted before the current deadline, but it's worth extending it just in case.

### DIFF
--- a/src/core/lib/experiments/experiments.yaml
+++ b/src/core/lib/experiments/experiments.yaml
@@ -67,7 +67,7 @@
   test_tags: [resource_quota_test]
 - name: event_engine_client
   description: Use EventEngine clients instead of iomgr's grpc_tcp_client
-  expiry: 2024/01/01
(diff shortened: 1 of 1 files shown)


src/core/lib/experiments/experiments.yaml
```diff
@@ -67,7 +67,7 @@
   test_tags: [resource_quota_test]
 - name: event_engine_client
   description: Use EventEngine clients instead of iomgr's grpc_tcp_client
-  expiry: 2023/10/01
+  expiry: 2024/01/01
```
ctiller, line 70: I would recommend pushing this a few weeks out... if it gets missed don't want to add work for folks during holidays.

drfloob: Good point! done.

Vignesh2208 (approved):

PR #34311, opened 2023-09-11
[CSM] Add target selector for csm channels

yashykt:

### DIFF
--- a/src/cpp/ext/csm/csm_observability.cc
+++ b/src/cpp/ext/csm/csm_observability.cc
@@ -20,8 +20,12 @@
 
 #include "src/cpp/ext/csm/csm_observability.h"
 
+#include <string>
 #include <utility>
 
+#include <grpc/support/log.h>
+
+#include "src/core/lib/uri/uri_parser.h"
 #include "src/cpp/ext/otel/otel_plugin.h"
 
 namespace grpc {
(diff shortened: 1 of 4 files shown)


src/cpp/ext/csm/csm_observability.cc
```diff
@@ -71,8 +75,27 @@ CsmObservabilityBuilder& CsmObservabilityBuilder::SetTargetAttributeFilter(
 
 absl::StatusOr<CsmObservability> CsmObservabilityBuilder::BuildAndRegister() {
   builder_.BuildAndRegisterGlobal();
+  builder_.SetTargetSelector(CsmChannelTargetSelector);
   return CsmObservability();
 }
 
+bool CsmChannelTargetSelector(absl::string_view target) {
+  auto uri = grpc_core::URI::Parse(target);
```
markdroth, line 83: Note that we support a [default prefix mechanism](https://github.com/grpc/grpc/blob/ce8838f05cb8d0374aba9edd06c7d38b0d5453a6/src/core/lib/resolver/resolver_registry.cc#L130), which means that the target specified by the application when they create a channel may not actually parse as a URI.  By default, [the default prefix is `dns:///`](https://github.com/grpc/grpc/blob/ce8838f05cb8d0374aba9edd06c7d38b0d5453a6/src/core/lib/resolver/resolver_registry.cc#L68), which means that users can just create a channel for `server.example.com:80`, and we'll treat that as `dns:///server.example.com:80`, even though the original string doesn't parse as a URI.

Is the string that we pass to this function the original target string used by the application, or is it already canonicalized to be a URI?  If it's the original string, then it may not parse as a URI.

In theory, there could be a case where the default prefix is set to `xds:`, in which case the URI could fail to parse here even if it is actually a CSM channel.  But there's no public API for setting the default prefix, so I don't think we really need to worry about this.

That having been said, if we are passing in the original target string that may not parse as a URI, then I think we should not emit the error log on line 85, because that will happen on a fairly routine basis with people omitting the `dns:///` prefix.

markdroth (approved):

src/cpp/ext/csm/csm_observability.cc
yashykt: We use the canonical form of the target (the one that gets inserted in GRPC_ARG_SERVER_URI)

yashykt: I can still remove the log but I think we'll always have the scheme added in, so should be fine.. unless we add in something like inproc to this as well without the scheme

PR #34312, opened 2023-09-11
[CSM] Add a server selector based on channel args

yashykt:
I've added channel args to `CreateNewServerCallTracer` on the `ServerCallTracerFactory`.
The motivation is for CSM Observability where the OTel plugin will be configured to only do stats on servers which are xDS enabled, so I plan to check this via channel args.
In the future, with the new scopes for metrics, I think I'll be able to change this to only check once per server or server connection instead of per call.


src/core/lib/surface/channel.h
```diff
@@ -174,6 +176,7 @@ class Channel : public RefCounted<Channel>,
   MemoryAllocator allocator_;
   std::string target_;
   const RefCountedPtr<grpc_channel_stack> channel_stack_;
+  const ChannelArgs channel_args_;
```
markdroth, line 179: I'd like input from @ctiller here, but it seems like it could significantly increase memory usage to store this here -- and it would affect both the client and server side, even if OT is not configured.

It doesn't seem like it would be very hard to add the layer we'd need to avoid this.  Just have `ServerCallTracerFactory` return a separate object for the connection, which we pass the channel args to and then save in the OT server filter.  Then we have *that* object return the `ServerCallTracer`.  That way, there'd be less memory increase, and it would happen only if OT is used.

src/cpp/ext/csm/csm_observability.cc
```diff
@@ -70,6 +73,11 @@ CsmObservabilityBuilder& CsmObservabilityBuilder::SetTargetAttributeFilter(
 }
 
 absl::StatusOr<CsmObservability> CsmObservabilityBuilder::BuildAndRegister() {
+  builder_.SetServerSelector([](const grpc_core::ChannelArgs& args) {
+    // Directly using the channel arg string to avoid a dependency on xDS.
```
markdroth, line 77: I'd prefer not to duplicate the string in two places, since it's possible that they could get out of sync.

I don't think we need a full xDS dependency to avoid this.  We can create a standalone header file that defines this channel arg key, and we can depend on that header in both the xDS code and here.

src/core/lib/surface/channel.h
yashykt: oh I think I've got a smarter method. Instead of getting it from the xds server config fetcher, I can add it to the server's channel args, and I think those are already available. Let me try it out

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
```diff
@@ -260,10 +260,11 @@ void PythonOpenCensusServerCallTracer::RecordEnd(
 
 grpc_core::ServerCallTracer*
 PythonOpenCensusServerCallTracerFactory::CreateNewServerCallTracer(
-    grpc_core::Arena* arena) {
+    grpc_core::Arena* arena, const grpc_core::ChannelArgs& args) {
```
XuanWang-Amos, line 263: Will the new xds channel arg be added for Python? Or is it already exist?

src/core/lib/surface/channel.h
yashykt: The new way works without adding any additional memory requirements.

> Just have ServerCallTracerFactory return a separate object for the connection, which we pass the channel args to and then save in the OT server filter.

Note that we don't have a filter for OT on the server side. If we had one, I could have just used the channel args from that filter. Note that I still want to add a layer for a `Server` scope, but it seems like it's not needed for this PR anymore :)

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
yashykt: Yeah, so what I'm thinking is that in the xDS server builder, we'll just add a channel arg that will be sent to the ServerCallTracer. When you start working on this piece, we can talk about it.

src/cpp/ext/csm/csm_observability.cc
yashykt: Done

src/core/ext/xds/xds_enabled_server.h
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#ifndef GRPC_SRC_CORE_EXT_XDS_XDS_ENABLED_SERVER_H
+#define GRPC_SRC_CORE_EXT_XDS_XDS_ENABLED_SERVER_H
+
+#include <grpc/support/port_platform.h>
```
markdroth, line 20: I don't think this include is actually needed.  I think our tools enforce it only if there is at least one include in the file, and in this case we don't need any.

src/core/BUILD
```diff
@@ -4178,6 +4178,15 @@ grpc_cc_library(
     deps = ["//:gpr_platform"],
 )
 
+grpc_cc_library(
+    name = "xds_enabled_server",
+    hdrs = [
+        "ext/xds/xds_enabled_server.h",
+    ],
+    language = "c++",
+    deps = ["//:gpr_platform"],
```
markdroth, line 4187: If you remove the include as per my other comment, then I think we don't think we need this dependency.

src/core/lib/surface/channel.cc
```diff
@@ -150,7 +150,6 @@ absl::StatusOr<RefCountedPtr<Channel>> Channel::CreateWithBuilder(
     compression_options.enabled_algorithms_bitset =
         *enabled_algorithms_bitset | 1 /* always support no compression */;
   }
-
```
markdroth, line 153: Please restore this blank line, so that there's no change needed to this file in this PR.

markdroth (approved):
Looks great!  Comments are all minor.

src/core/lib/surface/channel.cc
yashykt: Done

src/core/BUILD
yashykt: Done

src/core/ext/xds/xds_enabled_server.h
yashykt: Removed

src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -195,8 +196,14 @@ void OpenTelemetryServerCallTracer::RecordEnd(
 
 grpc_core::ServerCallTracer*
 OpenTelemetryServerCallTracerFactory::CreateNewServerCallTracer(
-    grpc_core::Arena* arena) {
-  return arena->ManagedNew<OpenTelemetryServerCallTracer>();
+    grpc_core::Arena* arena, const grpc_core::ChannelArgs& args) {
+  // Create the server call tracer only if there is no server selector
+  // registered or if the server selector returns true.
+  if (OTelPluginState().server_selector == nullptr ||
```
ctiller, line 202: Why are we looking at channel args on a per-call basis?

yashykt: Yeah, I was thinking about eventually having a Server-Channel scoped object tracer that would then create the per call `ServerCallTracer`..  Since we haven't agreed on the design there, I didn't want to jump the gun.

I guess, for now, I could create an API on the `ServerCallTracerFactory` for `IsServerChannelTraced()` and then save the bool on the Channel object? 

In the future, when we have more stuff on ServerChannel tracing, I could replace the bool with the `ServerChannelTracer`. What do you think?

markdroth: I don't think we should add anything to the `Channel` object.  That will increase per-channel memory on both client and server, even when OT is not in use.

I think we should just go ahead and create the server connection-scoped object.  None of this is public API yet, so we can change it later if we need to.

yashykt: > I don't think we should add anything to the Channel object. That will increase per-channel memory on both client and server, even when OT is not in use.

you're right. I got around this by just creating a `IsServerTraced()` method on the server call tracer factory, and saving it on the server. So, we only check it once per server.

src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -196,8 +197,18 @@ void OpenTelemetryServerCallTracer::RecordEnd(
 grpc_core::ServerCallTracer*
 OpenTelemetryServerCallTracerFactory::CreateNewServerCallTracer(
     grpc_core::Arena* arena) {
+  // Create the server call tracer only if there is no server selector
```
markdroth, line 200: I think this comment can go away now.

markdroth (approved):
This approach looks much better!  Just one minor nit, otherwise looks great!

src/cpp/ext/otel/otel_server_call_tracer.cc
yashykt: done

PR #34313, opened 2023-09-11
[ssa test] Add test for per-route SSA configuration

eugeneo:


test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -449,34 +482,57 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
       GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 1, kNumEchoRpcs / 3));
 }
 
-TEST_P(OverrideHostTest, EnablePerRoute) {
-  CreateAndStartBackends(2);
+TEST_P(OverrideHostTest, EnabledInRouteConfig) {
+  CreateAndStartBackends(1);
```
markdroth, line 486: I think this test needs two endpoints.  With only one endpoint, we're not actually checking that SSA is working, because the traffic would always go to the same endpoint even without the feature enabled.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -449,34 +482,57 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
...
   // All requests go to the backend we specified
-  CheckRpcSendOk(DEBUG_LOCATION, 5, RpcOptions().set_metadata(session_cookie));
+  CheckRpcSendOk(DEBUG_LOCATION, 5,
+                 RpcOptions().set_metadata({*session_cookie}));
   EXPECT_EQ(backends_[0]->backend_service()->request_count(), 5);
 }
 
+TEST_P(OverrideHostTest, DifferentPerRoute) {
+  constexpr absl::string_view kOverriddenCookie = "overridden-cookie-name";
+  CreateAndStartBackends(1);
```
markdroth, line 507: Same problem here.  Without multiple backends, we don't actually know that the SSA feature is working.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -449,34 +482,57 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
...
+      BuildStatefulSessionRouteConfig("/grpc.testing.EchoTestService/Echo2",
+                                      kOverriddenCookie);
+  *route_config.mutable_virtual_hosts(0)->add_routes() =
+      default_route_config_.virtual_hosts(0).routes(0);
+  SetListenerAndRouteConfiguration(
+      balancer_.get(), BuildListenerWithStatefulSessionFilter(), route_config);
+  balancer_->ads_service()->SetEdsResource(
+      BuildEdsResource(EdsResourceArgs({{"locality0", {CreateEndpoint(0)}}})));
+  WaitForAllBackends(DEBUG_LOCATION);
+  // Disabled for "echo1" method
```
markdroth, line 522: Please test that even with a cookie set, RPCs to echo1 will ignore the cookie and get distributed across multiple backends.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -449,34 +482,57 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
...
+  auto cookies = GetCookiesForBackend(
+      DEBUG_LOCATION, 0, 1, RpcOptions().set_rpc_method(METHOD_ECHO1));
+  ASSERT_EQ(cookies.empty() ? "" : cookies.front().name, "");
+  // Overridden for "echo2" method
+  cookies = GetCookiesForBackend(DEBUG_LOCATION, 0, 1,
```
markdroth, line 527: Please test that RPCs to echo2 honor the override cookie name but not the original cookie name.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -449,34 +482,57 @@ TEST_P(OverrideHostTest, ClusterGoneHostStays) {
...
+                                 RpcOptions().set_rpc_method(METHOD_ECHO2));
+  ASSERT_EQ(cookies.empty() ? "" : cookies.front().name, kOverriddenCookie);
+  // Default for "echo" method
+  cookies = GetCookiesForBackend(DEBUG_LOCATION, 0, 1,
```
markdroth, line 531: Please test that RPCs to the default route honor the original cookie name.

markdroth (commented):
Thanks for doing this!

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: Done

test/cpp/end2end/xds/xds_override_host_end2end_test.cc

test/cpp/end2end/xds/xds_override_host_end2end_test.cc

eugeneo (commented):
Thank you for your review. Updated the tests.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -153,6 +154,12 @@ class OverrideHostTest : public XdsEnd2endTest {
     return {};
   }
 
+  static std::pair<std::string, std::string> BuildCookieHeader(
```
markdroth, line 157: This could be a method on the `Cookie` struct.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -516,21 +522,57 @@ TEST_P(OverrideHostTest, DifferentPerRoute) {
...
+  ASSERT_EQ(echo_cookie.empty() ? "" : echo_cookie.front().name, kCookieName);
+  // Echo1 endpoint ignores cookies
+  CheckRpcSendOk(DEBUG_LOCATION, 6,
+                 RpcOptions()
+                     .set_metadata({
+                         BuildCookieHeader(echo_cookie.front()),
+                         BuildCookieHeader(echo2_cookie.front()),
+                     })
+                     .set_rpc_method(METHOD_ECHO1));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 3);
```
markdroth, line 549: Please also check that the second backend got the other half of the requests.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -516,21 +522,57 @@ TEST_P(OverrideHostTest, DifferentPerRoute) {
...
+  // Echo2 honours overwritten cookie
+  backends_[0]->backend_service()->ResetCounters();
+  CheckRpcSendOk(DEBUG_LOCATION, 6,
+                 RpcOptions()
+                     .set_metadata({BuildCookieHeader(echo_cookie.front())})
+                     .set_rpc_method(METHOD_ECHO2));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 3);
```
markdroth, line 556: Same here: the other backend should have gotten the other 3 requests.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -516,21 +522,57 @@ TEST_P(OverrideHostTest, DifferentPerRoute) {
...
+                 RpcOptions()
+                     .set_metadata({BuildCookieHeader(echo_cookie.front())})
+                     .set_rpc_method(METHOD_ECHO));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 6);
+  backends_[0]->backend_service()->ResetCounters();
+  CheckRpcSendOk(DEBUG_LOCATION, 6,
+                 RpcOptions()
+                     .set_metadata({BuildCookieHeader(echo2_cookie.front())})
+                     .set_rpc_method(METHOD_ECHO));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 3);
```
markdroth, line 575: Same here: the other backend should have gotten the other 3 requests.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -516,21 +522,57 @@ TEST_P(OverrideHostTest, DifferentPerRoute) {
...
+  // Echo1 endpoint ignores cookies
+  CheckRpcSendOk(DEBUG_LOCATION, 6,
+                 RpcOptions()
+                     .set_metadata({
+                         BuildCookieHeader(echo_cookie.front()),
+                         BuildCookieHeader(echo2_cookie.front()),
+                     })
+                     .set_rpc_method(METHOD_ECHO1));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 3);
+  // Echo2 honours overwritten cookie
```
markdroth, line 550: ...but not the cookie from the top-level config.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -516,21 +522,57 @@ TEST_P(OverrideHostTest, DifferentPerRoute) {
...
+                     .set_metadata({BuildCookieHeader(echo_cookie.front())})
+                     .set_rpc_method(METHOD_ECHO2));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 3);
+  backends_[0]->backend_service()->ResetCounters();
+  CheckRpcSendOk(DEBUG_LOCATION, 6,
+                 RpcOptions()
+                     .set_metadata({BuildCookieHeader(echo2_cookie.front())})
+                     .set_rpc_method(METHOD_ECHO2));
+  EXPECT_EQ(backends_[0]->backend_service()->request_count(), 6);
+  // Echo honours original cookie
```
markdroth, line 563: ...but not the override cookie.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -126,25 +122,18 @@ class OverrideHostTest : public XdsEnd2endTest {
...
   }
 
-  // Send requests until a desired backend is hit and returns cookie name/value
-  // pairs. Empty collection is returned if the backend was never hit.
-  // For weighted clusters, more than one request per backend may be necessary
-  // to obtain the cookie. max_requests_per_backend argument specifies
-  // the number of requests per backend to send.
-  std::vector<std::pair<std::string, std::string>>
-  GetAffinityCookieHeaderForBackend(
+  std::vector<Cookie> GetCookiesForBackend(
```
markdroth, line 125: Instead of having this return a vector, you could pass in the cookie name and have it return only the one you're actually interested in.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: Right! Done.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: Done.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: Done

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: Done

test/cpp/end2end/xds/xds_override_host_end2end_test.cc

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
eugeneo: I wanted to be able to assert in the test that there's no cookie under a different name. E.g. there should be no cookie with an original name when route config overwrites the name. Filtering by name would mean this method returns an `optional`, so code will not be more that much simpler.

eugeneo (commented):
Updated.

markdroth (approved):

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -59,6 +59,10 @@ class OverrideHostTest : public XdsEnd2endTest {
     std::string value;
     std::set<std::string> attributes;
 
+    std::pair<std::string, std::string> cookie_header() const {
```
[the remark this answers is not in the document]
eugeneo: Done! Thanks.

PR #34315, opened 2023-09-11
[EventEngine] Delete OriginalThreadPool, remove work_stealing experiment

drfloob:
This has been stable for a bit, everywhere that the EventEngine is enabled. Going forward, I think the event_engine_{client|listener} experiments can probably be used to regulate thread-pool-specific issues.


test/core/event_engine/thread_pool_test.cc
```diff
@@ -186,28 +184,19 @@ void ScheduleTwiceUntilZero(ThreadPool* p, std::atomic<int>& runcount, int n) {
...
-  // TODO(hork): this is nerfed due to the original thread pool taking eons to
-  // finish running 2M closures in some cases (usually < 10s, sometimes over
-  // 90s). Reset the branch factor to 20 when all thread pool runtimes
-  // stabilize.
   TypeParam p(8);
   std::atomic<int> runcount{0};
   // Our first thread pool implementation tried to create ~1M threads for this
   // test.
-  int branch_factor = 18;
+  int branch_factor = 20;
```
ctiller, line 191: what's the impact on runtime for this?

ctiller (approved):

test/core/event_engine/thread_pool_test.cc
drfloob: Trivial. Runtime with a 20x branch factor on Linux RBE is 180ms. With the OriginalThreadPool, it was minutes.

PR #34319, opened 2023-09-12
Fix crash issue of TTS-Service

ashrithaudpikar:
:Detailed Notes:
[RDX_CRASH][webos] /usr/sbin/tts-service
                   in __poll (libc.so.6 + 0xde830)
Program terminated with signal SIGSEGV, Segmentation fault.


yashykt:
I don't understand the intention here but this branch is not accepting new commits.. Feel free to open a PR against master with a rationale

PR #34321, opened 2023-09-12
[client_channel] second attempt: SubchannelWrapper hops into WorkSerializer before destruction

markdroth:
Original PR was #34307, reverted in #34318 due to internal test failures.

The first commit is a revert of the revert.  The second commit contains the fix.

The original idea here was that `SubchannelWrapper::Orphan()`, which is called when the strong refcount reaches 0, would take a new weak ref and then hop into the `WorkSerializer` before dropping that weak ref, thus ensuring that the `SubchannelWrapper` is destroyed inside the `WorkSerializer` (which is needed because the `SubchannelWrapper` dtor cleans up some state in the channel related to the subchannel).  The problem is that `DualRefCounted<>::Unref()` itself actually increments the weak ref count before calling `Orphan()` and then decrements it afterwards.  So in the case where the `SubchannelWrapper` is unreffed outside of the `WorkSerializer` and no other thread happens to be holding the `WorkSerializer`, the weak ref that we were taking in `Orphan()` was unreffed inline, which meant that it wasn't actually the last weak ref -- the last weak ref was the one taken by `DualRefCounted<>::Unref()`, and it wasn't released until after the `WorkSerializer` was released.

To this this problem, we move the code from the `SubchannelWrapper` dtor that cleans up the channel's state into the `WorkSerializer` callback that is scheduled in `Orphan()`.  Thus, regardless of whether or not the last weak ref is released inside of the `WorkSerializer`, we are definitely doing that cleanup inside the `WorkSerializer`, which is what we actually care about.

Also adds an experiment to guard this behavior.


ctiller:
Please wrap this in an experiment.

markdroth:
I've added an experiment for this.

ctiller (approved):

PR #34326, opened 2023-09-12
[ssa test] Test TTL attribute on cookie

eugeneo:


test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -217,8 +216,9 @@ class OverrideHostTest : public XdsEnd2endTest {
            num_requests;
   }
 
-  static Route BuildStatefulSessionRouteConfig(absl::string_view match_prefix,
-                                               absl::string_view cookie_name) {
+  static Route BuildStatefulSessionRouteConfig(
+      absl::string_view match_prefix, absl::string_view cookie_name,
+      absl::optional<google::protobuf::Duration> opt_duration = absl::nullopt) {
```
markdroth, line 221: Suggest passing this in as `grpc_core::Duration` instead of `google::protobuf::Duration` and then converting to protobuf form inside this function.  That will make it a bit more ergonomic for callers.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -94,7 +94,6 @@ class OverrideHostTest : public XdsEnd2endTest {
               std::string(it->second).c_str());
       values.emplace_back(ParseCookie(it->second));
       EXPECT_FALSE(values.back().value.empty());
-      EXPECT_THAT(values.back().attributes, ::testing::Contains("HttpOnly"));
```
markdroth, line 97: I think it's better to leave this here, so that we're checking this in every test.  We want this attribute to always be set, no matter what else we may be setting.

test/cpp/end2end/xds/xds_override_host_end2end_test.cc
```diff
@@ -262,11 +265,15 @@ TEST_P(OverrideHostTest, HappyPath) {
                          CreateEndpoint(1, HealthStatus::UNKNOWN)}}})));
   WaitForAllBackends(DEBUG_LOCATION);
   // Get cookie for backend #0.
-  auto session_cookie = GetAffinityCookieHeaderForBackend(DEBUG_LOCATION, 0);
-  ASSERT_TRUE(session_cookie.has_value());
+  auto cookies = GetCookiesForBackend(DEBUG_LOCATION, 0);
+  ASSERT_EQ(cookies.size(), 1);
```
markdroth, line 269: Suggest writing this as:

```
EXPECT_THAT(
    cookies, ::testing::ElementsAre(
        ::testing::AllOf(
            ::testing::Field(name, &Cookie::name, kCookieName),
            ::testing::Field(attributes, &Cookie::attributes,
                             ::testing::UnorderedElementsAre("HttpOnly")),
            ::testing::Field(value, &Cookie::value, ::testing::Not(::testing::IsEmpty())))));
```

markdroth (approved):

eugeneo (commented):
Thank you for the review. Really liked that matcher syntax.

PR #34327, opened 2023-09-12
[Build] re-enable grpc++_unsecure library without ssl dependencies

dawidcha:
This is a follow up to https://github.com/grpc/grpc/pull/34103

That pull request explicitly aimed to introduce shared library builds for Windows (DLLs) while effecting zero material change to the existing build pipelines. That aspiration meant that the grpc++_unsecure library had to be effectively excluded from the build (because including it would have also included a dependency on openssl, which makes no sense given its purpose)

This PR addresses that by:
* Extracting the single function in grpc_tls_certificate_provider with a dependency on openssl into a separate compilation unit
* Including that new .cc file into the grpc library
* Including grpc_tls_certificate_provider and one other source file into grpc_unsecure for the Windows DLL build only.
* Reinstating the grpc++_unsecure library which is a prerequisite for many tests.
* Regenerating all files affected by the changes in Bazel BUILD that introduce the new source file.

This change does affect the operation of other build pipelines - I have confirmed that it does not break the Linux Bazel build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,6 +441,10 @@ set(gRPC_UPB_GEN_DUPL_SRC
   src/core/ext/upb-generated/src/proto/grpc/health/v1/health.upb.c
   src/core/ext/upb-generated/src/proto/grpc/gcp/transport_security_common.upb.c
 )
+set(gRPC_ADDITIONAL_DLL_SRC
+  src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.cc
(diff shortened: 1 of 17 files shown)


veblush:
Thanks for the PR! It mostly looks good. Would you fix the Sanity checks and IWYU?

veblush (approved):
Thanks for the PR. All failing tests doesn't seem relevant.

veblush:
@yashykt Would you do a second review?

yashykt (approved):

PR #34345, opened 2023-09-14
[Build] Fix DLL build when run with gRPC_BUILD_TESTS=ON

dawidcha:
This is a follow up from https://github.com/grpc/grpc/pull/34327

The goal of this PR is to annotate the API with dllimport/export annotations to a sufficient extent that the gRPC package successfully compiles with gRPC_BUILD_TESTS=ON

As such, the bulk of the changes are to headers and c/c++ source to provide these annotations, this being done largely through trial and error.

A few other changes were needed to get this to work:
* A blanket application of the XXX_DLL_IMPORT macro for all compilation units caused conflicts for some test compilations. As a result, a more targeted approach was adopted such that these macros were only applied if the corresponding libraries were actually a dependency.
* A few of the tests had a dependency on one or more upb-generated files which as noted before can no longer be accessed from the grpc library. The approach taken was to include *all* upb-generated source files for those tests. I'm working on code for protobuf/upb which would permit dllimport/export annotations for upb-generated source and when/if that is accepted/merged we can dispense with all these duplications
* I added a new test: [test/distrib/cpp/run_distrib_test_dll_cmake.bat](https://github.com/grpc/grpc/compare/master...dawidcha:win_shared_tests?expand=1#diff-be63ff29b5644ed404b8a08db994400c553b28cbc545b4503a149880ae35a115) to confirm that the Windows DLL build is working - I'm not sure where I would need to add this in order to have it run as a validation?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,10 @@ set(PACKAGE_TARNAME       "${PACKAGE_NAME}-${PACKAGE_VERSION}")
 set(PACKAGE_BUGREPORT     "https://github.com/grpc/grpc/issues/")
 project(${PACKAGE_NAME} LANGUAGES C CXX)
 
+if(BUILD_SHARED_LIBS AND MSVC)
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+endif()
+
 set(gRPC_INSTALL_BINDIR "bin" CACHE STRING "Installation directory for executables")
 set(gRPC_INSTALL_LIBDIR "lib" CACHE STRING "Installation directory for libraries")
 set(gRPC_INSTALL_INCLUDEDIR "include" CACHE STRING "Installation directory for headers")
@@ -267,6 +271,8 @@ if(MSVC)
   set(_gRPC_C_CXX_FLAGS "${_gRPC_C_CXX_FLAGS} /wd4503")
   # Tell MSVC to build grpc using utf-8
   set(_gRPC_C_CXX_FLAGS "${_gRPC_C_CXX_FLAGS} /utf-8")
+  # Inconsistent object sizes can cause stack corruption and should be treated as an error
+  set(_gRPC_C_CXX_FLAGS "${_gRPC_C_CXX_FLAGS} /we4789")
 endif()
 if (MINGW)
   add_definitions(-D_WIN32_WINNT=0x600)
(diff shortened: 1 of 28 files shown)


veblush:
Thank you for the PR! You must have spent a lot of time figuring out where the annotation is needed.

One thing I want to know is that you need all tests in DLL? If you (and packages such as vcpkg) only need gRPC++ DLL, I think having a distrib test would be fine. Otherwise, we need to garden this annotation issue across the code. 

> I added a new test: [test/distrib/cpp/run_distrib_test_dll_cmake.bat](https://github.com/grpc/grpc/compare/master...dawidcha:win_shared_tests?expand=1#diff-be63ff29b5644ed404b8a08db994400c553b28cbc545b4503a149880ae35a115) to confirm that the Windows DLL build is working - I'm not sure where I would need to add this in order to have it run as a validation?

I think this `run_distrib_test_dll_cmake` sounds good. I need to set a corresponding one internally. Maybe I can do it first so that you can see the result in this PR.

dawidcha:
> One thing I want to know is that you need all tests in DLL? If you (and packages such as vcpkg) only need gRPC++ DLL, I think having a distrib test would be fine. Otherwise, we need to garden this annotation issue across the code.

I think that getting to a point where the tests compile, aside from just feeling like the right thing to do, also goes some way to addressing the question of what the minimal useful API is. In other words how much of the API needs to be exposed to be generally useful to consumers of the gRPC library who want to compile it as a DLL. I know what that minimal API looks like from our standpoint - and I have another PR ready to go that exposes the API points used in our code, but I'm also aware that we are something of an edge case in this respect.

dawidcha:
> I think this `run_distrib_test_dll_cmake` sounds good. I need to set a corresponding one internally. Maybe I can do it first so that you can see the result in this PR.

What needs to happen here?

veblush:
> > I think this `run_distrib_test_dll_cmake` sounds good. I need to set a corresponding one internally. Maybe I can do it first so that you can see the result in this PR.
> 
> What needs to happen here?

https://github.com/grpc/grpc/pull/34425 just got merged so I'd suggest having a separate PR to have a proper [run_distrib_test_cmake_for_dll.bat](https://github.com/grpc/grpc/blob/master/test/distrib/cpp/run_distrib_test_cmake_for_dll.bat). (This needs to use DLL properly) and you should be able to see the result of test in one of presubmit tests.

h-vetinari:
FWIW, this would be great to have for conda-forge. I've been wanting to [enable](https://github.com/conda-forge/grpc-cpp-feedstock/pull/311) more test coverage, and since we're building the shared libs on windows, that would require a build with `-DgRPC_BUILD_TESTS=ON` to compile (and pass the tests; or at least some subset). Perhaps that's not realistic right now, but it'd certainly be a great long-term goal. Either way, after this PR we can probably close https://github.com/grpc/grpc/issues/33032 and track remaining tasks in a separate issue.

dawidcha:
Regenerated CMakeLists.txt to fix the sanity test failure

dawidcha:
@veblush Failing tests don't seem related to the changes - are we seeing some transient failures again?

PR #34350, opened 2023-09-14
[OTel C++] Method attribute filtering

yashykt:
Implements method attribute filtering as defined by the gRFC https://github.com/grpc/proposal/pull/380

### DIFF
--- a/src/cpp/ext/otel/otel_call_tracer.h
+++ b/src/cpp/ext/otel/otel_call_tracer.h
@@ -100,7 +100,8 @@ class OpenTelemetryCallTracer : public grpc_core::ClientCallTracer {
 
   explicit OpenTelemetryCallTracer(OpenTelemetryClientFilter* parent,
                                    grpc_core::Slice path,
-                                   grpc_core::Arena* arena);
+                                   grpc_core::Arena* arena,
+                                   bool registered_method);
   ~OpenTelemetryCallTracer() override;
 
   std::string TraceId() override {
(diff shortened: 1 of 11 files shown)


src/core/lib/surface/server.cc
```diff
@@ -1209,6 +1210,11 @@ void Server::ChannelData::InitTransport(RefCountedPtr<Server> server,
   grpc_transport_perform_op(transport, op);
 }
 
+bool Server::ChannelData::IsRegisteredMethod(const grpc_slice& host,
+                                             const grpc_slice& path) {
+  return GetRegisteredMethod(host, path) != nullptr;
```
markdroth, line 1215: As we've discussed, the down-side of this is that we will wind up doing the registered method lookup twice for every server call, which seems sub-optimal for performance.  Can we have this code cache the result of `GetRegisteredMethod()` in a call context element somehow, so that the code in `Server::ChannelData::MakeCallPromise()` and `Server::CallData::StartNewRpc()` can just look at that value instead of calling `GetRegisteredMethod()` again?

ctiller: +1... we absolutely cannot do a second lookup here. Let's chat offline on how to do this.

yashykt: Yeah, adding it in another call context is relatively straightforward, but as you mentioned it would be a additional memory cost for all grpc calls, not just OTel enabled calls.

yashykt: Updated

src/cpp/ext/otel/otel_call_tracer.h
```diff
@@ -123,11 +124,14 @@ class OpenTelemetryCallTracer : public grpc_core::ClientCallTracer {
   void RecordAnnotation(absl::string_view /*annotation*/) override;
   void RecordAnnotation(const Annotation& /*annotation*/) override;
 
+  absl::string_view MethodForStats() const;
```
markdroth, line 127: I think this method can be private.

markdroth (approved):
Looks great!

src/cpp/ext/otel/otel_call_tracer.h
yashykt: Done

src/cpp/server/dynamic_thread_pool.h
markdroth, line 1: The changes in this file look like an accident.

yashykt: ha! sorry, yes

markdroth (approved):

PR #34361, opened 2023-09-15
[ObjC] Remove grpc core podspec module map

HannahShiSFB:
Not really needed, this should help the firestore upgrade issue

Tested with 
https://github.com/wu-hui/ReproGrpcCyclic


jtattermusch (approved):
LGTM as long as all the tests are green.

sampajano (approved):
thanks! :)

PR #34380, opened 2023-09-15
[Observability Testing] register prometheus exporter

stanley-cheung:
Working towards testing against CSM Observability. Added ability to register a prometheus exporter with our Opentelemetry plugin. This will allow our metrics to be available at the standard prometheus port `:9464`.

### DIFF
--- a/test/cpp/interop/observability_client.cc
+++ b/test/cpp/interop/observability_client.cc
@@ -20,6 +20,9 @@
 #include <unordered_map>
 
 #include "absl/flags/flag.h"
+#include "opentelemetry/exporters/prometheus/exporter_factory.h"
+#include "opentelemetry/exporters/prometheus/exporter_options.h"
+#include "opentelemetry/sdk/metrics/meter_provider.h"
 
 #include <grpc/grpc.h>
 #include <grpc/support/alloc.h>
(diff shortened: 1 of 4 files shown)


stanley-cheung:
Bleh, there's an external dependency build error with bazel 6 after we brought in these opentelemetry dependencies. I was hoping I can just downgrade my bazel to version 5 but apparently as long as my observability test stuff is a build target in these certain directories it will be checked by the repo bazel (at version 6) anyways. So looks like I do have to find a way to resolve that external error first...

yashykt (approved):

test/cpp/interop/observability_client.cc
```diff
@@ -214,6 +220,35 @@ int main(int argc, char** argv) {
...
+    // pods
+    opts.url = "0.0.0.0:9464";
+    auto prometheus_exporter =
+        opentelemetry::exporter::metrics::PrometheusExporterFactory::Create(
+            opts);
+    auto meter_provider =
+        std::make_shared<opentelemetry::sdk::metrics::MeterProvider>();
+    meter_provider->AddMetricReader(std::move(prometheus_exporter));
+    grpc::internal::OpenTelemetryPluginBuilder otel_builder;
+    otel_builder.EnableMetric(
```
yashykt, line 238: you don't need to set these

yashykt: they'll be enabled by default

yashykt: Just setting the meter provider and registering would be enough

stanley-cheung: Removed those lines. Thanks.

PR #34385, opened 2023-09-17
[ObjC] cf event engine supports resolve recursively from on_resolve callback

HannahShiSFB:
If the client calls LookupHostname again within the on_resolve callback, it re-acquires the `request_mu_` before releasing it which results in deadlock.

With this PR it extracts the request and releases the lock before calling on_resolve callback so it won't deadlock any more.


sampajano (approved):
Thanks for the fix, Hannah!

Issue #34135, opened 2023-08-23
Mac M2 Pro Max grpc-tools error: symbol not found in flat namespace '_CFRelease'


piotratlz:
### What version of gRPC and what language are you using?
gRPC 1.57.0 / Python

### What operating system (Linux, Windows,...) and version?
Mac M2 Pro Max

### What runtime / compiler are you using (e.g. python version or version of gcc)
Python 3.11.4

### What did you do?

I install grpc packages:
```bash
pip install grpcio grpcio-tools googleapis-common-protos
```
and generate proto code:
```bash
python -m grpc_tools.protoc --proto_path=$grpc_out ./*.proto --python_out=$grpc_out --grpc_python_out=$grpc_out
```

### What did you expect to see?
Generated code success message.

### What did you see instead?

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/piotr/.local/lib/python3.11/site-packages/grpc_tools/protoc.py", line 20, in <module>
    from grpc_tools import _protoc_compiler
ImportError: dlopen(/Users/piotr/.local/lib/python3.11/site-packages/grpc_tools/_protoc_compiler.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '_CFRelease'
```

### Anything else we should know about your project / environment?

`conda 23.7.3`

gnossen:
Closing as a duplicate of https://github.com/grpc/grpc/issues/33714

piotratlz:
Sorry @gnossen but this is not a duplicate. The issue you linked is related to grpcio but I have an issue with grpcio-tools.
I am still unable to generate Python code using grpc-tools package. Please reopen.

XuanWang-Amos:
Oh we might need take a look again. Did the two workarounds mentioned help?

piotratlz:
Unfortunately they did not.

XuanWang-Amos:
Ok I just tried again, I'm able to reproduce for grpcio-tools and both of the workarounds works for me, you'll need uninstall grpcio-tools (better uninstall grpcio too just to be safe).

After uninstall, you can install grpcio-tool again using one of the following ways:
1. Using Conda: `conda install grpcio-tools`
2. Install using pip: `GRPC_PYTHON_LDFLAGS=" -framework CoreFoundation" pip install grpcio-tools --no-binary :all:`

Tsopic:
[...]

Issue #33714, opened 2023-07-14
Unresolved symbols with PyPi grpcio installed in Conda environment on Mac OS


jjyao:
### What version of gRPC and what language are you using?

Grpc 1.56.0
Python

### What operating system (Linux, Windows,...) and version?
macOS Ventura, x86

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.

```
Successfully installed grpcio-1.56.0
(ray) jjyao@jjyao-C02FX025MD6T:~/Workspace/ray2 master$ python
Python 3.8.11 (default, Aug  6 2021, 08:56:27) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import grpc
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/envs/ray/lib/python3.8/site-packages/grpc/__init__.py", line 22, in <module>
    from grpc import _compression
  File "/opt/anaconda3/envs/ray/lib/python3.8/site-packages/grpc/_compression.py", line 20, in <module>
    from grpc._cython import cygrpc
ImportError: dlopen(/opt/anaconda3/envs/ray/lib/python3.8/site-packages/grpc/_cython/cygrpc.cpython-38-darwin.so, 0x0002): symbol not found in flat namespace '_kCFStreamPropertySocketNativeHandle'
```

### What did you see instead?

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

gnossen:
How was this artifact compiled? From source? From pypi (if so, which artifact specifically)? From Anaconda?

meixiaba:
I see the same error on MacOS  (Apple M2 Max. Ventura 13.4.1 ) too.

ImportError: dlopen(/Users/lechen/miniconda3/envs/py311/lib/python3.11/site-packages/grpc/_cython/cygrpc.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '_kCFStreamPropertySocketNativeHandle'

This error happened when I use pip inside conda, I tried different python env, py311, py3.8.

[...]

PR #34395, opened 2023-09-18
[Python dist] Explicitly link Core Foundation framework for MacOS

XuanWang-Amos:
Some user reported an issue that symbols from CoreFoundation are missing when install grpc/grpc-tools inside Conda:
* https://github.com/grpc/grpc/issues/33714
* https://github.com/grpc/grpc/issues/34135

This PR adds the `-framework CoreFoundation` flag so no workarounds are required from user.


XuanWang-Amos:
@gnossen PTAL. Tested locally with and without Conda, installing from source both works.

gnossen (approved):

PR #34396, opened 2023-09-18
[reland][EventEngine] Move combiner executor usage to EventEngine

drfloob:
Relands #31713


src/core/lib/iomgr/combiner.cc
```diff
@@ -164,15 +164,14 @@ static void move_next() {
...
-  push_last_on_exec_ctx(lock);
-}
-
 static void queue_offload(grpc_core::Combiner* lock) {
   move_next();
   GRPC_COMBINER_TRACE(gpr_log(GPR_INFO, "C:%p queue_offload", lock));
-  grpc_core::Executor::Run(&lock->offload, absl::OkStatus());
+  lock->event_engine->Run([lock] {
+    grpc_core::ApplicationCallbackExecCtx callback_exec_ctx;
+    grpc_core::ExecCtx exec_ctx;
```
ctiller, line 172: this needs to be initialized with flags=0, otherwise the `GRPC_EXEC_CTX_FLAG_IS_FINISHED` gets defaulted in, which will cause an immediate offload every offload

drfloob: Good catch, thanks.

ctiller (approved):

PR #34398, opened 2023-09-18
[Python 3.12] Support Python 3.12

XuanWang-Amos:
Support Python 3.12.

### Testing
* Passed all Distribution Tests.
* Also tested locally by installing 3.12 artifact.


tools/run_tests/artifacts/build_artifact_python.bat
```diff
@@ -19,9 +19,19 @@ set PATH=C:\%1;C:\%1\scripts;%PATH%
 set PATH=C:\msys64\mingw%2\bin;C:\tools\msys64\mingw%2\bin;%PATH%
 :end_mingw64_installation
 
+@rem check if building for Python3.7 in windows
```
gnossen, line 22: Interesting. Let's remove this when we drop 3.7 (which should be soon)

gnossen (approved):
Amazing job, Xuan! Thanks for making this happen!

rcodescu:
Does older grpcio versions support python3.12 as well? Any version older than 1.59 doesn't have a wheel built and building from source is still failing. I'm trying with the python3.12-alpine3.18 docker image.

PR #34402, opened 2023-09-19
[Python Documentation] fix a typo

CsterKuroi:

### DIFF
--- a/src/python/grpcio/README.rst
+++ b/src/python/grpcio/README.rst
@@ -60,7 +60,7 @@ package named :code:`python-dev`).
   $ git submodule update --init
 
   # For the next two commands do `sudo pip install` if you get permission-denied errors
-  $ pip install -requirements.txt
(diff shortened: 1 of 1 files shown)


XuanWang-Amos:
Nice catch! Do you mind sign the CLA so we can merge it?

src/python/grpcio/README.rst
```diff
@@ -60,7 +60,7 @@ package named :code:`python-dev`).
   $ git submodule update --init
 
   # For the next two commands do `sudo pip install` if you get permission-denied errors
-  $ pip install -rrequirements.txt
+  $ pip install -requirements.txt
```
gnossen, line 63: This should actually be `-r requirements.txt`. `requirements.txt` is an argument to the `-r` flag. I _believe_ it should be working as it currently is, but isn't the most aesthetically pleasing.

PR #34412, opened 2023-09-19
[CSM] Update labels

yashykt:
Changes - 
* Remove `csm.remote_workload_pod_name` and `csm.remote_workload_container_name`.
* Add `csm.remote_workload_name`, the value for which is sent through MetadataExchange, from the `CSM_WORKLOAD_NAME` env var. (Note that this is not added in local labels.)
* Add a local `csm.canonical_service` (@markdroth, please verify the key that we want here) that is read from `CSM_CANONICAL_SERVICE_NAME` env var, and we continue to send it over via MetadataExchange

### DIFF
--- a/src/cpp/ext/csm/metadata_exchange.cc
+++ b/src/cpp/ext/csm/metadata_exchange.cc
@@ -69,7 +69,7 @@ constexpr absl::string_view kMetadataExchangeCanonicalServiceKey =
     "canonical_service";
 // The keys that will be used for the local attributes when recording metrics.
 constexpr absl::string_view kCanonicalServiceAttribute =
-    "csm.canonical_service";
+    "csm.workload_canonical_service";
 constexpr absl::string_view kMeshIdAttribute = "csm.mesh_id";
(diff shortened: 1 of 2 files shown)


src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -70,13 +68,13 @@ constexpr absl::string_view kMetadataExchangeProjectIdKey = "project_id";
 constexpr absl::string_view kMetadataExchangeCanonicalServiceKey =
     "canonical_service";
 // The keys that will be used for the local attributes when recording metrics.
+constexpr absl::string_view kCanonicalServiceAttribute =
+    "csm.canonical_service";
```
markdroth, line 72: I think we want this to be `csm.workload_canonical_service`.

markdroth (approved):

src/cpp/ext/csm/metadata_exchange.cc
yashykt: done

PR #34416, opened 2023-09-20
[exec-ctx] Remove ScopedTimeCache from ExecCtx on iOS

dconeybe:
Fix a crash on older iOS versions due to problematic thread-local variable initialization.

See https://github.com/firebase/firebase-ios-sdk/issues/11509

Basically, there appears to be a bug in Xcode where it generates assembly for thread-local variable initialization that is susceptible to a crash. For example, on arm64 the generated assembly relies on registers like x8 and x10 being preserved by the thread-local variable initialization routine; however, in some cases this thread-local variable initialization calls functions like `ImageLoaderMachOCompressed::doBindFastLazySymbol` which clobber these registers, leaving their values indeterminate when the caller resumes. When those indeterminate values are later used as memory addresses they are invalid and result in a crash.

This PR works around this bug by removing the `ScopedTimeCache` member variable from the `ExecCtx` class on iOS. This is a reasonable workaround because `ScopedTimeCache` is only a slight optimization for data centers that entirely doesn't matter for mobile.

See https://github.com/dconeybe/TlsCrashIos12 for a demo of this crash.

Googlers see b/300501963 for full details.


dconeybe:
I don't have permissions to access the kokoro test results. Would you be able to give me access to the GCP project so I can view the test results?

sampajano:
> I don't have permissions to access the kokoro test results. Would you be able to give me access to the GCP project so I can view the test results?

@dconeybe Thanks so much for working on this!

I've seeing 2 failures right now but they don't really seem related to your change. (I don't know how to get your GCP access.. I think maybe you have to part of the grpc "organization" to see it? But i'm not sure i can add you to it.. @ctiller FYI in case you have an idea here :))

There is a merge conflict now though so you probably need to rebase your PR.

I'll run the tests for you again and we can probably merge it if the tests seem non-related to this.

Thanks!

ctiller (commented):
Leaving an LGTM here and letting Eryu sort out any build/test issues.

Also noting that I'm ok with the ifdef mess here primarily because we're planning on removing ExecCtx in the medium term - otherwise I'd be hoping we could find a cleaner approach.

ctiller (approved):

dconeybe:
Thank you for the review, @ctiller. I completely agree that the `#if` added by this PR are highly undesirable. We (Firestore users and the Firestore SDK development team) appreciate your cooperation in solving this unusual issue.

HannahShiSFB (approved):

HannahShiSFB:
LGTM, Thanks

sampajano (approved):
Thanks everyone!!

I've enabled auto merge and hope it'll be merged soon! 😃

PR #34419, opened 2023-09-20
[cmake build metadata] Categorize cmake C-core and C++ tests correctly

jtattermusch:
Currently most of tests built by cmake are categorized as C++ tests, even though they actually live under `test/core`.
The problem has historical reasons
- in the past the only tests that used gtest were C++ tests (since then many C-core tests also started using gtest)
- in the past tests had some implicit dependencies based on whether they were C-core tests (language="c" in build_autogenerated.yaml) or C++ tests (language="c++" in build_autogenerated.yaml) and tests that depended on gtest needed to treated as C++ tests. This has also changed.

Since neither of the above limitations exists anymore, we can rely on the categorization of C-core/C++ tests based on the existing default categorization, based on test path prefix: https://github.com/grpc/grpc/blob/490f6a3ee9a9c08f7f700842a285534fbb703df6/tools/buildgen/extract_metadata_from_bazel_xml.py#L955

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5518,6 +5518,7 @@ if(_gRPC_PLATFORM_LINUX OR _gRPC_PLATFORM_POSIX)
       third_party/googletest/googletest
       third_party/googletest/googlemock/include
       third_party/googletest/googlemock
+      ${_gRPC_PROTO_GENS_DIR}
   )
 
   target_link_libraries(alts_concurrent_connectivity_test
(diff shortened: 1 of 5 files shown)


jtattermusch:
The side effect of this change should be that the test build load will be split more evenly between `run_tests.py -l c --build_only` (which currently only builds a handful of tests) and `run_tests.py -l c++ --build_only` (which currently builds almost all the tests)

jtattermusch:
Failures that need to be looked at:
https://source.cloud.google.com/results/invocations/852d24cc-9954-4b91-a537-c59d2cf4ef9c/targets

jtattermusch:
One last remaining problem seems to be that "Basictests C windows" are now taking too long (because many tests have now become C-core tests again).

veblush:
Closing this as this PR didn't get make progress.

PR #34426, opened 2023-09-20
[pick_first] implement Happy Eyeballs

markdroth:


src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
```diff
@@ -607,14 +658,81 @@ void PickFirst::SubchannelList::SubchannelData::OnConnectivityStateChange(
...
+    // React to the connectivity state.
+    ReactToConnectivityStateLocked();
     return;
   }
-  // Ignore any other updates for subchannels we're not currently trying to
-  // connect to.
-  if (Index() != subchannel_list_->attempting_index_) return;
-  // React to the connectivity state.
-  ReactToConnectivityStateLocked();
+  // Otherwise, process connectivity state change.
```
eugeneo, line 676: nit: I feel like SubchannelData::OnConnectivityStateChange is becoming really long and hard to follow. Should it be split in several methods?

eugeneo (approved):

src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc
markdroth: I think it will be easier to read once the experiment goes away and we have only the new code to deal with.

PR #34428, opened 2023-09-21
[bazelified tests] Make bazelified C basictests build only

jtattermusch:
- make C-core basictests use `--build_only` when running as bazelified tests. This is because the volume of C core tests is expected to grow very significantly after https://github.com/grpc/grpc/pull/34419 and  currently the non-bazelified counterpart of the tests (the presubmit grpc_basictests_c_cpp_build_only job) is also "build only".
- make the linux presubmit job `grpc_basictests_c_cpp_build_only` a noop, since the bazelified tests already give the same coverage on presubmit.

### DIFF
--- a/tools/bazelify_tests/test/BUILD
+++ b/tools/bazelify_tests/test/BUILD
@@ -27,19 +27,19 @@ generate_bazel_distribtests(name = "bazel_distribtests_linux")
 
 # C/C++
 grpc_run_tests_harness_test(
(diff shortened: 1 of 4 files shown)


jtattermusch:
Basic Tests C/C++ Linux [Build Only] now takes 3mins: https://source.cloud.google.com/results/invocations/98bca4b5-8e29-48b6-9411-a0985412bad5/targets

drfloob (approved):

PR #34429, opened 2023-09-21
[bazelified tests] Bazelify tests from "linux/grpc_bazel_build" and make the original test job a noop.

jtattermusch:
Bazelify tests from "linux/grpc_bazel_build" kokoro job by creating 3 bazelified tests - "build with strict warning", "build with no_xds=True" and "build with no_xds=True negative test".

- also make the original "linux/grpc_bazel_build" kokoro job a no-op (since bazelified tests now provide the same coverage).

### DIFF
--- a/tools/bazelify_tests/grpc_run_simple_command_test.sh
+++ b/tools/bazelify_tests/grpc_run_simple_command_test.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# Copyright 2023 The gRPC Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+ARCHIVE_WITH_SUBMODULES="$1"
+shift
+
+# Extract grpc repo archive
+tar -xopf ${ARCHIVE_WITH_SUBMODULES}
+cd grpc
+
+# Remove the override the "do not detect toolchain" setting that was set
+# by the from .bazelrc configuration for the remote build.
+# TODO(jtattermusch): find a better solution to avoid breaking toolchain detection.
+export BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=0
+
+# Run command passed as args
+"$@"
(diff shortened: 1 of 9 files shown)


tools/bazelify_tests/grpc_run_simple_command_test.sh
```diff
@@ -0,0 +1,31 @@
...
+set -ex
+
+ARCHIVE_WITH_SUBMODULES="$1"
+shift
+
+# Extract grpc repo archive
+tar -xopf ${ARCHIVE_WITH_SUBMODULES}
+cd grpc
+
+# Remove the override the "do not detect toolchain" setting that was set
```
drfloob, line 25: Delete "Remove the"

tools/bazelify_tests/grpc_run_simple_command_test.sh
```diff
@@ -0,0 +1,31 @@
...
+# by the from .bazelrc configuration for the remote build.
```
drfloob, line 26: "by the from ..." should maybe just be "by .bazelrc for the remote build"

tools/bazelify_tests/test/bazel_build_with_grpc_no_xds_negative_test_linux.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+# Test that builds that need xDS do not build with --define=grpc_no_xds=true
```
drfloob, line 18: This is a bit tough to parse. I suggest something like "Ensure that tests that require XDS do not build with --define=grpc_no_xds=true"

jtattermusch:
"Bazel Basic build for C/C++" now takes ~3mins: https://source.cloud.google.com/results/invocations/ad5f2a72-a488-47f6-88ff-b2a796dff37f/targets

The newly added tests are here:
https://source.cloud.google.com/results/invocations/be31f365-66fa-4663-9af3-1b4d37c5e8a6/targets;query=bazel_build

Other advantages:
- each of the bazel builds now has a separate log (accessible via "target log" in the resultstore UI), so it's much easier to look at the build at determine the problem if there's a failure (compared to browsing one gigantic concatenated log that has multiple bazel invocations in it)
- each of the bazel build tests now has a separate success/failure status (so it's easy to see which one has failed, without influencing the other tests).

tools/bazelify_tests/grpc_run_simple_command_test.sh
jtattermusch: done.

tools/bazelify_tests/grpc_run_simple_command_test.sh
jtattermusch: done.

tools/bazelify_tests/test/bazel_build_with_grpc_no_xds_negative_test_linux.sh
jtattermusch: done.

jtattermusch:
@drfloob  all comments addressed, this should be good to go. PTAL.

drfloob (approved):

PR #34432, opened 2023-09-21
[RegisteredMethod] Set information on initial metadata about whether method is registered or not

yashykt:
Summary - 
On the server-side, we are changing the point at which we decide whether a method is registered or not from the surface to the transport at the point where we are done receiving initial metadata and before we invoke the recv_initial_metadata_ready closures from the filters. The main motivation for this is to allow filters to check whether the incoming method is a registered or not. The exact use-case is for observability where we only want to record the method if it is registered. We store the information about the registered method in the initial metadata.

On the client-side, we also set information about whether the method is registered or not in the outgoing initial metadata.

Since we are effectively changing the lookup point of the registered method, there are slight concerns of this being a potentially breaking change, so we are guarding this with an experiment to be safe.

Changes - 
* Transport API changes -
  * Along with `accept_stream_fn`, a new callback `registered_method_matcher_cb` will be sent down as a transport op on the server side. When initial metadata is received on the server side, this callback is invoked. This happens before invoking the `recv_initial_metadata_ready` closure.
* Metadata changes -
  * We add a new non-serializable metadata trait `GrpcRegisteredMethod()`. On the client-side, the value is a uintptr_t with a value of 1 if the call has a registered/known method, or 0, if it's not known. On the server side, the value is a (ChannelRegisteredMethod*). This metadata information can be used throughout the stack to check whether a call is registered or not.
* Server Changes -
  * When a new transport connection is accepted, the server sets `registered_method_matcher_cb` along with `accept_stream_fn`. This function checks whether the method is registered or not and sets the RegisteredMethod matcher in the metadata for use later.
* Client Changes -
  * Set the metadata on call creation on whether the  method is registered or not.

### DIFF
--- a/src/core/lib/experiments/experiments.yaml
+++ b/src/core/lib/experiments/experiments.yaml
@@ -206,7 +206,7 @@
   expiry: 2024/01/15
   owner: hork@google.com
   test_tags: []
-- name: rm_lookup_point_transport
+- name: registered_method_lookup_in_transport
   description:
     Change registered method's lookup point to transport
   expiry: 2024/03/31
--- a/src/core/lib/experiments/experiments.yaml
+++ b/src/core/lib/experiments/experiments.yaml
@@ -206,3 +206,9 @@
   expiry: 2024/01/15
   owner: hork@google.com
   test_tags: []
+- name: rm_lookup_point_transport
+  description:
+    Change registered method's lookup point to transport
+  expiry: 2024/03/31
+  owner: yashkt@google.com
+  test_tags: []
--- a/src/core/lib/experiments/experiments.yaml
+++ b/src/core/lib/experiments/experiments.yaml
@@ -211,4 +211,4 @@
     Change registered method's lookup point to transport
   expiry: 2024/03/31
   owner: yashkt@google.com
-  test_tags: []
+  test_tags: ["surface_registered_method_lookup"]
--- a/src/core/lib/experiments/rollouts.yaml
+++ b/src/core/lib/experiments/rollouts.yaml
@@ -100,5 +100,5 @@
   default: true
 - name: combiner_offload_to_event_engine
   default: true
-- name: rm_lookup_point_transport
+- name: registered_method_lookup_in_transport
   default: true
--- a/src/core/lib/experiments/rollouts.yaml
+++ b/src/core/lib/experiments/rollouts.yaml
@@ -100,3 +100,5 @@
   default: true
 - name: combiner_offload_to_event_engine
   default: true
+- name: rm_lookup_point_transport
+  default: true
--- a/src/core/lib/transport/transport.h
+++ b/src/core/lib/transport/transport.h
@@ -526,6 +526,8 @@ typedef struct grpc_transport_op {
   bool set_accept_stream = false;
   void (*set_accept_stream_fn)(void* user_data, grpc_transport* transport,
                                const void* server_data) = nullptr;
+  void (*set_registered_method_matcher_fn)(
+      void* user_data, grpc_core::ServerMetadata* metadata) = nullptr;
   void* set_accept_stream_user_data = nullptr;
   /// set the callback for accepting new streams based upon promises;
   /// this is a permanent callback, unlike the other one-shot closures.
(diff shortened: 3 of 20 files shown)


src/core/lib/experiments/rollouts.yaml
```diff
@@ -100,3 +100,5 @@
   default: true
 - name: combiner_offload_to_event_engine
   default: true
+- name: rm_lookup_point_transport
```
markdroth, line 103: Suggest calling this `registered_method_lookup_in_transport`.

src/core/lib/experiments/experiments.yaml
```diff
@@ -206,3 +206,9 @@
   expiry: 2024/01/15
   owner: hork@google.com
   test_tags: []
+- name: rm_lookup_point_transport
+  description:
+    Change registered method's lookup point to transport
+  expiry: 2024/03/31
+  owner: yashkt@google.com
+  test_tags: []
```
markdroth, line 214: Is there a set of tests that we should run with this experiment disabled, just to make sure that path continues to work until we remove it?

markdroth (approved):

src/core/lib/experiments/experiments.yaml
yashykt: Internal would be running with the experiment disabled, so we are fine?

markdroth: We can live with just that if we have to.  But if we can identify a set of tests to run in OSS, that seems better, because then we'd detect a problem before the PR is merged instead of blocking the import.

yashykt: I can disable the experiment for  `test/cpp/end2end:end2end_test` and that gives us coverage for chttp2 and interop, so that should be good enough?

src/core/lib/experiments/rollouts.yaml
yashykt: Done

src/core/lib/transport/transport.h
```diff
@@ -526,6 +526,8 @@ typedef struct grpc_transport_op {
   bool set_accept_stream = false;
   void (*set_accept_stream_fn)(void* user_data, grpc_transport* transport,
                                const void* server_data) = nullptr;
+  void (*set_registered_method_matcher_fn)(
```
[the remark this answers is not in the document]
yashykt: Done

src/core/lib/experiments/experiments.yaml
markdroth: Sounds fine.  Thanks!

yashykt: Done

test/cpp/end2end/end2end_test.cc
```diff
@@ -2337,8 +2337,12 @@ INSTANTIATE_TEST_SUITE_P(
 }  // namespace grpc
 
 int main(int argc, char** argv) {
-  grpc::testing::TestEnvironment env(&argc, argv);
+  // Use this test as a canary to make sure that method lookup in the surface
```
ctiller, line 2340: Please use the experiments framework to do this (by tagging this test and including it in the test_tags in the experiments.yaml file) - the method works and keeps things auditable, whereas individuals growing their own techniques does not.

yashykt: I would prefer to use that as well, but `test_tags` are for enabling experiments. I want to disable the experiment instead

yashykt: And it's mostly fine? since, when this experiment goes away, this line will go away as well

markdroth: Yash, the tests that have the tags specified in experiments.yaml get an additional run with the opposite of the default for the experiment.  So if the experiment is disabled by default, then those tests get an additional run with the experiment enabled, but conversely, if the experiment is enabled by default, then those tests get an additional run with the experiment disabled.

So I think Craig's right, we should use the tags in experiments.yaml for this.

yashykt: Yeah, I had a chat with Craig. I didn't realize that it disabled experiments if the default was true.

yashykt: Done

markdroth (approved):
Looks great!

PR #34435, opened 2023-09-21
[xds ssa] Remove environment variable protection for stateful affinity

eugeneo:


markdroth (approved):
Looks great!  Thanks for doing this!

gnossen:
This PR introduced a regression in the SSA interop tests. It is not safe to remove the guarding env var from the interop templates in the same PR as we disable the usage of the env var on master. It is only safe to do so once all branches tested using the feature have backported the removal of the env var and all of those branches have completed their image build.

[This](https://github.com/grpc/grpc/pull/34460) will fix things for the moment.

PR #34439, opened 2023-09-22
Revert "[ObjC] Remove grpc core podspec module map"

HannahShiSFB:
Reverts grpc/grpc#34361

Not sure why CI didn't capture it and why my previous test works, but looks like it broke at least src/objective-c/examples/SwiftSample build.
Will see how CI goes in this PR.

### DIFF
--- a/gRPC-Core.podspec
+++ b/gRPC-Core.podspec
@@ -65,6 +65,9 @@ Pod::Spec.new do |s|
   # following lets users write `#include <grpc/grpc.h>`.
   s.header_dir = name
 
+  # The module map created automatically by Cocoapods doesn't work for C libraries like gRPC-Core.
+  s.module_map = 'include/grpc/module.modulemap'
+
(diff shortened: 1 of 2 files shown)


drfloob:
@sampajano If breakages here merit rollbacks, then it's probably worth looking into why this test is not exercised on PRs.

drfloob (approved):

sampajano:
> @sampajano If breakages here merit rollbacks, then it's probably worth looking into why this test is not exercised on PRs.

@drfloob Agreed!

@HannahShiSFB Is this something you would be able to help check? Thanks!

sampajano (approved):
Thanks!

HannahShiSFB:
> > @sampajano If breakages here merit rollbacks, then it's probably worth looking into why this test is not exercised on PRs.
> 
> @drfloob Agreed!
> 
> @HannahShiSFB Is this something you would be able to help check? Thanks!

I can't see the inside of the CI. Could Jan @jtattermusch maybe?

PR #34444, opened 2023-09-22
[Test] Changed gcc-7 to gcc-8

veblush:
Bumping gcc 7 to 8 to workaround the ongoing gcc segfault problem when building Protobuf C++. Currently Foundational C++ requires gcc 7 so this is a temporary measure to make the test green. We need to either make a decision to change the minimum version of gcc in the Foundational C++ or find a way to support gcc 7 without gcc segfault soon.


eugeneo (approved):

jtattermusch:
This breaks the [Bazel RBE Non-Bazel Tests](https://source.cloud.google.com/results/invocations/b3f6fcf0-55b1-40d7-a728-e3449f5f06ce/targets) completely by not updating 
https://github.com/grpc/grpc/blob/98704e7e4d5167fa2497e76fb28b88fc39be709a/tools/bazelify_tests/test/portability_tests.bzl#L56

PR #34452, opened 2023-09-22
[PSM Interop] Rename GAMMA tests to CSM tests

sergiitk:
Notes:
- GAMMA tests are now a subset of a wider CSM scope
- Some "gamma" files to ensure graceful renames
- Should be safe to merge as is


gnossen (approved):
Looks great! Thanks for doing this Sergii!

PR #34454, opened 2023-09-22
[Python 3.12] Increase test timeout

XuanWang-Amos:
Passed manual run: https://fusion2.corp.google.com/invocations/0e337438-8573-4b54-ba1c-242b1ad58586

### DIFF
--- a/tools/run_tests/artifacts/distribtest_targets.py
+++ b/tools/run_tests/artifacts/distribtest_targets.py
@@ -193,6 +193,7 @@ class PythonDistribTest(object):
                 % (self.docker_suffix, self.arch),
                 "test/distrib/python/run_source_distrib_test.sh",
                 copy_rel_path="test/distrib",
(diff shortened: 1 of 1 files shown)


gnossen:
A few data points here. I ran this locally and it passed in just a few minutes. Peeking inside the docker image, I can confirm that we are still getting build parallelization.

I SSH'd to the build executor to check out what might be going on there since the build seems to be significantly slower there. We're using a 32 core machine with each core at around 2.2GHz. It is at 100% capacity for the duration of this build, with 4 distribtest containers running at the same time. This means this is almost certainly a legitimate timeout just caused by how much code is being compiled and how many resources we've devoted to the task.

The solution is either to increase the individual task timeout as we've done here, or to decrease the level of parallelization between different docker image runs so that each individual run will take less time. For the moment, I think this solution is a good one.

gnossen (approved):

Issue #32598, opened 2023-03-11
Improve RPC failure status message when TSI handshaker creation fails


markdroth:
This is related to #22883, but I'm splitting it off into a separate issue since it relates to the security code.  This will build on improvements previously made in #30077.

We recently discovered a case where RPCs started failing with the message ["Failed to create security handshaker"](https://github.com/grpc/grpc/blob/ac7faf75ba9b3c373a68acd18f08eb974e603ad5/src/core/lib/security/transport/security_handshaker.cc#L595), which isn't a very useful message.  This happens [when the TSI handshaker is null](https://github.com/grpc/grpc/blob/ac7faf75ba9b3c373a68acd18f08eb974e603ad5/src/core/lib/security/transport/security_handshaker.cc#L660), but the current API does not pass through an error message to say *why* we failed to create the TSI handshaker.  That's the thing we need to fix here.

The first part of this change will be easy: we need to change the [`SecurityHandshakerCreate()` function](https://github.com/grpc/grpc/blob/ac7faf75ba9b3c373a68acd18f08eb974e603ad5/src/core/lib/security/transport/security_handshaker.cc#L655) to take the TSI handshaker as `absl::StatusOr<tsi_handshaker*>`, so that we can pass in a status instead of a handshaker if we fail to create the handshaker.  The function can then pass this status to the `FailHandshaker` instance, so that it can be returned from the handshaker.  From there, it will make its way into RPC failure status messages via the client channel code.

The second part of this change will be a bit more work, which is to change all of the TSI implementations to return a useful status indicating the underlying failure.  Looking at the TLS TSI handshaker as an example, it looks like it will fail to create the handshaker in the following cases:
[...]

Issue #30228, opened 2022-07-06
add C++ e2e tests for TSI handshaker errors


markdroth:
As a follow-up to #30077, we should write some C++ end-to-end tests that show that we fail RPCs with the right status messages when TSI handshaking fails in various ways.

PR #34455, opened 2023-09-22
[ssl] Improve logging for SSL hostname check and ALPN negotiation errors.

matthewstevenson88:
I've been asked to help debug several issues over the last few weeks, and these logs would have helped quickly identify the root cause.

### DIFF
--- a/src/core/lib/security/security_connector/ssl_utils.cc
+++ b/src/core/lib/security/security_connector/ssl_utils.cc
@@ -135,7 +135,7 @@ grpc_error_handle grpc_ssl_check_alpn(const tsi_peer* peer) {
         "Cannot check peer: missing selected ALPN property.");
   }
   if (!grpc_chttp2_is_alpn_version_supported(p->value.data, p->value.length)) {
-    gpr_log(GPR_INFO, "Cannot check peer: invalid ALPN vlaue.");
+    gpr_log(GPR_INFO, "Cannot check peer: invalid ALPN value.");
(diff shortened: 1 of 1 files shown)


markdroth:
I don't think we want to unconditionally log like this.  This is going to really flood logfiles with meaningless error messages in a number of cases, especially once I land my Happy Eyeballs change (#34426).

In general, if the security handshake fails, TSI should return an error message from the handshaker, and gRPC should ensure that the error message winds up as part of the status failure message for any RPC failed as a result of that connection failure (see #22883).  I did a bit of work to improve this in #30077, but there's still more work to be done here (see #32598 and #30228 for a couple of missing things, and there are probably others).  I would ultimately like to see the TSI API revamped such that it returns `absl::Status` for failures instead logging errors and returning a enum value; libraries should really never log in this kind of low-level code.

If looking at the RPC status failure message is not sufficient for debugging purposes, another option would be to look at channelz.  That should show you the failure messages for individual subchannel connection failures.

If you really absolutely need log output to debug things, then you should use a [`TraceFlag`](https://github.com/grpc/grpc/blob/c76730114a952f9732e4ae68d6bdb3c1a7c13973/src/core/lib/debug/trace.h#L51) to guard the log output, so that users don't see it unless they explicitly enable it.

matthewstevenson88:
Thanks for the feedback Mark. I think having an "ssl" trace flag may be a useful feature. I'll close this PR and circle back.

PR #34459, opened 2023-09-23
[test] Remove grpc core podspec module map

HannahShiSFB:


sampajano (approved):
@HannahShiSFB Thanks!!

Could you please explain the reason for this change and its effects / implications?

Thakns!

HannahShiSFB:
This is only for test, it does fail now, but for the earlier PR with the same changes, it succeeded without "Basic Tests ObjC Examples" being executed.

PR #34485, opened 2023-09-26
[EventEngine] Extend the `event_engine_dns` experiment

yijiem:

### DIFF
--- a/src/core/lib/experiments/experiments.yaml
+++ b/src/core/lib/experiments/experiments.yaml
@@ -112,7 +112,7 @@
 - name: event_engine_dns
   description:
     If set, use EventEngine DNSResolver for client channel resolution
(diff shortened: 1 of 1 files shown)


src/core/lib/experiments/experiments.yaml
```diff
@@ -112,7 +112,7 @@
 - name: event_engine_dns
   description:
     If set, use EventEngine DNSResolver for client channel resolution
-  expiry: 2023/10/01
+  expiry: 2024/01/01
```
drfloob, line 115: please push this out a few weeks into January, so nobody gets nagged to act on an expiry over the winter holidays.

yijiem: Makes sense, done.

drfloob (approved):

PR #34496, opened 2023-09-27
[python] fix empty rpc error

Willjay90:
Currently, the bidirectional stream exception show empty error.

```
def YourBidirectionalStreamingRPCMethod(self, request_iterator, context):
    try:
        for request in request_iterator:
            # Your logic to process the incoming request and generate responses
            response = your_response_generator(request)
            yield response
    except grpc.RpcError as rpc_error:
        print(f"Error: {rpc_error}")  # no error will be shown now, after the change, if cancelled, we will see `cancelled`
```
No error message in `grpc.RpcError()`
```
class RpcError(Exception):
    """Raised by the gRPC library to indicate non-OK-status RPC termination."""
```
```
def _raise_rpc_error(state: _RPCState) -> None:
    rpc_error = grpc.RpcError() # no error message here
    state.rpc_errors.append(rpc_error)
    raise rpc_error
```

### DIFF
--- a/src/python/grpcio/grpc/_server.py
+++ b/src/python/grpcio/grpc/_server.py
@@ -154,7 +154,7 @@ class _RPCState(object):
 
 
 def _raise_rpc_error(state: _RPCState) -> None:
-    rpc_error = grpc.RpcError()
+    rpc_error = grpc.RpcError(state.client)
     state.rpc_errors.append(rpc_error)
     raise rpc_error
 


gnossen:
So the issue here ultimately comes down to "the `__str__` and/or `__repr__` methods of `grpc.RpcError` are not giving me as much information as I want." That opens up the discussion to "what information _do_ we want out of these exceptions?" I'm not sure that the `state.client` string on its own is a good answer to that question.

@Willjay90 Do you have any thoughts about this?

Willjay90:
From my understanding, since bidirectional stream is using [_RequestIterator](https://github.com/grpc/grpc/blob/cf0d1094a1d36adee12f21758ca56c5d849903c2/src/python/grpcio/grpc/_server.py#L474) which will monitor the `_RPCState`  status change and raise exception based on `state.client` (as far as I know this is current [status](https://grpc.github.io/grpc/core/md_doc_statuscodes.html))
```
if self._state.client is _CANCELLED:
     _raise_rpc_error(self._state)
```
However, when we're raising the exception, we raise an empty exception.
```
def _raise_rpc_error(state: _RPCState) -> None:
    rpc_error = grpc.RpcError() # no error message here
    state.rpc_errors.append(rpc_error)
    raise rpc_error
```
The logic above looks like the following code to me according to [grpc.RpcError](https://grpc.github.io/grpc/python/_modules/grpc.html#RpcError)'s definition.
```
def _raise_rpc_error(state: _RPCState) -> None:
    rpc_error = Exception()
    state.rpc_errors.append(rpc_error)
    raise rpc_error
```

Ideally, we should give reason about the failure.
For example, i started a grpc server and a client and I quit the client (ctrl+c) and currently I'm only able to see `state.client` as the reason of failure.
```
def _raise_rpc_error(state: _RPCState) -> None:
    print(state.code, state.client, state.details) # None, cancelled, None
    rpc_error = grpc.RpcError()
    state.rpc_errors.append(rpc_error)
    raise rpc_error
```

That's the reason why I put `state.client`, which is `cancelled` in this case, in the Exception (without modifying `grpc.RpcError` class)

Ideally, we should raise exception with `code` and `description` which can be referenced [here](https://grpc.github.io/grpc/core/md_doc_statuscodes.html)

Another confusion part to me is that somehow we are managing the state `state.rpc_errors.append(rpc_error)` but raise another object (`grpc.RpcError`)

My suggestion is either return the failure which is `state.client` or probably rethink what should be raised, instead of `raise Exception()`

XuanWang-Amos:
> "rethink what should be raised"

We agree with it, but in this case a more clear message like "Client has cancelled the call" will help more.

PR #34497, opened 2023-09-27
[EventEngine] Fix WinSocket/IOCP notification race

drfloob:
It was possible for threads that call `WinSocket::NotifyOnRead` and `WinSocket::NotifyOnWrite` to race against IOCP poller threads, causing poller events to be missed.

In the most common usage, in some thread (E), the Endpoint would make an async (overlapped) read or write using `WSARecv` or `WSASend` respectively, then use the socket's `NotifyOn*` methods to have callbacks executed when data was ready. If data was already available, those callbacks would be scheduled for execution immediately. Meanwhile, if overlapped events came in for some socket, some IOCP poller thread (P) would inform the socket that data was ready, and if notification callbacks were already present, they would be scheduled for execution immediately. It was possible for thread (E) to see no data available, and thread (P) to not see any notification callbacks registered. This resulted in registered callbacks that would never be called, for data that had already been received.

### DIFF
--- a/src/core/lib/event_engine/windows/win_socket.cc
+++ b/src/core/lib/event_engine/windows/win_socket.cc
@@ -93,12 +93,20 @@ void WinSocket::NotifyOnReady(OpState& info, EventEngine::Closure* closure) {
     thread_pool_->Run(closure);
     return;
   };
-  if (std::exchange(info.has_pending_iocp_, false)) {
-    thread_pool_->Run(closure);
-  } else {
-    EventEngine::Closure* prev = nullptr;
-    GPR_ASSERT(info.closure_.compare_exchange_strong(prev, closure));
+  {
+    grpc_core::MutexLock lock(&info.ready_mu_);
+    if (!std::exchange(info.has_pending_iocp_, false)) {
+      // No overlapped results have been returned for this socket. Assert that
+      // there is no notification callback yet, set the notification callback,
+      // and return.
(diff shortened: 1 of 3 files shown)


drfloob:
Passes internal Windows build.

src/core/lib/event_engine/windows/win_socket.cc
```diff
@@ -93,12 +93,20 @@ void WinSocket::NotifyOnReady(OpState& info, EventEngine::Closure* closure) {
...
-    EventEngine::Closure* prev = nullptr;
-    GPR_ASSERT(info.closure_.compare_exchange_strong(prev, closure));
+  {
+    grpc_core::MutexLock lock(&info.ready_mu_);
+    if (!std::exchange(info.has_pending_iocp_, false)) {
+      // No overlapped results have been returned for this socket. Assert that
+      // there is no notification callback yet, set the notification callback,
+      // and return.
+      EventEngine::Closure* prev = nullptr;
+      GPR_ASSERT(info.closure_.compare_exchange_strong(prev, closure));
```
ctiller, line 103: Seems like we could do this without an atomic now?

drfloob: Indeed. Fixed.

drfloob: Passes the linter.

ctiller (approved):

PR #34499, opened 2023-09-27
[Build] Add missing absl includes

DanilChapovalov:
To avoid depending on transitive includes, specially avoid relying on transitive include status.h -> str_cat.h that is removed in the latest version of abseil

### DIFF
--- a/src/core/lib/iomgr/event_engine_shims/endpoint.cc
+++ b/src/core/lib/iomgr/event_engine_shims/endpoint.cc
@@ -18,7 +18,9 @@
 #include <atomic>
 #include <memory>
 
+#include "absl/functional/any_invocable.h"
 #include "absl/status/status.h"
+#include "absl/status/statusor.h"
 #include "absl/strings/string_view.h"
 
 #include <grpc/event_engine/event_engine.h>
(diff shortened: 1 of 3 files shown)


drfloob:
How did you identify these were missing? Did you have a build failure? Or are you trying to compile against a newer abseil than what gRPC currently supports?

iwyu runs on every PR, I suspect it will decide that these includes are not necessary. But I'll trigger the CI and we'll see.

DanilChapovalov:
I got compilation errors in grpc while trying to import latest version of abseil into chromium.
Then I look into files that failed to compile and manually search for absl symbols file is using. That are the includes I'm adding. I haven't check all the files, just the files that are blocking my import.

DanilChapovalov:
Example of the build error:
https://ci.chromium.org/ui/p/chromium/builders/try/android-x64-cast/371969/overview

drfloob:
Interesting, alright then. IWYU is happy, but clang format is complaining. You'll need to run `tools/distrib/clang_format_code.sh` to get the includes sorted correctly.

drfloob (approved):

drfloob:
Looks good. All of the breaking tests are unrelated, there are know issues being fixed right now. They are not build errors.

apolcyn (approved):

PR #34500, opened 2023-09-27
[build] Add compiler checks for alignment problems

drfloob:


tools/bazel.rc
```diff
@@ -140,6 +140,7 @@ build:ubsan --copt=-DGRPC_UBSAN
 build:ubsan --copt=-DNDEBUG
 build:ubsan --copt=-fno-sanitize=function,vptr
 build:ubsan --linkopt=-fsanitize=undefined
+build:ubsan --define=use_strict_warning=true
```
veblush, line 143: Is this required? If it isn't, we don't want to have this as we cannot control what version of clang for ubsan, giving a different result.

drfloob: We can remove it. But it is the reason we caught the unused variables. Interestingly, UBSAN uses clang-15 at the moment. https://btx-internal.corp.google.com/invocations/07d3f537-6149-45dc-bd54-9bab403dc47d/log

drfloob: Removed.

drfloob:
> I'm curious how UBSAN builds use these warning flags. `use_strict_warning=1` will be passed somewhere else?

They'd be used as any other. And I often enable strict warnings on my workstation.

veblush (approved):

PR #34503, opened 2023-09-27
[EventEngine] Fix ThreadCount aligned storage for C++14 and earlier

drfloob:


gnossen:
Will need to be backported to v1.59.x

ctiller (approved):

PR #34511, opened 2023-09-27
[PSM Interop] Double the active channel timeout for GAMMA tests

sergiitk:
Waiting for an active channel takes less time in non-gamma test suites because they only start waiting after already waited for the TD backends to be created and report healthy.

In GAMMA, these resources are created asynchronously by Kubernetes. To compensate for this, we double the timeout for GAMMA tests.


sergiitk:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/b0dca063-35c4-43fd-bc08-73ebc699c6a7)
- [ ] [grpc/core/master/linux/grpc_xds_k8s_lb](https://source.cloud.google.com/results/invocations/9c96aab4-0902-4fb4-9f6d-99a7779dbb1c)

gnossen (approved):

PR #34518, opened 2023-09-28
[Interop] Make C++ xDS Interop Client and Server Responsive to SIGTERM

gnossen:
This PR adds `tini` to the C++ interop client and server to make them responsive to SIGTERM.


gnossen:
This will need to be backported to 1.59.x

eugeneo (approved):
Yay! Ctrl+C will now work 😀

tools/dockerfile/interoptest/grpc_interop_cxx_xds/Dockerfile.xds_server
```diff
@@ -27,6 +27,13 @@ COPY . .
 RUN tools/bazel build //test/cpp/interop:xds_interop_server
 RUN cp -rL /workdir/bazel-bin/test/cpp/interop/xds_interop_server /artifacts/
 
+# Download grpcurl
+RUN curl -L https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz -o grpcurl.tar.gz && \
```
sergiitk, line 31: Not too wild about having to embed grpcurl in the image. This will require making this change in all lang-specific images.
I'm not opposing this, especially short-term. But we should discuss alternatives.

sergiitk (changes requested):

tools/dockerfile/interoptest/grpc_interop_cxx_xds/Dockerfile.xds_server
```diff
@@ -38,5 +45,12 @@ RUN apt-get update \
...
+COPY --from=0 /workdir/grpcurl /bin/grpcurl
+
+# TODO: The reflection server is not currently fully functional. Copying the
+# protos is required for CLI RPC tools run against localhost of this container
+# image until this is fixed.
+COPY --from=0 /workdir/src/proto  /src/proto/
+COPY tools/dockerfile/interoptest/grpc_interop_cxx_xds/xds_server_entrypoint.sh /
 
-ENTRYPOINT ["/xds_interop_server"]
+ENTRYPOINT ["/xds_server_entrypoint.sh"]
```
sergiitk, line 56: Let's add a todo to replace it with [tini](https://github.com/krallin/tini) or similar.

tools/dockerfile/interoptest/grpc_interop_cxx_xds/Dockerfile.xds_server
gnossen: There are only two types of pre-stop hook: exec and HTTP. There's no gRPC pre-stop hook. I don't see any alternative here.

tools/dockerfile/interoptest/grpc_interop_cxx_xds/Dockerfile.xds_server
gnossen: Done.

tools/dockerfile/interoptest/grpc_interop_cxx_xds/Dockerfile.xds_server
sergiitk: Yea, I've confirmed this too. However, `exec` still provides alternatives.

sergiitk (commented):
Let's give it a test build first

tools/dockerfile/interoptest/grpc_interop_cxx_xds/Dockerfile.xds_server
```diff
@@ -39,4 +43,7 @@ RUN apt-get update \
 
 COPY --from=0 /artifacts ./
 
-ENTRYPOINT ["/xds_interop_server"]
+# tini serves as PID 1 and enables the server to properly respond to signals.
+COPY --from=0 /tini /tini
+
+ENTRYPOINT ["/tini", "--",  "/xds_interop_server"]
```
sergiitk, line 49:
```suggestion
ENTRYPOINT ["/tini", "-g", "-vv", "--",  "/xds_interop_server"]
```

sergiitk (commented):
So apparently even existing images won't die with SIGTERM normally. They still hang around until SIGKILL is issued, which is after 30s default grace period.

Let's:
1. rebase this
2. make the same change to the client
3. give it a test run

sergiitk:
For posterity, here's something interesting I discovered about container not immediately exiting after sigterm (such as cpp-server without tini):

Case where there's a blocking prestop hook and we don't release it, grace period 2 min for illustration purposes. Example timeline:
1. 0 min: k8s receives deployment delete option.
2. 2 min:
    - k8s recognizes the hook didn't complete within grace period, sends SIGTERM to the main container.
    - k8s waits for the main container to react to SIGTERM and exit for 2 more minutes.
3. 4 min: k8s recognizes the main container didn't exit after another grace period, sends the SIGKILL.

gnossen:
@sergiitk I think this does what you want now.

sergiitk:
Did you rebase / merged master?

sergiitk:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/1f02b9fd-bbf6-42c2-83ac-934dd1b46405)
- [ ] TBD

sergiitk (approved):

PR #34525, opened 2023-09-28
[AbortError] Reapply "[AbortError] And and check AbortError while abort"

XuanWang-Amos:
Reverts grpc/grpc#34515

This PR reapplies AbortError change as the previous one was reverted.

This change was mentioned in this gRFC: [L105: Python Add New Error Types](https://github.com/grpc/proposal/blob/master/L105-python-expose-new-error-types.md)


src/python/grpcio/grpc/__init__.py
```diff
@@ -1217,8 +1219,8 @@ def abort(self, code, details):
             termination of the RPC.
 
         Raises:
-          Exception: An exception is always raised to signal the abortion the
-            RPC to the gRPC runtime.
+          AbortError: An grpc.AbortError is always raised to signal the abortion
```
gnossen, line 1222: Nit: "an" -> "a". Here and throughout. Not usually important, but this will end up in the API docs.

gnossen (approved):
LGTM modulo the one nit.

src/python/grpcio/grpc/__init__.py
XuanWang-Amos: Right, better fix this.

gnossen (approved):

Issue #34530, opened 2023-09-28
[Python] Awaiting for `call.code()` dead-locks in the interceptor


alexeykudinkin:
### What version of gRPC and what language are you using?

1.57.0

### What operating system (Linux, Windows,...) and version?

Linux

### What runtime / compiler are you using (e.g. python version or version of gcc)

3.8

### What did you do?

While implementing a client-interceptor for Python AIO gRPC we've stumbled upon a dead-lock in the following code:

```
async def intercept_stream_unary(
        self,
        continuation: Callable[
            [ClientCallDetails, RequestType], StreamUnaryCall[RequestType, ResponseType]
        ],
        client_call_details: ClientCallDetails,
        request_iterator: AsyncIterable[RequestType],
    ) -> AsyncIterable[ResponseType]:
 
        def _emit_metrics(code: StatusCode):
           # ...

        try:
            unary_call = await continuation(client_call_details, request_iterator)
            # Following will dead-lock
            status_code = await unary_call.code()

        except AioRpcError as re:
            status_code = re.code()
            raise re
        finally: 
            _emit_metrics(status_code)

        return unary_call
```

I've traced it down to the following line leading to a dead-lock:
https://github.com/grpc/grpc/blob/90c72c0b777dddc8b48f24e2a50fe16ec5924205/src/python/grpcio/grpc/aio/_interceptor.py#L523

Essentially, when using `READER_WRITER` style (doing `stream.write(...)` calls), this would dead-lock preventing the `stream.write` call to be able to write into the queue and subsequently into the channel, before interceptors awaitable completes.

I looked at it from different angles and couldn't rationalize this particular await, so i'm thinking about removing it in a PR to address this.

### What did you expect to see?

This pattern of `await call` is working fine for other RPC types (unary, server-streaming), and is broken only for client-streaming.

### What did you see instead?

I see gRPC client-streaming dead-locking with an interceptor above.

XuanWang-Amos:
Hi, sorry for the late reply, after some investigation, I think this is expected behavior and the proposed fix won't help.

This is what happens for AIO READER_WRITER style stream-unary call with interceptors:
* First calls `call.write()`.
* It [awaits `self._interceptors_task`](https://github.com/grpc/grpc/blob/a0c1027bb3da36bfa5d3eb2630fd62dd79d41608/src/python/grpcio/grpc/aio/_interceptor.py#L561).
* `_interceptors_task` blocks on`await unary_call.code()`.
[...]

PR #34536, opened 2023-09-28
[Python][Callable] Fixing `_InterceptedStreamRequestMixin` to avoid deadlocking in client intereceptors

alexeykudinkin:
Issue is detailed https://github.com/grpc/grpc/issues/34530

Changes:

 - Removing awaiting on the interceptors task to allow `stream.write` invocations to channel requests into the async generator writing them out into the channel


sreenithi:
The linked issue was closed saying that it is expected behavior and this PR will not resolve it. Do we have any repro code to verify if this PR actually solves the deadlock issue?

asheshvidyut:
Closing this as stale and the related issue is also closed.

PR #34539, opened 2023-09-29
[ssl] Do not crash if ssl session cache capacity is zero.

matthewstevenson88:
This behavior is dangerous because we will crash when the cache is created, which is not necessarily on application startup and is likely when you first try to establish an SSL connection. Instead, we log an error. If the SSL library attempts to put a session ticket in the cache it will fail to do so, but everything else will continue as normal. In particular, we will always seamlessly fall back to a full SSL handshake.

Along the way, we also ensure that you cannot put a null `SSL_SESSION` into the cache, which would lead to a segfault when it is fetched from the cache.

### DIFF
--- a/src/core/tsi/ssl/session_cache/ssl_session_cache.cc
+++ b/src/core/tsi/ssl/session_cache/ssl_session_cache.cc
@@ -98,7 +98,10 @@ SslSessionLRUCache::Node* SslSessionLRUCache::FindLocked(
 }
 
 void SslSessionLRUCache::Put(const char* key, SslSessionPtr session) {
-  if (session == nullptr) return;
+  if (session == nullptr) {
+    gpr_log(GPR_ERROR, "Attempted to put null SSL session in session cache.");
+    return;
+  }
   grpc_core::MutexLock lock(&lock_);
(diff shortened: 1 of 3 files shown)


src/core/tsi/ssl/session_cache/ssl_session_cache.cc
```diff
@@ -94,6 +98,7 @@ SslSessionLRUCache::Node* SslSessionLRUCache::FindLocked(
 }
 
 void SslSessionLRUCache::Put(const char* key, SslSessionPtr session) {
+  if (session == nullptr) return;
```
gtcooke94, line 101: Should we do any error logging here - seems this would be a pretty unexpected scenario?

gtcooke94 (commented):
Minor comment - LGTM, will take another look when tests are green

src/core/tsi/ssl/session_cache/ssl_session_cache.cc
matthewstevenson88: Done.

gtcooke94 (approved):
LGTM

PR #34549, opened 2023-09-29
[Cord] Fix MakeCordFromSlice memory bug

veblush:
Fix `MakeCordFromSlice` function to fail to carry an inlined slice which can give a corrupted result when deserializing a Cord.

### DIFF
--- a/include/grpcpp/support/proto_buffer_reader.h
+++ b/include/grpcpp/support/proto_buffer_reader.h
@@ -189,6 +189,7 @@ class ProtoBufferReader : public grpc::protobuf::io::ZeroCopyInputStream {
   // This function takes ownership of slice and return a newly created Cord off
   // of it.
   static absl::Cord MakeCordFromSlice(grpc_slice slice) {
+    // slice_for_cord is created to keep inlined data of the given slice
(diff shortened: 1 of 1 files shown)


include/grpcpp/support/proto_buffer_reader.h
```diff
@@ -189,10 +189,16 @@ class ProtoBufferReader : public grpc::protobuf::io::ZeroCopyInputStream {
   // This function takes ownership of slice and return a newly created Cord off
   // of it.
   static absl::Cord MakeCordFromSlice(grpc_slice slice) {
+    grpc_slice* slice_for_cord = new grpc_slice;
```
drfloob, line 192: Please add a comment describing why a slice may need to be on the heap.

It might also be valuable to have a `GRPC_SLICE_UNDO_INLINING` macro, which could avoid this juggling.

drfloob (approved):

include/grpcpp/support/proto_buffer_reader.h
veblush: I added a comment explaining why. Thanks!

PR #34560, opened 2023-09-30
[PSM Interop] Fix the 10-minute-teardown issue in GAMMA tests

sergiitk:
We shouldn't just set `termination_grace_period_seconds=600` by default for all gamma tests extending `GammaXdsKubernetesTestCase`.

This is what's causing the deployment deletion issue:

> `framework.helpers.retryers.RetryError: Retry error calling framework.xds_k8s_testcase.IsolatedXdsKubernetesTestCase.cleanup: 1 attempts exhausted. Last exception: RetryError: Retry error calling framework.infrastructure.k8s.KubernetesNamespace.get_deployment: timeout 0:05:00 (h:mm:ss) exceeded. Check result callback returned False.`

We wait for 5 minutes, while  the deployment is happily handing for 10. Then the second cleanup retry kills it - but not before waiting for another 5 minutes.

I think `self.force = False` may be solving another issue triggered by the get_deployment retry timeout: because we start over deleting the resources by name and some of them are deleted from the first attempt we get 404. And I'm pretty sure we don't do error-handling correctly when deleting CRD-based resources - which cascades into even more unnecessary retries.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/framework/test_app/runners/k8s/gamma_server_runner.py
+++ b/tools/run_tests/xds_k8s_test_driver/framework/test_app/runners/k8s/gamma_server_runner.py
@@ -69,7 +69,7 @@ class GammaServerRunner(KubernetesServerRunner):
         safilter_name: str = "ssa-filter",
         sapolicy_name: str = "ssa-policy",
         bepolicy_name: str = "backend-policy",
-        termination_grace_period_seconds: Optional[int] = None,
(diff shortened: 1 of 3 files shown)


sergiitk:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/4e39b713-68f6-4804-8261-03ee5b0544c0)

gnossen (approved):

PR #34587, opened 2023-10-04
[build][cpp] Add cmake option to skip HTTP downloads

### LINKED ISSUE #30385: CMake: Config Step unconditionally downloading 3rd party APIs
Since Commit 99752b173cfa2f the CMake build is unconditionally downloading different 3rd party APIs during configuration if their submodule was not initialized. They are not needed for the actual build. Is this really desired?

Details:

The mentioned commit added the `download_archive` cmake-function and downloads envoy-api, googleapis, opencensus-proto and xds if the respective submodules have not been initialized.
From my understanding those packages are only needed when you want to generate/consume certain proto files for the mentioned APIs.

I'm asking because in the buildroot package we only generate the cpp libraries and those downloaded packages are never used. Further the download is (for some reason) failing in our build infrastructure which eventually breaks the package build.

Would it make sense to add a CMake Switch to disable this download?

matthewjmiller1:
Fixes #30385 for C++/cmake.

Add a `gRPC_DOWNLOAD_ARCHIVES` switch that is on by default but can be disabled for environments building without Internet connectivity.

Passes build tests:

```
$ python tools/run_tests/run_tests.py -l c++ --build_only 
PASSED: tools/run_tests/helper_scripts/build_cxx.sh [time=7116.3sec, retries=0:0SUCCESS: All tests passed

=== run_tests.py DEBUG INFO ===
command: "tools/run_tests/run_tests.py -l c++ --build_only"
===============================
$
```

And tested in a Docker container with `eth0` down to ensure we can now build when `-DgRPC_DOWNLOAD_ARCHIVES="OFF"` is passed to cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,7 @@ set(gRPC_BUILD_MSVC_MP_COUNT 0 CACHE STRING "The maximum number of processes for
 option(gRPC_BUILD_TESTS "Build tests" OFF)
 option(gRPC_BUILD_CODEGEN "Build codegen" ON)
 option(gRPC_BACKWARDS_COMPATIBILITY_MODE "Build libraries that are binary compatible across a larger number of OS and libc versions" OFF)
+option(gRPC_DOWNLOAD_ARCHIVES "Download archives for empty 3rd party directories" ON)
 
 set(gRPC_INSTALL_default ON)
 if(NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
@@ -344,7 +345,7 @@ if(_gRPC_PLATFORM_LINUX OR _gRPC_PLATFORM_POSIX)
 endif()
 
 # Setup external proto library at third_party/envoy-api with 2 download URLs
-if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api)
+if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api AND gRPC_DOWNLOAD_ARCHIVES)
   # Download the archive via HTTP, validate the checksum, and extract to third_party/envoy-api.
   download_archive(
     ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api
@@ -353,7 +354,7 @@ if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api)
     data-plane-api-9d6ffa70677c4dbf23f6ed569676206c4e2edff4
   )
 endif()
-if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api)
+if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api AND gRPC_DOWNLOAD_ARCHIVES)
   # Download the archive via HTTP, validate the checksum, and extract to third_party/envoy-api.
   download_archive(
     ${CMAKE_CURRENT_SOURCE_DIR}/third_party/envoy-api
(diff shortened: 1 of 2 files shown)


matthewjmiller1:
The Mergeable check fails b/c the PR doesn't have labels, but I don't think I have permissions to add labels to PRs for this project.

veblush:
Can you try to use `gRPC_BUILD_TESTS` instead of `gRPC_DOWNLOAD_ARCHIVES` which you added? I assume that those downloaded artifacts are only for testing so we can achieve this without adding a new flag.

And you need to modify its corresponding template file as well. (Please check out the sanity checks)

matthewjmiller1:
> Can you try to use `gRPC_BUILD_TESTS` instead of `gRPC_DOWNLOAD_ARCHIVES` which you added? I assume that those downloaded artifacts are only for testing so we can achieve this without adding a new flag.

I looked into this and as best I can tell these artifacts are completely separate from the tests and are only used in relation to gRPC's [xDS features](https://grpc.github.io/grpc/core/md_doc_grpc_xds_features.html) (e.g, in `tools/distrib/python/xds_protos/build.py`).

I was able to successfully build with `gRPC_BUILD_TESTS` on and `gRPC_DOWNLOAD_ARCHIVES` off, so it seems like building the tests is orthogonal to downloading these artifacts, that downloading the artifacts is not a dependency for building the tests.

> And you need to modify its corresponding template file as well. (Please check out the sanity checks)

I'll add a commit that updates `templates/CMakeLists.txt.template`.

I poked around the documentation some but am not sure if I also need to generate artifacts from that template for commit or if the generated artifacts are just ephemeral as part of a build. If I need to generate artifacts for commit, please let me know.

Also not sure if I can/need to run any sanity checks related to the template change locally or just wait for the CI to do the sanity checks. I'll add the commit but if there are any local sanity checks I should run, please let me know.

matthewjmiller1:
Any updates on this? The review comment was addressed nearly a month ago and this should be a relatively small change to review.

veblush:
> I was able to successfully build with gRPC_BUILD_TESTS on and gRPC_DOWNLOAD_ARCHIVES off, so it seems like building the tests is orthogonal to downloading these artifacts, that downloading the artifacts is not a dependency for building the tests.

I thought that those artifacts are required for the test then it sounds a bit strange to me. It seems that nothing requires those downloaded achieves? Did you see the cases where it fails to build with `gRPC_DOWNLOAD_ARCHIVES` off?

> I poked around the documentation some but am not sure if I also need to generate artifacts from that template for commit or if the generated artifacts are just ephemeral as part of a build. If I need to generate artifacts for commit, please let me know.

You can just run `tools/distrib/sanitize.sh` to generate those files from templates.

> Any updates on this? The review comment was addressed nearly a month ago and this should be a relatively small change to review.

Sorry about that. I happened to miss this.

matthewjmiller1:
> I thought that those artifacts are required for the test then it sounds a bit strange to me. It seems that nothing requires those downloaded achieves? Did you see the cases where it fails to build with gRPC_DOWNLOAD_ARCHIVES off?

I don't see a path for the cmake build to fail with `gRPC_DOWNLOAD_ARCHIVES` off.

The only place I see these downloaded files used is in `tools/distrib/python/xds_protos`. Maybe if one runs its `build_validate_upload.sh` script, then `build.py` won't find these directories. Why it's currently structured that cmake then downloads those instead of the Python scripts for `xds_protos`, I don't know. Maybe that could be a separate effort by someone more familiar with the `xds_protos` Python packaging to move this logic out of cmake to the scripts in there.

> You can just run tools/distrib/sanitize.sh to generate those files from templates.

This runs successfully. I don't see any generated files added that git is tracking, so I'm assuming nothing needs committed for this.

Let me know if there's anything else that needs done for this PR. It looks like all the tests now pass on the 2nd run.

matthewjmiller1:
@veblush @yashykt @eugeneo 
Are we waiting on anything else to close this PR? I think it's a relatively simple cmake change but it's been pending for a few months now. Thanks.

veblush (approved):
Thanks for the change. I'm now leading towards getting this merged.

The reason why I'm hesitant to have this one is to find a way to solve the problem without having a knob in the first place such as making the cmakefile not download at all. And any knobs of cmakefile are considered a semi-public API so it should be reviewed carefully.

But I think it's not easy to solve it without this knob. So I'm okay to have this one. In the future, we're going to revisit this issue to solve the issue.

veblush:
@yashykt One more review, please.

yashykt (approved):

PR #34607, opened 2023-10-05
[flakiness] Fix regex comparison which causes client_lb test to flake at a high rate internally when event engine is enabled.

Vignesh2208:


ctiller (approved):
image

markdroth (approved):

PR #34614, opened 2023-10-06
[Fix] new clang -Wthread-safety-analysis warnings

aeubanks:
Returning a reference to a GUARDED_BY variable while only holding the lock in the getter lets callers access the variable without holding the lock.

See https://github.com/llvm/llvm-project/pull/67776.

This is only used for testing, so just return a copy for simplicity.


yashykt:
can you remove the const please?

aeubanks:
> can you remove the const please?

done

veblush (approved):
Thanks for the PR!

PR #34628, opened 2023-10-09
[fuzzing] Avoid referencing an uninitialized fork_fd_list.

amatteis:
This avoids MSAN violations when fuzzing.


drfloob:
Thanks for your contribution. I can take a look at it after you've signed the CLA.

PR #34632, opened 2023-10-09
[ruby] create debug symbol packages

apolcyn:
Introduce `grpc-native-debug` gems containing debug symbol packages that complement the shared libraries shipped in pre-compiled binary gems.

Example package layout (`1.60.0.dev` is the "version" in this example):

```
$ find grpc-native-debug-1.60.0.dev-x86_64-linux/
grpc-native-debug-1.60.0.dev-x86_64-linux/
grpc-native-debug-1.60.0.dev-x86_64-linux/symbols
grpc-native-debug-1.60.0.dev-x86_64-linux/symbols/grpc-1.60.0.dev-x86_64-linux-ruby-3.0.dbg
grpc-native-debug-1.60.0.dev-x86_64-linux/symbols/grpc-1.60.0.dev-x86_64-linux-ruby-2.7.dbg
grpc-native-debug-1.60.0.dev-x86_64-linux/symbols/grpc-1.60.0.dev-x86_64-linux-ruby-3.2.dbg
grpc-native-debug-1.60.0.dev-x86_64-linux/symbols/grpc-1.60.0.dev-x86_64-linux-ruby-3.1.dbg
grpc-native-debug-1.60.0.dev-x86_64-linux/README.md
grpc-native-debug-1.60.0.dev-x86_64-linux/version.rb
grpc-native-debug-1.60.0.dev-x86_64-linux/platform.rb
```

<h3>Intended Usage</h3>

1) Fetch the `grpc-native-debug` package matching the `RUBY_PLATFORM` and version of the `grpc` gem being debugged. For example, if using `grpc-1.60-x86_64-linux.gem`, then fetch  `grpc-native-debug-1.60-x86_64-linux.gem`.

2) Unpack the gem and find the `symbols` directory

3) Find the `.dbg` file matching the version of ruby being used (there are separate shared libraries for each ruby version so we also need separate `.dbg` files).

4) Use the symbols however needed. For example, to load in gdb one can follow https://stackoverflow.com/questions/30281766/need-to-load-debugging-symbols-for-shared-library-in-gdb.

<h3>Caveats</h3>
- currently only works for linux x86 and x86_64 libraries

<h3>Why not just include debug symbols in the shared libraries we ship?</h3>

If we stopped stripping the shared libraries in binary gems today, we'd multiply the package size by an order of magnitude. 

For example, on linux x86_64, a stripped and compressed grpc_c.so is about 4MB. The same thing unstripped is ~80MB. There's a further multiplying effect because we include ~4 different extension libraries in each binary gem - one for each ruby minor version. Considering that the linux x86_64 gem is ~18MB today, > 300MB would be a 10-20X size increase.


alto-ruby (approved):

stanley-cheung (approved):

apolcyn:
cc @dazuma FYI

PR #34633, opened 2023-10-09
[FuzzingEventEngine] fix TickForDuration() bug and add test

markdroth:
The `TickForDuration()` method was using `grpc_core::Timestamp::Now()` to get the current time, but that was not in sync with the `now_` value inside the Fuzzing EE itself, with the result that after two subsequent 250ms increments, timers were not being properly fired.  I've added a test that demonstrates this failure without the fix.


ctiller (approved):

drfloob:
Excellent find.

PR #33646, opened 2023-07-10
[C++] Fix clang's -Winconsistent-missing-override in proto reader/writer.

Romain-Geissler-1A:
The "override" is not added on purpose to remain compatible with Protobuf < 22.x, as already written in the comment on top of these two functions.

CC @veblush as the author of this code.

Note: I am personally not super enthousiastic about this change. As an alternative, I can propose to selectively add the `override` keyword, based on the value of the `PROTOBUF_VERSION` macro (comparing it to `4022000`). Tell me if you prefer this version instead.

PR #34635, opened 2023-10-10
[C++] Use GOOGLE_PROTOBUF_VERSION instead of PROTOBUF_VERSION.

Romain-Geissler-1A:
Indeed it seems that on Protobuf side, PROTOBUF_VERSION is defined in port_def.inc which always comes with a matching include of port_undef.inc which explicitly undef all macros, among them PROTOBUF_VERSION. GOOGLE_PROTOBUF_VERSION doesn't suffer from this issue.

This is a followup of #33646.


veblush:
LGTM.

veblush (approved):

yashykt (approved):

PR #34644, opened 2023-10-10
[fuzzing] Avoid referencing an uninitialized fork_fd_list.

amatteis:
This avoids MSAN violations when fuzzing.


drfloob:
Could this be fixed with a one-line change, initializing `grpc_fd.fork_fd_list` to nullptr?

```
struct grpc_fd {
  ...
  // Only used when GRPC_ENABLE_FORK_SUPPORT=1
  grpc_fork_fd_list* fork_fd_list = nullptr;
  ...
};
```

drfloob:
Theis problem is unique to their environment, and will need to be fixed elsewhere.

PR #34657, opened 2023-10-11
[chaotic-good] Add chaotic good client transport read (roll-forward)

nanahpang:
Roll forward #34191, which is reverted due to error `2023-10-09 22:01:18,569 FAILED: cmake/build/client_transport_test --gtest_filter=ClientTransportTest.AddMultipleStreams  GRPC_POLL_STRATEGY=none` (Removed uses_event_engine=False, uses_polling=False in test build).


Vignesh2208 (approved):

ctiller:
Did we figure out the fuzzer bug yet?

nanahpang:
Yes, it is due to the frame.serialize() is the missing actual frame header value: [link](https://github.com/grpc/grpc/blob/ffe663416360d0bd7ca104f7de341a787734214c/src/core/ext/transport/chaotic_good/frame.cc#L51) (it did't cause issue before because the header is always set to nullptr, but in this PR we actually assign the header value). I am changing it to `FrameSerialize(FrameHeader header)`.

PR #34661, opened 2023-10-11
[core/security] Adding metrics header to sts request for external account credentials

aeitzman:


rockspore:
Is there any additional context as for why this header is needed?

rockspore (approved):
LGTM. @matthewstevenson88 Would you mind also taking a look?

matthewstevenson88:
After an offline discussion, redirecting this review to @markdroth.

src/core/lib/security/credentials/external/external_account_credentials.h
```diff
@@ -101,6 +101,10 @@ class ExternalAccountCredentials
       HTTPRequestContext* ctx, const Options& options,
       std::function<void(std::string, grpc_error_handle)> cb) = 0;
 
+  virtual std::string CredentialSourceType();
```
markdroth, line 104: This method should return `absl::string_view` instead of `std::string`.

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -321,7 +335,7 @@ void ExternalAccountCredentials::ExchangeToken(
   memset(&request, 0, sizeof(grpc_http_request));
   grpc_http_header* headers = nullptr;
```
markdroth, line 336: Suggest restructuring this a bit to avoid some duplication:

```
const bool add_authorization_header =
    !options_.client_id.empty() && !options_.client_secret.empty();
request.hdr_count = add_authorization_header ? 3 : 2;
auto* headers = static_cast<grpc_http_header*>(
    gpr_malloc(sizeof(grpc_http_header) * request.hdr_count));
headers[0].key = gpr_strdup("Content-Type");
headers[0].value = gpr_strdup("application/x-www-form-urlencoded");
headers[1].key = gpr_strdup("x-goog-api-client");
headers[1].value = gpr_strdup(MetricsHeaderValue().c_str());
if (add_authorization_header) {
  std::string raw_cred =
      absl::StrFormat("%s:%s", options_.client_id, options_.client_secret);
  char* encoded_cred =
      grpc_base64_encode(raw_cred.c_str(), raw_cred.length(), 0, 0);
  std::string str = absl::StrFormat("Basic %s", std::string(encoded_cred));
  headers[2].key = gpr_strdup("Authorization");
  headers[2].value = gpr_strdup(str.c_str());
}
```

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -273,6 +273,20 @@ std::string ExternalAccountCredentials::debug_string() {
                          grpc_oauth2_token_fetcher_credentials::debug_string());
 }
 
+std::string ExternalAccountCredentials::MetricsHeaderValue() {
+  return absl::StrFormat("gl-cpp/%s auth/%s google-byoid-sdk source/%s sa-impersonation/%v config-lifetime/%v",
+      "unknown",
```
markdroth, line 278: This string constant should just be inlined into the format string rather than being substituted in at runtime.

test/core/security/credentials_test.cc
```diff
@@ -2385,6 +2387,10 @@ class TestExternalAccountCredentials final : public ExternalAccountCredentials {
                                  std::vector<std::string> scopes)
       : ExternalAccountCredentials(std::move(options), std::move(scopes)) {}
 
+  std::string GetMetricsValue(){
```
markdroth, line 2390: Why is this method needed?  Why can't the caller simply call `MetricsHeaderValue()`?

test/core/security/credentials_test.cc
```diff
@@ -2393,6 +2399,84 @@ class TestExternalAccountCredentials final : public ExternalAccountCredentials {
...
+      "https://foo.com:5555/token_info",  // token_info_url;
+      credential_source,                  // credential_source;
+      "quota_project_id",                 // quota_project_id;
+      "client_id",                        // client_id;
+      "client_secret",                    // client_secret;
+      "",                                 // workforce_pool_user_project;
+  };
+  TestExternalAccountCredentials creds(options, {});
+
+  GPR_ASSERT(strcmp(creds.GetMetricsValue().c_str(),
```
markdroth, line 2423: Please write this as:

```
EXPECT_EQ(
    creds.MetricsHeaderValue(),
    absl::StrFormat(
        "gl-cpp/unknown auth/%s google-byoid-sdk source/unknown "
        "sa-impersonation/false config-lifetime/false",
        grpc_version_string()));
```

markdroth (commented):
Sorry for the delay on this.

I've triggered the tests to run.  Please address any failures that show up in addition to my comments.

test/core/security/credentials_test.cc
aeitzman: I wanted to keep MetricsHeaderValue() protected on the base external_account_credentials, wrapping it in a public method in the test credential here lets the tests call it.

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -274,19 +274,20 @@ std::string ExternalAccountCredentials::debug_string() {
 }
 
 std::string ExternalAccountCredentials::MetricsHeaderValue() {
-  return absl::StrFormat("gl-cpp/unknown"
-      " auth/%s google-byoid-sdk source/%s sa-impersonation/%v config-lifetime/%v",
-      grpc_version_string(),
-      CredentialSourceType(),
+  return absl::StrFormat(
+      "gl-cpp/unknown"
```
markdroth, line 278: I think this string can be split over 2 lines instead of 3.

src/core/lib/security/credentials/external/external_account_credentials.cc
```diff
@@ -347,10 +348,11 @@ void ExternalAccountCredentials::ExchangeToken(
         absl::StrFormat("%s:%s", options_.client_id, options_.client_secret);
     char* encoded_cred =
         grpc_base64_encode(raw_cred.c_str(), raw_cred.length(), 0, 0);
```
markdroth, line 350: Suggest changing this to use `absl::Base64Escape()`, so that you don't need to use C-style strings that need to be manually freed.

markdroth (approved):

PR #34666, opened 2023-10-11
[interop] Add absl dependency to interop clients

Vignesh2208:
Similar to https://github.com/grpc/grpc/pull/33830 but adds the absl dependency on the client side.

Requires a cherrypick

### DIFF
--- a/src/python/grpcio_tests/tests/interop/client.py
+++ b/src/python/grpcio_tests/tests/interop/client.py
@@ -216,7 +216,6 @@ def _test_case_from_arg(test_case_arg):
 
 
 def test_interoperability(args):
-    args = parse_interop_client_args()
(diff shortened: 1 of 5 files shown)


src/python/grpcio_tests/tests/interop/client.py
```diff
@@ -214,7 +215,7 @@ def _test_case_from_arg(test_case_arg):
         raise ValueError('No test case "%s"!' % test_case_arg)
 
 
-def test_interoperability():
+def test_interoperability(args):
     args = parse_interop_client_args()
```
gnossen: Doesn't this line override the `args` you've passed in?

gnossen (approved):

PR #34674, opened 2023-10-12
[build] Fix tools/distrib/iwyu.sh

yijiem:


markdroth:
This is a dup of https://github.com/grpc/grpc/pull/34673.

PR #34678, opened 2023-10-12
[xds_override_host] fix vector initialization bug

markdroth:
Instead of initializing the vector to N elements whose values are all null, we should be reserving N elements before populating the vector.

This bug didn't actually break anything, but it did cause us to allocate twice the amount of space that we should have.


eugeneo (approved):
Oops, I wrote too much Java in the past!

PR #34687, opened 2023-10-13
[PSM Interop] Log Google Cloud API debug header part 2

XuanWang-Amos:
We're seeing too many debug headers, change it to only log header in case of error.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/framework/infrastructure/gcp/compute.py
+++ b/tools/run_tests/xds_k8s_test_driver/framework/infrastructure/gcp/compute.py
@@ -29,6 +29,7 @@ logger = logging.getLogger(__name__)
 
 DEBUG_HEADER_IN_RESPONSE = "x-encrypted-debug-headers"
 DEBUG_HEADER_KEY = "X-Return-Encrypted-Headers"
+HTTP_OK_STATUS = 200
 
 
 class ComputeV1(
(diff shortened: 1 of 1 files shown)


tools/run_tests/xds_k8s_test_driver/framework/infrastructure/gcp/compute.py
```diff
@@ -585,7 +586,7 @@ def _operation_status_done(operation):
 
     @staticmethod
     def _log_debug_header(resp: httplib2.Response):
-        if DEBUG_HEADER_IN_RESPONSE in resp:
+        if DEBUG_HEADER_IN_RESPONSE in resp and resp.status != HTTP_OK_STATUS:
```
sergiitk, line 589:
```suggestion
        if DEBUG_HEADER_IN_RESPONSE in resp and resp.status >= 300:
```

There are other commonly used positive 2xx http responses, f.e. `201 Created` is a very common positive status. 
Let's just do what [googleapis library](https://github.com/googleapis/google-api-python-client/blob/6fa72cd3e8e747395f9569c6f246ca4d12459b03/googleapiclient/http.py#L1597-L1598C68) does: `resp.status >= 300`.

tools/run_tests/xds_k8s_test_driver/framework/infrastructure/gcp/compute.py
```diff
@@ -29,6 +29,7 @@
 
 DEBUG_HEADER_IN_RESPONSE = "x-encrypted-debug-headers"
 DEBUG_HEADER_KEY = "X-Return-Encrypted-Headers"
+HTTP_OK_STATUS = 200
```
sergiitk, line 32:
```suggestion
```

No need, see the other comment.

sergiitk (changes requested):

sergiitk (approved):

PR #34692, opened 2023-10-14
[CSM C++] Fix behavior when peer does not send metadata

yashykt:
@stanley-cheung noticed a bug where CSM labels were not being added on metrics if the peer was not also CSM Observability enabled.

This PR fixes the behavior to add in the local labels in this case, as well as add the remote workload type label with the value of unknown.

### DIFF
--- a/src/cpp/ext/csm/metadata_exchange.cc
+++ b/src/cpp/ext/csm/metadata_exchange.cc
@@ -205,7 +205,7 @@ class MeshLabelsIterable : public LabelsIterable {
   explicit MeshLabelsIterable(
       const std::vector<std::pair<absl::string_view, std::string>>&
           local_labels,
-      absl::optional<grpc_core::Slice> remote_metadata)
+      grpc_core::Slice remote_metadata)
       : local_labels_(local_labels), metadata_(std::move(remote_metadata)) {}
 
   absl::optional<std::pair<absl::string_view, absl::string_view>> Next()
@@ -252,8 +252,8 @@ class MeshLabelsIterable : public LabelsIterable {
 
   void ResetIteratorPosition() override { pos_ = 0; }
 
-  // Returns true if the peer sent base64 encoded "x-envoy-peer-metadata"
-  // metadata.
+  // Returns true if the peer sent a non-empty base64 encoded
+  // "x-envoy-peer-metadata" metadata.
   bool GotRemoteLabels() const {
     return GetDecodedMetadata().struct_pb != nullptr;
   }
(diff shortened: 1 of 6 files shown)


src/cpp/ext/otel/otel_plugin.h
```diff
@@ -65,9 +65,11 @@ class LabelsInjector {
  public:
   virtual ~LabelsInjector() {}
   // Read the incoming initial metadata to get the set of labels to be added to
-  // metrics.
+  // metrics. If \a received_peer_metadata is not nullptr, it is set to true if
+  // peer's metadata was received on the wire, false otherwise.
   virtual std::unique_ptr<LabelsIterable> GetLabels(
-      grpc_metadata_batch* incoming_initial_metadata) = 0;
+      grpc_metadata_batch* incoming_initial_metadata,
+      bool* received_peer_metadata) = 0;
```
markdroth, line 72: Instead of exposing this bit in the `LabelsInjector` API and handling this in the OTel call tracers, I think we should instead handle this inside of the `LabelsInjector` impl itself.  The reason I say that is that in the future, if we replace metadata exchange with a metadata lookup service, the concept of "received metadata from peer" will no longer be relevant, and we won't want to have this state machinery stick around.

yashykt: hmmm... that would mean making the LabelsInjector API hold state for the RPC

markdroth: Good point.  Is there a way to do this inside of `MeshLabelsIterable`, not in the `LabelsInjector` itself?

yashykt: I could change the API such that you have to pass back the `LabelsIterable` returned from `GetLabels` in the `AddLabels` function

markdroth: That seems fine for the server side.  But it also has to work on the client side, where `AddLabels()` is called before `GetLabels()`.

yashykt: We don't need this ability on the client anyway, so it'd be fine with a nullptr on the client side.
It's just the server that has to check whether the client sent metadata before it attaches metadata itself, whereas the client always attaches metadata.

markdroth: That sounds fine.

yashykt: Done

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -301,8 +309,8 @@ class MeshLabelsIterable : public LabelsIterable {
 
   const std::vector<std::pair<absl::string_view, std::string>>& local_labels_;
   // Holds either the metadata slice or the decoded proto struct.
-  mutable absl::variant<grpc_core::Slice, StructPb> metadata_;
-  mutable GcpResourceType type_;
+  mutable absl::variant<absl::optional<grpc_core::Slice>, StructPb> metadata_;
```
markdroth, line 312: Instead of putting `absl::optional<>` inside `absl::variant<>`, consider using `absl::monostate` as another possible type inside the `variant<>`.  I think that should give a more compact memory representation, although it may not be worth if it it makes the code more complicated.

markdroth (approved):

src/cpp/ext/csm/metadata_exchange.cc
yashykt: I can also simply say that an empty slice is not a valid value for the metadata, and that makes it more simple

markdroth: Yeah, that works too.

PR #34693, opened 2023-10-14
[chttp2] Fix outgoing data stats

yashykt:
Earlier, the grpc message-length prefix for outgoing data messages was incorrectly being counted towards `data_bytes` instead of `framing_bytes`. This PR fixes it.

Note that the incoming stats collection properly attributes the grpc message-length prefix to `framing_bytes`.

This change will affect all stats plugins (OpenCensus and OpenTelemetry) that make use of this  information for metrics.


ctiller (approved):

ctiller (changes requested):
Could we have a test that shows that whatever implementation we have is correct?

ctiller:
Probably also worth wrapping this in an experiment: it's a user visible behavior change and it's likely to have at least one load bearing alert triggered

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -121,6 +121,7 @@ TEST_F(OTelPluginEnd2EndTest, ClientAttemptSentTotalCompressedMessageSize) {
           &data[kMetricName][0].point_data);
   ASSERT_NE(point_data, nullptr);
   ASSERT_EQ(point_data->count_, 1);
+  ASSERT_EQ(absl::get<int64_t>(point_data->max_), 5);
```
ctiller, line 124: I don't think this is the right layer to add a test of chttp2 stats collection:
- it's quite disconnected from the code under test, which means when it breaks the author of that pr will have to spend substantial time connecting the dots
- it locks otel usage to chttp2, and it's likely at this point that we'll see otel + chaotic good usage next year
- it ties our testing to otel: if we decide to move to a different telemetry library in the future (this has happened before) we lose test coverage on our transports

I'd recommend instead adding a core end2end test that injects a filter or a custom call tracer, and exercise it just with http2 transports.

yashykt: I can add a http2 specific test (since it'll probably also be useful for future stats stuff), but for these metrics, the transport doesn't matter. All transports should be recording the same data size.

ctiller: We should test the chttp2 framing sizes too though.

yashykt: yeah, added. There might be something interesting going on with the framing bytes though. I want to spend a bit more time over there, so added a TODO for now.

test/core/end2end/tests/http2_stats.cc
```diff
@@ -0,0 +1,257 @@
...
+  EXPECT_EQ(client_transport_stats.outgoing.data_bytes,
+            send_from_client.size());
+  EXPECT_EQ(client_transport_stats.incoming.data_bytes,
+            send_from_server.size());
+  auto server_transport_stats = FakeServerCallTracer::transport_stream_stats();
+  EXPECT_EQ(server_transport_stats.outgoing.data_bytes,
+            send_from_server.size());
+  EXPECT_EQ(server_transport_stats.incoming.data_bytes,
+            send_from_client.size());
+  // TODO(yashykt): Add tests for framing bytes as well
```
ctiller, line 248: maybe let's just check that they're at least 23 on the client side? (9 bytes headers frame, 9 bytes data, 5 bytes grpc)?

yashykt: there is some stuff that needs to be fixed there, i'll prefer doing that in a follow-up PR

ctiller (approved):

PR #34711, opened 2023-10-17
[chaotic-good] Fix frame serialization

nanahpang:
This PR aims to fix the issue found in frame_fuzzer after #34191 was merged, where frame serialization is missing the frame header info and causes mismatch with the original frame. 

This PR needs to be merged before #34657.


ctiller (approved):

nanahpang:
Thanks for the review!

PR #34712, opened 2023-10-17
[Fix Python deadlock] Guard grpc_ssl_credentials_create with nogil

XuanWang-Amos:

### DIFF
--- a/src/python/grpcio_tests/tests/unit/_api_test.py
+++ b/src/python/grpcio_tests/tests/unit/_api_test.py
@@ -14,6 +14,7 @@
 """Test of gRPC Python's application-layer API."""
 
 import logging
+import threading
 import unittest
 
 import grpc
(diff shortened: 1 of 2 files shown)


src/python/grpcio_tests/tests/unit/_api_test.py
```diff
@@ -115,6 +116,21 @@ def test_secure_channel(self):
         channel = grpc.secure_channel("google.com:443", channel_credentials)
         channel.close()
 
+    def test_multiple_secure_channel(self):
+        def create_secure_channel():
+            channel_credentials = grpc.ssl_channel_credentials()
```
gnossen, line 121: Nit: You should probably synchronize with a [`WaitGroup`](https://github.com/grpc/grpc/blob/6d669e9c788bcdce9c7896a7d0af227c85c33372/src/python/grpcio_tests/tests/unit/test_common.py#L119) here so that the test is more likely to actually run the Core functions simultaneously.

gnossen (approved):

PR #34722, opened 2023-10-18
[ruby] upgrade and pin rake compiler dock

apolcyn:
Follow up to https://github.com/grpc/grpc/pull/34632 which upgraded rake compiler dock images to 1.3.1

It doesn't make sense to pull a rake compiler dock gem different from the version of our rake compiler dock docker images, so let's pin the version.


stanley-cheung (approved):

chadlwilson:
This one probably needs to be rebased off master (and changed to lock `= 1.8.0` given current state of master :-)

Issue #34690, opened 2023-10-13
set_response_compression unavailable


DarkMatter9309:
### What version of gRPC and what language are you using?
gRPC: 1.59.0
Python: 3.10.9

### What operating system (Linux, Windows,...) and version?
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.3 LTS
Release:        22.04
Codename:       jammy

### What runtime / compiler are you using (e.g. python version or version of gcc)
Python: 3.10.9

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.

Based on the example provided in https://github.com/grpc/grpc/tree/master/examples/python/compression
I am trying to set_response_compression() for context but get the below error:
**AttributeError: 'grpc._cython.cygrpc._ServicerContext' object has no attribute 'set_response_compression'**

### What did you expect to see?
When I run dir(grpc._cython.cygrpc._ServicerContext), I expect to see set_response_compression and when provided the gRPC will compress the response

### What did you see instead?
When I run dir(grpc._cython.cygrpc._ServicerContext), I see set_compression and when provided the gRPC does not compress the data

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

XuanWang-Amos:
Hi, are you using Aysncio in your code? If that's the case, you might need call `set_compression()` instead:
https://github.com/grpc/grpc/blob/a0c1027bb3da36bfa5d3eb2630fd62dd79d41608/src/python/grpcio/grpc/_cython/_cygrpc/aio/server.pyx.pxi#L226

XuanWang-Amos:
Well, it turns out that our example contains an error. We never actually triggered server compression. There is no method called `set_response_compression`. Instead, you should use `set_compression`. I will create a pull request to fix the example shortly.

Thank you for bringing this to our attention!

PR #34745, opened 2023-10-19
[Python Example] Fix compression example

XuanWang-Amos:
Fix: https://github.com/grpc/grpc/issues/34690

Also added generated proto files to example folder so we can run example directly.


examples/python/compression/client.py
```diff
@@ -21,9 +21,8 @@
 import logging
 
 import grpc
-
-from examples.protos import helloworld_pb2
-from examples.protos import helloworld_pb2_grpc
+import helloworld_pb2
```
gnossen, line 24: I'm wondering how this works under Bazel. PYTHONPATH should be pointing to the repo root in that case. How are we picking up `helloworld_pb2` there?

I'm all for enabling running the example under both Bazel and non-Bazel, but we have to enable that deterministically.

examples/python/compression/client.py
```diff
@@ -40,12 +39,13 @@ def run_client(channel_compression, call_compression, target):
         target, compression=channel_compression
     ) as channel:
         stub = helloworld_pb2_grpc.GreeterStub(channel)
-        response = stub.SayHello(
-            helloworld_pb2.HelloRequest(name="you"),
-            compression=call_compression,
-            wait_for_ready=True,
-        )
-        print("Response: {}".format(response))
+        for _ in range(10):
```
gnossen, line 42: Please hoist the `10` to a constant.

examples/python/compression/client.py
XuanWang-Amos: Looks like in the Bazel test we're using `subprocess` to start client and server, so technically those scripts was started  at parent directory?
https://github.com/grpc/grpc/blob/352c33986e5069abbdc8496915d8f913230c15df/examples/python/compression/test/compression_example_test.py#L22-L24

I think that's why we have `examples/python/compression` added to PYTHONPATH.

gnossen: Resolved offline. This actually seems to be behavior provided by Bazel to emulate a script being run from its own directory. Surprising, but harmless.

gnossen (approved):

PR #34752, opened 2023-10-19
[PSM Interop] Add PodMonitoring resource to psm xds interop testing framework

stanley-cheung:
Add a `PodMonitoring` resource type to the PSM interop testing framework. This is needed so that GMP (Google Managed Prometheus) can scrape the matching GKE pods Prometheus endpoint for Prometheus metrics.

### DIFF
--- a/tools/run_tests/xds_k8s_test_driver/kubernetes-manifests/csm/pod-monitoring.yaml
+++ b/tools/run_tests/xds_k8s_test_driver/kubernetes-manifests/csm/pod-monitoring.yaml
@@ -1,0 +1,13 @@
+---
+apiVersion: monitoring.googleapis.com/v1
+kind: PodMonitoring
+metadata:
+  namespace: ${namespace_name}
+  name: ${deployment_id}-gmp
+spec:
+  selector:
+    matchLabels:
+      deployment_id: ${deployment_id}
+  endpoints:
+  - port: 9464
+    interval: 10s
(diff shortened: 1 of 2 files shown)


tools/run_tests/xds_k8s_test_driver/kubernetes-manifests/csm/pod-monitoring.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: PodMonitoring
+metadata:
+  namespace: ${namespace_name}
+  name: ${deployment_id}-gmp
+spec:
+  selector:
+    matchLabels:
+      deployment_id: ${deployment_id}
+  endpoints:
+  - port: 9464
```
sergiitk, line 12: Is this supposed to be static?

sergiitk (approved):
LGTM

tools/run_tests/xds_k8s_test_driver/kubernetes-manifests/csm/pod-monitoring.yaml
stanley-cheung: @sergiitk Should I make it something that got generated each time from the framework and pass it as a variable?

sergiitk: I can't answer this without seeing how it's used. I'm OK with merging this PR as is - I don't see any invocations that use this template at the moment anyway. Then we can replace it with a variable once you actually render and apply the template.

stanley-cheung: 1. The <language> client / server containers will expose things in that port in this other PR https://github.com/grpc/grpc/pull/34832. The port number `9464` is also hard-coded there, but yes in a next set of PR we can make it variable
2. The framework changes to render the template is here: https://github.com/grpc/grpc/pull/34835

PR #34758, opened 2023-10-20
[tls] Fix ownership bugs in TlsCredentialsOptions and grpc_tls_credentials_options.

matthewstevenson88:
Currently it is very easy to use the `TlsCredentialsOptions` in such a way that it produces a memory leak. For example, the code block
```
{
  TlsCredentialsOptions options;
}
```
produces a memory leak. This PR fixes up the ownership bugs in this class and its `grpc_tls_credentials_options`, the C-core analogue.


include/grpcpp/security/tls_credentials_options.h
```diff
@@ -108,10 +109,14 @@ class TlsCredentialsOptions {
   void set_crl_provider(std::shared_ptr<CrlProvider> crl_provider);
 
   // ----- Getters for member fields ----
-  // Get the internal c options. This function shall be used only internally.
-  grpc_tls_credentials_options* c_credentials_options() const {
+  // Returns the internal c options. The caller does not take ownership of the
+  // returned pointer. This function shall be used only internally.
+  grpc_tls_credentials_options* mutable_c_credentials_options() {
```
gtcooke94, line 114: If we only want it used internally can we take it out of this header?

include/grpcpp/security/tls_credentials_options.h
```diff
@@ -108,10 +109,14 @@ class TlsCredentialsOptions {
...
     return c_credentials_options_;
   }
+  // Returns a deep copy of the internal c options. The caller takes ownership
```
gtcooke94, line 117: Same about internal only use in this header

test/cpp/client/credentials_test.cc
```diff
@@ -419,8 +423,89 @@ TEST(CredentialsTest, TlsChannelCredentialsWithCrlProviderAndDirectory) {
   GPR_ASSERT(channel_credentials.get() != nullptr);
 }
 
-// TODO(gtcooke94) - Add test to make sure Tls*CredentialsOptions does not leak
-// when not moved into TlsCredentials
+TEST(CredentialsTest, TlsCredentialsOptionsDoesNotLeak) {
+  // TlsCredentialsOptions does not leak.
```
gtcooke94, line 427: What do you think about separating each of these into individual test cases?

I prefer it - imagine this starts failing sometime, it's immediately somewhat annoying to start to debug which of the 5 tests in this one test is failing.

gtcooke94 (commented):
Looks good overall, have a few questions

test/cpp/client/credentials_test.cc
matthewstevenson88: Done.

gtcooke94 (approved):
LGTM

PR #34768, opened 2023-10-20
[ALTS] Allow key buffer borrowing in alts_zero_copy_grpc_protector

rockspore:
Change the `alts_zero_copy_grpc_protector_create()`, and its related functions, to optionally borrow the key buffer instead of copying it


src/core/tsi/alts/zero_copy_frame_protector/alts_zero_copy_grpc_protector.h
```diff
@@ -23,33 +23,37 @@
...
-/// This method returns TSI_OK on success or a specific error code otherwise.
-///
+//
+// This method creates an ALTS zero-copy grpc protector.
+//
+//- key: a symmetric key used to seal/unseal frames. The given buffer may be
+//  mutated if copy_key is set to false.
+//- key_size: the size of symmetric key.
+//- copy_key: when true, the given key will not be mutated and a copy of it
+//  will be made for the underlying encrypter. If false, the caller must ensure
```
matthewstevenson88, line 35: micro-nit: s/encrypter/crypter (for consistency with the rest of the language)

src/core/tsi/alts/crypt/aes_gcm.cc
```diff
@@ -687,21 +691,32 @@ grpc_status_code gsec_aes_gcm_aead_crypter_create(const uint8_t* key,
...
         gpr_malloc(sizeof(gsec_aes_gcm_aead_rekey_data)));
-    memcpy(aes_gcm_crypter->rekey_data->nonce_mask, key + kKdfKeyLen,
-           kAesGcmNonceLength);
+    if (copy_key) {
+      aes_gcm_crypter->rekey_data->nonce_mask =
+          static_cast<uint8_t*>(gpr_malloc(kAesGcmNonceLength));
+      memcpy(aes_gcm_crypter->rekey_data->nonce_mask, key + kKdfKeyLen,
+             kAesGcmNonceLength);
+    } else {
+      aes_gcm_crypter->rekey_data->nonce_mask = key + kKdfKeyLen;
```
matthewstevenson88, line 705: What's the plan here for when we do the rekey? Are we going to do that in a follow-up? Or do we seamlessly mutate the existing key even when we don't own it?

test/core/tsi/alts/crypt/aes_gcm_test.cc
```diff
@@ -803,8 +804,9 @@ static void gsec_test_create_random_aes_gcm_crypter(gsec_aead_crypter** crypter,
                                                     bool rekey) {
   uint8_t* key;
   gsec_test_random_array(&key, key_length);
-  gsec_aes_gcm_aead_crypter_create(key, key_length, nonce_length, tag_length,
-                                   rekey, crypter, nullptr);
+  gsec_aes_gcm_aead_crypter_create(key, key_length, /*copy_key=*/true,
```
matthewstevenson88, line 807: Could we run all of these tests twice, once with copy_key=true and once with copy_key=false?

matthewstevenson88 (commented):
Generally LGTM, just a few comments.

src/core/tsi/alts/crypt/aes_gcm.cc
rockspore: IIUC of https://github.com/grpc/grpc/blob/948fd6bc9c9c0837c736c32fd229a4d9e5033f26/src/core/tsi/alts/crypt/aes_gcm.cc#L256, the derived key is stored separately in its own stack. If we want to protect that, we would have to derive this key on top of the non-owning buffer. However, since this derivation needs the original key, it seems to me we have only two choices:

1. Introduce SecretData (or its equivalent) here to protect this local buffer.
2. Have the caller provide two key buffers, one for mutation, one for read.

Whichever we do, including nothing, I'd prefer to do it after this PR. Since this is the one we need to unblock internal change.

Let me know if I'm missing something obvious.

test/core/tsi/alts/crypt/aes_gcm_test.cc
rockspore: Done.

src/core/tsi/alts/zero_copy_frame_protector/alts_zero_copy_grpc_protector.h
rockspore: Done.

src/core/tsi/alts/crypt/aes_gcm.cc
rockspore: Sounds good! Done.

src/core/tsi/alts/crypt/aes_gcm.cc
```diff
@@ -655,12 +665,10 @@ static grpc_status_code aes_gcm_new_evp_cipher_ctx(
   return GRPC_STATUS_OK;
 }
 
-grpc_status_code gsec_aes_gcm_aead_crypter_create(const uint8_t* key,
-                                                  size_t key_length,
-                                                  size_t nonce_length,
-                                                  size_t tag_length, bool rekey,
-                                                  gsec_aead_crypter** crypter,
-                                                  char** error_details) {
+grpc_status_code gsec_aes_gcm_aead_crypter_create(
```
matthewstevenson88, line 668: I think we missed a copy of the key in the `aes_gcm_new_evp_cipher_ctx` function. Is there an easy way to deal with that buffer? I'm also ok doing this in a follow-up if you'd like.

test/core/tsi/alts/crypt/aes_gcm_test.cc
matthewstevenson88: It looks like the tests in this file are still only run with copy_key=true?

rockspore: Good point! I duplicated those tests taking copy_key and run them with either true or false in different test cases.

src/core/tsi/alts/crypt/aes_gcm.cc
rockspore: This is again a local buffer that is similar to the rekey we discussed earlier. I added the same TODO here.

matthewstevenson88 (commented):
Discussed offline, we're going to see if we can abstract out the key, and if not then we will come back to this.

rockspore:
Decided to go with a different way as in #34814

PR #34769, opened 2023-10-20
[e2e tests] refactor code for determining local IP address

markdroth:


test/core/util/resolve_localhost_ip46.cc
```diff
@@ -59,4 +59,20 @@ void LocalhostResolves(bool* ipv4, bool* ipv6) {
...
 }
 
+bool RunningWithIPv6Only() {
+  bool localhost_resolves_to_ipv4 = false;
+  bool localhost_resolves_to_ipv6 = false;
+  LocalhostResolves(&localhost_resolves_to_ipv4, &localhost_resolves_to_ipv6);
+  return !localhost_resolves_to_ipv4 && localhost_resolves_to_ipv6;
+}
+
+absl::string_view LocalIp() {
```
eugeneo, line 69: Almost all calls to this method are as a part of string cancatenation:
```
absl::StrCat(grpc_core::LocalIp(), ":", backends_[0]->port())
```
I would suggest considering having that in the utility function.

test/core/util/resolve_localhost_ip46.cc
```diff
@@ -59,4 +59,20 @@ void LocalhostResolves(bool* ipv4, bool* ipv6) {
...
+  return RunningWithIPv6Only() ? "[::1]" : "127.0.0.1";
+}
+
+std::string LocalIpUri(int port) {
```
eugeneo, line 73: Most calls to this function look like this:
```
grpc_core::URI::Parse(grpc_core::LocalIpUri(port))
```
I would suggest having a function for that to reduce the code duplication.

eugeneo (approved):
Approved, but I suggest slightly higher-level API.

test/core/util/resolve_localhost_ip46.cc
markdroth: Only 5 of 38 calls use that pattern.  I don't think that's worth further refactoring.

eugeneo (approved):

PR #34790, opened 2023-10-25
[xds] C++ XdsServerBuilder Remove src/... header dependency from public header

yashykt:


markdroth (approved):
Thanks for cleaning this up!

For bonus points, it might be a good idea to add a presubmit check to ensure that nothing in the include tree includes anything in the src tree.

yashykt:
> For bonus points, it might be a good idea to add a presubmit check to ensure that nothing in the include tree includes anything in the src tree.

will add one

PR #34794, opened 2023-10-25
[chaotic-good] Fix frame fuzzer error.

nanahpang:
The frame fuzzer constructs random bytes and frame header to test serialization and deserialization. There are error cases where the frame header isn't constructed correctly and isn't align with the serialized bytes. Added checks in frame header parser and initialize frame header's flags and header/trailer length in serialization.


src/core/ext/transport/chaotic_good/frame_header.cc
```diff
@@ -60,9 +60,15 @@ absl::StatusOr<FrameHeader> FrameHeader::Parse(const uint8_t* data) {
   header.flags = BitSet<2>::FromInt(flags);
   header.stream_id = ReadLittleEndianUint32(data + 4);
   header.header_length = ReadLittleEndianUint32(data + 8);
+  if (header.flags.is_set(0) && !(header.header_length > 0)) {
```
ctiller, line 63: how about:
`if (header.flags.is_set(0) && header.header_length <= 0)`

src/core/ext/transport/chaotic_good/frame_header.cc
```diff
@@ -60,9 +60,15 @@ absl::StatusOr<FrameHeader> FrameHeader::Parse(const uint8_t* data) {
...
+    return absl::InvalidArgumentError("Invalid header length");
+  }
   header.message_length = ReadLittleEndianUint32(data + 12);
   header.message_padding = ReadLittleEndianUint32(data + 16);
   header.trailer_length = ReadLittleEndianUint32(data + 20);
+  if (header.flags.is_set(1) && !(header.trailer_length > 0)) {
```
ctiller, line 69: here too

src/core/ext/transport/chaotic_good/frame_header.cc
nanahpang: Updated, thanks for the suggestion.

src/core/ext/transport/chaotic_good/frame_header.cc
nanahpang: Done.

ctiller (approved):

Issue #22285, opened 2020-03-10
Windows AF_UNIX/create_from_fd support


bridgewaterrobbie:
### Is your feature request related to a problem? Please describe.
Windows now supports AF_UNIX socket functionality. The create_from_fd feature, as far as I can tell browsing past issues, was partly done to support using AF_UNIX sockets with GRPC. However, currently there is no way to make use of this feature in Windows as it as key parts of it are only implemented in the grpc POSIX specific files

### Describe the solution you'd like
Defining GPR_SUPPORT_CHANNELS_FROM_FD should compile on windows and support these features

### Describe alternatives you've considered
I will now be looking into making my own custom channel to see if that can work for my use case, though there seems to be limited information on how exactly to do that.

### Additional context
Add any other context about the feature request here.

bridgewaterrobbie:
I would still find this a valuable feature if time can be found for it. Sorry if I shouldn't "Bump" stale tickets though

guantaol:
I am not sure who should take this at the moment but assign it to Yash first. Please feel free to reassign it to someone who is more appropriate.

drfloob:
It looks like we've had years of a non-answer to this, apologies. Here's my best summary of the situation. This has been requested a handful of times, and it's challenging to implement. There are a number of reasons, and I am most familiar with the challenge that AF_UNIX would need to be treated differently with respect to security concerns on Windows vs Posix. Unix sockets are currently considered "trusted" within gRPC, and that might need to change (some research would need to be done here by the right people, anyway). It's unlikely the team will have resources to spend on this any time soon.

PR #34801, opened 2023-10-26
[EventEngine] Support AF_UNIX for windows

youyuanwu:
#22285 
Unix domain socket has been added to windows: [AF_UNIX comes to Windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/)

Golang net pkg has adopted this long ago: 
https://go-review.googlesource.com/c/go/+/125456
https://go-review.googlesource.com/c/sys/+/132555
grpc-go already support this.

AF_UNIX on windows is seamlessly integrated with winsock API. 
The modification needed are:
* Set the right address family AF_UNIX depending on the address config, instead of using AF_INET6 all the time.
* Ignore socket options for tcp.


test/core/end2end/end2end_test_suites.cc
```diff
@@ -550,7 +550,29 @@ class InsecureFixtureWithPipeForWakeupFd : public InsecureFixture {
...
+std::string GetTempDir(){
+#ifdef WIN32
+  // windows temp dir usually exceeds uds length.
+  // std::array<char, MAX_PATH + 1> buff = {};
+  // DWORD len = GetTempPathA(buff.size(), buff.data());
+  // if (len == 0) {
+  //   GPR_ASSERT(false);
+  // }
+  // return std::string(buff.data(), len);
+  std::string temp_dir = "C:/tmp/";
```
youyuanwu, line 562: May need to find a better way to use the temp dir on windows.
Currently the test makes a dir from the root of the c drive.

The temp dir should be "C:\Users\user1\AppData\Local\Temp\grpc_fullstack_test.xxx.xxx" which exceeds uds name length.

src/core/lib/iomgr/tcp_client_windows.cc
```diff
@@ -145,6 +145,8 @@ static int64_t tcp_connect(grpc_closure* on_done, grpc_endpoint** endpoint,
   grpc_error_handle error;
   async_connect* ac = NULL;
   absl::StatusOr<std::string> addr_uri;
+  int af;
```
youyuanwu, line 148: Can someone tell me how to enable UseEventEngineClient so that I can test the event engine client code path?

youyuanwu:
@drfloob Can you review?

drfloob:
Hi @youyuanwu. This feature has been in discussion for a while. In short, we need to consider whether unix sockets can be treated identically within gRPC across platforms, or if there are subtle semantic differences that require a bit more scrutiny. There are a few folks reviewing the concept now, and we'll get back with you soon.

Thank you for the contribution regardless!

youyuanwu:
The PR has been tested manually with grpc-go, which already supports AF_UNIX on windows.
run_tests.py with c++ tests passes locally.

lostobject:
I have also done some testing of this with my own C++ server and C# client (grpc-dotnet which already supports AF_UNIX). No issues found.

drfloob:
Sorry for the delay, @youyuanwu. We've gotten consensus that this seems pretty safe to implement, so we can continue with the code review. Unfortunately, I'm going to be out for the remainder of the year, so we'll have to pick it up in January, but I'll do a quick first pass today.

test/core/end2end/end2end_test_suites.cc
drfloob: This may work for our test suite. If so, it may be sufficient to add a TODO to improve later if we need to. I'll kick off the tests shortly.

test/core/end2end/end2end_test_suites.cc
```diff
@@ -550,7 +550,29 @@ class InsecureFixtureWithPipeForWakeupFd : public InsecureFixture {
...
+  if (CreateDirectoryA(temp_dir.c_str(), NULL) == 0 &&
+    ERROR_ALREADY_EXISTS != GetLastError())
+  {
+    GPR_ASSERT(false); // temp dir create failed
```
drfloob, line 566: Prefer `grpc_core::Crash(absl::StrCat("Could not create temp dir: ", temp_dir));`

test/core/end2end/end2end_test_suites.cc
```diff
@@ -550,7 +550,29 @@ class InsecureFixtureWithPipeForWakeupFd : public InsecureFixture {
...
+#ifdef WIN32
+  // windows temp dir usually exceeds uds length.
+  // std::array<char, MAX_PATH + 1> buff = {};
+  // DWORD len = GetTempPathA(buff.size(), buff.data());
+  // if (len == 0) {
+  //   GPR_ASSERT(false);
+  // }
+  // return std::string(buff.data(), len);
+  std::string temp_dir = "C:/tmp/";
+  if (CreateDirectoryA(temp_dir.c_str(), NULL) == 0 &&
```
drfloob, line 563: If this implementation ends up working, we can create this directory once and return a static variable. That would let us eliminate the lambda captures below.

src/core/lib/iomgr/tcp_client_windows.cc
youyuanwu: iomgr and EventEngine both work now.

src/core/ext/filters/client_channel/resolver/binder/binder_resolver.cc
```diff
@@ -24,8 +24,13 @@
 #ifdef GRPC_HAVE_UNIX_SOCKET
 
 #include <string.h>
+#ifdef WIN32
```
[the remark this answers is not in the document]
youyuanwu: This works for 64-bit as well. Changed to use GPR_WINDOWS.

WIN32 indicates WIN32 API set is available and does not distinguish 32 vs 64 bits. To check 32 vs 64 bits, some other macro needs to be used but I do not remember which one.

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -292,7 +300,22 @@ EventEngine::ConnectionHandle WindowsEventEngine::Connect(
     return EventEngine::ConnectionHandle::kInvalid;
   }
   // bind the local address
-  auto local_address = ResolvedAddressMakeWild6(0);
+  ResolvedAddress local_address;
+  if(addr_family == AF_UNIX)
+  {
+    // Note: For ConnectEx() to work for AF_UNIX, the sock needs to be bind to
```
drfloob, line 306: I don't quite follow the logic in the second paragraph, why describe the alternative? Maybe the first sentence is all that's required here. I've corrected some grammar errors, please take a look.

```
// Note: For ConnectEx() to work with AF_UNIX, the socket needs to be bound to
// the local address of a unnamed socket.
```

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -67,11 +67,40 @@ void WindowsEventEngineListener::SinglePortSocketListener::
...
+    return;
+  }
+  struct sockaddr_un* un = reinterpret_cast<struct sockaddr_un*>(
+      const_cast<sockaddr*>(resolved_addr.address()));
+  // There is nothing to unlink for an abstract unix socket
+  if (un->sun_path[0] == '\0' && un->sun_path[1] != '\0') {
+    return;
+  }
+  // for windows we need to remove the file instead of unlink
+  // stat api not working, so use win api
```
drfloob, line 84: This comment is unnecessary, we are using the windows APIs everywhere.

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -67,11 +67,40 @@ void WindowsEventEngineListener::SinglePortSocketListener::
...
+  if (resolved_addr.address()->sa_family != AF_UNIX) {
+    return;
+  }
+  struct sockaddr_un* un = reinterpret_cast<struct sockaddr_un*>(
+      const_cast<sockaddr*>(resolved_addr.address()));
+  // There is nothing to unlink for an abstract unix socket
+  if (un->sun_path[0] == '\0' && un->sun_path[1] != '\0') {
+    return;
+  }
+  // for windows we need to remove the file instead of unlink
```
drfloob, line 83: grammar nit: please capitalize `For` and add a period at the end of the sentence.

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -109,7 +138,11 @@ absl::Status WindowsEventEngineListener::SinglePortSocketListener::Start() {
 
 absl::Status
 WindowsEventEngineListener::SinglePortSocketListener::StartLocked() {
-  SOCKET accept_socket = WSASocket(AF_INET6, SOCK_STREAM, IPPROTO_TCP, NULL, 0,
+  const EventEngine::ResolvedAddress addr = listener_sockname();
+  const int addr_family = addr.address()->sa_family;
+  const int protocol = addr_family == AF_UNIX ? 0 : IPPROTO_TCP;
+  SOCKET accept_socket = WSASocket(addr_family, SOCK_STREAM, protocol,
```
drfloob, line 144: similar to the above comment, this should be `AF_INET6` for tcp.

src/core/lib/event_engine/windows/windows_listener.cc
```diff
@@ -313,8 +357,11 @@ absl::StatusOr<int> WindowsEventEngineListener::Bind(
     out_addr = ResolvedAddressMakeWild6(out_port);
   }
   // open the socket
-  SOCKET sock = WSASocket(AF_INET6, SOCK_STREAM, IPPROTO_TCP, nullptr, 0,
-                          IOCP::GetDefaultSocketFlags());
+  const int addr_family = out_addr.address()->sa_family;
```
drfloob, line 360: similar to the above comment, this should be `AF_INET6` for tcp.

include/grpc/support/port_platform.h
```diff
@@ -165,6 +165,7 @@
 #define GPR_WINDOWS_STAT
 #define GPR_WINDOWS_STRING
 #define GPR_WINDOWS_TIME
+#define GRPC_HAVE_UNIX_SOCKET 1
```
drfloob, line 168: The network-related defines are mostly in `src/core/lib/iomgr/port.h`, and you have defined it in both places. I suggest only defining this macro in `port.h`, and ensuring that `port.h` is included in every file that uses it.

src/core/lib/iomgr/tcp_client_windows.cc
```diff
@@ -159,14 +161,24 @@ static int64_t tcp_connect(grpc_closure* on_done, grpc_endpoint** endpoint,
     addr = &addr6_v4mapped;
   }
 
-  sock = WSASocket(AF_INET6, SOCK_STREAM, IPPROTO_TCP, NULL, 0,
+  // extract family
+  af = grpc_sockaddr_get_family(addr);
```
drfloob, line 165: similar to the above comment, this should be `AF_INET6` for tcp.

src/core/lib/iomgr/tcp_client_windows.cc
```diff
@@ -183,7 +195,20 @@ static int64_t tcp_connect(grpc_closure* on_done, grpc_endpoint** endpoint,
     goto failure;
   }
 
-  grpc_sockaddr_make_wildcard6(0, &local_address);
+  if (af == AF_UNIX) {
+    // Note: For ConnectEx() to work for AF_UNIX, the sock needs to be bind to
```
drfloob, line 199: Please adjust the comment as discussed above in the EventEngine version.

src/core/lib/iomgr/tcp_client_windows.cc
```diff
@@ -194,9 +219,9 @@ static int64_t tcp_connect(grpc_closure* on_done, grpc_endpoint** endpoint,
 
   socket = grpc_winsocket_create(sock, "client");
   info = &socket->write_info;
+
```
drfloob, line 222: nit: for consistency with the codebase, please remove blank lines from inside function bodies.

drfloob (commented):
Sorry again for the delay. This is getting closer! I've done another review, please let me know if you have any questions.

test/core/end2end/end2end_test_suites.cc
youyuanwu: The solution works. Changed to a const global variable. The directory is created only once now.
In future we might want to use a local directory for this test.

include/grpc/support/port_platform.h
youyuanwu: Removed. port.h is used when needed.

youyuanwu:
@drfloob The last test run had failures when event engine listener is turned off. I cannot see the test run links anymore and do not remember the config to turn on/off, so please rerun the pipeline again.
Let me know if that use case needs to be supported as well.

dpservis:
I am updating this page 100 times a day, looking for the moment when the PR will be merged in production. This is a great achievement and I am looking forward to using it.

drfloob:
@youyuanwu 

You can see the previous failures by hovering your mouse over the red "x" on the right, next to every individual commit
image

> I do not remember the config to turn on/off, so please rerun the pipeline again.

I kicked off CI again, but you can disable the `event_engine_listener` experiment with the env var `GRPC_EXPERIMENTS=-event_engine_listener`, or if you're running tests through bazel: `bazel test --test_env=GRPC_EXPERIMENTS=-event_engine_listener`.

You'll likely want to test with:
* `GRPC_EXPERIMENTS=event_engine_client,event_engine_listener`
* `GRPC_EXPERIMENTS=-event_engine_client,-event_engine_listener`

youyuanwu:
@drfloob All failed windows tests has the config: experiment=no_event_engine_listener
Does this feature need to support no_event_engine_listener? I only changed event engine listener. I assume that there is another iomgr path for listener that I never changed. If this is required I will give a shot later this week.

drfloob:
> Does this feature need to support no_event_engine_listener? I only changed event engine listener. I assume that there is another iomgr path for listener that I never changed. If this is required I will give a shot later this week.

Support for iomgr is necessary (that experiment enables iomgr listeners, effectively). There is indeed another file that I missed. [`src/core/lib/iomgr/tcp_server_windows.cc`](https://github.com/grpc/grpc/blob/6cd53e1c80bbf5d4a5ca16cd3e1231304c792f05/src/core/lib/iomgr/tcp_server_windows.cc). I'm guessing we're looking at a too-small sockaddr.

youyuanwu:
Status: experiment=-event_engine_listener is supported now.

One more finding:
This change might have exposed a separate potential deadlock bug in windows event engine client.
With `$env:GRPC_EXPERIMENTS="event_engine_client` enabled. The client of uds will hit assert in debug mode about potential deadlock if server is not listening. This is related to grpc_core::ExecCtx dispatching pending tasks that uses locks. Will create a spearate issue about this, since I do not think the current change is the culprit.

@drfloob Can you rerun the ci?

drfloob:
I see the mutex re-acquisition problem https://source.cloud.google.com/results/invocations/419395f4-3a67-4bd1-9c73-f3fe2a43c1e5/targets/%2F%2Ftest%2Fcore%2Fend2end:connectivity_test@experiment%3Dpromise_based_client_call/log

youyuanwu:
> I see the mutex re-acquisition problem https://source.cloud.google.com/results/invocations/419395f4-3a67-4bd1-9c73-f3fe2a43c1e5/targets/%2F%2Ftest%2Fcore%2Fend2end:connectivity_test@experiment%3Dpromise_based_client_call/log

Yes. This is the symptom. The easiest repro is to create any uds client that try connect to a non-existent server. Somewhere inside "event_engine_tcp_client_connect()" has the problem.

drfloob:
@youyuanwu I have a fix in the linked PR. Thanks for helping us find some bugs!

youyuanwu:
@drfloob I rebased with master with your change. The locking issue has been fixed by your change.

Vignesh2208 (approved):

drfloob:
@youyuanwu Your change is merged, thanks for adding this feature!

youyuanwu:
@drfloob Thanks for the review and help. I will leave the documentation change to you, to say that uds is no longer unix only.

dpservis:
Hey, did this make it into main?

drfloob:
> Hey, did this make it into main?

Yes. https://github.com/grpc/grpc/commit/a58f83c947be52d4b622aa0d32435323b81a8e5e

PR #34829, opened 2023-10-30
[observability] Switch from datetime.utcnow to time.perf_counter

mnito:
datetime.utcnow() raises a deprecation warning on Python 3.12, so this change prevents that deprecation warning while also switching to a monotonic clock for calculating RPC latency

cc @XuanWang-Amos

### DIFF
--- a/src/python/grpcio/grpc/_channel.py
+++ b/src/python/grpcio/grpc/_channel.py
@@ -14,7 +14,6 @@
 """Invocation-side implementation of gRPC Python."""
 
 import copy
-from datetime import datetime
 import functools
 import logging
 import os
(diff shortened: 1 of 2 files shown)


XuanWang-Amos:
Thanks for the PR! Just out of curiosity, are you currently using the Python observability?

mnito:
> Thanks for the PR! Just out of curiosity, are you currently using the Python observability?

No, we are not currently using it -- I actually did not know about the built-in observability features until I saw the deprecation warnings. Over this past year we have been making improvements to our APM monitoring so this is definitely something good to know about.

XuanWang-Amos:
> > Thanks for the PR! Just out of curiosity, are you currently using the Python observability?
> 
> No, we are not currently using it -- I actually did not know about the built-in observability features until I saw the deprecation warnings. Over this past year we have been making improvements to our APM monitoring so this is definitely something good to know about.

Got it, observability is still a work in progress, but I'm glad to hear that we have users who are interested in it.

XuanWang-Amos (approved):

PR #34839, opened 2023-11-01
[build] Speed up clang_format

drfloob:
This reduces clang-format duration to under 10s on my 128 core machine. Previously I stopped the process at 10 minutes. Major changes:

Eliminates 18,000 files
```
find ... \
    -and -not -path '*/cmake/build/*' \
    -and -not -path '*/huffman_geometries/*' \
```

Also, this now runs `$CPU_COUNT` number of parallel clang-format processes with a roughly equal number of files. Previously, this script was formatting one file at a time in separate processes (`xargs -n 1`).

### DIFF
--- a/tools/dockerfile/grpc_clang_format/clang_format_all_the_things.sh
+++ b/tools/dockerfile/grpc_clang_format/clang_format_all_the_things.sh
@@ -50,7 +50,6 @@ do
     -and -not -name grpc_tls_credentials_options.h \
     -and -not -name grpc_tls_credentials_options_comparator_test.cc \
     -and -not -path '*/cmake/build/*' \
-    -and -not -path '*/huffman_geometries/*' \
     `"
   done
 done
--- a/tools/dockerfile/grpc_clang_format/clang_format_all_the_things.sh
+++ b/tools/dockerfile/grpc_clang_format/clang_format_all_the_things.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -e
+set -ex
 
 # directories to run against
 DIRS="examples/cpp examples/android/binder src/core/lib src/core/tsi src/core/ext src/cpp test/core test/cpp include src/compiler src/ruby src/objective-c tools/distrib/python src/python/grpcio_observability"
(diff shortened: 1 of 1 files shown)


ctiller:
can we keep huffman_geometries? (or at least have a recipe to fix them next time I change that code generator)

tools/dockerfile/grpc_clang_format/clang_format_all_the_things.sh
```diff
@@ -32,7 +32,26 @@ for dir in $DIRS
...
+    -and -not -name '*.pb.cc' \
+    -and -not -name '*.pbobjc.h' \
+    -and -not -name '*.pbobjc.m' \
+    -and -not -name '*.pbrpc.h' \
+    -and -not -name '*.pbrpc.m' \
+    -and -not -name end2end_tests.cc \
+    -and -not -name grpc_shadow_boringssl.h \
+    -and -not -name grpc_tls_credentials_options.h \
+    -and -not -name grpc_tls_credentials_options_comparator_test.cc \
+    -and -not -path '*/cmake/build/*' \
```
veblush, line 52: I'm curious why cmake files happen to reach here. Do we have cmake files under any directory from `DIRS`?

drfloob:
> can we keep huffman_geometries? (or at least have a recipe to fix them next time I change that code generator)

@ctiller I added them back in. clang-format struggles with them, those 233 files double the run time on my system.

tools/dockerfile/grpc_clang_format/clang_format_all_the_things.sh
drfloob: It depends on the state of your local environment, and I suppose the state of the CI build environment. I didn't go out of my way to run anything cmake-related, but I get 5901 entries for `find . -type d -path '*/cmake/build/*' | wc -l`

drfloob:
```
$ find examples/cpp -type d -path '*/cmake/build/*' | wc -l
1154
```

veblush: oh i see. examples! sg

veblush (approved):

tools/dockerfile/grpc_clang_format/clang_format_all_the_things.sh
```diff
@@ -42,9 +60,11 @@ if [ -n "$CHANGED_FILES" ]; then
   files=$(comm -12 <(echo $files | tr ' ' '\n' | sort -u) <(echo $CHANGED_FILES | tr ' ' '\n' | sort -u))
 fi
 
+FILES_PER_PROCESS="$(expr $(echo "$files" | grep -o '\n' | wc -l) / $CPU_COUNT + 1)"
```
ctiller, line 63: I expect there's a sweet spot somewhere between `1` and `files/cpus` here -- there's likely to be one batch here that's far more expensive than the others and we spend undue time waiting for it to complete.

I'd suggest quickly trying `files/cpus/2`, `files/cpus/5`, `files/cpus/10` before submission and choosing the best (the additional divisor would create smaller batches so that we'd be able to more effectively backfill cpus that finish earlier).

drfloob: This might depend on the system specs, right? Number of cores and HD/SSD speed, for example. I suspect you're right, but I'll leave further environment tuning for another time.

drfloob: On my 128 core system with SSD:

files/cpus/1: 20s
files/cpus/2: 22s
files/cpus//3: 27s
files/cpus//5: 38s

PR #34841, opened 2023-11-01
[build] Speed up the strict_build CI job

drfloob:
The `bazel_build_with_strict_warnings_linux` job was taking nearly an hour to complete. This change brings it down to ~10m by splitting up the build targets into 7 separate RBE actions.


drfloob:
In CI, the strict build went from 50m to 25m. The longest job is still building `//test/core/end2end/...`. All other jobs complete in ~10m. If we could split that set of targets in half somehow, I'd expect a 15m run time.

veblush (approved):

PR #34847, opened 2023-11-01
[Python Misc] Use template to generate _parallel_compile_patch.py

XuanWang-Amos:
`_parallel_compile_patch.py` exist in multiple places, we should use template system to generate the content.


templates/src/python/_parallel_compile_patch.include
```diff
@@ -0,0 +1,65 @@
+"""Patches the compile() to allow enable parallel compilation of C/C++.
```
gnossen, line 1: Nit: Maybe change the filename to `_parallel_compile_patch.py.include` to indicate this is a Python file.

gnossen (approved):

PR #34857, opened 2023-11-02
[Python] Do not call PyEval_InitThreads

musicinmybrain:
Since Python 3.7, it’s called by [`Py_Initialize`](https://docs.python.org/3/c-api/init.html#c.Py_Initialize) and doesn’t need to be called manually, so this removal has no effect.

Since Python 3.9, it does nothing, and in [Python 3.13](https://docs.python.org/3.13/whatsnew/3.13.html), it [will be removed](https://github.com/python/cpython/issues/105182), which makes this change necessary.

References:

https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads

https://github.com/python/cpython/issues/105182

----

Together with the changes discussed in https://github.com/grpc/grpc/issues/33918 and with a [Cython fix](https://github.com/cython/cython/pull/5791) to be released in 3.0.5, this allowed us to build and test our `grpc` package with Python 3.13.0a1 and Cython 3.0.4.

### DIFF
--- a/src/python/grpcio/grpc/_cython/cygrpc.pyx
+++ b/src/python/grpcio/grpc/_cython/cygrpc.pyx
@@ -80,14 +80,7 @@ include "_cygrpc/aio/server.pyx.pxi"
 #
 # initialize gRPC
 #
-cdef extern from "Python.h":
-
-  int PyEval_InitThreads()
-
 cdef _initialize():
-  # We have Python callbacks called by c-core threads, this ensures the GIL
-  # is initialized.
-  PyEval_InitThreads()
   grpc_set_ssl_roots_override_callback(
           <grpc_ssl_roots_override_callback>ssl_roots_override_callback)
 


musicinmybrain:
Is there a list of the proper component names for use in PR titles somewhere? Poking through recently-merged PR’s, they seem to be more or less `ad hoc`.

musicinmybrain:
For completeness, the supported Python versions are:

https://github.com/grpc/grpc/blob/a62c2607a64edf9484197172108f6163a45f919e/setup.py#L107-L112

hroncok:
> ...and Cython 3.0.4.

3.0.5 if it matters

gnossen (approved):
Thank you for the chance @musicinmybrain !

I'm somewhat wondering why this was added to begin with. Ostensibly, the reason was that the gRPC C++ library performs callbacks on threads not created by the Python interpreter, but instead via raw `pthread_create` calls. The documentation I'm reading for both `PyEval_InitThreads` and `Py_Initialize` suggests that they do _per process_ initialization, not per-thread initialization. If the initialization is per-thread, then the presence of this call makes sense and we'd probably need to replace it with a call to `Py_Initialize`. If the initialization is _per process_ as the documentation currently suggests, then I don't see why this call was necessary to begin with -- we call `_initialize` during `grpc` module load, so there is no danger of a C++-spawned thread attempting to run Python code without `PyEval_InitThreads` having been called first.

Regardless, the fact that `PyEval_InitThreads` is a no-op on 3.9+ and no one has complained, is pretty strong evidence that this change should be just fine.

Thanks again for the thorough investigation and fix! Sounds like Cython 3 support isn't far away for us.

gnossen:
CC @XuanWang-Amos

PR #34859, opened 2023-11-02
[tls] Allow skipping server cert verification when no default roots are present.

matthewstevenson88:
This PR fixes a bug identified in #29667, where the TLS channel credentials still require a trust bundle even if the user has explicitly opted to not verify the server certificate. This PR is based on #29810.

### DIFF
--- a/src/core/tsi/ssl_transport_security.cc
+++ b/src/core/tsi/ssl_transport_security.cc
@@ -2061,9 +2061,6 @@ tsi_result tsi_create_ssl_client_handshaker_factory_with_options(
 
   if (factory == nullptr) return TSI_INVALID_ARGUMENT;
   *factory = nullptr;
-  if (options->pem_root_certs == nullptr && options->root_store == nullptr) {
-    return TSI_INVALID_ARGUMENT;
-  }
 
 #if OPENSSL_VERSION_NUMBER >= 0x10100000
   ssl_context = SSL_CTX_new(TLS_method());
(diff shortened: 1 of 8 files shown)


src/core/tsi/ssl_transport_security.cc
```diff
@@ -2061,9 +2061,6 @@ tsi_result tsi_create_ssl_client_handshaker_factory_with_options(
 
   if (factory == nullptr) return TSI_INVALID_ARGUMENT;
   *factory = nullptr;
-  if (options->pem_root_certs == nullptr && options->root_store == nullptr) {
```
gtcooke94, line 2064: Is there a time where we do still want this return? Perhaps if `skip_server_certificate_verification == false` or something?

test/cpp/end2end/tls_credentials_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  EXPECT_TRUE(result.ok());
+  if (!result.ok()) {
+    gpr_log(GPR_ERROR, "Echo failed: %d, %s, %s",
+            static_cast<int>(result.error_code()),
+            result.error_message().c_str(), result.error_details().c_str());
+  }
+  EXPECT_NE(response.message(), kMessage);
+}
+
+TEST_F(TlsCredentialsTest, SkipServerCertificateVerification) {
```
gtcooke94, line 132: I don't exactly see how this test is verifying we are skipping server cert verification - should we maybe use bad certs/revoked certs for the server?

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Good call, done.

test/cpp/end2end/tls_credentials_test.cc
matthewstevenson88: We're not giving the client any roots (except maybe the OS trust store) and the server is presenting a custom cert, so the client must be skipping it since it does not have the root needed to verify. Does that help clarify?

gtcooke94: Yes that helps, I think it's worth a brief comment explaining

gtcooke94 (approved):

test/cpp/end2end/tls_credentials_test.cc
matthewstevenson88: Done.

Issue #12554, opened 2017-09-13
Stripping call credentials from composite channel credentials should be recursive


ebf42:
The current implementation of duplicate_without_call_credentials for composite channel credentials appears incorrect; it simply increments the refcount on the inner creds:

https://github.com/grpc/grpc/blob/3bf7b714a5458aadfc1e2761ede1213f69ef7959/src/core/lib/security/credentials/composite/composite_credentials.c#L242-L248

In particular, this seems incorrect if the inner creds are themselves composite. It appears this could be fixed simply by recursively calling duplicate_without_call_credentials on the inner creds.

jboeuf:
CC'ing @justinburke and @dgquintas FYI.

ebf42:
Prompted by the stale bot, I took another look; I see the code in question has changed from C to C++, but the bug still appears to exist:
https://github.com/grpc/grpc/blob/a5d4bc14834b8f10fbe65a818ea7eb6c7a8d6da2/src/core/lib/security/credentials/composite/composite_credentials.h#L41-L44

That is, if 'inner_creds_' is itself composite, it seems to me this won't work as intended.

The fix would seem to be as simple as:
```c++
return inner_creds_->duplicate_without_call_credentials();
```

apolcyn:
friendly ping @jiangtaoli2016 if you have time to take a look?

markdroth:
Fixed in https://github.com/grpc/grpc/pull/34860.

PR #34860, opened 2023-11-02
[Security] Fix bug for layered composite creds

gtcooke94:
Address https://github.com/grpc/grpc/issues/12554

The API for `duplicate_without_call_credentials` says 
```
// Creates a version of the channel credentials without any attached call
// credentials. This can be used in order to open a channel to a non-trusted
// gRPC load balancer.
```

As the impl stands right now, because of that description, in the case of layered composite creds, I think the right behavior would be to call down until you get the base cred with no call cred.

In discussing with the team, we do wonder if the use-case of layered composite creds is really something that should be a feature, or if we should be checking during the creation of composite creds to make sure we aren't layering composite creds? @markdroth can you give your thoughts?

### DIFF
--- a/test/core/security/credentials_test.cc
+++ b/test/core/security/credentials_test.cc
@@ -3878,7 +3878,7 @@ TEST(CredentialsTest, TestCompositeChannelCredsCompareSuccess) {
   grpc_channel_credentials_release(composite_creds_2);
 }
 
-TEST(CredentialsTest, RecursiveCompositeCredsCompareSuccess) {
+TEST(CredentialsTest, RecursiveCompositeCredsDuplicateWithoutCallCreds) {
   auto* insecure_creds = grpc_insecure_credentials_create();
   auto inner_fake_creds = MakeRefCounted<fake_call_creds>();
   auto outer_fake_creds = MakeRefCounted<fake_call_creds>();
--- a/test/core/security/credentials_test.cc
+++ b/test/core/security/credentials_test.cc
@@ -3878,6 +3878,23 @@ TEST(CredentialsTest, TestCompositeChannelCredsCompareSuccess) {
   grpc_channel_credentials_release(composite_creds_2);
 }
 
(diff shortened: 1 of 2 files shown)


test/core/security/credentials_test.cc
```diff
@@ -3878,6 +3878,23 @@ TEST(CredentialsTest, TestCompositeChannelCredsCompareSuccess) {
...
+  auto inner_fake_creds = MakeRefCounted<fake_call_creds>();
+  auto outer_fake_creds = MakeRefCounted<fake_call_creds>();
+  auto* inner_composite_creds = grpc_composite_channel_credentials_create(
+      insecure_creds, inner_fake_creds.get(), nullptr);
+  auto* outer_composite_creds = grpc_composite_channel_credentials_create(
+      inner_composite_creds, outer_fake_creds.get(), nullptr);
+  auto duplicate_without_call_creds =
+      outer_composite_creds->duplicate_without_call_credentials();
+  auto type = duplicate_without_call_creds->type();
+  ASSERT_NE(type, inner_composite_creds->type());
```
markdroth, line 3892: Why are we checking that the type does *not* equal something, rather than checking that the type is what we actually expect here?  And actually, why are we checking the type at all rather than just checking that we get the right channel creds instance back?

```
EXPECT_EQ(insecure_creds, duplicate_without_call_creds.get());
```

Note that this should be EXPECT instead of ASSERT.

markdroth (commented):
> In discussing with the team, we do wonder if the use-case of layered composite creds is really something that should be a feature, or if we should be checking during the creation of composite creds to make sure we aren't layering composite creds?

This is definitely the right fix.  There's no reason in principle why we should disallow nested composite creds.

test/core/security/credentials_test.cc
gtcooke94: All good points, I was so focused on making sure we _didn't_ return a composite creds I overlooked just checking that we returned exactly the right thing.
Addressed

markdroth (approved):

Issue #28382, opened 2021-12-16
Cannot force use of only TLS 1.3 ciphers


marekp0:
### What version of gRPC and what language are you using?

C++ 1.38.1 (also tested on 1.42.0, same results; also tried with both git-submodule-provided boringssl and ubuntu-provided openssl, same results)

### What operating system (Linux, Windows,...) and version?

Linux, Ubuntu 20.04

### What runtime / compiler are you using (e.g. python version or version of gcc)

gcc 9.3.0

### What did you do?

```
export GRPC_SSL_CIPHER_SUITES=TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256
./my-grpc-server
```

### What did you expect to see?

A working grpc server

### What did you see instead?

```
E1216 11:39:49.763531003  128930 ssl_transport_security.cc:828] Invalid cipher list: TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256.
E1216 11:39:49.763755474  128930 ssl_security_connector.cc:265] Handshaker factory creation failed with TSI_INVALID_ARGUMENT.
E1216 11:39:49.763841798  128930 server_secure_chttp2.cc:120] {"created":"@1639683589.763815388","description":"Unable to create secure server with credentials of type Ssl","file":"<redacted>/grpc/src/core/ext/transport/chttp2/server/secure/server_secure_chttp2.cc","file_line":102}
```

Some other things I noticed:

- `SSL_CTX_set_cipher_list()` is for TLS 1.2 ciphers only, while `SSL_CTX_set_ciphersuites()` is for TLS 1.3 only ([source](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_ciphersuites.html))
- From grepping the grpc source code, it doesn't seem like grpc ever calls `SSL_CTX_set_ciphersuites()`. Therefore, I'd conclude that it's impossible to set the TLS 1.3 ciphers in grpc from the `GRPC_SSL_CIPHER_SUITES` environment variable, despite [what the list of ciphers defaults to](https://github.com/grpc/grpc/blob/master/src/core/lib/security/security_connector/ssl_utils.cc#L71).

sanchayanghosh:
This PR [#205](https://github.com/grpc/proposal/pull/205) has outlined the proposal for exposing the OpenSSL API to enable customization over TLS versions, ciphers, etc.

[...]

PR #34861, opened 2023-11-02
[tls] Add set min/max TLS version APIs to TLS credentials APIs.

rockspore:
Address #28382. This is a recreation of #31368 except e2e tests are not handled here (yet).


include/grpcpp/security/tls_credentials_options.h
```diff
@@ -108,6 +108,17 @@ class TlsCredentialsOptions {
 
   void set_crl_provider(std::shared_ptr<CrlProvider> crl_provider);
 
+  // Sets the minimum TLS version that will be negotiated during the TLS
+  // handshake. If not set, the underlying SSL library will pick the version
```
matthewstevenson88, line 112: Same here RE min/max being constrained.

include/grpcpp/security/tls_credentials_options.h
```diff
@@ -108,6 +108,17 @@ class TlsCredentialsOptions {
...
+  // automatically.
+  // @param tls_version: The minimum TLS version.
+  void set_min_tls_version(grpc_tls_version tls_version);
+  // Sets the maximum TLS version that will be negotiated during the TLS
+  // handshake. If not set, the underlying SSL library will pick the version
```
matthewstevenson88, line 117: Same here RE min/max being constrained.

src/core/lib/security/credentials/tls/tls_credentials.cc
```diff
@@ -46,6 +46,12 @@ bool CredentialOptionSanityCheck(grpc_tls_credentials_options* options,
     gpr_log(GPR_ERROR, "TLS credentials options is nullptr.");
     return false;
   }
+  // In this case, there will be non-retriable handshake errors.
+  if (options->min_tls_version() == grpc_tls_version::TLS1_3 &&
+      options->max_tls_version() == grpc_tls_version::TLS1_2) {
+    gpr_log(GPR_ERROR, "TLS min version must not be higher than max version.");
```
matthewstevenson88, line 52: Should we use a bit more informative error message here? E.g. if min or max < 1.2, can say the min is too low? if max or min > 1.3, say not supported? and then also have a min <= max check?

matthewstevenson88 (commented):
Just a few minor comments.

Also, once #34859 is merged, could we add to the e2e test where we force TLS 1.2 and force TLS 1.3?

src/core/lib/security/credentials/tls/tls_credentials.cc
rockspore: Added those checks and tests. PTAL

include/grpcpp/security/tls_credentials_options.h
rockspore: Done.

include/grpcpp/security/tls_credentials_options.h
rockspore: Done.

include/grpc/grpc_security.h
```diff
@@ -815,6 +815,26 @@ GRPCAPI void grpc_tls_certificate_provider_release(
...
+ * automatically.
+ */
+GRPCAPI void grpc_tls_credentials_options_set_min_tls_version(
+    grpc_tls_credentials_options* options, grpc_tls_version min_tls_version);
+
+/**
+ * EXPERIMENTAL API - Subject to change
+ *
+ * Sets the maximum TLS version that will be negotiated during the TLS
+ * handshake. If not set, the underlying SSL library will pick the version
```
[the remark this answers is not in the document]
rockspore: Fixed.

include/grpc/grpc_security.h
```diff
@@ -815,6 +815,26 @@ GRPCAPI void grpc_tls_certificate_provider_release(
  */
 GRPCAPI grpc_tls_credentials_options* grpc_tls_credentials_options_create(void);
 
+/**
+ * EXPERIMENTAL API - Subject to change
+ *
+ * Sets the minimum TLS version that will be negotiated during the TLS
+ * handshake. If not set, the underlying SSL library will pick the version
```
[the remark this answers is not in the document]
rockspore: Fixed.

rockspore (commented):
Thanks. Will added e2e tests as a follow up.

test/cpp/server/credentials_test.cc
```diff
@@ -277,6 +277,40 @@ TEST(CredentialsTest, MultipleServerCredentialsOneCrlProviderDoesNotLeak) {
   EXPECT_NE(server_creds_2, nullptr);
 }
 
+TEST(
+    CredentialsTest,
+    TlsChannelCredentialsWithStaticDataCertificateProviderAndGoodMinMaxTlsVersions) {
```
gtcooke94, line 282: Does it particularly matter that this is `WithStaticDataCertificateProvider`? It's not really involved in the test, right? It's just a requirement to have to build `TlsServerCredentials`?
If that's all true, I think we can remove that part from the name of the test and have the test name focus on what is actually being exercised

gtcooke94 (approved):
LGTM, one minor comment

src/core/lib/security/credentials/tls/grpc_tls_credentials_options.cc
```diff
@@ -149,3 +149,15 @@ void grpc_tls_credentials_options_set_crl_provider(
   GPR_ASSERT(options != nullptr);
   options->set_crl_provider(provider);
 }
+
+void grpc_tls_credentials_options_set_min_tls_version(
```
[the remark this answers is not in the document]
rockspore: Those fields are already there in C-core layer so that python script was taken care of before IIUC. The methods I introduce are just for the C++ or other wrapping langs to set the versions, as the option itself is opaque in the header.

test/cpp/server/credentials_test.cc
rockspore: Yeah server side must have a root store (trust bundle) with this TLS API I think.

matthewstevenson88 (approved):

test/cpp/server/credentials_test.cc
rockspore: Scratch what I said, so I think the ctor of the options requires a cert provider. If a nullptr works I will remove this chunk.

markdroth (approved):

PR #34864, opened 2023-11-02
[Security] Handshaker status return

gtcooke94:
Address the first (easy) half of https://github.com/grpc/grpc/issues/32598

### DIFF
--- a/src/core/lib/security/transport/security_handshaker.cc
+++ b/src/core/lib/security/transport/security_handshaker.cc
@@ -592,16 +592,14 @@ class FailHandshaker : public Handshaker {
   void DoHandshake(grpc_tcp_server_acceptor* /*acceptor*/,
                    grpc_closure* on_handshake_done,
                    HandshakerArgs* args) override {
-    grpc_error_handle error = GRPC_ERROR_CREATE(absl::StrCat(
-        "Failed to create security handshaker: ", status_.ToString()));
-    grpc_endpoint_shutdown(args->endpoint, error);
+    grpc_endpoint_shutdown(args->endpoint, status_);
     grpc_endpoint_destroy(args->endpoint);
     args->endpoint = nullptr;
     args->args = ChannelArgs();
     grpc_slice_buffer_destroy(args->read_buffer);
     gpr_free(args->read_buffer);
     args->read_buffer = nullptr;
-    ExecCtx::Run(DEBUG_LOCATION, on_handshake_done, error);
+    ExecCtx::Run(DEBUG_LOCATION, on_handshake_done, status_);
   }
 
  private:
--- a/src/core/lib/security/transport/security_handshaker.cc
+++ b/src/core/lib/security/transport/security_handshaker.cc
@@ -664,9 +664,9 @@ RefCountedPtr<Handshaker> SecurityHandshakerCreate(
         absl::Status(handshaker.status().code(),
                      absl::StrCat("Failed to create security handshaker: ",
                                   handshaker.status().message())));
+  } else if (*handshaker == nullptr) {
     // TODO(gtcooke94) Once all TSI impls are updated to pass StatusOr<> instead
     // of null, we should change this to use absl::InternalError().
-  } else if (*handshaker == nullptr) {
     return MakeRefCounted<FailHandshaker>(
         absl::UnknownError("Failed to create security handshaker."));
   } else {
--- a/src/core/lib/security/transport/security_handshaker.cc
+++ b/src/core/lib/security/transport/security_handshaker.cc
@@ -603,8 +603,8 @@ class FailHandshaker : public Handshaker {
   }
 
  private:
-  absl::Status status_;
   ~FailHandshaker() override = default;
+  absl::Status status_;
 };
 
 //
--- a/src/core/lib/security/transport/security_handshaker.cc
+++ b/src/core/lib/security/transport/security_handshaker.cc
@@ -660,10 +660,15 @@ RefCountedPtr<Handshaker> SecurityHandshakerCreate(
   // If no TSI handshaker was created, return a handshaker that always fails.
   // Otherwise, return a real security handshaker.
   if (!handshaker.ok()) {
-    return MakeRefCounted<FailHandshaker>(handshaker.status());
+    return MakeRefCounted<FailHandshaker>(
+        absl::Status(handshaker.status().code(),
+                     absl::StrCat("Failed to create security handshaker: ",
+                                  handshaker.status().message())));
+    // TODO(gtcooke94) Once all TSI impls are updated to pass StatusOr<> instead
+    // of null, we should change this to use absl::InternalError().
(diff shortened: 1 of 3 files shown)


matthewstevenson88 (approved):
LGTM, but please wait for @markdroth's approval before merging.

src/core/lib/security/transport/security_handshaker.cc
```diff
@@ -652,14 +655,17 @@ class ServerSecurityHandshakerFactory : public HandshakerFactory {
...
-    tsi_handshaker* handshaker, grpc_security_connector* connector,
-    const ChannelArgs& args) {
+    absl::StatusOr<tsi_handshaker*> handshaker,
+    grpc_security_connector* connector, const ChannelArgs& args) {
   // If no TSI handshaker was created, return a handshaker that always fails.
   // Otherwise, return a real security handshaker.
-  if (handshaker == nullptr) {
-    return MakeRefCounted<FailHandshaker>();
+  if (!handshaker.ok()) {
+    return MakeRefCounted<FailHandshaker>(handshaker.status());
```
markdroth, line 663: Please change `handshaker.status()` to:

```
absl::Status(
    handshaker.status().code(),
    absl::StrCat("Failed to create security handshaker: ", handshaker.status().message())
```

This way, we're adding a useful prefix to the status message but leaving the code as-is.

src/core/lib/security/transport/security_handshaker.cc
```diff
@@ -652,14 +655,17 @@ class ServerSecurityHandshakerFactory : public HandshakerFactory {
...
+  } else if (*handshaker == nullptr) {
+    return MakeRefCounted<FailHandshaker>(absl::UnknownError(
+        "Handshaker creation was passed a nullptr tsi_handshaker."));
```
markdroth, line 666: Let's stick with the existing message "Failed to create security handshaker:".

src/core/lib/security/transport/security_handshaker.cc
```diff
@@ -652,14 +655,17 @@ class ServerSecurityHandshakerFactory : public HandshakerFactory {
...
+    absl::StatusOr<tsi_handshaker*> handshaker,
+    grpc_security_connector* connector, const ChannelArgs& args) {
   // If no TSI handshaker was created, return a handshaker that always fails.
   // Otherwise, return a real security handshaker.
-  if (handshaker == nullptr) {
-    return MakeRefCounted<FailHandshaker>();
+  if (!handshaker.ok()) {
+    return MakeRefCounted<FailHandshaker>(handshaker.status());
+  } else if (*handshaker == nullptr) {
+    return MakeRefCounted<FailHandshaker>(absl::UnknownError(
```
markdroth, line 665: Please add a TODO that once all TSI implementations are updated to pass a `StatusOr<>` instead of null, we should change this to use `absl::InternalError()`.

src/core/lib/security/transport/security_handshaker.cc
```diff
@@ -585,13 +586,14 @@ void SecurityHandshaker::DoHandshake(grpc_tcp_server_acceptor* /*acceptor*/,
...
  public:
+  explicit FailHandshaker(absl::Status status) : status_(std::move(status)) {}
   const char* name() const override { return "security_fail"; }
   void Shutdown(grpc_error_handle /*why*/) override {}
   void DoHandshake(grpc_tcp_server_acceptor* /*acceptor*/,
                    grpc_closure* on_handshake_done,
                    HandshakerArgs* args) override {
-    grpc_error_handle error =
-        GRPC_ERROR_CREATE("Failed to create security handshaker");
+    grpc_error_handle error = GRPC_ERROR_CREATE(absl::StrCat(
```
markdroth, line 595: Once you make the changes I suggested below, we won't need to do this here.  We can remove the local `error` variable and just directly use `status_` in its place.

src/core/lib/security/transport/security_handshaker.cc
```diff
@@ -660,10 +660,15 @@ RefCountedPtr<Handshaker> SecurityHandshakerCreate(
   // If no TSI handshaker was created, return a handshaker that always fails.
   // Otherwise, return a real security handshaker.
   if (!handshaker.ok()) {
-    return MakeRefCounted<FailHandshaker>(handshaker.status());
+    return MakeRefCounted<FailHandshaker>(
+        absl::Status(handshaker.status().code(),
+                     absl::StrCat("Failed to create security handshaker: ",
+                                  handshaker.status().message())));
+    // TODO(gtcooke94) Once all TSI impls are updated to pass StatusOr<> instead
```
markdroth, line 667: Please move this TODO to line 670.

src/core/lib/security/transport/security_handshaker.cc
markdroth: We don't need to call `GRPC_ERROR_CREATE()` at all.  Just pass `status_` to `ExecCtx::Run()` instead of `error`.

Note that `grpc_error_handle` used to be a custom type, but these days it's actually a typedef for `absl::Status`.  But there's a lot of legacy machinery for adding structured annotations to the `absl::Status` to simulate the legacy behavior of `grpc_error_handle`, and we're trying to slowly get rid of that tech debt in the code base.  `GRPC_ERROR_CREATE()` is one of the functions that add those legacy annotations, which is why I'm saying that we should get rid of it here.

src/core/lib/security/transport/security_handshaker.cc
gtcooke94: Done

src/core/lib/security/transport/security_handshaker.cc
gtcooke94: > these days it's actually a typedef for absl::Status

Ah this was my missing link, changed

src/core/lib/security/transport/security_handshaker.cc
```diff
@@ -585,24 +586,24 @@ void SecurityHandshaker::DoHandshake(grpc_tcp_server_acceptor* /*acceptor*/,
...
     args->args = ChannelArgs();
     grpc_slice_buffer_destroy(args->read_buffer);
     gpr_free(args->read_buffer);
     args->read_buffer = nullptr;
-    ExecCtx::Run(DEBUG_LOCATION, on_handshake_done, error);
+    ExecCtx::Run(DEBUG_LOCATION, on_handshake_done, status_);
   }
 
  private:
+  absl::Status status_;
```
markdroth, line 606: Data members need to come after the dtor.

https://google.github.io/styleguide/cppguide.html#Declaration_Order

markdroth (approved):
Just one remaining nit, we can get this landed once that's addressed.

src/core/lib/security/transport/security_handshaker.cc
gtcooke94: Fixed

markdroth (approved):

PR #34874, opened 2023-11-03
[ssl] Support Windows system roots.

matthewstevenson88:
This PR is copied from #34276, since I did not have permissions to add commits to it. That PR has been verified to work (see the top-level description). This PR just makes the gRPC tests pass (e.g. adding includes, clang formatting).


src/core/lib/security/security_connector/load_system_roots_windows.cc
```diff
@@ -0,0 +1,90 @@
...
+#include <grpc/support/alloc.h>
+#include <grpc/support/log.h>
+
+#include "src/core/lib/gpr/useful.h"
+#include "src/core/lib/security/security_connector/load_system_roots.h"
+
+namespace grpc_core {
+namespace {
+
+std::string utf8Encode(const std::wstring& wstr) {
```
gtcooke94, line 41: If you used any reference besides the docs for `WideCharToMultiByte` could you add a comment?

src/core/lib/security/security_connector/load_system_roots_windows.cc
```diff
@@ -0,0 +1,90 @@
...
+  if (!hRootCertStore) {
+    return grpc_empty_slice();
+  }
+
+  // Load all root certificates from certificate store.
+  PCCERT_CONTEXT pCert = NULL;
+  while ((pCert = CertEnumCertificatesInStore(hRootCertStore, pCert)) != NULL) {
+    // Append each certificate in PEM format.
+    DWORD size = 0;
+    CryptBinaryToStringW(pCert->pbCertEncoded, pCert->cbCertEncoded,
```
gtcooke94, line 68: I've been having some issues of my own with the `W` vs. `A` postfixed APIs for ascii vs. unicode, I'm wondering how the CI will like this

src/core/lib/security/security_connector/load_system_roots_windows.cc
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  CertCloseStore(hRootCertStore, 0);
+  if (bundle_string.size() == 0) {
+    return grpc_empty_slice();
+  }
+
+  char* result_bundle_string =
+      static_cast<char*>(gpr_zalloc(bundle_string.size() + 1));
+  strcpy(result_bundle_string, bundle_string.data());
```
gtcooke94, line 83: I had some issues with `strcpy` mixing with the `gpr` alloc methods before causing ASAN errors with alloc/dealloc mismatch. If I understand them right, I think this is actually fine, but I wanted to point out that `gpr_strdup` exists to duplicate strings. https://github.com/grpc/grpc/blob/52f9e011f3106c0bd184163cf4ec74625c110480/include/grpc/support/string_util.h#L32-L34

I think you could just do
`char* result_bundle_string = gpr_strdup(bundle_string.data());`

src/core/lib/security/security_connector/load_system_roots_windows.cc
matthewstevenson88: Resolving since I didn't use any references, just relied on the sample PR that was provided.

markdroth, line 41: Please call this `Utf8Encode()`, as per the style guide.

https://google.github.io/styleguide/cppguide.html#Function_Names

src/core/lib/security/security_connector/load_system_roots_windows.cc
```diff
@@ -0,0 +1,90 @@
...
+#include "src/core/lib/gpr/useful.h"
+#include "src/core/lib/security/security_connector/load_system_roots.h"
+
+namespace grpc_core {
+namespace {
+
+std::string utf8Encode(const std::wstring& wstr) {
+  if (wstr.empty()) return std::string();
+
+  int sizeNeeded = WideCharToMultiByte(CP_UTF8, 0, &wstr[0], (int)wstr.size(),
```
markdroth, line 44: Please name variables like `size_needed`, `str_to`, etc, as per the style guide.

https://google.github.io/styleguide/cppguide.html#Variable_Names

src/core/lib/security/security_connector/load_system_roots_windows.cc
markdroth: In this case, you can just use `grpc_slice_from_cpp_string(std::move(bundle_string))`.

src/core/lib/security/security_connector/load_system_roots_windows.cc
```diff
@@ -0,0 +1,90 @@
...
+                                       NULL, 0, NULL, NULL);
+  std::string strTo(sizeNeeded, 0);
+  WideCharToMultiByte(CP_UTF8, 0, &wstr[0], (int)wstr.size(), &strTo[0],
+                      sizeNeeded, NULL, NULL);
+  return strTo;
+}
+
+}  // namespace
+
+grpc_slice LoadSystemRootCerts() {
```
markdroth, line 54: Not directly related to this PR, but at some point it would be good to change this function to return `grpc_core::Slice` instead of `grpc_slice`.  The former is a C++ type with much more idiomatic semantics.

test/core/security/system_roots_test.cc
```diff
@@ -47,6 +50,7 @@
 namespace grpc {
 namespace {
 
+#if defined(GPR_LINUX) || defined(GPR_FREEBSD) || defined(GPR_APPLE)
 TEST(AbsoluteFilePathTest, ConcatenatesCorrectly) {
```
markdroth, line 54: I assume there's a reason we can't run these tests on Windows?

src/core/lib/security/security_connector/load_system_roots_windows.cc
matthewstevenson88: Done (sorry this was ported from an older sample PR and I forgot to fix up the readability issues).

src/core/lib/security/security_connector/load_system_roots_windows.cc
matthewstevenson88: Done.

src/core/lib/security/security_connector/load_system_roots_windows.cc
matthewstevenson88: Seems to be working, thankfully. :)

markdroth (approved):
Looks great!

gtcooke94 (approved):

PR #34876, opened 2023-11-03
[alpn] Remove grpc-exp experimental ALPN protocol.

### LINKED ISSUE #21619: Remove grpc-exp from ALPN
grpc-exp was an experiment that didn't go anywhere. We should remove this vestige.

Java issue: grpc/grpc-java#6591

matthewstevenson88:
This fixes #21619. This experimental ALPN protocol has already been removed from the other gRPC stacks.

### DIFF
--- a/src/core/ext/transport/chttp2/alpn/alpn.cc
+++ b/src/core/ext/transport/chttp2/alpn/alpn.cc
@@ -25,7 +25,7 @@
 #include "src/core/lib/gpr/useful.h"
 
 // in order of preference
-static const char* const supported_versions[] = {"grpc-exp", "h2"};
+static const char* const supported_versions[] = {"h2"};
 
 int grpc_chttp2_is_alpn_version_supported(const char* version, size_t size) {
   size_t i;
--- a/test/core/handshake/client_ssl.cc
+++ b/test/core/handshake/client_ssl.cc
@@ -153,27 +153,19 @@ static int alpn_select_cb(SSL* /*ssl*/, const uint8_t** out, uint8_t* out_len,
   *out_len = static_cast<uint8_t>(
       strlen(reinterpret_cast<const char*>(alpn_preferred)));
 
-  // Validate that the ALPN list includes "h2" and "grpc-exp", that "grpc-exp"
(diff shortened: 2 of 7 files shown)


rockspore (approved):

src/core/ext/transport/chttp2/alpn/alpn.cc
```diff
@@ -25,7 +25,7 @@
 #include "src/core/lib/gpr/useful.h"
 
 // in order of preference
-static const char* const supported_versions[] = {"grpc-exp", "h2"};
+static const char* const supported_versions[] = {"h2"};
```
gtcooke94, line 28: Do we plan to have any other supported versions? Otherwise I wonder if it's worth it to just remove this complexity altogether and have a non-arr const value for `h2`

gtcooke94 (approved):
LGTM, one comment if you want to make the change bigger

src/core/ext/transport/chttp2/alpn/alpn.cc
matthewstevenson88: I'm planning a follow-up PR to clean up this file (some of the APIs, e.g. `grpc_chttp2_get_alpn_version_index`, are now unused). Do you mind if we defer this to the cleanup PR?

matthewstevenson88:
@markdroth @ejona86 Could one of you confirm that this should be safe? The `grpc-exp` has already been removed from Go and Java, but I'd like to confirm that there are no edge cases for the C-core or wrapped languages that are still dependent on this ALPN protocol.

ctiller:
There's talk in the http2 community of introducing a new version of the
http2 protocol (which would get a new string) to properly deal with rapid
reset.

We should probably keep this machinery around for that.

matthewstevenson88:
@markdroth or @ejona86 Could one of you confirm that this should be safe, as far as you're aware?

ejona86 (commented):
I believe this to be safe

matthewstevenson88:
Thanks Eric! Will merge when the repo is unlocked.

PR #34878, opened 2023-11-03
[TSI] not configure SNI if it looks like an IP address

rockspore:
IP addressed are not allowed as host names. So we want to ignore the given server_name_indication when it looks like an IP.

### DIFF
--- a/src/core/tsi/ssl_transport_security.cc
+++ b/src/core/tsi/ssl_transport_security.cc
@@ -1831,8 +1831,10 @@ tsi_result tsi_ssl_client_handshaker_factory_create_handshaker(
     const char* server_name_indication, size_t network_bio_buf_size,
     size_t ssl_bio_buf_size, tsi_handshaker** handshaker) {
   return create_tsi_ssl_handshaker(
-      factory->ssl_context, 1, server_name_indication, network_bio_buf_size,
-      ssl_bio_buf_size, &factory->base, handshaker);
+      factory->ssl_context, 1,
+      looks_like_ip_address(server_name_indication) ? nullptr
+                                                    : server_name_indication,
+      network_bio_buf_size, ssl_bio_buf_size, &factory->base, handshaker);
 }
 
 void tsi_ssl_client_handshaker_factory_unref(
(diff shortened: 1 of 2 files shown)


rockspore:
What I am not sure is if this would introduce any performance concern since we are examining the char array every time creating a client handshaker.

matthewstevenson88:
> What I am not sure is if this would introduce any performance concern since we are examining the char array every time creating a client handshaker.

I think this should be fine, since the `looks_like_ip_address` function is just doing a linear pass of the target name, and target names are generally relatively short.

matthewstevenson88 (approved):
LGTM, but @gtcooke94 could you also take a look?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1831,8 +1831,10 @@ tsi_result tsi_ssl_client_handshaker_factory_create_handshaker(
     const char* server_name_indication, size_t network_bio_buf_size,
     size_t ssl_bio_buf_size, tsi_handshaker** handshaker) {
   return create_tsi_ssl_handshaker(
-      factory->ssl_context, 1, server_name_indication, network_bio_buf_size,
-      ssl_bio_buf_size, &factory->base, handshaker);
+      factory->ssl_context, 1,
+      looks_like_ip_address(server_name_indication) ? nullptr
```
gtcooke94, line 1835: I think it makes a little bit more sense to do this check in `create_tsi_ssl_handshaker` - that's where other checks regarding `server_name_indication` already are.
Plus this method is just a passthrough right now, so I think it's easier to keep this fn call clean and do it all in `create_tsi_ssl_handshaker`

gtcooke94:
Also I see this comment on `looks_like_ip` - do we need to enhance this to generally handle all IPs?
https://github.com/grpc/grpc/blob/78825f79caec2a4a2d6afa952f7b9bad954ccb6f/src/core/tsi/ssl_transport_security.cc#L237-L239

src/core/tsi/ssl_transport_security.cc
rockspore: Makes sense. Done.

rockspore:
> Also I see this comment on `looks_like_ip` - do we need to enhance this to generally handle all IPs?
> 
> https://github.com/grpc/grpc/blob/78825f79caec2a4a2d6afa952f7b9bad954ccb6f/src/core/tsi/ssl_transport_security.cc#L237-L239

If I didn't miss any edge cases, anything that `looks_like_ip` returns true for is definitely NOT a valid host name, even if it's not an IP. So we should be good for now. A more rigorous check doesn't seem cheap to me.

gtcooke94 (approved):

PR #34891, opened 2023-11-06
[chttp2] Remove unused fields from chttp2_transport

ananda1066:


src/core/ext/transport/chttp2/transport/internal.h
```diff
@@ -559,9 +552,6 @@ struct grpc_chttp2_transport final
   // What percentage of rst_stream frames on the server should cause a ping
   // frame to be generated.
   uint8_t ping_on_rst_stream_percent;
-
-  /// write execution state of the transport
-  grpc_chttp2_write_state write_state = GRPC_CHTTP2_WRITE_STATE_IDLE;
```
Vignesh2208, line 564: I think write_state is used in the transport. It shouldn't be deleted

ananda1066: Yeah I just moved it up in the struct for better cache alignment. It's on line 504 now.

Vignesh2208 (approved):

ctiller (approved):

PR #34899, opened 2023-11-07
[boringssl] Update boringssl 202311062145

markdroth:


stanley-cheung (approved):

markdroth:
Looks like this is blocked on the cmake upgrade, so we're going to skip this boringssl upgrade for now.

PR #34919, opened 2023-11-10
[deps] Fix breakage when built against protobuf head

acozzette:
A recent change to protobuf's `rules_python` version caused gRPC's Bazel build to break. Part of the problem seemed to be that gRPC used a different name for the dependency, `io_bazel_rules_python`, and this resulted in two copies of `rules_python` getting included in the build. This commit makes gRPC use the `rules_python` name so that it can choose its own version to pin and so that there's only one copy. That resulted in another breakage, but I was able to fix that by upgrading to the most recent release. I dropped the patch to `rules_python`, since hopefully it is no longer needed.
    
Another necessary fix was to update references to `@utf8_range`, since this has been merged into the protobuf Bazel repo.


acozzette:
I pushed another change that should fix the utf8_range breakage, but I couldn't figure out a good way to set this up so that it's compatible with both protobuf v25.0 and protobuf head. This fixes it against protobuf head but not protobuf v25.0. So with this change, `./tools/buildgen/generate_projects.sh` will not work unless you first update the protobuf dependency in `grpc_deps.bzl`. Is this OK? Or maybe we should just bump the protobuf version in `grpc_deps.bzl` to address this?

yashykt:
@gnossen @XuanWang-Amos ping

gnossen:
@acozzette Is this completely resolved by https://github.com/grpc/grpc/pull/34945 ?

acozzette:
I have forgotten some of the context about this but it looks to me like #34945 should have fixed this.

PR #34925, opened 2023-11-13
[EventEngine] Require EventEngine::Cancel return value to be used

drfloob:


markdroth:
I have mixed feelings about this.  On one hand, I agree in principle that requiring callers to think about this is a good thing.  But on the other hand, a quick scan of callers of `Cancel()` suggests that not caring about the result is actually the most common pattern, and making the common case more cumbersome seems sub-optimal.

I'd like input from @ctiller on this.

ctiller:
Mixed feelings too.

I can see bugs creeping in on critical paths if we do not do this: there we want to avoid allocations for callback creation, which means eschewing RAII in the AnyInvocable creation for example -- and in those cases we will need to do some cleanup on the Cancel path.

On the other hand, most of these will not be that path... so maybe we live with the occasional unforced error?

If we do do this, I'd much rather see `std::ignore = foo->Cancel()` than `(void)foo->Cancel()` because it calls out what's actually happening far better.

drfloob:
I have a slight preference for the added verbosity, since we gain clarity of intention. It's a gift to future developers bootstrapping on this project, knowing that these call sites are meant to ignore the result. There's also the fact that we can avoid some cancellation oversights as you both pointed out, but I agree there aren't many places where it makes much difference.

drfloob:
Consensus is that this is not really worth the added verbosity. Closing this PR.

PR #34930, opened 2023-11-13
[ci/cd] Hash-pin GitHub Actions, set up dependabot to keep them updated

### LINKED ISSUE #34929: Hash-pin GitHub Actions used in workflows
When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v4`). However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

These hashes can be automatically updated by dependabot. You'll receive a single periodic PR updating the hashes and version comments of all Actions with new versions (see [this PR](https://github.com/pnacht/libarchive/pull/9)).

I'll send a PR pinning the Actions along with this issue.

---

Disclosure: My name is Pedro and I work with Google and the [OpenSSF](https://openssf.org/) to improve the supply-chain security of the open-source ecosystem.

pnacht:
Fixes #34929.

This PR hash-pins all Actions used in workflows and sets up dependabot to keep them up-to-date.

Dependabot will send at most one PR per month. That PR will update the hashes and version comments of all Actions with new versions.

I also suggest you enable Dependabot Security Updates in the repo's [Code security & analysis](https://github.com/grpc/grpc/settings/security_analysis) settings (if you haven't already). This will make Dependabot send a PR as soon as a dependency is found to have a vulnerability.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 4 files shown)


eugeneo (approved):

yashykt (approved):

sergiitk:
@XuanWang-Amos neat, we should do the same to the migrated psm-interop workflow: https://github.com/grpc/psm-interop/blob/main/.github/workflows/psm-interop.yaml

PR #28963, opened 2022-02-24
Switch .py shebang to python3

jtattermusch:
Followup for python3 migration.

PR #34940, opened 2023-11-14
[benchmark] Use python3 for subprocesses, too

lepistone:
After #28963, the Python entry points run with "python3" found on the
PATH.

It makes sense to use the same for subprocesses, for two reasons:

- There might be no "python" in the PATH, as in my Mac with brew
- Having a different Python for subprocesses, especially if it's a
  Python 2, is asking for trouble.


gnossen (approved):

lepistone:
The builds fails because of missing labels. Can you fix that @gnossen ? Thank you!

lepistone:
Hi @gnossen and happy new year! Do you think we can merge this? Thank you!

PR #34941, opened 2023-11-14
Fix Python DeprecationWarning: 'pipes'

lepistone:
Starting from Python 3.11, the pipes module produces this warning:

DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13

Turns out that in this repo the pipes module is only used for the
"quote" function which is turn directly taken from the shlex module [1].

The shlex module is not deprecated as of today and is already used in
other places in this repo. The function shlex.quote has been around
since the ancient Python 3.3.

[1] https://github.com/python/cpython/blob/3.11/Lib/pipes.py#L64-L66

### DIFF
--- a/tools/run_tests/run_performance_tests.py
+++ b/tools/run_tests/run_performance_tests.py
@@ -21,8 +21,8 @@ import collections
 import itertools
 import json
 import os
-import pipes
 import re
+import shlex
 import sys
 import time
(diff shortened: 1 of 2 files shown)


lepistone:
I don't know about the failing checks. They seem unrelated.

PR #34942, opened 2023-11-14
[xDS] Delay ADS read when watchers need time

### LINKED ISSUE #34099: xDS: implement ADS stream flow control mechanism
CC @ejona86 @dfawley @ctiller 

We need to implement a flow control push-back mechanism for the ADS stream inside the XdsClient.  Specifically:
- Change the watcher API such that when it returns a new resource, it also sends a callback that will be invoked when the watcher has finished processing the update.  (The callback can be held in an RAII object to make sure it gets called at the appropriate time.)
- In the XdsClient, whenever we get an ADS response and send updates to the watchers, we will wait for all watchers to invoke their callback before we start the next read on the ADS stream.  (When a read does not get started right away, that implicitly causes flow-control push-back.)
- The read state here needs to be tracked only within the context of a single stream -- i.e., within the context of a single `AdsCallState` object.  If the stream ends (and the `AdsCallState` instance is destroyed) before all of the callbacks are invoked, then we don't need to retain any of this throttling state to the next `AdsCallState` instance.

Note that this will require changes to the `XdsTransport` API.  Currently, the `XdsTransport::StreamingCall` object will automatically start the next read after the previous one completes.  Now we will need to change the API to allow the `XdsClient` to explicitly start the next read when it is ready.  And that will require some internal changes as well (CC @allanrbo).

eugeneo:
Fixes #34099.


src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
```diff
@@ -244,7 +250,8 @@ class XdsClusterResolverLb : public LoadBalancingPolicy {
       // in methods of this class rather than in lambdas to work around an MSVC
       // bug.
       void OnResourceChangedHelper(
-          std::shared_ptr<const XdsEndpointResource> update) {
+          std::shared_ptr<const XdsEndpointResource> update,
+          RefCountedPtr<XdsApi::ReadDelayHandle> /*read_delay_handle*/) {
```
markdroth, line 254: I think this needs to be passed through to `OnEndpointChanged()`.

src/core/ext/xds/xds_server_config_fetcher.cc
```diff
@@ -292,7 +297,8 @@ class XdsServerConfigFetcher::ListenerWatcher::FilterChainMatchManager::
         filter_chain_match_manager_(std::move(filter_chain_match_manager)) {}
 
   void OnResourceChanged(
-      std::shared_ptr<const XdsRouteConfigResource> route_config) override {
+      std::shared_ptr<const XdsRouteConfigResource> route_config,
+      RefCountedPtr<ReadDelayHandle> /* read_delay_handle */) override {
```
markdroth, line 301: If the parameter is not given a name, does it actually stay in scope until the function returns?  I'm not sure, but we should verify.

markdroth (commented):
I haven't yet reviewed the tests, but here are my comments on the code.

src/core/ext/xds/xds_client.cc
```diff
@@ -1032,14 +1036,21 @@ void XdsClient::ChannelState::AdsCallState::OnRequestSent(bool ok) {
...
+bool XdsClient::ChannelState::AdsCallState::OnRecvMessage(
     absl::string_view payload) {
   {
     MutexLock lock(&xds_client()->mu_);
-    if (!IsCurrentCallOnChannel()) return;
+    if (!IsCurrentCallOnChannel()) return true;
+    RefCountedPtr<XdsApi::ReadDelayHandle> read_delay_handle =
+        MakeRefCounted<XdsApi::ReadDelayHandle>([ref = Ref()]() {
+          if (ref->call_ != nullptr) {
+            ref->call_->Read();
```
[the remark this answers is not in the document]
eugeneo: Please review. It is now in `XdsClient::ReadDelayHandle::~ReadDelayHandle`

src/core/ext/xds/xds_api.h
```diff
@@ -49,6 +51,20 @@ class XdsClient;
 // - CSDS response generation
 class XdsApi {
  public:
+  class ReadDelayHandle : public InternallyRefCounted<ReadDelayHandle> {
+   public:
+    explicit ReadDelayHandle(absl::AnyInvocable<void()> read)
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/xds/xds_api.h
```diff
@@ -49,6 +51,20 @@ class XdsClient;
 // - CSDS response generation
 class XdsApi {
  public:
+  class ReadDelayHandle : public InternallyRefCounted<ReadDelayHandle> {
```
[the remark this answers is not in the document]
eugeneo: Done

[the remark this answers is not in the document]
eugeneo: Done

test/core/ext/xds/xds_transport_grpc_test.cc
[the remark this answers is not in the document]
eugeneo: Done, but not sure if I will be able to build this test case for this PR. I may remove this and work separately.

src/core/ext/xds/xds_transport.h
```diff
@@ -61,7 +60,7 @@ class XdsTransportFactory : public InternallyRefCounted<XdsTransportFactory> {
       virtual void SendMessage(std::string payload) = 0;
 
       // Resumes reading messages from the stream.
```
markdroth, line 62: This should say "Starts a recv_message operation on the stream."

src/core/ext/xds/xds_transport_grpc.cc
```diff
@@ -205,15 +205,18 @@ void GrpcXdsTransportFactory::GrpcXdsTransport::GrpcStreamingCall::
...
+void GrpcXdsTransportFactory::GrpcXdsTransport::GrpcStreamingCall::
+    StartRecvMessage() {
   grpc_op op;
   memset(&op, 0, sizeof(op));
   op.op = GRPC_OP_RECV_MESSAGE;
-  op.data.recv_message.recv_message = &self->recv_message_payload_;
-  GPR_ASSERT(self->call_ != nullptr);
+  op.data.recv_message.recv_message = &recv_message_payload_;
+  GPR_ASSERT(call_ != nullptr);
   // Reuses the "OnResponseReceived" ref taken in ctor.
```
markdroth, line 217: Once you make the change that I suggested elsewhere to require the `XdsClient` to call `StartRecvMessage()` even for the initial read, I think this will need to take a new ref:

```
Ref(DEBUG_LOCATION, "OnResponseReceived").release();
```

And we'll need `OnResponseReceived()` to unconditionally unref.

src/core/ext/xds/xds_transport.h
```diff
@@ -58,6 +59,9 @@ class XdsTransportFactory : public InternallyRefCounted<XdsTransportFactory> {
       // Only one message will be in flight at a time; subsequent
       // messages will not be sent until this one is done.
       virtual void SendMessage(std::string payload) = 0;
+
+      // Resumes reading messages from the stream.
+      virtual void Read() = 0;
```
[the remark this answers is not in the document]
markdroth: > From the code, it looks like the connection starts by sending a request that will ultimately call `OnResponseReceived`, which will trigger read, as far as I understand.

Right, that's how it works today.  What I'm suggesting is that we change it such that creating a call does *not* automatically start a read; instead, the caller needs to explicitly call `StartRecvMessage()`.  This is consistent with the semantics of the C++ callback API, in which creating a reactor does not automatically start a read, so you need to explicitly request one.

src/core/ext/xds/xds_client.h
```diff
@@ -236,6 +239,19 @@ class XdsClient : public DualRefCounted<XdsClient> {
     absl::Status status_;
   };
 
+ public:
+  class ReadDelayHandle : public RefCounted<ReadDelayHandle> {
```
markdroth, line 243: I'd prefer not to have to put a part of the public API between a bunch of private things like this.  Instead, please do what I originally suggested, which is to have a base class here that does not depend on anything private, and then a concrete subclass in the .cc file that does depend on the private types.  In other words, here in the .h file, we can say this:

```
 public:
  class ReadDelayHandle : public RefCounted<ReadDelayHandle> {
   public:
    static RefCountedPtr<ReadDelayHandle> NoWait() { return nullptr; }
  };
```

Then in the .cc file, we can say something like this:

```
class AdsCallState {
 public:
  // ...

 private:
  class AdsReadDelayHandle : public ReadDelayHandle {
   public:
    explicit AdsReadDelayHandle(RefCountedPtr<AdsCallState> ads_calld)
        : ads_calld_(std::move(ads_calld)) {}
    ~AdsReadDelayHandle() override;

   private:
    RefCountedPtr<AdsCallState> ads_calld_;
  };
};
```

src/core/ext/xds/xds_client.h
```diff
@@ -236,6 +239,19 @@ class XdsClient : public DualRefCounted<XdsClient> {
...
+   public:
+    explicit ReadDelayHandle(WeakRefCountedPtr<ChannelState> channel_state);
```
markdroth, line 245: This should hold a ref to the `AdsCallState` object, not the `ChannelState` object.

src/core/ext/xds/xds_client.cc
```diff
@@ -117,6 +119,9 @@ class XdsClient::ChannelState::AdsCallState
 
   RetryableCall<AdsCallState>* parent() const { return parent_.get(); }
   ChannelState* chand() const { return parent_->chand(); }
+  XdsTransportFactory::XdsTransport::StreamingCall* call() const {
```
markdroth, line 122: I don't think we need this.

src/core/ext/xds/xds_client.cc
```diff
@@ -261,7 +266,7 @@ class XdsClient::ChannelState::AdsCallState
         ResourceState& state = authority_state.resource_map[type_][name_.key];
         state.meta.client_status = XdsApi::ResourceMetadata::DOES_NOT_EXIST;
         ads_calld_->xds_client()->NotifyWatchersOnResourceDoesNotExist(
-            state.watchers, XdsApi::ReadDelayHandle::NoWait());
+            state.watchers, XdsClient::ReadDelayHandle::NoWait());
```
markdroth, line 269: I don't think we need the `XdsClient::` prefix, since we're already inside that class.

Same thing throughout.

src/core/ext/xds/xds_client.cc
```diff
@@ -1036,10 +1047,15 @@ void XdsClient::ChannelState::AdsCallState::OnRecvMessage(
     absl::string_view payload) {
   {
     MutexLock lock(&xds_client()->mu_);
-    if (!IsCurrentCallOnChannel()) return;
+    if (!IsCurrentCallOnChannel()) {
+      call_->StartRecvMessage();
```
markdroth, line 1051: This is extremely counter-intuitive -- why would we need to start a new read on a stream that has been shut down?  I think the fact that you needed to add this is a symptom of the ref-counting problem I pointed out elsewhere.  If you make the changes I suggested to require the caller to explicitly start each read and to take a new ref for each read, then I don't think this call will be necessary.

src/core/ext/xds/xds_client.cc
```diff
@@ -1383,6 +1398,8 @@ void XdsClient::ChannelState::LrsCallState::OnRequestSent(bool /*ok*/) {
 void XdsClient::ChannelState::LrsCallState::OnRecvMessage(
     absl::string_view payload) {
   MutexLock lock(&xds_client()->mu_);
+  auto cleanup =
+      absl::MakeCleanup([call = call_.get()]() { call->StartRecvMessage(); });
```
markdroth, line 1402: This should be moved down below line 1404.  If we are not the current call, then we don't need to start another read.

src/core/ext/filters/client_channel/lb_policy/xds/xds_cluster_resolver.cc
markdroth: Actually, sorry, I think I steered you wrong on this -- the lambda won't go out of scope until after we return from `OnEndpointChanged()`, so there's not actually any reason why we need to pass the handle down there.

In fact, I don't even think the handle needs to be plumbed into any of these `*Helper()` methods.  I think it's sufficient to capture it in the lambda that runs in the `WorkSerializer`.

src/core/ext/xds/xds_server_config_fetcher.cc
```diff
@@ -157,7 +157,9 @@ class XdsServerConfigFetcher::ListenerWatcher
       std::shared_ptr<const XdsListenerResource> listener,
       RefCountedPtr<ReadDelayHandle> read_delay_handle) override;
 
-  void OnError(absl::Status status) override;
+  void OnError(
+      absl::Status status,
+      RefCountedPtr<XdsClient::ReadDelayHandle> read_delay_handle) override;
```
markdroth, line 162: Don't need `XdsClient::` prefix due to the `using` statement above.

test/core/xds/xds_transport_fake.h
```diff
@@ -66,6 +68,8 @@ class FakeXdsTransportFactory : public XdsTransportFactory {
 
     void Orphan() override;
 
+    void StartRecvMessage() override { read_count_ += 1; }
```
markdroth, line 71: I think we probably want something a little more sophisticated here:
- We should ensure that there is only one pending read at a time.  A read is considered pending between when `StartRecvMessage()` is called and when `EventHandler::OnRecvMessage()` is invoked.
- We should never invoke `EventHandler::OnRecvMessage()` without first seeing `StartRecvMessage()`.  This means that if the test calls `SendMessageToClient()` but `XdsClient` has not yet called `StartRecvMessage()`, the message should stay queued until `XdsClient` does call `StartRecvMessage()`.
- We probably want to expose a method for the test to check whether there is a pending read, so that we can write a test that shows that `XdsClient` does not start a new read until all watcher notifications have been completed.

test/core/xds/xds_client_ads_stream_wait_test.cc
```diff
@@ -0,0 +1,93 @@
...
+#include "test/core/util/test_config.h"
+#include "test/core/xds/xds_client_test_lib.h"
+
+namespace grpc_core {
+namespace testing {
+namespace {
+
+using XdsClientNotifyWatchersDone = XdsClientTestBase;
+
+TEST_F(XdsClientNotifyWatchersDone, Basic) {
```
markdroth, line 38: I don't think we should put this test in its own file.  Instead, please just add this to the existing xds_client_test.cc file, and move the test class from xds_client_test_lib.h back into xds_client_test.cc.

test/core/xds/xds_client_ads_stream_wait_test.cc
```diff
@@ -0,0 +1,93 @@
...
+  CheckRequestNode(*request);  // Should be present on the first request.
+  // Send a response.
+  stream->SendMessageToClient(
+      ResponseBuilder(XdsFooResourceType::Get()->type_url())
+          .set_version_info("1")
+          .set_nonce("A")
+          .AddFooResource(XdsFooResource("foo1", 6))
+          .Serialize());
+  // XdsClient should have delivered the response to the watcher.
+  auto resource = watcher->WaitForNextResourceAndHandle();
```
markdroth, line 63: Here we need a way to control when the handle is returned, so that we can check that the `XdsClient` did not start a new read.  Then we can free the handle and verify that it *did* start a new read.

test/core/xds/xds_transport_grpc_test.cc
```diff
@@ -0,0 +1,210 @@
...
+
+  std::thread server_thread_;
+  Mutex mu_;
+  CondVar cond_;
+  std::string url_ ABSL_GUARDED_BY(mu_);
+  AdsService ads_service_ ABSL_GUARDED_BY(mu_);
+  std::unique_ptr<grpc::Server> server_;
+};
+
+TEST(GrpcTransportTest, DISABLED_WaitsWithAdsRead) {
```
markdroth, line 161: Is this test going to stay here?  I'm assuming it's not ready for review yet.

test/core/xds/BUILD
```diff
@@ -179,6 +193,25 @@ grpc_cc_test(
     ],
 )
 
+grpc_cc_test(
+    name = "xds_client_ads_stream_wait_test",
+    srcs = ["xds_client_ads_stream_wait_test.cc"],
+    external_deps = ["gtest"],
+    language = "C++",
+    # shard_count = 10,
```
markdroth, line 201: Please remove commented-out lines.

markdroth (commented):

test/core/xds/xds_transport_fake.h
eugeneo: Done

test/core/xds/xds_client_ads_stream_wait_test.cc
eugeneo: Will move it. Currently it just makes it easier for me.

test/core/xds/xds_client_ads_stream_wait_test.cc
eugeneo: I check it before releasing the handle:
```
  EXPECT_EQ(stream->read_count(), 1);
  resource = absl::nullopt;
  EXPECT_EQ(stream->read_count(), 2);
```

test/core/xds/xds_transport_grpc_test.cc
eugeneo: Removed (to branch on my PC)

test/core/xds/BUILD
eugeneo: Will do.

test/core/xds/xds_transport_fake.cc
```diff
@@ -120,15 +124,45 @@ void FakeXdsTransportFactory::FakeStreamingCall::CompleteSendMessageFromClient(
...
-    event_handler = event_handler_->Ref();
+  ReleasableMutexLock lock(&mu_);
+  to_client_messages_.emplace_back(payload);
+  lock.Release();
+  DispatchPendingMessageToClient();
+}
+
+void FakeXdsTransportFactory::FakeStreamingCall::
+    DispatchPendingMessageToClient() {
+  ReleasableMutexLock lock(&mu_);
```
markdroth, line 155: Similar to above, please use an inner scope and `MutexLock` instead of `ReleaseableMutexLock`.

test/core/xds/xds_client_ads_stream_wait_test.cc
```diff
@@ -59,21 +58,37 @@ TEST_F(XdsClientNotifyWatchersDone, Basic) {
           .set_nonce("A")
           .AddFooResource(XdsFooResource("foo1", 6))
           .Serialize());
+  stream->SendMessageToClient(
```
markdroth, line 61: Please add a comment indicating that this message won't actually get read by the `XdsClient` until we reset the handle from the watcher notifications.

test/core/xds/xds_transport_fake.h
```diff
@@ -117,6 +127,9 @@ class FakeXdsTransportFactory : public XdsTransportFactory {
     std::deque<std::string> from_client_messages_ ABSL_GUARDED_BY(&mu_);
     bool status_sent_ ABSL_GUARDED_BY(&mu_) = false;
     bool orphaned_ ABSL_GUARDED_BY(&mu_) = false;
+    size_t read_count_ ABSL_GUARDED_BY(&mu_) = 0;
```
markdroth, line 130: Suggest calling this `reads_started_`, to make it clear that we're counting reads started instead of reads completed.

test/core/xds/xds_client_ads_stream_wait_test.cc
```diff
@@ -0,0 +1,108 @@
...
+namespace grpc_core {
+namespace testing {
+namespace {
+
+using XdsClientNotifyWatchersDone = XdsClientTestBase;
+
+TEST_F(XdsClientNotifyWatchersDone, Basic) {
+  InitXdsClient();
+  // Start a watch for "foo1".
+  auto watcher = StartFooWatch("foo1");
```
markdroth, line 40: I suggest starting watchers for two different resources and having them both in the same response, so that we can verify that we don't start the next read until *both* watchers free their read delay handles.

test/core/xds/xds_client_ads_stream_wait_test.cc
```diff
@@ -59,21 +58,37 @@ TEST_F(XdsClientNotifyWatchersDone, Basic) {
...
                /*version_info=*/"1", /*response_nonce=*/"A",
                /*error_detail=*/absl::OkStatus(),
                /*resource_names=*/{"foo1"});
-  EXPECT_EQ(stream->read_count(), 0);
-  resource->second.reset();
   EXPECT_EQ(stream->read_count(), 1);
+  resource = absl::nullopt;
+  EXPECT_EQ(stream->read_count(), 2);
+  resource = watcher->WaitForNextResourceAndHandle();
+  ASSERT_NE(resource, absl::nullopt);
```
markdroth, line 83: After this, please verify that we actually get the expected value for the resource.

test/core/xds/xds_client_ads_stream_wait_test.cc
markdroth: This is making it harder to review the PR, because I can't easily see what you've changed in the test class.

markdroth (commented):

test/core/xds/xds_client_ads_stream_wait_test.cc
eugeneo: Done

test/core/xds/xds_client_ads_stream_wait_test.cc
```diff
@@ -59,21 +58,37 @@ TEST_F(XdsClientNotifyWatchersDone, Basic) {
...
   request = WaitForRequest(stream.get());
   ASSERT_TRUE(request.has_value());
   CheckRequest(*request, XdsFooResourceType::Get()->type_url(),
                /*version_info=*/"1", /*response_nonce=*/"A",
                /*error_detail=*/absl::OkStatus(),
                /*resource_names=*/{"foo1"});
-  EXPECT_EQ(stream->read_count(), 0);
-  resource->second.reset();
   EXPECT_EQ(stream->read_count(), 1);
+  resource = absl::nullopt;
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/xds/xds_client_ads_stream_wait_test.cc
eugeneo: Done

src/core/ext/xds/xds_client.cc
```diff
@@ -702,6 +715,8 @@ absl::Status XdsClient::ChannelState::AdsCallState::AdsResponseParser::
         fields.type_url.c_str(), fields.version.c_str(), fields.nonce.c_str(),
         fields.num_resources);
   }
+  result_.read_delay_handle =
```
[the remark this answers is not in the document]
eugeneo: Done

eugeneo (commented):
Thank you for the review. I had to merge to fix merge conflicts.

Please take another look.

test/core/xds/xds_client_test.cc
```diff
@@ -252,28 +257,37 @@ class XdsClientTest : public ::testing::Test {
...
 
+      std::shared_ptr<const ResourceStruct> WaitForNextResource(
+          absl::Duration timeout = absl::Seconds(1),
+          SourceLocation location = SourceLocation()) {
+        auto resource_and_handle =
+            WaitForNextResourceAndHandle(timeout, location);
+        if (!resource_and_handle.has_value()) {
+          return nullptr;
+        }
+        return resource_and_handle->resource;
```
markdroth, line 288: `std::move()`

test/core/xds/xds_client_test.cc
```diff
@@ -2695,6 +2719,85 @@ TEST_F(XdsClientTest, FederationChannelFailureReportedToWatchers) {
   EXPECT_TRUE(stream2->Orphaned());
 }
 
+TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
+  InitXdsClient();
+  // Start watches for "foo1" and "foo2".
+  auto watcher1 = StartFooWatch("foo1");
+  auto watcher2 = StartFooWatch("foo2");
+  // Watchers should initially not see any resource reported.
```
markdroth, line 2727: This isn't useful to check here.  This test is about the read-delay handles, not about the basic watcher API functionality.

test/core/xds/xds_client_test.cc
```diff
@@ -2695,6 +2719,85 @@ TEST_F(XdsClientTest, FederationChannelFailureReportedToWatchers) {
   EXPECT_TRUE(stream2->Orphaned());
 }
 
+TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
+  InitXdsClient();
+  // Start watches for "foo1" and "foo2".
+  auto watcher1 = StartFooWatch("foo1");
```
markdroth, line 2725: Please check the first request seen by the server after starting the first watcher but before starting the second watcher.  In other words, instead of starting both watches and then checking both requests, please start a watch, check the request, then start the second watch, then start the second request.

test/core/xds/xds_client_test.cc
```diff
@@ -2695,6 +2719,85 @@ TEST_F(XdsClientTest, FederationChannelFailureReportedToWatchers) {
...
+  auto watcher2 = StartFooWatch("foo2");
+  // Watchers should initially not see any resource reported.
+  EXPECT_FALSE(watcher1->HasEvent());
+  EXPECT_FALSE(watcher2->HasEvent());
+  // XdsClient should have created an ADS stream.
+  auto stream = WaitForAdsStream();
+  ASSERT_TRUE(stream != nullptr);
+  // XdsClient should have sent a subscription request on the ADS stream.
+  auto request = WaitForRequest(stream.get());
+  ASSERT_TRUE(request.has_value());
```
markdroth, line 2735: Instead of just asserting that we got a request, please check the contents of the request via `CheckRequest()`.

Same thing for the next request.

test/core/xds/xds_client_test.cc
```diff
@@ -2695,6 +2719,85 @@ TEST_F(XdsClientTest, FederationChannelFailureReportedToWatchers) {
...
+  EXPECT_EQ(resource1->resource->value, 8);
+  EXPECT_EQ(watcher2->WaitForNextResourceAndHandle(), absl::nullopt);
+  // XdsClient should have sent an ACK message to the xDS server.
+  request = WaitForRequest(stream.get());
+  ASSERT_TRUE(request.has_value());
+  CheckRequest(*request, XdsFooResourceType::Get()->type_url(),
+               /*version_info=*/"2", /*response_nonce=*/"B",
+               /*error_detail=*/absl::OkStatus(),
+               /*resource_names=*/{"foo1", "foo2"});
+  resource1->read_delay_handle.reset();
```
markdroth, line 2787: Please check that `stream->reads_started()` is still 2 right before this.

test/core/xds/xds_client_fuzzer.cc
```diff
@@ -147,20 +147,26 @@ class Fuzzer {
         : resource_name_(std::move(resource_name)) {}
 
     void OnResourceChanged(
```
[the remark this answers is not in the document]
markdroth: Did you check whether this is feasible?  If it's not, that's fine, but it would be good to do this if there is a way to do so.

test/core/xds/xds_transport_fake.cc
```diff
@@ -120,15 +124,45 @@ void FakeXdsTransportFactory::FakeStreamingCall::CompleteSendMessageFromClient(
   CompleteSendMessageFromClientLocked(ok);
 }
 
+void FakeXdsTransportFactory::FakeStreamingCall::StartRecvMessage() {
+  absl::optional<std::string> pending;
+  ReleasableMutexLock lock(&mu_);
+  ASSERT_FALSE(read_pending_);
```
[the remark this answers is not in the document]
markdroth: It looks like you removed the assert and replaced it with a log message, which doesn't actually accomplish what I was asking for here.  There are two properties we need here:
1. The test will fail if the expectation is violated.  The log message will not actually cause the test to fail, so no one will notice this problem unless they happen to be looking at the test log for some other reason.
2. The test will fail in a reasonable amount of time rather than hanging indefinitely and timing out.  If the XdsClient does start a second read before the first one completes, then returning here means that we will never complete that second read, which will very likely cause the test to hang.

Please do the following:
- Replace `bool read_pending_` with `size_t num_pending_reads_`.
- Change this code to trigger a test failure (see below for options) if `num_pending_reads_` is non-zero, but do *not* return from this function -- instead, increment `num_pending_reads_` and continue.
- Change `MaybeDeliverMessageToClient()` to be a loop that keeps iterating while `num_pending_reads_ > 0` and decrements the counter each iteration.

There are two options for how to trigger a test failure:
1. Call `EXPECT_EQ(num_pending_reads_, 0)`.  This is the easiest option, if it works right in xds_client_fuzzer, where we're not using gtest.
2. If the `EXPECT` approach doesn't work with xds_client_fuzzer, then we can use a dependency injection approach.  We can add a method to `FakeXdsTransportFactory` to set a callback to be invoked when there are multiple pending reads.  Then each of xds_client_test and xds_client_fuzzer can set that callback to do something appropriate.

markdroth (commented):
This is getting closer!

test/core/xds/xds_client_fuzzer.cc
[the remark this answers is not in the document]
eugeneo: Following your other comment.

test/core/xds/xds_transport_fake.cc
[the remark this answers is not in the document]
eugeneo: > Call EXPECT_EQ(num_pending_reads_, 0).

Requires dependency on gtest, seems like fuzzer does not depend on it.

> If the EXPECT approach doesn't work with xds_client_fuzzer, then we can use a dependency injection approach.

I implemented this approach. Please take a look.

test/core/xds/xds_transport_fake.h
```diff
@@ -235,6 +245,7 @@ class FakeXdsTransportFactory : public XdsTransportFactory {
         ABSL_GUARDED_BY(&mu_);
     std::map<std::string /*method*/, RefCountedPtr<FakeStreamingCall>>
         active_calls_ ABSL_GUARDED_BY(&mu_);
+    std::function<void()> too_many_pending_reads_callback_;
```
markdroth, line 248: I don't think we need to store a copy of the callback in the `FakeXdsTransport` instance.  `FakeXdsTransport` already holds a ref to `FakeXdsTransportFactory`, so it can instead just access `factory_->too_many_pending_reads_callback_`.

test/core/xds/xds_transport_fake.h
```diff
@@ -128,11 +131,15 @@ class FakeXdsTransportFactory : public XdsTransportFactory {
     bool status_sent_ ABSL_GUARDED_BY(&mu_) = false;
     bool orphaned_ ABSL_GUARDED_BY(&mu_) = false;
     size_t reads_started_ ABSL_GUARDED_BY(&mu_) = 0;
-    bool read_pending_ ABSL_GUARDED_BY(&mu_) = false;
+    size_t num_pending_reads_ ABSL_GUARDED_BY(&mu_) = 0;
     std::deque<std::string> to_client_messages_ ABSL_GUARDED_BY(&mu_);
+    std::function<void()> too_many_pending_reads_callback_;
```
markdroth, line 136: Similar to my comment below, I don't think we need to store a copy of the callback in `FakeStreamingCall`.  Instead, we can access it via `transport_->factory_->too_many_pending_reads_callback_`.

test/core/xds/xds_client_fuzzer.cc
```diff
@@ -57,7 +57,9 @@ class Fuzzer {
       // Leave xds_client_ unset, so Act() will be a no-op.
       return;
     }
-    auto transport_factory = MakeOrphanable<FakeXdsTransportFactory>();
+    auto transport_factory = MakeOrphanable<FakeXdsTransportFactory>([]() {
+      gpr_assertion_failed(__FILE__, __LINE__, "Multiple concurrent reads");
```
markdroth, line 61: Please use `GPR_ASSERT()` instead of expanding the macro.

test/core/xds/xds_client_test.cc
```diff
@@ -2771,7 +2777,6 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
   resource1->read_delay_handle.reset();
   EXPECT_EQ(stream->reads_started(), 1);
   resource2->read_delay_handle.reset();
-  EXPECT_EQ(stream->reads_started(), 2);
```
markdroth: Please restore this line.  We want to check that another read has been started at this point, and then check it again later to make sure the next read hasn't been started yet before we trigger it.

markdroth (commented):
This looks great!  Just a few remaining minor comments.

test/core/xds/xds_client_fuzzer.cc
eugeneo: Done. I introduced a fake variable to make failure more descriptive

test/core/xds/xds_transport_fake.h
eugeneo: Done

test/core/xds/xds_transport_fake.h
eugeneo: Done. Except I added an accessor for factory as the field is private.

test/core/xds/xds_client_test.cc
eugeneo: Done

eugeneo (commented):
Thank you for the review

test/core/xds/xds_client_fuzzer.cc
```diff
@@ -58,7 +58,8 @@ class Fuzzer {
       return;
     }
     auto transport_factory = MakeOrphanable<FakeXdsTransportFactory>([]() {
-      gpr_assertion_failed(__FILE__, __LINE__, "Multiple concurrent reads");
+      bool has_pending_read = false;
```
markdroth, line 61: I think this can just be:

```
Crash("Multiple concurrent reads");
```

test/core/xds/xds_transport_fake.h
```diff
@@ -212,6 +205,8 @@ class FakeXdsTransportFactory : public XdsTransportFactory {
 
     void RemoveStream(const char* method, FakeStreamingCall* call);
 
+    RefCountedPtr<FakeXdsTransportFactory> factory() const { return factory_; }
```
markdroth, line 208: This should return a raw pointer, not a `RefCountedPtr<>`.  Otherwise, we're needlessly taking a new ref every time we call this.

test/core/xds/xds_transport_fake.cc
```diff
@@ -120,15 +126,50 @@ void FakeXdsTransportFactory::FakeStreamingCall::CompleteSendMessageFromClient(
   CompleteSendMessageFromClientLocked(ok);
 }
 
+void FakeXdsTransportFactory::FakeStreamingCall::StartRecvMessage() {
+  absl::optional<std::string> pending;
```
markdroth, line 130: This variable appears to be unused.  I think it can be removed.

markdroth (approved):
Looks good!  Feel free to merge after addressing the remaining comments.

test/core/xds/xds_client_fuzzer.cc
eugeneo: Done

test/core/xds/xds_transport_fake.h
eugeneo: Done

test/core/xds/xds_transport_fake.cc
eugeneo: Done

eugeneo (commented):
Thank you. I applied the suggestions and merged master.

markdroth:
I don't think your merge worked right.

PR #35002, opened 2023-11-16
[Python Observability] Allow create call with registered method

XuanWang-Amos:
Based on [OpenTelemetry Metrics gRFC](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md#opentelemetry-metrics), we should recored unregistered RPC method name as `other`, this PR adds the ability to pass register method information when creating a call.

We'll consider calls created using generated stubs as registered, note that this won't prevent user from setting `registered_method=True` when creating calls manually.

This is also enabled for simple stub flow but **NOT enabled for AsyncIO**, we'll add that later when start working on AsyncIO Observability.


src/compiler/python_generator.cc
```diff
@@ -657,7 +658,7 @@ bool PrivateGenerator::PrintServiceClass(
           out->Print("options, channel_credentials,\n");
           out->Print(
               "insecure, call_credentials, compression, wait_for_ready, "
-              "timeout, metadata)\n");
```
gnossen, line 660: Please update [this file](https://github.com/grpc/grpc/blob/master/examples/python/helloworld/helloworld_pb2_grpc.py) as a visual indicator of what these changes look like. [The `Makefile`](https://github.com/grpc/grpc/blob/master/examples/python/Makefile) will work as long as you've installed a dev version of `grpcio-tools`

src/python/grpcio/grpc/_channel.py
```diff
@@ -1224,12 +1230,14 @@ def __init__(
         method: bytes,
         request_serializer: SerializingFunction,
         response_deserializer: DeserializingFunction,
+        registered_method: bool,
```
gnossen, line 1233: This parameter name should start with an underscore to indicate that it's for internal usage only. We probably also want to add a docstring mentioning its purpose.

src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi
```diff
@@ -233,18 +234,31 @@ cdef void _call(
...
+      if registered_method:
+        registered_call_handle = grpc_channel_register_call(
+          channel_state.c_channel, <const char *>method,
+          host_ptr, NULL)
+        call_state.c_call = grpc_channel_create_registered_call(
+            channel_state.c_channel, NULL, flags,
+            c_completion_queue, registered_call_handle,
+            _timespec_from_time(deadline), NULL)
+      else:
+        call_state.c_call = grpc_channel_create_call(
```
gnossen, line 258: I'm worried that with this change, the vast majority of our testing will go down this branch, but the vast majority of our users will be going down the other branch. Can we either parameterize our existing tests on method registration or just manually set `registered` in most of them?

src/python/grpcio/grpc/_channel.py
```diff
@@ -1533,22 +1551,25 @@ class _StreamStreamMultiCallable(grpc.StreamStreamMultiCallable):
...
     _context: Any
+    _registered_method: bool
 
     # pylint: disable=too-many-arguments
     def __init__(
         self,
         channel: cygrpc.Channel,
         managed_call: IntegratedCallFactory,
         method: bytes,
-        request_serializer: Optional[SerializingFunction] = None,
```
gnossen, line 1543: Nice catch!

Note to self: This is not a backward-incompatible change because the _constructor_ of this class is a private API. It's just the other methods that constitute public API because instances of this class are returned from `Channel.stream_stream`

src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi
```diff
@@ -233,18 +234,31 @@ cdef void _call(
...
       else:
         host_slice = _slice_from_bytes(host)
         host_slice_ptr = &host_slice
-      call_state.c_call = grpc_channel_create_call(
-          channel_state.c_channel, NULL, flags,
-          c_completion_queue, method_slice, host_slice_ptr,
-          _timespec_from_time(deadline), NULL)
+        host_ptr = <const char *>host
+      if registered_method:
+        registered_call_handle = grpc_channel_register_call(
```
gnossen, line 250: We'll want to get clarification from @ctiller on this ([reference](https://github.com/grpc/grpc/pull/1261)), but I don't think the intent with this function is to call it on a per-call basis, but instead, to call it on a per *channel* basis, with the idea being that it saves on per-call string operations on the method/service.

However, I know that the primary value we intend to get from this is registration for the purposes of observability, not a cost-savings on string operations. And it would take some work so that we plumb through a pre-registration handle rather than a bool. So it might be okay to leave things as they are, but something to think about.

If you did decide to do this restructure, I think it would look like this:

- Add a new gRPC Python channel method called something like `_create_registered_method` that returns an object that's basically a Python int containing the `void*` returned by a call to `grpc_channel_register_call`
- Call `_created_registered_method` once for each method in the `Stub` class constructor and pass the returned Python integer through to  `unary_unary`, `unary_stream`, etc.
- Plumb the handle down here to `_call` if supplied
-  if the handle is supplied, call `grpc_channel_create_registered_call` with it. If not, call `grpc_channel_create_call`.

If you don't do the restructure, you _should_ make sure that repeatedly calling `grpc_channel_register_call` many times repeatedly is not going to cause issues for us, for example unbounded memory growth.

src/python/grpcio/grpc/_interceptor.py
```diff
@@ -684,57 +684,73 @@ def subscribe(
     def unsubscribe(self, callback: Callable):
         self._channel.unsubscribe(callback)
 
+    # pylint: disable=arguments-differ
     def unary_unary(
         self,
         method: str,
         request_serializer: Optional[SerializingFunction] = None,
         response_deserializer: Optional[DeserializingFunction] = None,
+        registered_method: Optional[bool] = False,
```
gnossen, line 693: Please add a test to ensure that `registered_method` gets plumbed through interceptors properly.

src/python/grpcio/grpc/aio/_channel.py
```diff
@@ -478,11 +478,13 @@ async def channel_ready(self) -> None:
             await self.wait_for_state_change(state)
             state = self.get_state(try_to_connect=True)
 
+    # pylint: disable=arguments-differ,unused-argument
     def unary_unary(
         self,
         method: str,
         request_serializer: Optional[SerializingFunction] = None,
         response_deserializer: Optional[DeserializingFunction] = None,
+        registered_method: Optional[bool] = False,
```
gnossen, line 487: Please add a TODO noting that we should support `registered_method` under asyncio and a comment explaining why this is currently a no-op.

src/compiler/python_generator.cc
```diff
@@ -657,7 +658,7 @@ bool PrivateGenerator::PrintServiceClass(
...
+              "timeout, metadata, _registered_method=True)\n");
```
gnossen, line 661: Looks like this pushes the generated line beyond 80 characters. Can you reformat the generated line so we stay below 80?

XuanWang-Amos:
@gnossen PTALA.

src/python/grpcio/grpc/aio/_channel.py
```diff
@@ -478,11 +478,20 @@ async def channel_ready(self) -> None:
             await self.wait_for_state_change(state)
             state = self.get_state(try_to_connect=True)
 
+    # TODO(xuanwn): Implement this method after we have
+    # observability for Asyncio.
+    def _get_registered_call_handle(self, method: str) -> int:
+        pass
```
gnossen, line 484: Shouldn't this fail type checking?

src/python/grpcio/grpc/_cython/_cygrpc/channel.pyx.pxi
```diff
@@ -542,3 +551,13 @@ cdef class Channel:
 
   def close_on_fork(self, code, details):
     _close(self, code, details, True)
+
+  def register_method(self, method, host):
```
gnossen, line 555: What does `host` represent here? Is this the remote host, i.e. the target string? If we never pass a value for `host`, will we have issues if we have two concurrent channels pointed at two different target strings but using the same methods? If so, should we be setting a value every time?

If not, and it's safe to unconditionally pass `NULL` here, can we just get rid of the `host` parameter to this method and not have to worry about it?

examples/python/helloworld/helloworld_pb2_grpc.py
```diff
@@ -19,7 +19,20 @@ def __init__(self, channel):
...
                 request_serializer=helloworld__pb2.HelloRequest.SerializeToString,
                 response_deserializer=helloworld__pb2.HelloReply.FromString,
-                )
+                _registered_call_handle=channel._get_registered_call_handle(
+                    '/helloworld.Greeter/SayHello'))
+        self.SayHelloStreamReply = channel.unary_stream(
+                '/helloworld.Greeter/SayHelloStreamReply',
+                request_serializer=helloworld__pb2.HelloRequest.SerializeToString,
+                response_deserializer=helloworld__pb2.HelloReply.FromString,
+                _registered_call_handle=channel._get_registered_call_handle(
```
gnossen, line 28: Instead of duplicating the method string and requiring creating a semi-private method on `Channel`, could we instead just have a kwarg called `_registered` which takes a `bool` and has a default of `False`? The smaller change to the API surface seems much more appealing.

This is similar to what you've done for the simple stubs versions.

src/python/grpcio/grpc/_simple_stubs.py
```diff
@@ -178,33 +192,51 @@ def get_channel(
             channel_data = self._mapping.get(key, None)
             if channel_data is not None:
                 channel = channel_data[0]
+                registered_method_handles = channel_data[2]
+                # Register a new call handle if we're calling a registered method for an
+                # existing channel and this method is not registered.
+                if (
+                    _registered_method
+                    and method not in registered_method_handles.keys()
```
gnossen, line 200: It seems odd that we're doing this book-keeping both at the `ChannelCache` level and at the `Channel` level. Why can't this block of added code just become

```
call_handle = channel._get_registered_call_handle(method)
```

src/python/grpcio/grpc/_simple_stubs.py
```diff
@@ -219,6 +251,7 @@ def unary_unary(
     wait_for_ready: Optional[bool] = None,
     timeout: Optional[float] = _DEFAULT_TIMEOUT,
     metadata: Optional[Sequence[Tuple[str, Union[str, bytes]]]] = None,
+    _registered_method: Optional[bool] = False,
```
gnossen, line 254: Why `Optional[bool]` instead of just `bool`?

src/python/grpcio/grpc/_simple_stubs.py
```diff
@@ -268,15 +301,23 @@ def unary_unary(
         unset, defaults to 60 seconds. Supply a value of None to indicate that
         no timeout should be enforced.
       metadata: Optional metadata to send to the server.
+      _registered_method: INTERNAL USE ONLY. Whether this RPC is made for a
```
gnossen, line 304: Let's not document this in the public-facing docstrings. If we want to keep the comment, let's add it a few lines below.

src/python/grpcio/grpc/_simple_stubs.py
gnossen: Ah, yes. You're right. Good catch.

examples/python/helloworld/helloworld_pb2_grpc.py
```diff
@@ -67,4 +99,41 @@ def SayHello(request,
...
+            target,
+            options=(),
+            channel_credentials=None,
+            call_credentials=None,
+            insecure=False,
+            compression=None,
+            wait_for_ready=None,
+            timeout=None,
+            metadata=None):
+        return grpc.experimental.unary_stream(request, target, '/helloworld.Greeter/SayHelloStreamReply',
```
gnossen, line 116: Can we put each argument on its own line? While we can't _guarantee_ that the lines won't exceed 80 characters because they contain user-provided strings, there's a much lower chance of this happening if each argument has its own line.

We don't provide a guarantee of any particular formatting protocol with our generated code, but we do have plenty of people who complain when we break common rules, such as 80 characters in length. The result is that people often either exclude `*pb2*.py` from their formatter's configuration or they manually format their files after generation.

src/python/grpcio/grpc/_channel.py
```diff
@@ -1695,6 +1719,8 @@ def create(
           event_handler: A behavior to call to handle the events resultant from
             the operations on the call.
           context: Context object for distributed tracing.
+          _registered_call_handle: An int representing the call handle of the
```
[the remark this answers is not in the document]
gnossen: Ah, you're right. This docstring does _not_ end up in the public docs. The method name must have spilled off of my screen. We actually used to have docstrings _only_ for the public APIs. My mistake. 

It should be fine to have `_registered_call_handle` documented here.

gnossen (approved):

PR #35008, opened 2023-11-16
[core] Ensure ChannelArgs::SetObject only allows conforming shared_ptr<T> classes

drfloob:
ChannelArgs shared_ptr only supports types that extend `enable_shared_from_this`. `args.SetObject<shared_ptr<X>>(x)` with a non-comforming type X will now fail with something like:

```
./src/core/lib/channel/channel_args.h:453:12: error: no matching member function for call to 'Set'
    return Set(ChannelArgNameTraits<T>::ChannelArgName(), std::move(p));
           ^~~
test/core/channel/channel_args_test.cc:352:32: note: in instantiation of function template specialization 'grpc_core::ChannelArgs::SetObject<X>' requested here
  grpc_core::ChannelArgs b = a.SetObject(x);
                               ^
..
```

### DIFF
--- a/src/core/lib/channel/channel_args.h
+++ b/src/core/lib/channel/channel_args.h
@@ -125,6 +125,16 @@ struct ChannelArgTypeTraits<
   };
 };
 
+// Define a check for shared_ptr supported types, which must extend
+// enable_shared_from_this.
+template <typename T>
+struct SupportedSharedPtrType
+    : std::integral_constant<
+          bool, std::is_base_of<std::enable_shared_from_this<T>, T>::value> {};
+template <>
+struct SupportedSharedPtrType<grpc_event_engine::experimental::EventEngine>
+    : std::true_type {};
+
 // Specialization for shared_ptr
 // Incurs an allocation because shared_ptr.release is not a thing.
 template <typename T>
@@ -132,6 +142,10 @@ struct is_shared_ptr : std::false_type {};
 template <typename T>
 struct is_shared_ptr<std::shared_ptr<T>> : std::true_type {};
 template <typename T>
+// struct ChannelArgTypeTraits<
+//     T, absl::enable_if_t<is_shared_ptr<T>::value &&
+//     SupportedSharedPtrType<T>::value,
+//                          void>> {
 struct ChannelArgTypeTraits<T,
                             absl::enable_if_t<is_shared_ptr<T>::value, void>> {
   static void* TakeUnownedPointer(T* p) { return p; }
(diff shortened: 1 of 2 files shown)


src/core/lib/channel/channel_args.h
```diff
@@ -386,9 +395,10 @@ class ChannelArgs {
   }
   template <typename T>
   GRPC_MUST_USE_RESULT absl::enable_if_t<
-      std::is_same<
-          const grpc_arg_pointer_vtable*,
-          decltype(ChannelArgTypeTraits<std::shared_ptr<T>>::VTable())>::value,
+      std::is_same<const grpc_arg_pointer_vtable*,
```
ctiller, line 398: Not sure if this is better, but worth a try?

What happens if we write:

```
  GRPC_MUST_USE_RESULT absl::enable_if_t<
      std::is_same<
          const grpc_arg_pointer_vtable*,
          decltype(ChannelArgTypeTraits<std::shared_ptr<T>>::VTable())>::value,
      ChannelArgs>
  Set(absl::string_view name, std::shared_ptr<T> value) const {
     static_assert(SupportedSharedPtrType<T>::value);
```

that is - removing it from the enable_if, and hard failing the compilation if it isn't a supported type

we lose the ability to overload more supported types (seems we don't need that?)
we gain a pretty clear error message though?

ctiller (approved):

src/core/lib/channel/channel_args.h
drfloob: I'm not sure what you mean by losing the ability to overload more types, we can further specialize `SupportedSharedPtrType` right?

The error is much better, I like this approach.

```
In file included from test/core/channel/channel_args_test.cc:19:
./src/core/lib/channel/channel_args.h:399:5: error: static_assert failed due to requirement 'SupportedSharedPtrType<X>::value' "Type T must extend std::enable_shared_from_this to be added into ChannelArgs as a shared_ptr<T>"
```

src/core/lib/channel/channel_args.h
```diff
@@ -125,13 +125,27 @@ struct ChannelArgTypeTraits<
...
+    : std::true_type {};
+
 // Specialization for shared_ptr
 // Incurs an allocation because shared_ptr.release is not a thing.
 template <typename T>
 struct is_shared_ptr : std::false_type {};
 template <typename T>
 struct is_shared_ptr<std::shared_ptr<T>> : std::true_type {};
 template <typename T>
+// struct ChannelArgTypeTraits<
```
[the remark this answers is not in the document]
drfloob: Deleted dead code. Thanks.

ctiller (approved):

PR #35042, opened 2023-11-20
[ObjC] add privacy manifests to grpc podspec

HannahShiSFB:
Not sure if we need to do the same for all the podspec files.

Will create a separate PR for swift package.


sampajano (approved):
Thanks a lot for the change!! 😃

sampajano:
@HannahShiSFB There is a "Distribution Tests C++ Windows DLL — Kokoro build finished" failure on CI. I'm guessing it's not related?

Could you sync and upload again maybe?

sampajano:
@ctiller Hi Craig! I wonder if i'm expected to see an auto-generated internal CL for this change? I'm not able to find that just yet. Am we missing someting? Thanks! 😃

PR #35043, opened 2023-11-20
[Security - Documentation] Mark crl_directory as deprecated

gtcooke94:


markdroth (approved):

markdroth:
Is there also an exposed method in the C++ API that we need to deprecate as well?

Issue #34872, opened 2023-11-03
The C++ xDS interop client provides empty metadata for each request in `LoadBalancerStatsResponse` if no metadata was requested


murgatroid99:
This behavior can be seen in [this log](https://fusion.corp.googleusercontent.com/file/autoview/googlefile%3A%2Fblobstore%2Fprod%2Fguitar%2Fguitar-logs%2Fcloudnet-sdn-staging-team%2Fa60a7fdd-4ccf-3340-aca3-7f2191750d46%2F7b03984a-a3c9-4d6d-abde-b99e958bb8e6%2F0%2Faa5e3333-faaa-4ce9-af37-b8956758e737%2F__cloud_cluster_testing_e2e_test_network_proxyless_grpc_psm_interop_legacy_xds_v3_cpp_v1.59.x_test%2F3bc79a46-e480-40c4-bc87-4a9f46c95ef3%2Fundeclared_outputs.zip/archiveEntry/00-test-results/circuit_breaking/00-test.log?invocationId=449a80b3-e2c7-46a3-9c2f-04e9e096d674&actionName=invocations%2F449a80b3-e2c7-46a3-9c2f-04e9e096d674%2Ftargets%2F%252F%252Fcloud%252Fcluster%252Ftesting%252Fe2e_test%252Fnetwork%252Fproxyless_grpc%253Apsm_interop_legacy_xds_v3_cpp_v1.59.x_test%2FconfiguredTargets%2Fdefault%2Factions%2Ftest&contentType=ansi) of the `circuit_breaking` test running in the legacy interop framework, which does not set `metadata_keys` at all in the `LoadBalancerStatsRequest`.

Note: in protobuf, an unset repeated field is equivalent to a repeated field with no entries.

PR #35045, opened 2023-11-20
[test] Do not return empty values if there's no metadata

eugeneo:
Changes C++ Xds server to only return metadata when there is at least one non-empty entry.

Fixes: #34872

### DIFF
--- a/test/cpp/interop/xds_stats_watcher.cc
+++ b/test/cpp/interop/xds_stats_watcher.cc
@@ -56,10 +56,8 @@ bool HasNonEmptyMetadata(
         metadata_by_peer) {
   for (const auto& entry : metadata_by_peer) {
     for (const auto& rpc_metadata : entry.second.rpc_metadata()) {
-      for (const auto& metadata : rpc_metadata.metadata()) {
-        if (!metadata.key().empty() || !metadata.value().empty()) {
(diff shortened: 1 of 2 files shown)


test/cpp/interop/xds_stats_watcher.cc
```diff
@@ -113,8 +126,12 @@ LoadBalancerStatsResponse XdsStatsWatcher::WaitForRpcStatsResponse(
                [this] { return rpcs_needed_ == 0; });
   response.mutable_rpcs_by_peer()->insert(rpcs_by_peer_.begin(),
                                           rpcs_by_peer_.end());
-  response.mutable_metadatas_by_peer()->insert(metadata_by_peer_.begin(),
-                                               metadata_by_peer_.end());
+  // Return metadata if at least one RPC had relevant metadata. Note that empty
+  // entries would be returned for RCPs with no relevant metadata in this case.
```
yijiem, line 130: s/RCPs/RPCs :)

yijiem (approved):

test/cpp/interop/xds_stats_watcher.cc
eugeneo: Done

PR #35064, opened 2023-11-21
[tcp] Defend against connect errno being overwritten.

benjaminp:
`grpc_tcp_client_create_from_prepared_fd` distinguishes "in-progress" `connect(2)` errors from fatal errors. However, it does a bunch of external calls between calling `connect(2`) and checking the `errno`. These calls may not preserve `errno`.

This change parallels defensive `errno` saving pattern in the event_engine.

### DIFF
--- a/src/core/lib/iomgr/tcp_client_posix.cc
+++ b/src/core/lib/iomgr/tcp_client_posix.cc
@@ -336,6 +336,7 @@ int64_t grpc_tcp_client_create_from_prepared_fd(
     err = connect(fd, reinterpret_cast<const grpc_sockaddr*>(addr->addr),
                   addr->len);
   } while (err < 0 && errno == EINTR);
+  int connect_errno = (err < 0) ? errno : 0;
 
   auto addr_uri = grpc_sockaddr_to_uri(addr);
   if (!addr_uri.ok()) {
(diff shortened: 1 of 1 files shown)


yashykt (approved):

src/core/lib/iomgr/tcp_client_posix.cc
```diff
@@ -336,6 +336,7 @@ int64_t grpc_tcp_client_create_from_prepared_fd(
     err = connect(fd, reinterpret_cast<const grpc_sockaddr*>(addr->addr),
                   addr->len);
   } while (err < 0 && errno == EINTR);
+  int saved_errno = (err < 0) ? errno : 0;
```
drfloob, line 339: nit: I suggest calling this `connect_errno`.

drfloob (commented):
Thanks for the fix! I left one bit of feedback.

This logic is already handled correctly in the EventEngine fork, so this PR should be sufficient.

src/core/lib/iomgr/tcp_client_posix.cc
benjaminp: I agree that's a better name. I used `saved_errno` to parallel the same logic in the event engine: https://github.com/grpc/grpc/blob/85cede5967eefd92703890b236939fa845d3801b/src/core/lib/event_engine/posix_engine/posix_engine.cc#L253-L262

drfloob: I noticed ... I'd love to see both variable names changed :)

benjaminp: :)

I've now changed this one. I could send another PR for event_engine. (I think it would also be worthwhile in event_engine to fix a theoretical bug by adding the `(err < 0) ? errno : 0` condition I've written here; `errno` is undefined if the call succeeds.)

drfloob (approved):

src/core/lib/iomgr/tcp_client_posix.cc
drfloob: I'd gladly review another fix PR! Thanks for this, I'll move forward with testing & landing it.

PR #35066, opened 2023-11-21
[benchmark] Exclude C++ from OSS benchmarks CI.

paulosjca:
QPS worker build is currently failing.


paulosjca:
Issue is fixed, PR is no longer required.

PR #35089, opened 2023-11-23
[ruby] Fix seg faults in tests

stanhu:
This pull request fixes two items:

1.  Improve type checking for SSL private key and certs. Previously if you passed the wrong types in `GRPC::Core::ServerCredentials.new` for `private_key` and `cert_chain` the Ruby C extension would attempt to case the values to string pointers and seg fault. This commit verifies that they are strings and raises an exception if they are not.

2. Fix duplicate definition of `load_test_certs`. `spec/generic/client_stub_spec.rb` and `spec/server_spec.rb` had slightly different definitions of `load_test_certs`. The conflicting definitions caused `spec/generic/client_stub_spec.rb` to seg fault since `private_key` was the incorrect type.


stanhu:
@apolcyn Could you review this? I was puzzled why running `bundle exec rspec` inside `src/ruby` was seg faulting.

stanhu:
As I said in https://github.com/grpc/grpc/pull/35089#issuecomment-1825069477, inside `src/ruby`, I ran:

```shell
bundle exec rake compile
bundle exec rspec
```

It fails consistently for me with Ruby 3.2.2.

apolcyn:
ruby tests are red

stanhu:
@apolcyn Can you retry the tests again? I think the cert path was off.

stanhu:
Thanks, looks like all the relevant tests pass now. It seems that different definitions of `load_test_certs` were being used, depending on what order the tests were run. I had missed another duplicate in `src/ruby/spec/generic/rpc_server_spec.rb`.

PR #34481, opened 2023-09-26
[ruby] Fix compilation errors on clang 16

stanhu:
As announced in https://discourse.llvm.org/t/clang-16-notice-of-potentially-breaking-changes/65562, clang 16 defaults `-Wincompatible-function-pointer-types` to be on. This causes a number of hard errors due to implicit conversions between `void *` and `void`, such as:

```
../../../../src/ruby/ext/grpc/rb_channel.c:770:47: error: incompatible function pointer types passing 'VALUE (VALUE)' (aka 'unsigned long (unsigned long)') to parameter of type 'VALUE (*)(void *)' (aka 'unsigned long (*)(void *)') [-Wincompatible-function-pointer-types]
  g_channel_polling_thread = rb_thread_create(run_poll_channels_loop, NULL);
                                              ^~~~~~~~~~~~~~~~~~~~~~
/root/.rbenv/versions/3.1.4/include/ruby-3.1.0/ruby/internal/intern/thread.h:190:32: note: passing argument to parameter 'f' here
VALUE rb_thread_create(VALUE (*f)(void *g), void *g);
                               ^
../../../../src/ruby/ext/grpc/rb_channel.c:780:41: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
void grpc_rb_channel_polling_thread_stop() {
                                        ^
                                         void
../../../../src/ruby/ext/grpc/rb_channel.c:786:30: error: incompatible function pointer types passing 'void (void *)' to parameter of type 'void *(*)(void *)' [-Wincompatible-function-pointer-types]
  rb_thread_call_without_gvl(run_poll_channels_loop_unblocking_func, NULL, NULL,
```

This commit fixes these pointer types using wrapper functions where necessary.

This issue was also raised on the FreeBSD port of the grpc gem: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=271540

PR #35194, opened 2023-12-01
[Ruby] fix incompatible function pointer type error calling rb_thread_create

### LINKED ISSUE #35148: Make fails with "incompatible function pointer type" on Apple m3

All grpc gem versions from 1.54.2 thru 1.59.2, ruby

OS: macOS 14.1.1
Chip: M3 Max

ruby: 3.1.2 (and 3.2.2)
gcc: 13.2.0 (homebrew)
llvm: 17.0.6 (homebrew)
Command Line Tools: 15.0.0.0.1.1694021235
Xcode: 15.0.1

### What did you do?

I'm trying to install the grpc gem on a new Macbook Pro with an M3 Max chip using the command `gem install grpc -v 1.54.2`

I've also tried cloning the repo & building from source/running tests

I expected the gem to install after it builds from source

### What did you see instead?

I see 1 error that causes the grpc build to fail:

```
rb_channel.c:764:40: error: incompatible function pointer types passing 'VALUE (VALUE)' (aka 'unsigned long (unsigned long)') to parameter of type 'VALUE (*)(void *)' (aka 'unsigned long (*)(void *)') [-Wincompatible-function-pointer-types]
  764 |   background_thread = rb_thread_create(run_poll_channels_loop, NULL);
      |                                        ^~~~~~~~~~~~~~~~~~~~~~
```

I've tried installing with both default macOS binaries (gcc/clang, etc.) with similar results but a different set of error messages (ranlib, etc.)

I tried setting the relevant paths (LDFLAGS, CPPFLAGS, etc.) per the homebrew instructions, and without the paths.

I tried installing with ruby 3.2.2 with the same error

I tried installing 1.60.0.pre and that install worked ~so there may be a patch/workaround for it that I can upgrade to~. I can't use this as a workaround in the current situation, because of cascading dependencies. 

I'd love to find out that I'm missing something in the install, or that I've configured something wrong. The only solution that comes to mind is a) buying a new laptop (without an M3 chip); b) Forgetting about locally installing any of this and work inside of a docker container

I'm wondering if this is an issue related to the m3 chipset

alto-ruby:
rb_thread_create accept VALUE (*f)(void *g) for its first arg: https://github.com/ruby/ruby/blob/ruby_3_2/include/ruby/internal/intern/thread.h#L190

fixes #35148

### DIFF
--- a/src/ruby/ext/grpc/rb_channel.c
+++ b/src/ruby/ext/grpc/rb_channel.c
@@ -726,7 +726,7 @@ static void run_poll_channels_loop_unblocking_func(void* arg) {
 }
 
 // Poll channel connectivity states in background thread without the GIL.
-static VALUE run_poll_channels_loop(VALUE arg) {
+static VALUE run_poll_channels_loop(void* arg) {
   (void)arg;
   gpr_log(
       GPR_DEBUG,
--- a/src/ruby/ext/grpc/rb_event_thread.c
+++ b/src/ruby/ext/grpc/rb_event_thread.c
@@ -116,7 +116,7 @@ static void grpc_rb_event_unblocking_func(void* arg) {
 
 /* This is the implementation of the thread that handles auth metadata plugin
  * events */
-static VALUE grpc_rb_event_thread(VALUE arg) {
+static VALUE grpc_rb_event_thread(void* arg) {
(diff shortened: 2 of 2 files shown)


alto-ruby:
In favor of #34481

PR #35203, opened 2023-12-02
Update "upb::SymbolTable" to "upb::DefPool"

ericsalo:
The old name has been deprecated for a while so this should be safe.

### DIFF
--- a/src/core/ext/xds/xds_api.cc
+++ b/src/core/ext/xds/xds_api.cc
@@ -57,7 +57,7 @@
 namespace grpc_core {
 
 XdsApi::XdsApi(XdsClient* client, TraceFlag* tracer,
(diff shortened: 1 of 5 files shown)


src/core/ext/xds/xds_api.cc
```diff
@@ -57,7 +57,7 @@
 namespace grpc_core {
 
 XdsApi::XdsApi(XdsClient* client, TraceFlag* tracer,
-               const XdsBootstrap::Node* node, upb::SymbolTable* symtab,
+               const XdsBootstrap::Node* node, upb::DefPool* symtab,
```
markdroth, line 60: Please also change the name of the variables everywhere from "symtab" to "def_pool".

PR #35208, opened 2023-12-04
[upb] Update "upb::SymbolTable" to "upb::DefPool"

ericsalo:
The old name has been deprecated for a while so this should be safe. Also update variable names accordingly.


markdroth (approved):

ericsalo:
Hmm at least one of these failures appears to be a timeout but it seems that I don't have permission to kick off a rerun.

markdroth:
I've triggered a re-run for the two failed tests.

PR #35210, opened 2023-12-05
[CSM O11Y] CSM Service Label Plumbing from LB Policies to CallAttemptTracer

yijiem:


test/core/xds/xds_cluster_resource_type_test.cc
```diff
@@ -1409,6 +1411,35 @@ TEST_F(HostOverrideStatusTest, CanExplicitlySetToEmpty) {
   EXPECT_EQ(resource.override_host_statuses.ToString(), "{}");
 }
 
+using TelemetryLabelTest = XdsClusterTest;
+
+TEST_F(TelemetryLabelTest, ValidConfig) {
+  Cluster cluster;
+  cluster.set_type(cluster.EDS);
+  cluster.mutable_eds_cluster_config()->mutable_eds_config()->mutable_self();
+  ::google::protobuf::Struct telemetry_labels;
```
markdroth, line 1420: I think this can be written a bit more idiomatically, and with less unnecessary copying, using something like this:

```
auto& filter_map = *cluster.mutable_metadata()->mutable_filter_metadata();
auto& label_map = *filter_map["com.google.csm.telemetry_labels"]->mutable_fields();
*label_map["service_name"].mutable_string_value() = service_name;
*label_map["service_namespace"].mutable_string_value() = service_namespace;
```

test/cpp/ext/otel/otel_test_library.cc
```diff
@@ -38,11 +39,60 @@
...
+  grpc_core::ArenaPromise<grpc_core::ServerMetadataHandle> MakeCallPromise(
+      grpc_core::CallArgs call_args,
+      grpc_core::NextPromiseFactory next_promise_factory) override {
+    using CallAttemptTracer = grpc_core::ClientCallTracer::CallAttemptTracer;
+    auto* call_context = grpc_core::GetContext<grpc_call_context_element>();
+    auto* call_tracer = static_cast<CallAttemptTracer*>(
+        call_context[GRPC_CONTEXT_CALL_TRACER].value);
+    EXPECT_NE(call_tracer, nullptr);
+    auto service_labels =
+        std::make_shared<std::map<std::string, std::string>>();
```
markdroth, line 68: I think this can be written something like this:

```
auto service_labels =
    std::make_shared<std::map<std::string, std::string>>({
        {"service_name", service_name_},
        {"service_namespace", service_namespace_},
    });
```

test/cpp/ext/csm/metadata_exchange_test.cc
```diff
@@ -328,6 +338,26 @@ TEST_P(MetadataExchangeTest, ClientDoesNotSendMetadata) {
             "unknown");
 }
 
+TEST_P(MetadataExchangeTest, VerifyCsmServiceLabels) {
+  Init(/*metric_names=*/{grpc::experimental::OpenTelemetryPluginBuilder::
+                             kClientAttemptDurationInstrumentName},
+       /*enable_client_side_injector=*/true,
+       /*add_service_labels_in_call=*/true);
```
markdroth, line 345: I'm not crazy about how the data we're injecting is hidden in the test framework, not something you can see just be reading the individual test.  Instead, how about we pass in a map of labels to inject, and then then the test can check that those values are present in the OT metrics data?

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -304,6 +311,58 @@ TEST_P(CdsTest, ClusterChangeAfterAdsCallFails) {
...
+  return false;
+}
+
+TEST_P(CdsTest, VerifyCsmServiceLabelsParsing) {
+  CreateAndStartBackends(1);
+  // Populates EDS resources.
+  EdsResourceArgs args({{"locality0", CreateEndpointsForBackends()}});
+  balancer_->ads_service()->SetEdsResource(
+      BuildEdsResource(args, kDefaultClusterName));
+
```
markdroth, line 337: Nit: Please don't add unnecessary blank lines within functions.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -304,6 +311,58 @@ TEST_P(CdsTest, ClusterChangeAfterAdsCallFails) {
...
+  // Populates service labels to CDS resources.
+  auto cluster = default_cluster_;
+  google::protobuf::Struct labels;
```
markdroth, line 340: Same comment as in xds_cluster_resource_type_test.cc about writing this more idomatically and with less copying.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -304,6 +311,58 @@ TEST_P(CdsTest, ClusterChangeAfterAdsCallFails) {
...
+  }
+  return false;
+}
+
+TEST_P(CdsTest, VerifyCsmServiceLabelsParsing) {
+  CreateAndStartBackends(1);
+  // Populates EDS resources.
+  EdsResourceArgs args({{"locality0", CreateEndpointsForBackends()}});
+  balancer_->ads_service()->SetEdsResource(
+      BuildEdsResource(args, kDefaultClusterName));
```
markdroth, line 336: There's no reason to set the EDS resource name to be the same as the CDS resource name.  Just remove the second parameter to `BuildEdsResource()` and let it default to the right value.  And you can also remove line 347, where we set the CDS resource to point to an EDS resource of the same name.

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -304,6 +311,58 @@ TEST_P(CdsTest, ClusterChangeAfterAdsCallFails) {
...
+                          &fake_client_call_tracer_factory);
+  ResetStub(/*failover_timeout_ms=*/0, &channel_args);
+
+  // Sends an RPC and verifies that the service labels are recorded in the fake
+  // client call tracer.
+  EXPECT_TRUE(SendRpc().ok());
+  EXPECT_THAT(fake_client_call_tracer_factory.GetLastFakeClientCallTracer()
+                  ->GetLastCallAttemptTracer()
+                  ->GetoptionalLabels(),
+              VerifyCsmServiceLabels());
```
markdroth, line 363: Instead of moving the actual check out into a matcher, please just do it inline, so that all of the expectations are directly specified in the test case:

```
EXPECT_THAT(
    fake_client_call_tracer_factory.GetLastFakeClientCallTracer()
        ->GetLastCallAttemptTracer()
        ->GetoptionalLabels(),
    ::testing::ElementsAre(
        ::testing::Pair(
            OptionalLabelComponent::kXdsServiceLabels,
            ::testing::ElementsAre(
                ::testing::Pair("service_name", "myservice"),
                ::testing::Pair("service_namespace", "mynamespace")))));
```

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -304,6 +311,58 @@ TEST_P(CdsTest, ClusterChangeAfterAdsCallFails) {
...
+  // Injects a fake client call tracer factory.
+  grpc_core::FakeClientCallTracerFactory fake_client_call_tracer_factory;
+  ChannelArguments channel_args;
+  channel_args.SetPointer(GRPC_ARG_INJECT_FAKE_CLIENT_CALL_TRACER_FACTORY,
+                          &fake_client_call_tracer_factory);
+  ResetStub(/*failover_timeout_ms=*/0, &channel_args);
+
+  // Sends an RPC and verifies that the service labels are recorded in the fake
+  // client call tracer.
+  EXPECT_TRUE(SendRpc().ok());
```
markdroth, line 359: Please use `CheckRpcSendOk(DEBUG_LOCATION)`.

markdroth (commented):
This looks really good overall!  Comments are mostly minor things, but there are a couple of more significant issues.

src/cpp/ext/otel/otel_client_filter.cc
```diff
@@ -182,7 +183,14 @@ void OpenTelemetryCallTracer::OpenTelemetryCallAttemptTracer::
            {OpenTelemetryStatusKey(),
             grpc_status_code_to_string(
                 static_cast<grpc_status_code>(status.code()))}}};
-  KeyValueIterable labels(injected_labels_.get(), additional_labels);
+  std::unique_ptr<LabelsIterable> optional_labels;
+  if (OpenTelemetryPluginState().labels_injector != nullptr) {
+    optional_labels =
+        OpenTelemetryPluginState().labels_injector->GetLabelsFromOptionalLabels(
```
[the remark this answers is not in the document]
markdroth: Slight correction: We'd actually have to pass the vector into `KeyValueIterable` as a pointer, not a reference, since there are cases where there is no optional labels vector to use.

test/cpp/ext/csm/metadata_exchange_test.cc
yijiem: Done.

test/cpp/ext/otel/otel_test_library.cc
yijiem: No longer needed.

yijiem (commented):
Thanks Mark!

src/cpp/ext/otel/otel_client_filter.cc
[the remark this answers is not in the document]
yashykt: So, the flow of things is - 
1) CDS receives the cluster which has the metadata information containing the service name and namespace labels.
2) These labels are inserted into `optional_labels_vector_` through the call tracer. `optional_labels_vector_` is really an array of maps. Each map contains the labels added by a specific component. Right now, we have just one map added for CSM. In the future, we can have more.
3) We give this map to the PluginOption (which in the future, we can have many of), and then each PluginOption is going to look at the vector and decide which labels it wants to insert. Note that as part of this, it can also change the label key. In our case, the CSM Observability code fetches the map with the service name and namespace labels, and returns an iterator which returns labels in the exact format we need for CSM, i.e. with the "csm." prefix.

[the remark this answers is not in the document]
yashykt: If we were to pass `optional_labels_vector_` directly to `KeyValueIterable`, we'd lose the ability for PluginOptions to decide which labels they want from the optional labels and in what format.

src/cpp/ext/otel/otel_call_tracer.h
```diff
@@ -91,13 +91,20 @@ class OpenTelemetryCallTracer : public grpc_core::ClientCallTracer {
...
 
    private:
     const OpenTelemetryCallTracer* parent_;
     const bool arena_allocated_;
     // Start time (for measuring latency).
     absl::Time start_time_;
     std::unique_ptr<LabelsIterable> injected_labels_;
+    // The indices of the vector correspond to the OptionalLabelComponent enum.
+    std::array<std::shared_ptr<std::map<std::string, std::string>>,
+               static_cast<size_t>(OptionalLabelComponent::kSize)>
```
yashykt, line 106: This is not a vector anymore :)

src/core/ext/xds/xds_cluster.cc
```diff
@@ -645,6 +646,34 @@ absl::StatusOr<std::shared_ptr<const XdsClusterResource>> CdsResourceParse(
...
+  const envoy_config_core_v3_Metadata* metadata =
+      envoy_config_cluster_v3_Cluster_metadata(cluster);
+  if (metadata != nullptr) {
+    google_protobuf_Struct* telemetry_labels;
+    if (envoy_config_core_v3_Metadata_filter_metadata_get(
+            metadata,
+            StdStringToUpbString(
+                absl::string_view("com.google.csm.telemetry_labels")),
+            &telemetry_labels)) {
+      size_t iter = -1;
```
yashykt, line 661: probably not a big deal, but you don't need to initialize `iter` and `fields_entry`

test/core/util/fake_stats_plugin.cc
```diff
@@ -0,0 +1,85 @@
+// Copyright 2023 The gRPC Authors.
```
yashykt, line 1: Is this supposed to be 2024 now?

yashykt (approved):

src/cpp/ext/otel/otel_client_filter.cc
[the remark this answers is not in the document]
markdroth: Okay, that makes sense.  And I agree that we do want the plugin to be able to decide which labels they want from the optional labels.  But then I think we need to find a way to implement this that doesn't add so much per-call allocation.

How about adding a method like this to `LabelsInjector`:

```
// Returns true if the label should be included.
bool UseLabel(absl::string_view key, absl::string_view value);
```

Then we have `KeyValueIterable` call that method on each individual optional label.  I think this would do what we need without any allocation, wouldn't it?

test/core/util/fake_stats_plugin.cc
markdroth: We started writing this code in 2023, so I think that's the right year to use.

markdroth (commented):
This is looking really good!  The only remaining substantive issue is the one about the per-call allocations.

src/cpp/ext/otel/otel_client_filter.cc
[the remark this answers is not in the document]
yashykt: Note that we can have multiple plugin options after my change in https://github.com/grpc/grpc/pull/35434. So, if we were to implement that strategy, it would look something like - 
Each time the `KeyValueIterable` is used (3 times at the moment), we will have to loop over each label and ask each plugin option as to what label, if any, it wants to insert.

Downsides - 
1) We have to run this thrice, and we end up creating the strings that many times as well.
2) We lose the ability for plugin options to look at multiple labels at once and create a single label. (We don't have a usecase at present, but we might want it in the future.)

test/cpp/ext/csm/metadata_exchange_test.cc
```diff
@@ -342,7 +342,8 @@ TEST_P(MetadataExchangeTest, VerifyCsmServiceLabels) {
   Init(/*metric_names=*/{grpc::experimental::OpenTelemetryPluginBuilder::
                              kClientAttemptDurationInstrumentName},
        /*enable_client_side_injector=*/true,
-       /*add_service_labels_in_call=*/true);
+       // Injects CSM service labels to be recorded in the call.
+       {{"service_name", "myservice"}, {"service_namespace", "mynamespace"}});
```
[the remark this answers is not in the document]
yashykt: resolved on chat

test/core/xds/xds_cluster_resource_type_test.cc
```diff
@@ -1448,7 +1448,7 @@ TEST_F(TelemetryLabelTest, MissingMetadataField) {
   ASSERT_TRUE(decode_result.resource.ok()) << decode_result.resource.status();
   auto& resource =
       static_cast<const XdsClusterResource&>(**decode_result.resource);
-  EXPECT_THAT(*resource.telemetry_labels, ::testing::IsEmpty());
+  EXPECT_THAT(resource.telemetry_labels, ::testing::IsNull());
```
markdroth, line 1451: Can just say `EXPECT_EQ(resource.telemetry_labels, nullptr)`.

Same on line 1469.

yijiem: Done.

markdroth (commented):
This is getting really close!

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -427,5 +428,36 @@ void ServiceMeshLabelsInjector::AddLabels(
...
+             opentelemetry::common::AttributeValue)>
+        callback) const {
+  // According to the CSM Observability Metric spec, if the control plane fails
+  // to provide these labels, the client will set their values to "unknown".
+  // These default values are set below.
+  absl::string_view service_name = "unknown";
+  absl::string_view service_namespace = "unknown";
+  // Performs JSON label name format to CSM Observability Metric spec format
+  // conversion.
+  for (const auto& optional_labels : optional_labels_span) {
```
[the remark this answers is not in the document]
yijiem: Done.

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -427,5 +428,36 @@ void ServiceMeshLabelsInjector::AddLabels(
...
+    if (optional_labels == nullptr) {
+      continue;
+    }
+    if (optional_labels->find("service_name") != optional_labels->end()) {
```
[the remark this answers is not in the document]
yijiem: Done.

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -104,15 +120,21 @@ class KeyValueIterable : public opentelemetry::common::KeyValueIterable {
...
     size += additional_labels_.size();
+    for (const auto& optional_labels : optional_labels_) {
+      size += optional_labels->size();
+    }
     return size;
   }
 
  private:
   LabelsIterable* injected_labels_iterable_;
+  const ActivePluginOptionsView* active_plugin_options_view_;
```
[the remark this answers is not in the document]
yijiem: Done.

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -50,14 +50,19 @@ class KeyValueIterable : public opentelemetry::common::KeyValueIterable {
  public:
   explicit KeyValueIterable(
       LabelsIterable* injected_labels_iterable,
+      const ActivePluginOptionsView* active_plugin_options_view,
```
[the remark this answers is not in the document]
yijiem: Done.

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -104,15 +120,21 @@ class KeyValueIterable : public opentelemetry::common::KeyValueIterable {
       }
     }
     size += additional_labels_.size();
+    for (const auto& optional_labels : optional_labels_) {
+      size += optional_labels->size();
```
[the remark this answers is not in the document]
yijiem: Good catch! I added a new API (`size_t GetOptionalLabelsSize(absl::Span<const std::shared_ptr<std::map<std::string, std::string>>> optional_labels) const`) to the LabelInjector for the PluginOption to return the actual size with a reference of the optional_labels. PTAL.

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -442,15 +446,14 @@ bool ServiceMeshLabelsInjector::AddOptionalLabels(
...
-  for (const auto& optional_labels : optional_labels_span) {
-    if (optional_labels == nullptr) {
-      continue;
-    }
-    if (optional_labels->find("service_name") != optional_labels->end()) {
-      service_name = optional_labels->at("service_name");
-    }
-    if (optional_labels->find("service_namespace") != optional_labels->end()) {
-      service_namespace = optional_labels->at("service_namespace");
+  if (!optional_labels_span.empty()) {
```
markdroth, line 449: I think it would be safer to write this as:

```
if (optional_labels_span.size() > OptionalLabelComponent::kXdsServiceLabels) {
```

That way, if `kXdsServiceLabels` was changed to not be the first element, we'd still do the right thing if the span was non-empty but not long enough.

markdroth (approved):
This looks great!  Just one remaining minor comment; feel free to merge after addressing.

src/cpp/ext/csm/metadata_exchange.cc
yijiem: Makes sense. Done.

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -72,6 +77,20 @@ class KeyValueIterable : public opentelemetry::common::KeyValueIterable {
         }
       }
     }
+    if (OpenTelemetryPluginState().labels_injector != nullptr &&
+        !OpenTelemetryPluginState().labels_injector->AddOptionalLabels(
+            optional_labels_, callback)) {
+      return false;
+    }
+    if (active_plugin_options_view_ != nullptr) {
+      active_plugin_options_view_->ForEach(
```
yashykt, line 86: `ForEach` needs to be modified to stop iterating over the plugin options and return false if any of plugin option returns false.
After that, `ForEach` needs to return the false itself, so that we can replicate the logic that if any callback returned false, this function returns a false as well

yashykt (changes requested):

src/cpp/ext/otel/key_value_iterable.h
yijiem: Done. PTAL. Thanks!

yashykt (approved):

PR #35218, opened 2023-12-05
[examples] Implement custom metrics reporting from server

eugeneo:


examples/cpp/orca/orca_server.cc
```diff
@@ -0,0 +1,96 @@
...
+using grpc::Status;
+using helloworld::Greeter;
+using helloworld::HelloReply;
+using helloworld::HelloRequest;
+
+ABSL_FLAG(uint16_t, port, 50051, "Server port for the service");
+
+// Logic and data behind the server's behavior.
+class GreeterServiceImpl final : public Greeter::Service {
+  Status SayHello(ServerContext* context, const HelloRequest* request,
```
yashykt, line 46: Prefer callback

PR #35219, opened 2023-12-05
[example] Add WAIT_FOR_READY example

yashykt:


eugeneo (approved):
I approved, though I would suggest making it more focused.

PR #35226, opened 2023-12-06
binder: skip all tests on ios

jdcormie:
They don't build due to -DNO_GRPC_BINDER.

### DIFF
--- a/test/core/transport/binder/end2end/BUILD
+++ b/test/core/transport/binder/end2end/BUILD
@@ -34,7 +34,6 @@ grpc_cc_library(
         "absl/time",
         "absl/types:variant",
     ],
-    tags = ["no_test_ios"],
     deps = [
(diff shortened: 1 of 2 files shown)


test/core/transport/binder/end2end/BUILD
```diff
@@ -34,6 +34,7 @@ grpc_cc_library(
         "absl/time",
         "absl/types:variant",
     ],
+    tags = ["no_test_ios"],
```
sifmelcara, line 37: This is a `grpc_cc_library` target, do we need the tag here?

jdcormie: fixed thanks!

sifmelcara:
LGTM. @eugeneo would you like to also take a look?

cc @rock1246

PR #35238, opened 2023-12-07
[upb] cast to (upb_Message*) when calling upb functions

ericsalo:
upb currently defines upb_Message to be void, which has allowed us to be lax with the types when calling upb functions that take a (upb_Message*) as an arg.

If that ever changes (which may happen quite soon) things will break, so this is a pre-emptive patch to protect against such a change.


markdroth:
Please use C++-style cases instead of C-style casts.

ericsalo:
Will do. Closing for now.

PR #35247, opened 2023-12-07
[Security - TLS] DirectoryReloaderCrlProvider construction time bug fix

gtcooke94:
The `DirectoryReloaderProvider` currently segfaults on construction if grpc_init() is not called before construction. This is because when creating the `DirectoryReloaderCrlProvider` we [call GetDefaultEventEngine](https://github.com/grpc/grpc/blob/a58f3f2df531c3fae2afaef39d37186cf9206778/src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc#L152), and getting the default event engine requires that `grpc_init` is called.

This PR adds a test that catches the segfault and adds `grpc_init` and `grpc_shutdown` to the ctor and dtor of `DirectoryReloaderCrlProvider` so that the test passes.


src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc
```diff
@@ -148,19 +148,36 @@ absl::StatusOr<std::shared_ptr<CrlProvider>> CreateDirectoryReloaderCrlProvider(
     return absl::InvalidArgumentError("Refresh duration minimum is 60 seconds");
   }
   auto provider = std::make_shared<DirectoryReloaderCrlProvider>(
-      refresh_duration, reload_error_callback,
-      grpc_event_engine::experimental::GetDefaultEventEngine(),
+      refresh_duration, reload_error_callback, nullptr,
```
matthewstevenson88, line 151: micro-nit: Please add `/*event_engine=*/` in front of nullptr.

test/core/security/grpc_tls_directory_reloader_crl_provider_test.cc
```diff
@@ -0,0 +1,58 @@
...
+static constexpr absl::string_view kCrlDirectory =
+    "test/core/tsi/test_creds/crl_data/crls";
+
+namespace grpc_core {
+namespace testing {
+
+// This test is needed separately from the tests in grpc_tls_crl_provider_test
+// because the DirectoryReloaderCrlProvider gets the default event engine on
+// construction. To get the default event engine, grpc_init must have been
+// called, otherwise a segfault occurs. This test checks that grpc_init gets
```
matthewstevenson88, line 41: nit on the last sentence: Could we maybe say "This test checks that no segfault occurs." instead of "This test checks that grpc_init gets called..."?

test/cpp/end2end/crl_provider_test.cc
```diff
@@ -220,6 +221,48 @@ TEST_F(CrlProviderTest, CrlProviderRevokedServer) {
...
+  auto certificate_provider =
+      std::make_shared<experimental::StaticDataCertificateProvider>(
+          root_cert, identity_key_cert_pairs);
+  grpc::experimental::TlsChannelCredentialsOptions options;
+  options.set_certificate_provider(certificate_provider);
+  options.watch_root_certs();
+  options.set_root_cert_name("root");
+  options.watch_identity_key_cert_pairs();
+  options.set_identity_cert_name("identity");
+  std::string root_crl = grpc_core::testing::GetFileContents(kRootCrlPath);
```
matthewstevenson88, line 251: nit: Is this variable needed?

test/cpp/end2end/crl_provider_test.cc
```diff
@@ -220,6 +221,48 @@ TEST_F(CrlProviderTest, CrlProviderRevokedServer) {
   DoRpc(server_addr_, options, false);
 }
 
+TEST_F(CrlProviderTest, CrlProviderValidReloader) {
```
matthewstevenson88, line 224: To make sure I understand correctly, the difference between this test and the `CrlProviderValid` test is that this test reloads from a directory whereas the other test is explicitly given the CRL at construction time. Is that correct?

Could you edit the test names to help clarify the difference? And/or maybe add a comment to explain this?

matthewstevenson88, line 224: (Can be deferred to a follow-up PR if you like) Should we have an e2e test where we try to reload from an empty directory?

matthewstevenson88 (commented):
Thanks for the PR Greg!

test/core/security/grpc_tls_directory_reloader_crl_provider_test.cc
```diff
@@ -0,0 +1,58 @@
...
+#include "src/core/lib/security/credentials/tls/grpc_tls_crl_provider.h"
+#include "test/core/util/test_config.h"
+
+static constexpr absl::string_view kCrlDirectory =
+    "test/core/tsi/test_creds/crl_data/crls";
+
+namespace grpc_core {
+namespace testing {
+
+// This test is needed separately from the tests in grpc_tls_crl_provider_test
```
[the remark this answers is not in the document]
gtcooke94: I had it in [this file](https://github.com/grpc/grpc/blob/5f92a67f948ea7436d78d7bc9be9ee0de883b64b/test/core/security/grpc_tls_crl_provider_test.cc#L100) as a different test class, and it was passing. I'm not very clear on how `grpc_init` interacts with different instantiations of test classes, so I thought it might need it's own file?

test/core/security/grpc_tls_directory_reloader_crl_provider_test.cc
gtcooke94: Changed

test/cpp/end2end/crl_provider_test.cc
gtcooke94: This is the same e2e test but with the directory reloader vs. the static provider

gtcooke94: I think the edge cases are covered in unit tests for the directory reloader, I don't see much value in having e2e tests that test the same edge cases, what do you think?

test/cpp/end2end/crl_provider_test.cc
gtcooke94: Removed

test/cpp/end2end/crl_provider_test.cc
matthewstevenson88: Thanks for clarifying!

matthewstevenson88: The reason I mention this here is that may be good to have an e2e test that ensures that the DynamicReloaderCrlProvider's GetCrl() method plays well with the code in ssl_transport_security.cc when it returns an either a missing or empty CRL. If you think the unit test coverage between DynamicReloaderCrlProvider and ssl_transport_security.cc is sufficient, that's totally fine to leave it, but just wanted to call this out as a place where we should ensure there is adequate testing.

matthewstevenson88 (approved):
LGTM from my end, since the last comment thread I started is non-blocking.

Please wait for @markdroth's approval before merging though. :)

test/cpp/end2end/crl_provider_test.cc
gtcooke94: I suppose it can't hurt to expand the coverage - I'll make a not to myself, I don't think it belongs in this PR

PR #35248, opened 2023-12-07
[Build] Expose Completion Queue and related API

dawidcha:
Expose Completion Queue and related classes and symbols for Windows DLL build

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,10 @@ set(PACKAGE_TARNAME       "${PACKAGE_NAME}-${PACKAGE_VERSION}")
 set(PACKAGE_BUGREPORT     "https://github.com/grpc/grpc/issues/")
 project(${PACKAGE_NAME} LANGUAGES C CXX)
 
+if(BUILD_SHARED_LIBS AND MSVC)
+  set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+endif()
+
 set(gRPC_INSTALL_BINDIR "bin" CACHE STRING "Installation directory for executables")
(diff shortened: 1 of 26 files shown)


dawidcha:
Thanks for the approval and the help with CLA. Is this ready to be merged now?

veblush:
David, I have two questions about this change.

- As DLL build primarily relies on `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` feature, I thought that a few edge cases actually need an explicit annotation but this PR doesn't seem to be the case. Is there something changed? 
- Why do you need to change `rb_grpc_imports.generated.h`? This is for Ruby.

dawidcha:
> David, I have two questions about this change.
> 
> * As DLL build primarily relies on `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` feature, I thought that a few edge cases actually need an explicit annotation but this PR doesn't seem to be the case. Is there something changed?
> * Why do you need to change `rb_grpc_imports.generated.h`? This is for Ruby.

So looking again at this, I'm reminded that when I worked on it I was looking to expose more of the API than was just needed for our purposes, in the hope that it could be more useful to more people. Let me see if I can reduce the number of changes to the bare minimum.

The ruby header got changed as far as I remember because some symbols from it were unresolved in the link step. I'll confirm that that is still needed.

dawidcha:
Hi @veblush,

I have reworked this to get to a minimal set of changes that still work with our code. I've abandoned the attempt to annotate classes and functions with dllimport/export tags. This results in slightly less efficient calling for clients, but it certainly reduces the number of changes needed and likely means that the result is less fragile to code reorganizations.

I've created a new PR here https://github.com/grpc/grpc/pull/35484 and will abandon this one.

PR #35250, opened 2023-12-07
[upb] add explicit casting to (upb_Message*) when calling into upb

ericsalo:
Now with improved C++-style casting!


ericsalo:
Abandoning for the moment.

PR #35254, opened 2023-12-08
[PSM Interop] Configure CODEOWNERS

sergiitk:
Configures CODEOWNERS for the PSM Interop framework `/tools/run_tests/xds_k8s_test_driver/` to be @sergiitk, @XuanWang-Amos, @gnossen.

We need this change to be able to prevent unexpected changes to the the framework while it's being moved to the dedicated grpc/psm-interop repo.


markdroth (approved):
Note that the way we have our repo configured, ownership really just means that you get CCed on PRs; it does not require that one of these people approves every PR.

PR #35284, opened 2023-12-12

stanley-cheung:


veblush:
Oh we cannot update boringssl until we bump cmake version in a few test cases.

PR #35285, opened 2023-12-12
[Protobuf] Update third_party/protobuf to 3.25.1

stanley-cheung:


veblush:
This cannot be merged as it is. We need to bundle upb in the same PR.

PR #35292, opened 2023-12-12
[python O11Y] Implement OpenTelemetry

XuanWang-Amos:

### DIFF
--- a/src/python/grpcio_observability/grpc_observability/__init__.py
+++ b/src/python/grpcio_observability/grpc_observability/__init__.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from grpc_observability._gcp_observability import GCPOpenCensusObservability
+from grpc_observability._open_telemetry_observability import (
+    OpenTelemetryObservability,
+)
+from grpc_observability._open_telemetry_plugin import OpenTelemetryPlugin
 
-__all__ = ("GCPOpenCensusObservability",)
+__all__ = ("OpenTelemetryObservability", "OpenTelemetryPlugin")
--- a/src/python/grpcio_observability/grpc_observability/_open_telemetry_plugin.py
+++ b/src/python/grpcio_observability/grpc_observability/_open_telemetry_plugin.py
@@ -18,6 +18,7 @@ from typing import Any, Dict, Iterable, List, Optional, Union
 # pytype: disable=pyi-error
 import grpc
 from grpc_observability import _open_telemetry_measures
+from grpc_observability._cyobservability import MetricsName
 from grpc_observability._observability import StatsData
 from opentelemetry.sdk.metrics import Counter
 from opentelemetry.sdk.metrics import Histogram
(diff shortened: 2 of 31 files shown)


src/python/grpcio_observability/grpc_observability/__init__.py
```diff
@@ -12,6 +12,9 @@
...
 # limitations under the License.
 
-from grpc_observability._gcp_observability import GCPOpenCensusObservability
+from grpc_observability._open_telemetry_observability import (
+    OpenTelemetryObservability,
+)
+from grpc_observability._open_telemetry_plugin import OpenTelemetryPlugin
 
-__all__ = ("GCPOpenCensusObservability",)
+__all__ = ("OpenTelemetryObservability", "OpenTelemetryPlugin")
```
gnossen, line 20: Is `OpenTelemetryObservability` tied to GCP? Are other languages including "GCP" in the name of their equivalent API?

src/python/grpcio_tests/tests/observability/_open_telemetry_observability_test.py
```diff
@@ -0,0 +1,232 @@
...
+    def force_flush(self, timeout_millis: float = 10_000) -> bool:
+        return True
+
+    def record_metric(self, metrics_data: MetricsData) -> None:
+        for resource_metric in metrics_data.resource_metrics:
+            for scope_metric in resource_metric.scope_metrics:
+                for metric in scope_metric.metrics:
+                    self.all_metrics["name"].add(metric.name)
+                    for data_point in metric.data.data_points:
+                        for key, value in data_point.attributes.items():
```
gnossen, line 82: Nit: Could use `defaultdict(set)` for `self.all_metrics["label"]` and then just do `self.all_metrics["label"][key].add(value)`

src/python/grpcio_tests/tests/observability/_open_telemetry_observability_test.py
```diff
@@ -0,0 +1,232 @@
...
+            _test_server.unary_unary_filtered_call(port=port)
+
+        self._validate_metrics_exist(self.all_metrics)
+        method_values = self.all_metrics["label"].get(GRPC_METHOD_LABEL)
+        self.assertTrue(GRPC_OTHER_LABEL_VALUE in method_values)
+        self.assertTrue(FILTERED_METHOD_NAME not in method_values)
+
+    def _validate_metrics_exist(self, all_metrics: Dict[str, Any]) -> None:
+        # Sleep here to make sure we have at least one export from OTel MetricExporter.
+        time.sleep(OTEL_EXPORT_INTERVAL_S)
```
gnossen, line 215: In general, I'm very wary of using statically timed sleeps in a test. They tend to produce flakiness because, even if it can be proven that two threads of execution _should_ interleave a certain way because of sleeps with specific times they have in their implementation, thread pre-emption creates cases where they don't _always_ interleave the way intended. I strongly suggest that you make this assertion _eventual_. In other words, retry multiple times up to a configurable (long -- multiple seconds) timeout. Check out this [`assert_eventually`](https://github.com/grpc/grpc/blob/6d35000c17d96819184f344183b07104f5b7eb16/src/python/grpcio_tests/tests_py3_only/unit/_simple_stubs_test.py#L170) higher order function we already have for cases like this.

We should also characterize how flaky this test is already by running under bazel 1000 times. We should aim for fewer than 2 failures/timeouts out of 1000.

src/python/grpcio_tests/tests/observability/_open_telemetry_observability_test.py
```diff
@@ -0,0 +1,232 @@
...
+        self.assertGreater(len(all_metrics["name"]), 0)
+        self._validate_metrics_names(all_metrics["name"])
+
+    def _validate_metrics_names(self, metric_names: Set[str]) -> None:
+        for base_metric in _open_telemetry_measures.base_metrics():
+            if base_metric.name not in metric_names:
+                logger.error(
```
gnossen, line 222: [`assertTrue` takes a second argument](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue) for a message to print to `stderr`

src/python/grpcio_observability/grpc_observability/server_call_tracer.cc
```diff
@@ -226,7 +226,7 @@ void PythonOpenCensusServerCallTracer::RecordEnd(
   if (PythonCensusStatsEnabled()) {
     const uint64_t request_size = GetOutgoingDataSize(final_info);
     const uint64_t response_size = GetIncomingDataSize(final_info);
-    double elapsed_time_ms = absl::ToDoubleMilliseconds(elapsed_time_);
+    double elapsed_time_ms = absl::ToDoubleSeconds(elapsed_time_);
```
gnossen, line 229: Seems like there's now a mismatch between the variable name and the unit used? Is this double supposed to hold seconds or milliseconds?

gnossen (commented):
Just a first batch. Not a full review yet.

gnossen (commented):
I think I've reviewed everything now. Mostly looks good.

src/python/grpcio_observability/grpc_observability/_open_telemetry_plugin.py
```diff
@@ -0,0 +1,231 @@
...
+from opentelemetry.sdk.metrics import Histogram
+from opentelemetry.sdk.metrics import Meter
+from opentelemetry.sdk.metrics import MeterProvider
+
+GRPC_METHOD_LABEL = "grpc.method"
+GRPC_TARGET_LABEL = "grpc.target"
+GRPC_OTHER_LABEL_VALUE = "other"
+
+
+class LabelInjector(abc.ABC):
```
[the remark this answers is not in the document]
gnossen: Please assign an issue against yourself to create a test [similar to this one](https://github.com/grpc/grpc/blob/75e5ebcb14a7a16e1cb5994853e2a6215dbf97e8/src/python/grpcio_tests/tests/unit/_api_test.py#L114) for the `grpc_observability` package so that we can protect against regressions in the public API.

PR #35295, opened 2023-12-12
[EventEngine] Replace Executor with EE::Run in Client Reactor

drfloob:


include/grpc/support/call.h
```diff
@@ -0,0 +1,29 @@
+// Copyright 2023 The gRPC Authors
```
ctiller, line 1: Let's move this to `include/grpc/impl/call.h`
... I believe the stance is that `support` is stable enough that we need a grfc for, but `impl` is not.

drfloob: Done!

ctiller (approved):

PR #35300, opened 2023-12-13
[ObjC] add PrivacyInfo.xcprivacy to Package.swift

HannahShiSFB:
Privacy info for swift package.


sampajano (approved):
Thanks for the change! 😃

PR #35347, opened 2023-12-19
binder: rename two non-public classes to camel case.

jdcormie:
https://google.github.io/styleguide/cppguide.html#Type_Names


jdcormie:
Good afternoon ctiller! Do you think I should move these classes into a namespace in lieu of the `grpc_binder_` prefix? If so, do you think that should be part of this PR or a follow up? Thanks!

jdcormie:
Friendly ping to @ctiller for my question above.

ctiller:
This transport has been deleted.

PR #35348, opened 2023-12-19
[OTel] Experimental API for metrics

yashykt:
Provide a public experimental API and bazel compatible build target for OpenTelemetry metrics.

Details -
* New `OpenTelemetryPluginBuilder` class that provides the API specified in https://github.com/grpc/proposal/blob/master/A66-otel-stats.md
* The existing `grpc::internal::OpenTelemetryPluginBuilder` class is moved to `grpc::internal::OpenTelemetryPluginBuilderImpl` for disambiguation.
* Renamed `OTel` in some instances to `OpenTelemetry` for consistency.

### DIFF
--- a/include/grpcpp/ext/otel_plugin.h
+++ b/include/grpcpp/ext/otel_plugin.h
@@ -51,43 +51,43 @@ OpenTelemetryServerCallSentTotalCompressedMessageSizeInstrumentName();
 absl::string_view
 OpenTelemetryServerCallRcvdTotalCompressedMessageSizeInstrumentName();
 
+/// The most common way to use this API is -
+///
+/// OpenTelemetryPluginBuilder().SetMeterProvider(provider).BuildAndRegister();
+///
+/// The set of instruments available are -
+/// grpc.client.attempt.started
+/// grpc.client.attempt.duration
+/// grpc.client.attempt.sent_total_compressed_message_size
+/// grpc.client.attempt.rcvd_total_compressed_message_size
+/// grpc.server.call.started
+/// grpc.server.call.duration
+/// grpc.server.call.sent_total_compressed_message_size
+/// grpc.server.call.rcvd_total_compressed_message_size
 class OpenTelemetryPluginBuilder {
  public:
   OpenTelemetryPluginBuilder();
-  // If `SetMeterProvider()` is not called, no metrics are collected.
+  /// If `SetMeterProvider()` is not called, no metrics are collected.
   OpenTelemetryPluginBuilder& SetMeterProvider(
       std::shared_ptr<opentelemetry::metrics::MeterProvider> meter_provider);
-  // If set, \a target_attribute_filter is called per channel to decide whether
-  // to record the target attribute on client or to replace it with "other".
-  // This helps reduce the cardinality on metrics in cases where many channels
-  // are created with different targets in the same binary (which might happen
-  // for example, if the channel target string uses IP addresses directly).
+  /// If set, \a target_attribute_filter is called per channel to decide whether
+  /// to record the target attribute on client or to replace it with "other".
+  /// This helps reduce the cardinality on metrics in cases where many channels
+  /// are created with different targets in the same binary (which might happen
+  /// for example, if the channel target string uses IP addresses directly).
   OpenTelemetryPluginBuilder& SetTargetAttributeFilter(
       absl::AnyInvocable<bool(absl::string_view /*target*/) const>
           target_attribute_filter);
-  // If set, \a generic_method_attribute_filter is called per call with a
-  // generic method type to decide whether to record the method name or to
-  // replace it with "other". Non-generic or pre-registered methods remain
-  // unaffected. If not set, by default, generic method names are replaced with
-  // "other" when recording metrics.
+  /// If set, \a generic_method_attribute_filter is called per call with a
+  /// generic method type to decide whether to record the method name or to
+  /// replace it with "other". Non-generic or pre-registered methods remain
+  /// unaffected. If not set, by default, generic method names are replaced with
+  /// "other" when recording metrics.
   OpenTelemetryPluginBuilder& SetGenericMethodAttributeFilter(
       absl::AnyInvocable<bool(absl::string_view /*generic_method*/) const>
           generic_method_attribute_filter);
-  // Registers a global plugin that acts on all channels and servers running on
-  // the process.
-  // The most common way to use this API is -
-  //
-  // OpenTelemetryPluginBuilder().SetMeterProvider(provider).BuildAndRegister();
-  //
-  // The set of instruments available are -
-  // grpc.client.attempt.started
-  // grpc.client.attempt.duration
-  // grpc.client.attempt.sent_total_compressed_message_size
-  // grpc.client.attempt.rcvd_total_compressed_message_size
-  // grpc.server.call.started
-  // grpc.server.call.duration
-  // grpc.server.call.sent_total_compressed_message_size
-  // grpc.server.call.rcvd_total_compressed_message_size
+  /// Registers a global plugin that acts on all channels and servers running on
+  /// the process.
   void BuildAndRegisterGlobal();
 
  private:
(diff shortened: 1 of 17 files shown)


include/grpcpp/ext/otel_plugin.h
```diff
@@ -0,0 +1,99 @@
...
+  // generic method type to decide whether to record the method name or to
+  // replace it with "other". Non-generic or pre-registered methods remain
+  // unaffected. If not set, by default, generic method names are replaced with
+  // "other" when recording metrics.
+  OpenTelemetryPluginBuilder& SetGenericMethodAttributeFilter(
+      absl::AnyInvocable<bool(absl::string_view /*generic_method*/) const>
+          generic_method_attribute_filter);
+  // Registers a global plugin that acts on all channels and servers running on
+  // the process.
+  // The most common way to use this API is -
```
markdroth, line 78: The rest of this comment should probably go above the class.

include/grpcpp/ext/otel_plugin.h
```diff
@@ -0,0 +1,99 @@
...
+absl::string_view OpenTelemetryServerCallDurationInstrumentName();
+absl::string_view
+OpenTelemetryServerCallSentTotalCompressedMessageSizeInstrumentName();
+absl::string_view
+OpenTelemetryServerCallRcvdTotalCompressedMessageSizeInstrumentName();
+
+class OpenTelemetryPluginBuilder {
+ public:
+  OpenTelemetryPluginBuilder();
+  // If `SetMeterProvider()` is not called, no metrics are collected.
```
markdroth, line 57: Comments that should appear in our doxygen doc need three slashes instead of two.

src/cpp/ext/csm/csm_observability.cc
```diff
@@ -48,40 +48,46 @@ namespace experimental {
 // CsmObservabilityBuilder
 //
 
+CsmObservabilityBuilder::CsmObservabilityBuilder()
+    : builder_(
+          std::make_unique<grpc::internal::OpenTelemetryPluginBuilderImpl>()) {}
+
+CsmObservabilityBuilder::~CsmObservabilityBuilder() = default;
```
markdroth, line 55: Why is this needed?

src/cpp/ext/otel/BUILD
```diff
@@ -41,6 +41,7 @@ grpc_cc_library(
         "otel_client_filter.h",
         "otel_plugin.h",
         "otel_server_call_tracer.h",
+        "//:include/grpcpp/ext/otel_plugin.h",
```
markdroth, line 44: Is it going to cause a problem to have this header in two different targets?

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -264,4 +264,76 @@ void OpenTelemetryPluginBuilder::BuildAndRegisterGlobal() {
 }
 
 }  // namespace internal
+
+namespace experimental {
+absl::string_view OpenTelemetryClientAttemptStartedInstrumentName() {
```
markdroth, line 269: We may have already discussed this, but is there a reason why these need to be functions instead of just constexpr constants?

include/grpcpp/ext/otel_plugin.h
```diff
@@ -0,0 +1,99 @@
...
+#include <memory>
+
+#include "absl/functional/any_invocable.h"
+#include "absl/strings/string_view.h"
+#include "opentelemetry/metrics/meter_provider.h"
+
+namespace grpc {
+
+namespace internal {
+class OpenTelemetryPluginBuilderImpl;
```
markdroth, line 36: I assume that once we have the plugin option API in place, there will no longer be any need for having two different layers here (the private API wrapped inside of the public API)?

markdroth (approved):

include/grpcpp/ext/otel_plugin.h
yashykt: Yeah, I think once the experimental `CsmObservability` API goes away, we won't need to do this.

include/grpcpp/ext/otel_plugin.h
yashykt: ah, thanks!

include/grpcpp/ext/otel_plugin.h
yashykt: done!

src/cpp/ext/csm/csm_observability.cc
yashykt: The issue is that `csm_observability.h` only has a forward declaration of `grpc::internal::OpenTelemetryPluginBuilderImpl`. Since we are using this forward declared type inside a unique_ptr, the default destructor of the class can not be used. (That would end up invoking the destructor of the unique_ptr from the header which still doesn't know the type definition resulting in an incomplete type error of the form - 
```
error: invalid application of 'sizeof' to an incomplete type 'grpc::internal::OpenTelemetryPluginBuilderImpl'
```
.) 
So, to avoid this, we define the destructor in the cc file where we have the definition of `grpc::internal::OpenTelemetryPluginBuilderImpl`.

src/cpp/ext/otel/BUILD
yashykt: This is an internal target that we can change whenever, so it should be fine

src/cpp/ext/otel/otel_plugin.cc
yashykt: I had tried that some years ago but had ran into issues. Let me try it again and see if it works now.

yashykt: pushed a canary change to see if the tests are happy with it. If they are, I'll use it for all.

include/grpcpp/ext/otel_plugin.h
markdroth: It seems like we need the CSM plugin option implemented before we can remove the `CsmObservability` API.  And once we have the CSM plugin option implemented, the `CsmObservability` code can use that plugin option in the public API, so it won't need to reach inside the internal API anymore.  Or am I missing something here?

yashykt: I see what you mean. I wasn't originally planning on changing the implementation of `CsmObservability` to use the plugin option, but I like that idea. I will be able to remove the internal API earlier. Will do that. Thanks!

src/cpp/ext/otel/otel_plugin.cc
yashykt: seems to be working. Replaced all

include/grpcpp/ext/otel_plugin.h
```diff
@@ -67,8 +53,27 @@ OpenTelemetryServerCallRcvdTotalCompressedMessageSizeInstrumentName();
 /// grpc.server.call.rcvd_total_compressed_message_size
```
markdroth, line 53: Probably no need to list the metrics in the comment now that the constants are in the header file directly.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -61,26 +61,24 @@ absl::string_view OpenTelemetryTargetKey() { return "grpc.target"; }
 namespace {
 absl::flat_hash_set<std::string> BaseMetrics() {
```
markdroth, line 62: Does this need to be `std::string` instead of `absl::string_view`?

markdroth (commented):
Just a couple of lingering comments -- feel free to address or ignore.

src/cpp/ext/otel/otel_plugin.cc
yashykt: I want to intentionally store the set as `std::string` instead of `absl::string_view`. The case I'm worried about is when in the future we allow users to provide the set of metric names that they want enabled and they provide non-constant strings

include/grpcpp/ext/otel_plugin.h
yashykt: done

PR #35371, opened 2023-12-21
[CSM] Add support for GCE resources

yashykt:
Add support for GCE resources in CSM Observability.

Additionally, fix a bug where we were not adding the remote workload's canonical service label for unknown resource types.

Also, if zone and region are both specified, zone takes precedence.


src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -226,17 +229,34 @@ class MeshLabelsIterable : public LabelsIterable {
...
     }
-    // Only handle GKE type for now.
+    if (pos_ == local_labels_size + 2) {
+      return std::make_pair(
+          kPeerCanonicalServiceAttribute,
+          GetStringValueFromUpbStruct(struct_pb.struct_pb,
+                                      kMetadataExchangeCanonicalServiceKey,
+                                      struct_pb.arena.ptr()));
+    }
     switch (remote_type_) {
```
markdroth, line 239: I think this could be written more clearly using some helper functions, so that we can use the same code for the fixed labels and for any remote type:

```
absl::Span<const RemoteAttribute> GetAttributesForType(GcpResourceType remote_type) {
  switch (remote_type) {
    case GcpResourceType::kGke:
      return kGkeAttributeList;
    case GcpResourceType::kGce:
      return kGceAttributeList;
    default:
      return {};
  }
}

static constexpr std::array<RemoteAttribute, 2> kFixedAttributes = {
    RemoteAttribute{kPeerTypeAttribute, kMetadataExchangeTypeKey},
    RemoteAttribute{kPeerCanonicalServiceAttribute,
                    kPeerCanonicalServiceAttribute},
};

absl::optional<std::pair<absl::string_view, absl::string_view>> NextFromAttributeList(
    const StructPb& struct_pb, absl::Span<const RemoteAttribute> attributes, size_t start_index) {
  const size_t index = pos_ - start_index;
  if (index >= attributes.size()) return absl::nullopt;
  ++pos_;
  const auto& attribute = attributes[index];
  return std::make_pair(
      attribute.otel_attribute,
      GetStringValueFromUpbStruct(
          struct_pb.struct_pb,
          attribute.metadata_attribute,
          struct_pb.arena.ptr()));  
}
```

Then the code here can handle basically everything except local labels without duplication:

```
const size_t fixed_attribute_end = local_labels_size + kFixedAttributes.size();
if (pos_ < fixed_attribute_end) {
  return NextFromAttributeList(struct_pb, kFixedAttributes, local_labels_size);
}
auto attributes = GetAttributesForType(remote_type_);
return NextFromAttributeList(struct_pb, attributes, fixed_attribute_end);
```

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -226,17 +229,34 @@ class MeshLabelsIterable : public LabelsIterable {
                                 struct_pb.struct_pb, kMetadataExchangeTypeKey,
                                 struct_pb.arena.ptr()));
     }
-    // Only handle GKE type for now.
+    if (pos_ == local_labels_size + 2) {
+      return std::make_pair(
```
markdroth, line 233: If the compiler will accept it, I suggest just saying `{x, y}` instead of `std::make_pair(x, y)`.

Same thing throughout this method.

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -274,16 +294,21 @@ class MeshLabelsIterable : public LabelsIterable {
...
-                   kMetadataExchangeWorkloadNameKey},
-      GkeAttribute{kPeerNamespaceNameAttribute,
-                   kMetadataExchangeNamespaceNameKey},
-      GkeAttribute{kPeerClusterNameAttribute, kMetadataExchangeClusterNameKey},
-      GkeAttribute{kPeerLocationAttribute, kMetadataExchangeLocationKey},
-      GkeAttribute{kPeerProjectIdAttribute, kMetadataExchangeProjectIdKey},
-      GkeAttribute{kPeerCanonicalServiceAttribute,
-                   kMetadataExchangeCanonicalServiceKey},
+  static constexpr std::array<RemoteAttribute, 5> kGkeAttributeList = {
+      RemoteAttribute{kPeerWorkloadNameAttribute,
```
markdroth, line 298: Do you need to say `RemoteAttribute` here, or can you just use braces and let the compiler figure it out?

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -226,17 +229,34 @@ class MeshLabelsIterable : public LabelsIterable {
                                 struct_pb.struct_pb, kMetadataExchangeTypeKey,
                                 struct_pb.arena.ptr()));
     }
-    // Only handle GKE type for now.
+    if (pos_ == local_labels_size + 2) {
+      return std::make_pair(
+          kPeerCanonicalServiceAttribute,
```
markdroth, line 234: I believe that you need to change the `Size()` method as well to account for these changes (both the canonical service label and the GCE labels).

Please make sure the tests catch this problem, so that we don't miss it on future changes.

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -364,13 +391,13 @@ ServiceMeshLabelsInjector::ServiceMeshLabelsInjector(
       opentelemetry::sdk::resource::SemanticConventions::kK8sNamespaceName);
   absl::string_view cluster_name_value = GetStringValueFromAttributeMap(
       map, opentelemetry::sdk::resource::SemanticConventions::kK8sClusterName);
-  absl::string_view cluster_location_value = GetStringValueFromAttributeMap(
+  absl::string_view location_value = GetStringValueFromAttributeMap(
       map, opentelemetry::sdk::resource::SemanticConventions::
-               kCloudRegion);  // if regional
-  if (cluster_location_value == "unknown") {
-    cluster_location_value = GetStringValueFromAttributeMap(
+               kCloudAvailabilityZone);  // if zonal
```
markdroth, line 396: Looks like zone is taking precedence over region now.  Is that switch intentional?

src/cpp/ext/csm/metadata_exchange.cc
yashykt: Doesn't work without it :(

src/cpp/ext/csm/metadata_exchange.cc
yashykt: Yeah, if everything works normally, only one of `zone`/`region` would be set, and the precedence wouldn't matter, but my reading from internal threads said that if for some reason both `zone` and `region` are set, then `zone` is more specific than `region`.

src/cpp/ext/csm/metadata_exchange.cc
yashykt: No, doesn't work. I believe the issue is that the return type is `absl::optional<std::pair<**>>` and not just `std::pair<**>` which the compiler is not fine auto deducing when returning `{**}`

src/cpp/ext/csm/metadata_exchange.cc
yashykt: Yeah, you are right. `Size()` is a weird function which is not used by OpenTelemetry, and hence it doesn't fail, but given that it might end up being used in the future, we should make sure that it remains correct. 

With your suggested change, `Size()` no longer needs to be changed in the future :)

src/cpp/ext/csm/metadata_exchange.cc
yashykt: Suggestion accepted. Thanks!

src/cpp/ext/csm/metadata_exchange.cc
yashykt: Thinking of a test that might be useful here

yashykt: Maybe I'll add a test that verifies the functioning of the `MeshLabelsIterable` class itself in isolation

markdroth: The code looks good now.  It would still be good to add a test, since if we ever add a new category of labels, we'd still need to remember to update the `Size()` method.  But given that we're blocking the release, feel free to defer that to a separate PR.

markdroth (approved):

src/cpp/ext/csm/metadata_exchange.cc
yashykt: I think I got a decent idea. Gonna send you a follow-up PR

PR #35377, opened 2023-12-21
Add java patches 1.58.1 and 1.60.1 to client_matrix.py

larry-safran:
java release 1.58.1 was missed and just released 1.60.1


larry-safran:
Did this through UI, will redo through command line

PR #35384, opened 2023-12-22
Performance: allow to configure BigQuery project

lepistone:
Continue supporting the current grpc-testing that I suppose is used
inside of Google, but also allow to configure a different project to
upload results to.

The format "project_id.dataset_id.table_id" is common for BigQuery so it
seems idiomatic to do it in this way. Adding a separate command line
option would be more complicated because it would require changes all
the way down the chain (at least in the entry point for the test driver
and in the LoadTest controller).

### DIFF
--- a/tools/run_tests/performance/bq_upload_result.py
+++ b/tools/run_tests/performance/bq_upload_result.py
@@ -29,10 +29,12 @@ gcp_utils_dir = os.path.abspath(
 sys.path.append(gcp_utils_dir)
 import big_query_utils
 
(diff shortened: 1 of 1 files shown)


lepistone:
I think the failed check is because a maintainer needs to label the PR. Can you please give a look? Thank you!

yashykt:
@paulosjca can you take a look please?

paulosjca (approved):

yashykt (approved):

PR #35385, opened 2023-12-24
Fixes for C++20 in promise_like.h

mattpaletta:
The GRPC commit: https://github.com/grpc/grpc/commit/45aecbe3555c1da36cfbfaff0622646923d5137c does not build in C++20 because it relies on `std::result_of`, which was removed in C++20.  I believe the intention was to use the Abseil-defined implementation as the include for it was already there.

This MR is intended to be squashed.


markdroth:
I don't think we should be using an absl internal like this. @veblush

mattpaletta:
Looks like it's been fixed elsewhere. Closing.

Issue #35391, opened 2023-12-26
Failure in distribtest_python with Bazel 7


veblush:
### What version of gRPC and what language are you using?

gRPC HEAD

### What operating system (Linux, Windows,...) and version?

Linux

### What runtime / compiler are you using (e.g. python version or version of gcc)

Bazel 7

### What did you do?

bazel test //tools/bazelify_tests/test:bazel_distribtest_7.0.0_distribtest_python

### What did you expect to see?

Test pass

### What did you see instead?

Test failure with the following log ([full log](https://btx-internal.corp.google.com/invocations/053c045d-5d0c-44c1-8d5a-2525baedfeff/targets/%2F%2Ftools%2Fbazelify_tests%2Ftest:bazel_distribtest_7.0.0_distribtest_python;config=7d4ce66f87adb1c57986d96cd1fe88e6071dec5ffe33127d11d170a64d49b595/log))

```
'external/some_other_repo/proto/rnal/some_other_repo/proto/my_pb2.py' was not created
ERROR: /b/f/w/_tmp/bb723e14ec6b476e87942a3726f9e044/_bazel_anon/5abca540aa9d78e1243ad3a42569afac/external/some_other_repo/proto/BUILD:17:14: output 'external/some_other_repo/proto/rnal/some_other_repo/proto/my_pb2.pyi' was not created
ERROR: /b/f/w/_tmp/bb723e14ec6b476e87942a3726f9e044/_bazel_anon/5abca540aa9d78e1243ad3a42569afac/external/some_other_repo/proto/BUILD:17:14: ProtocInvocation external/some_other_repo/proto/rnal/some_other_repo/proto/my_pb2.py failed: not all outputs were created or valid
INFO: From ProtoCompile external/com_google_protobuf/python/google/protobuf/descriptor_pb2.py:
external/com_google_protobuf/.: warning: directory does not exist.
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 113.150s, Critical Path: 30.48s
INFO: 2419 processes: 1224 internal, 1195 linux-sandbox.
ERROR: Build did NOT complete successfully
```

veblush:
@gnossen & @XuanWang-Amos I disabled this test for now with Bazel 7. Would you take a look at this? It might be an easy fix.

XuanWang-Amos:
Checking...

Looks like some issue with building proto library:
[...]

PR #35392, opened 2023-12-26
Fix Bazel 7 build (more)

veblush:
It turned out that the previous change missed two things which this PR has

- Fix function `_dockerized_genrule` not to have `timeout` and `flaky` which Bazel doesn't have. (Bazel 7 may either drop these arguments or become more strict about passing unrecognized ones)
- Disabled python bazel distribution tests with Bazel 7. This needs to be addressed by https://github.com/grpc/grpc/issues/35391.


eugeneo (approved):

veblush:
Merged forcefully as copybara won't handle this.

Issue #22089, opened 2020-02-20
No event engine could be initialized from epoll1


simPod:
### What version of gRPC and what language are you using?
PHP 7.3.14 with grpc ext 1.27.0

### What operating system (Linux, Windows,...) and version?

macos catalina 

### What did you do?

I set 
```
grpc.enable_fork_support = 1
grpc.poll_strategy = epoll1
```

which gives 

```
php -i
E0220 12:47:37.882918000 4551400896 ev_posix.cc:230]                   No event engine could be initialized from epoll1
[1]    27272 abort      php -i
```

What am I doing wrong?

769344359:
https://github.com/grpc/grpc/blob/6dde2f43f70e10759bff06e09207d48f66beff36/doc/core/grpc-polling-engines.md#other-polling-engine-implementations-poll-and-windows-polling-engine

may be it do not support in max 
try poiling-engine

stanley-cheung:
Unfortunately these 2 flags 
```
grpc.enable_fork_support = 1
grpc.poll_strategy = epoll1
```
are not supported in MacOS. PHP fork() supported is limited to linux and CLI.

aakarim:
The patch here seems to solve this problem and is used in e.g. [Nix](https://github.com/NixOS/nixpkgs/blob/release-24.11/pkgs/development/php-packages/grpc/skip-darwin-test.patch). Could that not be merged in?

PR #35403, opened 2023-12-28
php: skip epoll1 test on darwin

shyim:
See https://github.com/grpc/grpc/issues/22089


shyim:
I can't add labels to make mergeable happy

PR #35404, opened 2023-12-28
[php]: add with-grpc-dir to configure and pkg-config support

shyim:
This allows me to link against system installed grpc and save a lot of build time from ~20 minutes to just 5 seconds.

See https://github.com/NixOS/nixpkgs/pull/277292 for usage

### DIFF
--- a/src/php/ext/grpc/config.m4
+++ b/src/php/ext/grpc/config.m4
@@ -13,29 +13,10 @@ if test "$PHP_TESTS" != "no"; then
 fi
 
 if test "$PHP_GRPC" != "no"; then
-  dnl Write more examples of tests here...
+  PKG_CHECK_MODULES([GRPC], [grpc])
 
-  dnl # --with-grpc -> check with-path
-  SEARCH_PATH="/usr/local /usr"     # you might want to change this
-  SEARCH_FOR="include/grpc/grpc.h"  # you most likely want to change this
-  if test -r $PHP_GRPC/$SEARCH_FOR; then # path given as parameter
-    GRPC_DIR=$PHP_GRPC
-  else # search default path list
-    AC_MSG_CHECKING([for grpc files in default path])
-    for i in $SEARCH_PATH ; do
-      if test -r $i/$SEARCH_FOR; then
-        GRPC_DIR=$i
-        AC_MSG_RESULT(found in $i)
-      fi
-    done
-  fi
-  if test -z "$GRPC_DIR"; then
-    AC_MSG_RESULT([not found])
-    AC_MSG_ERROR([Please reinstall the grpc distribution])
-  fi
-
-  dnl # --with-grpc -> add include path
-  PHP_ADD_INCLUDE($GRPC_DIR/include)
+  PHP_EVAL_INCLINE($GRPC_CFLAGS)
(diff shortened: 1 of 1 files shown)


shyim:
Do we need for the PHP extension the exact same GRPC version as the extension? Otherwise, we could add here a detection with pkg-config.

shyim:
I added now also pkg-config support to find the grpc version.

jtojnar:
> Do we need for the PHP extension the exact same GRPC version as the extension? Otherwise, we could add here a detection with pkg-config.

That is already not guaranteed when looking for the header in `/usr` so nothing special.

src/php/ext/grpc/config.m4
```diff
@@ -7,35 +7,51 @@ PHP_ARG_ENABLE(coverage, whether to include code coverage symbols,
...
+
+    GRPC_VERSION=`$PKG_CONFIG grpc --modversion`
+		AC_MSG_RESULT([found version $GRPC_VERSION])
+
+    PHP_GRPC_LIBS=`$PKG_CONFIG grpc --libs`
+		PHP_GRPC_INCS=`$PKG_CONFIG grpc --cflags`
+
+    PHP_EVAL_LIBLINE($PHP_GRPC_LIBS, AMQP_SHARED_LIBADD)
+    PHP_EVAL_INCLINE($PHP_GRPC_INCS)
+  else
```
jtojnar, line 36: Why not use PHP pkg-config macros?

https://github.com/php/php-src/commit/cc9c5d80570c0a3184017a921099b674f5e119e9

src/php/ext/grpc/config.m4
```diff
@@ -7,35 +7,17 @@ PHP_ARG_ENABLE(coverage, whether to include code coverage symbols,
...
-    done
-  fi
-  if test -z "$GRPC_DIR"; then
-    AC_MSG_RESULT([not found])
-    AC_MSG_ERROR([Please reinstall the grpc distribution])
-  fi
-
-  dnl # --with-grpc -> add include path
-  PHP_ADD_INCLUDE($GRPC_DIR/include)
+  AC_PATH_PROG(PKG_CONFIG, pkg-config, no)
```
jtojnar, line 20: what is this for now?

daynis-olman:
This PR is the only solid fix i was able to locate for the issue where the gRpc takes almost 45 minutes to install on Docker / EC2 LAMP Stack environments (PHP 8.1+). Is there any reason this PR is not being merged for almost 2 years now?

sameeraksc:
@jtojnar Can you please check ? we are wasting 30 mins to install :-(

ste93cry:
I’d really love to see this resolved, either by the author addressing the outstanding comments or by the maintainers applying the final touches before merging. At work we have a project that takes an extremely long time to build; compiling gRPC and the Protobuf extension alone can take nearly an hour. The build process also consumes a huge amount of resources. I haven’t yet investigated why or how to optimize the build process (I’m sure there are ways to reduce CPU and memory usage), but having this fixed here would still make a big difference.

shyim:
@pawbhard sorry for the ping, but i fixed the CI/CD. Could you trigger it again?

PR #35407, opened 2023-12-28
Use absl::string_view instead of self-implemented grpc::string_ref

georgthegreat:


markdroth:
Unfortunately, this would be an API breaking change for us, because `grpc::string_ref` has the `starts_with()` and `ends_with()` methods that are not present on `absl::string_view`.

georgthegreat:
absl::string_view provides these methods: https://github.com/abseil/abseil-cpp/blob/4038192a57cb75f7ee671f81a3378ff4c74c4f8e/absl/strings/string_view.h#L605

I wonder why they have been put under ifdef though.

markdroth:
Interesting!  I'm pretty sure it wasn't there the last time we looked at this.  It looks like this was added relatively recently by @derekmauro in https://github.com/abseil/abseil-cpp/commit/0f110600fcd84bd99e559eaf88b8ab5b4c32cce0.

Derek, is it intentional that the `absl::string_view` `starts_with()` and `ends_with()` methods are not enabled by default?  gRPC is still using C++14, and we currently have a `grpc::string_ref` class that we'd ideally like to replace with `absl::string_view`, but we can't do that due to the presence of these two methods on `grpc::string_ref`.  If these methods were enabled on `absl::string_view` by default, then I think that would allow us to do this replacement.  Is there any way we can make that happen?

markdroth:
Okay, thanks, that makes sense.  And I assume it would also not be okay to just use the absl implementation instead of aliasing to `std::string_view` until C++20, presumably because there are existing C++17 users who are expecting `absl::string_view` and `std::string_view` to be completely interchangeable?

derekmauro:
Yes. It seems like you would have to wait until C++20 to make this change.

markdroth:
Okay, thanks for all the info!

georgthegreat:
@derekmauro, can we give a bit more consideration to this?

First of all, it looks like there is very little reason to have `absl::string_view` (standing for a standalone interpretation) to provide starts_with / ends_with at all! The only combination these methods will be available is:
`ABSL_OPTION_USE_STD_STRING_VIEW = 0` and `ABSL_INTERNAL_CPLUSPLUS_LANG >= 202002L` which is a pretty weird combination. Why would someone having C++20 want to fallback to a standalone implementation of string_view?

@markdroth, will you consider approving a PR which uses inheritance from `absl::string_view`?

derekmauro:
> First of all, it looks like there is very little reason to have absl::string_view (standing for a standalone interpretation) to provide starts_with / ends_with at all! The only combination these methods will be available is:
ABSL_OPTION_USE_STD_STRING_VIEW = 0 and ABSL_INTERNAL_CPLUSPLUS_LANG >= 202002L which is a pretty weird combination. Why would someone having C++20 want to fallback to a standalone implementation of string_view?

Oh, there's a story to that.

Have you ever noticed that `absl::string_view` can be constructed from a `char*` that is `nullptr`? That's because `string_view` evolved from an internal Google type called `StringPiece` that accepted `nullptr` arguments. Except when C++ standardized it, that behavior was rejected (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0903r2.pdf). I think the reasons for rejecting it were sound.

This essentially left us with a very large codebase where we could not adopt `std::string_view`, hence we are still using `absl::string_view` in C++20. It was someone's job to clean this up, but it hasn't happened yet.

Unfortunately this won't be reconsidered.

derekmauro:
> > @markdroth, will you consider approving a PR which uses inheritance from `absl::string_view`?
> 
> @derekmauro Is it legal to inherit from an absl type? I don't see anything on https://abseil.io/about/compatibility that explicitly prohibits it, but it's unusual enough that I want to make sure it's okay.
> 
> If absl has no problem with it, we can think about whether there are any issues on the gRPC side.

You should consider this type to be a `std` type. We didn't mark it `final`. I would advise against inheriting from it though. There is little to gain and you are just asking for code to be broken in subtle ways.

markdroth:
Yeah, I agree, it does feel like it could cause problems down the road.  I don't think we want to go in that direction.

georgthegreat:
Can we inherit from std::string_view then?

I was unable to find any reference to the minimum c++ standard that is required to build grpc.

georgthegreat:
Well, it looks like there is no solution indeed.

PR #35412, opened 2023-12-29
[Python Bazel] Allow overwriting grpc_library in python_rules.bzl

mering:
For example it can be replaced with grpcio from PyPi to significantly speed up the initial build.

### DIFF
--- a/bazel/python_rules.bzl
+++ b/bazel/python_rules.bzl
@@ -222,11 +222,11 @@ def _generate_pb2_grpc_src_impl(context):
     py_info = _merge_pyinfos(
         [
             p,
-            context.attr._grpc_library[PyInfo],
+            context.attr.grpc_library[PyInfo],
         ] + [dep[PyInfo] for dep in context.attr.py_deps],
     )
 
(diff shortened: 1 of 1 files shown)


gnossen:
Thanks for the contribution!

In principle, I think this idea is fine -- being able to use the version of `grpcio` on PyPi seems fine, if a bit against the spirit of Bazel. One thing we'll definitely want, however, is test coverage for this new public field. You can add that [here](https://github.com/grpc/grpc/blob/master/test/distrib/bazel/python/BUILD).

We'll also want to make this new field explicit by adding it to the argument list of [the public macro](https://github.com/grpc/grpc/blob/f31b1695dfe3a529e981dd1f6f373019caa7c226/bazel/python_rules.bzl#L277).

gnossen:
@mering Is this still required?

mering:
@gnossen Thanks for getting back to this. I found a workaround which removed the immediate need for us. We might still want this in the future in order to reduce initial build time for our customers using our SDK in the future. I will close it for now and will re-open in case we need it in the future.

mering:
@sloretz FYI

PR #35433, opened 2024-01-03
ssl: More comprehensive testing of SSL credentials, part 1.

matthewstevenson88:
First of several PRs to improve the e2e testing for the SSL credentials API.


test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+
+  void ExpectNonResumedSession(
+      std::shared_ptr<const AuthContext> auth_context) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_SSL_SESSION_REUSED_PROPERTY);
+    ASSERT_EQ(properties.size(), 1u);
+    EXPECT_EQ("false", ToString(properties[0]));
+  }
+
+  void ExpectOk(const absl::Status& status) {
```
gtcooke94, line 180: Do you think this is cleaner than `EXPECT_TRUE(status.ok())`? It feels like we're adding an extra step

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+
+  void ExpectTransportSecurityLevel(
+      std::shared_ptr<const AuthContext> auth_context) {
+    std::vector<grpc::string_ref> properties = auth_context->FindPropertyValues(
+        GRPC_TRANSPORT_SECURITY_LEVEL_PROPERTY_NAME);
+    ASSERT_EQ(properties.size(), 1u);
+    EXPECT_EQ(ToString(properties[0]), "TSI_PRIVACY_AND_INTEGRITY");
+  }
+
+  void ExpectPeerCommonName(std::shared_ptr<const AuthContext> auth_context,
```
gtcooke94, line 201: For convention, these `Expect*` that take the auth context and an expected value, consider naming them `Expect*Equals` or something like that to match fns like `EXPECT_EQ`

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+  void ExpectPeerSubjectAltName(std::shared_ptr<const AuthContext> auth_context,
+                                const absl::flat_hash_set<std::string>& sans) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_X509_SAN_PROPERTY_NAME);
+    ASSERT_EQ(properties.size(), sans.size());
+    absl::flat_hash_set<std::string> observed_sans;
+    for (size_t i = 0; i < properties.size(); ++i) {
+      observed_sans.insert(ToString(properties[i]));
+    }
+    EXPECT_EQ(sans, observed_sans);
```
gtcooke94, line 226: nit: reverse the order to be consistent in this test file

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+  void ExpectPeerDnsSans(std::shared_ptr<const AuthContext> auth_context,
+                         const absl::flat_hash_set<std::string>& dns_sans) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_PEER_DNS_PROPERTY_NAME);
+    ASSERT_EQ(properties.size(), dns_sans.size());
+    absl::flat_hash_set<std::string> observed_sans;
+    for (size_t i = 0; i < properties.size(); ++i) {
+      observed_sans.insert(ToString(properties[i]));
+    }
+    EXPECT_EQ(dns_sans, observed_sans);
```
gtcooke94, line 256: nit: reverse the order to be consistent in this test file

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+  void ExpectPeerUriSans(std::shared_ptr<const AuthContext> auth_context,
+                         const absl::flat_hash_set<std::string>& uri_sans) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_PEER_URI_PROPERTY_NAME);
+    ASSERT_EQ(properties.size(), uri_sans.size());
+    absl::flat_hash_set<std::string> observed_sans;
+    for (size_t i = 0; i < properties.size(); ++i) {
+      observed_sans.insert(ToString(properties[i]));
+    }
+    EXPECT_EQ(uri_sans, observed_sans);
```
gtcooke94, line 270: nit: reverse the order to be consistent in this test file (looks like all the `sans` related ones are reversed from the others)

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+
+  auto resumed_handshake_context = DoRpc(ssl_options, cache);
+  ExpectOk(resumed_handshake_context.status());
+  ExpectResumedSession(*resumed_handshake_context);
+  ExpectTransportSecurityType(*resumed_handshake_context);
+  ExpectTransportSecurityLevel(*resumed_handshake_context);
+  ExpectPeerCommonName(*resumed_handshake_context, "*.test.google.com");
+  ExpectPeerSubject(
+      *resumed_handshake_context,
+      ""
```
gtcooke94, line 399: Does this line do anything?

gtcooke94 (commented):
This is looking really good, I'm liking the refactor to using params

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+      channel_args.SetPointer(std::string(GRPC_SSL_SESSION_CACHE_ARG), cache);
+    }
+    if (override_ssl_target_name) {
+      channel_args.SetSslTargetNameOverride("foo.test.google.fr");
+    }
+
+    auto creds = grpc::SslCredentials(ssl_options);
+    SetTlsVersion(GetParam().tls_version, creds.get());
+    std::shared_ptr<Channel> channel = grpc::CreateCustomChannel(
+        server_addr_, grpc::SslCredentials(ssl_options), channel_args);
```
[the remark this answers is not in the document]
matthewstevenson88: Good catch, done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+  void ExpectPeerEmailSans(std::shared_ptr<const AuthContext> auth_context,
+                           const absl::flat_hash_set<std::string>& email_sans) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_PEER_EMAIL_PROPERTY_NAME);
+    ASSERT_EQ(properties.size(), email_sans.size());
+    absl::flat_hash_set<std::string> observed_sans;
+    for (size_t i = 0; i < properties.size(); ++i) {
+      observed_sans.insert(ToString(properties[i]));
+    }
+    EXPECT_EQ(email_sans, observed_sans);
```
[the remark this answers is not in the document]
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,301 @@ class SslCredentialsTest : public ::testing::Test {
...
+  void ExpectPeerIpSans(std::shared_ptr<const AuthContext> auth_context,
+                        const absl::flat_hash_set<std::string>& ip_sans) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_PEER_IP_PROPERTY_NAME);
+    ASSERT_EQ(properties.size(), ip_sans.size());
+    absl::flat_hash_set<std::string> observed_sans;
+    for (size_t i = 0; i < properties.size(); ++i) {
+      observed_sans.insert(ToString(properties[i]));
+    }
+    EXPECT_EQ(ip_sans, observed_sans);
```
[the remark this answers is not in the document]
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Not sure what clang format did there...Fixed. )

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1140,6 +1140,10 @@ void tsi_ssl_session_cache_unref(tsi_ssl_session_cache* cache) {
   tsi::SslSessionLRUCache::FromC(cache)->Unref();
 }
 
+size_t tsi_ssl_session_cache_size(tsi_ssl_session_cache* cache) {
+  return tsi::SslSessionLRUCache::FromC(cache)->Size();
```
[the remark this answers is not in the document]
matthewstevenson88: At present this is only used in testing, agreed. However, this is IMO a very natural API to add to the current list. For example, if we intend to clean up the resumption stack during the de-experimentalization work, then likely adding logs for the state/size of the session cache will be a useful part of that, and it would be used there. This is a loosely held opinion so please push back if you don't feel that this is the right place for this API.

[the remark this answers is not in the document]
gtcooke94: I'm good with it then - at first I also thought it might be doing a copy in `FromC` which would be an inefficient helper for size, but it looks like it's just a static pointer cast under the hood

Feel free to resolve if this all sounds good to you

test/cpp/end2end/ssl_credentials_test.cc
gtcooke94: Good point, this makes a failure more diagnosable

gtcooke94 (approved):
LGTM

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,252 @@ class SslCredentialsTest : public ::testing::Test {
...
+    grpc::Status result = stub->Echo(&context, request, &response);
+    if (!result.ok()) {
+      return absl::Status(static_cast<absl::StatusCode>(result.error_code()),
+                          result.error_message());
+    }
+    EXPECT_EQ(response.message(), kMessage);
+    return context.auth_context();
+  }
+
+  void ExpectResumedSession(std::shared_ptr<const AuthContext> auth_context) {
```
markdroth, line 165: Passing in a `shared_ptr` takes an unnecessary ref.  Suggest passing this in as `const AuthContext&` instead.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,252 @@ class SslCredentialsTest : public ::testing::Test {
...
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_SSL_SESSION_REUSED_PROPERTY);
+    ASSERT_EQ(properties.size(), 1u);
```
markdroth, line 168: This pattern of putting expectations in helper functions is sub-optimal, because when tests fail, the error message will point to the helper function, not to the actual test where the problem is occurring.  Instead, I suggest structuring this as helper functions to extract the data you want, but leave the actual expectations for the value of that data in the test.  For example:

```
static absl::string_view GetAuthContextProperty(const AuthContext&, const char* property) {
  std::vector<grpc::string_ref> properties =
      auth_context.FindPropertyValues(property);
  if (properties.size() != 1) return "";
  return absl::string_view(properties[0].data(), properties[0].size());
}
```

Then in the test, you can say something like this:

```
EXPECT_EQ(GetAuthContextProperty(*auth_context, GRPC_SSL_SESSION_REUSED_PROPERTY),
          "false");
```

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,252 @@ class SslCredentialsTest : public ::testing::Test {
...
+
+  void ExpectNonResumedSession(
+      std::shared_ptr<const AuthContext> auth_context) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_SSL_SESSION_REUSED_PROPERTY);
+    ASSERT_EQ(properties.size(), 1u);
+    EXPECT_EQ("false", ToString(properties[0]));
+  }
+
+  void ExpectOk(const absl::Status& status) {
```
markdroth, line 180: I think you'll get a better error message by just saying `EXPECT_EQ(status, absl::OkStatus())`.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -85,51 +131,252 @@ class SslCredentialsTest : public ::testing::Test {
...
+
+  void ExpectPeerSubjectEquals(std::shared_ptr<const AuthContext> auth_context,
+                               absl::string_view subject) {
+    std::vector<grpc::string_ref> properties =
+        auth_context->FindPropertyValues(GRPC_X509_SUBJECT_PROPERTY_NAME);
+    ASSERT_EQ(properties.size(), 1u);
+    EXPECT_EQ(ToString(properties[0]), subject);
+  }
+
+  void ExpectPeerSubjectAltNameEquals(
```
markdroth, line 218: Instead of having this method, you can do the following in the individual tests:

```
absl::string_view subject_alt_name =
    GetAuthContextProperty(*auth_context, GRPC_X509_SAN_PROPERTY_NAME);
EXPECT_THAT(
    absl::StrSplit(subject_alt_name, ","),
    ::testing::UnorderedElementsAre(
        "*.test.google.fr", "waterzooi.test.google.be",
        "*.test.youtube.com", "192.168.1.3"));
```

Same thing for the other SAN methods below.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Good idea, done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Good point! Done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Good call. I adjusted this slightly and created a separate `GetAuthContextPropertyAsSet` function, because the `GetAuthContextProperty` function sometimes returns strings (e.g. the common name) that contain commas.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -162,153 +164,23 @@ class SslCredentialsTest : public ::testing::TestWithParam<SslOptions> {
...
-    ASSERT_EQ(properties.size(), 1u);
-    EXPECT_EQ("true", ToString(properties[0]));
+        auth_context.FindPropertyValues(property);
+    if (properties.size() != 1) return "";
+    return absl::string_view(properties[0].data(), properties[0].size());
   }
 
-  void ExpectNonResumedSession(
-      std::shared_ptr<const AuthContext> auth_context) {
+  static absl::flat_hash_set<std::string> GetAuthContextPropertyAsSet(
```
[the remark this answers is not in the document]
matthewstevenson88: Thanks Mark and apologies for the long-delayed reply.

Agreed with your general approach, a few minor changes in execution.

- Changed to return a list.
- Changed to avoid duplicating functionality, except the implementation turned out to be simpler if `GetAuthContextPropertyAsList` has the logic to parse out the properties from the auth context and then we can call this function to implement `GetAuthContextProperty`.
- Done RE `ToString`.
- Done RE `absl::string_view`.
- Re using `absl::StrSplit`: no longer applies since I inverted the order of which function calls which.

matthewstevenson88:
@markdroth Would you mind taking another look at this PR when you have a sec? Thanks!

matthewstevenson88:
@markdroth Friendly ping :)

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -77,96 +129,297 @@ class SslCredentialsTest : public ::testing::Test {
...
+    grpc_channel_credentials* c_creds = grpc_ssl_credentials_create(
+        options.pem_root_certs.empty() ? nullptr
+                                       : options.pem_root_certs.c_str(),
+        options.pem_private_key.empty() ? nullptr : &pem_key_cert_pair, nullptr,
+        nullptr);
+    grpc_ssl_credentials* ssl_creds =
+        reinterpret_cast<grpc_ssl_credentials*>(c_creds);
+    ssl_creds->set_min_tls_version(GetParam().tls_version);
+    ssl_creds->set_max_tls_version(GetParam().tls_version);
+    auto creds = std::make_shared<WrappedChannelCredentials>(ssl_creds);
```
markdroth, line 154: Why is this now using the C-core API instead of the public C++ API?

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -15,11 +15,17 @@
...
+#include "src/core/lib/security/credentials/ssl/ssl_credentials.h"
+
 #include <memory>
+#include <string>
+#include <thread>
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "absl/container/flat_hash_set.h"
```
markdroth, line 27: I don't think these two includes are actually needed.

test/cpp/end2end/BUILD
```diff
@@ -1077,6 +1078,8 @@ grpc_cc_test(
         "//src/core/tsi/test_creds:server1.pem",
     ],
     external_deps = [
+        "absl/container:flat_hash_set",
```
markdroth, line 1081: I don't think these two deps are needed.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -29,34 +35,80 @@
...
+
 constexpr char kCaCertPath[] = "src/core/tsi/test_creds/ca.pem";
 constexpr char kServerCertPath[] = "src/core/tsi/test_creds/server1.pem";
 constexpr char kServerKeyPath[] = "src/core/tsi/test_creds/server1.key";
 constexpr char kClientCertPath[] = "src/core/tsi/test_creds/client.pem";
 constexpr char kClientKeyPath[] = "src/core/tsi/test_creds/client.key";
 constexpr char kMessage[] = "Hello";
 
-class SslCredentialsTest : public ::testing::Test {
+struct ToStringJoinFormatter {
```
markdroth, line 78: Doesn't look like this is used anywhere.

test/cpp/end2end/ssl_credentials_test.cc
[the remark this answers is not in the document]
markdroth: This approach looks good!  I think I was misreading the code earlier when I suggested `StrSplit()` -- we weren't actually splitting a string but rather returning a list when the property was present more than once.  Sorry for the confusion.

markdroth (commented):
Just one noteworthy comment, which is about why we're no longer using the public C++ API.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -77,96 +129,297 @@ class SslCredentialsTest : public ::testing::Test {
...
 
+  absl::StatusOr<std::shared_ptr<const AuthContext>> DoRpc(
+      const SslCredentialsOptions& options, grpc_ssl_session_cache* cache,
+      bool override_ssl_target_name = true) {
+    ChannelArguments channel_args;
+    if (GetParam().use_session_cache) {
+      channel_args.SetPointer(std::string(GRPC_SSL_SESSION_CACHE_ARG), cache);
+    }
+    if (override_ssl_target_name) {
+      channel_args.SetSslTargetNameOverride("foo.test.google.fr");
```
gtcooke94, line 140: nit: consider pulling this magic value into a named constant for clarity

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -176,30 +429,169 @@ TEST_F(SslCredentialsTest, ConcurrentResumption) {
...
+  server_addr_ = absl::StrCat("localhost:",
+                              std::to_string(grpc_pick_unused_port_or_die()));
+  absl::Notification notification;
+  server_thread_ = new std::thread([&]() {
+    std::string root_cert = GetFileContents(kServerCertPath);
+    RunServer(&notification, root_cert);
+  });
+  notification.WaitForNotification();
+
+  std::string root_cert = GetFileContents(kClientCertPath);
```
gtcooke94, line 508: Should this be `kCaCertPath`?

gtcooke94 (approved):
Looks good with one question

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: The test needs to be parametrized on the TLS version to get better coverage, but the decision was made a while ago to not expose the min/max TLS version setters for the C++-layer SSL credentials (only for TLS credentials).

If you'd prefer, I'm happy to use the public C++ API in the TLS 1.3 case, and only revert to the C-core API in the TLS 1.2 case.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Done.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Good catch, removed.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Good catch, fixed.

test/cpp/end2end/BUILD
matthewstevenson88: Good catch, removed.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Good catch, removed.

test/cpp/end2end/BUILD
```diff
@@ -1078,7 +1078,6 @@ grpc_cc_test(
         "//src/core/tsi/test_creds:server1.pem",
     ],
     external_deps = [
-        "absl/container:flat_hash_set",
         "absl/strings",
```
markdroth, line 1081: This dep is not needed either.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -24,7 +24,6 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
-#include "absl/container/flat_hash_set.h"
 #include "absl/strings/str_split.h"
```
markdroth, line 27: This include is not needed either.

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -523,8 +517,7 @@ TEST_P(SslCredentialsTest, ClientCertificateIsUntrusted) {
           grpc_ssl_client_certificate_request_type::
               GRPC_SSL_REQUEST_AND_REQUIRE_CLIENT_CERTIFICATE_AND_VERIFY) {
     EXPECT_EQ(auth_context.status().code(), absl::StatusCode::kUnavailable);
-    EXPECT_THAT(auth_context.status().message(),
-                HasSubstr("CERTIFICATE_VERIFY_FAILED"));
+    EXPECT_FALSE(auth_context.status().message().empty());
```
markdroth, line 520: Why was this changed to not check the status message?  In general, it's better to check the full text of status messages in tests, since having useful messages for users is important for debuggability, so we do actually want tests to check that we're emitting useful messages.

test/cpp/end2end/ssl_credentials_test.cc
markdroth: This is an e2e test for the C++ SSL creds API.  If a piece of functionality isn't exposed in that API, why are we testing it here?  It seems like if you want to test the underlying C-core API, that should be done in a C-core test instead.

matthewstevenson88: Totally fair. Switched.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Greg noticed a small mistake in the test setup, which changed the error message. Specifically, the new client-side error message is (correctly) something that doesn't have any TLS-specific info it, it's just something like "failed to connect to all addresses: ... : Socket closed". Changed this back to check that the "Socket closed" substring is present (rather than checking the full error message, which I'm not confident will be stable).

markdroth: Why does the status message not include any TLS-specific info?  I thought I'd fixed that in #30077.  If not, then we still have some work to do here.

Hmm, is the problem here that the server's verification of the client's cert doesn't happen until after the handshake is done?  If so, that's really unfortunate, because this error message doesn't tell users anything about what the actual problem is.

matthewstevenson88: Your guess is correct: The server verifies the client cert after the client-side thinks the TLS handshake has finished, so the server-side logs this error (e.g. something with "CERTIFICATE_VERIFY_FAILED ...") and then the server closes the connection, so all that the client sees is that the connection was closed.

I agree that it's an unfortunate situation (other TLS stacks, e.g. Golang's crypto/tls, have the same issue). One idea would be to add some state to the "Socket closed" message (e.g. "Socket closed after TLS handshake completed and before received any application data from server"). Interested to hear your thoughts on this.

markdroth: Okay.  I don't see an immediate way to solve this, so let's not block this PR on it.  But please add a TODO here describing the problem and noting that we need to ultimately find a way to improve this error message for users.

markdroth (approved):

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -176,30 +403,165 @@ TEST_F(SslCredentialsTest, ConcurrentResumption) {
...
+  server_addr_ = absl::StrCat("localhost:",
+                              std::to_string(grpc_pick_unused_port_or_die()));
+  absl::Notification notification;
+  server_thread_ = new std::thread([&]() {
+    std::string root_cert = GetFileContents(kCaCertPath);
+    RunServer(&notification, root_cert);
+  });
+  notification.WaitForNotification();
+
+  std::string root_cert = GetFileContents(kClientCertPath);
```
gtcooke94, line 447: This one too

test/cpp/end2end/ssl_credentials_test.cc
```diff
@@ -176,30 +403,165 @@ TEST_F(SslCredentialsTest, ConcurrentResumption) {
...
+  if (GetParam().request_type == grpc_ssl_client_certificate_request_type::
+                                     GRPC_SSL_DONT_REQUEST_CLIENT_CERTIFICATE) {
+    return;
+  }
+
+  server_addr_ = absl::StrCat("localhost:",
+                              std::to_string(grpc_pick_unused_port_or_die()));
+  absl::Notification notification;
+  server_thread_ = new std::thread([&]() {
+    std::string root_cert = GetFileContents(kServerCertPath);
```
gtcooke94, line 477: Maybe this one? I think it might be okay because this test is wanting the server root to not trust the client cert, so you just use what should be the servers cert as the root, which then won't chain to the Client's cert?

If that's the case can you add a comment explaining this?

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: This is correct, because you don't want the client to trust the server's certificate. Added a comment to clarify.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: This is correct, because you don't want the server to trust the client's certificate. Added a comment to clarify.

test/cpp/end2end/ssl_credentials_test.cc
matthewstevenson88: Done.

PR #35434, opened 2024-01-03
[OTel] Add a PluginOption API

yashykt:


src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -148,6 +167,11 @@ class OpenTelemetryServerCallTracer : public grpc_core::ServerCallTracer {
   grpc_core::Slice path_;
   std::unique_ptr<LabelsIterable> injected_labels_;
   bool registered_method_;
+  std::vector<bool>
```
markdroth, line 170: Same comment as elsewhere: We can probably save memory by using a bitmask here.

src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -56,7 +56,14 @@ namespace {
 
 class OpenTelemetryServerCallTracer : public grpc_core::ServerCallTracer {
  public:
-  OpenTelemetryServerCallTracer() : start_time_(absl::Now()) {}
+  explicit OpenTelemetryServerCallTracer(const grpc_core::ChannelArgs& args)
+      : start_time_(absl::Now()) {
+    for (const auto& plugin_option :
```
markdroth, line 61: Instead of duplicating all of this code to determine the active set of options and then iterate over them, I wonder if it could be refactored into a common view object.  I'm thinking something like this:

```
class ActivePluginOptionsView {
 public:
  explicit ActivePluginOptionsView(absl::string_view target) {
    for (size_t i = 0; i < OpenTelemetryPluginState().plugin_options.size(); ++i) {
      const auto& plugin_option = OpenTelemetryPluginState().plugin_options[i];
      if (plugin_option != nullptr &&
          plugin_option->IsActiveOnClientChannel(target)) {
        active_mask_.set(i);
      }
    }
  }

  void ForEach(absl::FunctionRef<void(const OpenTelemetryPluginOption&)> func) const {
    for (size_t i = 0; i < OpenTelemetryPluginState().plugin_options.size(); ++i) {
      const auto& plugin_option = OpenTelemetryPluginState().plugin_options[i];
      if (active_mask_.is_set(i)) func(*plugin_option);
    }
  }

 private:
  BitSet<64> active_mask_;
};
```

You can instantiate that object in both the server and client code, and then you can call it using something like this:

```
active_plugin_options_view_.ForEach(
    [&](const OpenTelemetryPluginOption& plugin_option) {
      auto* labels_injector = plugin_option.labels_injector();
      if (labels_injector != nullptr) {
        labels_injector->AddLabels(send_initial_metadata, injected_labels_from_plugin_options_[i].get());
      }
    });
```

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -653,6 +654,273 @@ TEST_F(OpenTelemetryPluginEnd2EndTest,
...
+  }
+
+ private:
+  bool enabled_on_client_;
+  bool enabled_on_server_;
+  std::unique_ptr<CustomLabelInjector> label_injector_;
+};
+
+TEST_F(OpenTelemetryPluginOptionEnd2EndTest, Basic) {
+  std::vector<std::unique_ptr<grpc::experimental::OpenTelemetryPluginOption>>
```
markdroth, line 736: Could write this as:

```
std::vector<std::unique_ptr<grpc::experimental::OpenTelemetryPluginOption>>
    plugin_option_list = {
      std::make_unique<CustomPluginOption>(
          /*enabled_on_client=*/true, /*enabled_on_server=*/true,
          std::make_pair("key", "value")));
    };
```

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -653,6 +654,273 @@ TEST_F(OpenTelemetryPluginEnd2EndTest,
...
+  const auto& client_attributes =
+      data["grpc.client.attempt.duration"][0].attributes.GetAttributes();
+  EXPECT_EQ(client_attributes.size(), 4);
+  EXPECT_EQ(absl::get<std::string>(client_attributes.at("key")), "value");
+  // Verify server side metric
+  ASSERT_EQ(data["grpc.server.call.duration"].size(), 1);
+  const auto& server_attributes =
+      data["grpc.server.call.duration"][0].attributes.GetAttributes();
+  EXPECT_EQ(server_attributes.size(), 2);
+  EXPECT_EQ(server_attributes.find("key"), server_attributes.end());
```
markdroth, line 816: Please write this as:

```
EXPECT_THAT(server_attributes, ::testing::Not(::testing::Contains("key")));
```

Same thing throughout.

markdroth (commented):
This looks good overall!

Comments are mostly minor.  The only really significant one is the one about using `unique_ptr<>` in the public API.

include/grpcpp/ext/otel_plugin.h
```diff
@@ -95,6 +98,11 @@ class OpenTelemetryPluginBuilder {
   OpenTelemetryPluginBuilder& SetGenericMethodAttributeFilter(
       absl::AnyInvocable<bool(absl::string_view /*generic_method*/) const>
           generic_method_attribute_filter);
+  /// Add a plugin option to add to the opentelemetry plugin being built. At
+  /// present, this type is an opaque type. Ownership of \a option is
```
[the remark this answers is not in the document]
yashykt: Yeah, you are right, this is better. This makes ownership obvious instead of requiring documentation.

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: done

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: The issue with this is that initializer_list is always copied and std::unique_ptr can not be copied. Using `emplace_back`/`push_back` is the only thing i've gotten to work.

src/cpp/ext/otel/otel_server_call_tracer.cc
yashykt: Suggestion taken with some modifications.. I think the total number of lines is higher than before but I like it since we factored out the common code

src/cpp/ext/otel/otel_server_call_tracer.cc
yashykt: Suggestion taken

include/grpcpp/ext/csm_observability.h
```diff
@@ -88,6 +88,18 @@ class CsmObservabilityBuilder {
...
+
+/// Creates an OpenTelemetryPluginOption that would add additional labels on
+/// gRPC metrics to enhance observability for CSM users.
+///
+/// Sample Usage -
+/// OpenTelemetryPluginBuilder()
+///     .SetMeterProvider(provider)
+///     .AddPluginOption(MakeCsmOpenTelemetryPluginOption())
+///     .BuildAndRegisterGlobal();
+OpenTelemetryPluginOption* MakeCsmOpenTelemetryPluginOption();
```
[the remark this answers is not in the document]
yashykt: done

src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -57,13 +57,11 @@ namespace {
...
-         OpenTelemetryPluginState().plugin_options) {
-      active_plugin_options_.push_back(plugin_option != nullptr &&
-                                       plugin_option->IsActiveOnServer(args));
-    }
-  }
+      : start_time_(absl::Now()),
+        active_plugin_options_view_(
+            ActivePluginOptionsView::MakeForServer(args)),
+        injected_labels_from_plugin_options_(
+            OpenTelemetryPluginState().plugin_options.size()) {}
```
markdroth, line 64: This could be wasteful if there are a large number of plugin options and only a small subset of them are active.

Note that since the channel args should be the same for all calls on a server, we should be able to determine the number (and set of) active options for the connection and not have to do it per-call.  But I'm guessing we can't actually do that until we implement the rest of the stats plugin infrastructure, so maybe just add a TODO for now?

(Technically, the channel args could actually be different for each individual connection on a given server.  But if we only match against the top-level channel args for the server once when the server starts up and then use the resulting set of active options for all calls on all connections, we can ignore those differences.)

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -733,8 +730,9 @@ class CustomPluginOption
 };
 
 TEST_F(OpenTelemetryPluginOptionEnd2EndTest, Basic) {
-  std::vector<std::unique_ptr<grpc::experimental::OpenTelemetryPluginOption>>
-      plugin_option_list;
+  std::vector<
+      std::unique_ptr<grpc::internal::InternalOpenTelemetryPluginOption>>
+      plugin_option_list(1);
```
markdroth, line 735: Passing a size to the ctor is the equivalent of calling `resize()`, not `reserve()` -- it actually adds that number of default-constructed elements to the vector.  So what you're doing here is constructing a vector with 2 elements, the first of which is set to null.  I don't think that's what you intended.

Same for all tests.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -813,12 +812,15 @@ TEST_F(OpenTelemetryPluginOptionEnd2EndTest, ClientOnlyPluginOption) {
   const auto& server_attributes =
       data["grpc.server.call.duration"][0].attributes.GetAttributes();
   EXPECT_EQ(server_attributes.size(), 2);
-  EXPECT_EQ(server_attributes.find("key"), server_attributes.end());
+  EXPECT_THAT(
+      server_attributes,
+      ::testing::Not(::testing::Contains(std::make_pair("key", "value"))));
```
markdroth, line 817: We want to check that the key is not present, even if it has a different value.  Please change this to `::testing::Contains(::testing::Key("key"))`.

Same thing throughout.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -849,19 +851,22 @@ TEST_F(OpenTelemetryPluginOptionEnd2EndTest, ServerOnlyPluginOption) {
...
+      attributes,
+      ::testing::Not(::testing::Contains(std::make_pair("key", "value"))));
   // Verify server side metric
   ASSERT_EQ(data["grpc.server.call.duration"].size(), 1);
   const auto& server_attributes =
       data["grpc.server.call.duration"][0].attributes.GetAttributes();
   EXPECT_EQ(server_attributes.size(), 3);
   EXPECT_EQ(absl::get<std::string>(server_attributes.at("key")), "value");
-}
+}  // namespace
```
markdroth, line 863: This is the end of a test, not the end of a namespace.

markdroth (approved):
This looks really good!  Remaining comments are relatively minor.

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: Of course! Don't know what I was thinking! Thanks!

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: Thanks!

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: Done

include/grpcpp/ext/otel_plugin.h
```diff
@@ -102,7 +105,7 @@ class OpenTelemetryPluginBuilder {
   /// present, this type is an opaque type. Ownership of \a option is
   /// transferred when `AddPluginOption` is invoked.
   OpenTelemetryPluginBuilder& AddPluginOption(
```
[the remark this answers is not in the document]
yashykt: Done

src/cpp/ext/otel/otel_server_call_tracer.cc
yashykt: Yeah, exactly. TODO added

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -203,10 +200,52 @@ class OpenTelemetryPluginBuilderImpl {
...
+    return ActivePluginOptionsView(
+        [target](const InternalOpenTelemetryPluginOption& plugin_option) {
+          return plugin_option.IsActiveOnClientChannel(target);
+        });
+  }
+
+  static ActivePluginOptionsView MakeForServer(
+      const grpc_core::ChannelArgs& args) {
+    return ActivePluginOptionsView(
+        [args](const InternalOpenTelemetryPluginOption& plugin_option) {
```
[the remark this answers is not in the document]
yashykt: Ofcourse! Good catch!

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -203,10 +200,52 @@ class OpenTelemetryPluginBuilderImpl {
       generic_method_attribute_filter_;
   absl::AnyInvocable<bool(const grpc_core::ChannelArgs& /*args*/) const>
       server_selector_;
-  std::vector<std::unique_ptr<experimental::OpenTelemetryPluginOption>>
+  std::vector<std::unique_ptr<InternalOpenTelemetryPluginOption>>
       plugin_options_;
 };
 
+class ActivePluginOptionsView {
```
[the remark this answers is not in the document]
yashykt: Done

PR #35440, opened 2024-01-03
[test-infra] Update all Ubuntu docker images to v20.04 or newer

drfloob:


tools/dockerfile/push_testing_images.sh
```diff
@@ -86,8 +86,8 @@ fi
 for DOCKERFILE_DIR in "${ALL_DOCKERFILE_DIRS[@]}"
 do
   # Generate image name based on Dockerfile checksum. That works well as long
-  # as can count on dockerfiles being written in a way that changing the logical
```
eugeneo, line 89: nit: I suggest reverting this file

eugeneo (approved):
LGTM if the tests pass...

tools/dockerfile/push_testing_images.sh
drfloob: Done. Also, the test failures are all unrelated.

Issue #35368, opened 2023-12-20
gRPC does not disable TLS renegotiation, as required by HTTP/2


davidben:
TLS has a very problematic feature, renegotiation, which causes a single connection to run multiple handshakes  This causes a lot of problems and is incompatible with multiplexed protocols like HTTP/2. Indeed HTTP/2 is quite clear on this:

> A deployment of HTTP/2 over TLS 1.2 MUST disable renegotiation. 

https://www.rfc-editor.org/rfc/rfc9113.html#section-9.2.1

When built against BoringSSL, gRPC is fine. BoringSSL disables renegotiation by default. When built against OpenSSL, however, gRPC is exposed to renegotiations and I suspect has never tested any of those codepaths. The `SSL_OP_NO_RENEGOTIATION` may be used to turn renegotiation off. You all should set it if you aren't planning on testing renegotiation.

drfloob:
Pinging @gtcooke94 on this one, since I'm not steeped in gRPC+SSL. I do see logic in SSL read/write code that explicitly logs an "unsupported" failure if a peer tries to renegotiate an SSL connection.  https://github.com/grpc/grpc/blob/cdec4fa61cf08c4420ad3a1ccb209771e9afeb8a/src/core/tsi/ssl_transport_security_utils.cc#L74-L76

davidben:
That does not disable renegotiation. Renegotiation is a scenario that can cause `SSL_write` to block on transport read and vice versa, but detecting that does is not sufficient to disable renegotiation. Rather, disabling renegotiation is a requirement for you all to not handle that case. Having not disabled renegotiation, that code is also broken.

See the issue description for what you all would actually need to do to rejected renegotiation.

PR #35444, opened 2024-01-03
Disable renegotiation

gtcooke94:
Disable renegotiation for TLS
This option was added in OpenSSL 1.1.1, thus why it is inside of the include guards.

https://github.com/grpc/grpc/issues/35368


matthewstevenson88:
I think the issue linked in the top comment is incorrect. #35368 has the correct rationale for why this change is being made.

matthewstevenson88 (approved):

gtcooke94:
Thanks for catching that, looks like I accidentally backspaced a number

matthewstevenson88 (approved):

src/core/tsi/ssl_transport_security.cc
```diff
@@ -2074,6 +2074,9 @@ tsi_result tsi_create_ssl_client_handshaker_factory_with_options(
   ssl_context = SSL_CTX_new(TLS_method());
 #else
   ssl_context = SSL_CTX_new(TLSv1_2_method());
+#endif
+#if OPENSSL_VERSION_NUMBER >= 0x10101000
```
davidben, line 2078: Optional: `#if SSL_OP_NO_RENEGOTIATION` would work too

davidben (approved):

PR #35455, opened 2024-01-04
[Python AIO] Change default code() return to StatusCode.OK instead of None

XuanWang-Amos:

### DIFF
--- a/src/python/grpcio/grpc/_cython/_cygrpc/aio/server.pyx.pxi
+++ b/src/python/grpcio/grpc/_cython/_cygrpc/aio/server.pyx.pxi
@@ -17,6 +17,7 @@ import inspect
 import traceback
 import functools
 
+import grpc
 
 cdef int _EMPTY_FLAG = 0
 cdef str _RPC_FINISHED_DETAILS = 'RPC already finished.'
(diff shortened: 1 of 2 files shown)


gnossen:
I agree that returning `StatusCode.OK` is more correct here, but this is a backwards-incompatible change and it doesn't seem to be a clear-cut case where the docs said one thing and the implementation did another. Especially for the sync stack, it might not be worth it to change this behavior.

At the very least, I think we'll want to send out a request for comments on the grpc-io mailing list to make sure that this change wouldn't be disastrous.

XuanWang-Amos:
> I agree that returning `StatusCode.OK` is more correct here, but this is a backwards-incompatible change and it doesn't seem to be a clear-cut case where the docs said one thing and the implementation did another. Especially for the sync stack, it might not be worth it to change this behavior.
> 
> At the very least, I think we'll want to send out a request for comments on the grpc-io mailing list to make sure that this change wouldn't be disastrous.

Make sense, while working on change sync stack, I also noticed that it won't be an easy and clean change, we have multiple conditions depends on code to be `None` so we can't simply change the default value.

I'll put this PR on hold for now and investigate the user's level of interest in this change.

PR #35467, opened 2024-01-05
[Test] Fix bug in waiting for the Orca OOB report

eugeneo:
Make sure there is no unnecessary delays when there are multiple reports in the queue.

This change also adds a test for the custom LB policy.


test/cpp/interop/backend_metrics_lb_policy_test.cc
```diff
@@ -0,0 +1,143 @@
...
+  SimpleResponse res;
+  ASSERT_EQ(stub.UnaryCall(&ctx, req, &res).error_code(), grpc::OK);
+  // This report is sent on start, available immediately
+  auto report = tracker.WaitForOobLoadReport(
+      [](auto report) { return report.cpu_utilization() == 0.5; },
+      absl::Milliseconds(1), 3);
+  ASSERT_TRUE(report.has_value());
+  EXPECT_EQ(report->cpu_utilization(), 0.5);
+  for (size_t i = 0; i < 3; i++) {
+    // Wait for slightly more that 1 min
```
yijiem, line 123: This should be `// Wait for slightly more than 1 second` ?

yijiem (approved):

test/cpp/interop/backend_metrics_lb_policy_test.cc
eugeneo: Done. Thanks for pointing out!

PR #35480, opened 2024-01-08
[EventEngine] Enable the EventEngine DNS Resolver on Windows

drfloob:


drfloob:
This introduces a number of failures, including some in the resolver component test. @yijiem I'll leave this to you to land later this month. Note how I temporarily disabled the goaway_server test for EventEngine, since it's iomgr-specific. I did that to try to get this landed in short order to make it into the 1.61 release, but we should rewrite that to use a custom EventEngine DNS resolver for the 1.62 release. It should not be that much work.

PR #35484, opened 2024-01-09
[Build] Minimal changes for enabling DLL operation on Windows

dawidcha:

### DIFF
--- a/include/grpcpp/support/proto_buffer_reader.h
+++ b/include/grpcpp/support/proto_buffer_reader.h
@@ -148,7 +148,7 @@ class ProtoBufferReader : public grpc::protobuf::io::ZeroCopyInputStream {
             *slice(), GRPC_SLICE_LENGTH(*slice()) - backup_count(),
             GRPC_SLICE_LENGTH(*slice()) - backup_count() + count)));
       }
-      int64_t take = std::min(backup_count(), static_cast<int64_t>(count));
+      int64_t take = (std::min)(backup_count(), static_cast<int64_t>(count));
       set_backup_count(backup_count() - take);
       count -= take;
       if (count == 0) {
--- a/templates/CMakeLists.txt.template
+++ b/templates/CMakeLists.txt.template
@@ -29,6 +29,11 @@
 
   third_party_proto_prefixes = {lib.proto_prefix for lib in external_proto_libraries}
 
+  gpr_libs = ['gpr']
+  grpc_libs = ['grpc', 'grpc_unsecure']
+  grpcxx_libs = ['grpc++', 'grpc++_unsecure']
+  protoc_libs = ['benchmark_helpers', 'grpc++_reflection', 'grpcpp_channelz']
+
   def third_party_proto_import_path(path):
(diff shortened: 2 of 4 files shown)


dawidcha:
This is a follow up from https://github.com/grpc/grpc/pull/35248

I've retained the changes to the CMakeLists.txt template file which make the macro definitions for import/export more reliable, but it turns out no further annotations are needed to get this to compile if I fall back to the slightly less efficient non-explicit DLL binding for client code.

Note that as before these changes do not affect the way the non-Windows, non-shared builds work.

veblush (approved):

dawidcha:
@yashykt are you able to approve this change?

yashykt (approved):

include/grpcpp/support/proto_buffer_reader.h
```diff
@@ -148,7 +148,7 @@ class ProtoBufferReader : public grpc::protobuf::io::ZeroCopyInputStream {
             *slice(), GRPC_SLICE_LENGTH(*slice()) - backup_count(),
             GRPC_SLICE_LENGTH(*slice()) - backup_count() + count)));
       }
-      int64_t take = std::min(backup_count(), static_cast<int64_t>(count));
+      int64_t take = (std::min)(backup_count(), static_cast<int64_t>(count));
```
yashykt, line 151: why are there brackets around std::min?

dawidcha: brackets around std::min avoid issues where an older header that #define's a min 'function'

dawidcha: Looks like mergeable has been running for a couple of days now. Does it need restarting? Or do I need to rebase once more?

templates/CMakeLists.txt.template
```diff
@@ -250,12 +268,6 @@
       return '\n'.join(lines)
     return _func
 
-  def add_dll_annotation_macros(lib, prefix):
-    return """set_target_properties(%s PROPERTIES DEFINE_SYMBOL "%s_DLL_EXPORTS")
```
h-vetinari, line 254: Also, I'm not 100% sure how thing still work after removing
```
set_target_properties(%s PROPERTIES DEFINE_SYMBOL "%s_DLL_EXPORTS")
```
That had been intentional to make it less reliant on CMake magic (which creates `<target>_exports` automatically AFAIR), and more explicit overall. However, since the symbols being to switch between import/export in `port_platform.h` haven't been renamed, I'm less sure what's happening. Normally, I'd say the build would have to fail, but so far it compiled... 🤷

h-vetinari (commented):
Some questions after our windows [builds](https://github.com/conda-forge/grpc-cpp-feedstock/pull/352) for 1.62 broke.

Also, the following comment is out of date w.r.t. both points:
https://github.com/grpc/grpc/blob/d379dc24c5f6dc428da74ab6ce02cb2e5648caca/include/grpc/support/port_platform.h#L63-L66

PR #35488, opened 2024-01-09
[ruby] Fix use-after-free for post-fork channel recreation

XrXr:
The `grpc_channel_args` is retained on the Ruby object and used for recreating the channel after forking in
grpc_rb_channel_maybe_recreate_channel_after_fork(). Previously, the key for each argument was taken from a Ruby string directly, which could be invalidated if the Ruby string is modified or moved by the GC. Duplicate the string for the key instead, so we own it.

Reproducer in https://github.com/grpc/grpc/issues/35489


apolcyn:
Thank you for the repro and pull request.

I added essentially your repro, as a test in https://github.com/grpc/grpc/pull/35565

Can you please include the test changes from https://github.com/grpc/grpc/pull/35565 into this PR?

apolcyn (approved):

XrXr:
@apolcyn Thanks! I added the test changes.

src/ruby/end2end/fork_test_repro_35489.rb
```diff
@@ -0,0 +1,64 @@
...
+  to_parent_r, to_parent_w = IO.pipe
+  # Note gRPC has not yet been initialized, otherwise we would need to call prefork
+  # before spawn and postfork_parent after.
+  # TODO(apolcyn): consider redirecting server's stderr to a file
+  Process.spawn(RbConfig.ruby, echo_server_path, in: to_child_r, out: to_parent_w, err: "server_log")
+  to_child_r.close
+  to_parent_w.close
+  child_port = to_parent_r.gets.strip
+  STDERR.puts "server running on port: #{child_port}"
+  key = "grpc." * 100000 # large enough to malloc backing storage
```
apolcyn, line 47: style checker wants 100000 to be written as `100_000` can you please fix?

ditto line below

apolcyn (commented):
Linter complains:

src/ruby/end2end/fork_test_repro_35489.rb:1:1: W: Lint/ScriptPermission: Script file fork_test_repro_35489.rb doesn't have execute permission.
#!/usr/bin/env ruby
^^^^^^^^^^^^^^^^^^^

Can you please fix?

apolcyn (approved):

PR #35506, opened 2024-01-10
Fix deprecation warnings from upb 

ericsalo:
First attempt at a PR, which I am quite certain will fail a great many CI tests but hopefully we can fix them all easily enough.


ericsalo:
Closing for now, will return with a clean patch soon.

PR #35526, opened 2024-01-11
[CSM] De-experimentalize CSM OTel Plugin Option

yashykt:


include/grpcpp/ext/csm_observability.h
```diff
@@ -101,7 +103,6 @@ class CsmObservabilityBuilder {
 std::unique_ptr<grpc::OpenTelemetryPluginOption>
```
markdroth, line 103: Shouldn't need `grpc::` prefix here.

markdroth (approved):

include/grpcpp/ext/csm_observability.h
yashykt: Done

markdroth (approved):

PR #35527, opened 2024-01-11
Use MSBuildProjectExtension to detect language in Google.Protobuf.Tools.targets

wannabesrevenge:
Prior to this commit, $(Language) is used in Google.Protobuf.Tools.targets to identify whether a project is C# or not. The issue with this approach is that $(Language) is not set by default in MSBUILD. It is set in the Microsoft.CSharp.targets file here: https://github.com/dotnet/msbuild/blob/45f3aed1944548f123ce928d1253e38b5008dccf/src/Tasks/Microsoft.CSharp.CurrentVersion.targets#L38 . This causes the Google.Protobuf.Tools.targets to have a dependency on Microsoft.CSharp.targets file to be called first.

To improve on this situation, this PR seeks to remove this dependency by using the well-defined MSBUILD properties that are available at any time during the build process. Specifically, we can identify whether or not a project is a C# project by checking the project extension with the $(MSBuildProjectExtension) property. If $(MSBuildProjectExtension) == 'csproj', then we can say that this is a C# project.

### DIFF
--- a/src/csharp/Grpc.Tools/build/_protobuf/Google.Protobuf.Tools.targets
+++ b/src/csharp/Grpc.Tools/build/_protobuf/Google.Protobuf.Tools.targets
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
     <!-- We allow a non-C# generator be set by the user, but skip adding outputs to Compile in this case. -->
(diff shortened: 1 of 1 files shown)


wannabesrevenge:
There are additional locations that use `$(Language) == "C#` outside of my changes. I would be happy to update this PR to include those other locations.

tonydnewell:
See my comment on the issue https://github.com/grpc/grpc/issues/35540#issuecomment-1891763039

Grpc.Tools is working as intended.

PR #35535, opened 2024-01-12
[SOT] Show warnings to the release manager about the commits that do not have PR info (CL-first workflow)

yijiem:


tools/release/release_notes.py
```diff
@@ -86,6 +86,40 @@
 API_URL = "https://api.github.com/repos/grpc/grpc/pulls/"
 
 
+def print_commits_wo_pr(commits_wo_pr):
+    """Print commit and CL info for the commits that are submitted with CL-first workflow and warn the release manager to check manually."""
+    import re
```
drfloob, line 91: I think this falls under the Python style guide's "conditional imports" clause, since the modules would be lazily imported if the function were called. I don't think this is an exceptional case, so I'd recommend doing all imports at the top level.

tools/release/release_notes.py
```diff
@@ -86,6 +86,40 @@
 API_URL = "https://api.github.com/repos/grpc/grpc/pulls/"
 
 
+def print_commits_wo_pr(commits_wo_pr):
```
drfloob, line 89: This method is a no-op if `commits_wo_pr` is empty. Can we just not call the function instead of making the repeated `if commits_wo_pr` checks?

yijiem: Done.

tools/release/release_notes.py
yijiem: Done.

PR #35545, opened 2024-01-12
[PSM Interop] Add a payload to xds interop client when sending RPCs

stanley-cheung:
When testing CSM Observability, we discovered that the c++ xds interop client is not sending any payload with the `UnaryCall` RPCs so most of the metrics will have a value of 0.

Adding a payload to the xds interop client here.

We need this fix so that we can verify that the metrics are recording the right number of bytes being sent / received. So we need a non-trivial payload to be sent with the `UnaryCall` RPC between the xds interop client and server.

### DIFF
--- a/test/cpp/interop/xds_interop_client.cc
+++ b/test/cpp/interop/xds_interop_client.cc
@@ -67,8 +67,10 @@ ABSL_FLAG(int32_t, stats_port, 50052,
           "Port to expose peer distribution stats service.");
 ABSL_FLAG(std::string, rpc, "UnaryCall",
           "a comma separated list of rpc methods.");
-ABSL_FLAG(int32_t, request_payload_size, 0, "request payload size in bytes.");
-ABSL_FLAG(int32_t, response_payload_size, 0, "response payload size in bytes.");
+ABSL_FLAG(int32_t, request_payload_size, 0,
+          "request payload size in bytes of zeros.");
+ABSL_FLAG(int32_t, response_payload_size, 0,
+          "response payload size in bytes of zeros.");
 ABSL_FLAG(std::string, metadata, "", "metadata to send with the RPC.");
 ABSL_FLAG(std::string, expect_status, "OK",
           "RPC status for the test RPC to be considered successful");
@@ -157,10 +159,14 @@ class TestClient {
       }
     }
     SimpleRequest request;
-    request.set_response_size(absl::GetFlag(FLAGS_response_payload_size));
-    std::string payload(absl::GetFlag(FLAGS_request_payload_size), '\0');
-    request.mutable_payload()->set_body(
-        payload.c_str(), absl::GetFlag(FLAGS_request_payload_size));
+    if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
+      request.set_response_size(absl::GetFlag(FLAGS_response_payload_size));
+    }
+    if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
+      std::string payload(absl::GetFlag(FLAGS_request_payload_size), '\0');
+      request.mutable_payload()->set_body(
+          payload.c_str(), absl::GetFlag(FLAGS_request_payload_size));
+    }
     call->context.set_deadline(deadline);
     call->result.saved_request_id = saved_request_id;
     call->result.rpc_type = ClientConfigureRequest::UNARY_CALL;
--- a/test/cpp/interop/xds_interop_client.cc
+++ b/test/cpp/interop/xds_interop_client.cc
@@ -68,9 +68,11 @@ ABSL_FLAG(int32_t, stats_port, 50052,
 ABSL_FLAG(std::string, rpc, "UnaryCall",
           "a comma separated list of rpc methods.");
 ABSL_FLAG(int32_t, request_payload_size, 0,
-          "Set the SimpleRequest.payload.body to a string of repeated 0 (zero) ASCII characters of the given size in bytes.");
+          "Set the SimpleRequest.payload.body to a string of repeated 0 (zero) "
+          "ASCII characters of the given size in bytes.");
 ABSL_FLAG(int32_t, response_payload_size, 0,
-          "Ask the server to respond with SimpleResponse.payload.body of the given length (may not be implemented on the server).");
+          "Ask the server to respond with SimpleResponse.payload.body of the "
+          "given length (may not be implemented on the server).");
 ABSL_FLAG(std::string, metadata, "", "metadata to send with the RPC.");
 ABSL_FLAG(std::string, expect_status, "OK",
           "RPC status for the test RPC to be considered successful");
--- a/test/cpp/interop/xds_interop_client.cc
+++ b/test/cpp/interop/xds_interop_client.cc
@@ -121,6 +121,9 @@ struct RpcConfig {
   ClientConfigureRequest::RpcType type;
   std::vector<std::pair<std::string, std::string>> metadata;
   int timeout_sec = 0;
+  std::string request_payload;
+  int request_payload_size = 0;
+  int response_payload_size = 0;
 };
 struct RpcConfigurationsQueue {
   // A queue of RPC configurations detailing how RPCs should be sent.
@@ -159,13 +162,12 @@ class TestClient {
       }
     }
     SimpleRequest request;
-    if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
-      request.set_response_size(absl::GetFlag(FLAGS_response_payload_size));
+    if (config.response_payload_size > 0) {
+      request.set_response_size(config.response_payload_size);
     }
-    if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
-      std::string payload(absl::GetFlag(FLAGS_request_payload_size), '\0');
-      request.mutable_payload()->set_body(
-          payload.c_str(), absl::GetFlag(FLAGS_request_payload_size));
+    if (config.request_payload_size > 0) {
+      request.mutable_payload()->set_body(config.request_payload.c_str(),
+                                          config.request_payload_size);
     }
     call->context.set_deadline(deadline);
     call->result.saved_request_id = saved_request_id;
@@ -345,6 +347,15 @@ class XdsUpdateClientConfigureServiceImpl
       if (metadata_iter != metadata_map.end()) {
         config.metadata = metadata_iter->second;
       }
+      if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
+        config.request_payload_size = absl::GetFlag(FLAGS_request_payload_size);
+        std::string payload(config.request_payload_size, '\0');
+        config.request_payload = payload;
+      }
+      if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
+        config.response_payload_size =
+            absl::GetFlag(FLAGS_response_payload_size);
+      }
       configs.push_back(std::move(config));
     }
     {
@@ -482,6 +493,15 @@ void BuildRpcConfigsFromFlags(RpcConfigurationsQueue* rpc_configs_queue) {
     auto metadata_iter = metadata_map.find(config.type);
     if (metadata_iter != metadata_map.end()) {
       config.metadata = metadata_iter->second;
+    }
+    if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
+      config.request_payload_size = absl::GetFlag(FLAGS_request_payload_size);
+      std::string payload(config.request_payload_size, '\0');
+      config.request_payload = payload;
+    }
+    if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
+      config.response_payload_size =
+          absl::GetFlag(FLAGS_response_payload_size);
     }
     configs.push_back(std::move(config));
   }
--- a/test/cpp/interop/xds_interop_client.cc
+++ b/test/cpp/interop/xds_interop_client.cc
@@ -349,7 +349,7 @@ class XdsUpdateClientConfigureServiceImpl
       }
       if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
         config.request_payload_size = absl::GetFlag(FLAGS_request_payload_size);
-        std::string payload(config.request_payload_size, '\0');
+        std::string payload(config.request_payload_size, '0');
         config.request_payload = payload;
       }
       if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
@@ -496,7 +496,7 @@ void BuildRpcConfigsFromFlags(RpcConfigurationsQueue* rpc_configs_queue) {
     }
     if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
       config.request_payload_size = absl::GetFlag(FLAGS_request_payload_size);
-      std::string payload(config.request_payload_size, '\0');
+      std::string payload(config.request_payload_size, '0');
       config.request_payload = payload;
     }
     if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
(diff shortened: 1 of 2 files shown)


yashykt (approved):

stanley-cheung:
The framework changes to pass these parameters is [here](https://github.com/grpc/psm-interop/pull/20).

test/cpp/interop/xds_interop_client.cc
```diff
@@ -154,11 +156,16 @@ class TestClient {
             std::chrono::system_clock::now() + std::chrono::seconds(INT_MAX);
       }
     }
+    SimpleRequest request;
+    request.set_response_size(absl::GetFlag(FLAGS_response_payload_size));
+    std::string payload(absl::GetFlag(FLAGS_request_payload_size), '\0');
```
sergiitk, line 161: What's the contents of the payload here? Random memory contents? repeated `\0`s?

test/cpp/interop/xds_interop_client.cc
```diff
@@ -154,11 +156,16 @@ class TestClient {
...
+    request.mutable_payload()->set_body(
```
sergiitk, line 162: Since `Payload payload = 3;` is a message (therefore it's nullable, we should only set the `payload` when `request_payload_size` > 0 to avoid sending unnecessary data.

test/cpp/interop/xds_interop_client.cc
```diff
@@ -67,6 +67,8 @@ ABSL_FLAG(int32_t, stats_port, 50052,
           "Port to expose peer distribution stats service.");
 ABSL_FLAG(std::string, rpc, "UnaryCall",
           "a comma separated list of rpc methods.");
+ABSL_FLAG(int32_t, request_payload_size, 0, "request payload size in bytes.");
```
sergiitk, line 70: should we clarify what's the contents of the payload be?

sergiitk (changes requested):

test/cpp/interop/xds_interop_client.cc
```diff
@@ -67,6 +67,8 @@ ABSL_FLAG(int32_t, stats_port, 50052,
...
+ABSL_FLAG(int32_t, response_payload_size, 0, "response payload size in bytes.");
```
sergiitk, line 71: Ok, apparently this field results in an error right now: https://github.com/grpc/grpc/blob/e36025d3d24182d0c512b767a8e3687ae2fd9640/test/cpp/interop/interop_server.cc#L213C14-L216

Let's avoid it until it's implemented in the server.

sergiitk: Ooops, misread it. Deleting this thread.

test/cpp/interop/xds_interop_client.cc
stanley-cheung: Repeated `\0` for `request_payload_size` times.

stanley-cheung: Done the same way here in the regular `interop_client.cc` https://github.com/grpc/grpc/blob/master/test/cpp/interop/interop_client.cc#L250

test/cpp/interop/xds_interop_client.cc
stanley-cheung: seems fine to me, as the content isn't important as long as they get to the desired length. It's kind of documented here: https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#large_unary

` body: 271828 bytes of zeros `

stanley-cheung: Updated the comment

test/cpp/interop/xds_interop_client.cc
stanley-cheung: Fixed

test/cpp/interop/xds_interop_client.cc
sergiitk: I'm not to good with c, so I don't really know what this would do

test/cpp/interop/xds_interop_client.cc
sergiitk: I see. I'll give it another though.

sergiitk (commented):
Thanks for the changes. After looking at it, I think flag parse, and payload prep shouldn't happen in this `AsyncUnaryCall`.

stanley-cheung:
> Thanks for the changes. After looking at it, I think flag parse, and payload prep shouldn't happen in this `AsyncUnaryCall`.

There are other `absl::GetFlag` calls in these similar functions. @yashykt are they OK?

test/cpp/interop/xds_interop_client.cc
stanley-cheung: @yashykt Does this look OK?

stanley-cheung: > I'm not to good with c, so I don't really know what this would do

And just for reference, https://cplusplus.com/reference/string/string/string/ this is the 6th version of the constructor.

```
fill (6) | string (size_t n, char c);

(6) fill constructor
Fills the string with n consecutive copies of character c.
```

test/cpp/interop/xds_interop_client.cc
```diff
@@ -154,11 +158,20 @@ class TestClient {
             std::chrono::system_clock::now() + std::chrono::seconds(INT_MAX);
       }
     }
+    SimpleRequest request;
+    if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
+      request.set_response_size(absl::GetFlag(FLAGS_response_payload_size));
+    }
+    if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
+      std::string payload(absl::GetFlag(FLAGS_request_payload_size), '\0');
```
sergiitk, line 166: Wait, is this a `0` char? I thought `\0` is the null char, e.g. `0x0`.

test/cpp/interop/xds_interop_client.cc
```diff
@@ -154,11 +158,20 @@ class TestClient {
             std::chrono::system_clock::now() + std::chrono::seconds(INT_MAX);
       }
     }
+    SimpleRequest request;
+    if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
```
sergiitk, line 162: Confirmed with @yashykt. Flag parsing can be expensive, and `AsyncUnaryCall` method runs in a loop with as fast as the qps argument set.

Instead, we should parse flags early, do proper validation, and provide them to `AsyncUnaryCall` via `RpcConfig& config`. Let's add new fields `response_size` and `payload` (pointing to precreated payload body) to the `RpcConfig` struct.

sergiitk (changes requested):

test/cpp/interop/xds_interop_client.cc
sergiitk: Thanks!

test/cpp/interop/xds_interop_client.cc
sergiitk: https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#large_unary

OK, these specs are not for psm (xds) interop, but for the regular grpc interop tests. We don't even have `large_unary` test case.

XuanWang-Amos:
@sergiitk, do we need do the same for Python client?

sergiitk:
@XuanWang-Amos Eventually, but how soon is a question for Yash and Gina.

test/cpp/interop/xds_interop_client.cc
stanley-cheung: I left it as it for now. I looked at Java's implementation https://github.com/grpc/grpc-java/blob/master/interop-testing/src/main/java/io/grpc/testing/integration/AbstractInteropTest.java#L501 and it looks like it's the null character as well?

test/cpp/interop/xds_interop_client.cc
stanley-cheung: Fixed.

test/cpp/interop/xds_interop_client.cc
stanley-cheung: There are quite some inconsistencies among the existing interop client / server (not this Xds client / server) how we use the null char. But that doesn't seem to matter much at all, because only the length is being checked.

stanley-cheung: Chatted offline with @yashykt, setting this to the zero `0` character now.

yashykt:
> There are other absl::GetFlag calls in these similar functions.  are they OK?

Think of them as an optimization. We should ideally just parse those flags once during the lifetime of process.

yashykt:
> Eventually, but how soon is a question for Yash and Gina.

Let's get C++ working properly first. Once we know what all needs to be changed, we'll port them to other languages as well.

stanley-cheung:
> > There are other absl::GetFlag calls in these similar functions.  are they OK?
> 
> Think of them as an optimization. We should ideally just parse those flags once during the lifetime of process.

It's done in the latest commit.

test/cpp/interop/xds_interop_client.cc
```diff
@@ -470,6 +494,14 @@ void BuildRpcConfigsFromFlags(RpcConfigurationsQueue* rpc_configs_queue) {
     if (metadata_iter != metadata_map.end()) {
       config.metadata = metadata_iter->second;
     }
+    if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
+      config.request_payload_size = absl::GetFlag(FLAGS_request_payload_size);
+      std::string payload(config.request_payload_size, '0');
+      config.request_payload = payload;
+    }
+    if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
```
sergiitk, line 502: Instead, let's add validation that the flag value is not negative, and exit via `GPR_ASSERT(0);` it's not.

test/cpp/interop/xds_interop_client.cc
```diff
@@ -470,6 +494,14 @@ void BuildRpcConfigsFromFlags(RpcConfigurationsQueue* rpc_configs_queue) {
     if (metadata_iter != metadata_map.end()) {
       config.metadata = metadata_iter->second;
     }
+    if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
```
sergiitk, line 497: Please move the if block outside the `for` loop. Inside the for loop, print a warning log message when `request_payload_size` > 0, but an `EmptyCall` is requested.

test/cpp/interop/xds_interop_client.cc
```diff
@@ -332,6 +347,15 @@ class XdsUpdateClientConfigureServiceImpl
       if (metadata_iter != metadata_map.end()) {
         config.metadata = metadata_iter->second;
       }
+      if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
+        config.request_payload_size = absl::GetFlag(FLAGS_request_payload_size);
+        std::string payload(config.request_payload_size, '0');
+        config.request_payload = payload;
```
sergiitk, line 353: Please print a warning log message when `request_payload_size` > 0, but an `EmptyCall` is requested.

sergiitk (changes requested):
I'd like to add:
- Initial value validation when parsing flags: so that failed validation terminates the process
- Warning log on an incorrect config state: when a request_payload_size is > 0, but rpc_type is EmptyCall. Note that neither EmptyCall request nor response supports the payload body.

test/cpp/interop/xds_interop_client.cc
```diff
@@ -154,11 +161,19 @@ class TestClient {
             std::chrono::system_clock::now() + std::chrono::seconds(INT_MAX);
       }
     }
+    SimpleRequest request;
+    if (config.response_payload_size > 0) {
+      request.set_response_size(config.response_payload_size);
+    }
```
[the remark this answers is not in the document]
stanley-cheung: Done

test/cpp/interop/xds_interop_client.cc
stanley-cheung: Done

test/cpp/interop/xds_interop_client.cc
stanley-cheung: Done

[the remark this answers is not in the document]
stanley-cheung: Done

test/cpp/interop/xds_interop_client.cc
stanley-cheung: Done

test/cpp/interop/xds_interop_client.cc
```diff
@@ -332,6 +347,15 @@ class XdsUpdateClientConfigureServiceImpl
       if (metadata_iter != metadata_map.end()) {
         config.metadata = metadata_iter->second;
       }
+      if (absl::GetFlag(FLAGS_request_payload_size) > 0) {
```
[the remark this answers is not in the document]
stanley-cheung: Done

test/cpp/interop/xds_interop_client.cc
```diff
@@ -332,6 +347,15 @@ class XdsUpdateClientConfigureServiceImpl
...
+        config.request_payload_size = absl::GetFlag(FLAGS_request_payload_size);
+        std::string payload(config.request_payload_size, '0');
+        config.request_payload = payload;
+      }
+      if (absl::GetFlag(FLAGS_response_payload_size) > 0) {
```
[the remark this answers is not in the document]
stanley-cheung: The logic and structure is slightly different. Can we do this optimization next time?

test/cpp/interop/xds_interop_client.cc
```diff
@@ -67,6 +67,10 @@ ABSL_FLAG(int32_t, stats_port, 50052,
           "Port to expose peer distribution stats service.");
 ABSL_FLAG(std::string, rpc, "UnaryCall",
           "a comma separated list of rpc methods.");
+ABSL_FLAG(int32_t, request_payload_size, 0,
+          "request payload size in bytes of zeros.");
+ABSL_FLAG(int32_t, response_payload_size, 0,
+          "response payload size in bytes of zeros.");
```
sergiitk, line 73:
```suggestion
          "Ask the server to respond with SimpleResponse.payload.body of the given length (may not be implemented on the server).");
```

sergiitk (commented):
Looks good - just one question

test/cpp/interop/xds_interop_client.cc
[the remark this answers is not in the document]
sergiitk: Looks good in the last revision, no need to change!

test/cpp/interop/xds_interop_client.cc
```diff
@@ -470,6 +509,22 @@ void BuildRpcConfigsFromFlags(RpcConfigurationsQueue* rpc_configs_queue) {
     if (metadata_iter != metadata_map.end()) {
       config.metadata = metadata_iter->second;
     }
+    if (request_payload_size > 0 &&
+        config.type == ClientConfigureRequest::EMPTY_CALL) {
+      gpr_log(GPR_DEBUG,
```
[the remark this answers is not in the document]
sergiitk: I see your other comment, `GPR_WARNING` doesn't exist. Let's make it `GPR_ERROR` then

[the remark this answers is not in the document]
sergiitk: Yep, all looks good with GRP_ERROR:
```
❯ ./bazel-bin/test/cpp/interop/xds_interop_client --request_payload_size=1 --rpc="EmptyCall"
E0125 19:14:57.554702027   26043 xds_interop_client.cc:514]            request_payload_size should not be set for EMPTY_CALL
```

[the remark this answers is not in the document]
stanley-cheung: apparently there's no `GPR_WARNING`, only `DEBUG` or `INFO` or `ERROR`. https://grpc.github.io/grpc/cpp/support_2log_8h.html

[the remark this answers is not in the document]
stanley-cheung: Changed to `GPR_ERROR`. Thanks.

stanley-cheung:
All 4 of these jobs passed!

sergiitk (approved):

PR #35565, opened 2024-01-16
[ruby] Fix SEGV after fork due to use-after-free

apolcyn:
Fix https://github.com/grpc/grpc/issues/35489

Channel args are re-created when using a channel post-fork. So we need to maintain ownership of keys too.


apolcyn:
https://github.com/grpc/grpc/pull/35488/files instead

PR #35594, opened 2024-01-18
[ruby] [v1.61.x] Build/test ruby 3.3 and build native gems with Ruby 3.3 support

chadlwilson:
Cherry-pick of #35399 for `1.61` since release branch appears to have been cut for pre-release builds.

It'd be good to land this for `1.61` since Ruby 3.3 has been out for a few weeks and this is one of the most popular native gems which is blocking some testing/upgrades by users who can't/don't want to build from source.

I read `CONTRIBUTING.md` but not entirely sure if this is needed, or whether the team have a habit of merging from master selectively prior to release. Let me know if this PR is unnecessary for consideration for `1.61` release.

---
Updates ruby-compiler-dock to 1.4.0 which brings Ruby 3.3 final support per https://github.com/rake-compiler/rake-compiler-dock/releases/tag/1.4.0 and starts cross-compiling for ruby 3.3.

As noted in #31991 and https://github.com/grpc/grpc/pull/35399#issuecomment-1886117733 the tests currently don't run on all ruby versions, so this is still an existing gap which this PR does not close, but also doesn't appear to make worse.

FYI @alto-ruby @apolcyn

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -144,7 +144,7 @@ task 'gem:native', [:plat] do |t, args|
   verbose = ENV['V'] || '0'
 
   grpc_config = ENV['GRPC_CONFIG'] || 'opt'
-  ruby_cc_versions = ['3.2.0', '3.1.0', '3.0.0', '2.7.0'].join(':')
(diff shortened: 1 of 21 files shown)


chadlwilson:
Done in #35614

PR #34966, opened 2023-11-16
bazel: rename `@com_github_cncf_udpa` to `@com_github_cncf_xds`

sergiitk:
The repo has been renamed from cncf/udpa to cncf/xds long time ago, this is just to reflect this in the bazel repo name. Otherwise this is a noop.

Similar change to Envoy: https://github.com/envoyproxy/envoy/pull/30802.

PR #35603, opened 2024-01-19
[build] Update `envoy-api` and `xds` dependencies

eugeneo:
Closes: #34966


tools/buildgen/extract_metadata_from_bazel_xml.py
```diff
@@ -88,6 +88,9 @@ def __init__(
     "com_github_cncf_udpa": ExternalProtoLibrary(
```
markdroth, line 88: Why is this still needed?  What breaks if this is removed?

Since we've removed `com_github_cncf_udpa` everywhere else, I don't see why we need to retain it here.

tools/buildgen/extract_metadata_from_bazel_xml.py
```diff
@@ -575,6 +578,7 @@ def _expand_upb_proto_library_rules(bazel_rules):
         ("@com_google_protobuf//", "src/"),
         ("@com_google_googleapis//", ""),
         ("@com_github_cncf_udpa//", ""),
```
markdroth, line 580: Same question here: Why is this still needed?

tools/distrib/python/xds_protos/build.py
```diff
@@ -65,22 +61,20 @@ def localize_path(p):
...
-def add_test_import(
-    proto_package_path: str, file_name: str, service: bool = False
-):
+def add_test_import(proto_package_path: str, file_name: str, service: bool = False):
     TEST_IMPORTS.append(
         "from %s import %s\n"
         % (
-            proto_package_path.replace("/", "."),
-            file_name.replace(".proto", "_pb2"),
+            proto_package_path.replace("/", ".").replace("-", "_"),
```
markdroth, line 68: In what cases does the proto package path include a `-` character?  I don't think that should happen.

markdroth (commented):
Overall, looks great!  Just a few questions that I'd like to understand.  Thanks!

tools/distrib/python/xds_protos/build.py
eugeneo: There was some message from envoy, like `test-data` - I do not have that CI log. @gnossen wrote this change.

tools/buildgen/extract_metadata_from_bazel_xml.py
eugeneo: No longer needed. Deleted.

tools/buildgen/extract_metadata_from_bazel_xml.py
eugeneo: No longer needed. Deleted.

markdroth (approved):
Looks great!

markdroth (approved):

PR #35608, opened 2024-01-19
Enable service config resolution with c-ares by default

deeedob:
Currently, service configuration lookup in DNS TXT records is disabled by default and requires clients to set `GRPC_ARG_SERVICE_CONFIG_DISABLE_RESOLUTION` to false, as introduced in commit 93b7acd.

Considering my understanding of service configuration and the merged [error handling proposal](https://github.com/grpc/proposal/pull/100), it may seem reasonable to make it the default now.


markdroth:
I don't think we want to enable this by default.  There are potential security issues involving DNS TXT records, and I think clients should have to explicitly opt-in to using them rather than looking at them by default.

PR #35612, opened 2024-01-19
 [csds] Make grpc_dump_xds_configs return ClientStatusResponse

eugeneo:


gnossen (approved):

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -230,14 +236,28 @@ void SetXdsFallbackBootstrapConfig(const char* config) {
...
       grpc_core::ChannelArgs(), "grpc_dump_xds_configs()");
-  if (!xds_client.ok()) {
-    // If we aren't using xDS, just return an empty string.
-    return grpc_empty_slice();
+  upb::Arena arena;
+  // Following two containers should survive till serialization
+  std::vector<std::unique_ptr<std::string>> string_pool;
+  auto response = envoy_service_status_v3_ClientStatusResponse_new(arena.ptr());
+  // If we aren't using xDS, just return a response with empty config. This is
+  // consistent with older gRPC implementation
```
markdroth, line 249: No need for this last sentence.

src/cpp/server/csds/csds.cc
```diff
@@ -34,19 +34,18 @@ namespace grpc {
...
-using envoy::service::status::v3::ClientConfig;
 using envoy::service::status::v3::ClientStatusRequest;
 using envoy::service::status::v3::ClientStatusResponse;
 
 namespace {
 
-absl::StatusOr<ClientConfig> DumpClientConfig() {
-  ClientConfig client_config;
+absl::StatusOr<ClientStatusResponse> DumpClientStatusResponse() {
+  ClientStatusResponse client_config;
```
markdroth, line 43: Please rename this to `response`.

src/cpp/server/csds/csds.cc
```diff
@@ -59,7 +58,7 @@ Status ClientStatusDiscoveryService::StreamClientStatus(
   ClientStatusRequest request;
   while (stream->Read(&request)) {
     ClientStatusResponse response;
```
markdroth, line 60: I don't think we need a separate variable anymore.  Instead, let's just say:

```
absl::StatusOr<ClientStatusResponse> response = DumpClientStatusResponse();
if (!response.ok()) {
  // ...
}
stream->Write(*response);
```

src/cpp/server/csds/csds.cc
```diff
@@ -87,7 +86,7 @@ Status ClientStatusDiscoveryService::FetchClientStatus(
     return Status(static_cast<StatusCode>(s.status().raw_code()),
                   s.status().ToString());
   }
-  *response->add_config() = std::move(s.value());
+  *response = s.value();
```
markdroth, line 89: This should use `std::move()` to avoid an unnecessary copy.

test/core/xds/xds_client_fuzzer.cc
```diff
@@ -225,6 +225,14 @@ class Fuzzer {
     return &bootstrap.server();
   }
 
+  void DumpClientConfigBinary() {
+    upb::Arena arena;
+    auto client_config = envoy_service_status_v3_ClientConfig_new(arena.ptr());
+    std::vector<std::unique_ptr<std::string>> string_pool;
+    MutexLock lock(xds_client_->mutex());
```
markdroth, line 232: Suggest using a test spouse pattern to get access to the protected methods here.

markdroth (commented):
This looks like the right general approach!

test/core/xds/xds_client_fuzzer.cc
eugeneo: My first attempy at using this pattern, please review.

src/cpp/server/csds/csds.cc
eugeneo: Done

eugeneo (commented):

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -230,14 +236,28 @@ void SetXdsFallbackBootstrapConfig(const char* config) {
...
+  auto client_config = envoy_service_status_v3_ClientStatusResponse_add_config(
+      response, arena.ptr());
+  std::vector<std::unique_ptr<grpc_core::MutexLock>> client_locks;
```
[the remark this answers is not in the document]
markdroth: Interesting!  I thought that `reserve()` just allocated space without instantiating the objects, but it looks like it does actually need to default-construct them.

Note that `emplace_back()` does not require a move constructor.  The whole point of `emplace_back()` over `push_back()` is that it instantiates the object in place rather than moving or copying.  So the only problem here is actually the `reserve()`.

Thinking about this further, I think that once we actually have multiple `XdsClient` instances, we're going to have to construct a vector of refs to those instances anyway, and once we do that, it doesn't really make sense to also construct a vector of locks.  I think this is one of those rare cases where it will make sense to manually lock and unlock instead of using `MutexLock`:

```
// Grab a ref to every XdsClient instance.
std::vector<RefCountedPtr<GrpcXdsClient>> xds_clients;
{
  MutexLock lock(g_mu);
  xds_clients.reserve(g_xds_client_map->size());
  for (const auto& p : *g_xds_client_map) {
    const std::string& client_scope = p.first;
    auto xds_client = p.second->RefIfNonZero(DEBUG_LOCATION, "csds");
    if (xds_client != nullptr) {
      xds_clients.push_back(xds_client.TakeAsSubclass<GrpcXdsClient>());
    }
  }
}
// Construct response proto.
upb::Arena arena;
auto* response = envoy_service_status_v3_ClientStatusResponse_new(arena.ptr());
// For each XdsClient, acquire the lock and append its config to the response proto.
// Note that we continue to hold the lock until after we serialize,
// so that we don't need to copy data.
std::set<std::string> string_pool;
for (auto& xds_client : xds_clients) {
  auto* client_config = envoy_service_status_v3_ClientStatusResponse_add_config(
      response, arena.ptr());
  xds_client->mu()->Lock();
  xds_client->DumpClientConfig(&string_pool, arena.ptr(), client_config);
}
// Serialize the upb message to bytes.
size_t output_length;
char* output = envoy_service_status_v3_ClientStatusResponse_serialize(
    response, arena.ptr(), &output_length);
grpc_slice result = grpc_slice_from_cpp_string(std::string(output, output_length));
// Release mutexes.
for (auto& xds_client : xds_clients) {
  xds_client->mu()->Unlock();
}
return result;
```

We can use basically this same pattern now, even though there's only one `XdsClient`.

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -230,14 +236,28 @@ void SetXdsFallbackBootstrapConfig(const char* config) {
 }  // namespace grpc_core
 
 // The returned bytes may contain NULL(0), so we can't use c-string.
-grpc_slice grpc_dump_xds_configs(void) {
+grpc_slice grpc_dump_xds_configs(void) ABSL_NO_THREAD_SAFETY_ANALYSIS {
```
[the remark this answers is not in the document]
markdroth: If you switch to manually locking and unlocking as I suggested elsewhere, then we should not need `ABSL_NO_THREAD_SAFETY_ANALYSIS` here.

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -178,6 +184,29 @@ absl::StatusOr<RefCountedPtr<GrpcXdsClient>> GrpcXdsClient::GetOrCreate(
...
 
+grpc_slice GrpcXdsClient::DumpAllClientConfigs()
+    ABSL_NO_THREAD_SAFETY_ANALYSIS {
+  auto xds_client =
+      GrpcXdsClient::GetOrCreate(ChannelArgs(), "grpc_dump_xds_configs()");
+  upb::Arena arena;
+  // Following two containers should survive till serialization
+  std::set<std::string> string_pool;
+  auto response = envoy_service_status_v3_ClientStatusResponse_new(arena.ptr());
+  // If we aren't using xDS, just return a response with empty config.
```
markdroth, line 195: I don't think we actually want to add a `ClientConfig` to the response if there is no `XdsClient`.  I think we can just return an empty response proto in that case.  This is consistent with what we'll be doing once we do have multiple `XdsClient` instances.

Even if this is a change in the current behavior, I don't think it should cause any problems.

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -178,6 +184,29 @@ absl::StatusOr<RefCountedPtr<GrpcXdsClient>> GrpcXdsClient::GetOrCreate(
   return xds_client;
 }
 
+grpc_slice GrpcXdsClient::DumpAllClientConfigs()
+    ABSL_NO_THREAD_SAFETY_ANALYSIS {
+  auto xds_client =
+      GrpcXdsClient::GetOrCreate(ChannelArgs(), "grpc_dump_xds_configs()");
```
markdroth, line 190: There's not much point in creating a new `XdsClient` if we don't already have one.  I suggest doing something like this:

```
RefCountedPtr<GrpcXdsClient> xds_client;
{
  MutexLock lock(g_mu);
  xds_client =
      g_xds_client->RefIfNonZero(DEBUG_LOCATION, "csds")
          .TakeAsSubclass<GrpcXdsClient>();
}
```

Then just check if `xds_client` is not null.

eugeneo (commented):
Thanks for the review. I believe I addressed all the comments but GitHub UI does not allow me to respond to some comments...

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -184,26 +185,39 @@ absl::StatusOr<RefCountedPtr<GrpcXdsClient>> GrpcXdsClient::GetOrCreate(
...
 
-grpc_slice GrpcXdsClient::DumpAllClientConfigs()
-    ABSL_NO_THREAD_SAFETY_ANALYSIS {
-  auto xds_client =
-      GrpcXdsClient::GetOrCreate(ChannelArgs(), "grpc_dump_xds_configs()");
+grpc_slice GrpcXdsClient::DumpAllClientConfigs() {
+  std::vector<RefCountedPtr<GrpcXdsClient>> xds_clients;
+  {
+    MutexLock lock(g_mu);
+    RefCountedPtr<GrpcXdsClient> xds_client =
```
markdroth, line 192: This would be both clearer and slightly more efficient (one less move) if written as:

```
if (g_xds_client != nullptr) {
  auto xds_client = g_xds_client->RefIfNonZero(DEBUG_LOCATION, "DumpAllClientConfigs");
  if (xds_client != nullptr) {
    xds_clients.emplace_back(xds_client.TakeAsSubclass<GrpcXdsClient>());
  }
}
```

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -184,26 +185,39 @@ absl::StatusOr<RefCountedPtr<GrpcXdsClient>> GrpcXdsClient::GetOrCreate(
...
-  // If we aren't using xDS, just return a response with empty config.
-  auto client_config = envoy_service_status_v3_ClientStatusResponse_add_config(
-      response, arena.ptr());
-  std::vector<std::unique_ptr<MutexLock>> client_locks;
-  if (xds_client.ok()) {
-    client_locks.emplace_back(std::make_unique<MutexLock>((*xds_client)->mu()));
-    (*xds_client)->DumpClientConfig(&string_pool, arena.ptr(), client_config);
-  }
+  absl::c_for_each(
+      xds_clients, [&](const auto& xds_client) ABSL_NO_THREAD_SAFETY_ANALYSIS {
```
markdroth, line 206: When I suggested disabling the thread-safety analysis on a lambda, I thought that we could do that only for the lambda that unlocked the mutex.  If we need to do it in this lambda as well, then it's not really any better than doing it for the whole function, because the main thing that we want the thread safety for here is to ensure that we're holding the lock when we call `xds_client->DumpClientConfig()`.  Disabling it for the whole function is a lot more readable, and we shouldn't give up that readability unless we're actually getting something for it.

At this point, I suggest going back to just disabling thread-safety for the whole function.  But just to make sure that doesn't also disable the checks for our use of `g_mu`, let's move the code that constructs the vector of `XdsClient` refs into its own function, which we can call from here.

eugeneo: Done.

src/core/ext/xds/xds_client_grpc.cc
eugeneo: Done

eugeneo (commented):
Thanks for the reviews! I applied your suggestions and refactored the Python test to make sure XdsClient is properly setup for the test.

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -185,39 +198,27 @@ absl::StatusOr<RefCountedPtr<GrpcXdsClient>> GrpcXdsClient::GetOrCreate(
...
-                  .TakeAsSubclass<GrpcXdsClient>();
-    if (xds_client != nullptr) {
-      xds_clients.emplace_back(std::move(xds_client));
-    }
-  }
+grpc_slice GrpcXdsClient::DumpAllClientConfigs()
+    ABSL_NO_THREAD_SAFETY_ANALYSIS {
+  auto xds_clients = GetAllXdsClients();
   upb::Arena arena;
   // Following two containers should survive till serialization
```
markdroth, line 205: There's only one container here now. :)

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -185,39 +198,27 @@ absl::StatusOr<RefCountedPtr<GrpcXdsClient>> GrpcXdsClient::GetOrCreate(
...
   auto response = envoy_service_status_v3_ClientStatusResponse_new(arena.ptr());
-  absl::c_for_each(
-      xds_clients, [&](const auto& xds_client) ABSL_NO_THREAD_SAFETY_ANALYSIS {
-        auto client_config =
-            envoy_service_status_v3_ClientStatusResponse_add_config(
-                response, arena.ptr());
-        xds_client->mu()->Lock();
-        xds_client->DumpClientConfig(&string_pool, arena.ptr(), client_config);
-      });
+  for (const auto& xds_client : xds_clients) {
```
markdroth, line 208: Please add a comment here indicating that we are holding the locks in the individual XdsClient instances until we serialize, because the proto object that we're building contains pointers to data in the XdsClient caches, and we need to ensure that that data doesn't change until we serialize the proto.

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -185,39 +198,27 @@ absl::StatusOr<RefCountedPtr<GrpcXdsClient>> GrpcXdsClient::GetOrCreate(
...
-        g_xds_client == nullptr
-            ? nullptr
-            : g_xds_client->RefIfNonZero(DEBUG_LOCATION, "DumpAllClientConfigs")
-                  .TakeAsSubclass<GrpcXdsClient>();
-    if (xds_client != nullptr) {
-      xds_clients.emplace_back(std::move(xds_client));
-    }
-  }
+grpc_slice GrpcXdsClient::DumpAllClientConfigs()
+    ABSL_NO_THREAD_SAFETY_ANALYSIS {
```
markdroth, line 202: Please add a comment indicating that the reason for inhibiting the thread-safety checks is that the compiler can't handle the way we're acquiring mutexes in one loop and then releasing them in another loop.

markdroth (approved):
This looks good!

I've requested a few minor comment changes, but feel free to merge after doing that.  Thanks!

eugeneo (commented):
Thanks for the review! Applied the suggestions and will proceed to merge not.

markdroth (approved):

PR #35615, opened 2024-01-19
ssl_transport_security.cc: Make it compile with libressl

t-b:
Compiling grpc latest master [1] on Windows 10 Pro using MSVC 2019 and using LibreSSL 3.8.2 as SSL provider (package) does currently not work, as the functions X509_STORE_set_get_crl/X509_STORE_set_check_crl and the define SSL_OP_NO_RENEGOTIATION are not present.

Employ a workaround seen in the surrounding code by making the code block only for openssl available.

[1]: 24f89637 ([Test] Removed obsolete node test from run_tests.py (#35525), 2024-01-19)

### DIFF
--- a/src/core/tsi/ssl_transport_security.cc
+++ b/src/core/tsi/ssl_transport_security.cc
@@ -2082,7 +2082,7 @@ tsi_result tsi_create_ssl_client_handshaker_factory_with_options(
 #else
   ssl_context = SSL_CTX_new(TLSv1_2_method());
 #endif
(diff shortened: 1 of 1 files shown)


gtcooke94:
@markdroth this seems fine to me, can you take a look. We don't explicitly support LibreSSL nor make any guarantees about it's behavior, but I don't mind doing little things to let it compile.

I think the test failures are unrelated flakes

gtcooke94 (approved):

markdroth (approved):

PR #35542, opened 2024-01-12
[CocoaPods] Re-work how privacy manifests are exposed

ncooke3:
Follow-up to add privacy manifests to #35042

- [x] Update podspec template files
- [x] Align on manifest update:
```
git grep mach_absolute_time
src/core/lib/gpr/posix/time.cc:static uint64_t g_time_start = mach_absolute_time();
src/core/lib/gpr/posix/time.cc:                ((double)(mach_absolute_time() - g_time_start)) * g_time_scale;
```

cc: @paulb777

PR #35616, opened 2024-01-21
[ObjC] add privacy manifests to boringssl grpc

HannahShiSFB:
BoringSSL-GRPC uses sources from https://github.com/google/boringssl.git, which doesn't contain PrivacyInfo.xcprivacy file.
This PR embeds the manifest data into the podspec and extract during pod install, similar to file boringssl_prefix_symbols.h

--
Will need to regenerate the podspec after #35542 merged.


sampajano (approved):
@HannahShiSFB Thanks for the change!! 😃

sampajano:
@HannahShiSFB Ahh there's a conflict now! Would you mind fixing it?

PR #35631, opened 2024-01-23
[otel] Re-structure otel_plugin_test initialization

yashykt:


test/cpp/ext/otel/otel_test_library.h
```diff
@@ -54,31 +54,94 @@ class MockMetricReader : public opentelemetry::sdk::metrics::MetricReader {
...
+    return *this;
+  }
+
+  OpenTelemetryPluginTestConfiguration& set_labels_injector(
+      std::unique_ptr<grpc::internal::LabelsInjector> injector) {
+    labels_injector = std::move(injector);
+    return *this;
+  }
+
+  OpenTelemetryPluginTestConfiguration& set_test_no_meter_provider(bool flag) {
```
markdroth, line 77: I suggest inverting this: call the parameter `use_meter_provider`, make it true by default, and use this method to disable it when appropriate.

test/cpp/ext/otel/otel_test_library.h
```diff
@@ -54,31 +54,94 @@ class MockMetricReader : public opentelemetry::sdk::metrics::MetricReader {
...
-  void Init(
-      const absl::flat_hash_set<absl::string_view>& metric_names,
-      opentelemetry::sdk::resource::Resource resource =
-          opentelemetry::sdk::resource::Resource::Create({}),
-      std::unique_ptr<grpc::internal::LabelsInjector> labels_injector = nullptr,
-      bool test_no_meter_provider = false,
-      const std::map<std::string, std::string>& labels_to_inject = {},
-      absl::AnyInvocable<bool(absl::string_view /*target*/) const>
-          target_selector = absl::AnyInvocable<bool(absl::string_view) const>(),
+struct OpenTelemetryPluginTestConfiguration {
```
markdroth, line 57: Suggest moving this inside of `OpenTelemetryPluginEnd2EndTest` and calling it `Options`.

markdroth (approved):

test/cpp/ext/otel/otel_test_library.h
yashykt: note that there is just one test that needs to disable meter provider. All others need it, which means that all other tests would need to set this option if we invert it.

markdroth: I'm not suggesting changing the default, just inverting the semantics to avoid a double negative.

Right now, the one test that needs to not use a MeterProvider does `set_test_no_meter_provider(true)`.  What I'm suggesting is that that test should instead say `set_use_meter_provider(false)`.  No other test should need any changes.

yashykt: ah, misread the comment

test/cpp/ext/otel/otel_test_library.h
yashykt: done

test/cpp/ext/otel/otel_test_library.h
yashykt: done

markdroth (approved):

PR #35632, opened 2024-01-23
[EventEngine] Dump all thread stacks when the thread pool is wedged

drfloob:
This should help us debug rare flakes where the thread pool waits indefinitely


src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -271,6 +310,37 @@ void WorkStealingThreadPool::WorkStealingThreadPoolImpl::Postfork() {
...
+}
+
+void WorkStealingThreadPool::WorkStealingThreadPoolImpl::DumpStacksAndCrash() {
+  grpc_core::MutexLock lock(&thd_set_mu_);
+  gpr_log(GPR_ERROR,
+          "Pool did not quiesce in time, gRPC will not shut down cleanly. "
+          "Dumping all %zu thread stacks.",
+          thds_.size());
+  for (const auto tid : thds_) {
+    grpc_core::Thread::Signal(tid, kDumpStackSignal);
```
ctiller, line 332: This is likely to have some unintended consequences when we're linked into someone elses binary -- signals are a limited resource, and their assignment needs to be at the whim of the binary owner, not some random library they link in.

Consequently I'd recommend placing this under some kind of environment variable and only enabling it when we're running under our own CI.

drfloob: Done. I've added an internal env var that controls this behavior, and all `grpc_cc_test`s set it.

I also updated the logic to check the timeout relative to the last count change. Some stress tests may take a while to quiesce, which is fine. Generally in these circumstances, the pool will grow & shrink, so I'm using the count to indicate wedgedness.

src/core/lib/gprpp/time.h
```diff
@@ -49,8 +49,9 @@
     uint64_t now = grpc_core::Timestamp::FromTimespecRoundDown( \
                        gpr_now(GPR_CLOCK_MONOTONIC))            \
                        .milliseconds_after_process_epoch();     \
-    uint64_t prev_tsamp = prev.exchange(now);                   \
-    if (now - prev_tsamp > (n) * 1000) {                        \
+    if (prev == 0) prev = now;                                  \
```
apolcyn, line 52: this assignment is racey right?

drfloob: In terms of memory access errors? I don't see how, this macro should be called in a loop, single-threaded, in which these would be local variables. In terms of comparison in the subsequent equality check, it's ok if there's some time drift here, this log rate can be loose.

See `GRPC_LOG_EVERY_N_SEC` above. This change is a fix, it was inconsistently logging at the expected rate.

apolcyn: I was thinking just about logical raciness affecting the time/rate.

e..g. two threads could write to `prev` on lines 52 and 54.

Though as discussed, neither of these cases shouldn't really make a meaningful difference in the rate math

drfloob: I see your point, good catch. As discussed, there is some potential time slop here, we could improve it using something like a thread_local, but I agree it's likely not worth it. This method may already log messages slower than `n` if it's called infrequently, or called with some rate that does not evenly divide `n`.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -125,6 +168,11 @@ thread_local WorkQueue* g_local_queue = nullptr;
 
 WorkStealingThreadPool::WorkStealingThreadPool(size_t reserve_threads)
     : pool_{std::make_shared<WorkStealingThreadPoolImpl>(reserve_threads)} {
+  static bool logged = false;
```
apolcyn, line 171: is this class's ctor (i.e. `MakeThreadPool`) not thread safe? (wondering if `logged` should be atomic)

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -221,8 +269,12 @@ void WorkStealingThreadPool::WorkStealingThreadPoolImpl::Quiesce() {
   // running instead of zero.
   bool is_threadpool_thread = g_local_queue != nullptr;
   work_signal()->SignalAll();
-  living_thread_count_.BlockUntilThreadCount(is_threadpool_thread ? 1 : 0,
-                                             "shutting down");
+  auto threads_were_shut_down = living_thread_count_.BlockUntilThreadCount(
+      is_threadpool_thread ? 1 : 0, "shutting down",
+      kBlockUntilThreadCountTimeout);
```
apolcyn, line 274: Wondering if it's worth preserving the behavior of an infinite timeout if `g_log_verbose_failures == false`.

Otherwise, what happens if we reach the `kBlockUntilThreadCountTimeout` timeout and `g_log_verbose_failures == false` ?

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
drfloob: Good catch. I removed the check, and now only trace-log the message ... on every thread pool instantiation. There's usually one, except for in tests. I don't think we have a good pattern today for a thread-safe post-grpc-init once-init situation, which is what I'd reach for.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
drfloob: Good catch! I agree, fixed.

apolcyn (approved):

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
apolcyn: > I don't think we have a good pattern today for a thread-safe post-grpc-init once-init situation, which is what I'd reach for.

Maybe `gpr_once` ?

Current code LGTM though

drfloob:
Closed in favor of https://github.com/grpc/grpc/pull/35656

PR #35633, opened 2024-01-23
[CSM O11Y] Fix issue when CSM optional labels are present in server metrics

yijiem:

### DIFF
--- a/src/cpp/ext/csm/metadata_exchange.cc
+++ b/src/cpp/ext/csm/metadata_exchange.cc
@@ -462,18 +462,17 @@ void ServiceMeshLabelsInjector::AddLabels(
 }
 
 bool ServiceMeshLabelsInjector::AddOptionalLabels(
-    Entity entity,
+    bool is_client,
     absl::Span<const std::shared_ptr<std::map<std::string, std::string>>>
         optional_labels_span,
     opentelemetry::nostd::function_ref<
         bool(opentelemetry::nostd::string_view,
              opentelemetry::common::AttributeValue)>
         callback) const {
-  if (entity == Entity::kServer) {
+  if (!is_client) {
     // Currently the CSM optional labels are only set on client.
     return true;
   }
-  GPR_ASSERT(entity == Entity::kClient);
   // According to the CSM Observability Metric spec, if the control plane fails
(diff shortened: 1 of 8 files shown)


src/cpp/ext/otel/otel_plugin.h
```diff
@@ -66,6 +66,15 @@ class LabelsIterable {
 // through the OpenTelemetry plugin.
 class LabelsInjector {
  public:
+  // Some optional labels may only be set on client or server. We use this enum
+  // to specify the entity of the caller for the APIs that need this
+  // information.
+  enum class Entity {
```
yashykt, line 72: Throughout gRPC Core, we've just been using a simple `is_client` bool for this. If we were to do this, I would prefer doing it all over.

markdroth: +1 -- let's just use a bool here.  We're not expecting to have any context other than client or server in the future, so there's no need to leave room for future options.

markdroth:
I assume this will need to be back-ported to 1.61 before the release goes out.

src/cpp/ext/otel/otel_plugin.h
yijiem: Done.

yashykt (approved):

src/cpp/ext/csm/metadata_exchange.cc
```diff
@@ -493,9 +498,13 @@ bool ServiceMeshLabelsInjector::AddOptionalLabels(
 }
 
 size_t ServiceMeshLabelsInjector::GetOptionalLabelsSize(
+    bool is_client,
     absl::Span<const std::shared_ptr<std::map<std::string, std::string>>>)
     const {
-  return 2;
+  if (is_client) {
```
markdroth, line 504: Can write this as `return is_client ? 2 : 0`.

markdroth (approved):

src/cpp/ext/csm/metadata_exchange.cc
yijiem: Done.

markdroth (approved):

markdroth (approved):

PR #35640, opened 2024-01-23
[DO NOT MERGE] sergiitk observability psm interop image experiments

sergiitk:


sergiitk:
- [ ] [grpc/core/master/linux/psm-csm](https://source.cloud.google.com/results/invocations/6034d22d-9596-43a5-b362-b379071ef58e)

PR #35641, opened 2024-01-23
[Security - CrlProvider] Use a better mechanism for Crl Lookup and add some verification helpers

gtcooke94:
This PR does 2 distinct things, I can unbundle them if desired

1) Add functions in `ssl_transport_security_utils` and associated tests that will eventually be used for additional Crl validation (the logic of actually doing this will be in a future PR), so other than the tests these fns are currently unused.

2) Remove the use of `X509_NAME_oneline` - it is not a guaranteed stable way to get the issuer name for lookups. Instead, use the DER encoding via `i2d_X509_NAME` - the results in a non-human readable string that is stable for lookup, and necessitated some change to the CrlProvider test code that previously used a human readable string for this value.

Neither should result in behavior changes.


src/core/tsi/ssl_transport_security_utils.h
```diff
@@ -142,6 +142,23 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+// return: 0 if equal, otherwise a value per X509_NAME_cmp
+int verify_crl_cert_issuer_names_match(X509_CRL* crl, X509* issuer);
+
+// Verifies the certificate in question has the cRLSign bit present
+// return: true if cRLSign bit is present, 0 otherwise
+bool verify_crl_sign_bit(X509* issuer);
+
+// Gets a stable representation of the issuer name from an X509 certificate
+// return: a std::string of the DER encoding of the X509_NAME issuer name
+std::string IssuerFromCert(X509* cert);
```
matthewstevenson88, line 160: Rather than returning an empty string on error, could we change the return type to e.g. `std::optional<std::string>` or `absl::StatusOr<std::string>`?

src/core/tsi/ssl_transport_security_utils.h
```diff
@@ -142,6 +142,23 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
                                  unsigned char* unprotected_bytes,
                                  size_t* unprotected_bytes_size);
 
+// Verifies the CRL signature using X509_CRL_verify
+// https://www.openssl.org/docs/man3.0/man3/X509_CRL_verify.html
+// return: 1 if valid, 0 if invalid, and -1 is there was a problem
+int verify_crl_signature(X509_CRL* crl, X509* issuer);
```
matthewstevenson88, line 148: For this function and the ones below, can we use C++-style naming, e.g. `VerifyCrlSignature`?

matthewstevenson88, line 148: I'm not a big fan of using an `int` as a "status" return type, since it is error prone IMO. Could we instead return `absl::Status` or a `bool`?

This puts the burden of correctly interpreting the return type of the OpenSSL APIs on us, rather than also passing this burden to the callers of this API.

Similarly below.

src/core/tsi/ssl_transport_security_utils.h
```diff
@@ -142,6 +142,23 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
                                  unsigned char* unprotected_bytes,
                                  size_t* unprotected_bytes_size);
 
+// Verifies the CRL signature using X509_CRL_verify
```
matthewstevenson88, line 145: My 2 cents: The API description should describe what the API does, rather than leaning on explaining the implementation details.

Similarly below.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +248,40 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+    EVP_PKEY_free(ikey);
+    return 1;
+  }
+  int ret = X509_CRL_verify(crl, ikey);
+  EVP_PKEY_free(ikey);
+  return ret;
+}
+
+int verify_crl_cert_issuer_names_match(X509_CRL* crl, X509* issuer) {
+  return X509_NAME_cmp(X509_get_issuer_name(issuer), X509_CRL_get_issuer(crl));
```
matthewstevenson88, line 265: Please check that both `X509_get_issuer_name(issuer)` and `X509_CRL_get_issuer(crl)` are not null before passing them into `X509_NAME_cmp`, otherwise this will cause a segfault.

matthewstevenson88, line 265: Should we be using `X509_get_subject_name(issuer)` instead of `X509_get_issuer_name(issuer)` here?

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +248,40 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+    // check.
+    EVP_PKEY_free(ikey);
+    return 1;
+  }
+  int ret = X509_CRL_verify(crl, ikey);
+  EVP_PKEY_free(ikey);
+  return ret;
+}
+
+int verify_crl_cert_issuer_names_match(X509_CRL* crl, X509* issuer) {
```
matthewstevenson88, line 264: Please check that both `crl` and `issuer` are not null before proceeding, otherwise we would segfault when passing them into `X509_get_issuer_name` or `X509_CRL_get_issuer`.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +248,40 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
 
+int verify_crl_signature(X509_CRL* crl, X509* issuer) {
+  EVP_PKEY* ikey = X509_get_pubkey(issuer);
+  if (ikey == nullptr) {
+    // Can't verify signature because we couldn't get the pubkey, fail the
+    // check.
+    EVP_PKEY_free(ikey);
+    return 1;
+  }
+  int ret = X509_CRL_verify(crl, ikey);
```
matthewstevenson88, line 259: Please check that `crl` is not null before proceeding, otherwise we would segfault when passing it into `X509_CRL_verify`.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +248,40 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+bool verify_crl_sign_bit(X509* issuer) {
+  return (X509_get_key_usage(issuer) & KU_CRL_SIGN) != 0;
+}
+
+std::string IssuerFromCert(X509* cert) {
+  if (cert == nullptr) {
+    return "";
+  }
+  X509_NAME* issuer = X509_get_issuer_name(cert);
+  int len;
```
matthewstevenson88, line 277: Any reason this is broken up? I would have expected this to be written as:
```
unsigned char* buf = nullptr;
int len = i2d_X509_NAME(issuer, &buf);
```

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +248,40 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+  EVP_PKEY_free(ikey);
+  return ret;
+}
+
+int verify_crl_cert_issuer_names_match(X509_CRL* crl, X509* issuer) {
+  return X509_NAME_cmp(X509_get_issuer_name(issuer), X509_CRL_get_issuer(crl));
+}
+
+bool verify_crl_sign_bit(X509* issuer) {
+  return (X509_get_key_usage(issuer) & KU_CRL_SIGN) != 0;
```
matthewstevenson88, line 269: Please check that `issuer` is not null, otherwise we would segfault when we pass into `X509_get_key_usage`.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +248,40 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
   return result;
 }
 
+int verify_crl_signature(X509_CRL* crl, X509* issuer) {
+  EVP_PKEY* ikey = X509_get_pubkey(issuer);
```
matthewstevenson88, line 252: Please check that `issuer` is not null before passing it into `X509_get_pubkey`.

src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc
```diff
@@ -57,11 +57,13 @@ std::string IssuerFromCrl(X509_CRL* crl) {
   if (crl == nullptr) {
     return "";
   }
-  char* buf = X509_NAME_oneline(X509_CRL_get_issuer(crl), nullptr, 0);
-  std::string ret;
-  if (buf != nullptr) {
-    ret = buf;
-  }
+  X509_NAME* issuer = X509_CRL_get_issuer(crl);
```
matthewstevenson88, line 60: Can we also make this code block into a helper function in `ssl_transport_security_utils.{h,cc}`? This way it can be reused elsewhere if needed, and we can test it in the same place as the similar APIs.

src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc
```diff
@@ -57,11 +57,13 @@ std::string IssuerFromCrl(X509_CRL* crl) {
...
+  int len;
```
matthewstevenson88, line 61: Any reason to break up the code like this? I would have expected it to be written as:
```
unsigned char* buf = nullptr;
int len = i2d_X509_NAME(issuer, &buf);
```

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,135 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
 
 #endif  // OPENSSL_IS_BORINGSSL
 
+X509_CRL* read_crl(absl::string_view crl_string) {
```
matthewstevenson88, line 448: Can we move these "ReadCrl" and "ReadCert" methods to a test utility file so that we can reuse them, rather than duplicating the implementations?

test/core/security/grpc_tls_crl_provider_test.cc
```diff
@@ -88,9 +86,38 @@ class FakeDirectoryReader : public DirectoryReader {
...
+    X509* issuer = read_cert(cert_string);
+    crl_issuer = IssuerFromCert(issuer);
+    std::string intermediate_string =
+        GetFileContents(kCrlIntermediateIssuerPath.data());
+    X509* intermediate_issuer = read_cert(intermediate_string);
+    intermediate_crl_issuer = IssuerFromCert(intermediate_issuer);
+  }
+
+ protected:
+  std::string crl_issuer;
```
matthewstevenson88, line 102: nit: Generally we have the private methods listed before the private member fields.

matthewstevenson88, line 102: nit: Please add "_" at the end of these fields, so e.g. `crl_issuer_` instead of `crl_issuer`. Similarly for `intermediate_crl_issuer`.

test/core/security/grpc_tls_crl_provider_test.cc
```diff
@@ -88,9 +86,38 @@ class FakeDirectoryReader : public DirectoryReader {
...
+  std::string intermediate_crl_issuer;
+  X509* read_cert(absl::string_view cert_string) {
```
matthewstevenson88, line 104: nit: Can we use C++ naming and call it "ReadCert" instead? Also would "ParsePemCert" be a more appropriate name?

matthewstevenson88, line 104: nit: Would `pem_cert` be a more appropriate variable name than `cert_string`?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,135 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+    return nullptr;
+  }
+  X509* cert = PEM_read_bio_X509(cert_bio, nullptr, nullptr, nullptr);
+  BIO_free(cert_bio);
+  return cert;
+}
+
+TEST(CrlUtils, VerifySignatureValid) {
+  absl::StatusOr<Slice> crl_slice = LoadFile(kValidCrl, false);
+  ASSERT_TRUE(crl_slice.ok()) << crl_slice.status();
```
matthewstevenson88, line 474: optional: Consider using `ASSERT_EQ(crl_slice.status(), absl::OkStatus());` so that we get more information on a failure. Similarly throughout.

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,135 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+      BIO_new_mem_buf(cert_string.data(), static_cast<int>(cert_string.size()));
+  // Errors on BIO
+  if (cert_bio == nullptr) {
+    return nullptr;
+  }
+  X509* cert = PEM_read_bio_X509(cert_bio, nullptr, nullptr, nullptr);
+  BIO_free(cert_bio);
+  return cert;
+}
+
```
matthewstevenson88, line 471: Please add unit tests for all of the nullptr behaviors mentioned above.

matthewstevenson88 (commented):
Thanks for the PR Greg! Did an initial pass.

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Done

gtcooke94: I don't think so, but it depends on how we call it

In this case, we're asking whether a given certificate and a given crl have the same issuer. Because a CRL issued by issuer A would be the right CRL to revoke a cert issued by issuer A. But a cert issued by issuer A would have a different subject.
To check the subject name against the CRL we'd have to go one up the chain, because with Direct CRLs the subject one up the chain is the issuer.

Does that make sense to you?

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Done

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Again, copying from OpenSSL reference will clean up

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: Definitely, don't know what I was thinking with the naming 🤦

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: Not addressing this yet, so please leave open. Will make the text better

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: My plan was to have this layer be super close to OpenSSL, so I was using their returns.
Then I was going to add another layer that used our internal object representations and returned bools or statuses - does that sound good to you?

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: Done

test/core/security/grpc_tls_crl_provider_test.cc
gtcooke94: Fixed

gtcooke94: Fixed, my bad

test/core/security/grpc_tls_crl_provider_test.cc
gtcooke94: Done

gtcooke94: Yes, changed

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Yes - is there an existing test utils file that these fit in well?

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Will do, haven't done yet so don't resolve this

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Done

gtcooke94 (commented):
Addressed most comments, have a few ones that I will continue to work on

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: Ok this should be addressed

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Have added nullptr test

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+  X509* issuer = read_cert(issuer_slice->as_string_view());
+  EXPECT_EQ(VerifyCrlSignature(crl, issuer), 0);
+}
+
+TEST(CrlUtils, VerifySignatureModifiedContent) {
+  absl::StatusOr<Slice> crl_slice = LoadFile(kModifiedContent, false);
+  ASSERT_EQ(crl_slice.status(), absl::OkStatus()) << crl_slice.status();
+  absl::StatusOr<Slice> issuer_slice = LoadFile(kCrlIssuer, false);
+  ASSERT_EQ(issuer_slice.status(), absl::OkStatus());
+  X509_CRL* crl = read_crl(crl_slice->as_string_view());
```
matthewstevenson88, line 507: nit: Can we simplify this into 1 line?

```
EXPECT_EQ(ReadCrl(crl_slice->as_string_view(), nullptr);
```

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+  ASSERT_EQ(crl_slice.status(), absl::OkStatus()) << crl_slice.status();
+  X509_CRL* crl = read_crl(crl_slice->as_string_view());
+  X509* issuer = nullptr;
+  EXPECT_EQ(VerifyCrlSignature(crl, issuer), -1);
+}
+
+TEST(CrlUtils, VerifySignatureNullCrlAndCert) {
+  X509_CRL* crl = nullptr;
+  X509* issuer = nullptr;
+  EXPECT_EQ(VerifyCrlSignature(crl, issuer), -1);
```
matthewstevenson88, line 550: nit: Can we simplify this to 1 line?
```
EXPECT_EQ(VerifyCrlSignature(/*crl=*/nullptr, /*issuer=*/nullptr), -1);
```

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+}
+
+TEST(CrlUtils, VerifyIssuerNamesDontMatch) {
+  absl::StatusOr<Slice> crl_slice = LoadFile(kValidCrl, false);
+  ASSERT_EQ(crl_slice.status(), absl::OkStatus()) << crl_slice.status();
+  absl::StatusOr<Slice> issuer_slice = LoadFile(kLeafCert, false);
+  ASSERT_EQ(issuer_slice.status(), absl::OkStatus());
+  X509_CRL* crl = read_crl(crl_slice->as_string_view());
+  X509* issuer = read_cert(issuer_slice->as_string_view());
+  EXPECT_NE(VerifyCrlCertIssuerNamesMatch(crl, issuer), 0);
```
matthewstevenson88, line 570: Any reason to not check the exact return value here?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+
+TEST(CrlUtils, VerifySignatureNullCert) {
+  absl::StatusOr<Slice> crl_slice = LoadFile(kIntermediateCrl, false);
+  ASSERT_EQ(crl_slice.status(), absl::OkStatus()) << crl_slice.status();
+  X509_CRL* crl = read_crl(crl_slice->as_string_view());
+  X509* issuer = nullptr;
+  EXPECT_EQ(VerifyCrlSignature(crl, issuer), -1);
+}
+
+TEST(CrlUtils, VerifySignatureNullCrlAndCert) {
```
matthewstevenson88, line 547: I think this test only really catches the "is crl null" case and we need a separate test for the "is issuer null" case.

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+  EXPECT_EQ(VerifyCrlCertIssuerNamesMatch(crl, issuer), 0);
+  EXPECT_EQ(VerifyCrlSignature(crl, issuer), 0);
+}
+
+TEST(CrlUtils, VerifyIssuerNameNullCrl) {
+  absl::StatusOr<Slice> issuer_slice = LoadFile(kCrlIssuer, false);
+  ASSERT_EQ(issuer_slice.status(), absl::OkStatus());
+  X509_CRL* crl = nullptr;
+  X509* issuer = read_cert(issuer_slice->as_string_view());
+  EXPECT_NE(VerifyCrlCertIssuerNamesMatch(crl, issuer), 0);
```
matthewstevenson88, line 590: Any reason to not check the exact return value here?

matthewstevenson88, line 590: nit: Can we inline the `/*crl=*/nullptr` argument, as mentioned above?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+}
+
+TEST(CrlUtils, VerifyIssuerNameNullCert) {
+  absl::StatusOr<Slice> crl_slice = LoadFile(kIntermediateCrl, false);
+  ASSERT_EQ(crl_slice.status(), absl::OkStatus()) << crl_slice.status();
+  X509_CRL* crl = read_crl(crl_slice->as_string_view());
+  X509* issuer = nullptr;
+  EXPECT_NE(VerifyCrlCertIssuerNamesMatch(crl, issuer), 0);
```
matthewstevenson88, line 598: nit: Can we inline the `/*issuer=*/nullptr` argument, as mentioned above?

matthewstevenson88, line 598: Any reason to not check the exact return value here?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+}
+
+TEST(CrlUtils, VerifyIssuerNameNullCrlAndCert) {
+  X509_CRL* crl = nullptr;
+  X509* issuer = nullptr;
+  EXPECT_NE(VerifyCrlCertIssuerNamesMatch(crl, issuer), 0);
```
matthewstevenson88, line 604: nit: Same here RE inlining the nullptr arguments.

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+
+TEST(CrlUtils, VerifyIssuerNameNullCert) {
+  absl::StatusOr<Slice> crl_slice = LoadFile(kIntermediateCrl, false);
+  ASSERT_EQ(crl_slice.status(), absl::OkStatus()) << crl_slice.status();
+  X509_CRL* crl = read_crl(crl_slice->as_string_view());
+  X509* issuer = nullptr;
+  EXPECT_NE(VerifyCrlCertIssuerNamesMatch(crl, issuer), 0);
+}
+
+TEST(CrlUtils, VerifyIssuerNameNullCrlAndCert) {
```
matthewstevenson88, line 601: Is this test adding anything new beyond `VerifyIssuerNameNullCrl` and `VerifyIssuerNameNullCert`?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+TEST(CrlUtils, VerifyCrlSignBitMissing) {
+  absl::StatusOr<Slice> issuer_slice = LoadFile(kLeafCert, false);
+  ASSERT_EQ(issuer_slice.status(), absl::OkStatus());
+  X509* issuer = read_cert(issuer_slice->as_string_view());
+  EXPECT_FALSE(VerifyCrlSignBit(issuer));
+}
+
+TEST(CrlUtils, VerifyCrlSignBitNullCert) {
+  X509* issuer = nullptr;
+  EXPECT_FALSE(VerifyCrlSignBit(issuer));
```
matthewstevenson88, line 623: nit: Can we inline `/*issuer=*/nullptr` here?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+}
+
+TEST(CrlUtils, VerifyCrlSignBitMissing) {
+  absl::StatusOr<Slice> issuer_slice = LoadFile(kLeafCert, false);
+  ASSERT_EQ(issuer_slice.status(), absl::OkStatus());
+  X509* issuer = read_cert(issuer_slice->as_string_view());
+  EXPECT_FALSE(VerifyCrlSignBit(issuer));
+}
+
+TEST(CrlUtils, VerifyCrlSignBitNullCert) {
```
matthewstevenson88, line 621: Do we also need a "nullptr" test for `IssuerFromCert`?

src/core/tsi/ssl_transport_security_utils.h
```diff
@@ -142,6 +144,22 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
                                  unsigned char* unprotected_bytes,
                                  size_t* unprotected_bytes_size);
 
+// Verifies that `crl` was signed by `issuer`
```
matthewstevenson88, line 147: micro-nit: Please add periods at the end of the function comments.

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
 
 #endif  // OPENSSL_IS_BORINGSSL
 
+X509_CRL* read_crl(absl::string_view crl_string) {
```
matthewstevenson88, line 448: Same here RE C++ naming.

matthewstevenson88, line 448: Same here RE changing the variable name to `crl_pem`.

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +445,184 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+  // Errors on BIO
+  if (crl_bio == nullptr) {
+    return nullptr;
+  }
+  X509_CRL* crl = PEM_read_bio_X509_CRL(crl_bio, nullptr, nullptr, nullptr);
+  BIO_free(crl_bio);
+  return crl;
+}
+
+X509* read_cert(absl::string_view cert_string) {
```
matthewstevenson88, line 460: Same here RE changing the variable name to `cert_pem`.

test/core/security/grpc_tls_crl_provider_test.cc
```diff
@@ -88,9 +86,43 @@ class FakeDirectoryReader : public DirectoryReader {
...
+    std::string pem_cert = GetFileContents(kRootCert.data());
+    X509* issuer = ReadPemCert(pem_cert);
+    auto base_crl_issuer = IssuerFromCert(issuer);
+    ASSERT_TRUE(base_crl_issuer.ok());
+    base_crl_issuer_ = *base_crl_issuer;
+    std::string intermediate_string =
+        GetFileContents(kCrlIntermediateIssuerPath.data());
+    X509* intermediate_issuer = ReadPemCert(intermediate_string);
+    auto intermediate_crl_issuer = IssuerFromCert(intermediate_issuer);
+    ASSERT_TRUE(intermediate_crl_issuer.ok());
```
matthewstevenson88, line 101: nit/optional: consider using `ASSERT_EQ(intermediate_crl_issuer.status(), absl::OkStatus());`

src/core/tsi/ssl_transport_security_utils.h
matthewstevenson88: IIUC, we don't need to feed these values back into OpenSSL anywhere, do we?

If we don't need to be compatible with OpenSSL, then wouldn't it be safest to immediately change to a safer return type?

Please continue to push back if I missed something or if you disagree. :)

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Pinging this, since it seems to still be the case.

src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc
matthewstevenson88: Pinging this since it still seems to be open.

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: OK so to confirm, what we want to do here is: given a cert and CRL, do they have the same issuer. Is that correct?

If yes, then I think the problem is using `issuer` as the variable name rather than e.g. `cert`?

WDYT?

src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc
matthewstevenson88: Thanks for investigating. I think it would be good to pull out the `ssl_transport_security_util` into its own BUILD target as you suggest, so that we can avoid duplicating OpenSSL code, avoid divergent behavior between the copies of the code, etc. Totally fine doing this in a separate PR though. Can you leave a TODO for this?

gtcooke94: Done

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: Done

test/core/security/grpc_tls_crl_provider_test.cc
gtcooke94: Done

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Done

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Addressed in moving to a util file

gtcooke94: Done

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Done

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: I'm of the (changeable) opinion that it's clearer as-is and follows the same structure as the existing tests right now, namely reading the CRL then doing the assert on it. Combining them makes it just a little bit harder to see what is being tested

But I don't feel super strongly, just leaning that way.

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: I have a test for just null CRL and just null cert above this, figured I'd have the complete set even if the test doesn't do much else

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Same as my previous comment about that potentially making it _less_ clear what is being asserted/tested

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: I believe it can return any value != 0, negative if the first is "less" and positive if the first is "more"

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Same as other comment

gtcooke94: Same readability and consistency comment from other comment chains

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Same readability and consistency comment from other comment chains

gtcooke94: Same as other comment

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Same as other comment

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Not particularly, just tests the full permutation of arguements

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Ah good catch, I don't really have appropriate tests for it.

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Same comment about structure

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: No, but imagine we have another fn that looks like this `bool VerifyCrlSignature(OurCrlType, OurCertType);`

Then it handles the conversion of our types to call this fn, and the conversion of that output to something meaningful to grpc, and keeps the core code where this will be called cleaner by not having to do a bunch of type translations

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Ah that makes a lot of sense, I've fixed it

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: I think this comment is on an old commit, I've fixed it on the tip of this branch

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,216 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+      (const unsigned char*)"Internet Widgits Pty Ltd", -1, -1, 0));
+  ASSERT_TRUE(X509_NAME_add_entry_by_txt(
+      expected, "CN", MBSTRING_ASC, (const unsigned char*)"testca", -1, -1, 0));
+  int len = i2d_X509_NAME(expected, &buf);
+  std::string expected_name(reinterpret_cast<char const*>(buf), len);
+  OPENSSL_free(buf);
+  X509_NAME_free(expected);
+
+  ASSERT_EQ(issuer.status(), absl::OkStatus());
+  std::cout << *issuer << "xxx\n";
```
matthewstevenson88, line 629: nit: Please remove the print statement.

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,216 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+  ASSERT_TRUE(X509_NAME_add_entry_by_txt(
+      expected, "CN", MBSTRING_ASC,
+      (const unsigned char*)"intermediatecert.example.com", -1, -1, 0));
+  int len = i2d_X509_NAME(expected, &buf);
+  std::string expected_name(reinterpret_cast<char const*>(buf), len);
+
+  OPENSSL_free(buf);
+  X509_NAME_free(expected);
+  ASSERT_EQ(issuer.status(), absl::OkStatus());
+  std::cout << *issuer << "xxx\n";
```
matthewstevenson88, line 651: nit: Please remove the print statement.

src/core/tsi/ssl_transport_security_utils.h
matthewstevenson88: But wouldn't the scenario you describe be made simpler and less error-prone by having the return type of our `VerifyCrlSignature` function be `bool` instead of `int`? Said differently, is there any scenario where the caller needs to distinguish between "something was invalid" and "there was a problem"?

IMO, we should make an API like this one as safe and easy-to-use as possible. And if there is a reason to make the return type more complex in the future, we can always change it.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +251,56 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+  int ret = X509_CRL_verify(crl, ikey);
+  EVP_PKEY_free(ikey);
+  return ret;
+}
+
+int VerifyCrlCertIssuerNamesMatch(X509_CRL* crl, X509* cert) {
+  if (cert == nullptr || crl == nullptr) {
+    return 1;
+  }
+  X509_NAME* cert_name = X509_get_issuer_name(cert);
```
matthewstevenson88, line 274: nit: Should this be `cert_issuer_name` instead?

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: It looks like there are a couple instances where this still occurs, I'll comment on them.

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,216 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+  ASSERT_TRUE(X509_NAME_add_entry_by_txt(expected, "ST", MBSTRING_ASC,
+                                         (const unsigned char*)"Some-State", -1,
+                                         -1, 0));
+  ASSERT_TRUE(X509_NAME_add_entry_by_txt(
+      expected, "O", MBSTRING_ASC,
+      (const unsigned char*)"Internet Widgits Pty Ltd", -1, -1, 0));
+  ASSERT_TRUE(X509_NAME_add_entry_by_txt(
+      expected, "CN", MBSTRING_ASC, (const unsigned char*)"testca", -1, -1, 0));
+  int len = i2d_X509_NAME(expected, &buf);
+  std::string expected_name(reinterpret_cast<char const*>(buf), len);
```
matthewstevenson88, line 624: nit: Should we call this "expected_issuer_name_der"?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,216 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+  auto issuer = IssuerFromCert(cert);
+
+  // Build the known name for comparison
+  unsigned char* buf = nullptr;
+  X509_NAME* expected = X509_NAME_new();
+  ASSERT_TRUE(X509_NAME_add_entry_by_txt(
+      expected, "CN", MBSTRING_ASC,
+      (const unsigned char*)"intermediatecert.example.com", -1, -1, 0));
+  int len = i2d_X509_NAME(expected, &buf);
+  std::string expected_name(reinterpret_cast<char const*>(buf), len);
```
matthewstevenson88, line 646: nit: Should we call this "expected_issuer_name_der"?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,216 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+
+  OPENSSL_free(buf);
```
matthewstevenson88, line 648: nit: Can we move the "free" calls up so that they are logically grouped with the rest of the "build the known name" logic?

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,216 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+
+TEST(CrlUtils, IssuerFromLeaf) {
+  absl::StatusOr<Slice> cert_slice = LoadFile(kLeafCert, false);
+  ASSERT_EQ(cert_slice.status(), absl::OkStatus());
+  X509* cert = ReadPemCert(cert_slice->as_string_view());
+  auto issuer = IssuerFromCert(cert);
+
+  // Build the known name for comparison
+  unsigned char* buf = nullptr;
+  X509_NAME* expected = X509_NAME_new();
```
matthewstevenson88, line 641: optional: s/expected/expected_issuer_name

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,216 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+
+TEST(CrlUtils, IssuerFromIntermediateCert) {
+  absl::StatusOr<Slice> cert_slice = LoadFile(kIntermediateCrlIssuer, false);
+  ASSERT_EQ(cert_slice.status(), absl::OkStatus());
+  X509* cert = ReadPemCert(cert_slice->as_string_view());
+  auto issuer = IssuerFromCert(cert);
+
+  // Build the known name for comparison
+  unsigned char* buf = nullptr;
+  X509_NAME* expected = X509_NAME_new();
```
matthewstevenson88, line 612: optional: s/expected/expected_issuer_name

src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc
```diff
@@ -53,15 +53,23 @@ namespace grpc_core {
...
   }
-  char* buf = X509_NAME_oneline(X509_CRL_get_issuer(crl), nullptr, 0);
-  std::string ret;
-  if (buf != nullptr) {
-    ret = buf;
+  X509_NAME* issuer = X509_CRL_get_issuer(crl);
+  if (issuer == nullptr) {
+    return absl::InvalidArgumentError("crl cannot have null issuer");
   }
+  unsigned char* buf;
```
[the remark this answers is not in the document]
gtcooke94: Done

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +251,56 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+    return 1;
+  }
+  X509_NAME* crl_issuer_name = X509_CRL_get_issuer(crl);
+  if (crl_issuer_name == nullptr) {
+    return 1;
+  }
+  return X509_NAME_cmp(cert_name, crl_issuer_name);
+}
+
+bool VerifyCrlSignBit(X509* issuer) {
```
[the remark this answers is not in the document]
gtcooke94: I guess the rationale to use `issuer` here is that you are expected to pass the cert for the issuer of some crl to this API, whereas for `VerifyCrlCertIssuerNamesMatch` you'd be expected to pass in a cert and crl that were both issued by the same CA, WDYT

I think it could help someone reading the code differentiate which specific cert in the chain they are supposed to pass?

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Changed

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +251,56 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+  }
+  return (X509_get_key_usage(issuer) & KU_CRL_SIGN) != 0;
+}
+
+absl::StatusOr<std::string> IssuerFromCert(X509* cert) {
+  if (cert == nullptr) {
+    return absl::InvalidArgumentError("cert cannot be null");
+  }
+  X509_NAME* issuer = X509_get_issuer_name(cert);
+  unsigned char* buf;
```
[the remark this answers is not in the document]
gtcooke94: Fixed

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Done

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Thoughts - either way really to me. The current implementation inside of `GetIssuerFromCert` returns the DER serialization because that's the safe/stable thing to do, but there's nothing that says we _have_ to do this forever, there could be some other format in the future. So I don't think the fact it is DER is more of an implementation detail?

But I don't mind adding it if you think it makes the code more readable

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: My bad

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Done

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: See other comment

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: I removed all the blank lines because IIRC it's not in grpc style to have blank lines

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: My bad

test/core/tsi/transport_security_test_lib.h
```diff
@@ -238,4 +240,10 @@ struct SelfSignedCertificateOptions {
 std::string GenerateSelfSignedCertificate(
     const SelfSignedCertificateOptions& options);
 
+// Returns the OpenSSL representation of a PEM cert.
+X509* ReadPemCert(absl::string_view pem_cert);
+
+// Returns the OpenSSL representation of a CRL
```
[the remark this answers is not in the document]
gtcooke94: Added

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: I figured there might be a case where we'd want to differentiate between "signature failed verification" and "something weird happened and we couldn't verify the signature", which is what the int return allows for, and we'd convert in the middle layer.

But, I don't feel super strongly, and the code is definitely easier to reason with when we just return bools.
Since I don't have a compelling reason yet for the fine grained I'll swap to bools, if we need more control later we can switch back.

matthewstevenson88: Thank you!

test/core/tsi/ssl_transport_security_utils_test.cc
matthewstevenson88: My 2 cents: Specifying the cert format in the string will be helpful when we need to look back at this code. :)

test/core/tsi/ssl_transport_security_utils_test.cc
matthewstevenson88: Same here.

src/core/tsi/ssl_transport_security_utils.cc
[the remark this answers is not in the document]
matthewstevenson88: I think what you describe is one use case, but someone in the future could be interested in checking if their CA cert has this bit set, in which case they should not be looking at the issuer of the CA cert.

To that end, WDYT of changing the signature to:
```
bool HasCrlSignBit(X509* cert);
```

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Ok, will change then

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Changed

src/core/tsi/ssl_transport_security_utils.cc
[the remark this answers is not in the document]
gtcooke94: That sounds good

src/core/tsi/ssl_transport_security_utils.h
```diff
@@ -142,6 +144,22 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+// Verifies that `crl` was signed by `issuer.
+// return: true if valid, false otherwise.
+bool VerifyCrlSignature(X509_CRL* crl, X509* issuer);
+
+// Verifies the CRL issuer and certificate issuer name match.
+// return: true if equal, false if not.
+bool VerifyCrlCertIssuerNamesMatch(X509_CRL* crl, X509* cert);
+
+// Verifies the certificate in question has the cRLSign bit present.
+// return: true if cRLSign bit is present, 0 otherwise.
```
matthewstevenson88, line 156: nit: s/0 otherwise/false otherwise

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +251,55 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+    return false;
+  }
+  EVP_PKEY* ikey = X509_get_pubkey(issuer);
+  if (ikey == nullptr) {
+    // Can't verify signature because we couldn't get the pubkey, fail the
+    // check.
+    EVP_PKEY_free(ikey);
+    return false;
+  }
+  int ret = X509_CRL_verify(crl, ikey);
```
matthewstevenson88, line 265: nit: Since `X509_CRL_verify` can return values other than 0 and 1, should we change this to the following?

```
bool ok = (X509_CRL_verify(crl, ikey) == 1);
return ok;
```

matthewstevenson88:
Generally looks good, modulo the 2 open comments and the failing Windows tests (which seem to be due to this CL).

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: Done

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Done

matthewstevenson88 (approved):

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: I believe this has been resolved for a time, please double check

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: With the changed structure here I think this comment is not relevant anymore, PTAL

gtcooke94:
@davidben  PTAL

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -247,4 +251,62 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+  if (crl_issuer_name == nullptr) {
+    return false;
+  }
+  return X509_NAME_cmp(cert_issuer_name, crl_issuer_name) == 0;
+}
+
+bool HasCrlSignBit(X509* cert) {
+// X509_get_key_usage is not in 1.0.2. Just make this return true.
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+  return true;
```
bob-beck, line 288: If my memory serves me correctly and you must support old openssl, in 1.0.2 this was just exposed and didn't have an accessor, so you can just look at cert->ex_kusage and check the flag, rather than just giving up and allowing any certificate to be used to sign a CRL.

src/core/tsi/ssl_transport_security_utils.h
```diff
@@ -142,6 +144,23 @@ tsi_result SslProtectorUnprotect(const unsigned char* protected_frames_bytes,
...
+// return: true if valid, false otherwise.
+bool VerifyCrlSignature(X509_CRL* crl, X509* issuer);
+
+// Verifies the CRL issuer and certificate issuer name match.
+// return: true if equal, false if not.
+bool VerifyCrlCertIssuerNamesMatch(X509_CRL* crl, X509* cert);
+
+// Verifies the certificate in question has the cRLSign bit present.
+// OpenSSL >= 1.1.1 return: true if cRLSign bit is present, false otherwise.
+// OpenSSL == 1.0.2 return: true always
```
bob-beck, line 157: you should fix this comment when you fix this ;)

test/core/tsi/ssl_transport_security_utils_test.cc
```diff
@@ -429,6 +447,182 @@ INSTANTIATE_TEST_SUITE_P(FrameProtectorUtil, FlowTest,
...
+
+TEST_F(CrlUtils, VerifyIssuerNameNullCrlAndCert) {
+  EXPECT_FALSE(VerifyCrlCertIssuerNamesMatch(nullptr, nullptr));
+}
+
+TEST_F(CrlUtils, HasCrlSignBitExists) { EXPECT_TRUE(HasCrlSignBit(root_ca_)); }
+
+// On OpenSSL 1.0.2 this would be expected to just pass all the time because
+// X509_get_key_usage doesn't exist, so we make it a passthrough
+#if OPENSSL_VERSION_NUMBER >= 0x10100000
```
bob-beck, line 570: as above, I think you can fix this and then make this test run unconditionally.

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Just want to double check a behavior with you
> X509_get_key_usage() returns the value of the key usage extension. If key usage is present will return zero or more of the flags: KU_DIGITAL_SIGNATURE, KU_NON_REPUDIATION, KU_KEY_ENCIPHERMENT, KU_DATA_ENCIPHERMENT, KU_KEY_AGREEMENT, KU_KEY_CERT_SIGN, KU_CRL_SIGN, KU_ENCIPHER_ONLY or KU_DECIPHER_ONLY corresponding to individual key usage bits. If key usage is absent then UINT32_MAX is returned.

UINT32_MAX will mean that every usage is seen as valid - I've done this for 1.0.2 in the case of a missing extension too, does that sound right to you

src/core/tsi/ssl_transport_security_utils.h
gtcooke94: Fixed

test/core/tsi/ssl_transport_security_utils_test.cc
gtcooke94: Removed the include guards and comments

src/core/tsi/ssl_transport_security_utils.h
bob-beck: if the test works with 1.0.2 you're good to go :)

src/core/tsi/ssl_transport_security_utils.cc
bob-beck: yeah if there's no key usage present then effectively that's what you need to do.

bob-beck (approved):

src/core/lib/security/credentials/tls/grpc_tls_crl_provider.cc
```diff
@@ -53,15 +53,22 @@ namespace grpc_core {
 namespace experimental {
 
 namespace {
-std::string IssuerFromCrl(X509_CRL* crl) {
+// TODO(gtcooke94) Move ssl_transport_security_utils to it's own BUILD target
+// and add this to it.
+absl::StatusOr<std::string> IssuerFromCrl(X509_CRL* crl) {
   if (crl == nullptr) {
     return "";
```
markdroth, line 60: Shouldn't this return a non-OK status as well?  Previously, we were failing if the returned string was empty.

gtcooke94: Oh yes, good catch, fixing it

gtcooke94: Was trying to get test coverage, and I think that, given the current codepaths it is not currently _actually_ possible to have `crl == nullptr` as an input to this function as its only caller does the same check (and this fn is just defined in the the anonymous namespace in this `.cc` file, so it's not defined elsewhere).

markdroth (approved):

PR #35648, opened 2024-01-24
[EventEngine] Add QueryExtension Interfaces for *SupportsFd

drfloob:
This PR:
* adds FD extensions to the public headers
* Adds the query extension interface to EventEngine, Listener, and Endpoint, via a new `Extensible` interface
* Refactors the PosixEventEngine to use the Extensible interface.

PR #35649, opened 2024-01-24
[EventEngine] Convert passthru_endpoint to an EventEngine::Endpoint

drfloob:


drfloob:
Depends on #35648

drfloob:
No longer necessary, the passthru endpoint was removed in https://github.com/grpc/grpc/pull/35986

PR #35659, opened 2024-01-25
[otel] Return absl::Status as a return from BuildAndRegisterGlobal

yashykt:
Just to be future-proof, I'm amending the `void` return status of `BuildAndRegisterGlobal` in `OpenTelemetryPluginBuilder` to absl::Status.

This will be backported to 1.61 as well.


test/cpp/ext/csm/csm_observability_test.cc
```diff
@@ -64,9 +64,10 @@ TEST(CsmChannelTargetSelectorTest, XdsTargetsWithTDAuthority) {
 }
 
 TEST(CsmPluginOptionTest, Basic) {
-  OpenTelemetryPluginBuilder()
-      .AddPluginOption(MakeCsmOpenTelemetryPluginOption())
-      .BuildAndRegisterGlobal();
+  ASSERT_TRUE(OpenTelemetryPluginBuilder()
```
markdroth, line 67: I suggest writing this as `EXPECT_EQ(..., absl::OkStatus())`, so that if the test fails, we get an error message indicating what status was returned.

Same for all tests.

markdroth (approved):

test/cpp/ext/csm/csm_observability_test.cc
yashykt: done

Issue #35271, opened 2023-12-11
bzlmod support


mering:
### Is your feature request related to a problem? Please describe.
Bazels bzlmod resolves fundamental problems which could not be solved with the `WORKSPACE` approach. For example, it allows to only specify direct dependencies and resolve transitive ones automatically.

### Describe the solution you'd like
One could start with the downstream patches available in the [Bazel central registry](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/grpc).

### Describe alternatives you've considered
Continue maintaining downstream patches.

### Additional context
- [Bzlmod Migration Guide](https://bazel.build/external/migration)

veblush:
Currently we don't have bandwidth on this so I put `help wanted` label until we will be able to do.

AustinSchuhBRT:
I tried the grpc rules in BCR, and got the following error:

```
ERROR: /home/austin/local/server/api/BUILD.bazel:23:16: no such package '@@[unknown repo 'com_github_grpc_grpc' requested from @@]//': The repository '@@[unknown repo 'com_github_grpc_grpc' requested from @@]' could not be resolved: No repository visible as '@com_github_grpc_grpc' from main repository and referenced by '//api:api_proto_grpc_cc'
```

`BUILD` file has:
```
load("@rules_proto//proto:defs.bzl", "proto_library")    
load("@rules_go//go:def.bzl", "go_library")    
load("@rules_go//proto:def.bzl", "go_proto_library")    
load("@grpc//bazel:cc_grpc_library.bzl", "cc_grpc_library")    
    
proto_library(    
    name = "api_proto",    
    srcs = ["api.proto"],    
    visibility = ["//visibility:public"],    
    deps = [    
        "//brt/datalog:log_events_proto",    
    ],    
)    
    
cc_proto_library(    
    name = "api_proto_cc",    
    visibility = ["//visibility:public"],    
    deps = [    
        ":api_proto",    
    ],    
)    
    
cc_grpc_library(    
    name = "api_proto_grpc_cc",    
    srcs = [    
        ":api_proto",    
    ],    
    grpc_only = True,    
    visibility = ["//visibility:public"],    
    deps = [    
        ":api_proto_cc",    
    ],    
)    
```

My `MODULE.bazel` file has:
```
bazel_dep(name = "grpc", version = "1.65.0")
```

Feels like something is still missing, probably around the decision to rename `com_github_grpc_grpc` to `grpc` inside BCR.

[...]

PR #35666, opened 2024-01-25
[Bazel] Use Python headers from rules_python

### LINKED ISSUE #19956: @local_config_python://python_include fails to build under Bazel RBE
https://github.com/grpc/grpc/pull/19926 was (unwittingly) the first time any Python target was run under Bazel RBE. It brought several `py_proto_library` targets under the `test/` directory, which caused `@local_config_python://python_include` to be built under Bazel RBE. This repository rule ([originally copied from Tensorflow](https://github.com/tensorflow/tensorflow/blob/master/third_party/py/python_configure.bzl)) pulls Python headers from the system for use while compiling the C extension. These repository rules were attempts to hack support in for building a Cython-backed C extension with Bazel. After instrumenting the Skylark a bit, it appears that the operation that copies/symlinks the Python headers from system library directories to the Bazel sandbox is failing.

Example: https://source.cloud.google.com/results/invocations/8719a22e-e12d-4bfd-b92d-edb83579e0d6/log

I manually copied in Tensorflow's up-to-date version of `local_config_python` and found that their code also appears not work under RBE.

Ideally, there would be a solution officially supported by the Bazel team as in https://github.com/bazelbuild/bazel/issues/701.

### LINKED ISSUE #24665: Bazel rules should use the python headers from the python toolchain

### LINKED ISSUE #31892: Hermetic bazel python builds
When I use bazel to build python gRPC servers (and, I suppose, clients), I get "undefined symbol" cython errors if by bazel python rules version doesn't match my system's python version. For example, when I try to "import grpc" with a bazel python version of 3.9 and a system python version of 3.10, I get the following error:
```
==================== Test output for //coultech/thor/backend:candidate_service_test:
  File "/home/philstahlfeld/.cache/bazel/_bazel_philstahlfeld/06691ff925eb71d440deffe948c90182/sandbox/linux-sandbox/3255/execroot/__main__/bazel-out/k8-opt/bin/coultech/thor/backend/candidate_service_test.runfiles/__main__/coultech/thor/backend/candidate_service_test.py", line 5, in <module>
    import grpc
  File "/home/philstahlfeld/.cache/bazel/_bazel_philstahlfeld/06691ff925eb71d440deffe948c90182/sandbox/linux-sandbox/3255/execroot/__main__/bazel-out/k8-opt/bin/coultech/thor/backend/candidate_service_test.runfiles/com_github_grpc_grpc/src/python/grpcio/grpc/__init__.py", line 22, in <module>
    from grpc import _compression
  File "/home/philstahlfeld/.cache/bazel/_bazel_philstahlfeld/06691ff925eb71d440deffe948c90182/sandbox/linux-sandbox/3255/execroot/__main__/bazel-out/k8-opt/bin/coultech/thor/backend/candidate_service_test.runfiles/com_github_grpc_grpc/src/python/grpcio/grpc/_compression.py", line 15, in <module>
ImportError: /home/philstahlfeld/.cache/bazel/_bazel_philstahlfeld/06691ff925eb71d440deffe948c90182/sandbox/linux-sandbox/3255/execroot/__main__/bazel-out/k8-opt/bin/coultech/thor/backend/candidate_service_test.runfiles/com_github_grpc_grpc/src/python/grpcio/grpc/_cython/cygrpc.so: undefined symbol: PyDescr_IsData
```
If I switch my bazel python version to 3.10, the test passes. This indicates that the cython rules are looking at a locally installed version of python and not the one that will be used in the `py_test` or `py_binary` rules.

The cython rules seem to use the [local python headers](https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/bazel/cython_library.bzl#L74).

Performing a `bazel test` or a `bazel run` should pass/fail independently of the locally installed python version. 

One option I've explored is using pip to install the grpcio package and that works fine if I just wanted to do "import grpc". However, I want to use the `py_proto_library` and `py_grpc_library` targets provided by this repo.

### Additional context
Add any other context about the feature request here.
I'm happy to implement this if someone from the gRPC team can approve my design. When adding the gRPC rules to the WORKSPACE, I recommend adding an optional `python_headers` parameter to `grpc_deps` so the WORKSPACE would look as follows:
```
python_register_toolchains(
    name = "python3_9",
    # Available versions are listed in @rules_python//python:versions.bzl.
    # We recommend using the same version your team is already standardized on.
    python_version = "3.9",
)

http_archive(
    name = "com_github_grpc_grpc",
    ...
)

load("@com_github_grpc_grpc//bazel:grpc_deps.bzl", "grpc_deps")

grpc_deps(
    python_headers="@python3_9//:python_headers",
)

load("@com_github_grpc_grpc//bazel:grpc_extra_deps.bzl", "grpc_extra_deps")

grpc_extra_deps()
```
The `python_headers` parameter would then be passed through to the location where the headers [are bound](https://github.com/grpc/grpc/blob/master/bazel/grpc_python_deps.bzl#L35). The cython rules can then be updated to refer to `//external:python_headers` [here](https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/bazel/cython_library.bzl#L74).

Note that I've tested this out locally and it yield results that are independent from the locally installed python.

### LINKED ISSUE #37251: Unable to build Python gRPC service with Bazel
gRPC 1.63.1
python 3.12.1

macOS Sonoma 14.5

python 3.12.1
```sh
rgenter@RGENTER-M-040N ~ % gcc --version
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
rgenter@RGENTER-M-040N ~ % 
```

### What did you do?
I have a test that demonstrates that with Bazel 7.2.0, protobuf 27.1 and gRPC 1.63.1, a gRPC service written in Python fails to build on a MacBook Pro with an Apple M3 Pro chip.

My MODULE.bazel file:

```
"""Test demonstrating failure between Bazel, gRPC and Python."""
module(name = "test", version = "1.0")

bazel_dep(name = "grpc", version = "1.63.1", repo_name = "com_github_grpc_grpc")
bazel_dep(name = "protobuf", version = "27.1")

use_repo_rule("@@com_github_grpc_grpc//bazel:grpc_deps.bzl", "grpc_deps")
use_repo_rule("@@com_github_grpc_grpc//bazel:grpc_extra_deps.bzl", "grpc_extra_deps")
use_repo_rule("@@com_github_grpc_grpc//baze::grpc_python_deps.bzl", "grpc_python_deps")
```

My BUILD.bazel file:

```
package(default_visibility = ["//visibility:public"])

load("@com_github_grpc_grpc//bazel:cc_grpc_library.bzl", "cc_grpc_library")
load("@com_github_grpc_grpc//bazel:python_rules.bzl", "py_grpc_library", "py_proto_library")

proto_library(
    name = "service_proto",
    srcs = ["service.proto"],
)

cc_proto_library(
    name = "service_cc_proto",
    deps = [":service_proto"],
)

cc_grpc_library(
    name = "service_cc_grpc",
    srcs = [":service_proto"],
    grpc_only = True,
    deps = [":service_cc_proto"],
)

py_proto_library(
    name = "service_py_proto",
    deps = [":service_proto"],
)

py_grpc_library(
    name = "service_py_grpc",
    srcs = [":service_proto"],
    deps = [":service_py_proto"],
)
```

My service.proto file:

```
syntax = "proto3";

package test;

service MyService {
    rpc MyMethod(MyRequest) returns (MyResponse);
}

message MyRequest {
    optional string request = 1;
}

message MyResponse {
    optional string response = 1;
}
```

When I do:

```sh
$ bazel clean --expunge
$ bazel build //:service_py_grpc
```

I get:

```
ERROR: /private/var/tmp/_bazel_rgenter/a1a8280059cd89e552f80330d54f1415/external/grpc~/src/python/grpcio/grpc/_cython/BUILD.bazel:25:12: Linking external/grpc~/src/python/grpcio/grpc/_cython/cygrpc.so failed: (Exit 1): cc_wrapper.sh failed: error executing CppLink command (from target @@grpc~//src/python/grpcio/grpc/_cython:cygrpc.so) external/bazel_tools~cc_configure_extension~local_config_cc/cc_wrapper.sh @bazel-out/darwin_arm64-fastbuild/bin/external/grpc~/src/python/grpcio/grpc/_cython/cygrpc.so-2.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
Undefined symbols for architecture arm64:
  "_PyArg_UnpackTuple", referenced from:
      __Pyx_Coroutine_Throw(_object*, _object*) in cygrpc.o
      __Pyx_async_gen_athrow_send(__pyx_PyAsyncGenAThrow*, _object*) in cygrpc.o
  "_PyAsyncGen_Type", referenced from:
      __Pyx_PyGen_Send(PyGenObject*, _object*) in cygrpc.o
  "_PyBaseObject_Type", referenced from:
      __Pyx_InBases(_typeobject*, _typeobject*) in cygrpc.o
      __Pyx_setup_reduce(_object*) in cygrpc.o
      __pyx_tp_new_7_cython_6cygrpc__ChannelArg(_typeobject*, _object*, _object*) in cygrpc.o
      __pyx_tp_new_7_cython_6cygrpc__ChannelArgs(_typeobject*, _object*, _object*) in cygrpc.o
      __pyx_tp_new_7_cython_6cygrpc_Call(_typeobject*, _object*, _object*) in cygrpc.o
      __pyx_tp_new_7_cython_6cygrpc__CallState(_typeobject*, _object*, _object*) in cygrpc.o
      __pyx_tp_new_7_cython_6cygrpc__ChannelState(_typeobject*, _object*, _object*) in cygrpc.o
      ...
```

followed by several thousand more lines complaining about undefined symbols, all of which start with _Py or __Py. This seems to come from Cython, which gRPC apparently depends upon for Python.

If I try to build a C++ service using:

```sh
$ bazel clean --expunge
$ bazel build //:service_cc_grpc
```

The build succeeds.

A successful build.

### What did you see instead?

See above.

Nothing that I can think of. I'm happy to answer questions.

NOTE: This is also filed as https://github.com/grpc/grpc.io/issues/1307

mering:
This has several advantages like choosing the headers from the correct runtime (see bazelbuild/rules_python#1287) and is a requirement for #35271.

This feature was added in `rules_python` v0.24.0.

CC @rickeylev who implemented this feature in rules_python and suggested this approach.

Fixes https://github.com/grpc/grpc/issues/24665
Fixes https://github.com/grpc/grpc/issues/31892
Fixes https://github.com/grpc/grpc/issues/19956
Fixes https://github.com/grpc/grpc/issues/37251


gnossen:
I'm incredibly excited to find out that this feature has been added to `rules_python`. This had previously looked like it would require some difficult upstream changes.

I suspect that there may be some backward-compatibility concerns we need to look at, like "What versions of Bazel will this work with? Can we fall back to the old behavior for any versions that don't support it?" and "Can we fall back to the old behavior if the project is using an older version of rules_python that does not support `current_py_cc_headers`". We may be able to drop backwards compatibility in some cases, but if so, we'll need to send out a notice on the `grpc-io` mailing list first. I'm happy to help work through these issues.

rickeylev:
In theory, it should work with Bazesl 5.4 and later. It's a pretty basic toolchain and was introduced when rules_python still supported Bazel 5.

However, rules_python no longer tests against Bazel 5 or supports it[1]. I don't recall any changes we've made that would break Bazel 5, but can't be certain. More likely, some other part of rules_python will fall over for an earlier Bazel version.

> Can we fall back to the old behavior for any versions that don't support it?

If you can version detect on rules_python, I suppose so. There isn't a direct way to do that. I suppose a repo rule could look for the `@rules_python//python/cc/BUILD.bazel` file. That would indicate whether this code is present.

What might work better is to have a repo rule figure out the situation, then write something to a bzl file, and then you load that and use it as the target. This is also a good way to use an environment variable to control things for opt-in/out type of behavior.

Note that this feature first appears in 0.24.0 and is automatically available for bzlmod builds, but isn't automatically available workspace builds until 0.28.0 (see https://github.com/bazelbuild/rules_python/issues/1669). For workspace builds, between rules_python 0.24.0 and 0.28.0, a user just has to manually call e.g. `register_toolchains("@python_3_11_linux_x86_64//:py_cc_toolchain")` (or something like that, I forget the exact string) in their workspace config (all the other code was there, I just forgot the line to actually _register_ the toolchain in workspace builds :facepalm: )

HTH!

[1] Best effort basis, patches might be accepted if they're reasonable.

mering:
I added removal of `local_config_python` to this PR as it appears to me that this is not needed any more with this PR. When searching for `python_headers` string in this repository it seems that there is even  more clean up potential.

gnossen:
It looks like there are still some build issues here:

```
ERROR: /root/.cache/bazel/_bazel_root/954bb7512d44d20015390af6e76121c6/external/rules_python/python/cc/BUILD.bazel:14:22: While resolving toolchains for target @rules_python//python/cc:current_py_cc_headers: No matching toolchains found for types @rules_python//python/cc:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@rules_python//python/cc:toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
ERROR: Analysis of target '//src/python/grpcio_observability/grpc_observability:_cyobservability.so' failed; build aborted: 
INFO: Elapsed time: 22.436s
INFO: 0 processes.
```

It looks like we'll have to actually add a toolchain in our WORKSPACE file.

rickeylev:
> It looks like there are still some build issues here:
> No matching toolchains found for types @rules_python//python/cc:toolchain_type.

This sounds like the bug fixed in 0.28.0 I mentioned.

* Can you verify the version of rules_python being used in the build?
* Can you link to a gist with the output of running with `--toolchain_resolution_debug=.*python.*` ? It's probably going to be relatively short output that shows it doesn't see anything with the //python/cc:toolchain_type identifier.

> Add a C++ toolchain

You had C++ building working already, right? So your C++ toolchain part is probably fine.

The py cc toolchain thing it can't find is the C-side of the python runtime information (there's two rules_python toolchain types: //python:toolchain_type, which is information about the interpreter to run a program, and //python/cc:toolchain_type, which is C information for building an interpreter or C extensions for one)

mering:
@rickeylev Thanks for the hint towards updating `rules_python`. I updated to the latest version in my branch.

@gnossen can you re-trigger the tests?

mering:
Looks like gRPC depends on Protobuf's `@system_python`:
https://github.com/grpc/grpc/blob/bb3edd22b36fb7aaea53666c8b010c2a9e316c80/WORKSPACE#L75-L78

which in turn depends on `@rules_python//python:pip.bzl` `pip_install` which was removed in `rules_python` 0.28.0:
https://github.com/protocolbuffers/protobuf/blob/0ce48b30b5c925fc2e30fa2d8f234fc9e3611276/bazel/system_python.bzl#L39

mering:
This has finally been merged to Protobuf in https://github.com/protocolbuffers/protobuf/pull/16942. Waiting for the next release to unblock this PR.

agronskiy:
Hey @mering, come here through https://github.com/grpc/grpc/issues/19956 as the latter blocks the usage of RBE, what's your take on the future of this fix? Thank you! Otherwise feels like I'd need to patch on the fly?

mering:
Protobuf dependency has been updated to v28 in #37682, so this is finally ready to go now!

mering:
@gnossen @veblush could you PTAL? Thanks!

veblush:
Thanks! LGTM but I'll let @gnossen approve this.

gnossen:
It looks like there are [some build issues](https://btx.cloud.google.com/invocations/fe9c2599-a2b3-4241-82d5-988044d7ae87/log):

```
ERROR: /root/.cache/bazel/_bazel_root/954bb7512d44d20015390af6e76121c6/external/rules_python/python/cc/BUILD.bazel:26:19: While resolving toolchains for target @@rules_python//python/cc:current_py_cc_libs (cf24561): No matching toolchains found for types @@rules_python//python/cc:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@@rules_python//python/cc:toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
ERROR: Analysis of target '//src/python/grpcio/grpc/_cython:cygrpc.so' failed; build aborted: Analysis failed
INFO: Elapsed time: 14.959s, Critical Path: 0.75s
INFO: 48 processes: 1 internal, 47 processwrapper-sandbox.
ERROR: Build did NOT complete successfully
FAILED: 
ERROR: No test targets were found, yet testing was requested
```

mering:
@gnossen This sounds like the issue we observed previously and which is supposed to be fixed by `rules_python` 0.28.0. Could you please follow the instructions of @rickeylev in https://github.com/grpc/grpc/pull/35666#issuecomment-1944257958 to debug this further? Maybe the test is still using an old version of `rules_python` specified in a place which I didn't find?

bazel/cython_library.bzl
```diff
@@ -72,7 +72,10 @@ def pyx_library(name, deps = [], py_deps = [], srcs = [], **kwargs):
         native.cc_binary(
             name = shared_object_name,
             srcs = [stem + ".cpp"],
-            deps = deps + ["@local_config_python//:python_headers"],
+            deps = deps + [
+            "@rules_python//python/cc:current_py_cc_headers",
+            "@rules_python//python/cc:current_py_cc_libs",
```
rickeylev, line 77: You probably don't want py_cc_libs here -- that is going to try and link in the libpython.so files. The previous code, if its name is to be believed, just gets the headers (which is what you want -- lazy binding of the runtime symbols to come from whatever the py runtime the extension is used with).

Ah, actually, looking at variety.tpl, I think what it was doing is more nuanced. If I'm reading that variety.tpl code correctly, `:python_headers` was just the headers for non-windows, but the headers and cc_import(<libs>) for windows. I think the equivalent here would be:

```
deps = [current_py_cc_headers] + select({
"windows": [current py cc libs],
"default: []
})
```

This probably isn't _the_ problem, but would manifest as the sort of weird behavior ODR violates can result in.

rickeylev:
Marcel asked me to take a closer look at this -- I tried reading through some of the code, but it's been hard to be entirely sure of the workspace setup being used.

I'm getting the feeling that the normal toolchains rule_python assumes (the hermetic ones, aka python-build-standalone) aren't being setup, though. In particular, I couldn't find the call to py_repositories() or any python_register_toolchains() calls. This would mean it falls back to `@bazel_tools//tools/python:autodetecting_toolchain`, which, despite its name, doesn't autodetect anything, and doesn't set up any of the pycc toolchains that you need. To verify this, I need the output from running with `--toolchain_resolution_debug=.*` (stick it in a gist -- it'll be very long). This will, at the least, confirm for us what toolchains are being used or not used.

Two main suggestions I have to try are:
(1) Call py_repositories() and python_register_toolchains in workspace. See docs here: https://rules-python.readthedocs.io/en/latest/toolchains.html#workspace-configuration (that page also talks about how to setup a custom toolchain, fwiw)

(2) Implement your own py cc toolchain by:
* Restore python_configure.bzl et al.
* As part of `python_configure()`, instantiate the same set of targets as seen here: https://github.com/bazelbuild/rules_python/blob/main/python/private/local_runtime_repo_setup.bzl
  python_configure() already creates the cc_library; the missing part is the py_cc_toolchain() part. That can be loaded from `@rules_python//python/cc:py_cc_toolchain.bzl` (if memory serves)
* In workspace, call `register_toolchains("@local_config_python//:all")`


I'll have to find some time on the weekend to unwind things and diagnose further.

Final note: a "secret" feature in rules_python 0.35.0+ is some code to use a locally installed python. Basically what's in that python_configure.bzl. Something you can try is something like this: https://github.com/bazelbuild/rules_python/blob/main/tests/integration/local_toolchains/MODULE.bazel

That's a MODULE file, but its using the repo-rules directly, as workspace would do, so it should be easily translatable to workspace and work. In workspace, load the names in the use_repo calls and copy/paste the invocations and register_toolchains() call. You'll need to load some private paths to try this out. If it _does_ work, though, let me know, and we can create a public API for them.

bazel/cython_library.bzl
mering: Indeed originally, I only depended on the headers but pulled in the patch from https://github.com/bazelbuild/bazel-central-registry/pull/2685 later.

I used your `select()` suggestion. Let's see what the tests are doing with this.

mering:
> Marcel asked me to take a closer look at this -- I tried reading through some of the code, but it's been hard to be entirely sure of the workspace setup being used.
> 
> I'm getting the feeling that the normal toolchains rule_python assumes (the hermetic ones, aka python-build-standalone) aren't being setup, though. In particular, I couldn't find the call to py_repositories() or any python_register_toolchains() calls. This would mean it falls back to `@bazel_tools//tools/python:autodetecting_toolchain`, which, despite its name, doesn't autodetect anything, and doesn't set up any of the pycc toolchains that you need. To verify this, I need the output from running with `--toolchain_resolution_debug=.*` (stick it in a gist -- it'll be very long). This will, at the least, confirm for us what toolchains are being used or not used.
> 
> Two main suggestions I have to try are: 
> (1) Call py_repositories() and python_register_toolchains in workspace. See docs here: https://rules-python.readthedocs.io/en/latest/toolchains.html#workspace-configuration (that page also talks about how to setup a custom toolchain, fwiw)
> 
> (2) Implement your own py cc toolchain by:
> 
> * Restore python_configure.bzl et al.
> * As part of `python_configure()`, instantiate the same set of targets as seen here: https://github.com/bazelbuild/rules_python/blob/main/python/private/local_runtime_repo_setup.bzl
>   python_configure() already creates the cc_library; the missing part is the py_cc_toolchain() part. That can be loaded from `@rules_python//python/cc:py_cc_toolchain.bzl` (if memory serves)
> * In workspace, call `register_toolchains("@local_config_python//:all")`
> 
> I'll have to find some time on the weekend to unwind things and diagnose further.
> 
> Final note: a "secret" feature in rules_python 0.35.0+ is some code to use a locally installed python. Basically what's in that python_configure.bzl. Something you can try is something like this: https://github.com/bazelbuild/rules_python/blob/main/tests/integration/local_toolchains/MODULE.bazel
> 
> That's a MODULE file, but its using the repo-rules directly, as workspace would do, so it should be easily translatable to workspace and work. In workspace, load the names in the use_repo calls and copy/paste the invocations and register_toolchains() call. You'll need to load some private paths to try this out. If it _does_ work, though, let me know, and we can create a public API for them.

@rickeylev Thanks a lot for your extensive reply and pointers!

The `py_repositories()` call is in https://github.com/grpc/grpc/blob/dbb5164ac7322d2a396675aab9cee7979203126d/bazel/grpc_extra_deps.bzl#L76

bazel/cython_library.bzl
mering: Reading the original bug more closely, it appeared on MacOS, so only linking against it on Windows probably won't be sufficient. I reverted the `select()` again but registered a hermetic toolchain as per the docs to replace the previous custom `python_configure()`.

mering:
I just found out that pybind11_bazel used the same python_configure.bzl previously and migrated to hermetic Python in https://github.com/pybind/pybind11_bazel/commit/f77df954ab235f7487d284abe68ebe429d6f7b2b.

sergiitk:
Superseded by #41684

PR #35681, opened 2024-01-26
[Python O11Y] Update deps file when building new abseil-cpp

XuanWang-Amos:
* Generate observability and grpcio-tools dependency file when building project.
* Also removed deps file from black formatter.


gnossen (approved):
LGTM. But be aware that `generate_projects.sh` is run manually by C++ developers, so if we've accidentally introduced a dependency here, we may have to respond to any issues this introduces for them.

PR #35687, opened 2024-01-27
[Build Reproducibility] Sort input file list

bmwiedemann:
Sort input file list
so that the `.so` file builds in a reproducible way
in spite of non-deterministic filesystem readdir order and http://bugs.python.org/issue30461

See https://reproducible-builds.org/ for why this is good.

This patch was done while working on reproducible builds for openSUSE.


gnossen (approved):

gnossen:
Thanks for the fix!

veblush (approved):

PR #35776, opened 2024-01-31
[EventEngine] Fix HttpCli test to deail with EventEngine async execution

drfloob:
One test (`CallerPollentsAreNotReferencedAfterCallbackIsRan`) was not waiting for a callback to be run before finishing when the EventEngine client is in use. With the iomgr implementation, `exec_ctx.Flush();` previously guaranteed that the callback would be executed before the test ended, but the EventEngine runs it in a separate thread, with the EventEngine shim providing its own isolated ExecCtx. This fix simply waits for the callback to be run before exiting, in the same way many other tests in this file wait on iomgr-based callback execution.

### DIFF
--- a/test/core/http/httpcli_test.cc
+++ b/test/core/http/httpcli_test.cc
@@ -43,7 +43,6 @@
 
 #include "src/core/ext/filters/client_channel/resolver/dns/c_ares/grpc_ares_wrapper.h"
 #include "src/core/lib/gpr/subprocess.h"
-#include "src/core/lib/gprpp/notification.h"
 #include "src/core/lib/gprpp/status_helper.h"
 #include "src/core/lib/gprpp/time.h"
 #include "src/core/lib/gprpp/time_util.h"
@@ -148,7 +147,6 @@ struct RequestState {
   bool done = false;
   grpc_http_response response = {};
   grpc_pollset_set* pollset_set_to_destroy_eagerly = nullptr;
-  grpc_core::Notification callback_finished;
 };
 
 void OnFinish(void* arg, grpc_error_handle error) {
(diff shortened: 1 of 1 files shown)


apolcyn (approved):

markdroth (approved):
Thanks for fixing this!

markdroth (approved):

test/core/http/httpcli_test.cc
```diff
@@ -475,11 +475,19 @@ TEST_F(HttpRequestTest, CallerPollentsAreNotReferencedAfterCallbackIsRan) {
   http_request->Start();
   exec_ctx.Flush();
   http_request.reset();  // cancel the request
+  // With iomgr polling:
   // Since the request was cancelled, the on_done callback should be flushed
   // out on the ExecCtx flush below. When the on_done callback is ran, it will
```
apolcyn, line 480: note the part about eagerly destroying the pollset from within the callback is the same for both iomgr and EE (that's the crux of this test)

Maybe worth rewording:

```
// When the on_done callback is ran, it will
// eagerly destroy 'request_state.pollset_set_to_destroy_eagerly'. PollUntil's
// predicate should return true immediately.
//
// With iomgr polling:
// Since the request was cancelled, the on_done callback should be flushed
// out on the ExecCtx flush below.
// With EventEngine polling:
// Since the callback will be run asynchronously in another thread, with an
// independent ExecCtx, PollUntil is used here to ensure this test does not
// finish before the callback is run.


```

apolcyn (approved):

test/core/http/httpcli_test.cc
drfloob: Great point, thanks. Changed.

PR #35777, opened 2024-01-31
Allow overriding the QPS server target irrespective of the number of servers.

AshZhang:
This change enables test scenarios that involve name resolution, where multiple QPS server instances are load balanced behind a single hostname that the QPS clients must resolve. All of the server instances' IP addresses would be passed into the `QPS_WORKERS` environment variable so that the QPS server will be started on all server worker instances, and the single hostname would be passed in as the `qps_server_target_override` for clients to use.


AshZhang:
CC: @anicr7

anicr7 (approved):

test/cpp/qps/driver.cc
```diff
@@ -470,9 +470,6 @@ std::unique_ptr<ScenarioResult> RunScenario(
     }
   }
   if (qps_server_target_override.length() > 0) {
-    // overriding the qps server target only makes since if there is <= 1
-    // servers
-    GPR_ASSERT(num_servers <= 1);
```
drfloob, line 475: This is an unusual use case, and deserves some documentation. Having a server count > 1 and a only single overridden server target only makes sense if the target resolves to that many servers.

drfloob: I'm also curious how this would work with server metric reporting. Wouldn't only one server report back?

AshZhang: I've added some documentation and comments explaining the proper intended usage of `qps_server_target_override` when there are multiple servers.

In summary, if there are multiple servers and `qps_server_target_override` is specified, then the overriding target must load balance among all the servers; otherwise, the usage is incorrect. Only one server of the target, as chosen by the load balancer, should respond and report metrics for each client request.

PR #35789, opened 2024-02-01
[promise] add optional is_acceptable callback to Observable

markdroth:


src/core/lib/promise/observable.h
```diff
@@ -37,7 +38,12 @@ class Observable {
   void Set(T value) { state_->Set(std::move(value)); }
 
   // Returns a promise that resolves to a T when the value becomes != current.
-  auto Next(T current) { return Observer(state_, std::move(current)); }
+  // If a new value is available but is_acceptable is non-null and returns
+  // false for that value, does not return that value (i.e., remains pending).
+  auto Next(T current,
```
ctiller, line 43: I think this ends up branchier than what I'd like to see at this layer... we end up with an indirect call per poll, and an if to check whether it's available, and possibly an hard to spot allocation if `is_acceptable` is phrased a little bit wrong.

Instead, let's add a new variant of `Next`:

```
template <typename F>
auto NextWhen(F is_acceptable) {
  return ObserverWhen<F>(std::move(is_acceptable));
}
```

With `ObserverWhen` being templated on the function that's checking acceptability.

This eliminates any additional unnecessary branches and indirections and possibilities of allocation from this critical path.

ctiller: (this probably means making Observer an interface so that we can track it in State, but I think that's going to be the better tradeoff)

markdroth: I changed it to use a separate `NextWhen()` method and a separate `ObserverWhen` class.  However, I didn't use a templated callback type, because we need to store pointers to the observers in the `State` object, and if those observers are different types, then we would also need to template the `State` object, which basically means we'd need to template the entire `Observable` object.

The only reasonable way I see of doing this would be to have a base `Observer` type that we use in `State`, and then have two subclasses of that type that we instantiate in `Next()` and `NextWhen()`.  I wasn't sure if you'd be comfortable with a virtual method call in this code path, but if you'd prefer that approach, I can make those changes.

markdroth: Changed `Observer` to be virtual, as per our discussion.

ctiller (approved):

PR #35799, opened 2024-02-02
[examples] Update SSA client example

eugeneo:
1. Update cookies after all successful RPCs. Ensures that client adapts to changes in cluster configuration.
2. Reorganized client code to extract more snippets that could be copy-pasted.
3. Docker build now uses the same bazel build as outside code.


yashykt (approved):
Thanks!

PR #35803, opened 2024-02-03
[CSM] Modify CsmObservability to CsmOpenTelemetryPluginOption internally

yashykt:
Changes -
* `CsmObservability` API will now use the `CsmOpenTelemetryPluginOption` internally. After this change, `CsmObservability` will enable observability for all channels and servers. (Earlier, `CsmObservability` only enabled observability for CSM-enabled channels and servers.) CSM labels will still be added just for CSM-enabled channels and servers. 
* Also, we no longer need the ability to set `LabelInjector` on the `OpenTelemetryPluginBuilder` directly. Instead, we always use `PluginOption` to inject the `LabelInjector`. This simplifies the code as well.

Note that `SetTargetSelector` and `SetServerSelector` APIs on the `OpenTelemetryPluginBuilderImpl` are not being deleted yet since we might need them shortly. This is also why `OpenTelemetryPluginBuilderImpl` is not being deleted right now.

### DIFF
--- a/test/cpp/ext/csm/metadata_exchange_test.cc
+++ b/test/cpp/ext/csm/metadata_exchange_test.cc
@@ -127,7 +127,8 @@ class TestScenario {
 class MeshLabelsPluginOption
     : public grpc::internal::InternalOpenTelemetryPluginOption {
  public:
-  MeshLabelsPluginOption(const opentelemetry::sdk::common::AttributeMap& map)
+  explicit MeshLabelsPluginOption(
+      const opentelemetry::sdk::common::AttributeMap& map)
       : labels_injector_(
             std::make_unique<grpc::internal::ServiceMeshLabelsInjector>(map)) {}
 
--- a/test/cpp/ext/csm/metadata_exchange_test.cc
+++ b/test/cpp/ext/csm/metadata_exchange_test.cc
@@ -132,11 +132,11 @@ class MeshLabelsPluginOption
       : labels_injector_(
             std::make_unique<grpc::internal::ServiceMeshLabelsInjector>(map)) {}
 
-  bool IsActiveOnClientChannel(absl::string_view target) const override {
+  bool IsActiveOnClientChannel(absl::string_view /*target*/) const override {
     return true;
   }
 
-  bool IsActiveOnServer(const grpc_core::ChannelArgs& args) const override {
+  bool IsActiveOnServer(const grpc_core::ChannelArgs& /*args*/) const override {
     return true;
   }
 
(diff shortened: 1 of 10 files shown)


test/cpp/ext/csm/metadata_exchange_test.cc
```diff
@@ -121,6 +121,32 @@ class TestScenario {
   XdsBootstrapSource bootstrap_source_;
 };
 
+// A PluginOption that injects `ServiceMeshLabelsInjector`. (This is different
+// from CsmOpenTelemetryPluginOption since it does not restrict itself to just
```
markdroth, line 125: Why does this difference matter for this test?  Are we not using a channel or server that considers itself a CSM channel or server?

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -775,19 +775,15 @@ class CustomPluginOption
...
-      /*enabled_on_client*/ true, /*enabled_on_server*/ true,
-      std::make_pair("key", "value")));
   Init(
       std::move(Options()
                     .set_metric_names({grpc::OpenTelemetryPluginBuilder::
                                            kClientAttemptDurationInstrumentName,
                                        grpc::OpenTelemetryPluginBuilder::
                                            kServerCallDurationInstrumentName})
-                    .set_plugin_options(std::move(plugin_option_list))));
+                    .add_plugin_option(std::make_unique<CustomPluginOption>(
```
markdroth, line 784: I like the change to use `add_plugin_option()` instead of having to construct a vector!

markdroth (approved):
Just one question about a test, otherwise looks great!

test/cpp/ext/csm/metadata_exchange_test.cc
yashykt: No, this test does not use xds enabled channels or servers. 

Benefits of the current test approach - I can verify the workings of `ServiceMeshLabelsInjector` since I'm creating it directly and I'm able to pass in a resource that can behave as various environments (GCE, GKE, CloudRun, etc.)
For convenience, this doesn't use xds, but instead injects the labels that we want from xDS through a separate method. 
We also have separate tests within the xds end2end test suite that check that the service name and namespace labels are properly propagated to a fake stats plugin. (Gives the additional benefit of not depending on OTel from within the xds end2end test suite.)

We still get end-to-end test coverage using CsmObservability from the CI tests with GMP.

I'm thinking it'll still be nice to have an end2end test integrated with the xds end2end test library that uses the CSM O11y API as-is so that we can locally verify basic CSM functionality atleast without waiting on interop tests.

PR #35812, opened 2024-02-05
[CSM] Remove experimental CSM PluginOption API in favor of CsmObservability API

yashykt:
Also update interop tests to use `CsmObservability` API

### DIFF
--- a/test/cpp/interop/xds_interop_client.cc
+++ b/test/cpp/interop/xds_interop_client.cc
@@ -427,7 +427,7 @@ void RunTestLoop(std::chrono::duration<double> duration_per_query,
   GPR_UNREACHABLE_CODE(thread.join());
 }
 
-void EnableCsmObservability() {
+grpc::experimental::CsmObservability EnableCsmObservability() {
   gpr_log(GPR_DEBUG, "Registering Prometheus exporter");
   opentelemetry::exporter::metrics::PrometheusExporterOptions opts;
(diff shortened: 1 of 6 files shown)


yijiem (approved):

test/cpp/interop/xds_interop_client.cc
```diff
@@ -438,12 +438,11 @@ void EnableCsmObservability() {
...
-             .AddPluginOption(
-                 grpc::experimental::MakeCsmOpenTelemetryPluginOption())
-             .SetMeterProvider(std::move(meter_provider))
-             .BuildAndRegisterGlobal()
-             .ok());
+  auto observability = grpc::experimental::CsmObservabilityBuilder()
+                           .SetMeterProvider(std::move(meter_provider))
+                           .BuildAndRegister();
+  assert(observability.ok());
+  return *observability;
```
markdroth, line 445: This makes me realize that we should change `CsmObservability` to be a move-only type.  If we allow it to be copyable, then if we want to add something like a `unique_ptr<>` to it later, we will break existing code.

markdroth (commented):
Looks good, but it made me notice a problem that we need to fix, so let's do that at the same time.

test/cpp/interop/xds_interop_client.cc
yashykt: oh wow, that was a really good catch. Thanks Mark!

markdroth (approved):

PR #23385, opened 2020-07-04
Remove wrap_memcpy

veblush:
gRPC has been having the `wrap_memcpy` workaround to handle the breaking change of `memcpy` on glibc 2.14 ([ref1](https://www.win.tue.nl/~aeb/linux/misc/gcc-semibug.html), [ref2](https://stackoverflow.com/questions/35656696/explanation-of-memcpy-memmove-glibc-2-14-2-2-5)). This was necessary to build more portable artifacts which are expected to run on most linux distributions and we manged to have https://github.com/grpc/grpc/pull/5007 years ago for this.

Since we started to use manylinux2010-based docker images for artifacts, however, this became unnecessary since CentOS 6 has glibc 2.12 which is older than 2.14 so it doesn't have memcpy issue, which is another benefit of using manylinux2010.

This shouldn't be merged before #23344 trying to switch ruby's base image to manylinux2010 but it's been merged now.

[grpc_build_artifacts_multiplatform](https://g3c.corp.google.com/results/invocations/97118688-f212-4e66-ba53-5198a1fc15de): passed

PR #35826, opened 2024-02-06
[Core] Remove wrap_memcpy

veblush:
gRPC has been having the `wrap_memcpy` workaround to handle the breaking change of `memcpy` on glibc 2.14 ([ref1](https://www.win.tue.nl/~aeb/linux/misc/gcc-semibug.html), [ref2](https://stackoverflow.com/questions/35656696/explanation-of-memcpy-memmove-glibc-2-14-2-2-5)). This was necessary to build more portable artifacts which are expected to run on most linux distributions and we manged to have https://github.com/grpc/grpc/pull/5007 years ago for this.

Since we started to use manylinux2010-based docker images for artifacts, however, this became unnecessary since CentOS 6 has glibc 2.12 which is older than 2.14 so it doesn't have memcpy issue, which is another benefit of using manylinux2010.

Superseding https://github.com/grpc/grpc/pull/23385


apolcyn (approved):
one nit on the description: ruby builds off [manylinux2014 images](https://github.com/rake-compiler/rake-compiler-dock/blob/ef73c80e6fe4e5601bdec6dc9b90a1df13e0349b/Dockerfile.mri.erb#L4) now, (but I believe that shouldn't change anything)

PR #35831, opened 2024-02-06
[Python Distribution] Add a check for file size

XuanWang-Amos:
Add a step in distribution test which checks if any of the artifacts exceeds 80 MB since Pypi has a per-file size limit.

### DIFF
--- a/tools/internal_ci/helper_scripts/check_python_artifacts_size.sh
+++ b/tools/internal_ci/helper_scripts/check_python_artifacts_size.sh
@@ -18,7 +18,7 @@
 
 set -ex
 
-find . -path "*/artifacts/*" -size +10M | egrep '.*' && echo "Found Python artifacts larger than 100 MB." && FAILED="true"
+find . -path "*/artifacts/*" -size +80M | egrep '.*' && echo "Found Python artifacts larger than 80 MB." && FAILED="true"
 
 
 if [ "$FAILED" != "" ]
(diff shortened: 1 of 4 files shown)


tools/internal_ci/helper_scripts/check_python_artifacts_size.sh
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Checks if any of the Python artifacts exceeds a certern size limit since
+# Pypi has a per-file size limit.
+
+set -ex
+
+find . -path "*/artifacts/*" -size +80M | egrep '.*' && echo "Found Python artifacts larger than 80 MB." && FAILED="true"
```
gnossen, line 21: Since the artifacts are currently over 80MB, we have two options:

1. Temporarily bump this to 110MB and have a TODO (which we actually follow up on) or
2. Wait to merge this PR until we actually fix the file size.

It's up to you which one to go with.

gnossen (approved):

tools/internal_ci/helper_scripts/check_python_artifacts_size.sh
XuanWang-Amos: Since this PR won't be useful until next release, I'll merge it after we fix the size issue.

gnossen:
One half of the long-term solution for the mitigation provided by https://github.com/pypi/support/issues/3616

gnossen:
Now that https://github.com/grpc/grpc/pull/35832 is in, this should be good to go.

PR #35832, opened 2024-02-06
[Build] Strip armv7 artifacts

gnossen:


gnossen:
Manual run: https://fusion2.corp.google.com/invocations/81e1ad8a-35c0-4537-96e4-8f5c02bc7c8f/targets

gnossen:
Looks like we managed to shrink the artifacts by about 25x.

This is the other half of the long-term solution for the problem mitigated by https://github.com/pypi/support/issues/3616

eugeneo (approved):

PR #35835, opened 2024-02-07
[CSM] Disable metrics recording when CsmObservability goes out of scope

yashykt:
As discussed, this change adds scoping to `CsmObservability` such that when that object goes out of scope, new channels and servers don't record metrics. In the documentation, I've talked about how existing channels/servers are going to continue to record metrics but i've left room for us to change that behavior in the future.

The current way of doing this is through a global bool since there can only be one plugin right now, but we'll change this to use the global stats plugin registry in the future.

### DIFF
--- a/include/grpcpp/ext/csm_observability.h
+++ b/include/grpcpp/ext/csm_observability.h
@@ -38,11 +38,13 @@ class OpenTelemetryPluginBuilderImpl;
 
 namespace experimental {
 
-// This is a no-op at present, but in the future, this object would be useful
-// for performing cleanup.
+// When this object goes out of scope, CsmObservability will stop recording
+// metrics for new channels and servers. Existing channels and servers may
+// continue exporting metrics (implementation-dependent).
 class CsmObservability {
  public:
   CsmObservability() = default;
+  ~CsmObservability();
   // Disable copy constructor and copy-assignment operator.
   CsmObservability(const CsmObservability&) = delete;
   CsmObservability& operator=(const CsmObservability&) = delete;
--- a/src/cpp/ext/otel/otel_plugin.cc
+++ b/src/cpp/ext/otel/otel_plugin.cc
@@ -43,21 +43,6 @@
 namespace grpc {
 namespace internal {
 
-// TODO(yashykt): Extend this to allow multiple OpenTelemetry plugins to be
-// registered in the same binary.
-struct OpenTelemetryPluginState* g_otel_plugin_state_;
-
-const struct OpenTelemetryPluginState& OpenTelemetryPluginState() {
-  GPR_DEBUG_ASSERT(g_otel_plugin_state_ != nullptr);
-  return *g_otel_plugin_state_;
-}
-
-absl::string_view OpenTelemetryMethodKey() { return "grpc.method"; }
-
-absl::string_view OpenTelemetryStatusKey() { return "grpc.status"; }
-
-absl::string_view OpenTelemetryTargetKey() { return "grpc.target"; }
-
 namespace {
 absl::flat_hash_set<std::string> BaseMetrics() {
   return absl::flat_hash_set<std::string>{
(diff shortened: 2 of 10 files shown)


test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -680,6 +680,23 @@ TEST_F(OpenTelemetryPluginEnd2EndTest,
...
+  // Re-create server and channel after disabling the plugin.
+  ResetServerAndStub();
+  SendRPC();
+  auto data = ReadCurrentMetricsData(
+      [&](const absl::flat_hash_map<
+          std::string,
+          std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
+          /*data*/) { return false; });
+  // Since we've disabled the plugin, no metrics should be recorded.
+  ASSERT_TRUE(data.empty());
```
yijiem, line 697: `EXPECT_TRUE()`? `ASSERT_TRUE()` will return early which seems unnecessary here.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -80,8 +65,32 @@ absl::flat_hash_set<std::string> BaseMetrics() {
       std::string(grpc::OpenTelemetryPluginBuilder::
                       kServerCallRcvdTotalCompressedMessageSizeInstrumentName)};
 }
+
+// TODO(yashykt): This should go away when we move to the
```
yijiem, line 69: Does this mean that the `GlobalStatsPluginRegistry` will have an API to disable metrics recording for a specific (or a group of) `StatsPlugins`? What would that API look like? How can a user specify the `StatsPlugin`s that they want to disable? And would that only affect new channels/servers or existing ones as well?

yijiem (approved):

markdroth (commented):
I think we want a slightly different approach here.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -80,8 +65,32 @@ absl::flat_hash_set<std::string> BaseMetrics() {
...
+// GlobalStatsPluginRegistry.
+std::atomic<bool> g_otel_plugin_enabled(false);
+
 }  // namespace
 
+// TODO(yashykt): Extend this to allow multiple OpenTelemetry plugins to be
```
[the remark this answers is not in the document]
yashykt: I think we are on the same page about this.

The `g_otel_plugin_state_` is an existing thing that will go away once we have the GlobalStatsPluginRegistry. Currently, the limitation is that we can have only once otel stats plugin enabled, so I think it's fine to just have a single global that works for the otel stats plugin.

If we do this in the CSM plugin's target and server selector callbacks, it will just disable the CSM label addition, but the base OTel functionality will continue which I don't think we want.

include/grpcpp/ext/csm_observability.h
```diff
@@ -38,11 +38,13 @@ class OpenTelemetryPluginBuilderImpl;
 
 namespace experimental {
 
-// This is a no-op at present, but in the future, this object would be useful
-// for performing cleanup.
+// When this object goes out of scope, CsmObservability will stop recording
```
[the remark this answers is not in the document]
yashykt: I like it. Done!

src/cpp/ext/otel/otel_plugin.cc
yashykt: I think the OTelPluginBuilder will probably have new methods - 
1) OTelHandle Build();
2) void Register(OTelHandle);
3) void UnRegister(OTelHandle);

or `BuildAndRegister` returns a handle  which has a method on it to deregister. We discussed that we do not want deregistering to affect existing ones (to avoid synchronization). It'll just affect new ones.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -680,6 +680,23 @@ TEST_F(OpenTelemetryPluginEnd2EndTest,
   EXPECT_EQ(*status_value, "UNIMPLEMENTED");
 }
 
+TEST_F(OpenTelemetryPluginEnd2EndTest, DisableOTelPluginTest) {
+  Init(std::move(Options().set_metric_names(
+      {grpc::OpenTelemetryPluginBuilder::kClientAttemptDurationInstrumentName,
+       grpc::OpenTelemetryPluginBuilder::kServerCallDurationInstrumentName})));
+  grpc::internal::DisableOpenTelemetryPlugin();
```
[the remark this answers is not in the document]
yashykt: Yeah, this is the discussion we had a few months back when you were implementing the service name and namespace labels. At that point, we decided to breakdown the testing in two separate parts - 
1) Tests the xDS plumbing
2) Tests the Observability plumbing

We don't have a test that tests the whole thing locally. (We have CI coverage, but not a local thing.) I'm thinking of adding one soon though. It'll increase my confidence when I submit my PRs, and I won't have to wait for CI runs.

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: Done

src/cpp/ext/otel/otel_plugin.cc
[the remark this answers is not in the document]
yashykt: Done as discussed on chat

test/cpp/ext/csm/csm_observability_test.cc
```diff
@@ -40,7 +42,15 @@ TEST(GsmDependencyTest, GoogleCloudOpenTelemetryDependency) {
   EXPECT_NE(google::cloud::otel::MakeResourceDetector(), nullptr);
 }
 
-TEST(CsmChannelTargetSelectorTest, NonXdsTargets) {
+class CsmChannelTargetSelectorTest : public ::testing::Test {
+ protected:
+  ~CsmChannelTargetSelectorTest() override {
+    grpc_core::CoreConfiguration::Reset();
```
markdroth, line 48: Why is this needed?

markdroth (approved):
Looks great!  Just one question about the test; feel free to merge after answering.

PR #35836, opened 2024-02-07
[CSM] De-experimentalize CsmObservability API

yashykt:


include/grpcpp/ext/csm_observability.h
```diff
@@ -97,7 +95,15 @@ class CsmObservabilityBuilder {
   std::unique_ptr<grpc::internal::OpenTelemetryPluginBuilderImpl> builder_;
 };
 
+namespace experimental {
+// TODO(yashykt): Remove this once no longer needed.
```
yijiem, line 99: Who is using these right now?

yijiem (approved):

include/grpcpp/ext/csm_observability.h
yashykt: just in case someone is using it. They'll get warnings first, and then in coming releases we can remove it

yijiem (approved):

PR #35841, opened 2024-02-07
[http] allow empty te header

quantumsheep:
I stumbled upon errors when `:te` header wasn't set. It should be ok to have an empty `:te` header.

[As the RFC says](https://datatracker.ietf.org/doc/html/rfc9113#section-8.2.2-2):
> The only exception to this is the TE header field, which MAY be present in an HTTP/2 request; when it is, it MUST NOT contain any value other than "trailers".

This change allow empty `:te` header but we could also set the header to `trailers` if this causes issues further down in the code.

### DIFF
--- a/src/core/ext/filters/http/server/http_server_filter.cc
+++ b/src/core/ext/filters/http/server/http_server_filter.cc
@@ -96,10 +96,8 @@ ServerMetadataHandle HttpServerFilter::Call::OnClientInitialMetadata(
   }
 
   auto te = md.Take(TeMetadata());
-  if (te == TeMetadata::kTrailers) {
+  if (te == TeMetadata::kTrailers || !te.has_value()) {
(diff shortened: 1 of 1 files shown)


ejona86:
gRPC requires trailers. Do you know why te: trailers is missing? If it is a client that just isn't sending it, seems it could easily. If it is a proxy stripping it because it doesn't support trailers, it sounds like the RPC will fail anyway.

quantumsheep:
It's due to Envoy now [removing TE headers](https://github.com/envoyproxy/envoy/pull/30535).

Reading https://mailman.nginx.org/pipermail/nginx-devel/2018-March/010946.html, it seems like gRPC SHOULD contain a TE header so you're right. I guess I will have to deal that with Envoy, thanks.

PR #35851, opened 2024-02-07
[slice] remove slice base64 APIs in favor of absl base64 APIs

markdroth:


ctiller:
I think I'd prefer to see an experiment transitioning this first: we don't have terribly great coverage of this code, and there's enough nuance between various base64 flavors that I worry about bug for bug compatibility.

markdroth:
I've added an experiment here, as we discussed.

ctiller (approved):

Issue #23503, opened 2020-07-16
grpc_load_file() is flaky on MacOS when two threads attempt to read the same file


jtattermusch:
If two threads invoke `grpc_load_file` for the same filename concurrently, loading of the file might fail with this error:
```
E0603 07:33:03.955011000 123145516060672 verify_peer_options.cc:57]    load_file: {"created":"@1591169583.954981000","description":"Failed to load file","file":"src/core/lib/iomgr/load_file.cc","file_line":72,"filename":"src/core/tsi/test_creds/ca.pem","referenced_errors":[{"created":"@1591169583.954964000","description":"Is a directory","errno":21,"file":"src/core/lib/iomgr/load_file.cc","file_line":57,"os_error":"Is a directory","syscall":"fread"}]}
E0603 07:33:03.955061000 123145516060672 verify_peer_options.cc:57]    assertion failed: GRPC_LOG_IF_ERROR("load_file", grpc_load_file(SSL_CA_PATH, 1, &ca_slice))
```

Oddly, it reports "errno 21: Is a directory" but the real cause seems to be that even though each thread has called their own fopen (for reading) and they both have a different `FILE*` object, their invocations of `fread` and `fseek` still interfere.
(and it makes one of the reads fail occasionally).
https://github.com/grpc/grpc/blob/9b215e9f90410fbe6679e47b7d5ed6f3ca76bf77/src/core/lib/iomgr/load_file.cc#L33

Even though this conditions is rare, the `grpc_load_file` is a function that's used in the production code (not just in tests) and therefore should be fixed. It is mostly used for reading the files with certificates.

This is also the rootcause for verify_peer_options test being flaky (See b/160697968 for reference)

jtattermusch:
CC @jboeuf  as you've added this function in https://github.com/grpc/grpc/commit/026a417defcd13d0ae5e8a8ddb67c18ff02fa142#diff-e71d6ee1c67f937cfd448359b95ab99dR44 (and later refactored it to its current form).

jtattermusch:
A rough reproduction of this problem is available in https://github.com/grpc/grpc/pull/23502.

jtattermusch:
CC @yihuazhang  as the `grpc_load_file` function is mostly used in the security-related stack.

jtattermusch:
I suggest this two step approach:

[...]

PR #35856, opened 2024-02-08
[Security] add mutex lock to the file loading in FileExternalAccountCredentials

rockspore:
I'm following the suggestions in #23503.

The FileExternalAccountCredentials is the only place I found that could potentially read the same file concurrently (if multiple requests are made simultaneously with the same call credentials).

However, I can't produce any errors without the lock so am not entirely certain whether this is a necessary fix.

### DIFF
--- a/src/core/lib/iomgr/load_file.h
+++ b/src/core/lib/iomgr/load_file.h
@@ -21,14 +21,15 @@
 
 #include <grpc/support/port_platform.h>
 
-#include <stdio.h>
-
 #include <grpc/slice.h>
 
 #include "src/core/lib/iomgr/error.h"
 
 // Loads the content of a file into a slice. add_null_terminator will add
 // a NULL terminator if non-zero.
+// This API is NOT thread-safe and requires proper synchronization when used by
+// multiple threads, especially when they can happen to be reading from the same
+// file.
 grpc_error_handle grpc_load_file(const char* filename, int add_null_terminator,
                                  grpc_slice* output);
 
--- a/src/core/lib/security/credentials/external/file_external_account_credentials.h
+++ b/src/core/lib/security/credentials/external/file_external_account_credentials.h
@@ -25,6 +25,8 @@
 
 #include "absl/strings/string_view.h"
 
+#include <grpc/support/sync.h>
+
 #include "src/core/lib/gprpp/ref_counted_ptr.h"
 #include "src/core/lib/iomgr/error.h"
 #include "src/core/lib/security/credentials/external/external_account_credentials.h"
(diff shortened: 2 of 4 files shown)


gtcooke94 (approved):
This LGTM pending all the tests running

I'd also like to know if it's absolutely required at the point, so let's get Mark's eyes on it as well?

rockspore:
@markdroth I just noticed you opened #35857. I think this shall wait till yours get merged.

src/core/lib/iomgr/load_file.h
```diff
@@ -21,14 +21,15 @@
...
 
-#include <stdio.h>
-
 #include <grpc/slice.h>
 
 #include "src/core/lib/iomgr/error.h"
 
 // Loads the content of a file into a slice. add_null_terminator will add
 // a NULL terminator if non-zero.
+// This API is NOT thread-safe and requires proper synchronization when used by
```
markdroth, line 30: I am soon going to land #35857, which removes this API in favor of `grpc_core::LoadFile()`.  I suggest moving this comment to that function instead.

src/core/lib/security/credentials/external/file_external_account_credentials.h
```diff
@@ -52,6 +54,8 @@ class FileExternalAccountCredentials final : public ExternalAccountCredentials {
   std::string file_;
   std::string format_type_;
   std::string format_subject_token_field_name_;
+  // Used to lock the file loading.
+  gpr_mu mu_;
```
markdroth, line 58: Instead of using `gpr_mu` and `gpr_mu_lock()` / `gpr_mu_unlock()`, please use `grpc_core::Mutex` and `grpc_core::MutexLock`.

test/core/security/credentials_test.cc
```diff
@@ -3038,6 +3038,58 @@ TEST(CredentialsTest, TestFileExternalAccountCredsSuccessFormatJson) {
...
+      "client_secret",                    // client_secret;
+      "",                                 // workforce_pool_user_project;
+  };
+  grpc_error_handle error;
+  auto creds = FileExternalAccountCredentials::Create(options, {}, &error);
+  GPR_ASSERT(creds != nullptr);
+  GPR_ASSERT(error.ok());
+  GPR_ASSERT(creds->min_security_level() == GRPC_PRIVACY_AND_INTEGRITY);
+  for (int i = 0; i < 100; ++i) {
+    auto state = RequestMetadataState::NewInstance(
```
markdroth, line 3080: You're not actually spawning any new threads here, so everything is still happening single-threaded.  This is not actually going to exercise the synchronization you've added.

markdroth (commented):
I added a few comments, but then I realized that this PR is not actually necessary in the first place, because the oauth credential fetching framework ensures that there is only one request outstanding at a time for a given credentials instance:

https://github.com/grpc/grpc/blob/521b2ea17ca1c2dadb83f4f76358f78bf190066f/src/core/lib/security/credentials/oauth2/oauth2_credentials.cc#L324

rockspore:
Thanks for your review, @markdroth! Sorry I didn't catch that `start_fetch` boolean. And thanks for pointing out the flaw of the added test.

Nevertheless, I am closing the PR here.

PR #35861, opened 2024-02-08
[testing] Add StopListening helper and BidiStream handler to XDS e2e test lib

apolcyn:
As title. Pulling these additions out from a larger change.

Related: cl/563857636


markdroth (commented):
The `StopListening()` change looks good, but I'm not sure we actually need the `BidiStream` change.  Maybe split the latter out into a separate PR until we're sure we need it?

apolcyn:
> The StopListening() change looks good, but I'm not sure we actually need the BidiStream change. Maybe split the latter out into a separate PR until we're sure we need it?

I've found ways to restructure my tests to not use bidi streams. Removed that part

PR #35864, opened 2024-02-08
Remove assert that causes a crash

gtcooke94:
It's been noted that this assert causes a crash - really, when shouldn't crash. The function will already return an error that can be handled as needed during the `end` block. The only possible code path for this assert to crash would occur when an error is already being returned, so this should not result in any behavior change except _not_ crashing and returning that error.
b/262285099
b/262240495


gtcooke94:
Closing this PR because of https://github.com/grpc/grpc/pull/35857

PR #35865, opened 2024-02-08
[pick first] explicitly unset selected subchannel when promoting pending subchannel

apolcyn:
This change is needed in order to get certain LB-policy-unit-test-framework-based tests to pass, which involve updates on PF policies.

Without this change, in such tests, health watchers on the original subchannel can be left hanging around. The underlying reason is related to https://github.com/grpc/grpc/blob/842057d8d5a79d48538e4b276f46d56b4dfcdb16/test/core/client_channel/lb_policy/lb_policy_test_lib.h#L203.

Related: cl/563857636


src/core/load_balancing/pick_first/pick_first.cc
```diff
@@ -1022,6 +1022,7 @@ void PickFirst::SubchannelList::SubchannelData::ProcessUnselectedReadyLocked() {
               p, p->latest_pending_subchannel_list_.get(),
               p->subchannel_list_.get());
     }
+    p->UnsetSelectedSubchannel();
```
markdroth, line 1025: Can we add a test that covers this?  I'm thinking that we could get connected, then send a new update that has a READY subchannel, and then check the number of watchers to make sure it's 1 instead of 2.

apolcyn: Added a test.

To trigger the original issue though, we need to enable health checking, so I added a way to enable that in the tests.

The only downside I see of this test is it seems to mostly be testing the policy's interactions with the mocks in the LB policy test lib, but I think the code path is still useful to exercise.

test/core/client_channel/lb_policy/pick_first_test.cc
```diff
@@ -1103,6 +1106,49 @@ TEST_F(PickFirstTest, ShufflingDisabled) {
...
+  // Reapply the same update we did before. The the underlying
+  // subchannel will immediately become ready.
+  status =
+      ApplyUpdate(BuildUpdate({kAddress}, MakePickFirstConfig()), lb_policy());
+  EXPECT_TRUE(status.ok()) << status;
+  picker = ExpectState(GRPC_CHANNEL_READY);
+  EXPECT_EQ(ExpectPickComplete(picker.get()), kAddress);
+  // At this point, NumWatchers() should account for our
+  // subchannel connectivity watcher and our health watcher.
+  EXPECT_EQ(subchannel->NumWatchers(), 2);
```
markdroth, line 1149: If you haven't already done so, please verify that this expectation fails without the fix.

markdroth (approved):

apolcyn:
Yep the `NumWatchers()` check fails like this without the change:

```
Expected equality of these values:
  subchannel->NumWatchers()
    Which is: 3
  2
```

Passes with it.

PR #35868, opened 2024-02-08
[securit] Add status to the Cancel function of the TlsCertificateVerifier

rockspore:
This tries to address the issue mentioned in #34434.

Admittedly, it includes some to-be-wasted work if we eventually eliminate the C-core to C++ plumbing. But given that's not happening soon enough, I think it's worth fixing this now.

### DIFF
--- a/include/grpcpp/security/tls_certificate_verifier.h
+++ b/include/grpcpp/security/tls_certificate_verifier.h
@@ -23,6 +23,8 @@
 #include <utility>
 #include <vector>
 
+#include "absl/status/status.h"
+
 #include <grpc/grpc_security_constants.h>
 #include <grpc/status.h>
 #include <grpc/support/log.h>
@@ -116,7 +118,7 @@ class CertificateVerifier {
   // verification request is pending.
   //
   // request: the verification information associated with this request
-  void Cancel(TlsCustomVerificationCheckRequest* request);
+  void Cancel(TlsCustomVerificationCheckRequest* request, const absl::Status&);
 
   // Gets the core verifier used internally.
   grpc_tls_certificate_verifier* c_verifier() { return verifier_; }
@@ -181,7 +183,8 @@ class ExternalCertificateVerifier {
   // passed to Verify() with a status indicating the cancellation.
   //
   // request: the verification information associated with this request
-  virtual void Cancel(TlsCustomVerificationCheckRequest* request) = 0;
+  virtual void Cancel(TlsCustomVerificationCheckRequest* request,
+                      const absl::Status&) = 0;
 
  protected:
   ExternalCertificateVerifier();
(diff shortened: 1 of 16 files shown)


include/grpcpp/security/tls_certificate_verifier.h
```diff
@@ -116,7 +118,7 @@ class CertificateVerifier {
   // verification request is pending.
   //
   // request: the verification information associated with this request
-  void Cancel(TlsCustomVerificationCheckRequest* request);
+  void Cancel(TlsCustomVerificationCheckRequest* request, const absl::Status&);
```
matthewstevenson88, line 121: nit: Should we add a variable name for the `absl::Status` parameter, for consistency?

Similarly below.

src/core/lib/iomgr/load_file.h
```diff
@@ -21,14 +21,15 @@
...
 
-#include <stdio.h>
-
 #include <grpc/slice.h>
 
 #include "src/core/lib/iomgr/error.h"
 
 // Loads the content of a file into a slice. add_null_terminator will add
 // a NULL terminator if non-zero.
+// This API is NOT thread-safe and requires proper synchronization when used by
```
matthewstevenson88, line 30: It looks like this change (and the changes to file_external_account_credentials.{h,cc} got carried over from https://github.com/grpc/grpc/pull/35856). Should they be reverted in this PR?

src/core/lib/security/credentials/tls/grpc_tls_certificate_verifier.cc
```diff
@@ -210,9 +210,11 @@ int grpc_tls_certificate_verifier_verify(
 
 void grpc_tls_certificate_verifier_cancel(
     grpc_tls_certificate_verifier* verifier,
-    grpc_tls_custom_verification_check_request* request) {
+    grpc_tls_custom_verification_check_request* request, grpc_status_code code,
+    const char* error_details) {
   grpc_core::ExecCtx exec_ctx;
-  verifier->Cancel(request);
+  verifier->Cancel(request, absl::Status(static_cast<absl::StatusCode>(code),
+                                         error_details));
```
matthewstevenson88, line 217: nit: To be safe, we should probably check that `error_details` is not nullptr before passing into the `absl::Status`, e.g. we could write `error_details ? error_details : ""`.

src/cpp/common/tls_certificate_verifier.cc
```diff
@@ -240,7 +246,8 @@ void ExternalCertificateVerifier::CancelInCoreExternalVerifier(
     }
   }
   if (cpp_request != nullptr) {
-    self->Cancel(cpp_request);
+    self->Cancel(cpp_request, absl::Status(static_cast<absl::StatusCode>(code),
+                                           error_details));
```
matthewstevenson88, line 250: nit: Same here RE checking for nullptr.

test/core/security/credentials_test.cc
```diff
@@ -3038,6 +3038,58 @@ TEST(CredentialsTest, TestFileExternalAccountCredsSuccessFormatJson) {
   gpr_free(subject_token_path);
 }
 
+TEST(CredentialsTest,
```
matthewstevenson88, line 3041: Is this change also leftover from https://github.com/grpc/grpc/pull/35856 ?

matthewstevenson88 (commented):
Thanks for the PR @rockspore! Generally looks good, just a few minor comments. :)

PR #35875, opened 2024-02-09
[Security - TLS verifier] add status information to cancel

rockspore:
#35868 got polluted. This reopens it cleanly with all the comments from that PR addressed.

Original description:

 > This tries to address the issue mentioned in https://github.com/grpc/grpc/pull/34434.

 > Admittedly, it includes some to-be-wasted work if we eventually eliminate the C-core to C++ plumbing. But given that's not happening soon enough, I think it's worth fixing this now.

### DIFF
--- a/src/cpp/common/tls_certificate_verifier.cc
+++ b/src/cpp/common/tls_certificate_verifier.cc
@@ -141,10 +141,14 @@ bool CertificateVerifier::Verify(TlsCustomVerificationCheckRequest* request,
   return is_done;
 }
 
-void CertificateVerifier::Cancel(TlsCustomVerificationCheckRequest* request) {
-  GPR_ASSERT(request != nullptr);
-  GPR_ASSERT(request->c_request() != nullptr);
-  grpc_tls_certificate_verifier_cancel(verifier_, request->c_request());
+void CertificateVerifier::Cancel(TlsCustomVerificationCheckRequest* request,
+                                 const absl::Status& status) {
+  // A copy of the status message has to be made since the API takes const
+  // char*.
+  grpc_tls_certificate_verifier_cancel(
+      verifier_, request->c_request(),
+      static_cast<grpc_status_code>(status.code()),
+      std::string(status.message()).c_str());
 }
 
 void CertificateVerifier::AsyncCheckDone(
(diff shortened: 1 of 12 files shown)


src/cpp/common/tls_certificate_verifier.cc
```diff
@@ -141,10 +141,14 @@ bool CertificateVerifier::Verify(TlsCustomVerificationCheckRequest* request,
   return is_done;
 }
 
-void CertificateVerifier::Cancel(TlsCustomVerificationCheckRequest* request) {
-  GPR_ASSERT(request != nullptr);
```
gtcooke94, line 145: Is it fully safe to remove these?

test/core/util/tls_utils.h
```diff
@@ -130,7 +131,8 @@ class AsyncExternalVerifier {
                     void* callback_arg, grpc_status_code* sync_status,
                     char** sync_error_details);
 
-  static void Cancel(void*, grpc_tls_custom_verification_check_request*) {}
+  static void Cancel(void*, grpc_tls_custom_verification_check_request*,
```
gtcooke94, line 134: We just don't care about defining more than a no-op cancels because it's a test helper and we aren't testing cancel in this area?

test/cpp/util/tls_test_utils.cc
```diff
@@ -43,6 +40,7 @@ AsyncCertificateVerifier::AsyncCertificateVerifier(bool success)
     : success_(success),
       thread_("AsyncCertificateVerifierWorkerThread", WorkerThread, this) {
   thread_.Start();
+  thread_started_ = true;
```
gtcooke94, line 43: Do we need this bool? Won't it always be true since it's hard-set during the ctor and not touched elsewhere?

gtcooke94 (commented):
A few comments - also, this represents a breaking change in the public API since we're changing `grpc_security.h`, right?

src/cpp/common/tls_certificate_verifier.cc
rockspore: Oops. I didn't mean to. Added it back.

test/cpp/util/tls_test_utils.cc
rockspore: We have another ctor that doesn't set it (defined in the header since it only takes the status), and this bool defaults to false.

thread_.Join() will crash if it never started. That's why we need this boolean.

test/core/util/tls_utils.h
rockspore: We explicitly call `Cancel` in the test code, unlike in some cases we implement the interface and expect it to be called directly and correctly. So we don't need to do anything here.

rockspore:
> A few comments - also, this represents a breaking change in the public API since we're changing `grpc_security.h`, right?

Yes I think so. But it's experimental API so I suppose it's fine.

matthewstevenson88 (approved):
LGTM, but please also get sign-off from @gtcooke94.

rockspore:
To give some context why I closed it for posterity, it would cause many internal breakages that are tedious to fix. Given the required effort, we decided not to prioritize this.

PR #35890, opened 2024-02-13
[inproc] Fix WaitForConnected for inproc channels

yashykt:


markdroth (changes requested):
I'd prefer not to special-case this right now.  Once the call v3 stack and transport selection changes are in place, we'll be in a much better position to try to solve this in a more general way.  But adding the special case now will make it a bit harder to make progress on those projects.

PR #35892, opened 2024-02-13
[EventEngine] Fix lock acquisition ordering problem on Windows Connect

drfloob:
CC @youyuanwu

This fixes a lock acquisition ordering problem in WindowsEventEngine's Connect behavior, between a ConnectionState's internal mutex, and a HandshakeManager's mutex. The connection state mutex should not be held when calling the on_connected callback.

This should unblock #34801


ctiller (approved):
Would be great to get a test in place here (maybe something that forces lock acquisition during a callback)

drfloob:
> Would be great to get a test in place here (maybe something that forces lock acquisition during a callback)

Agreed. We'll have good coverage when the AF_UNIX PR is landed (see https://github.com/grpc/grpc/pull/34801#issuecomment-1937923505, currently blocked on this PR). Many tests will fail deterministically without this fix.

PR #35906, opened 2024-02-14
[Python Stub] Add version check to stubs generated by grpcio_tools

XuanWang-Amos:
The stubs generated by grpcio_tools should always be used with [the same or higher version of grpcio](https://github.com/grpc/grpc/blob/master/tools/distrib/python/grpcio_tools/setup.py#L313), this change will add a run time check for this requirement inside the generated stubs and therefor enforce this requirement.

Please note for now we're just printing a warning for incorrect usage, we'll **change it to an error** soon.

Example warning message:
```
/usr/local/google/home/xuanwn/workspace/misc/grpc/examples/python/helloworld/helloworld_pb2_grpc.py:21: RuntimeWarning: The grpc package installed is at version 1.60.1, but the generated code in helloworld_pb2_grpc.py depends on grpcio>=1.63.0.dev0. Please upgrade your grpc module to grpcio>=1.63.0.dev0 or downgrade your generated code using grpcio-tools<=1.60.1. This warning will become an error in 1.64.0, scheduled for release on May 14,2024.
```


examples/python/helloworld/helloworld_pb2_grpc.py
```diff
@@ -1,9 +1,38 @@
...
+def _first_version_is_lower(version1, version2):
+    """
+    Compares two versions in the format '1.60.1' or '1.60.1.dev0'.
+    Returns True if version1 is lower, False otherwise.
+    """
+    version1_list = version1.split('.')
+    version2_list = version2.split('.')
+
+    for i in range(3):
+        if version1_list[i] < version2_list[i]:
```
gnossen, line 20: This is doing lexicographic sort, right? We probably want to convert to an integer, right? Otherwise, we'll have some odd results:

```
>>> "3" < "400"
True
```

examples/python/helloworld/helloworld_pb2_grpc.py
```diff
@@ -1,9 +1,38 @@
...
 """Client and server classes corresponding to protobuf-defined services."""
 import grpc
+import warnings
 
 import helloworld_pb2 as helloworld__pb2
 
+GRPC_GENERATE_VERSION='1.62.0.dev0'
+GRPC_VERSION=grpc.__version__
+
+def _first_version_is_lower(version1, version2):
```
gnossen, line 11: Can we make this a protected API within `grpc` itself and just _attempt_ to load it here? If not, this seems like it's going to result in a _lot_ of duplication.

If the import fails, we can assume that the version of grpc is _older_ than `grpcio-tools` and throw the warning.

tools/distrib/python/grpcio_tools/grpc_tools/protoc.py
```diff
@@ -37,6 +38,9 @@ def main(command_arguments):
       command_arguments: a list of strings representing command line arguments to
           `protoc`.
     """
+    command_arguments += [
+        "--grpc_tools_version={}".format(grpc_version.VERSION)
```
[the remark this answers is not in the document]
XuanWang-Amos: Yeah this won't work for simple stub, I'll add version info to `protoc_get_services` too.

examples/python/helloworld/helloworld_pb2_grpc.py
```diff
@@ -5,32 +5,28 @@
...
-    # If we reach here, the versions are equal up to the compared elements.
-    # The version without dev0 will be considered lower.
-    return len(version1_list) < len(version2_list)
-
-if _first_version_is_lower(GRPC_GENERATE_VERSION, GRPC_VERSION):
+GRPC_GENERATED_VERSION = '1.63.0.dev0'
+GRPC_VERSION = grpc.__version__
+EXPECTED_ERROR_RELEASE = '1.64.0'
+SCHEDULED_RELEASE_DATE = 'May 14,2024'
+_version_not_support = False
```
gnossen, line 12: Nit: `_version_not_support` -> `_version_not_supported`. I wouldn't usually be this nitpicky, but this code is going to be duplicated in a _lot_ of places.

src/python/grpcio/grpc/_utilities.py
```diff
@@ -189,3 +189,32 @@ def channel_ready_future(channel: grpc.Channel) -> _ChannelReadyFuture:
     ready_future = _ChannelReadyFuture(channel)
     ready_future.start()
     return ready_future
+
+
+def first_version_is_lower(version1: str, version2: str) -> bool:
```
gnossen, line 194: Please add a note about where this function is used.

src/python/grpcio/grpc/_utilities.py
```diff
@@ -189,3 +189,32 @@ def channel_ready_future(channel: grpc.Channel) -> _ChannelReadyFuture:
...
+    """
```
gnossen, line 195: Please add some simple unit tests for this function. I used the following just to poke at it, but you may want to add some more:

```
print(first_version_is_lower("1.2.3", "1.2.4"))
print(first_version_is_lower("1.2.4", "1.2.3"))
print(first_version_is_lower("1.2.3dev0", "1.2.3"))
print(first_version_is_lower("1.2.3", "1.2.3dev0"))
```

tools/distrib/python/grpcio_tools/grpc_tools/_protoc_compiler.pyx
```diff
@@ -40,13 +43,15 @@ cdef extern from "grpc_tools/main.h" namespace "grpc_tools":
...
                         vector[cProtocWarning]* wrnings) nogil except +
-  int protoc_get_services(char* protobuf_path,
+  int protoc_get_services(char* protobuf_path, char* version,
                           vector[string]* include_path,
                           vector[pair[string, string]]* files_out,
                           vector[cProtocError]* errors,
                           vector[cProtocWarning]* wrnings) nogil except +
 
 def run_main(list args not None):
+  version = "--grpc_tools_version={}".format(grpc_version.VERSION)
```
gnossen, line 53: Instead of using this flag, can you just add a parameter to `protoc_main`? We should be completely in control of the signature.

tools/distrib/python/grpcio_tools/setup.py
```diff
@@ -293,7 +293,7 @@ def extension_modules():
     python_requires=">=3.7",
     install_requires=[
         "protobuf>=4.21.6,<5.0dev",
-        "grpcio>={version}".format(version=grpc_version.VERSION),
```
gnossen, line 296: We definitely still want this line.

tools/distrib/python/grpcio_tools/grpc_tools/protoc.py
```diff
@@ -37,6 +38,9 @@ def main(command_arguments):
       command_arguments: a list of strings representing command line arguments to
           `protoc`.
     """
+    # command_arguments += [
```
gnossen, line 41: Still needed?

src/python/grpcio/grpc/_utilities.py
```diff
@@ -189,3 +189,35 @@ def channel_ready_future(channel: grpc.Channel) -> _ChannelReadyFuture:
...
+        for i in range(3):
+            if int(version1_list[i]) < int(version2_list[i]):
+                return True
+            elif int(version1_list[i]) > int(version2_list[i]):
+                return False
+    except ValueError:
+        # Return false in case we can't convert version to int.
+        return False
+
+    # If we reach here, the versions are equal up to the compared elements.
```
gnossen, line 221: The standard format is `1.2.3dev0`, not `1.2.3.dev0`. So we won't be reaching this line of code if we have `dev0`. I think this is fine because, in general, we should not be comparing versions that have different numbers of period-delimited version segments though, so maybe just remove the comment?

src/python/grpcio_tests/tests/unit/_utilities_test.py
```diff
@@ -0,0 +1,38 @@
...
+
+import logging
+import unittest
+
+from grpc._utilities import first_version_is_lower
+
+
+class UtilityTest(unittest.TestCase):
+    def testVersionCheck(self):
+        self.assertFalse(first_version_is_lower("NOT_A_VERSION", "1.2.4"))
```
gnossen, line 24: I _do_ think some of this behavior is a little weird. For example, in general, exactly one of the three things should be true:

- `a < b`
- `b < a`
- `a == b`

But I _think_ this should be good enough for our purposes assuming that people don't do anything too weird when building manually.

tools/distrib/python/grpcio_tools/grpc_tools/protoc.py
XuanWang-Amos: No we don't need those anymore, deleted.

tools/distrib/python/grpcio_tools/setup.py
XuanWang-Amos: Opps

src/python/grpcio/grpc/_utilities.py
gnossen: Oh, you're right. My mistake.

tools/distrib/python/grpcio_tools/setup.py
gnossen: Still commented out.

gnossen (approved):

PR #35931, opened 2024-02-16
[Security - CrlProvider] Add AKID to CertificateInfo

gtcooke94:
This PR adds the Authority Key Identifier to CertificateInfo. This value _can be_ important in finding the right CRLs to use if there are Issuer name overlaps or a more complicated CA setup with multiple signing keys.

We should observe no behavior change in our `CrlProvider` implementations, this is just adding an important field for users who implement it themselves.


src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -313,4 +313,44 @@ absl::StatusOr<std::string> IssuerFromCert(X509* cert) {
   return ret;
 }
 
+absl::StatusOr<std::string> AkidFromCertificate(X509* cert) {
```
matthewstevenson88, line 316: Please check that `cert` is not null, otherwise we will segfault when we pass it into `X509_get_ext_by_NID`.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -313,4 +313,44 @@ absl::StatusOr<std::string> IssuerFromCert(X509* cert) {
...
+  int len = i2d_ASN1_OCTET_STRING(akid, &buf);
+  if (len <= 0) {
+    return absl::InvalidArgumentError("Could not get AKID from certificate.");
+  }
+  std::string ret(reinterpret_cast<char const*>(buf), len);
+  OPENSSL_free(buf);
+  return ret;
+}
+
+absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl) {
```
matthewstevenson88, line 336: Same here RE nullptr check on `crl`.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -313,4 +313,44 @@ absl::StatusOr<std::string> IssuerFromCert(X509* cert) {
...
+  ASN1_OCTET_STRING* akid = nullptr;
+  int j = X509_CRL_get_ext_by_NID(crl, NID_authority_key_identifier, -1);
+  // Can't have multiple occurrences
+  if (j >= 0) {
+    if (X509_CRL_get_ext_by_NID(crl, NID_authority_key_identifier, j) != -1) {
+      return absl::InvalidArgumentError("Could not get AKID from crlificate.");
```
matthewstevenson88, line 342: nit: s/crlificate/CRL

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -313,4 +313,44 @@ absl::StatusOr<std::string> IssuerFromCert(X509* cert) {
...
+    }
+    akid = X509_EXTENSION_get_data(X509_CRL_get_ext(crl, j));
+  }
+  unsigned char* buf = nullptr;
+  int len = i2d_ASN1_OCTET_STRING(akid, &buf);
+  if (len <= 0) {
+    return absl::InvalidArgumentError("Could not get AKID from crlificate.");
```
matthewstevenson88, line 349: nit: s/crlificate/CRL

matthewstevenson88 (commented):
Thanks for the PR Greg! Just a few minor things. :)

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Done

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Done

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: That's a fun find/replace error :)

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Done

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: And added tests

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -313,4 +313,50 @@ absl::StatusOr<std::string> IssuerFromCert(X509* cert) {
...
+  }
+  ASN1_OCTET_STRING* akid = nullptr;
+  int j = X509_get_ext_by_NID(cert, NID_authority_key_identifier, -1);
+  // Can't have multiple occurrences
+  if (j >= 0) {
+    if (X509_get_ext_by_NID(cert, NID_authority_key_identifier, j) != -1) {
+      return absl::InvalidArgumentError("Could not get AKID from certificate.");
+    }
+    akid = X509_EXTENSION_get_data(X509_get_ext(cert, j));
+  }
```
bob-beck, line 328: And if j is negative because we could not find an AKID in the certificate? 

I think you should check for this here and indicate what you intend to do in that case. 

Right now I think you are falling through calling i2d_ASN1_OCTET_STRING with a nullptr and then 
hitting the failure case of "Could not get AKID from certificate" - which is kind of ugly. 

If you intend to *require* an AKID in the certificate you should probably fail if j is negative here 
with a specific error that the certificate does not contain an AKID. Otherwise if you are going
to tolerate it not being there you need to do something else.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -313,4 +313,50 @@ absl::StatusOr<std::string> IssuerFromCert(X509* cert) {
...
+  ASN1_OCTET_STRING* akid = nullptr;
+  int j = X509_CRL_get_ext_by_NID(crl, NID_authority_key_identifier, -1);
+  // Can't have multiple occurrences
+  if (j >= 0) {
+    if (X509_CRL_get_ext_by_NID(crl, NID_authority_key_identifier, j) != -1) {
+      return absl::InvalidArgumentError("Could not get AKID from crl.");
+    }
+    akid = X509_EXTENSION_get_data(X509_CRL_get_ext(crl, j));
+  }
```
bob-beck, line 351: Same thing here.. check for negative return from X509_CRL_get_ext_by_NID and do what you want to do rather than falling through with a nullptr to the i2d call.

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: Added an `else` block aligned with the `if j >= 0` statement that should catch and return a bad status if j is negative, good catch!

matthewstevenson88 (approved):

PR #35932, opened 2024-02-16
[Python Otel] Allow start observability without context manager

XuanWang-Amos:
Allow start observability globally with a new API `start_open_telemetry_observability`.

### DIFF
--- a/examples/python/observability/observability_greeter_client.py
+++ b/examples/python/observability/observability_greeter_client.py
@@ -54,14 +54,11 @@ def run():
     with grpc.insecure_channel(target="localhost:50051") as channel:
         stub = helloworld_pb2_grpc.GreeterStub(channel)
         try:
-            response = stub.SayHello(
-                helloworld_pb2.HelloRequest(name="You")
-            )
+            response = stub.SayHello(helloworld_pb2.HelloRequest(name="You"))
             print(f"Greeter client received: {response.message}")
         except grpc.RpcError as rpc_error:
             print("Call failed with code: ", rpc_error.code())
     grpc_observability.end_open_telemetry_observability()
-
 
     # Sleep to make sure all metrics are exported.
     time.sleep(5)
--- a/examples/python/observability/observability_greeter_client.py
+++ b/examples/python/observability/observability_greeter_client.py
@@ -47,16 +47,21 @@ def run():
     provider = MeterProvider(metric_readers=[reader])
     otel_plugin = BaseOpenTelemetryPlugin(provider)
 
-    with grpc_observability.OpenTelemetryObservability(plugins=[otel_plugin]):
-        with grpc.insecure_channel(target="localhost:50051") as channel:
-            stub = helloworld_pb2_grpc.GreeterStub(channel)
-            try:
-                response = stub.SayHello(
-                    helloworld_pb2.HelloRequest(name="You")
-                )
-                print(f"Greeter client received: {response.message}")
-            except grpc.RpcError as rpc_error:
-                print("Call failed with code: ", rpc_error.code())
+    grpc_observability.start_open_telemetry_observability(plugins=[otel_plugin])
+    # You can also use context manager to start observability:
+    # with grpc_observability.OpenTelemetryObservability(plugins=[otel_plugin]):
+
+    with grpc.insecure_channel(target="localhost:50051") as channel:
+        stub = helloworld_pb2_grpc.GreeterStub(channel)
+        try:
+            response = stub.SayHello(
+                helloworld_pb2.HelloRequest(name="You")
+            )
+            print(f"Greeter client received: {response.message}")
+        except grpc.RpcError as rpc_error:
+            print("Call failed with code: ", rpc_error.code())
+    grpc_observability.end_open_telemetry_observability()
+
 
     # Sleep to make sure all metrics are exported.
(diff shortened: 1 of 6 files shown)


examples/python/observability/observability_greeter_client.py
```diff
@@ -47,16 +47,18 @@ def run():
...
-            try:
-                response = stub.SayHello(
-                    helloworld_pb2.HelloRequest(name="You")
-                )
-                print(f"Greeter client received: {response.message}")
-            except grpc.RpcError as rpc_error:
-                print("Call failed with code: ", rpc_error.code())
+    grpc_observability.start_open_telemetry_observability(plugins=[otel_plugin])
+    # You can also use context manager to start observability:
+    # with grpc_observability.OpenTelemetryObservability(plugins=[otel_plugin]):
```
gnossen, line 52: As a rule of thumb, people will only ever use the version of an API that you include in your example. So you might want to ask yourself:

- Should the context manager or the global call be the default version?
- Should we include a separate example that includes the other option?

I'll let you make the call on both of those questions.

src/python/grpcio_observability/grpc_observability/__init__.py
```diff
@@ -15,6 +15,17 @@
 from grpc_observability._open_telemetry_observability import (
     OpenTelemetryObservability,
 )
+from grpc_observability._open_telemetry_observability import (
```
gnossen, line 18: Why are these all on separate import statements but each importing a tuple with a single element? Maybe just a single import with a 3-tuple?

src/python/grpcio_observability/grpc_observability/_open_telemetry_observability.py
```diff
@@ -80,26 +112,39 @@ def __init__(
 
         self.exporter = _OpenTelemetryExporterDelegator(_plugins)
 
+    def __enter__(self):
+        global _OPEN_TELEMETRY_OBSERVABILITY  # pylint: disable=global-statement
+        with _observability_lock:
+            if _OPEN_TELEMETRY_OBSERVABILITY:
+                raise RuntimeError(
+                    "gPRC Python observability was already initiated!"
```
gnossen, line 120: Nit: "init" is more commonly an abbreviation for "initialized" rather than "initiated," even though technically both could work here.

src/python/grpcio_observability/grpc_observability/_open_telemetry_observability.py
```diff
@@ -80,26 +112,39 @@ def __init__(
 
         self.exporter = _OpenTelemetryExporterDelegator(_plugins)
 
+    def __enter__(self):
```
gnossen, line 115: It looks like this has a lot in common with the implementation of `start_open_telemetry_observability`. The only real difference seems to be that here, we _know_ which instance of `OpenTelemetryObservability` we want to set `_OPEN_TELEMETRY_OBSERVABILITY` to and in the case of `start_open_telemetry_observability`, we need to create a new one. How about we have a `def _start_open_telemetry_observability(o: OpenTelemetryObservability, plugins: Optional[Iterable[OpenTelemetryPlugin]])` function that both share?

src/python/grpcio_observability/grpc_observability/_open_telemetry_observability.py
```diff
@@ -80,26 +112,39 @@ def __init__(
...
+        with _observability_lock:
+            if _OPEN_TELEMETRY_OBSERVABILITY:
+                raise RuntimeError(
+                    "gPRC Python observability was already initiated!"
+                )
+            self.observability_init()
+            _OPEN_TELEMETRY_OBSERVABILITY = self
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
```
gnossen, line 126: Sam here. Can we share an implementation with `end_open_telemetry_observability`?

src/python/grpcio_tests/tests/observability/_open_telemetry_observability_test.py
```diff
@@ -146,6 +147,68 @@ def testRecordUnaryUnary(self):
...
+        )
+        server, port = _test_server.start_server()
+        self._server = server
+        _test_server.unary_unary_call(port=port)
+
+        self._validate_metrics_exist(self.all_metrics)
+        self._validate_all_metrics_names(self.all_metrics)
+        grpc_observability.end_open_telemetry_observability()
+
+    def testCallGlobalInitThrowErrorWhenGlobalCalled(self):
```
gnossen, line 164: Love the coverage for all of the combinations 👍

examples/python/observability/observability_greeter_client.py
XuanWang-Amos: One of the reasons why we added global init option is to keep in sync with other languages, since none of them have context manager option, I'll keep global call here as default version.

src/python/grpcio_observability/grpc_observability/__init__.py
XuanWang-Amos: It's automatically formatted by `isort`.

gnossen: This isn't about sorting though. I'm asking you to combine these into a single import statement.

gnossen (approved):

Issue #35943, opened 2024-02-20
assertion next_bdp_ping_timer_handle kInvalid


pmoosman:
### What version of gRPC and what language are you using?

1.60.1

### What operating system (Linux, Windows,...) and version?

MacOS 14.2.1

### What runtime / compiler are you using (e.g. python version or version of gcc)

python version

### What did you do?

I'm running Google BigQuery read requests in python using 16 sub-processes on MacOS.

### What did you expect to see?

I expected my read request to succeed.

### What did you see instead?

`ASSERTION FAILED: t->next_bdp_ping_timer_handle != TaskHandle::kInvalid`

Looking at the log files and reading the code, I think I see the problem.

In `chttp2_transport.cc` `finish_bdp_ping_locked()` there is the following code:

```
  t->next_bdp_ping_timer_handle =
      t->event_engine->RunAfter(next_ping - grpc_core::Timestamp::Now(), [t] {
        grpc_core::ApplicationCallbackExecCtx callback_exec_ctx;
        grpc_core::ExecCtx exec_ctx;
        next_bdp_ping_timer_expired(t.get());
      });
```

`next_bdp_ping_timer_expired()` calls `next_bdp_ping_timer_expired_locked()`

which is where the assert is happening:

  `GPR_ASSERT(t->next_bdp_ping_timer_handle != TaskHandle::kInvalid);`

`t->next_bdp_ping_timer_handle `is set with the results of RunAfter().

However, I think `next_bdp_ping_timer_expired_locked()` is getting called before
`t->next_bdp_ping_timer_handle` is set with the return value.

In `chttp2_transport.cc` `finish_bdp_ping_locked()` there is code that calculates the RunAfter() duration. I think this problem only occurs when that duration is 0 ms.

`next_ping - grpc_core::Timestamp::Now() is zero (0)`

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

[crash.log](https://github.com/grpc/grpc/files/14339202/crash.log)

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

pmoosman:
Looking at the code, I think the problem is the zero value for next_ping duration. In `posix_engine.cc` there is function called  `RunAfterInternal()`. If the `when` duration is less than or equal to zero, the call back is run immediately and an invalid handle is returned. (Code which was added by AJ Heller 08-16-2023).

[...]

PR #35953, opened 2024-02-20
[EventEngine] rm invalid assert

drfloob:
In the case where the timer delay is <= 0, RunAfter will return an invalid handle since the callback will not be cancellable. The asserting is invalid in this case. Based on how this handle is used, it seems there is no need for a check that the handle is set. It would only matter if `finish_bdp_ping` were called again before `next_bdp_ping_timer_expired` were called, which I don't think is possible.


drfloob:
Fixes #35943

ctiller (approved):

Hoover391github2 (approved):

drfloob:
> Thanks for this fix. Is there an ETA when this will make it into a release?

That's a good question. I should backport this as a patch fix to a few releases. As it sits today, this should be in the 1.63 release, mid-April.

drfloob:
@mads-gdl Are you experiencing the same crash? And if so, are you using gRPC directly, or are you using the python BigQuery client library?

drfloob:
@mads-gdl grpcio v1.62.1 has been released, this should fix your issue. Please let us know if things are still broken.

PR #35964, opened 2024-02-21
Update min PHP testing version from PHP 7.4 to 8.1

stanley-cheung:
- PHP 7.4 and 8.0 have been EOL https://www.php.net/supported-versions.php
- See go/drop-php-8.0 for more details

Most of the changes came from `generate_projects.sh` after updating the `tools/buildgen/plugins/expand_version.py` file. The change to `tools/internal_ci/helper_scripts/prepare_build_macos_rc` is manual.


stanley-cheung:
"Basic Tests PHP MacOS" failed. Looks like a `phpunit` version mismatch. Looks like there are more to be done here...

stanley-cheung:
Ok "Basic Tests PHP MacOS" now passed. Cautiously optimistic...

drfloob (approved):

PR #35967, opened 2024-02-22
[test] Apply the same workaround as #35573 to remaining cancel_after_invoke test cases

yijiem:

PR #35969, opened 2024-02-22
[test] Decrease ping timeout on CancelAfterInvoke tests

drfloob:
Supersedes #35967

### DIFF
--- a/test/core/end2end/tests/cancel_after_invoke.cc
+++ b/test/core/end2end/tests/cancel_after_invoke.cc
@@ -81,6 +81,8 @@ void CancelAfterInvoke5(CoreEnd2endTest& test,
 void CancelAfterInvoke4(CoreEnd2endTest& test,
                         std::unique_ptr<CancellationMode> mode,
                         Duration timeout) {
+  test.InitClient(ChannelArgs());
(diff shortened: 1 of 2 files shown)


yijiem (approved):

test/core/end2end/tests/cancel_after_invoke.cc
```diff
@@ -81,6 +81,8 @@ void CancelAfterInvoke5(CoreEnd2endTest& test,
 void CancelAfterInvoke4(CoreEnd2endTest& test,
                         std::unique_ptr<CancellationMode> mode,
                         Duration timeout) {
+  test.InitClient(ChannelArgs());
+  test.InitServer(ChannelArgs().Set(GRPC_ARG_PING_TIMEOUT_MS, 5000));
```
yijiem, line 85: I was going to add a comment like: `// This is a workaround for the flakiness that if the server ever enters GracefulShutdown for whatever reason while the client has already been shutdown, the test would not timeout and fail.`

drfloob: I didn't catch this before it merged. Taken care of in https://github.com/grpc/grpc/pull/35971

PR #35971, opened 2024-02-22
[test] Add comment about workaround (re #35969)

drfloob:
Followup from #35969 

This also fixes a few scenarios we had not caught earlier.


yijiem:
Can you also add to the 2 places in https://github.com/grpc/grpc/blob/master/test/core/end2end/tests/cancel_with_status.cc? Thanks!

yijiem (approved):

drfloob:
> Can you also add to the 2 places in https://github.com/grpc/grpc/blob/master/test/core/end2end/tests/cancel_with_status.cc? Thanks!

Done. Good catch.

PR #35977, opened 2024-02-22
[WRR] implement non-per-call metrics

markdroth:

Note that these new metrics are populated only if the `wrr_delegate_to_pick_first` experiment is enabled, which is the case by default.


src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -86,6 +88,41 @@ namespace {
...
 
+constexpr absl::string_view kMetricLabelLocality = "grpc.lb.locality";
+
+const auto kMetricRrFallback =
+    GlobalInstrumentsRegistry::RegisterUInt64Counter(
+        "grpc.lb.wrr.rr_fallback",
+        "EXPERIMENTAL.  Number of scheduler updates in which there were not "
+        "enough endpoints with valid weight, which caused the WRR policy to "
+        "fall back to RR behavior.",
+        "{updates}", {kMetricLabelTarget}, {kMetricLabelLocality}, false);
```
yashykt, line 99: let's follow the same convention here as opentelemetry https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units . s/{updates}/{update}

src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -86,6 +88,41 @@ namespace {
...
+
+const auto kMetricEndpointWeightNotYetUsable =
+    GlobalInstrumentsRegistry::RegisterUInt64Counter(
+        "grpc.lb.wrr.endpoint_weight_not_yet_usable",
+        "EXPERIMENTAL.  Number of endpoints from each scheduler update that "
+        "don't yet have usable weight information (i.e., either the load "
+        "report has not yet been received, or it is within the blackout "
+        "period).",
+        "{endpoints}", {kMetricLabelTarget}, {kMetricLabelLocality}, false);
```
yashykt, line 108: let's follow the same convention here as opentelemetry https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units . s/{endpoints}/{endpoint}

src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -86,6 +88,41 @@ namespace {
...
+
+const auto kMetricEndpointWeightStale =
+    GlobalInstrumentsRegistry::RegisterUInt64Counter(
+        "grpc.lb.wrr.endpoint_weight_stale",
+        "EXPERIMENTAL.  Number of endpoints from each scheduler update whose "
+        "latest weight is older than the expiration period.",
+        "{endpoints}", {kMetricLabelTarget}, {kMetricLabelLocality}, false);
```
yashykt, line 115: let's follow the same convention here as opentelemetry https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units . s/{endpoints}/{endpoint}

yashykt (approved):

test/cpp/end2end/xds/xds_wrr_end2end_test.cc
```diff
@@ -40,12 +42,25 @@ using ::envoy::extensions::load_balancing_policies::
     client_side_weighted_round_robin::v3::ClientSideWeightedRoundRobin;
 using ::envoy::extensions::load_balancing_policies::wrr_locality::v3::
     WrrLocality;
-using WrrTest = XdsEnd2endTest;
+
+class WrrTest : public XdsEnd2endTest {
+ protected:
+  void SetUp() override {
```
yashykt, line 48: what happens if in the future we put more stuff in `XdsEnd2endTest::SetUp`

src/core/lib/channel/metrics.h
```diff
@@ -97,9 +100,14 @@ class GlobalInstrumentsRegistry {
   static void ForEach(
       absl::FunctionRef<void(const GlobalInstrumentDescriptor&)> f);
 
-  static void TestOnlyResetGlobalInstrumentsRegistry();
+ private:
+  friend class GlobalInstrumentsRegistryTestPeer;
```
yashykt, line 104: I see, interesting way to separate out test code

test/cpp/end2end/xds/xds_wrr_end2end_test.cc
markdroth: This is an already-established pattern; we already have a bunch of places where subclasses override `SetUp()` so that individual tests can call `InitClient()` themselves.  We originally moved code from `SetUp()` to `InitClient()` for the specific purpose of supporting cases like this.  If we need to add more code in the future, we would add it in `InitClient()` instead of in `SetUp()`.

src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc
markdroth: Done.

src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc
markdroth: Done.

src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc
markdroth: Done.

src/core/lib/channel/metrics.h
markdroth: We use the test peer pattern in a number of places.  If we're adding just one or two simple test-only accessor methods, it seems fine to do it right in the class, but if it's more complex, a test peer is better, because it (a) avoids punching holes in the public API surface that would be used outside of tests and (b) avoids bloating the production code with test-only functionality.

yijiem (approved):
Thanks Mark!

src/core/load_balancing/weighted_round_robin/weighted_round_robin.cc
```diff
@@ -1418,14 +1460,28 @@ size_t WeightedRoundRobin::Picker::PickIndex() {
...
-    weights.push_back(endpoint.weight->GetWeight(
-        now, config_->weight_expiration_period(), config_->blackout_period()));
-  }
+    float weight = endpoint.weight->GetWeight(
+        now, config_->weight_expiration_period(), config_->blackout_period(),
+        &num_not_yet_usable, &num_stale);
+    weights.push_back(weight);
+    stats_plugins.RecordHistogram(
+        kMetricEndpointWeights, weight,
+        {wrr_->channel_control_helper()->GetTarget()}, {wrr_->locality_name_});
```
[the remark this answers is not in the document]
markdroth: Probably not worth it.  There's not much overhead here.

test/core/client_channel/lb_policy/weighted_round_robin_test.cc
```diff
@@ -989,6 +997,173 @@ TEST_F(WeightedRoundRobinTest, MultipleAddressesPerEndpoint) {
...
+                  // Picker update for second endpoint CONNECTING.
+                  0,
+                  // Picker update for second endpoint READY.
+                  0, 0,
+                  // Picker update for third endpoint CONNECTING.
+                  0, 0,
+                  // Picker update for third endpoint READY.
+                  0, 0, 0,
+                  // Weights for first two endpoints now start getting used.
+                  ::testing::DoubleNear(111.111115, 0.000001),
```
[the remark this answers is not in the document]
markdroth: https://github.com/grpc/proposal/blob/master/A58-client-side-weighted-round-robin-lb-policy.md#subchannel-weights

PR #35980, opened 2024-02-23
[Core] Fix listener linked list walk error

honsunrise:


yijiem:
What is the error? How does this change fix the error? Can you add a test or reproducible example?

PR #35983, opened 2024-02-23
[EventEngine] Set DSCP on listener socket

### LINKED ISSUE #35954: dscp value set through GRPC_ARG_DSCP channel argument does not cause the socket options to be updated on the GRPC server side.
What version of gRPC and what language are you using?
grpc C++ V1.60.0

What operating system (Linux, Windows,...) and version?
AlmaLinux 9

I recently upgraded to v1.60.0 to use the GRPC_ARG_DSCP channel argument that was added as part of https://github.com/grpc/grpc/pull/28322. But I noticed that after adding the following change to server side code when initializing and starting the grpc service,
"""
builder.AddChannelArgument(GRPC_ARG_DSCP, dscp); 
"""
The tos value was not getting set in the stream channel packets sent from the server to the client.

After looking at the grpc c++ core library code, I noticed that the "PosixTcpOptions" arg passed to PrepareSocket ( core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc ) has the custom dscp value set. But there is call to set dscp socket options inside the function. 

"""
absl::Status PrepareSocket(const PosixTcpOptions& options,
                           ListenerSocket& socket) {
  ....
  GRPC_RETURN_IF_ERROR(socket.sock.SetSocketNonBlocking(1));
  GRPC_RETURN_IF_ERROR(socket.sock.SetSocketCloexec(1));

  if (socket.addr.address()->sa_family != AF_UNIX &&
      !ResolvedAddressIsVSock(socket.addr)) {
    GRPC_RETURN_IF_ERROR(socket.sock.SetSocketLowLatency(1));
    GRPC_RETURN_IF_ERROR(socket.sock.SetSocketReuseAddr(1));
    GRPC_RETURN_IF_ERROR(socket.sock.SetSocketDscp(options.dscp));
    socket.sock.TrySetSocketTcpUserTimeout(options, false);
  }
  GRPC_RETURN_IF_ERROR(socket.sock.SetSocketNoSigpipeIfPossible());
  GRPC_RETURN_IF_ERROR(socket.sock.ApplySocketMutatorInOptions(
      GRPC_FD_SERVER_LISTENER_USAGE, options));
"""  

Adding "GRPC_RETURN_IF_ERROR(socket.sock.SetSocketDscp(options.dscp));" fixed the issue and I do see the tos value being set correctly in the outgoing packets from the server side. But I would like to get some confirmation if the fix is the right one and how this got missed in PR# 28322

bjosv:
Prepare the listener socket with the configured DSCP value to mark outgoing packets correctly.
This is already done correctly for clients.

During implementation and delivery of the DSCP feature both experiments `event_engine_listener` and `event_engine_client` were default `OFF`. Unfortunately the `event_engine_listener` experiment was not enabled correctly during verification, which resulted in that the working `iomgr` codepath was used instead.
This PR adds the missing action which I now have verified correctly.

Fixes #35954

### DIFF
--- a/src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
+++ b/src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
@@ -169,6 +169,7 @@ absl::Status PrepareSocket(const PosixTcpOptions& options,
       !ResolvedAddressIsVSock(socket.addr)) {
     GRPC_RETURN_IF_ERROR(socket.sock.SetSocketLowLatency(1));
     GRPC_RETURN_IF_ERROR(socket.sock.SetSocketReuseAddr(1));
+    GRPC_RETURN_IF_ERROR(socket.sock.SetSocketDscp(options.dscp));
     socket.sock.TrySetSocketTcpUserTimeout(options, false);
   }
   GRPC_RETURN_IF_ERROR(socket.sock.SetSocketNoSigpipeIfPossible());


drfloob (approved):
Thanks for the fix!

Vignesh2208 (approved):

PR #35984, opened 2024-02-23
[pick_first] implement non-per-call metrics

markdroth:


src/core/load_balancing/pick_first/pick_first.cc
```diff
@@ -693,6 +716,9 @@ void PickFirst::SubchannelList::SubchannelData::OnConnectivityStateChange(
     }
     // Any state change is considered to be a failure of the existing
     // connection.
+    stats_plugins.AddCounter(
+        kMetricDisconnections, 1,
```
yashykt, line 720: because the list is the current list, we assume that all subchannels were connected and an update over here has to mean that the new state is something other than ready.

In other words `GPR_DEBUG_ASSERT(new_state != GRPC_CHANNEL_READY)` should be fine?

yashykt (approved):

src/core/load_balancing/pick_first/pick_first.cc
markdroth: That assertion should pass, yes, but I don't see enough value in adding it.  The worst that happens here is that we drop a connection that is still working, in which case we'll reconnect soon anyway.  And note that we're already treating this as a disconnection -- the only change in this PR is that we're exporting a metric to indicate that this is happening.

PR #35990, opened 2024-02-23
[EventEngine] Do not accept connections if the listener is shut down

drfloob:
There's a race with the iomgr/EventEngine shims on listener shutdown, specifically for the PosixEventEngine implementation. This showed up as TSAN failures, such as in https://source.cloud.google.com/results/invocations/539def09-0647-4508-960a-f22b76e76dae/targets/%2F%2Ftest%2Fcore%2Fend2end:retry_cancel_after_first_attempt_starts_test@poller%3Depoll1/log. This PR alters the lifetime of the grpc_tcp_server object and its constituent members, letting the EventEngine listener manage it when EE listeners are enabled. Previously, some member objects were destroyed before the tcp server itself was destroyed.

Notes:
 * This is best reviewed once with whitespace diff turned off. Blocks was added to manage mutex lifetimes.
 * BUILD file changes are result of `fix_build_deps.py :iomgr`


src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -205,8 +238,7 @@ static grpc_error_handle CreateEventEngineListener(
           GPR_ASSERT(gpr_atm_no_barrier_load(&s->refs.count) == 0);
           grpc_event_engine::experimental::RunEventEngineClosure(
               shutdown_complete, absl_status_to_grpc_error(status));
-          delete s->fd_handler;
-          delete s;
+          finish_shutdown(s);
```
Vignesh2208, line 241: Is finish_shutdown just equivalent to `delete s->fd_handler; delete s;`  for the EventEngine case ?

src/core/lib/iomgr/tcp_server_posix.cc
```diff
@@ -117,63 +130,76 @@ static grpc_error_handle CreateEventEngineListener(
...
-              gpr_log(GPR_ERROR, "Failed getpeername: %s",
-                      grpc_core::StrError(errno).c_str());
-              close(fd);
+          grpc_pollset* read_notifier_pollset;
+          grpc_tcp_server_acceptor* acceptor;
+          void* cb_arg;
+          // Scoped for server lock, to ensure it's released before the callback
+          // is called.
+          {
+            grpc_core::MutexLockForGprMu lock(&s->mu);
```
Vignesh2208, line 139: Does the whole block need to be under the lock or just the if(s->shutdown) statement ?

src/core/lib/iomgr/tcp_server_posix.cc
drfloob: Not quite. In my PR description, I outlined that one of the key changes here is that none of the constituent objects are destroyed until the server is destroyed. That includes the mutex, the listeners, etc.. Those are now destroyed in `finish_shutdown`.

src/core/lib/iomgr/tcp_server_posix.cc
drfloob: See the lock's usage in `tcp_server_destroy` and `deactivated_all_ports`. If those can run concurrently with any of this accept logic (which includes statements like `s->listen_fd_to_index_map.find(listener_fd)`), then we have a destruction race.

Vignesh2208 (approved):

Issue #35905, opened 2024-02-14
Incorrect XML documentation comment generation for unusual comments


AlphaGremlin:
### What version of gRPC and what language are you using?
C#, NuGet package 2.61.0

### What operating system (Linux, Windows,...) and version?
Windows 10 22H2

### What runtime / compiler are you using (e.g. python version or version of gcc)
.Net 6.0

### What did you do?
Create a .proto file with a message and a field with a comment that inadvertently includes an additional leading slash.

```proto
syntax = "proto3";

message TestMessage {
    /// This has an extra leading slash, and breaks
    string BadField = 1;
    //This has the normal number of slashes, but no space
   string WeirdField = 2;
    // This has the normal number of slashes, and is fine
   string OkField = 3;
}
```

### What did you expect to see?
The generated C# code will have valid XML documentation comments, even with unusual but otherwise valid source comments.

### What did you see instead?
The commented text is placed straight into the raw output, resulting in a comment for `BadField` like:
```cs
/// <summary>
//// This has an extra leading slash, and breaks
/// </summary>
```

C# requires exactly three slashes for an XML documentation comment, so four is interpreted as a normal comment, resulting in some rather confusing warnings until you realise the cause.

```
warning CS1570: XML comment has badly formed XML -- 'Expected an end tag for element 'summary'.'
warning CS1570: XML comment has badly formed XML -- 'End tag was not expected at this location.'
warning CS1587: XML comment is not placed on a valid language element
```

Also, `WierdField` generates a comment like:
```cs
/// <summary>
///This has the normal number of slashes, but no space
/// </summary>
```

This suggests the text is taken directly after the `//` characters, with no whitespace added when generating a multi-line XML comment. The leading whitespace is not technically required, but it can result in this situation where the XML comment format gets broken.

### Anything else we should know about your project / environment?
This is a common typo, especially if your muscle memory is to type three slashes for field comments.

tonydnewell:
@apolcyn I'll take a look

tonydnewell:
I've just confirmed that this behaviour is the same in both the protocol buffers C# plugin and the gRPC C# plugin.

tonydnewell:
Some details - there are two places a simple fix could be made, to check if the line starts with a '`/`' and if so insert a space:
[...]

PR #36000, opened 2024-02-26
Fix csharp doc comments

tonydnewell:
Fix C# document comments when there is an extra / at the start of a comment line.
See https://github.com/grpc/grpc/issues/35905

Related to https://github.com/protocolbuffers/protobuf/pull/15955 which also fixes the issue in protoc compiler.

### DIFF
--- a/src/compiler/csharp_generator.cc
+++ b/src/compiler/csharp_generator.cc
@@ -82,7 +82,13 @@ bool GenerateDocCommentBodyImpl(grpc::protobuf::io::Printer* printer,
         printer->Print("///\n");
       }
       last_was_empty = false;
-      printer->Print("///$line$\n", "line", *it);
+      // If the comment has an extra slash at the start then this can cause the C# compiler
+      // to complain when generating the XML documentation
+      // Issue https://github.com/grpc/grpc/issues/35905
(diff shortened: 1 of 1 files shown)


apolcyn (approved):

apolcyn:
@tonydnewell clang format complains:

```
printer->Print("///\n");
       }
       last_was_empty = false;
-      // If the comment has an extra slash at the start then this can cause the C# compiler
-      // to complain when generating the XML documentation
-      // Issue [https://github.com/grpc/grpc/issues/35905](https://www.google.com/url?q=https://github.com/grpc/grpc/issues/35905&sa=D)
+      // If the comment has an extra slash at the start then this can cause the
+      // C# compiler to complain when generating the XML documentation Issue
+      // [https://github.com/grpc/grpc/issues/35905](https://www.google.com/url?q=https://github.com/grpc/grpc/issues/35905&sa=D)
       if (line[0] == '/') {
-        line.replace(0,1,"&#x2F;");
+        line.replace(0, 1, "&#x2F;");
       }
       printer->Print("///$line$\n", "line", line);
     }
```

apolcyn (approved):

PR #36001, opened 2024-02-26
[RLS] implement non-per-call metrics

markdroth:


src/core/lib/channel/metrics.h
```diff
@@ -97,6 +104,27 @@ class GlobalInstrumentsRegistry {
...
+      absl::string_view name, absl::string_view description,
+      absl::string_view unit, absl::Span<const absl::string_view> label_keys,
+      absl::Span<const absl::string_view> optional_label_keys,
+      bool enable_by_default);
+  static GlobalDoubleGaugeHandle RegisterDoubleGauge(
+      absl::string_view name, absl::string_view description,
+      absl::string_view unit, absl::Span<const absl::string_view> label_keys,
+      absl::Span<const absl::string_view> optional_label_keys,
+      bool enable_by_default);
+  static GlobalCallbackUInt64GaugeHandle RegisterCallbackUInt64Gauge(
```
yijiem, line 117: Note that it seems like OTel's asynchronous Gauge API only support `int64_t` integral type. CC @yashykt

src/core/lib/channel/metrics.h
```diff
@@ -110,6 +138,24 @@ class GlobalInstrumentsRegistry {
   GetInstrumentList();
 };
 
+// An interface for implementing callback-style metrics.
+// To be implemented by stats plugins.
+class CallbackMetricReporter {
```
yijiem, line 143: Why not move this into the `StatsPlugin` interface as a nested class? Since the `CallbackmetricReporter` probably needs to access members within the `StatsPlugin` to access the asynchronous instrument and making it a nested class may make that easier.

src/core/lib/channel/metrics.h
```diff
@@ -147,6 +193,18 @@ class StatsPlugin {
...
+      GlobalInstrumentsRegistry::GlobalUInt64GaugeHandle handle, uint64_t value,
+      absl::Span<const absl::string_view> label_values,
+      absl::Span<const absl::string_view> optional_values) = 0;
+  virtual void SetGauge(
+      GlobalInstrumentsRegistry::GlobalDoubleGaugeHandle handle, double value,
+      absl::Span<const absl::string_view> label_values,
+      absl::Span<const absl::string_view> optional_values) = 0;
+
+  virtual void AddCallback(RegisteredMetricCallback* callback) = 0;
+  virtual void RemoveCallback(RegisteredMetricCallback* callback) = 0;
```
yijiem, line 206: Don't we need to pass a certain type of instrument handle as an argument in this API? So that the StatsPlugin can find the async instrument to add the callback. We might be able to do something like:

```
using AsyncInstrumentHandle = absl::variant<GlobalInstrumentsRegistry::GlobalCallbackUInt64GaugeHandle, GlobalInstrumentsRegistry::GlobalCallbackDoubleGaugeHandle>;

virtual void AddCallback(AsyncInstrumentHandle handle, RegisteredMetricCallback* callback) = 0;
virtual void RemoveCallback(AsyncInstrumentHandle handle, RegisteredMetricCallback* callback) = 0;
```

If we want to minimize the API surface and to accommodate new async types in the future.

src/core/lib/channel/metrics.h
markdroth: Oh, interesting.  I guess that makes sense, since a gauge might want to be negative, unlike a counter.

Any reason we shouldn't make the same change in this API?

src/core/lib/channel/metrics.h
markdroth: A stats plugin that implements this interface can have that implementation nested within the stats plugin subclass.  The base class does not access any data, so there's no advantage of it being inside `StatsPlugin`.

Note that this interface is also needed by the gRPC plugin that generates the metric data, and that plugin does not otherwise need to use `StatsPlugin` directly at all; it only interacts with `StatsPluginGroup`.  So that seems to be a good reason not to put this inside `StatsPlugin`.

src/core/lib/channel/metrics.h
markdroth: The intent here is that a single callback can provide data for multiple metrics, so the callback is not associated with any one metric.

Consider a case like chttp2, where we will probably eventually have tens of metrics, all guarded by the same mutex.  In that case, we don't want to grab the lock separately for each individual metric, which is what we would need to do if we had a separate callback for each metric.  Instead, we want to be able to invoke a single callback that will grab the lock once and then report all of the metrics at once.

src/core/lib/channel/metrics.h
```diff
@@ -208,6 +230,22 @@ class GlobalStatsPluginRegistry {
...
+      }
+    }
+    void SetGauge(GlobalInstrumentsRegistry::GlobalDoubleGaugeHandle handle,
+                  double value,
+                  absl::Span<const absl::string_view> label_values,
+                  absl::Span<const absl::string_view> optional_values) {
+      for (auto& plugin : plugins_) {
+        plugin->SetGauge(handle, value, label_values, optional_values);
+      }
+    }
```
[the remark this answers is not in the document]
markdroth: Yeah, there are a few places where we could avoid duplication like that.  But I'll let you address that in a separate PR.

src/core/lib/channel/metrics.h
markdroth: I've changed the API to allow passing a list of metrics to be updated by each callback, along with a minimum interval at which the callback should be invoked.  These together will make it easier to implement caching in stats plugins like the OT plugin that do not support setting multiple metrics in a single callback.

src/core/lib/channel/metrics.h
markdroth: Changed the gauge APIs to use `int64_t` instead of `uint64_t`.

src/core/lib/channel/metrics.h
```diff
@@ -231,6 +318,32 @@ class GlobalStatsPluginRegistry {
...
+      const {
+    return metrics_;
+  }
+  Duration min_interval() const { return min_interval_; }
+
+ private:
+  GlobalStatsPluginRegistry::StatsPluginGroup& stats_plugin_group_;
+  absl::AnyInvocable<void(CallbackMetricReporter&)> callback_;
+  std::vector<GlobalInstrumentsRegistry::GlobalCallbackHandle> metrics_;
+  Duration min_interval_;
```
yijiem, line 344: Maybe add a comment to this? Like:

```
// Minimum interval for the registered metric callback to be called from the StatsPlugin.
Duration min_interval_;
```

test/core/util/fake_stats_plugin.h
```diff
@@ -355,8 +405,96 @@ class FakeStatsPlugin : public StatsPlugin {
...
+      GlobalInstrumentsRegistry::GlobalDoubleGaugeHandle handle,
+      absl::Span<const absl::string_view> label_values,
+      absl::Span<const absl::string_view> optional_values) {
+    auto iter = double_gauges_.find(handle.index);
+    if (iter == double_gauges_.end()) {
+      return absl::nullopt;
+    }
+    return iter->second.GetValue(label_values, optional_values);
+  }
+  void TriggerCallbacks() {
```
yijiem, line 428: Are we skipping the test for the CallbackGauge for now?

yijiem (approved):
Thanks Mark!

test/core/util/fake_stats_plugin.h
yijiem: Oops, I didn't notice it was used in the rls_end2end_test.cc. Resolved.

src/core/load_balancing/rls/rls.cc
```diff
@@ -690,6 +755,13 @@ class RlsLb : public LoadBalancingPolicy {
   // Updates the picker in the work serializer.
   void UpdatePickerLocked() ABSL_LOCKS_EXCLUDED(&mu_);
 
+  void MaybeExportPickCount(
+      GlobalInstrumentsRegistry::GlobalUInt64CounterHandle handle,
+      absl::string_view target, const PickResult& pick_result);
+
+  const std::string instance_id_;
+  std::unique_ptr<RegisteredMetricCallback> registered_metric_callback_;
```
yashykt, line 763: It looks like the `mu_` will be destroyed before the callback. That's going to result in a crash.

test/core/util/fake_stats_plugin.h
markdroth: No, you're right, we should have tests here showing how the API should work.  The ones in rls_end2end_test.cc are tests for RLS, not tests for the metrics framework.

I've added tests here.

src/core/lib/channel/metrics.h
markdroth: Added a comment on the accessor method, which is what stats plugin implementations actually need to use.

src/core/lib/channel/metrics.h
markdroth: Added comments.

src/core/lib/channel/metrics.h
```diff
@@ -231,6 +327,37 @@ class GlobalStatsPluginRegistry {
       ABSL_GUARDED_BY(mutex_);
 };
 
+// A metric callback that is registered with a stats plugin group.
+class RegisteredMetricCallback {
+ public:
+  RegisteredMetricCallback(
```
yashykt, line 333: do we want each component to have to reason about a reasonable `min_interval`? 

How about another constructor which sets the interval to a default that we think will work for most metrics?

src/core/load_balancing/rls/rls.cc
```diff
@@ -1852,7 +1943,27 @@ RlsLb::ResponseInfo RlsLb::RlsRequest::ParseResponseProto() {
 // RlsLb
 //
 
-RlsLb::RlsLb(Args args) : LoadBalancingPolicy(std::move(args)), cache_(this) {
+std::string GenerateUUID() {
+  absl::uniform_int_distribution<uint64_t> distribution;
+  absl::BitGen bitgen;
```
yashykt, line 1948: https://abseil.io/docs/cpp/guides/random we would normally want to reuse `BitGen` for multiple distributions, but in this case, we will be generating this relatively infrequently, so it's fine to create a new `BitGen` each time. Can we add this as a comment please?

yashykt (approved):

src/core/lib/channel/metrics.h
markdroth: This won't actually be instantiated directly by gRPC components; instead, they will use `StatsPluginGroup::RegisterCallback()` to create this object, and that method does have a default value for this parameter.

src/core/load_balancing/rls/rls.cc
markdroth: In the interest of making sure I get this merged today before I'm out of the office for a while, I'll let you add a comment in a follow-up PR, so that this doesn't need to re-run all of the tests.

PR #36011, opened 2024-02-27
[BUILD] Further simplification of Makefile

veblush:
Simplified `Makefile` which is supposed to be internal only for Ruby and PHP. This is mainly to have a fewer (just necessary) targets instead of all targets found in the Bazel BUILD. As an example, upb used to have a couple of granular targets and this is becoming a problem to upgrade Protobuf to the latest one. Currently, this is an internal use only so this should be fine as long as Ruby and PHP build work.

On top of this, I did a few clean-up and improvement:
- Makefile.template took so long in getting transivie_deps from the target. It can be done quickly by just looking at `transivie_deps`. It now took a few ms.
- `Bunch` in tools/buildgen/_utils.py was too simple to handle write-access to the dictionary which is now required to process `Makefile` so I changed it to handle write operations properly with minimal changes.
- `EMBED_CARES` is removed as this isn't working. Not embedding CARES isn't working.

Internal b/325632907


grpc.gyp
```diff
@@ -62,6 +62,10 @@
       '-Isrc/core/ext/upbdefs-gen',
```
gnossen, line 62: We can probably delete this whole file, right? JS is no longer supported in this repo.

gnossen (approved):

grpc.gyp
veblush: Yep, I think we can. https://github.com/grpc/grpc/pull/36077

veblush:
Copybara did something weird and forgot to close this. This change was merged with this commit; https://github.com/grpc/grpc/commit/7b98cda1f4c4a3836d02b4be4018926f62aaf42b

PR #36020, opened 2024-02-28
[XdsClient] implement non-per-call metrics

markdroth:


yashykt (approved):
lgtm

PR #36021, opened 2024-02-28
[xds] Support for multiple servers in the bootstrap.json

eugeneo:


test/core/xds/xds_client_test.cc
```diff
@@ -2792,6 +2798,47 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
   EXPECT_TRUE(stream->Orphaned());
 }
 
+TEST_F(XdsClientTest, Fallback) {
```
markdroth, line 2801: I don't think this test belongs in this PR, since the actual fallback functionality isn't here yet.

test/cpp/end2end/xds/xds_utils.cc
```diff
@@ -78,9 +78,12 @@ std::string XdsBootstrapBuilder::Build() {
 }
 
 std::string XdsBootstrapBuilder::MakeXdsServersText(
-    absl::string_view server_uri) {
+    absl::Span<const std::string> server_uris) {
+  std::vector<std::string> server_features;
```
markdroth, line 82: Why were these 4 lines moved up to the top of this function?  Seems like it flowed a bit better where they were before.

test/cpp/end2end/xds/xds_utils.h
```diff
@@ -43,7 +43,13 @@ class XdsBootstrapBuilder {
   // not already been set.
   XdsBootstrapBuilder& SetDefaultServer(const std::string& server,
                                         bool ignore_if_set = false) {
-    if (!ignore_if_set || top_server_.empty()) top_server_ = server;
+    if (!ignore_if_set || absl::c_find(servers_, server) == servers_.end()) {
```
markdroth, line 46: The `ignore_if_set` thing was a bit of a hack to allow tests to override the default server.  The idea is that the test framework [automatically sets the default server, but only if it was not already set by the application](https://github.com/grpc/grpc/blob/4ce4f4ea3aa2fd5863b47cfa2eee4de7cdd6306d/test/cpp/end2end/xds/xds_end2end_test_lib.cc#L504).  Given that pattern, I think we would want `ignore_if_set` to ignore the value if the current list is non-empty.  It should not matter whether the specified server is in the list or not, just that the list has already been populated.

Actually, I think it might be cleaner to just eliminate this `ignore_if_set` hack completely.  Instead, I suggest the following:

- In `XdsEnd2endTest`, add this method:
  ```
  XdsBootstrapBuilder MakeBootstrapBuilder() {
    return XdsBootstrapBuilder().SetServer(
        absl::StrCat("localhost:", balancer_->port()));
  }
  ```
- In all tests where we directly instantiate an `XdsBootstrapBuilder` but do not set the server, we should instead call `MakeBootstrapBuilder()`, so that the default server is set automatically.
- In cases where we instantiate an `XdsBootstrapBuilder` and *do* override the default server (it looks like this we do this in [only one place](https://github.com/grpc/grpc/blob/4ce4f4ea3aa2fd5863b47cfa2eee4de7cdd6306d/test/cpp/end2end/xds/xds_core_end2end_test.cc#L1283)), the code should continue to do what it's already doing.
- We can remove [the line in `InitClient()` that sets the default server](https://github.com/grpc/grpc/blob/4ce4f4ea3aa2fd5863b47cfa2eee4de7cdd6306d/test/cpp/end2end/xds/xds_end2end_test_lib.cc#L504).

markdroth (commented):
Thanks for doing this!

test/cpp/end2end/xds/xds_utils.h
eugeneo: I tried to contain the blast radius, hence the use of `absl::optional`. Please take another look.

eugeneo (commented):
Thank you for your comments. Please take another look.

test/core/xds/xds_bootstrap_test.cc
```diff
@@ -55,12 +55,27 @@ namespace grpc_core {
 namespace testing {
 namespace {
 
+MATCHER_P(EqXdsServer, name, "equals XdsServer") {
+  auto* server = static_cast<const GrpcXdsBootstrap::GrpcXdsServer*>(arg);
+  if (!::testing::ExplainMatchResult(::testing::Not(nullptr), server,
```
markdroth, line 60: If channel_creds_config is null, we still want to check the URI.  Suggest writing this as:

```
if (!::testing::ExplainMatchResult(::testing::Ne(nullptr), server, result_listener)) {
  return false;
}
bool ok = true;
ok |= ::testing::ExplainMatchResult(name, server->server_uri(), result_listener);
auto creds_config = server->channel_creds_config();
if (!::testing::ExplainMatchResult(::testing::Ne(nullptr), creds_config, result_listener)) {
  return false;
}
ok |= ::testing::ExplainMatchResult("fake", creds_config->type(), result_listener);
return ok;
```

test/core/xds/xds_bootstrap_test.cc
```diff
@@ -55,12 +55,27 @@ namespace grpc_core {
...
+                                     result_listener) ||
+      !::testing::ExplainMatchResult(::testing::Not(nullptr),
+                                     server->channel_creds_config(),
+                                     result_listener)) {
+    return false;
+  }
+  return ::testing::ExplainMatchResult(name, server->server_uri(),
+                                       result_listener) &&
+         ::testing::ExplainMatchResult(
+             "fake", server->channel_creds_config()->type(), result_listener);
```
markdroth, line 70: Instead of hard-coding the value "fake" here, please pass it in as a parameter to the matcher.

test/core/xds/xds_bootstrap_test.cc
```diff
@@ -203,8 +217,8 @@ TEST(XdsBootstrapTest, ValidWithoutNode) {
   auto bootstrap_or = GrpcXdsBootstrap::Create(json_str);
   ASSERT_TRUE(bootstrap_or.ok()) << bootstrap_or.status();
   auto bootstrap = std::move(*bootstrap_or);
-  auto* server =
-      static_cast<const GrpcXdsBootstrap::GrpcXdsServer*>(bootstrap->server(0));
+  auto* server = static_cast<const GrpcXdsBootstrap::GrpcXdsServer*>(
```
markdroth, line 220: Please use `ElementsAre(EqXdsServer(...))`.

Same thing in all places where we're checking the server.

markdroth (commented):
Comments are mostly cosmetic.  Please let me know if you have any questions.  Thanks!

test/core/xds/xds_bootstrap_test.cc
eugeneo: Done

test/cpp/end2end/xds/xds_end2end_test_lib.cc
```diff
@@ -480,9 +480,11 @@ std::vector<int> XdsEnd2endTest::GetBackendPorts(size_t start_index,
...
 }
 
-void XdsEnd2endTest::InitClient(XdsBootstrapBuilder builder,
-                                std::string lb_expected_authority,
-                                int xds_resource_does_not_exist_timeout_ms) {
+void XdsEnd2endTest::InitClient(
+    const absl::optional<XdsBootstrapBuilder>& maybe_builder,
+    std::string lb_expected_authority,
+    int xds_resource_does_not_exist_timeout_ms) {
+  auto builder = maybe_builder.value_or(MakeBootstrapBuilder());
```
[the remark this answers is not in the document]
eugeneo: Done

test/cpp/end2end/xds/xds_utils.h
```diff
@@ -41,11 +41,8 @@ class XdsBootstrapBuilder {
   }
   // If ignore_if_set is true, sets the default server only if it has
   // not already been set.
-  XdsBootstrapBuilder& SetDefaultServer(const std::string& server,
-                                        bool ignore_if_set = false) {
-    if (!ignore_if_set || absl::c_find(servers_, server) == servers_.end()) {
-      servers_.insert(servers_.begin(), {server});
-    }
+  XdsBootstrapBuilder& SetServer(const std::string& server) {
```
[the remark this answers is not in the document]
eugeneo: Done

BUILD
```diff
@@ -4258,6 +4258,7 @@ grpc_cc_library(
         "absl/strings",
         "absl/strings:str_format",
         "absl/types:optional",
+        "absl/types:span",
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/xds/xds_bootstrap.h
```diff
@@ -20,6 +20,9 @@
 #include <grpc/support/port_platform.h>
 
 #include <string>
+#include <vector>
+
+#include "absl/types/span.h"
```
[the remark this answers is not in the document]
eugeneo: Done

eugeneo (commented):
Thank you for the review. I implemented your suggestions, please take another look.

markdroth (approved):
Looks great!

Issue #36009, opened 2024-02-27
abort() should raise grpc.RpcError


awong-evoiq:
### What version of gRPC and what language are you using?
v1.62.0

The python abort() method raises a generic Exception, but it should raise a grpc.RpcError. https://github.com/grpc/grpc/blob/v1.62.0/src/python/grpcio/grpc/_server.py#L407

Based on the test code, it should be raising grpc.RpcError.
https://github.com/grpc/grpc/blob/v1.62.0/src/python/grpcio_tests/tests/unit/_abort_test.py

Note: this also causes a problem with https://github.com/lchenn/py-grpc-prometheus/blob/master/py_grpc_prometheus/prometheus_server_interceptor.py#L85. Using the abort() or abort_with_status() methods do not get logged as errors.

XuanWang-Amos:
The Exception was raised on server side and in the test we're testing client abort.

In general we think it's a good idea to also raise RpcError here, we did made similar changes before and it was reverted because it caused other issue: https://github.com/grpc/grpc/pull/34515

I'll take a look to see if it's possible to raise RpcError here.

awong-evoiq:
> The Exception was raised on server side and in the test we're testing client abort.
> 
> In general we think it's a good idea to also raise RpcError here, we did made similar changes before and it was reverted because it caused other issue: #34515
> 
> I'll take a look to see if it's possible to raise RpcError here.

Thank you for looking!

sjoerdvandenbos-prodrive:
Hey @XuanWang-Amos is there any news on this? The PRs seem to be cancelled. Can I help?

PR #36022, opened 2024-02-28
[python] server abort raises RpcError

awong-evoiq:
When grpc server code calls context.abort() or context.abort_with_status(), it should raise a better to raise a grpc.RpcError instead of just an Exception. This allows interceptor to distinguish between intentional aborts() and unexpected errors. It matches what happens when a client cancels an rpc call.

https://github.com/grpc/grpc/issues/36009


XuanWang-Amos:
Thanks for working on this! We already have a PR to fix this: https://github.com/grpc/grpc/pull/34525 which was reverted, we'll work on merge that again soon.

rkrdo:
@XuanWang-Amos any ETA on getting the feature merged again?

PR #36031, opened 2024-02-29
[Security - Revocation] Refactor how CRLs are checked internally

gtcooke94:
This PR changes how CRLs are handled purely internally. After discussing with davidben@, there are various problems with the `X509_STORE_set_get_crl` API and we shouldn't use it. This change keeps the behavior and external API the same, but instead of bulk pushing CRL information into OpenSSL, we instead iterate through the built chain and check each certificate for revocation, as well as doing the CRL validation ourselves.


src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+    int ret = CheckCertRevocation(provider, cert, issuer);
+    if (ret != 1) {
+      return ret;
+    }
+  }
   return 1;
 }
 
+// The custom verification function to set in OpenSSL.
+// This calls the standard OpenSSL proceudre (X509_verify_cert), then also
```
matthewstevenson88, line 1126: micro-nit: s/proceudre/procedure

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+static int CustomVerificationFunction(X509_STORE_CTX* ctx, void* arg) {
+  // Use arg maybe
+  int ret = X509_verify_cert(ctx);
+  if (ret <= 0) {
+    // Verification failed. We shouldn't expect to have a verified chain, so
+    // there is no need to attempt to extract the root cert from it or check
+    // anything else.
+    return ret;
+  }
+  ret = RootCertExtractCallback(ctx, arg);
```
matthewstevenson88, line 1138: Should we do this *after* the CRL checks?

Rationale: We should fail as fast as possible.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
-// When using CRL Providers, this function used to override the default
-// `check_crl` function in OpenSSL using `X509_STORE_set_check_crl`.
-// CrlProviders put the onus on the users to provide the CRLs that they want to
-// provide, and because we override default CRL fetching behavior, we can expect
-// some of these verification checks to fails for custom CRL providers as well.
-// Thus, we need a passthrough to indicate to OpenSSL that we've provided a CRL
-// and we are good with it.
-static int CheckCrlPassthrough(X509_STORE_CTX* /*ctx*/, X509_CRL* /*crl*/) {
+// Checks each certificate in the chain for revocation
+static int CheckChainRevocation(
```
matthewstevenson88, line 1098: Please comment on the return semantics of this function (e.g. return 1 is success, return 0 is failure).

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
           ->crl();
 
-  X509_CRL* copy = X509_CRL_dup(crl);
-  *crl_out = copy;
+  return X509_CRL_dup(crl);
+}
+
+// Perform the validation checks in RFC5280 6.3.3 to ensure the given CRL is
+// valid
+static int ValidateCrl(X509* cert, X509* issuer, X509_CRL* crl) {
```
matthewstevenson88, line 1039: Please comment on the return semantics of this function (e.g. return 1 is success, return 0 is failure).

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+    }
+  }
   return 1;
 }
 
+// The custom verification function to set in OpenSSL.
+// This calls the standard OpenSSL proceudre (X509_verify_cert), then also
+// extracts the root certificate in the built chain and does revocation checks
+// when a user has configured CrlProviders.
+static int CustomVerificationFunction(X509_STORE_CTX* ctx, void* arg) {
```
matthewstevenson88, line 1129: Please comment on the return semantics of this function (e.g. return 1 is success, return 0 is failure).

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+  valid = grpc_core::VerifyCrlSignature(crl, issuer);
+  return valid;
+}
+
+// Check if a given certificate is revoked
+// Returns 1 if the certificate is not revoked, 0 if the certificate is revoked
+static int CheckCertRevocation(grpc_core::experimental::CrlProvider* provider,
+                               X509* cert, X509* issuer) {
+  auto crl = GetCrlFromProvider(provider, cert);
+  if (absl::IsNotFound(crl.status())) {
```
matthewstevenson88, line 1067: Why is kNotFound special-cased here? Can you comment on this?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+  X509_REVOKED* rev;
+  if (X509_CRL_get0_by_cert(*crl, &rev, cert)) {
+    // cert is revoked
+    X509_CRL_free(*crl);
+    return 0;
+  }
+  // The certificate is not revoked
+  X509_CRL_free(*crl);
   return 1;
+  // RFC5280j - Not supporting reasons
```
matthewstevenson88, line 1093: Should these last 2 comment lines be removed?

matthewstevenson88 (commented):
Thanks for the PR Greg! Don't we need tests that guard the new functionality introduced in this PR?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Changed to bool and addressed

src/core/tsi/ssl_transport_security.cc
gtcooke94: Cleared it up

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+  return valid;
+}
+
+// Check if a given certificate is revoked
+// Returns 1 if the certificate is not revoked, 0 if the certificate is revoked
+static int CheckCertRevocation(grpc_core::experimental::CrlProvider* provider,
+                               X509* cert, X509* issuer) {
+  auto crl = GetCrlFromProvider(provider, cert);
+  if (absl::IsNotFound(crl.status())) {
+    // TODO(gtcooke94) knob for fail open vs. close
```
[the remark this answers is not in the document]
gtcooke94: Cleared it up as part of the above comment

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,128 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+    return 1;
+  } else if (!crl.ok()) {
+    // This is an unexpected error, return false
+    return 0;
+  }
+  // Validate the crl
+  // RFC5280 6.3.3(a-i)
+  if (!ValidateCrl(cert, issuer, *crl)) {
```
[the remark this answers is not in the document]
gtcooke94: I think our impl has the semantics that an `ok` status return means that the CRL won't be nullptr.
I don't see harm in adding an extra check, but I don't think the prior fn call should be able to return a good status with a nulltpr crl. WDYT?

src/core/tsi/ssl_transport_security.cc
gtcooke94: I don't think so - they are the next steps per the RFC, we just don't need to do anything about it, but I think they are worth calling out in the structure of the code. WDYT?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Done

src/core/tsi/ssl_transport_security.cc
gtcooke94: Fixed

src/core/tsi/ssl_transport_security.cc
gtcooke94: Done

src/core/tsi/ssl_transport_security.cc
gtcooke94: Good point, changed, PTAL

src/core/tsi/ssl_transport_security.cc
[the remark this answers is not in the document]
matthewstevenson88: SGTM.

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: That's fine, but maybe we should move them above the return statement in that case?

Also could we say "Not supported" instead of "Not supporting reasons"?

matthewstevenson88:
> Thanks for the PR Greg! Don't we need tests that guard the new functionality introduced in this PR?

Friendly ping on the testing question. :)

gtcooke94:
Ah missed that, sorry! The existing tests should cover all of this as it's purely internal logic. If implemented correctly no behavior should change

matthewstevenson88:
> Ah missed that, sorry! The existing tests should cover all of this as it's purely internal logic. If implemented correctly no behavior should change

Do we at least want tests that ensure we're correctly checking the signature on the CRL? Isn't that a functional difference between the current state and this PR?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Done

gtcooke94:
Added to the tests, PTAL

matthewstevenson88 (approved):

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,137 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+}
+
+// Check if a given certificate is revoked
+// Returns 1 if the certificate is not revoked, 0 if the certificate is revoked
+static int CheckCertRevocation(grpc_core::experimental::CrlProvider* provider,
+                               X509* cert, X509* issuer) {
+  auto crl = GetCrlFromProvider(provider, cert);
+  // Not finding a CRL is a specific behavior. Per RFC5280, not having a CRL to
+  // check for a given certificate means that we cannot know for certain if the
+  // status is Revoked or Unrevoked and instead if Undetermined. How a user
```
bob-beck, line 1070: "is Undetermined" ?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,137 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+  // handles an Undetermined CRL is up to them. We use absl::IsNotFound as an
+  // analogue for not finding the Crl from the provider, thus the certificate in
+  // question is undetermined.
```
bob-beck, line 1073: Capitalize Undetermined for consistency?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,137 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+  if (absl::IsNotFound(crl.status())) {
+    // TODO(gtcooke94) knob for undetermined being revoked or unrevoked. By
+    // default, unrevoked.
+    return 1;
+  } else if (!crl.ok()) {
+    // This is an unexpected error, return false
```
bob-beck, line 1079: X509_CRL_free(crl);

bob-beck, line 1079: You for sure do not need to X509_CRL_free(*crl) here?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,137 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+    X509_STORE_CTX* ctx, grpc_core::experimental::CrlProvider* provider) {
+#if OPENSSL_VERSION_NUMBER >= 0x10100000
+  STACK_OF(X509)* chain = X509_STORE_CTX_get0_chain(ctx);
+#else
+  STACK_OF(X509)* chain = X509_STORE_CTX_get_chain(ctx);
+#endif
+  if (chain == nullptr) {
+    return 0;
+  }
+  size_t chain_length = sk_X509_num(chain);
```
bob-beck, line 1117: sk_x509_num() returns an int, which can be -1, and you're about to jam it into a size_t which is unsigned.
it can return -1 on failure, which includes if the pointer is NULL. Rather than worry about this at all, I would
simply remove the superfluous check for chain being nullptr above, make chain_length and int, and check
for chain_length <= 0 below.

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,137 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
+  if (chain_length == 0) {
+    return 0;
+  }
+  // Loop to < chain_length - 1 because the last cert is the trust anchor/root
+  // which cannot be revoked
+  for (unsigned long i = 0; i < chain_length - 1; i++) {
```
bob-beck, line 1123: Why is i an unsigned long here?  if chain_length was a size_t it should really be a size_t, but either way
you are then jamming it into sk_X509_value(chain, i) which will expect it to be an int, not an unsigned long, or a size_t. 

If you change chain_length above to an int as I suggest, just make i an int here too and then there is nothing to be concerned about

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1026,28 +1025,137 @@ static int GetCrlFromProvider(X509_STORE_CTX* ctx, X509_CRL** crl_out,
...
 }
 
+// The custom verification function to set in OpenSSL using
+// X509_set_cert_verify_callback. This calls the standard OpenSSL procedure
+// (X509_verify_cert), then also extracts the root certificate in the built
+// chain and does revocation checks when a user has configured CrlProviders.
+// returns 1 on success, indicating a trusted chain to a root of trust was
+// found, 0 if a trusted chain could not be built.
+static int CustomVerificationFunction(X509_STORE_CTX* ctx, void* arg) {
+  int ret = X509_verify_cert(ctx);
```
bob-beck, line 1141: Possibly a complete side issue, not sure how much you trust the people sending you certs here, is there any chance someone would be sending you chains of unusual size to make you drown in a fire swamp of work both here and checking the CRL's on the chain after?  Do you set a depth limit (X509_VERIFY_PARAM_set_depth) by default?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Fixed

src/core/tsi/ssl_transport_security.cc
gtcooke94: Fixed

src/core/tsi/ssl_transport_security.cc
gtcooke94: I believe so - the `absl::StatusOr<Crl>` will only hold a crl in the case that the status is good.
So there won't be any CRL to free in this scope - the called function (`GetCrlFromProvider`) is responsible for making sure there aren't any leaks in the case it returns a bad status.

src/core/tsi/ssl_transport_security.cc
gtcooke94: I think I pulled the `size_t` from seeing how the fn is used in the boringssl codebase (https://github.com/google/boringssl/blob/1a118bbf76ab52698961a1c7ec717ab0080b037d/crypto/x509/x509_cmp.c#L201), and in my IDE my type hinting says it returns a `size_t`. I think I'm not quite clear on the return, could you help me understand? 
[here](https://github.com/google/boringssl/blob/1a118bbf76ab52698961a1c7ec717ab0080b037d/include/openssl/stack.h#L296) it looks like the sk_TYPE_num returns a `size_t`, but [here](https://github.com/openssl/openssl/blob/74fd6823884e27c18ec3fe7bd99b9bc02e6f31f3/include/openssl/stack.h#L29) it appears to return an `int`? I'm not sure if this could be some weirdness in how grpc's bazel dependencies are pulling BoringSSL?

Just want to make sure I get it right

src/core/tsi/ssl_transport_security.cc
gtcooke94: Fixed

src/core/tsi/ssl_transport_security.cc
gtcooke94: Fixed

src/core/tsi/ssl_transport_security.cc
gtcooke94: Fixed

src/core/tsi/ssl_transport_security.cc
gtcooke94: Ack, will fix depending on the result of the other comment.

src/core/tsi/ssl_transport_security.cc
gtcooke94: We do not, but setting a limit is a good point, that makes sense. @matthewstevenson88 any input here?

gtcooke94 (commented):
Thanks for the review! Have a question mainly around the `size_t` vs. `int` for the return of`sk_X509_num`

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Agreed that we should have a depth limit. Maybe we can choose some sensible defaults and provide a knob to allow users to tune these values. If that sounds reasonable, can we plan to do this in a follow-up? And let's a file a bug to keep track of this?

src/core/tsi/ssl_transport_security.cc
bob-beck: They are different in OpenSSL vs BoringSSL,  in BoringSSL they are size_t, in OpenSSL they are int.

src/core/tsi/ssl_transport_security.cc
gtcooke94: After some discussion, set to a max length of 100, which is a ridiculously long chain

src/core/tsi/ssl_transport_security.cc
gtcooke94: Changed `unsigned long` to `size_t` - I believe this should be okay given our discussions?

src/core/tsi/ssl_transport_security.cc
gtcooke94: Discussed offline, we are using `size_t` and then also checking against a max size.

src/core/tsi/ssl_transport_security.cc
bob-beck: ok

src/core/tsi/ssl_transport_security.cc
bob-beck: SGTM

src/core/tsi/ssl_transport_security.cc
bob-beck: ack. lgtm

src/core/tsi/ssl_transport_security.cc
bob-beck: indeed ridiculous but at least constrains it on old openssl's ok

Issue #19133, opened 2019-05-24
Allow use of existing sockets to implement systemd socket activation


bknl:
This is a feature request, may apply to all languages or just to C++.
I am on Linux and use C++, currently gRPC 1.14.2 (but will soon update to the latest version).

I would like to be able to create gRPC services in C++ that can use systemd socket activation, see [http://0pointer.de/blog/projects/socket-activation.html](url)
 
In order for that to work, the server reuses a socket (or sockets) provided to the  process at startup that can be retrieved by a small API (i.e. instead of  _socket(2)  ... bind(2)_ one gets a prebound listening socket from a call to _sd_listen_fds().

It looks like this could be done by adding an appropriate _grpc_socket_factory_ before  creating the server, but it isn't clear to me whether that is the intended path or whether there is another option that I am missing.

yorickvP:
My current solution: https://gist.github.com/yorickvP/8d523a4df2b10c5812fa7789e82b7c1b. I suppose providing an actual python API for the External TCP Listener stuff would help a lot in making this less ugly.

glittershark:
Any progress on this? I'm working on a project that needs this, and might want to take a crack at it myself if there hasn't been any work on it internally

yashykt:
Doesn't look like anyone has come forward to work on this yet.

eriksjolund:
Here are  some more arguments why socket activation support would be good to have:

Rootless Podman has since version 3.4 support for passing activated sockets to the container. Socket activation comes with two extra advantages when running rootless Podman:

* [Native network performance over the socket-activated socket](https://github.com/containers/podman/blob/main/docs/tutorials/socket_activation.md#native-network-performance-over-the-socket-activated-socket) when running rootless Podman. Network traffic is normally passed through the program __slirp4netns__ which comes with a performance penalty.  
* There is also a security advantage when running socket-activated containers with Podman. It's sometimes possible to run the container with __--network=none__. I wrote a blog about it https://www.redhat.com/sysadmin/socket-activation-podman

nipil:
I am interrested into this too.

I tried fiddling around using `unix:///proc/{pid}/fd/3` for example.

It only provided me the following error :

```
# E0224 10:47:04.777537851   40056 chttp2_server.cc:1060]
UNKNOWN:No address added out of total 1 resolved for 'unix:///proc/40056/fd/3'
{created_time:"2024-02-24T10:47:04.777053388+01:00",
[...]

PR #36035, opened 2024-03-02
[examples/cpp/systemd] switch to user units instead of requiring root permissions

nipil:
Drop the requirement for root priviledges to run systemd socket activation test

Additionnal changes :
- Does not autostart (no need during a test)
- Better cleanup logic (cleanup even if the test fails)

Tested with :
- systemd 247 on debian 11

release notes: no

### DIFF
--- a/examples/cpp/systemd_socket_activation/test.sh
+++ b/examples/cpp/systemd_socket_activation/test.sh
@@ -13,21 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Run this script as root
-
-clean() {
-    echo "Cleaning..."
-    systemctl stop sdsockact.socket
-    systemctl stop sdsockact.service
-    systemctl daemon-reload
-    rm /tmp/greeter_server
-    rm /tmp/greeter_client
-    rm /etc/systemd/system/sdsockact.service
-    rm /etc/systemd/system/sdsockact.socket
-}
+# Run this script as normal user
 
 fail() {
-    clean
     echo "FAIL: $@" >&2
     exit 1
 }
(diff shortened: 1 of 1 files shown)


examples/cpp/systemd_socket_activation/test.sh
```diff
@@ -54,17 +49,49 @@ ReusePort=true
 WantedBy=sockets.target
 EOF
 
-systemctl daemon-reload
-systemctl enable sdsockact.socket
-systemctl start sdsockact.socket
+    # reload after adding units
+    systemctl --user daemon-reload
+
+    ## starting at boot is not necessary for testing
```
drfloob, line 55: this can be deleted.

examples/cpp/systemd_socket_activation/test.sh
```diff
@@ -54,17 +49,49 @@ ReusePort=true
...
+
+if [ ${RESULT} -ne 0 ]; then
     fail "Response not received"
 fi
-
-popd
 pass "Response received"
-clean
+
+# teardown is called upon exit
```
drfloob, line 97: teardown_after is called upon exit

drfloob (commented):
Thanks for the improvement! Just a few corrections.

Also, is there any reason you didn't test with the EventEngine? iomgr is going to be disabled/deleted.

drfloob:
> Also, is there any reason you didn't test with the EventEngine? iomgr is going to be disabled/deleted.

I see this was never ported to EventEngine. Unfortunately, unless someone ports the code, systemd socket activation will stop working soon. This is a community-supported feature, we (the gRPC team) won't have bandwidth to help much. Do you have any interest @nipil ?

drfloob:
See https://github.com/grpc/grpc/issues/36085 for necessary pre-work.

examples/cpp/systemd_socket_activation/test.sh
nipil: Ok, i will remove it and update the branch

examples/cpp/systemd_socket_activation/test.sh
nipil: Correct

nipil:
> I see this was never ported to EventEngine. Unfortunately, unless someone ports the code, systemd socket activation will stop working soon. This is a community-supported feature, we (the gRPC team) won't have bandwidth to help much. Do you have any interest @nipil ?

I saw this is a community-supported feature, and if i spent time to update/fix the iomgr part, that is because i want to use it inside one of my projects, so i definitely have interest (as other might, as shown in #19133).

I am not a good C/C++ developer, so that is why i prefered _starting_ with debugging/understanding how it was implemented into iomgr, and get a quick return on investment in my projects, before trying to add the whole thing to event engine (where i would get nothing back if i failed to complete).

nipil:
Your comments have been included.

I tried to sync with master to remove an already merged commit from the pull request, but it seems i should have merged instead of rebased, so i ended up force-pushing my way out of it (i am no skilled with git).

drfloob (commented):
Can this PR be closed? Some of these changes are in https://github.com/grpc/grpc/pull/36200.

PR #36036, opened 2024-03-02
[unix path with systemd activation] fix scheme slashes + absolute path 

nipil:
According to doc/naming.md, when a unix path can be
- `unix:relative`, example `unix:foo` or `unix:../bar`
- `unix:absolute`, example `unix:/tmp/foo`
- `unix://absolute`, example `unix:///tmp/foo`

Each form works _when not using systemd socket activation_,
i.e. it handles the "triple slash path" remaining after absl::ConsumePrefix()

But _with systemd socket activation_, the `unix:///tmp/foo` is not working,
because the remaining `///tmp/foo` provided to `sd_is_socket_unix` does
not seem to be handled correctly, as it expects a "natural path" `/tmp/foo`

So it does not identify it as matching one of the provided file descriptor,
and the test fails because the server side tries to open another socket
on disk, overlapping the one systemd created and the client tried
to use, thus failing with a timeout. This pull request fixes that case.

I included an additional error management for the case when only two
slashes are provided, e.g unix://tmp/foo : as per doc/naming.md, this
should never be a valid case.

Tested with :
- systemd 247 on debian 11

release notes: no

### DIFF
--- a/src/core/lib/address_utils/parse_address.cc
+++ b/src/core/lib/address_utils/parse_address.cc
@@ -103,8 +103,17 @@ grpc_error_handle UnixSockaddrPopulate(absl::string_view path,
         "Path name should not have more than ", maxlen, " characters"));
   }
   un->sun_family = AF_UNIX;
-  path.copy(un->sun_path, path.size());
-  un->sun_path[path.size()] = '\0';
+  int offset_start = 0;
+  if (absl::StartsWith(path, "///")) {
+    std::cout << "offset" << std::endl;
+    offset_start = 2;
+  } else if (absl::StartsWith(path, "//")) {
+    return GRPC_ERROR_CREATE(absl::StrCat(
+        "Absolute path should start with either a single slash caracter, or triple slash characters"));
+  }
+  path.copy(un->sun_path, path.size() - offset_start, offset_start);
+  un->sun_path[path.size() - offset_start] = '\0';
+  std::cout << "un->sun_path " << un->sun_path <<  std::endl;
   resolved_addr->len = static_cast<socklen_t>(sizeof(*un));
   return absl::OkStatus();
 }
--- a/src/core/lib/address_utils/parse_address.cc
+++ b/src/core/lib/address_utils/parse_address.cc
@@ -105,7 +105,6 @@ grpc_error_handle UnixSockaddrPopulate(absl::string_view path,
   un->sun_family = AF_UNIX;
   int offset_start = 0;
   if (absl::StartsWith(path, "///")) {
-    std::cout << "offset" << std::endl;
     offset_start = 2;
   } else if (absl::StartsWith(path, "//")) {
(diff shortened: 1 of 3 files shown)


src/core/lib/address_utils/parse_address.cc
```diff
@@ -103,8 +103,17 @@ grpc_error_handle UnixSockaddrPopulate(absl::string_view path,
         "Path name should not have more than ", maxlen, " characters"));
   }
   un->sun_family = AF_UNIX;
-  path.copy(un->sun_path, path.size());
-  un->sun_path[path.size()] = '\0';
+  int offset_start = 0;
+  if (absl::StartsWith(path, "///")) {
```
drfloob, line 107: This could potentially be cleaner with 
```
if (absl::ConsumePrefix(&path, "//") && !absl::StartsWith(path, "/")) {
  return /* error */;
}
path.copy(un->sun_path, path.size());
...
```

src/core/lib/address_utils/parse_address.cc
```diff
@@ -103,8 +103,17 @@ grpc_error_handle UnixSockaddrPopulate(absl::string_view path,
...
+    std::cout << "offset" << std::endl;
+    offset_start = 2;
+  } else if (absl::StartsWith(path, "//")) {
+    return GRPC_ERROR_CREATE(absl::StrCat(
+        "Absolute path should start with either a single slash caracter, or triple slash characters"));
+  }
+  path.copy(un->sun_path, path.size() - offset_start, offset_start);
+  un->sun_path[path.size() - offset_start] = '\0';
+  std::cout << "un->sun_path " << un->sun_path <<  std::endl;
```
drfloob, line 116: Logging in grpc is done with `gpr_log(GPR_DEBUG, "un->sun_path: %s", un->sun_path);`

drfloob:
See https://github.com/grpc/grpc/issues/36085 for necessary pre-work. This won't be useful unless the feature is ported.

src/core/lib/address_utils/parse_address.cc
nipil: Theses debugging output should not have been left over, i'll purge them

src/core/lib/address_utils/parse_address.cc
nipil: Fixed in [21d840d](https://github.com/grpc/grpc/pull/36036/commits/21d840dabdb9e26a4b49ddab066c3552a7118516)

nipil:

src/core/lib/address_utils/parse_address.cc
drfloob: See https://github.com/abseil/abseil-cpp/blob/0941ce7fd95104ba5f4678c41802a3229db1d5eb/absl/strings/strip.h#L36-L40. The only reject path here requires the string to start with `//`

PR #36037, opened 2024-03-02
[python/systemd] allow compiling systemd support into python builds

nipil:
Add GRPC_PYTHON_BUILD_WITH_SYSTEMD environment variable to setup.py

The variable sets the required compile time define and adds the required link time library to the build process


gnossen:
@nipil Can you please provide some more details on the exact use case that this solves?

gnossen:
Nevermind, [I see now](https://github.com/grpc/grpc/blob/2aa1b6479f900ae9eaafbf0876567920308808cf/src/core/lib/iomgr/systemd_utils.cc#L34).

gnossen (approved):

PR #36039, opened 2024-03-02
[systemd/unix-abstract] fix socket activation for "unix-abstract:" naming

nipil:
systemd socket activation was not working with abstract unix sockets

sd_is_socket_unix() requires different length value depending on
the type of unix socket it is working on, so feed a value in 
accordance with the documentation

other additions : early return on error or empty addresses

Tested with :
- systemd 247 on debian 11


drfloob:
See https://github.com/grpc/grpc/issues/36085 for necessary pre-work. This won't be useful unless the feature is ported.

nipil:

PR #36040, opened 2024-03-03
[bazel/dbg copt] fix warning name for GCC

nipil:
Reproduced with :
- bazel-bootstrap 3.5.1+ds-3 (pulling bazel 6.4.0)
- setting CC to gcc and CXX to g++ otherwise bazel does not detect it
- overall, using debian 11 using default gcc/g++ versions 10.2.1

When using --config=dbg, a warning name produces an error at the beginning of the compilation :

`cc1plus: error: '-Werror=return-stack-address':
no option '-Wreturn-stack-address';
did you mean '-Wreturn-local-addr'?`

Fixed the error following the compiler advice for GCC

PLEASE NOTE : I guess `return-stack-address` might be related to CLANG
and `return-local-addr` is the GCC equivalent. I do not know CLANG,
nor am i good enough with bazel to know how to make this option flexible
(so any help will be appreciated)

### DIFF
--- a/tools/bazel.rc
+++ b/tools/bazel.rc
@@ -36,7 +36,7 @@ build:opt --compilation_mode=opt
 build:opt --copt=-Wframe-larger-than=16384
 
 build:dbg --compilation_mode=dbg
-build:dbg --copt=-Werror=return-stack-address
+build:dbg --copt=-Werror=return-local-addr
 
 # Dynamic link cause issues like: `dyld: malformed mach-o: load commands size (59272) > 32768`
 # https://github.com/bazelbuild/bazel/issues/9190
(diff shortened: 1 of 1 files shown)


yashykt:
let's just remove this statement

yashykt:
We really don't want to mention warning flags in the bazel.rc file. This `bazel.rc` file is mostly for gRPC's own CI testing, and we already have a list of flags enabled through https://github.com/grpc/grpc/blob/master/bazel/copts.bzl. This file just mentions flags for clang but it gives us good CI coverage.

nipil:
Included your feedback and removed the line.

yashykt (approved):

yijiem (approved):

PR #36041, opened 2024-03-03
[systemd/unix] handles relative socket path 

nipil:
systemd sd_is_socket_unix() does not seem to handle relative
socket path as expected

See https://lists.freedesktop.org/archives/systemd-devel/2024-March/050088.html
for a report on systemd mailing list, asking about the expected
behaviour for the path (relative or absolute) in that function

This fix rebuilds an absolute path to provide to systemd functions
so that the function can accurately match the path with the
provided fd and enable socket activation to work.

The commit bd23a8 on this PR is built over PR #36039

Tested with :
- systemd 247 on debian 11

### DIFF
--- a/src/core/lib/iomgr/systemd_utils.cc
+++ b/src/core/lib/iomgr/systemd_utils.cc
@@ -36,8 +36,21 @@ bool set_matching_sd_unix_fd(grpc_tcp_server* s,
                              const grpc_resolved_address* addr,
                              const int fd_start, const int n) {
   absl::StatusOr<std::string> addr_name = grpc_sockaddr_to_string(addr, true);
+  if (!addr_name.ok()) {
+    return false;
+  }
+  std::string &address = addr_name.value();
+  if (address.empty()) {
+    return false;
+  }
   for (int i = fd_start; i < fd_start + n; i++) {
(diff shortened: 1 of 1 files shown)


drfloob:
Please test with the event_engine_listener experiment enabled, iomgr is being deprecated.

drfloob:
> Please test with the event_engine_listener experiment enabled, iomgr is being deprecated.

My mistake. See https://github.com/grpc/grpc/issues/36085 for necessary pre-work. This won't be useful unless the feature is ported.

nipil:

src/core/lib/iomgr/systemd_utils.cc
```diff
@@ -36,8 +36,30 @@ bool set_matching_sd_unix_fd(grpc_tcp_server* s,
                              const grpc_resolved_address* addr,
                              const int fd_start, const int n) {
   absl::StatusOr<std::string> addr_name = grpc_sockaddr_to_string(addr, true);
+  if (!addr_name.ok()) {
+    return false;
+  }
+  std::string &address = addr_name.value();
```
drfloob, line 42: You can just use `addr_name->...` to avoid having two names for the same value. See https://abseil.io/tips/181.

PR #36034, opened 2024-03-01
[examples] fix binary path in systemd socket activation test script

nipil:
Fix failing copy operations due to incorrect path levels

PR #36042, opened 2024-03-03
[examples/systemd activation] multiple fixtures for testing different unix enpoints

nipil:
The test script only handled one socket type/name at a time.
Rewrite the script to handle multiple test fixtures, and add tests easily.

To keep existing behaviour, returns the number of FAILED test as exit value :
so when all tests pass, exits with 0, and if any test fails, exit with non-zero.

The tests provided in this PR all pass, when the following PR regarding systemd socket activation fixes are active :
- #36034 (included)
- #36035 (included)
- #36036
- #36039
- #36041

Tested with :
- systemd 247 on debian 11


drfloob:
This appears to depend on https://github.com/grpc/grpc/pull/36035. I'll review it afterwards. In the meantime, please test with the event_engine_listener experiment as well, iomgr is being deprecated.

drfloob (commented):
Can this PR be closed? Broader changes to these tests are in https://github.com/grpc/grpc/pull/36200.

PR #36045, opened 2024-03-03
[ssl] Use grpc_core::Slice in LoadSystemRoots.

matthewstevenson88:
This is some cleanup that was identified during the review of #34874.


matthewstevenson88:
@markdroth PTAL when you have a few moments. :)

src/core/lib/security/security_connector/load_system_roots.h
```diff
@@ -19,16 +19,14 @@
...
-namespace grpc_core {
+#include "src/core/lib/slice/slice.h"
 
-// TODO(matthewstevenson88): Update LoadSystemRootCerts to use Slice
-// instead of grpc_slice.
+namespace grpc_core {
 
 // Returns a slice containing roots from the OS trust store
-grpc_slice LoadSystemRootCerts();
+Slice LoadSystemRootCerts();
```
markdroth, line 29: While we're at it, how about changing this to return `absl::StatusOr<Slice>`, so that we have a more obvious way of returning an error?

src/core/lib/security/security_connector/load_system_roots_supported.cc
```diff
@@ -62,13 +62,13 @@ const char* kCertFiles[] = {"/etc/ssl/cert.pem"};
 const char* kCertDirectories[] = {""};
 #endif                      // GPR_APPLE
 
-grpc_slice GetSystemRootCerts() {
+Slice GetSystemRootCerts() {
   size_t num_cert_files_ = GPR_ARRAY_SIZE(kCertFiles);
```
markdroth, line 66: Pre-existing nit: This should not have an underscore suffix, since it's not a class data member.

src/core/lib/security/security_connector/load_system_roots_supported.cc
```diff
@@ -135,26 +135,25 @@ grpc_slice CreateRootCertsBundle(const char* certs_directory) {
       }
     }
   }
-  bundle_slice = grpc_slice_new(bundle_string, bytes_read, gpr_free);
-  return bundle_slice;
+  return Slice(grpc_slice_new(bundle_string, bytes_read, gpr_free));
```
markdroth, line 138: This should be:

```
return Slice::FromCopiedBuffer(bundle_string, bytes_read);
```

markdroth (commented):
Thanks for doing this cleanup!

markdroth:
I took care of this in https://github.com/grpc/grpc/pull/41478.

PR #36049, opened 2024-03-04
[xDS] add optional locality label to per-call metrics

markdroth:


src/core/lib/channel/call_tracer.h
```diff
@@ -129,6 +129,7 @@ class ClientCallTracer : public CallTracerAnnotationInterface {
    public:
     enum class OptionalLabelComponent : std::uint8_t {
       kXdsServiceLabels = 0,
+      kXdsLocalityLabels = 0,
```
yijiem, line 132: This probably needs to be:

```
enum class OptionalLabelComponent : std::uint8_t {
  kXdsServiceLabels = 0,
  kXdsLocalityLabels = 1,
  kSize = 2,  // keep last
};
```

src/core/ext/xds/xds_client_stats.h
```diff
@@ -63,7 +63,11 @@ class XdsLocalityName : public RefCounted<XdsLocalityName> {
   XdsLocalityName(std::string region, std::string zone, std::string sub_zone)
       : region_(std::move(region)),
         zone_(std::move(zone)),
-        sub_zone_(std::move(sub_zone)) {}
+        sub_zone_(std::move(sub_zone)),
+        locality_labels_(
+            std::make_shared<std::map<std::string, std::string>>()) {
+    locality_labels_->emplace("grpc.lb.locality", AsHumanReadableString());
```
yashykt, line 69: is this the format that we want the labels to be in?

src/core/ext/xds/xds_client_stats.h
```diff
@@ -63,7 +63,11 @@ class XdsLocalityName : public RefCounted<XdsLocalityName> {
   XdsLocalityName(std::string region, std::string zone, std::string sub_zone)
       : region_(std::move(region)),
         zone_(std::move(zone)),
-        sub_zone_(std::move(sub_zone)) {}
+        sub_zone_(std::move(sub_zone)),
+        locality_labels_(
+            std::make_shared<std::map<std::string, std::string>>()) {
```
yashykt, line 68: I'm guessing that this is being saved as an std::make_shared<std::map<std::string, std::string>> instead of just a `std::string` for the actual value to avoid the per-call map creation?

yashykt (approved):

src/core/ext/xds/xds_client_stats.h
markdroth: In general, I think the optional label keys should be the exact name of the label that we want to appear on the metrics, so that stats plugins (a) don't have to any transformations and (b) can more easily enable the same optional label for both per-call and non-per-call metrics.

We didn't get this quite right for the CSM labels, unfortunately, so we're stuck with a weird edge case there, but we should do the right thing for all subsequent labels.

src/core/ext/xds/xds_client_stats.h
markdroth: Yes, exactly.

yijiem (approved):
Thanks Mark!

test/cpp/end2end/xds/xds_cluster_end2end_test.cc
```diff
@@ -363,6 +363,13 @@ TEST_P(CdsTest, MetricLabels) {
               OptionalLabelComponent::kXdsLocalityLabels,
               ::testing::Pointee(::testing::ElementsAre(::testing::Pair(
                   "grpc.lb.locality", LocalityNameString("locality1")))))));
+  // TODO(yashkt, yijiem): This shutdown shouldn't actually be necessary.
+  // The only reason it's here is to add a delay before
+  // fake_client_call_tracer_factory goes out of scope, since there may
+  // be lingering callbacks in the call stack that are using the
+  // CallAttemptTracer even after we get here, which would then cause a
+  // crash.  Find a cleaner way to fix this.
+  balancer_->Shutdown();
```
yashykt, line 372: Yeah, I think that should be simple. We just need a check that the fake client call tracer is done.

src/core/ext/xds/xds_client_stats.h
yashykt: I forgot about this comment. I meant whether the value should be the thing provided by `AsHumanReadableString`.

markdroth: Is there some reason to use a different value?

yashykt: That has a format of `"{region=\"%s\", zone=\"%s\", sub_zone=\"%s\"}"`

If that's the format we want, shouldn't we add it in https://github.com/grpc/proposal/pull/419/files?

PR #36062, opened 2024-03-06
[ObjC] add boringssl swift package

HannahShiSFB:
gRPC swift package depends on [boringSSL-SwiftPM](https://github.com/firebase/boringSSL-SwiftPM), which hasn't been updated since more than two years ago (https://github.com/firebase/boringSSL-SwiftPM/pull/8).
Updating it is not trivial as the `Package.swift` file is maintained manually, a recent [upgrade to swift 5.5](https://github.com/firebase/boringSSL-SwiftPM/pull/16) took good amount of time only to repeatedly exclude one (sometimes two) non-code files and retry build to get new errors, until the build succeeds.

This PR uses the gRPC build system to generate the swift package file from the same boringssl version in third_party.
Will use the output to update the boringSSL-SwiftPM repo after this PR is merged.

Tested with
```
cp BoringSSL-Package.swift Package.swift
swift build --target openssl_grpc
swift test
```


sampajano (approved):
Wow! Looks non-trivial to do indeed! Thanks so much for working on this, Hannah!

(Just curious, what's the command to run to generate the .swift file again? :))

HannahShiSFB:
It's the same script we use to generate other files: tools/buildgen/generate_projects.sh

PR #36070, opened 2024-03-07
[Metrics] New OpenTelemetry Plugin Implementation of Stats Plugin

yijiem:


markdroth (commented):
This looks like a good start!

One significant theme to my comments here is that we need to avoid doing so much work and using so much memory on a per-call basis.  Anything that we can do at the channel level instead, we should.

src/core/lib/channel/metrics.h
```diff
@@ -159,6 +162,9 @@ class CallbackMetricReporter {
 
 class RegisteredMetricCallback;
 
+class ClientCallTracer;
```
yashykt, line 165: why not just include call_tracer.h?

src/core/lib/channel/metrics.h
```diff
@@ -235,57 +237,37 @@ class StatsPlugin {
 // function begins. This API is thread-safe.
 class GlobalStatsPluginRegistry {
  public:
-  class StatsPluginGroup {
+  class StatsPluginGroup : public std::vector<std::shared_ptr<StatsPlugin>> {
```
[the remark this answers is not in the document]
yashykt: yeah, it's considered bad practice to be deriving from std::vector

src/core/lib/surface/call.cc
```diff
@@ -822,6 +822,18 @@ grpc_error_handle FilterStackCall::Create(grpc_call_create_args* args,
     call->send_initial_metadata_.Set(
         GrpcRegisteredMethod(), reinterpret_cast<void*>(static_cast<uintptr_t>(
                                     args->registered_method)));
+    StatsPlugin::ChannelScope scope(channel->target(),
+                                    /*authority=*/"");
+    GlobalStatsPluginRegistry::GetStatsPluginsForChannel(scope).ForEach(
```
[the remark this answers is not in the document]
yashykt: +1

test/core/end2end/tests/http2_stats.cc
```diff
@@ -211,6 +185,49 @@ class FakeServerCallTracerFactory : public ServerCallTracerFactory {
   }
 };
 
+class FakeStatsPlugin : public grpc_core::StatsPlugin {
```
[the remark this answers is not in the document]
yashykt: +1

src/core/lib/surface/legacy_channel.cc
```diff
@@ -116,7 +116,9 @@ LegacyChannel::LegacyChannel(bool is_client, bool is_promising,
                              std::string target,
                              const ChannelArgs& channel_args,
                              RefCountedPtr<grpc_channel_stack> channel_stack)
-    : Channel(std::move(target), channel_args),
+    : Channel(channel_args.GetOwnedString(GRPC_ARG_SERVER_URI)
```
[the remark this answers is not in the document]
markdroth: I sent #36134 to do this.

yijiem (commented):
Thanks!

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -933,6 +938,368 @@ TEST_F(OpenTelemetryPluginOptionEnd2EndTest,
   EXPECT_EQ(absl::get<std::string>(server_attributes.at("key5")), "value5");
 }
 
+using OpenTelemetryPluginNPCMetricsTest = OpenTelemetryPluginEnd2EndTest;
+
+template <typename HandleType, typename ValueType>
+void VerifyCounter(
```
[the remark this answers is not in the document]
yijiem: Done. PTAL!

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -189,10 +189,30 @@ class OpenTelemetryPluginBuilderImpl {
...
+          filtered_target_(filtered_target) {}
+
+    ActivePluginOptionsView* active_plugin_options_view() const {
+      return active_plugin_options_view_.get();
+    }
+
+    absl::string_view filtered_target() const { return filtered_target_; }
+
+   private:
+    std::unique_ptr<ActivePluginOptionsView> active_plugin_options_view_;
```
markdroth, line 212: Why does this need to be dynamically allocated?  Why not simply inline it?

If the problem here is the declaration order, it seems like we could move the declaration of `ActivePluginOptionsView` up above this, and move the body of its methods into the .cc file.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -189,10 +189,30 @@ class OpenTelemetryPluginBuilderImpl {
   std::shared_ptr<std::set<absl::string_view>> optional_label_keys_;
 };
 
+class ActivePluginOptionsView;
```
markdroth, line 192: Consider moving these types inside of `OpenTelemetryPlugin`.  That would allow us to make more of the types and methods in `OpenTelemetryPlugin` private.

src/cpp/ext/otel/otel_server_call_tracer.h
```diff
@@ -21,28 +21,130 @@
...
+  }
+
+  bool IsSampled() override {
+    // Not implemented
+    return false;
+  }
+
+  // Please refer to `grpc_transport_stream_op_batch_payload` for details on
+  // arguments.
+  void RecordSendInitialMetadata(
```
markdroth, line 63: This method is long enough that it should probably move to the .cc file.

https://google.github.io/styleguide/cppguide.html#Inline_Functions

src/cpp/ext/otel/otel_server_call_tracer.h
```diff
@@ -21,28 +21,130 @@
...
+      return method;
+    }
+    return "other";
+  }
+
+  absl::Time start_time_;
+  absl::Duration elapsed_time_;
+  grpc_core::Slice path_;
+  bool registered_method_;
+  // TODO(yashykt): It's wasteful to do this per call. When we re-haul the stats
```
markdroth, line 142: We should probably take care of this TODO as part of this PR.  I think we can put this in the `ScopeConfig` object.  Note that we can have separate `ScopeConfig` subclasses for client-side and server-side, so that we don't waste memory on fields that we need on only one side.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -43,6 +43,92 @@ namespace {
...
+        std::string,
+        std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>& data,
+    absl::string_view metric_name,
+    absl::Span<const absl::string_view> label_keys,
+    absl::Span<const absl::string_view> label_values,
+    absl::Span<const absl::string_view> optional_label_keys,
+    absl::Span<const absl::string_view> optional_label_values) {
+  const auto& attributes = data[metric_name][0].attributes.GetAttributes();
+  EXPECT_EQ(attributes.size(), label_keys.size() + optional_label_keys.size());
+  for (int i = 0; i < label_keys.size(); i++) {
```
markdroth, line 57: Prefer pre-increment over post-increment.

https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement

Same throughout.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -43,6 +43,92 @@ namespace {
 #define GRPC_ARG_SERVER_SELECTOR_KEY "grpc.testing.server_selector_key"
 #define GRPC_ARG_SERVER_SELECTOR_VALUE "grpc.testing.server_selector_value"
 
+void VerifyAttributes(
```
markdroth, line 46: Suggest writing these functions as gmock matchers.  Also, instead of passing in the map and having them do the map lookup and then the vector iteration, suggest writing them to match against only one `opentelemetry::sdk::metrics::PointDataAttributes` and calling them in the context of `::testing::ElementsAre()`.  The goal here should be to write the expectations directly in the test using something like this:

```
EXPECT_THAT(data, ::testing::ElementsAre(
    ::testing::Pair(
        kMetricName,
        ::testing::ElementsAre(
            ::testing::AllOf(
                AttributesEq(kLabelKeys, kLabelValues,
                             kOptionalLabelKeys, kOptionalLabelValues),
                CounterResultEq(kCounterResult))))));
```

This will yield much more helpful error messages when the test fails.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -253,16 +362,18 @@ class ActivePluginOptionsView {
...
+      : otel_plugin_(otel_plugin) {
+    for (size_t i = 0; i < otel_plugin_->plugin_options().size(); ++i) {
+      const auto& plugin_option = otel_plugin_->plugin_options()[i];
       if (plugin_option != nullptr && func(*plugin_option)) {
         active_mask_.set(i);
       }
     }
   }
 
+  OpenTelemetryPlugin* otel_plugin_;
```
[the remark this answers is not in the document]
markdroth: Storing the pointer in `KeyValueIterable` is fine, because that object is only created on the stack, never dynamically allocated.

markdroth (commented):
This is definitely headed in the right direction!  Looks like there are still a number of comments from my original review not yet addressed, but I assume you just hadn't gotten to those yet.

src/cpp/ext/otel/otel_plugin.h
yijiem: Done.

src/cpp/ext/otel/otel_plugin.h
yijiem: Done.

yijiem (commented):
Thanks Mark!

test/core/end2end/tests/http2_stats.cc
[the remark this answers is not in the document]
yijiem: This test has some custom logic between its custom FakeClientCallTracer, FakeServerCallTracer and the test body. I was able to reuse most of the FakeStatsPlugin by inheriting and overriding the `GetClientCallTracer()` and `GetServerCallTracer()` functions.

src/cpp/ext/otel/otel_server_call_tracer.h
markdroth: I think I may have steered you wrong here.  When I read this TODO, I thought this was something that we were initializing per-call even though it didn't depend on per-call data, which is why I suggested initializing it per-server-connection instead.  But looking at the code, it looks like the value actually depends on the call's initial metadata.  So we're now storing a single copy for the connection but setting the data on a per-call basis, which isn't thread-safe and will cause conflicts between calls.

@yashykt should weigh in here about what this TODO was supposed to be about (maybe it was about moving to a metadata lookup service?), but unless he has a better idea, I think this needs to go back to being per-call.

markdroth (commented):
This is getting closer!

Note that there are still some unresolved comments from earlier review passes.

test/core/end2end/tests/http2_stats.cc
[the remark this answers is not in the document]
yijiem: There are synchronization logic between the custom client call tracer, the server call tracer and the test body. This change doesn't make it any worse though. Maybe we can figure out some way to clean it up later? CC @yashykt also.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,33 +186,173 @@ class OpenTelemetryPluginBuilderImpl {
...
+  }
+  Client& client() { return client_; }
+  Server& server() { return server_; }
+
+ private:
+  // Instruments for per-call metrics.
+  Client client_;
+  Server server_;
+  // Instruments for non-per-call metrics.
+  absl::flat_hash_map<
```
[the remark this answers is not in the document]
yijiem: Done. PTAL!

src/core/lib/channel/metrics.h
```diff
@@ -177,14 +178,22 @@ class StatsPlugin {
...
+  class ScopeConfig {
+   public:
+    virtual ~ScopeConfig() = default;
   };
 
   virtual ~StatsPlugin() = default;
 
-  virtual bool IsEnabledForChannel(const ChannelScope& scope) const = 0;
-  virtual bool IsEnabledForServer(const ChannelArgs& args) const = 0;
+  virtual std::pair<bool, std::shared_ptr<ScopeConfig>> IsEnabledForChannel(
```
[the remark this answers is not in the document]
yijiem: Done.

src/core/lib/channel/metrics.h
```diff
@@ -237,38 +247,39 @@ class StatsPlugin {
...
-  class StatsPluginGroup : public std::vector<std::shared_ptr<StatsPlugin>> {
+  class StatsPluginGroup {
    public:
-    StatsPluginGroup() : std::vector<std::shared_ptr<StatsPlugin>>() {}
-
-    void ForEach(absl::FunctionRef<void(std::shared_ptr<StatsPlugin>)> f) {
-      for (auto& plugin : *this) {
-        f(plugin);
-      }
+    void AddStatsPlugin(std::shared_ptr<StatsPlugin> plugin,
```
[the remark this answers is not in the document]
yijiem: Done.

markdroth (commented):
Mostly minor comments here.  The only significant remaining issue is my comment from a previous pass about the injected labels on the server side being set per-call but stored per-connection.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -47,33 +48,40 @@ namespace {
...
 }
 
 MATCHER_P(CounterResultEq, result, "") {
-  return opentelemetry::nostd::holds_alternative<
-             opentelemetry::sdk::metrics::SumPointData>(arg.point_data) &&
-         opentelemetry::nostd::get<std::remove_cv_t<decltype(result)>>(
-             opentelemetry::nostd::get<
-                 opentelemetry::sdk::metrics::SumPointData>(arg.point_data)
-                 .value_) == result;
+  return ::testing::ExplainMatchResult(
```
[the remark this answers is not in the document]
yijiem: This doesn't work and I really don't know how to debug with the compiler error messages:

```
MATCHER_P(CounterResultEq, result, "") {
  return ::testing::ExplainMatchResult(
      ::testing::VariantWith<opentelemetry::sdk::metrics::SumPointData>(
          ::testing::Field(
              &opentelemetry::sdk::metrics::SumPointData::value_,
              ::testing::VariantWith<std::remove_cv_t<decltype(result)>>(
                  ::testing::Eq(result)))),
      arg, result_listener);
}
```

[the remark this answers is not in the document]
yijiem: oh wait, it works now, my bad lol:

```
MATCHER_P(CounterResultEq, result, "") {
  return ::testing::ExplainMatchResult(
      ::testing::VariantWith<opentelemetry::sdk::metrics::SumPointData>(
          ::testing::Field(
              &opentelemetry::sdk::metrics::SumPointData::value_,
              ::testing::VariantWith<std::remove_cv_t<decltype(result)>>(
                  ::testing::Eq(result)))),
      arg.point_data, result_listener);
}
```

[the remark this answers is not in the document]
yijiem: This is amazing.

src/cpp/ext/otel/otel_server_call_tracer.h
yashykt: I think I added the TODO at the wrong line. This is about the `active_plugin_options_view_` which only needs to be calculated once per server. This can be removed

yijiem: Done. Reverted.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,237 @@ class OpenTelemetryPluginBuilderImpl {
...
+      return injected_labels_from_plugin_options_;
+    }
+
+   private:
+    ActivePluginOptionsView active_plugin_options_view_;
+    std::vector<std::unique_ptr<LabelsIterable>>
+        injected_labels_from_plugin_options_;
+  };
+
+  struct Client {
```
yashykt, line 306: now that this struct is outside of `OpenTelemetryPluginState`, let's call this `ClientMetrics` and `ServerMetrics` respectively

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,237 @@ class OpenTelemetryPluginBuilderImpl {
...
+  }
+  const absl::AnyInvocable<bool(absl::string_view /*generic_method*/) const>&
+  generic_method_attribute_filter() const {
+    return generic_method_attribute_filter_;
   }
+  const std::vector<std::unique_ptr<InternalOpenTelemetryPluginOption>>&
+  plugin_options() const {
+    return plugin_options_;
+  }
+  Client& client() { return client_; }
```
yashykt, line 383: Would it make sense to make the OTel call tracer classes subclasses of this?

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,237 @@ class OpenTelemetryPluginBuilderImpl {
...
+  Server& server() { return server_; }
 
-  std::bitset<64> active_mask_;
+ private:
+  // Instruments for per-call metrics.
+  Client client_;
+  Server server_;
+  // Instruments for non-per-call metrics.
+  struct Disabled {};
+  using InstrumentType = absl::variant<
```
yashykt, line 391: `InstrumentType` made me think that this is going to be an enum, but this is the actual OTel Instrument. Let's just call it `Instrument`?

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -425,9 +415,7 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
   using OptionalLabelKeys = std::vector<absl::string_view>;
   // Stores label keys and optional label keys for each instrument at
   // construction time.
-  absl::flat_hash_map<grpc_core::GlobalInstrumentsRegistry::UID,
-                      std::pair<LabelKeys, OptionalLabelKeys>>
-      label_keys_map_;
+  std::vector<std::pair<LabelKeys, OptionalLabelKeys>> label_keys_store_;
```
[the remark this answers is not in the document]
yashykt: Why do we need to store `label_keys` and `optional_label_keys`? Why can't we just get it from the instruments registry?

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,233 @@ class OpenTelemetryPluginBuilderImpl {
...
+  using InstrumentType = absl::variant<
+      Disabled, std::unique_ptr<opentelemetry::metrics::Counter<uint64_t>>,
+      std::unique_ptr<opentelemetry::metrics::Counter<double>>,
+      std::unique_ptr<opentelemetry::metrics::Histogram<uint64_t>>,
+      std::unique_ptr<opentelemetry::metrics::Histogram<double>>>;
+  struct Instrument {
+    InstrumentType otel_instrument;
+    // Stores label keys and optional label keys for each instrument at
+    // construction time.
+    std::vector<absl::string_view> label_keys;
```
yashykt, line 400: Why do we need to store these keys again? Why can't we just get it from the instrument registry? Unless there are synchronization costs, we should be fine.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -79,7 +73,71 @@ absl::flat_hash_set<std::string> BaseMetrics() {
...
+        return false;
+      }
+    }
+    if (enabled_optional_label_keys_ == nullptr) {
+      return true;
+    }
+    // Note that if there is duplicated enabled keys we will send them multiple
+    // times.
+    for (int i = 0; i < optional_label_keys_.size(); i++) {
+      if (enabled_optional_label_keys_->find(optional_label_keys_[i]) ==
```
yashykt, line 117: We don't want to do string comparisons. As suggested in https://github.com/grpc/proposal/pull/421/files, let's use a bit_set to determine whether an optional label in a particular position needs to be recorded or not

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,233 @@ class OpenTelemetryPluginBuilderImpl {
...
+  absl::AnyInvocable<bool(absl::string_view /*target*/) const> target_selector_;
+  absl::AnyInvocable<bool(const grpc_core::ChannelArgs& /*args*/) const>
+      server_selector_;
+  absl::AnyInvocable<bool(absl::string_view /*target*/) const>
+      target_attribute_filter_;
+  absl::AnyInvocable<bool(absl::string_view /*generic_method*/) const>
+      generic_method_attribute_filter_;
+  std::vector<std::unique_ptr<InternalOpenTelemetryPluginOption>>
+      plugin_options_;
+  std::shared_ptr<std::set<absl::string_view>> optional_label_keys_;
```
yashykt, line 415: Let's not save the optional label keys. Instead, let's save a bit set along with each created instrument to denote whether an optional label in a particular position needs to be recorded or not

include/grpcpp/ext/otel_plugin.h
```diff
@@ -106,6 +106,9 @@ class OpenTelemetryPluginBuilder {
   /// options can be added.
   OpenTelemetryPluginBuilder& AddPluginOption(
       std::unique_ptr<OpenTelemetryPluginOption> option);
+  // Records \a optional_label_key on all metrics that provide it.
+  OpenTelemetryPluginBuilder& AddOptionalLabel(
```
yashykt, line 110: I'm wondering if this should be guarded by an `experimental_type`

```
class experimental_type {
public:
    explicit experimental_type(OpenTelemetryPluginBuilder* builder) : builder_(builder) {}
// Records \a optional_label_key on all metrics that provide it.
  OpenTelemetryPluginBuilder& AddOptionalLabel(absl::string_view optional_label_key);
};

/// NOTE: The function experimental() is not stable public API. It is a view
  /// to the experimental components of this class. It may be changed or removed
  /// at any time.
  experimental_type experimental() { return experimental_type(this); }
```

include/grpcpp/ext/otel_plugin.h
```diff
@@ -106,6 +106,9 @@ class OpenTelemetryPluginBuilder {
...
+      absl::string_view optional_label_key);
```
yashykt, line 111: This does not have any of the other changes mentioned in https://github.com/grpc/proposal/pull/421/files. Are they gonna come in a separate PR?

include/grpcpp/ext/otel_plugin.h
yashykt: What do you think @markdroth ?

include/grpcpp/ext/otel_plugin.h
yijiem: Yeah, I think we will pair program that in a separate PR next :).

src/cpp/ext/otel/otel_server_call_tracer.cc
```diff
@@ -50,152 +49,54 @@
...
   path_ =
       recv_initial_metadata->get_pointer(grpc_core::HttpPathMetadata())->Ref();
-  active_plugin_options_view_.ForEach(
+  scope_config_->active_plugin_options_view().ForEach(
       [&](const InternalOpenTelemetryPluginOption& plugin_option,
           size_t index) {
         auto* labels_injector = plugin_option.labels_injector();
         if (labels_injector != nullptr) {
-          injected_labels_from_plugin_options_[index] =
+          scope_config_->injected_labels_from_plugin_options()[index] =
```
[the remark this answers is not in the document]
yijiem: Done. Reverted.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,237 @@ class OpenTelemetryPluginBuilderImpl {
...
+              ActivePluginOptionsView::MakeForServer(args, otel_plugin)),
+          injected_labels_from_plugin_options_(
+              otel_plugin->plugin_options().size()) {}
+
+    const ActivePluginOptionsView& active_plugin_options_view() const {
+      return active_plugin_options_view_;
+    }
+
+    std::vector<std::unique_ptr<LabelsIterable>>&
+    injected_labels_from_plugin_options() {
```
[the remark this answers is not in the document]
yijiem: Reverted.

src/cpp/ext/otel/otel_plugin.h
yijiem: Done.

src/cpp/ext/otel/otel_plugin.h
yashykt: Ah, they already are. Nice! Then, you don't need these accessors.

yijiem: Done. Removed.

src/cpp/ext/otel/otel_plugin.h
yijiem: Done.

src/cpp/ext/otel/otel_plugin.h
yijiem: Done.

src/cpp/ext/otel/otel_plugin.h
yijiem: `std::bitset<N>` is a fixed-size sequence of N bits (N needs to be specified during compile-time), are we certain that the optional label keys size for an instrument will be less than a reasonable bound (say 64)? Or we actually want to be dynamic and store something like a `std::vector<std::bitset<64>>` with each instrument?

src/cpp/ext/otel/otel_call_tracer.h
```diff
@@ -47,12 +47,14 @@
 namespace grpc {
 namespace internal {
 
-class OpenTelemetryCallTracer : public grpc_core::ClientCallTracer {
+class OpenTelemetryPlugin::ClientCallTracer
+    : public grpc_core::ClientCallTracer {
  public:
   class OpenTelemetryCallAttemptTracer : public CallAttemptTracer {
```
markdroth, line 53: This could be renamed to just `AttemptTracer`.  The other qualifications are already clear from the enclosing scope, so there isn't any real value in repeating them here.

src/cpp/ext/otel/otel_plugin.cc
markdroth: Yeah, we definitely want this.

Also, we need the same plumbing for the per-call metrics.  The only difference in that case is that the set of optional labels is dictated by both `AddOptionalLabels()` and by the plugin options.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -69,71 +69,45 @@ MATCHER_P4(AttributesEq, label_keys, label_values, optional_label_keys,
...
-             result_listener) &&
-         ::testing::ExplainMatchResult(
-             result,
-             opentelemetry::nostd::get<std::remove_cv_t<decltype(result)>>(
-                 opentelemetry::nostd::get<
-                     opentelemetry::sdk::metrics::SumPointData>(arg.point_data)
-                     .value_),
-             result_listener);
+template <typename T>
+auto Equal(T result) {
```
markdroth, line 73: Suggest calling this `IntOrDoubleEq()`.

src/cpp/ext/otel/otel_plugin.h
markdroth: I'm fine with saying that there can be no more than 64 optional labels for a given metric.

We do need to store a separate bitset for each instrument, because the set and order of optional labels will be different for each one.  I think we want something like this:

```
struct InstrumentData {
  Instrument instrument;
  std::bitset<64> optional_labels_bits;
};
std::vector<InstrumentData> instruments_;
```

include/grpcpp/ext/otel_plugin.h
markdroth: I don't think it needs to be experimental.  It seems like this will be a pretty basic use-case that we will want to support.

markdroth (commented):
This is looking really good!  I think the only substantive remaining issue is the optional label handling.

src/cpp/ext/otel/otel_plugin.h
yashykt: agreed

src/core/lib/channel/metrics.cc
```diff
@@ -340,7 +378,7 @@ GlobalStatsPluginRegistry::GetAllStatsPlugins() {
   MutexLock lock(&*mutex_);
   StatsPluginGroup group;
   for (const auto& plugin : *plugins_) {
-    group.push_back(plugin);
+    group.AddStatsPlugin(plugin, nullptr);
```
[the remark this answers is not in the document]
yashykt: I don't get this. Why would we want to create a call tracer?

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -38,6 +43,73 @@ namespace grpc {
...
+  return ::testing::ExplainMatchResult(
+      ::testing::VariantWith<opentelemetry::sdk::metrics::SumPointData>(
+          ::testing::Field(
+              &opentelemetry::sdk::metrics::SumPointData::value_,
+              ::testing::VariantWith<std::remove_cv_t<decltype(result)>>(
+                  IntOrDoubleEq(result)))),
+      arg.point_data, result_listener);
+}
+
+MATCHER_P4(HistogramResultEq, sum, min, max, count, "") {
```
yashykt, line 91: I'm not a fan of complicated matchers, but since this is already written, I'm fine with it remaining. My only concern is making debugging easier. My experience with such matchers is that on errors, it is really hard to figure out what went wrong, so at the very minimum, we need to print out what the actual result was so we can compare it to the expectations.

src/core/lib/channel/metrics.cc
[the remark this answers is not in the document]
markdroth: When we start an xDS-enabled gRPC server, we will use this method to get the stats plugins to use for the channel that we'll create to talk to the xDS server.  When we start a call on that xDS channel, we'll call each plugin's `GetClientCallTracer()` method to create a call tracer.  That method will expect to see a non-null `ScopedConfig`.  If we pass it null, it's probably going to crash.

In thinking about this, I realized that in addition to that problem, we actually do want to be a bit more selective than simply selecting *all* stats plugins.  I think what we actually want here is to select only those stats plugins that are not enabling themselves for only a subset of channels.  Given that, I suggest something like this:

- Add a method `std::pair<bool, std::shared_ptr<ScopedConfig>> StatsPlugin::IsEnabledForChannelsByDefault()`.
- Change the code here to call that method, and if it returns true, save the resulting config in the `StatsPluginGroup`.
- In the OT plugin, have `IsEnabledForChannelsByDefault()` return true if `target_selector_` is null.  It can return the same config that would be returned by `IsEnabledForChannel()`.
- Rename this method from `GetAllStatsPlugins()` to `GetStatsPluginsEnabledForChannelsByDefault()`.

markdroth (commented):
This looks great!  The only remaining blocking comment is the one about `GetAllStatsPlugins()`.

src/core/lib/channel/metrics.cc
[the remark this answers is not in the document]
markdroth: Sorry, I got that suggestion slightly wrong -- here's what I meant:

* Add a method `std::pair<bool, std::shared_ptr<ScopedConfig>> StatsPlugin::IsEnabledForServersByDefault()`.
* Change the code here to call that method, and if it returns true, save the resulting config in the `StatsPluginGroup`.
* In the OT plugin, have `IsEnabledForServersByDefault()` return true if `server_selector_` is null.  It can return the same config that would be returned by `IsEnabledForServer()`.
* Rename this method from `GetAllStatsPlugins()` to `GetStatsPluginsEnabledForServersByDefault()`.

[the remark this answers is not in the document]
yijiem: The ScopeConfig returns here might be an issue since it takes a `ChannelArg` which we don't have it here and seems like passing a default `ChannelArg` might not be the most sensible choice? https://github.com/grpc/grpc/blob/604a8b9155aacb98d9206f2d493bac0cf7f2aa74/src/cpp/ext/csm/csm_observability.cc#L52-L55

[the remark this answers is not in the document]
markdroth: New suggestion: Remove `GetAllStatsPlugins()` entirely and change the XdsClient code to instead call `GetStatsPluginsForServer()` with empty channel args.

src/core/lib/channel/metrics.cc
```diff
@@ -42,13 +39,12 @@ GlobalInstrumentsRegistry::RegisterUInt64Counter(
...
     bool enable_by_default) {
   auto& instruments = GetInstrumentList();
-  std::for_each(instruments.begin(), instruments.end(),
-                [&](const GlobalInstrumentDescriptor& descriptor) {
-                  if (descriptor.name == name) {
-                    Crash(absl::StrFormat(
-                        "Metric name %s has already been registered.", name));
-                  }
-                });
+  for (const auto& descriptor : instruments) {
```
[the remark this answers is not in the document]
yijiem: Done.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,224 @@ class OpenTelemetryPluginBuilderImpl {
...
-  static ActivePluginOptionsView MakeForClient(absl::string_view target) {
-    return ActivePluginOptionsView(
-        [target](const InternalOpenTelemetryPluginOption& plugin_option) {
-          return plugin_option.IsActiveOnClientChannel(target);
-        });
-  }
+  // Creates a convenience wrapper to help iterate over only those plugin
+  // options
+  // that are active over a given channel/server.
+  class ActivePluginOptionsView {
```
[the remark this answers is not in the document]
yijiem: Done.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -391,7 +390,11 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
       std::unique_ptr<opentelemetry::metrics::Counter<double>>,
       std::unique_ptr<opentelemetry::metrics::Histogram<uint64_t>>,
       std::unique_ptr<opentelemetry::metrics::Histogram<double>>>;
-  std::vector<Instrument> instruments_;
+  struct InstrumentData {
+    Instrument instrument;
+    std::bitset<64> optional_labels_bits;
```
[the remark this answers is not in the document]
yijiem: Done.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -108,14 +108,8 @@ class NPCMetricsKeyValueIterable
         return false;
       }
     }
-    if (enabled_optional_label_keys_ == nullptr) {
-      return true;
-    }
-    // Note that if there is duplicated enabled keys we will send them multiple
-    // times.
     for (int i = 0; i < optional_label_keys_.size(); i++) {
```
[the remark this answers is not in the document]
yijiem: Done.

src/core/lib/channel/metrics.cc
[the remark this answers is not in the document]
markdroth: Sorry, I think I steered you wrong here -- seeing this made me realize that the limitation of 64 is being imposed in the OT plugin, not here in the instrument registry, so we should enforce the limit in the OT plugin instead.

markdroth (commented):
Just one small comment remaining from my end!

src/core/lib/channel/metrics.cc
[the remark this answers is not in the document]
yijiem: Removed. `std::bitset<N>::set()` function will throw en exception if it is larger than what it can hold so I think we already have that there.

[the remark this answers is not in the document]
markdroth: Depending on an exception from `std::bitset<N>::set()` doesn't seem quite as good, because someone could define a metric with 65 optional labels, and it wouldn't trigger an exception until someone enables the 65th optional label.

markdroth (approved):
I'll go ahead and approve at this point.  Even if we need to add the missing check in a subsequent PR, we can do that.

Nice work on this!

src/core/lib/channel/metrics.cc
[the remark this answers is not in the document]
yijiem: Good point. Done. Though I still like the idea to have `static_assert()` in metrics.cc to check this limit during compile-time when we have better compile-time enforcement. But I also understand that this limit is imposed by the OT plugin so it shouldn't affect the instrument registry.

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: I see, let me test it out

src/core/lib/channel/metrics.cc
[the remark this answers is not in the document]
yashykt: I don't think this is correct. When we create an xDS channel and create calls on that channel, that xDS gRPC channel will just be treated as another gRPC channel. Instead of using the stats plugin group from the xDS component, it will create its own stats plugin group similar to all other client channels.

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -217,53 +186,237 @@ class OpenTelemetryPluginBuilderImpl {
...
+        if (plugin_option != nullptr && func(*plugin_option)) {
+          active_mask_.set(i);
+        }
       }
     }
+
+    std::bitset<64> active_mask_;
+  };
+
+  class ClientScopeConfig : public grpc_core::StatsPlugin::ScopeConfig {
```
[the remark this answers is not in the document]
yashykt: Let's keep it as is

src/core/lib/channel/metrics.cc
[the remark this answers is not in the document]
yashykt: > The ScopeConfig returns here might be an issue since it takes a ChannelArg which we don't have it here and seems like passing a default ChannelArg may cause some issue for CSM on server-side?

CSM is a plugin option. It doesn't affect whether the otel plugin is enabled for a channel/server.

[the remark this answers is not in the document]
yashykt: I'm not going to block over this because I think we haven't properly figured out our story for scoping on the server side yet.

We might need a way for an OTel plugin to say whether it is interested in servers or not, and with that I think `GetStatsPluginsForServer()` will work nicely.

yashykt (approved):

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -103,13 +103,13 @@ class OpenTelemetryPlugin::NPCMetricsKeyValueIterable
                        bool(opentelemetry::nostd::string_view,
                             opentelemetry::common::AttributeValue)>
                            callback) const noexcept override {
-    for (int i = 0; i < label_keys_.size(); i++) {
+    for (size_t i = 0; i < label_keys_.size(); i++) {
```
markdroth, line 106: Prefer preincrement over postincrement.

PR #35796, opened 2024-02-02
[Deps] Upgraded protobuf to v26.1

veblush:
This is to upgrade Protobuf to v26.1 along with upb, and utf8_range.

PR #36074, opened 2024-03-07
[Build] Making grpcio_tools ready for upcoming Protobuf change.

veblush:
Protobuf is beginning to rely more on upb, pulling in more C files. Hence, grpcio_tool needs the following changes to absorb this.

- Changing the setup.py to support both C and C++ by removing explicit language=c++. Rather it can choose the right compiler by its extension.
- Adding build_extensions injection to deal with C/C++ option conflict as grpcio does.
- Adding protobuf and upb directories to the build script so that it can find newly added source files in protobuf.

Tested by https://github.com/grpc/grpc/pull/35796

### DIFF
--- a/tools/distrib/python/grpcio_tools/setup.py
+++ b/tools/distrib/python/grpcio_tools/setup.py
@@ -259,7 +259,6 @@ def extension_modules():
             os.path.join("grpc_root", "include"),
         ]
         + CC_INCLUDES,
-        language="c++",
         define_macros=list(DEFINE_MACROS),
         extra_compile_args=list(EXTRA_COMPILE_ARGS),
         extra_link_args=list(EXTRA_LINK_ARGS),
--- a/tools/distrib/python/grpcio_tools/setup.py
+++ b/tools/distrib/python/grpcio_tools/setup.py
@@ -126,6 +126,40 @@ class BuildExt(build_ext.build_ext):
             filename = filename[: -len(orig_ext_suffix)] + new_ext_suffix
         return filename
 
+    def build_extensions(self):
+        # This special conditioning is here due to difference of compiler
+        #   behavior in gcc and clang. The clang doesn't take --stdc++11
+        #   flags but gcc does. Since the setuptools of Python only support
+        #   all C or all C++ compilation, the mix of C and C++ will crash.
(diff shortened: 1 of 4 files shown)


gnossen:
@XuanWang-Amos Can you please review this?

tools/distrib/python/grpcio_tools/setup.py
```diff
@@ -126,6 +126,40 @@ def get_ext_filename(self, ext_name):
...
+        #
+        #   If we are not using a permissive compiler that's OK with being
+        #   passed wrong std flags, swap out compile function by adding a filter
+        #   for it.
+        old_compile = self.compiler._compile
+
+        def new_compile(obj, src, ext, cc_args, extra_postargs, pp_opts):
+            if src.endswith(".c"):
+                extra_postargs = [
+                    arg for arg in extra_postargs if not "-std=c++" in arg
```
XuanWang-Amos, line 144: [Minor] I think it's more easy to understand by saying `if "-std=c++" not in arg`

tools/distrib/python/make_grpcio_tools.py
```diff
@@ -71,7 +71,9 @@
 # will be added to include path when building grpcio_tools
 CC_INCLUDES = [
     os.path.join("third_party", "abseil-cpp"),
+    os.path.join("third_party", "protobuf"),
     os.path.join("third_party", "protobuf", "src"),
+    os.path.join("third_party", "protobuf", "upb"),
```
XuanWang-Amos, line 76: Can we remove those two lines if we already have `os.path.join("third_party", "protobuf")`?

XuanWang-Amos (approved):

tools/distrib/python/grpcio_tools/setup.py
veblush: Good point. I'll take care of this in a separate PR to clean this up on other usage as well.

tools/distrib/python/make_grpcio_tools.py
veblush: If protobuf has other directories other than src and upb as an example, having a single protobuf is copying those unnecessary files?

PR #35923, opened 2024-02-16
Use clock_gettime on Apple platforms

dmaclach:
Apple platforms have supported clock_gettime for several years now (since 2016).

Simplify implementation and remove required api call to `mach_absolute_time`.

PR #36080, opened 2024-03-08
[ObjC] remove NSPrivacyAccessedAPICategorySystemBootTime from privacy manifest

HannahShiSFB:
#35923 removed the use of mach_absolute_time in gpr.

There is still one usage in protobuf, which they have removed the usage in https://github.com/protocolbuffers/protobuf/pull/15554 but we haven't update yet.
However gRPC library does not depends on protobuf but upb, so we should no longer need NSPrivacyAccessedAPICategorySystemBootTime any more.


sampajano (approved):
Woohoo! Thanks! 😃

sampajano (approved):
Thank you!!

PR #36087, opened 2024-03-08
[OTel] Add CI support for tests via CMake

yashykt:
Earlier, the tests just had bazel support. With CMake support added in #36063, we can also add CI CMake support for the tests. A major benefit of this is that we also get coverage for the various platforms that we test from our portability test suite.

### DIFF
--- a/tools/run_tests/helper_scripts/build_cxx.sh
+++ b/tools/run_tests/helper_scripts/build_cxx.sh
@@ -15,13 +15,32 @@
 
 set -ex
 
+# Set install path to avoid installing to system paths
 cd "$(dirname "$0")/../../.."
+mkdir -p cmake/install
+INSTALL_PATH="$(pwd)/cmake/install"
 
+# Install abseil-cpp since opentelemetry CMake uses find_package to find it.
+cd third_party/abseil-cpp
+mkdir build
+cd build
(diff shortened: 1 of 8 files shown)


tools/run_tests/helper_scripts/build_cxx.sh
```diff
@@ -15,13 +15,32 @@
...
+# Set install path to avoid installing to system paths
 cd "$(dirname "$0")/../../.."
+mkdir -p cmake/install
+INSTALL_PATH="$(pwd)/cmake/install"
 
+# Install abseil-cpp since opentelemetry CMake uses find_package to find it.
+cd third_party/abseil-cpp
+mkdir build
+cd build
+cmake -DCMAKE_CXX_STANDARD=14 -DABSL_BUILD_TESTING=OFF -DCMAKE_BUILD_TYPE="${MSBUILD_CONFIG}" -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" "$@" ..
```
veblush, line 27: Is it possible to omit this option for all cmake invocations?

veblush (approved):

tools/run_tests/helper_scripts/build_cxx.sh
yashykt: which option?

yashykt: The `-DCMAKE_CXX_STANDARD=14` is needed for MacOS builds

PR #36088, opened 2024-03-08
[server] Allow configuring max incoming connections at the server

Vignesh2208:


ctiller (commented):
we will probably want to make this a little more pluggable in the future, but this is a fine place to start

Vignesh2208:
> we will probably want to make this a little more pluggable in the future, but this is a fine place to start

Yeah I think there's some scope to make pluggable

PR #36095, opened 2024-03-12
[ObjC] log when got GRPC_QUEUE_TIMEOUT event from  completion queue

HannahShiSFB:
This is for the tat task app sometimes crash with "Unrecognized completion type".
However in obj-c wrapper, it is calling grpc_completion_queue_next with gpr_inf_future where it should never receive a timeout event.
This PR add a log and reschedule another grpc_completion_queue_next for a GRPC_OP_COMPLETE or GRPC_QUEUE_SHUTDOWN event, which is similar to the [gRPC C++ implementation](https://github.com/grpc/grpc/blob/v1.62.1/src/cpp/common/completion_queue_cc.cc#L80-L85)


src/objective-c/GRPCClient/private/GRPCCore/GRPCCompletionQueue.mm
```diff
@@ -69,6 +69,9 @@ - (instancetype)init {
             case GRPC_QUEUE_SHUTDOWN:
               grpc_completion_queue_destroy(unmanagedQueue);
               return;
+            case GRPC_QUEUE_TIMEOUT:
+              NSLog(@"GRPC_QUEUE_TIMEOUT, success: %d, tag: %p", event.success, event.tag);
+              break;
```
sampajano, line 74: Would this change silently disregard rpc errors?

I wonder if we should call `handler` with a failure event?

sampajano (commented):
Thanks for working on this, Hannah! One question inline!

src/objective-c/GRPCClient/private/GRPCCore/GRPCCompletionQueue.mm
HannahShiSFB: Not very sure but my understanding is that the completion queue is not supposed to timeout unless passed a finite timeout value.
The change in this PR will retry getting a event from the completion queue, which may indicate connection timeout in the event.

The crash report only had the stack trace, it may be helpful if we have the logs before the crash.

HannahShiSFB: I tried return an error from [pollset_work](https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/ev_apple.cc#L263) to simulate the queue timeout error, it appears the handler was not setup and trying to use it to propagate the error still crashes.


Screenshot 2024-04-21 at 8 44 07 AM

sampajano: Aha ok thank you! So are you saying that you'd prefer to eliminate the crash in this PR as well? (Besides only adds logging?)

In any case I trust your judgement here since you've given enough thoughts on this 😃 

We can chat when we meet :)

sampajano (approved):
Thanks for the change! 😊

src/objective-c/GRPCClient/private/GRPCCore/GRPCCompletionQueue.mm
sampajano: Please let me know what the client should expect to find after this change. (e.g. the crash goes away? And when to check logs? etc.) 😃

HannahShiSFB:
All green now

PR #36105, opened 2024-03-12
[CI] Moved to Bazel 7

veblush:
Let's use Bazel 7 by default.


gnossen:
@XuanWang-Amos Are we still skipping the Bazel 7 tests?

XuanWang-Amos:
> @XuanWang-Amos Are we still skipping the Bazel 7 tests?

Yes, but we're enabling them in this PR.

@veblush do you have updates? Last time I checked, we're still having virtual import issue mentioned here: https://github.com/grpc/grpc/issues/35391

veblush:
I tried to see whether this issue is still relevant but it turned out it's the case; [log](https://btx-internal.corp.google.com/invocations/bde45bbf-85c1-4fda-8530-f7a0e06e4e3a/targets/%2F%2Ftools%2Fbazelify_tests%2Ftest:bazel_distribtest_7.1.0_distribtest_python;config=daa91819a258a80cb1ea70735ab0d7c451588e3a7b9aeb02de68149cb1f255ae/tests)) so I need to disable `distribtest_python` for Bazel 7 again.

@gnossen @XuanWang-Amos What do you think about using Bazel 7 with `distribtest_python` disabled only for Bazel 7? (This test is still run with Bazel 6) I'd like to make this change as Bazel 7 support is needed and Bazel 6 will drop from the gRPC's supported Bazel versions ~6 months later.

veblush:
Oh I just noticed that both Xuan and Richard are our for 2 weeks. Then I'd like to get this merged first and consider rolling this back when they come back and if this has an issue with Python. I'd like to note that this PR doesn't change the test coverage. (`distribtest_python` will continue running with Bazel 6 only)

yashykt (approved):

PR #36113, opened 2024-03-13
[docs] Clarify expected semantics of deflate compression

akshayjshah:
I was surprised to see that the C++ core interprets "grpc-accept-encoding: deflate" like HTTP: it compresses messages using zlib with deflate compression, rather than using raw deflate. Matching HTTP semantics makes sense, of course, but this misnomer has historically confused HTTP implementations too.

This detail didn't seem important enough to warrant inclusion in the main protocol specification, so I've proposed an addition to `docs/compression.md` to document the expected behavior. I'm happy to move the addition around if you'd prefer it elsewhere.

### DIFF
--- a/doc/compression.md
+++ b/doc/compression.md
@@ -90,6 +90,15 @@ The inheritance of the compression configuration by child RPCs is left up to the
 implementation. Note that in the absence of changes to the parent channel, its
 configuration will be used.
 
+### Deflate Compression
+
+Like HTTP implementations, gRPC implementations MUST use the "deflate"
+compression to mean the zlib structure (defined in 
+[RFC 1950](https://datatracker.ietf.org/doc/html/rfc1950))
+with the deflate compression algorithm (defined in 
+[RFC 1951](https://datatracker.ietf.org/doc/html/rfc1951)).
+Servers and clients MUST NOT send raw deflate data.
+
 ### Test cases
 
 1. When a compression level is not specified for either the channel or the


yashykt:
@ejona86 @dfawley Can you verify that Java and Go do the same?

dfawley:
gRPC-Go only supports `"gzip"` out of the box.

akshayjshah:
> If C++ treats deflate like gzip, why does it support deflate? I wonder if it was implemented wrong first and then changed and the old flow wasn't removed.

For a protocol built on HTTP, the behavior makes some sense as implemented: it matches HTTP's `deflate` content-coding ([RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-deflate-coding)). This confusing choice of terminology has led to a ton of problems in HTTP implementations over the years, including [incorrect behavior in .NET](https://github.com/dotnet/runtime/issues/38022) and [IIS](https://stackoverflow.com/questions/9170338/why-are-major-web-sites-using-GZIP/9186091#9186091).

At this point, protocols like gRPC are in a tough spot: the standard is confusing, but deviating from the standard is just as bad.

FWIW, we only started looking into the expected behavior of `deflate` because it's explicitly mentioned in the main gRPC protocol specification:

> Content-Coding → "identity" / "gzip" / "deflate" / "snappy" / {custom}

I assume that block vs. stream compression makes `snappy` equally confusing, but I didn't propose a clarification because AFAIK none of the first-party gRPC implementations support snappy.

ejona86:
Looking at things, I think this is the expectation. You aren't saying gzip was used, which has additional framing on-top. Instead you are saying deflate means zlib, with adler32 and such. That's the standard approach; even png uses the zlib framing.

ejona86 (approved):

yashykt (approved):

PR #36116, opened 2024-03-13
[Build] Migrate Android NDK, android rules and platforms

sifmelcara:
This commit does the following 3 things for Android bazel rules, which depends on each other

1. Update NDK to 26.2.11394342 (and a few other Android things)
2. Use rules_android_ndk instead of builtin `native.android_ndk_repository`
    * `third_party/android/android_configure.bzl` was a workaround for users who don't set `$ANDROID_NDK_HOME` env var. Now with rules_android_ndk, we can declare repo without registering the toolchain. Instead users who need NDK toolchain should use `--extra_toolchains` to manually register the toolchain, as shown in README.md.
3. Migrate to platforms. See https://bazel.build/concepts/platforms
    * Currently we declares android platforms that is needed for binder transport APK. Later gRPC repo can gradually migrate to platforms for other platforms.
    * The value of `crosstool_top` will still be `//external:android/crosstool` (which is the default value), so existings android `config_settings` will still work. We should migrate them to match with `@platforms//os:android` constraint later.

The platforms migration needs Bazel 7.0 to work so we also override bazel version in testing scripts.


veblush:
Thanks for the change. LGTM and I'll approve this once all tests are passed.

veblush:
FYI, there is https://github.com/grpc/grpc/pull/36105 changing the default version of Bazel to 7, having a build issue in running "Bazel BinderTransport APK building" ([log](https://btx-internal.corp.google.com/invocations/69edd2fb-7a2e-494a-8a48-959815919a65)).

tools/internal_ci/linux/grpc_binder_transport_apk_build_in_docker.sh
```diff
@@ -18,19 +18,15 @@ set -ex
 echo $ANDROID_HOME
 echo $ANDROID_NDK_HOME
 
-# Build all targets using the strict warning option which leverages the
-# clang compiler to check if sources can pass a set of warning options.
-# CPU are specified because gRPC does not build with 32bit NDK (which has socklen_t
-# defined as int due to an accident).
-# The python option is for disabling python2 enforcement when packing APK
+# Android platforms only works with Bazel >= 7.0
+export OVERRIDE_BAZEL_VERSION=7.0.0
```
sifmelcara, line 22: Is this the right way to force Bazel 7.0?

third_party/android/android_configure.bzl
```diff
@@ -1,62 +0,0 @@
-"""Repository rule for Android SDK and NDK autoconfiguration.
```
sifmelcara, line 1: @gnossen for removal of this workaround (in case you have any concern)

tools/internal_ci/linux/grpc_binder_transport_apk_build_in_docker.sh
sifmelcara: This will break after Bazel 7.0.0 become unsupported, we can simply remove this when we drop support for Bazel < 7.0.0

sifmelcara:
> FYI, there is https://github.com/grpc/grpc/pull/36105 changing the default version of Bazel to 7, having a build issue in running "Bazel BinderTransport APK building" ([log](https://btx-internal.corp.google.com/invocations/69edd2fb-7a2e-494a-8a48-959815919a65)).

Got it, thanks! This PR should fix the Bazel 7 build issue

sifmelcara:
cc @jdcormie @rock1246

tools/internal_ci/linux/grpc_binder_transport_apk_build_in_docker.sh
veblush: It looks good. Yeah we're going to stop supporting Bazel 6.x ~6 months after Bazel 7 upgrade is done. Then, we can just remove this line.

veblush (approved):

veblush:
@gnossen Would you do a second review on this?

sifmelcara:
(rebased to use bazel 7.1.0 , and some fixes for copybara internal change)

veblush:
@drfloob AJ, would you do a second review? (I forgot @gnossen is out this week)

PR #36122, opened 2024-03-14
[Security Tests - Crl Provider] Fix inconsistent directory related tests

gtcooke94:
There were some failures in the Crl Directory related tests after https://github.com/grpc/grpc/pull/36031

This came down to https://github.com/grpc/grpc/pull/36031 adding some CRLs with bad qualities (invalid content/signatures, overriding issuer names) to the `test_creds/crl_data/crls` directory, which is used in the directory reloading tests. The tests began failing on some platforms because they were picking up these bad crls which were failing various checks, but the test was designed to assume that `test_creds/crl_data/crls` was a valid and good directory.
This PR moves the bad CRLs to their own directory to prevent this accidental mash-up of test data. It also adds debug logging to our custom verification stack.


src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -259,12 +259,19 @@ bool VerifyCrlSignature(X509_CRL* crl, X509* issuer) {
...
+  int ret = X509_CRL_verify(crl, ikey);
+  if (ret < 0) {
+    gpr_log(GPR_DEBUG,
+            "There was an unexpected problem checking the CRL signature.");
+  } else if (ret == 0) {
+    gpr_log(GPR_DEBUG, "CRL could not be verified.");
+  }
   EVP_PKEY_free(ikey);
-  return ret;
+  return ret == 1;
```
matthewstevenson88, line 274: Do we need a unit test that detects this functional change? Said differently, do we have any test CRL that would trigger `X509_CRL_verify` to return -1?

matthewstevenson88 (commented):
LGTM with 1 question about testing.

src/core/tsi/ssl_transport_security_utils.cc
gtcooke94: It's not a behavior change, previously the line was `bool ret = X509_CRL_verify(crl, ikey) == 1;`

Now we just temporarily look at the `int` because a `0` return means verification failed, but `-1` indicates that something happened that it couldn't verify at all.

matthewstevenson88: Whoops, I misread! Thanks for clarifying!

PR #36124, opened 2024-03-14
[Build] Massage grpcio_tools build configuration for protobuf upgrade.

veblush:
This is a prerequisite for the upcoming Protobuf v26 upgrade and has the following changes;

- Reduced the optimization level from `O3` to `O1` on Linux to workaround a segfault issue on `manylinux2014/x86`. This is believed to be caused by an aggressive but buggy optimization that `gcc` 10 made so workaround is to reduce the optimization level. `gcsio_tools` isn't that performance sensitive so we should be able to afford this change. (internal b/329134877)
- Added a Windows library dependency `Shell32.lib` to support the use of `CommandLineToArgvW`. (internal b/328455319)

### DIFF
--- a/tools/distrib/python/grpcio_tools/setup.py
+++ b/tools/distrib/python/grpcio_tools/setup.py
@@ -185,6 +185,9 @@ if EXTRA_ENV_COMPILE_ARGS is None:
         EXTRA_ENV_COMPILE_ARGS += " -fno-wrapv -frtti"
 if EXTRA_ENV_LINK_ARGS is None:
     EXTRA_ENV_LINK_ARGS = ""
+    # This is needed for protobuf/main.cc
+    if "win32" in sys.platform:
+        EXTRA_ENV_LINK_ARGS += " Shell32.lib"
     # NOTE(rbellevi): Clang on Mac OS will make all static symbols (both
     # variables and objects) global weak symbols. When a process loads the
     # protobuf wheel's shared object library before loading *this* C extension,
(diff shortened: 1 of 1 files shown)


tools/distrib/python/grpcio_tools/setup.py
```diff
@@ -152,14 +151,7 @@ def new_compile(obj, src, ext, cc_args, extra_postargs, pp_opts):
 
         self.compiler._compile = new_compile
 
-        try:
-            build_ext.build_ext.build_extensions(self)
-        except Exception as error:
```
XuanWang-Amos, line 157: Can we keep this `except` and diagnose part?

veblush: Usually we can get the build error (through stderr) regardless of this change and this code needs some change if we want to keep (e.g. class `CommandError` turned out to not exist in this module). Any value of having this one?

XuanWang-Amos: @gnossen Look like we're only diagnose `AttributeError` and `CompileError`, `AttributeError` is mainly for older versions of setuptools which has been there for 8 years and doesn't seems to be relevant anymore.  Not so sure about `CompileError` though, do you think we should still keep those diagnoses?

gnossen: In practice, no one's reporting this error, so I think it's likely that it's never being hit. It's probably fine to remove this try-except block.

XuanWang-Amos (approved):

PR #34903, opened 2023-11-08
Revert "[ruby] disable ruby fork tests (#34443)"

apolcyn:
This reverts commit 74048f118b36bcdf0e680be7ddc28bcad02a71bf.

Looks like https://github.com/grpc/grpc/issues/34442 has been fixed by https://github.com/grpc/grpc/pull/33733

FYI @casperisfine

PR #36129, opened 2024-03-14
[fork] Ensure the default EventEngine is not destroyed during fork

drfloob:


apolcyn (approved):
Noting to self that we could resume with https://github.com/grpc/grpc/pull/34903 assuming tests here are good

PR #36134, opened 2024-03-15
[channel] canonify target and set channel arg in only one place

markdroth:
Channel target will now be reported in canonified form (e.g., in channelz).


yijiem (approved):
Thanks Mark! Nice simplification.

PR #36136, opened 2024-03-16
[time] Use grpc_core::Timestamp for start times used to compute deadlines

### LINKED ISSUE #34547: RPC timeout set via service config is off by ~1s
grpcio==1.54.2

OSX Ventura 13.6

Python 3.11.4

### What did you do?
tl;dr I find that setting an RPC via service config needs ~1 full second higher than the client/server will actually take.  Happens across a variety of timings, and ONLY applies to the service config and NOT to setting a timeout at the actual RPC call.  

Started a simple gRPC server that sleeps for a small time.  Then perform a client call with a `timeout` in the service config for 1 second more than the configured sleep, which fails unexpectedly and in a shorter time than the timeout.  

The below example should sleep for .2s, should timeout in 1.2s, but the client returns a DEADLINE_EXCEEDED in less than .2s.

#### Server
```python
from concurrent import futures
import time

import grpc
import greeting_pb2
import greeting_pb2_grpc


class Greeter(greeting_pb2_grpc.GreeterServicer):
    def greet(self, request, context):
        print("Got request " + str(request))
        time.sleep(0.2)
        return greeting_pb2.ServerOutput(message="{0} {1}!".format(request.greeting, request.name))


def server():
    server = grpc.server(futures.ThreadPoolExecutor(max_workers=2))
    greeting_pb2_grpc.add_GreeterServicer_to_server(Greeter(), server)
    server.add_insecure_port("[::]:50051")
    print("gRPC starting")
    server.start()
    server.wait_for_termination()


server()
```

#### Client
```python
import time
import grpc
import greeting_pb2
import greeting_pb2_grpc

options = [("grpc.service_config", '{"methodConfig": [{"timeout": "1.2s", "name": [{}]}] }')]

channel = grpc.insecure_channel(
    target="dns:///localhost:50051",
    options=options,
)

stub = greeting_pb2_grpc.GreeterStub(channel)
start = time.perf_counter()
try:
    response = stub.greet(greeting_pb2.ClientInput(name="Me", greeting="Yo"))
except Exception as e:
    response = e
print(response, " in ", time.perf_counter() - start)
```

I expect this RPC to succeed.  

### What did you see instead?
This RPC fails in a shorter time than the timeout.  For the example above it fails at 0.14ish seconds rather than 1.2ish seconds.  

```
<_InactiveRpcError of RPC that terminated with:
	status = StatusCode.DEADLINE_EXCEEDED
	details = "Deadline Exceeded"
	debug_error_string = "UNKNOWN:Deadline Exceeded {grpc_status:4, created_time:"2023-09-29T11:00:52.966757-05:00"}"
>  in  0.14907191699603572
```

This will then succeed if I change the timeout to 1.3 seconds, and this pattern is repeatable for higher values.  e.g. a server that sleeps for 3 seconds needs a `serviceConfig` timeout of ~4.1 seconds to succeed.  

However, using the timeout in the RPC call instead seems fine.  It doesn't seem to have the same off-by-one issue that I find in the service config.

yashykt:
Fixes #34547

It turns out that there is a difference between the now value between gpr_cycle_counter and gpr_timespec when converted to grpc_core::Timestamp and this difference is around a second.

The main reason for this is - https://github.com/grpc/grpc/blob/c68e6c27ca54a8b7b98a21e6315c805f235d5dda/src/core/lib/gprpp/time.cc#L73 
which was introduced to fix time related bugs.
Additionally, we throw away the nsec component of the gpr_timespec to be efficient (since we are using std::atomic).

All in all, we don't want to mix and match `gpr_timespec` with `gpr_cycle_counter`. If we ever feel the need to, we should fix the discrepancy first.

### DIFF
--- a/src/core/lib/surface/call.cc
+++ b/src/core/lib/surface/call.cc
@@ -245,7 +245,14 @@ class Call : public CppImplOf<Call, grpc_call> {
   void HandleCompressionAlgorithmNotAccepted(
       grpc_compression_algorithm compression_algorithm) GPR_ATTRIBUTE_NOINLINE;
 
+  // This should only be used for call latency calculation (gpr_cycle_counter is
+  // nice for profiling), and not for any other reason such as checking for
+  // deadlines. There's currently a discrepancy when mixing gpr_cycle_counter
+  // and gpr_timespec when converting to grpc_core:::Timestamp. Prefer to use
+  // start_timestamp() for general gRPC functionality.
   gpr_cycle_counter start_time() const { return start_time_; }
+
+  grpc_core::Timestamp start_timestamp() const { return start_timestamp_; }
 
  private:
   RefCountedPtr<Channel> channel_;
@@ -269,6 +276,7 @@ class Call : public CppImplOf<Call, grpc_call> {
   mutable Mutex peer_mu_;
   Slice peer_string_;
   gpr_cycle_counter start_time_ = gpr_get_cycle_counter();
+  grpc_core::Timestamp start_timestamp_ = Timestamp::Now();
 };
 
 Call::ParentCall* Call::GetOrCreateParentCall() {
@@ -855,10 +863,10 @@ grpc_error_handle FilterStackCall::Create(grpc_call_create_args* args,
   }
   // initial refcount dropped by grpc_call_unref
   grpc_call_element_args call_args = {
-      call->call_stack(), args->server_transport_data,
-      call->context_,     path,
-      call->start_time(), call->send_deadline(),
-      call->arena(),      &call->call_combiner_};
+      call->call_stack(),      args->server_transport_data,
+      call->context_,          path,
+      call->start_timestamp(), call->send_deadline(),
+      call->arena(),           &call->call_combiner_};
   add_init_error(&error, grpc_call_stack_init(channel_stack, 1, DestroyCall,
                                               call, &call_args));
   // Publish this call to parent only after the call stack has been initialized.
--- a/src/core/client_channel/client_channel_filter.cc
+++ b/src/core/client_channel/client_channel_filter.cc
@@ -268,8 +268,13 @@ class ClientChannelFilter::FilterBasedCallData
       ABSL_EXCLUSIVE_LOCKS_REQUIRED(&ClientChannelFilter::resolution_mu_);
 
   void ResetDeadline(Duration timeout) override {
-    const Timestamp per_method_deadline =
-        Timestamp::FromCycleCounterRoundUp(call_start_time_) + timeout;
+    gpr_log(GPR_ERROR, "here");
+    const Timestamp per_method_deadline = call_start_time_ + timeout;
+    gpr_log(GPR_ERROR, "now: %s, deadline: %s, per_method: %s cycle counter:%s",
+            grpc_core::Timestamp::Now().ToString().c_str(),
+            deadline_.ToString().c_str(),
+            per_method_deadline.ToString().c_str(),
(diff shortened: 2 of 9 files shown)


src/core/lib/surface/call.cc
```diff
@@ -269,6 +276,7 @@ class Call : public CppImplOf<Call, grpc_call> {
   mutable Mutex peer_mu_;
   Slice peer_string_;
   gpr_cycle_counter start_time_ = gpr_get_cycle_counter();
+  Timestamp start_timestamp_ = Timestamp::Now();
```
markdroth, line 279: Seems like storing this timestamp in two different forms would increase per-call memory, which seems undesirable.  Do we really not have an alternative here?

yashykt: I was thinking that maybe the cycle counter is useful for profiling, but maybe that is not enough of a justification to increase the call size. I think I'll just save a `gpr_timespec`

yashykt: Changed

ctiller: Just to note:

```
sizeof(Timestamp)==8
sizeof(gpr_cycle_counter)==8
sizeof(gpr_timespec)==16
```

But I'm supportive of getting this down to one time query!

(and I think the ultimate fix here is to move `grpc_core::Timestamp` to be nanosecond precise)

markdroth: If `gpr_timespec` is larger, switching to that type is still a regression.

If we change `grpc_core::Timestamp` to nanosecond precision, would that change its size?  If not, maybe we should just go ahead and do that.

Are we worried about latency calculations using `Timestamp` instead of `gpr_cycle_counter` in terms of NTP clock smearing?

markdroth:
Alternatively, can we adjust for the 1-second offset systemically somehow?

src/core/lib/surface/call.cc
ctiller: `Timestamp` is monotonic, so no NTP clock interactions.

I'm not sure that we've done much deep thinking about storage format for nanosecond sized times.

One option would be what absl/gpr_timespec do: a seconds `uint64_t` and a microseconds `uint32_t` - this would increase the size of `Timestamp` everywhere but have some nice compatibility story.

Another would be integer nanoseconds from monotonic clock epoch - which has a maximum range of about 500 years (and worst case starting time of Jan 1 1970 I guess? -- so we'd get about 448 more years).

Another would be floating point nanoseconds from monotonic clock epoch - makes math marginally easier - and probably gets us the precision we need, but we'd need to study a bit.

yashykt: > Another would be integer nanoseconds from monotonic clock epoch - which has a maximum range of about 500 years (and worst case starting time of Jan 1 1970 I guess? -- so we'd get about 448 more years).

I think this is fine. We could probably also consider microsecond? 

> Another would be floating point nanoseconds from monotonic clock epoch - makes math marginally easier - and probably gets us the precision we need, but we'd need to study a bit.

I don't see the point of using floating point for time. We'd be losing precision compared to `uint64_t`. (The precision for 64 bit floating point is 53 bits.)

sarthak77:
@yashykt any timelines for merging this fix?

ctiller:
So I think the conclusion here is to migrate Timestamp to uint64_t nanos?

yashykt:
> So I think the conclusion here is to migrate Timestamp to uint64_t nanos?

That's my recollection

ctiller:
Closing then :)

PR #36143, opened 2024-03-18
[posix EventEngine] fix ResolvedAddressIsWildcard documentation

nipil:
Fix the wording of the description of the function for the return value : current version says the opposite of what the function does. The current behaviour is to return the port used, but only when the address _is actually_ a wildcard address : 0.0.0.0 (ipv4) or :: (ipv6). Plus mention both ipv4 and ipv6 in the description.


drfloob:
Thanks @nipil, good catch.

drfloob (approved):

eugeneo (approved):

PR #36145, opened 2024-03-18
[xds] Fallback implementation

eugeneo:
As per [gRFC A71](https://github.com/grpc/proposal/blob/master/A71-xds-fallback.md).


src/core/ext/xds/xds_client.cc
```diff
@@ -966,6 +1017,38 @@ XdsClient::XdsChannel::AdsCall::AdsCall(
...
+        std::find(channels.begin(), channels.end(), xds_channel());
+    // Skip if channel is not on the list
+    if (channel_it == channels.end()) {
+      continue;
+    }
+    gpr_log(GPR_INFO, "[xds_client %p] Falling forward to %s", this,
+            xds_channel()->server_.server_uri().c_str());
+    // Lower priority channels are no longer needed, connection is back!
+    channels.erase(channel_it + 1, channels.end());
+    // Restore all resource subscriptions
```
markdroth, line 1040: This should not be needed here.  As per comments elsewhere, we should just do this when the watchers are added, even if the channel status is not ok.

src/core/ext/xds/xds_client.cc
```diff
@@ -1074,6 +1157,7 @@ void XdsClient::XdsChannel::AdsCall::OnRecvMessage(absl::string_view payload) {
   {
     MutexLock lock(&xds_client()->mu_);
     if (!IsCurrentCallOnChannel()) return;
+    FallForwardIfNeededLocked();
```
markdroth, line 1160: I think this should move down to line 1176, where we set `seen_response_` and set the channel's status to OK.  We don't want to do this unless we've actually gotten back a valid response.

markdroth (commented):
This looks like a good start!  Haven't had a chance to review the tests yet but wanted to send you my comments on the code.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -0,0 +1,208 @@
...
+  void SetUp() override {
+    // Overrides SetUp from a base class so we can call InitClient per-test case
+  }
+
+  void TearDown() override {
+    fallback_balancer_->Shutdown();
+    XdsEnd2endTest::TearDown();
+  }
+
+  std::string SetupServer(BalancerServerThread* balancer, size_t backend,
```
markdroth, line 68: Please call this `SetXdsResourcesForTarget()`.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -0,0 +1,208 @@
...
+                          int server_id) {
```
markdroth, line 69: This parameter should be named `target_id`.  It should default to 0.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
markdroth, line 68: Please document that this returns the target name.

test/cpp/end2end/xds/BUILD
```diff
@@ -491,3 +491,25 @@ grpc_cc_test(
...
+        "no_test_ios",
+        "no_windows",
+        "xds_end2end_test",
+    ],  # TODO(jtattermusch): fix test on windows
+    deps = [
+        ":xds_end2end_test_lib",
+        "//:gpr",
+        "//:grpc",
+        "//:grpc++",
+        "//:grpcpp_csds",
```
markdroth, line 512: I don't think this dependency is needed.

test/cpp/end2end/xds/BUILD
```diff
@@ -491,3 +491,25 @@ grpc_cc_test(
...
+        "//test/core/util:scoped_env_var",
```
markdroth, line 513: This reminded me that we do need to use an env var guard around the bootstrap code that reads more than one server entry in a server list.

markdroth (commented):
Here are my comments on the tests.

eugeneo (commented):
Thank you for reviews. I am still adding test cases but the core code is ready for review.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -65,21 +62,26 @@ class XdsFallbackTest : public XdsEnd2endTest {
     XdsEnd2endTest::TearDown();
   }
 
-  std::string SetupServer(BalancerServerThread* balancer, size_t backend,
-                          int server_id) {
+  // Sets resources and returns target name
```
markdroth, line 65: I think the tests would be a bit more readable if we encoded the target name in the tests themselves and passed it in here, rather than constructing it here and returning it.  I'm thinking something like this:

```
void SetXdsResourcesForTarget(BalancerServerThread* balancer, size_t backend,
                              absl::string_vierw target = "") {
  // ...
  if (!target.empty()) {
    listener.set_name(std::string(target));
    cluster.set_name(absl::StrCat("cluster_", target));
    cluster.mutable_eds_cluster_config()->set_service_name(
        absl::StrCat("eds_", target));
    route_config.set_name(absl::StrCat("route_", target));
    route_config.mutable_virtual_hosts(0)
        ->mutable_routes(0)
        ->mutable_route()
        ->set_cluster(cluster.name());
  }
  // ...
}
```

Then in individual tests, something like:

```
constexpr absl::string_view kServerName2 = "server2.example.com";
SetXdsResourcesForTarget(balancer_.get(), 2, kServerName2);
SetXdsResourcesForTarget(fallback_balancer_.get(), 3, kServerName2);
```

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -65,21 +62,26 @@ class XdsFallbackTest : public XdsEnd2endTest {
...
       cluster.mutable_eds_cluster_config()->set_service_name(
-          absl::StrFormat("eds%d", server_id));
-      route_config.set_name(absl::StrFormat("route%d", server_id));
+          absl::StrFormat("eds%d", target_id));
+      route_config.set_name(absl::StrFormat("route%d", target_id));
       route_config.mutable_virtual_hosts(0)
           ->mutable_routes(0)
           ->mutable_route()
           ->set_cluster(cluster.name());
+    } else if (!authority.empty()) {
```
markdroth, line 84: I assume something is going to be added here to set xdstp-style names.

test/cpp/end2end/xds/xds_end2end_test_lib.h
```diff
@@ -439,7 +444,8 @@ class XdsEnd2endTest : public ::testing::TestWithParam<XdsTestType>,
   // Creates and starts a new balancer, running in its own thread.
   // Most tests will not need to call this; instead, they can use
   // balancer_, which is already populated with default resources.
-  std::unique_ptr<BalancerServerThread> CreateAndStartBalancer();
+  std::unique_ptr<BalancerServerThread> CreateAndStartBalancer(
+      absl::string_view debug_label = "Default Balancer");
```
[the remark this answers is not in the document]
markdroth: To be clear, I'm fine with having a fixed string for the default balancer now that you've added the address to the debug string.  However, I suggest that instead of encoding the default here, we do it [where we initialize `balancer_`](https://github.com/grpc/grpc/blob/04594ce92add74d8ec3b7bcb80c0f35549c2a41f/test/cpp/end2end/xds/xds_end2end_test_lib.cc#L383).  That way, other tests that call `CreateAndStartBalancer()` won't accidentally get the default string.

src/core/ext/xds/xds_client.cc
```diff
@@ -966,6 +1017,38 @@ XdsClient::XdsChannel::AdsCall::AdsCall(
   streaming_call_->StartRecvMessage();
 }
 
+// Make this channel active iff:
+// 1. Channel is on the list of authority channels
+// 2. Channel is not the last channel on the list (i.e. not the active channel)
+void XdsClient::XdsChannel::AdsCall::FallForwardIfNeededLocked() {
+  for (auto& a : xds_client()->authority_state_map_) {
+    auto channels = a.second.xds_channels;
```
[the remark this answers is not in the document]
markdroth: All I was really asking for here was to change `auto` to `auto&`.  I don't think we should pull the loop out of this method into the caller.

See my comment elsewhere about moving this method to `XdsClient`.

src/core/ext/xds/xds_client.cc
```diff
@@ -1646,13 +1731,20 @@ void XdsClient::WatchResource(const XdsResourceType* type,
               },
           DEBUG_LOCATION);
     }
-    // If the authority doesn't yet have a channel, set it, creating it if
-    // needed.
-    if (authority_state.xds_channel == nullptr) {
-      authority_state.xds_channel =
-          GetOrCreateXdsChannelLocked(*xds_server, "start watch");
+    RefCountedPtr<XdsClient::XdsChannel> xds_channel;
+    for (size_t i = 0; i < xds_servers.size(); ++i) {
```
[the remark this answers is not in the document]
markdroth: Good catch!

markdroth (commented):

src/core/ext/xds/xds_client.cc
```diff
@@ -340,7 +336,8 @@ class XdsClient::XdsChannel::AdsCall : public InternallyRefCounted<AdsCall> {
   std::vector<std::string> ResourceNamesForRequest(const XdsResourceType* type)
       ABSL_EXCLUSIVE_LOCKS_REQUIRED(&XdsClient::mu_);
 
-  void FallForwardIfNeededLocked()
+  void FallForwardForAuthorityLocked(
```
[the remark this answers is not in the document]
eugeneo: I did sum shuffling and came up with `XdsChannel::SetHealthyLocked` - I do think this is fairly logical. Please let me know what you think.

eugeneo (commented):
Thank you for the reviews! I am still working on remaining test cases (marked DISABLED on this PR)

eugeneo:
CC: @allanrbo

src/core/ext/xds/xds_client.cc
```diff
@@ -1565,6 +1600,56 @@ RefCountedPtr<XdsClient::XdsChannel> XdsClient::GetOrCreateXdsChannelLocked(
...
+        authority_state.xds_channels.back()->SubscribeLocked(
+            type_resource.first, {authority, key_state.first});
+      }
+    }
+    if (authority_state.xds_channels.back()->status().ok()) {
+      gpr_log(GPR_INFO, "[xds_client %p] Performed fallback to %s", this,
+              xds_servers[i]->server_uri().c_str());
+      return true;
+    } else {
+      gpr_log(GPR_ERROR, "Fallback channel %s is in state %s",
```
markdroth, line 1644: Instead of having two different messages here, I suggest something like this:

```
if (GRPC_TRACE_FLAG_ENABLED(grpc_xds_client_trace)) {
  gpr_log(GPR_INFO, "[xds_client %p] authority %s: added fallback server %s (%s)",
          this, authority.c_str(), xds_servers[i]->server_uri().c_str(),
          authority_state.xds_channels.back()->status().ToString().c_str());
}
if (authority_state.xds_channels.back()->status().ok()) return tre;
```

src/core/ext/xds/xds_client.cc
```diff
@@ -1565,6 +1600,56 @@ RefCountedPtr<XdsClient::XdsChannel> XdsClient::GetOrCreateXdsChannelLocked(
...
+              xds_servers[i]->server_uri().c_str(),
+              authority_state.xds_channels.back()->status().ToString().c_str());
+    }
+  }
+  gpr_log(GPR_INFO, "[xds_client %p] No fallback server", this);
```
markdroth, line 1649: This log message should include the authority.

test/cpp/end2end/xds/xds_end2end_test_lib.cc
```diff
@@ -290,22 +290,23 @@ XdsEnd2endTest::BalancerServerThread::BalancerServerThread(
...
+          // features.
           [&](const DiscoveryRequest& request) {
             EXPECT_TRUE(request.has_node());
             EXPECT_THAT(request.node().client_features(),
                         ::testing::UnorderedElementsAre(
                             "envoy.lb.does_not_support_overprovisioning",
                             "xds.config.resource-in-sotw"));
           },
-          debug_label,
+
```
markdroth, line 302: Please remove blank line.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -63,40 +64,65 @@ class XdsFallbackTest : public XdsEnd2endTest {
   }
 
   // Sets resources and returns target name
```
markdroth, line 66: Please remove this comment now.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -63,40 +64,65 @@ class XdsFallbackTest : public XdsEnd2endTest {
...
+    SetListenerAndRouteConfiguration(fallback_balancer_.get(), listener,
+                                     new_route_config);
   }
 
  protected:
   std::unique_ptr<BalancerServerThread> fallback_balancer_;
 };
 
 TEST_P(XdsFallbackTest, FallbackAndRecover) {
+  grpc_core::testing::ScopedEnvVar fallback_enabled(
```
markdroth, line 124: Please add a test that shows that fallback does *not* work if the env var is unset.

test/cpp/end2end/xds/xds_end2end_test_lib.h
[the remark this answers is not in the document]
markdroth: I'm not sure if you saw this comment.

markdroth (commented):
Comments are mostly cosmetic.  Just waiting for the remaining tests.

test/cpp/end2end/xds/xds_end2end_test_lib.h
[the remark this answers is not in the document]
eugeneo: Please take a look.

eugeneo (commented):
Thank you for the comments, addressed. Working on the last test case I planned.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -90,18 +91,32 @@ class XdsFallbackTest : public XdsEnd2endTest {
...
   }
 
-  void ConfigureAuthority(const std::string& eds_server_name,
-                          const std::string& cluster_name,
-                          const std::string& route_config_name,
-                          const std::string& listener_name,
-                          const EdsResourceArgs& args) {
-    fallback_balancer_->ads_service()->SetEdsResource(
-        BuildEdsResource(args, eds_server_name));
+  void ConfigureAuthority(BalancerServerThread* balancer,
```
markdroth, line 94: I think we should combine this with the existing `SetXdsResourcesForTarget()` method, as you had originally started to do:

```
  void SetXdsResourcesForServer(BalancerServerThread* balancer, size_t backend,
                                absl::string_view server_name = "",
                                absl::string_view authority = "") {
    Listener listener = default_listener_;
    RouteConfiguration route_config = default_route_config_;
    Cluster cluster = default_cluster_;
    // Default server uses default resources when no authority, to enable using more
    // test framework functions.
    if (!server_name.empty() || !authority.empty()) {
      auto get_resource_name =
          [&](absl::string_view resource_type) {
            if (authority.empty()) {
              if (resource_type == kLdsTypeUrl) return std::string(server_name);
              return absl::StrFormat("%s_%s", resource_type, server_name);
            }
            return absl::StrFormat("xdstp://%s/%s/%s", authority, resource_type, server_name);
          };
      listener.set_name(get_resource_name(kLdsTypeUrl));
      cluster.set_name(get_resource_name(kCdsTypeUrl);
      cluster.mutable_eds_cluster_config()->set_service_name(get_resource_name(kEdsTypeUrl);
      route_config.set_name(get_resource_name(kRdsTypeUrl));
      route_config.mutable_virtual_hosts(0)
          ->mutable_routes(0)
          ->mutable_route()
          ->set_cluster(cluster.name());
    }
    SetListenerAndRouteConfiguration(balancer, listener, route_config);
    balancer->ads_service()->SetCdsResource(cluster);
    balancer->ads_service()->SetEdsResource(BuildEdsResource(
        EdsResourceArgs(
            {{"locality0", CreateEndpointsForBackends(backend, backend + 1)}}),
        cluster.eds_cluster_config().service_name()));
  }
```

Note that this approach will avoid the config tear that you're handling in the `PerAuthorityFallback` test below, because the resource names will be the same on the primary and secondary servers.

Also note that I went back to using the term "server" instead of "target" here, since in the authority case, the string is not the full target URI.

markdroth (commented):
Just a few minor comments remaining!

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
eugeneo: Done.

eugeneo (commented):
Thanks for the suggestions, applied. I am working on unit tests. Should we merge this and have them in another PR?

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -0,0 +1,317 @@
...
+      /*failover_timeout_ms=*/0, std::string(kServer1Name).c_str(),
+      kAuthority1));
+  auto authority2_stub = grpc::testing::EchoTestService::NewStub(CreateChannel(
+      /*failover_timeout_ms=*/0, std::string(kServer2Name).c_str(),
+      kAuthority2));
+  ExpectBackendCall(authority1_stub.get(), 0, DEBUG_LOCATION);
+  ExpectBackendCall(authority2_stub.get(), 1, DEBUG_LOCATION);
+  // Primary balancer is up, its data will be used now.
+  balancer_->ads_service()->ClearADSFailure();
+  auto deadline = absl::Now() + absl::Seconds(5);
```
markdroth, line 264: We should multiply the timeout by `grpc_test_slowdown_factor()`.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -0,0 +1,317 @@
...
+  while (absl::Now() < deadline &&
+         (backends_[2]->backend_service()->request_count() == 0 ||
+          backends_[3]->backend_service()->request_count() == 0)) {
+    ClientContext context;
+    EchoRequest request;
+    EchoResponse response;
+    RpcOptions().SetupRpc(&context, &request);
+    Status status = authority1_stub->Echo(&context, request, &response);
+    if (!status.ok()) {
```
markdroth, line 273: This should no longer fail now that we're using the same resource names on the primary and fallback servers.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -0,0 +1,317 @@
...
+      gpr_log(GPR_ERROR, "RPC failed, may be caused by a config tear: %s",
+              status.error_message().c_str());
+    }
+    ClientContext context2;
+    EchoRequest request2;
+    EchoResponse response2;
+    RpcOptions().SetupRpc(&context2, &request2);
+    // RPCs may briefly fail if the config tear happens
+    status = authority2_stub->Echo(&context2, request2, &response2);
+    if (!status.ok()) {
```
markdroth, line 283: Same here: This should no longer fail now that we're using the same resource names on the primary and fallback servers.

test/cpp/end2end/xds/xds_fallback_end2end_test.cc
```diff
@@ -0,0 +1,317 @@
...
+    Status status = authority1_stub->Echo(&context, request, &response);
+    if (!status.ok()) {
+      gpr_log(GPR_ERROR, "RPC failed, may be caused by a config tear: %s",
+              status.error_message().c_str());
+    }
+    ClientContext context2;
+    EchoRequest request2;
+    EchoResponse response2;
+    RpcOptions().SetupRpc(&context2, &request2);
+    // RPCs may briefly fail if the config tear happens
```
markdroth, line 281: This comment is no longer correct.

markdroth (commented):
> I am working on unit tests. Should we merge this and have them in another PR?

I think the tests should be part of this PR.

eugeneo (commented):
Addressed the comments and added a test case. PTAL.

markdroth (commented):
This is getting close!

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3804,237 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+          .AddFooResource(XdsFooResource("foo1", 6))
+          .Serialize());
+  // XdsClient should have delivered the response to the watcher.
+  auto resource = watcher->WaitForNextResource();
+  ASSERT_NE(resource, nullptr);
+  EXPECT_EQ(resource->name, "foo1");
+  EXPECT_EQ(resource->value, 6);
+  TriggerConnectionFailure(primary_server,
+                           absl::UnavailableError("Server down"));
+  // XdsClient should have sent a subscription request on the ADS stream.
```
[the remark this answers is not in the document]
eugeneo: Done

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3804,237 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+          .set_version_info("1")
+          .set_nonce("A")
+          .AddFooResource(XdsFooResource("foo1", 6))
+          .Serialize());
+  // XdsClient should have delivered the response to the watcher.
+  auto resource = watcher->WaitForNextResource();
+  ASSERT_NE(resource, nullptr);
+  EXPECT_EQ(resource->name, "foo1");
+  EXPECT_EQ(resource->value, 6);
+  TriggerConnectionFailure(primary_server,
```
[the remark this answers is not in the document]
markdroth: > I tried my best to implement your suggestions, ptal.

This is a lot better, but the order of steps still isn't quite right.  I think there are two principles here:
- For each input we trigger on the client, we should check all of the results of that input before we provide the next input.
- When checking the results of a given input, we should always check the local results (i.e., watcher events and metric data) before we remote results (i.e., requests sent to the server), because we shouldn't have to wait for remote events to complete (which may involve waiting for activity in other threads) before seeing the local results.

Given that, I think the order we want is:

1. Input: Get initial response from primary server.
   1. Result (local): Resource is delivered to watcher.
   2. Result (local): Metrics show 1 resource update and 1 cached resource.
   3. Result (remote): Client sends ACK to server.
2. Input: Trigger connection failure to primary.
   1. Result (local): The error is reported to the watcher.
   2. Result (local): The metrics show the channel as being unhealthy.
3. Input: Trigger stream failure.
   1. Result (local): The metrics still show the channel as being unhealthy.
   2. Result (remote): The client starts a new stream and sends a subscription message.
4. Input: Start second watch for foo1 (already cached).
   1. Result (local): New watcher gets the cached resource.
   2. Result (local): New watcher gets the error from the channel state.
5. Input: Start watch for foo2 (not already cached).
    1. Result (local): Metrics show a healthy channel to the fallback server.
    2. Result (remote): Client sent a new request for both resources on the stream to the primary.
    3. Result (remote): Client created a stream to the fallback server and sent a request on that stream for both resources.
6. Input: Fallback server sends a response with both resources.
   1. Result (local): Watcher for foo2 gets resource.
   2. Result (local): Metrics show an update from fallback server.
   3. Result (remote): Client sends ACK to fallback server.
7. Input: Primary server sends a response containing both resources.
   1. Result (local): Metrics show that we've closed the channel to the fallback server and received resource updates from the primary server.
   2. Result (remote): The stream to the fallback server has been orphaned.

Please add a one-line comment before each input and each result check, so that it's clear what's happening.

> Note that no error is returned for the original watcher here:

That seems like a bug.  We're not triggering fallback here, so we should report the error to the watcher.

test/core/xds/xds_client_test.cc
```diff
@@ -3991,48 +4033,125 @@ TEST_F(XdsClientTest, FallbackReportsError) {
...
   auto error = watcher->WaitForNextError();
   ASSERT_TRUE(error.has_value());
   EXPECT_EQ(error->message(),
             "xDS channel for server default_xds_server: Server down (node "
             "ID:xds_client_test)");
+  TriggerConnectionFailure(fallback_server,
+                           absl::UnavailableError("Another server down"));
+  stream->MaybeSendStatusToClient(
+      absl::UnavailableError("Another server down"));
+
```
markdroth, line 4063: Please remove blank line.

test/core/xds/xds_client_test.cc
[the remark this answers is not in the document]
markdroth: Please fix ordering here to be consistent with the principles I oulined in the previous test.

test/core/xds/xds_client_test.cc
```diff
@@ -3991,48 +4033,125 @@ TEST_F(XdsClientTest, FallbackReportsError) {
...
+}
+
+TEST_F(XdsClientTest, FallbackOnStartup) {
+  FakeXdsBootstrap::FakeXdsServer primary_server;
+  FakeXdsBootstrap::FakeXdsServer fallback_server("fallback_xds_server");
+  // Regular operation
+  InitXdsClient(FakeXdsBootstrap::Builder().SetServers(
+      {primary_server, fallback_server}));
+  // Start a watch for "foo1".
+  auto watcher = StartFooWatch("foo1");
```
markdroth, line 4090: This input will cause the client to send a request on the stream before it sees the connectivity failure.

test/core/xds/xds_client_test.cc
```diff
@@ -3991,48 +4033,125 @@ TEST_F(XdsClientTest, FallbackReportsError) {
...
-               /*resource_names=*/{"foo1", "foo2"});
-  auto stream2 = WaitForAdsStream(fallback_server);
-  TriggerConnectionFailure(fallback_server,
-                           absl::UnavailableError("Another server down"));
-  ASSERT_TRUE(stream2 != nullptr);
-  // XdsClient should have sent a subscription request on the ADS stream.
-  request = WaitForRequest(stream2.get());
+               /*resource_names=*/{"foo1"});
+  // Recover to primary
+  TriggerConnectionFailure(primary_server, absl::OkStatus());
```
markdroth, line 4126: As mentioned above, you should not call this to restore connectivity.  Instead, just send a response on the stream.

markdroth (commented):
This is getting closer!

PLMK if you have any questions.  Thanks!

test/core/xds/xds_client_test.cc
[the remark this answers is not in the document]
eugeneo: Done

eugeneo (commented):
Thanks for the feedback!

src/core/ext/xds/xds_client.cc
```diff
@@ -1753,11 +1755,12 @@ void XdsClient::WatchResource(const XdsResourceType* type,
     // If this is the first watcher for this authority, add channels.
     // Note that a channel might have already triggered fallback
     // due to being used in a different authority.
-    if (authority_state.xds_channels.empty() ||
-        !authority_state.xds_channels.back()->status().ok()) {
-      for (const auto& server : xds_servers) {
+    if (fetch_needed && (authority_state.xds_channels.empty() ||
+                         !authority_state.xds_channels.back()->status().ok())) {
+      for (size_t i = authority_state.xds_channels.size();
+           i < xds_servers.size(); i++) {
```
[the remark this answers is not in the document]
eugeneo: Done.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  auto watcher = StartFooWatch("foo1");
+  EXPECT_THAT(GetServerConnections(), ::testing::ElementsAre(::testing::Pair(
+                                          kDefaultXdsServerUrl, true)));
+  // Result (local): Resource is delivered to watcher.
+  auto stream = WaitForAdsStream();
+  ASSERT_TRUE(stream != nullptr);
+  // XdsClient should have sent a subscription request on the ADS stream.
+  auto request = WaitForRequest(stream.get());
+  ASSERT_TRUE(request.has_value());
+  // Result (local): Metrics show 1 resource update and 1 cached resource.
```
markdroth, line 4053: That's not what the metrics below show.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  EXPECT_THAT(GetResourceCounts(),
+              ::testing::ElementsAre(::testing::Pair(
+                  ResourceCountLabelsEq(XdsClient::kOldStyleAuthority,
+                                        XdsFooResourceType::Get()->type_url(),
+                                        "requested"),
+                  1)));
+  TriggerConnectionFailure(primary_server,
+                           absl::UnavailableError("Server down"));
+  stream->MaybeSendStatusToClient(absl::UnavailableError("Server is down"));
```
markdroth, line 4062: Don't need to fail the stream in this case: we're acting as if the channel never got connected in the first place, so the original stream will stay queued when the channel reports the connectivity failure.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  ASSERT_NE(stream, nullptr);
+  request = WaitForRequest(stream.get());
+  ASSERT_TRUE(request.has_value());
+  CheckRequest(*request, XdsFooResourceType::Get()->type_url(),
+               /*version_info=*/"", /*response_nonce=*/"",
+               /*error_detail=*/absl::OkStatus(),
+               /*resource_names=*/{"foo1"});
+  TriggerConnectionFailure(fallback_server,
+                           absl::UnavailableError("Another server down"));
+  stream->MaybeSendStatusToClient(absl::UnavailableError("Fallback"));
```
markdroth, line 4074: Same as above: No need to fail the stream.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+          .set_version_info("1")
+          .set_nonce("A")
+          .AddFooResource(XdsFooResource("foo1", 6))
+          .Serialize());
+  // XdsClient should have delivered the response to the watcher.
+  auto resource = watcher->WaitForNextResource();
+  ASSERT_NE(resource, nullptr);
+  EXPECT_EQ(resource->name, "foo1");
+  EXPECT_EQ(resource->value, 6);
+  request = WaitForRequest(fallback_stream.get());
```
markdroth, line 4136: Please add a comment:

// Client sends an ACK.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  ASSERT_TRUE(request.has_value());
+  CheckRequest(*request, XdsFooResourceType::Get()->type_url(),
+               /*version_info=*/"1", /*response_nonce=*/"A",
+               /*error_detail=*/absl::OkStatus(),
+               /*resource_names=*/{"foo1"});
+  // Recover to primary
+  primary_stream = WaitForAdsStream(primary_server);
```
markdroth, line 4143: This isn't necessary -- we already got the stream above.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  primary_stream->SendMessageToClient(
+      ResponseBuilder(XdsFooResourceType::Get()->type_url())
+          .set_version_info("5")
+          .set_nonce("D")
+          .AddFooResource(XdsFooResource("foo1", 42))
+          .Serialize());
+  ASSERT_NE(primary_stream, nullptr);
```
markdroth, line 4150: This isn't necessary.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  fallback_stream.reset();
```
markdroth, line 4151: Instead of this, please say:

```
// Fallback stream will be closed.
EXPECT_TRUE(fallback_stream->Orphaned());
```

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  auto request = WaitForRequest(stream.get());
+  ASSERT_TRUE(request.has_value());
+  // Result (local): Metrics show 1 resource update and 1 cached resource.
+  EXPECT_THAT(GetResourceCounts(),
+              ::testing::ElementsAre(::testing::Pair(
+                  ResourceCountLabelsEq(XdsClient::kOldStyleAuthority,
+                                        XdsFooResourceType::Get()->type_url(),
+                                        "requested"),
+                  1)));
+  TriggerConnectionFailure(primary_server,
```
markdroth, line 4060: After triggering the connection failure, please check that `GetServerConnections()` shows a failed connection to the primary server and a working connection to the fallback server.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  // Fallback happens now
+  stream = WaitForAdsStream(fallback_server);
+  ASSERT_NE(stream, nullptr);
+  request = WaitForRequest(stream.get());
+  ASSERT_TRUE(request.has_value());
+  CheckRequest(*request, XdsFooResourceType::Get()->type_url(),
+               /*version_info=*/"", /*response_nonce=*/"",
+               /*error_detail=*/absl::OkStatus(),
+               /*resource_names=*/{"foo1"});
+  TriggerConnectionFailure(fallback_server,
```
markdroth, line 4072: After triggering the connection failure, please check that `GetServerConnections()` shows failed connections to both servers.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+               /*error_detail=*/absl::OkStatus(),
+               /*resource_names=*/{"foo1"});
+  // Send a response.
+  fallback_stream->SendMessageToClient(
+      ResponseBuilder(XdsFooResourceType::Get()->type_url())
+          .set_version_info("1")
+          .set_nonce("A")
+          .AddFooResource(XdsFooResource("foo1", 6))
+          .Serialize());
+  // XdsClient should have delivered the response to the watcher.
```
markdroth, line 4131: Please add a check here that `GetServerConnections()` shows a healthy connection to the fallback server.

test/core/xds/xds_client_test.cc
```diff
@@ -3779,6 +3803,367 @@ TEST_F(XdsClientTest, AdsReadWaitsForHandleRelease) {
...
+  primary_stream = WaitForAdsStream(primary_server);
+  primary_stream->SendMessageToClient(
+      ResponseBuilder(XdsFooResourceType::Get()->type_url())
+          .set_version_info("5")
+          .set_nonce("D")
+          .AddFooResource(XdsFooResource("foo1", 42))
+          .Serialize());
+  ASSERT_NE(primary_stream, nullptr);
+  fallback_stream.reset();
+  resource = watcher->WaitForNextResource();
```
markdroth, line 4152: Please add a check here that `GetServerConnections()` shows a healthy channel to the primary server and no channel to the fallback server.

src/core/ext/xds/xds_client.cc
```diff
@@ -1749,7 +1751,8 @@ void XdsClient::WatchResource(const XdsResourceType* type,
     // If this is the first watcher for this authority, add channels.
     // Note that a channel might have already triggered fallback
     // due to being used in a different authority.
-    if (authority_state.xds_channels.empty()) {
+    if (authority_state.xds_channels.empty() ||
+        !authority_state.xds_channels.back()->status().ok()) {
```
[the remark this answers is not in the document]
markdroth: Doesn't look like you addressed this comment.

markdroth (commented):
The tests are better but still need some work.  And there was one comment from my previous review pass that you didn't address.

PLMK if you have any questions.  Thanks!

test/core/xds/xds_client_test.cc
```diff
@@ -3834,46 +3834,74 @@ TEST_F(XdsClientTest, FallbackAndRecover) {
   // Send a response.
   stream->SendMessageToClient(
       ResponseBuilder(XdsFooResourceType::Get()->type_url())
-          .set_version_info("")
-          .set_nonce("")
+          .set_version_info("20")
+          .set_nonce("O")
           .AddFooResource(XdsFooResource("foo1", 6))
           .Serialize());
+  // Input: Get initial response from primary server.
```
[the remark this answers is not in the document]
eugeneo: Done

src/core/ext/xds/xds_client.cc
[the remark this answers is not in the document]
eugeneo: (Just noticed this comment - addressing)

test/core/xds/xds_client_test.cc
```diff
@@ -4128,28 +4113,33 @@ TEST_F(XdsClientTest, FallbackOnStartup) {
...
       ResponseBuilder(XdsFooResourceType::Get()->type_url())
           .set_version_info("5")
           .set_nonce("D")
           .AddFooResource(XdsFooResource("foo1", 42))
           .Serialize());
-  ASSERT_NE(primary_stream, nullptr);
-  fallback_stream.reset();
+  EXPECT_TRUE(fallback_stream->Orphaned());
   resource = watcher->WaitForNextResource();
+  EXPECT_THAT(GetServerConnections(), ::testing::ElementsAre(::testing::Pair(
```
markdroth, line 4141: Please move this down to line 4146, so that it's not right in between getting the resource from the watcher and checking the resource's value.

markdroth (approved):
Just one remaining comment.  Feel free to merge after addressing.

test/core/xds/xds_client_test.cc
eugeneo: Done. Thanks!

PR #35272, opened 2023-12-11
Replace bind() targets with alias() ones

mering:
Usage of `bind()` is [deprecated in WORKSPACE](https://bazel.build/reference/be/workspace#bind) and [not supported in Bzlmod](https://bazel.build/external/migration#bind-targets).

This is a required preparatory step for #35271.

An alternative would be to use the target names directly and completely get rid of `bind()` and `alias()`.

Fixes #13902, #20193.

PR #36157, opened 2024-03-21
[Bazel] fix: remove use of native.bind

jbedard:
`native.bind()` is not supported with bzlmod: https://bazel.build/external/migration#bind-targets

This is required when `grpc` is used in a project that does have bzlmod enabled and will be required if/when the grpc repo converts to bzlmod.


XuanWang-Amos:
Looks like we have a conflict, can you please rebase? We'll trigger the tests after that.

jbedard:
Updated

Wyverald:
Looks like most (if not all) of this PR is already checked in as part of https://github.com/grpc/grpc/pull/35272.

@jbedard, could you confirm? Maybe this can be closed now.

jbedard:
I haven't looked at this in a while. If you think this is no longer required we can just close it.

PR #36165, opened 2024-03-22
[build] fix compile faild in ppc64le platform

qiangzhao:
Fix data type, and function call, it's improper in VSX (Vector Scalar Extensions) instruction set in ppc64le to avoid build failure.


qiangzhao:
Hi @veblush :
Could you please take a look of this request if when you have time. I have test it can build pass in PPC64.
Thank you!

veblush:
Thanks for the PR but we don't modify this module as it's a copy of https://github.com/Cyan4973/xxHash. Would you try using the latest version of `xxhash.h` there instead?

PR #25586, opened 2021-03-01
Eliminate gRPC insecure build

yihuazhang:
This PR is a part of the effort for removing gRPC insecure build so that all gRPC targets can depend on secure builds. Another purpose of this PR is to refactor the BUILD file so that for the targets that do not rely on SSL, they will not link `libssl` targets. 

After this PR, there will be the following continuous efforts. 

- Refactor gRPC C++ so that each C++ credential target will have its own build target.  
- Clean up BUILD file by removing unnecessary targets.

---
This change is [Reviewable](https://reviewable.io/reviews/grpc/grpc/25586)

Issue #21589, opened 2020-01-06
Add AuthMetadataProcessor support to InsecureServerCredentials


BurningEnlightenment:
### Is your feature request related to a problem? Please describe.
I host multiple grpc services and http apps on a single machine behind an NGINX reverse proxy [[1]](https://www.nginx.com/blog/nginx-1-13-10-grpc/) which forwards incoming requests to the correct service based on TLS SNI. I.e. the client connects via TLS to the publicly exposed NGINX server.

I also rely on a third party authentication solution which I integrate in grpc via the `AuthMetadataProcessor`. However, I can only use the custom `AuthMetadataProcessor` with `SslServerCredentials` which implies that I'm currently forced to use TLS for the NGINX <=> grpc service connection.

It is pretty annoying to provision the grpc services with self signed certificates for the loopback address. Furthermore it is an outright waste of server resources to encrypt connections on the loopback adapter.

### Describe the solution you'd like
```cpp
std::shared_ptr<AuthMetadataProcessor> authProcessor  = /*...*/;
auto serverCredentials = grpc::InsecureServerCredentials();
serverCredentials->SetAuthMetadataProcessor(std::move(authPocessor));
```
To work the same way as it would with `SslServerCredentials()`.

### Describe alternatives you've considered
I wouldn't mind if another `ServerCredentials` implementation (maybe `MetadataServerCredentials`) would be provided in order to keep the completely unauthenticated case separated.

### Additional context

[1] https://www.nginx.com/blog/nginx-1-13-10-grpc/
[2] related #17942
[3] similar use case https://groups.google.com/forum/#!topic/grpc-io/ifL63H0kN48

veblush:
Jiangtao, can you take a look at this?

sazconaitbit:
im looking for the same, any news about this? thanks

veblush:
@yihuazhang Could you take a look at this?

toddtomkinson:
The current state of #25586 will NOT allow this, as the SetAuthMetadataProcessor in InsecureServerCredentials still does not allow an AuthMetadataProcessor to be set:

```
  void SetAuthMetadataProcessor(
      const std::shared_ptr<grpc::AuthMetadataProcessor>& processor) override {
    (void)processor;
    GPR_ASSERT(0);  // Should not be called on InsecureServerCredentials.
  }
```

It appears that the default implementation will only be used by classes that extend SecureServerCredentials.

markdroth:
[...]

PR #36176, opened 2024-03-26
[security] Refactor credentials types to remove special handling for insecure creds

drfloob:
Forked from #35957

This PR refactors the credentials types to remove Secure and Insecure Channel and Call credentials types. We standardize on a `c_creds()` accessor method for all credentials types, which can now be treated uniformly. This notably removes special-case handling of insecure credentials.

The special code-paths for insecure creds are no longer necessary in the wake of #25586.


include/grpcpp/security/server_credentials.h
```diff
@@ -89,16 +87,7 @@ class ServerCredentials : private grpc::internal::GrpcLibrary {
   // TODO(dgq): the "port" part seems to be a misnomer.
   virtual int AddPortToServer(const std::string& addr, grpc_server* server) = 0;
```
markdroth, line 88: Similar to my comments elsewhere about removing the wrapping subclasses on the client side, I think this base class should have a data member for the C-core creds and should have default implementations for these methods, so that subclasses need to override the methods only if they want non-default behavior, which should be a subset of cases (and most of that subset should go away once we implement transport selection in C-core).

src/cpp/server/insecure_server_credentials.cc
```diff
@@ -29,27 +29,36 @@ namespace grpc {
...
+  InsecureServerCredentialsImpl()
+      : creds_(grpc_insecure_server_credentials_create()) {}
+
+  ~InsecureServerCredentialsImpl() override {
+    grpc_server_credentials_release(creds_);
+  }
+
   int AddPortToServer(const std::string& addr, grpc_server* server) override {
     grpc_server_credentials* server_creds =
         grpc_insecure_server_credentials_create();
```
markdroth, line 41: This doesn't need to create a new creds object anymore; it can use `creds_` instead.

src/cpp/server/secure_server_credentials.h
```diff
@@ -72,10 +72,8 @@ class SecureServerCredentials final : public ServerCredentials {
   void SetAuthMetadataProcessor(
       const std::shared_ptr<grpc::AuthMetadataProcessor>& processor) override;
 
-  grpc_server_credentials* c_creds() { return creds_; }
-
  private:
-  SecureServerCredentials* AsSecureServerCredentials() override { return this; }
+  grpc_server_credentials* c_creds() const override { return creds_; }
```
markdroth, line 76: I think the `SecureServerCredentials` class can go away.

test/cpp/end2end/rls_end2end_test.cc
```diff
@@ -198,13 +199,13 @@ class RlsEnd2endTest : public ::testing::Test {
...
-        grpc_composite_channel_credentials_create(channel_creds, call_creds,
-                                                  nullptr));
+    auto creds =
+        WrapChannelCredentials(grpc_composite_channel_credentials_create(
+            channel_creds, call_creds, nullptr));
     call_creds->Unref();
     channel_creds->Unref();
     target_uri_ = absl::StrCat("fake:///", kServerName);
-    channel_ = grpc::CreateCustomChannel(target_uri_, std::move(creds), args);
+    channel_ = grpc::CreateCustomChannel(target_uri_, creds, args);
```
markdroth, line 208: Why remove `std::move()` here?

markdroth (commented):
Thanks for cleaning all of this up!

I do think the PR title should be about removing special code-paths for insecure creds.  At its heart, that's the main thing we're doing in this PR; all refactoring is just a consequence of that.

src/cpp/server/insecure_server_credentials.cc
drfloob: Done

src/cpp/client/secure_credentials.cc
```diff
@@ -361,26 +389,19 @@ std::shared_ptr<ChannelCredentials> CompositeChannelCredentials(
...
   return nullptr;
 }
 
 std::shared_ptr<CallCredentials> CompositeCallCredentials(
     const std::shared_ptr<CallCredentials>& creds1,
     const std::shared_ptr<CallCredentials>& creds2) {
-  SecureCallCredentials* s_creds1 = creds1->AsSecureCredentials();
-  SecureCallCredentials* s_creds2 = creds2->AsSecureCredentials();
-  if (s_creds1 != nullptr && s_creds2 != nullptr) {
+  if (creds1->c_creds() != nullptr && creds2->c_creds() != nullptr) {
```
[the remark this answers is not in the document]
drfloob: Done.

src/cpp/client/wrapped_credentials.cc
```diff
@@ -0,0 +1,115 @@
...
+        grpc_channel_create(target.c_str(), c_creds_, &channel_args),
+        std::move(interceptor_creators));
+  }
+
+  grpc_channel_credentials* const c_creds_ = nullptr;
+};
+
+// ---- WrappedCallCredentials ----
+
+class WrappedCallCredentials final : public CallCredentials {
```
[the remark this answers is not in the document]
drfloob: The class is removed, and replaced with a factory that friends CallCredentials.

include/grpcpp/security/server_credentials.h
drfloob: Done. The base class is now abstract with some state mirroring the old insecure credentials, and subclasses implement a subset of functionality.

src/cpp/server/secure_server_credentials.h
drfloob: As discussed, we were able to reduce the footprint, but making it go away was not necessarily the right thing to do just yet.

src/cpp/client/secure_credentials.cc
```diff
@@ -361,26 +389,19 @@ std::shared_ptr<ChannelCredentials> CompositeChannelCredentials(
...
   // incremented.
-  SecureChannelCredentials* s_channel_creds =
-      channel_creds->AsSecureCredentials();
-  SecureCallCredentials* s_call_creds = call_creds->AsSecureCredentials();
-  if (s_channel_creds && s_call_creds) {
-    return internal::WrapChannelCredentials(
-        grpc_composite_channel_credentials_create(
-            s_channel_creds->GetRawCreds(), s_call_creds->GetRawCreds(),
-            nullptr));
+  if (channel_creds->c_creds() != nullptr && call_creds->c_creds() != nullptr) {
```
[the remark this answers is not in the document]
drfloob: I removed the check against call_creds

src/cpp/server/secure_server_credentials.h
```diff
@@ -61,21 +55,12 @@ class AuthMetadataProcessorAsyncWrapper final {
 
 class SecureServerCredentials final : public ServerCredentials {
```
markdroth, line 56: Please add a TODO here to remove this class once we either (a) allow `AuthMetadataProcessor` to be used with any creds type as requested in #21589 or (b) find a way to remove `AuthMetadataProcessor` in favor of some new server-side interception API.

include/grpcpp/security/credentials.h
```diff
@@ -122,22 +109,29 @@ class ChannelCredentials : private grpc::internal::GrpcLibrary {
...
 class CallCredentials : private grpc::internal::GrpcLibrary {
  public:
+  ~CallCredentials() override;
+
   /// Apply this instance's credentials to \a call.
-  virtual bool ApplyToCall(grpc_call* call) = 0;
-  virtual grpc::string DebugString() {
-    return "CallCredentials did not provide a debug string";
-  }
+  virtual bool ApplyToCall(grpc_call* call);
```
markdroth, line 115: Do either  of these methods actually get overridden anywhere?  If not, maybe they don't need to be virtual.

This may technically be an API change, but in practice I'm not sure there's any way that anyone could have actually subclassed this in a useful way without using internal APIs.

include/grpcpp/security/credentials.h
```diff
@@ -122,22 +109,29 @@ class ChannelCredentials : private grpc::internal::GrpcLibrary {
...
  protected:
   friend std::shared_ptr<ChannelCredentials> CompositeChannelCredentials(
       const std::shared_ptr<ChannelCredentials>& channel_creds,
       const std::shared_ptr<CallCredentials>& call_creds);
-
   friend std::shared_ptr<CallCredentials> CompositeCallCredentials(
       const std::shared_ptr<CallCredentials>& creds1,
       const std::shared_ptr<CallCredentials>& creds2);
+  friend std::string grpc::testing::GetOauth2AccessToken();
+  friend std::shared_ptr<CallCredentials> MakeCallCredentials(
```
markdroth, line 127: Could avoid this by using a subclass pattern, similar to what I suggested elsewhere for `InsecureServerCredentials`.

Could also remove the friend statement for `CompositeCallCredentials()` with the same pattern.

include/grpcpp/security/server_credentials.h
```diff
@@ -68,37 +68,33 @@ std::shared_ptr<ServerCredentials> XdsServerCredentials(
...
-  virtual bool IsInsecure() const { return false; }
-
-  // TODO(yashkt): This is a hack that should be removed once we remove insecure
-  // builds and the indirect method of adding ports to a server.
-  virtual SecureServerCredentials* AsSecureServerCredentials() {
-    return nullptr;
-  }
+  virtual int AddPortToServer(const std::string& addr, grpc_server* server);
+
+  grpc_server_credentials* c_creds_ = nullptr;
```
markdroth, line 97: Don't need `= nullptr`, since it always gets set by the ctor.

src/cpp/client/wrapped_credentials.cc
```diff
@@ -0,0 +1,115 @@
...
+#include <grpc/grpc_security.h>
+#include <grpcpp/security/credentials.h>
+#include <grpcpp/support/channel_arguments.h>
+
+#include "src/core/lib/iomgr/exec_ctx.h"
+#include "src/core/lib/security/credentials/credentials.h"
+
+namespace grpc {
+
+class WrappedChannelCredentials final : public ChannelCredentials {
```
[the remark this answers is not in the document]
markdroth: I don't understand the conflict here.  First, I didn't actually propose removing the check for `c_creds()` being null for channel creds, only for call creds.  And second, even if I had, I don't see why that would prevent us from removing this wrapper class.

I think this should be structured the same way as the server creds, except that we don't need the equivalent of the `SecureServerCredentials` class, because there is no `AuthMetadataProcessor` issue on the client side.  In other words:
- The base class has the data member, a protected ctor that takes the c_creds arg, and a protected  non-virtual accessor method.
- For creds subclasses that are backed by C-core creds types, they can pass the C-core object to the base class ctor.
- For creds subclasses not backed by C-core creds types, they can pass nullptr to the base class ctor.

markdroth (commented):
This is getting closer!

src/cpp/client/wrapped_credentials.cc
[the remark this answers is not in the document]
drfloob: > I didn't actually propose removing the check for c_creds() being null for channel creds

That's the only point of contention. I made some of your proposed changes. We are left with a very minimal `WrappedChannelCredentials` type that serves to hide the public constructor. It would be identical code for ~10 different credentials types if we wanted to remove the wrapper. We can if you think having different concrete types is meaningful polymorphism.

src/cpp/server/xds_server_credentials.cc
```diff
@@ -23,16 +23,14 @@
...
-
 namespace grpc {
 
 std::shared_ptr<ServerCredentials> XdsServerCredentials(
     const std::shared_ptr<ServerCredentials>& fallback_credentials) {
   GPR_ASSERT(fallback_credentials != nullptr);
   GPR_ASSERT(fallback_credentials->c_creds() != nullptr);
-  return std::make_shared<SecureServerCredentials>(
-      grpc_xds_server_credentials_create(fallback_credentials->c_creds()));
+  return std::shared_ptr<ServerCredentials>(new ServerCredentials(
```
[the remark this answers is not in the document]
drfloob: See my comment above about friends and `fallback_credentials->c_creds()`. You can't access protected members of a base class object unless it's `this` or you're a friend, https://godbolt.org/z/4GW5aWPeP. As written, your proposal needs 2 friends.

BUILD
```diff
@@ -1216,9 +1216,15 @@ grpc_cc_library(
 grpc_cc_library(
```
markdroth, line 1216: Please add a TODO here that we should restructure the `grpc++_unsecure` and `grpc++` build targets in a similar way to how the `grpc_unsecure` and `grpc` targets were restructured in #25586: specifically, `grpc++` should essentially be the same as `grpc++_unsecure` except for additional dependencies to pull in additional credential types.

include/grpcpp/security/server_credentials.h
```diff
@@ -68,17 +68,21 @@ std::shared_ptr<ServerCredentials> XdsServerCredentials(
...
-      const std::shared_ptr<grpc::AuthMetadataProcessor>& processor) = 0;
+      const std::shared_ptr<grpc::AuthMetadataProcessor>& processor);
+
+ protected:
+  explicit ServerCredentials(grpc_server_credentials* creds);
+
+  grpc_server_credentials* c_creds() const { return c_creds_; }
 
  private:
+  // We need these friend declarations for access to c_creds.
```
markdroth, line 84: This comment is misleading: `Server` is a friend for access to `AddPortToServer()`, not for access to `c_creds`.

src/cpp/client/insecure_credentials.cc
```diff
@@ -27,46 +24,30 @@
 #include <grpcpp/support/channel_arguments.h>
 #include <grpcpp/support/client_interceptor.h>
 
-#include "src/cpp/client/create_channel_internal.h"
-
 namespace grpc {
-
 namespace {
-class InsecureChannelCredentialsImpl final : public ChannelCredentials {
+class InsecureChannelCredentialsImpl : public ChannelCredentials {
```
markdroth, line 29: I think this can still be `final`.

src/cpp/client/xds_credentials.cc
```diff
@@ -23,24 +23,21 @@
 #include <grpc/support/log.h>
 #include <grpcpp/security/credentials.h>
 
-#include "src/cpp/client/secure_credentials.h"
-
 namespace grpc {
+class XdsChannelCredentialsImpl : public ChannelCredentials {
```
markdroth, line 27: This can be `final`.

markdroth (approved):
This looks great!

Remaining comments are minor; feel free to merge after addressing.

src/cpp/client/call_credentials.cc
```diff
@@ -24,23 +22,16 @@
 namespace grpc {
 
 CallCredentials::CallCredentials(grpc_call_credentials* c_creds)
-    : c_creds_(c_creds) {
-  GPR_ASSERT(c_creds != nullptr);
```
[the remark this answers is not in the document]
drfloob: Nice, added back in. Looks like we removed the last instance of null call c-creds.

drfloob (commented):
All done. Thanks for the reviews!

src/cpp/client/channel_credentials.cc
```diff
@@ -44,4 +44,29 @@ std::shared_ptr<Channel> ChannelCredentials::CreateChannelWithInterceptors(
       std::move(interceptor_creators));
 }
 
+namespace {
+class InsecureChannelCredentialsImpl final : public ChannelCredentials {
```
markdroth, line 48: I think we still want insecure creds in its own file, so that we can eventually visibility-restrict it internally.

include/grpcpp/security/credentials.h
markdroth: Please do the same for `DebugString()`.

drfloob: Done

src/cpp/client/channel_credentials.cc
drfloob: Alright, I've added it back.

markdroth (approved):

PR #36177, opened 2024-03-26
Remove contradicting comment

georgthegreat:
```
(e.g., restarting a read-modify-write sequence).
```

actually stands for retrying conditional update request, which is covered by `(d)` ... `conflicting read-modify-write on the same resource`


georgthegreat:
> This is an appropriate example. 

I do not understand how read-modify-write can be retried without read executed by the client.
This is clearly a «something else must happen prior to retrying the operation».

I suggest changing this to something more clear, i. e.: transient failure in a subsystem needed to perform the request.

PR #36187, opened 2024-03-27
[build] create xds_end2end_test_lib visibility tag

apolcyn:
As title

### DIFF
--- a/bazel/grpc_build_system.bzl
+++ b/bazel/grpc_build_system.bzl
@@ -27,13 +27,13 @@
 Contains macros used throughout the repo.
 """
 
-load("@build_bazel_rules_apple//apple:ios.bzl", "ios_unit_test")
-load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
-load("@com_google_protobuf//bazel:upb_proto_library.bzl", "upb_proto_library", "upb_proto_reflection_library")
 load("//bazel:cc_grpc_library.bzl", "cc_grpc_library")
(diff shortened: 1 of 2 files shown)


bazel/grpc_build_system.bzl
```diff
@@ -120,6 +120,7 @@ def _update_visibility(visibility):
         "tsi": PRIVATE,
         "xds": PRIVATE,
         "xds_client_core": PRIVATE,
+        "xds_end2end_test_lib": PRIVATE,
```
markdroth, line 123: Please call this "xds_end2end_test_utils", to differentiate it from the "xds_end2end_test_lib" build target, which does not actually use this visibility tag. :)

bazel/grpc_build_system.bzl
```diff
@@ -27,13 +27,13 @@
 Contains macros used throughout the repo.
 """
 
+load("@build_bazel_rules_apple//apple:ios.bzl", "ios_unit_test")
```
markdroth, line 30: Why move these lines?

markdroth (approved):

bazel/grpc_build_system.bzl
apolcyn: These were from the internal formatter. I was unsure at first :) , but I've removed them to not add cruft to this PR's diff

bazel/grpc_build_system.bzl
apolcyn: done

markdroth (approved):

PR #36189, opened 2024-03-28
[OTel C++] Add API to set channel scope filter

yashykt:
Also addressing a TODO from previous PRs where `authority` is not being populated in the `ChannelScope`.


src/core/lib/surface/legacy_channel.cc
```diff
@@ -93,10 +93,8 @@ absl::StatusOr<OrphanablePtr<Channel>> LegacyChannel::Create(
     *(*r)->stats_plugin_group =
         GlobalStatsPluginRegistry::GetStatsPluginsForServer(args);
   } else {
-    // TODO(roth): Figure out how to populate authority here.
-    // Or maybe just don't worry about this if no one needs it until after
-    // the call v3 stack lands.
-    StatsPlugin::ChannelScope scope(target, "");
+    StatsPlugin::ChannelScope scope(
+        target, args.GetOwnedString(GRPC_ARG_DEFAULT_AUTHORITY).value_or(""));
```
markdroth, line 97: This isn't going to work as you expected, because in the majority of cases that channel arg is not actually set until much lower in the stack.  And unfortunately, it's not trivial to fix this in the current channel stack, although it will be easier in the new v3 stack.  That's why I had the TODO here to begin with.

For now, how about we simply remove the authority field from `ChannelScope`, since we don't actually need it right now?

include/grpcpp/ext/otel_plugin.h
```diff
@@ -57,6 +58,27 @@ class OpenTelemetryPluginOption {
 /// grpc.server.call.rcvd_total_compressed_message_size
 class OpenTelemetryPluginBuilder {
  public:
+  /// Configuration (scope) for a specific client channel.
+  class ChannelScope {
```
markdroth, line 62: Instead of duplicating the existing class, let's just use an alias for the underlying type.  I'm thinking something like this:

```
using grpc_core::StatsPlugin::ChannelScope;
```

I realize that this means exposing the stats plugin APIs from C-core, but I think that's the right thing to do here anyway.

src/core/lib/surface/legacy_channel.cc
yashykt: I know that the `client_channel_filter` and the resolvers can override the authority to be used, and if there is no override, we derive the authority from the target.

This is exactly why I'm wording it as "Returns the default authority set for the channel, empty if un-set". I could be specific and say that it only shows the default authority arg added by the application, but I want some wiggle room. Also, a reason why we should probably slap the EXPERIMENTAL label here.

include/grpcpp/ext/otel_plugin.h
yashykt: Exposing the stats plugin APIs from C-Core and directly using it from public APIs can be bad since those internal Core APIs can change without gRFCs.

Additionally, `grpc_core::StatsPlugin::ChannelScope` is defined in a `src/core` header which is not installed, so that won't work.

include/grpcpp/ext/otel_plugin.h
```diff
@@ -57,6 +58,27 @@ class OpenTelemetryPluginOption {
 /// grpc.server.call.rcvd_total_compressed_message_size
```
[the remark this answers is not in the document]
yashykt: yeah, we should. Let me send a follow-up PR

include/grpcpp/ext/otel_plugin.h
yashykt: done as discussed offline

src/core/lib/surface/legacy_channel.cc
yashykt: done as discussed offline

src/core/lib/surface/legacy_channel.cc
```diff
@@ -93,8 +93,11 @@ absl::StatusOr<OrphanablePtr<Channel>> LegacyChannel::Create(
     *(*r)->stats_plugin_group =
         GlobalStatsPluginRegistry::GetStatsPluginsForServer(args);
   } else {
-    StatsPlugin::ChannelScope scope(
-        target, args.GetOwnedString(GRPC_ARG_DEFAULT_AUTHORITY).value_or(""));
+    experimental::StatsPluginChannelScope scope(
+        target, args.GetOwnedString(GRPC_ARG_DEFAULT_AUTHORITY)
+                    .value_or(CoreConfiguration::Get()
```
markdroth, line 98: I've done some digging to remind myself of the flow here (details originally came from https://github.com/grpc/grpc/pull/14615#issuecomment-371273483), and I've convinced myself that this approach makes sense.  There's some more cleanup that will be needed here, but I'll take care of that in the v3 stack.

markdroth (approved):

src/core/lib/surface/legacy_channel.cc
yashykt: Thanks!

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -337,7 +338,8 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
 
   // StatsPlugin:
   std::pair<bool, std::shared_ptr<grpc_core::StatsPlugin::ScopeConfig>>
-  IsEnabledForChannel(const ChannelScope& scope) const override;
+  IsEnabledForChannel(const grpc_core::experimental::StatsPluginChannelScope&
```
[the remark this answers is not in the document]
yashykt: Done

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -276,8 +276,9 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
 
   class ClientScopeConfig : public grpc_core::StatsPlugin::ScopeConfig {
    public:
-    ClientScopeConfig(const OpenTelemetryPlugin* otel_plugin,
-                      const ChannelScope& scope)
+    ClientScopeConfig(
+        const OpenTelemetryPlugin* otel_plugin,
+        const grpc_core::experimental::StatsPluginChannelScope& scope)
```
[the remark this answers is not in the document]
yashykt: Done

PR #36190, opened 2024-03-28
[ObjC] increase Basic Tests C/C++ MacOS jobs

HannahShiSFB:
These tests are slow and can timeout, this diff increase the number of concurrent jobs to 4 and 8, which may help avoid  timeouts.

These parameters are also configured differently in [many other tests](https://github.com/search?q=repo%3Agrpc%2Fgrpc+inner_jobs&type=code).

```
2024-03-10 07:00:56,453 START: run_tests_c_macos_dbg_native
2024-03-10 07:06:44,335 PASSED: run_tests_c_macos_dbg_native [time=347.9sec, retries=0:0]
2024-03-10 07:06:44,335 START: run_tests_c_macos_opt_native
2024-03-10 07:12:54,717 PASSED: run_tests_c_macos_opt_native [time=370.4sec, retries=0:0]
2024-03-10 07:12:54,717 START: run_tests_c++_macos_dbg_native
2024-03-10 09:19:00,785 PASSED: run_tests_c++_macos_dbg_native [time=7566.1sec, retries=0:0]
2024-03-10 09:19:00,788 START: run_tests_c++_macos_opt_native

ERROR: Aborting VM command due to timeout of 14400 seconds
```

### DIFF
--- a/tools/internal_ci/macos/grpc_basictests_c_cpp.cfg
+++ b/tools/internal_ci/macos/grpc_basictests_c_cpp.cfg
@@ -27,5 +27,5 @@ action {
 
 env_vars {
   key: "RUN_TESTS_FLAGS"
-  value: "-f basictests macos corelang --internal_ci -j 1 --inner_jobs 4 --bq_result_table aggregate_results"
+  value: "-f basictests macos corelang --internal_ci -j 4 --inner_jobs 8 --bq_result_table aggregate_results"
 }
--- a/tools/internal_ci/macos/pull_request/grpc_basictests_c_cpp.cfg
+++ b/tools/internal_ci/macos/pull_request/grpc_basictests_c_cpp.cfg
@@ -28,5 +28,5 @@ action {
 env_vars {
   key: "RUN_TESTS_FLAGS"
   # on pull requests, only run the "dbg" configuration due to CI resource constraints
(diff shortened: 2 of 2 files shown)


sampajano (approved):
Awesome!! Thanks Hannah! 😃

By any chance you've tried these yourself? How much faster are the tests now, compared to before? Or if not, do you think they'd be 2-4x faster in theory?

Thanks!

HannahShiSFB:
It's hard to say for inner_jobs, run time can vary a log with the same configuration.
The 4 outer jobs should help run c/c++ tests with dbg/opt mode in parallel, which should avoid the CI timeout.

sampajano:
@HannahShiSFB Aha ok thanks for explaining, Hannah! Looking forward to the test timeout disappearing! 😃

sampajano:
@HannahShiSFB Hi Hannah..

Unfortunately this has not worked.. since tests are still timing out..

Example log:
https://btx.cloud.google.com/invocations/4a523df8-c692-4b62-a0c3-0fc04cad1c24/targets/grpc%2Fcore%2Fmaster%2Fmacos%2Fgrpc_basictests_c_cpp;config=default/log

Do you have other ideas of making the test run faster? Or should we disable part of it somehow?

Thanks!

HannahShiSFB:
Looks like this is how fast a single box can do, we probably need to separate the opt and dbg tests into two jobs so they run in separate VMs. Can we have Jan help on this?

PR #36216, opened 2024-04-02
[build] Restrict visibility for creating core credentials types.

drfloob:


include/grpc/grpc_security.h
```diff
@@ -121,39 +123,6 @@ GRPCAPI void grpc_ssl_session_cache_destroy(grpc_ssl_session_cache* cache);
 GRPCAPI grpc_arg
 grpc_ssl_session_cache_create_channel_arg(grpc_ssl_session_cache* cache);
```
markdroth, line 124: The SSL session cache stuff is also used by SslCreds / TlsCreds, so it should move where they are.

BUILD
```diff
@@ -2312,6 +2313,13 @@ grpc_cc_library(
     ],
 )
 
+grpc_cc_library(
+    name = "grpc_core_credentials",
+    hdrs = ["include/grpc/credentials.h"],
+    language = "c++",
+    visibility = ["@grpc:core_credentials"],
```
markdroth, line 2320: One potential problem here is that wrapped languages need access to these APIs.  Please make sure we're not breaking that -- e.g., I think we support building python via bazel.

BUILD
```diff
@@ -2312,6 +2313,13 @@ grpc_cc_library(
     ],
 )
 
+grpc_cc_library(
+    name = "grpc_core_credentials",
```
markdroth, line 2317: Ultimately, I'd like us to get to a place where each creds type has a separate BUILD target with its own headers and its own implementation in the same target.  But we don't need to do that as part of this PR.

third_party/protobuf
markdroth, line 1: Looks like you may be accidentally changing the protobuf version here.

BUILD
drfloob: Good point. It looks like the python bazel tests are happy, and one of those builds the auth example https://source.cloud.google.com/results/invocations/9b89d452-140c-459f-9c47-1780350c6fc9/targets. 

`//examples/python/auth:_auth_example_test                                PASSED in 0.9s`

drfloob (commented):
Thanks Mark! Changes made.

BUILD
```diff
@@ -2326,6 +2328,14 @@ grpc_cc_library(
     ],
 )
 
+# TODO(hork): split credentials types into their own source files and targets.
+grpc_cc_library(
+    name = "grpc_core_credentials",
```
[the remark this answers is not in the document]
drfloob: Done

tools/distrib/sanitize.sh
```diff
@@ -23,7 +23,6 @@ tools/distrib/check_naked_includes.py --fix || true
 tools/distrib/check_copyright.py --fix
 tools/distrib/add-iwyu.py
 tools/distrib/check_trailing_newlines.sh --fix
-tools/run_tests/sanity/check_port_platform.py --fix
```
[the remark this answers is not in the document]
drfloob: I removed it separately in https://github.com/grpc/grpc/pull/36257. It's gone from this PR now, just had to merge master again.

tools/codegen/core/gen_grpc_tls_credentials_options.py
```diff
@@ -469,6 +470,8 @@ def get_file_differences(file1, file2):
 
 #include <gmock/gmock.h>
 
+#include <grpc/credentials.h>
```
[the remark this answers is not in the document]
drfloob: As discussed, this is adding the header to both the `.h` and the `.cc` files. It's not redundant, in an IWYU-sense.

drfloob (commented):
Thanks Mark!

markdroth (approved):

gnossen (approved):
LGTM to Python changes.

PR #36230, opened 2024-04-02
[Python grpcio_tools] Update version requirement of protobuf in grpcio_tools

XuanWang-Amos:
Since we already finished v26.1 upgrade in https://github.com/grpc/grpc/pull/35796

### DIFF
--- a/tools/distrib/python/grpcio_tools/setup.py
+++ b/tools/distrib/python/grpcio_tools/setup.py
@@ -329,7 +329,7 @@ setuptools.setup(
     packages=setuptools.find_packages("."),
     python_requires=">=3.8",
     install_requires=[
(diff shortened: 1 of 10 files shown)


tools/distrib/python/grpcio_tools/setup.py
```diff
@@ -329,7 +329,7 @@ def extension_modules():
     packages=setuptools.find_packages("."),
     python_requires=">=3.7",
     install_requires=[
-        "protobuf>=4.21.6,<5.0dev",
+        "protobuf>=4.21.6,<6.0dev",
```
gnossen, line 332: Lower range also needs to be updated.

gnossen (approved):

gnossen (approved):

PR #36235, opened 2024-04-02
[Samples] Health checking example

eugeneo:


examples/cpp/health/health_client.cc
```diff
@@ -0,0 +1,140 @@
...
+    // Context for the client. It could be used to convey extra information to
+    // the server and/or tweak certain RPC behaviors.
+    ClientContext context;
+
+    // The actual RPC.
+    std::mutex mu;
+    std::condition_variable cv;
+    bool done = false;
+    Status status;
+    stub_->async()->SayHello(&context, &request, &reply,
```
yijiem, line 76: Sorry for the noob question, not familiar with how to use gRPC. Is there a synchronous API we can use here so that we don't need these?

yijiem (approved):
Thanks Eugene!

eugeneo (commented):
Thanks for the review!

PR #36254, opened 2024-04-04
[OTel C++] Add experimental optional locality label available to client per-attempt metrics

yashykt:
As per https://github.com/grpc/proposal/pull/419, the experimental optional label `grpc.lb.locality` is added to the follow per-call metrics - 
* grpc.client.attempt.duration
* grpc.client.attempt.sent_total_compressed_message_size
* grpc.client.attempt.rcvd_total_compressed_message_size


src/cpp/ext/otel/otel_plugin.h
```diff
@@ -398,6 +400,8 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
   // Instruments for per-call metrics.
   ClientMetrics client_;
   ServerMetrics server_;
+  std::vector<grpc_core::ClientCallTracer::CallAttemptTracer::OptionalLabelKey>
```
markdroth, line 403: I think this can be `std::bitset<OptionalLabelKey::kSize>`, using the same approach that we use for optional labels for the NPC metrics.

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -100,6 +103,22 @@ class OpenTelemetryPlugin::KeyValueIterable
         return false;
       }
     }
+    // Add per-call optional labels
+    if (record_optional_labels_) {
+      for (const auto& optional_label_key :
+           otel_plugin_->per_call_optional_label_keys_) {
+        auto it = optional_labels_.find(optional_label_key);
```
markdroth, line 110: Don't we need to check if the optional label is actually enabled here?  If it's enabled, then we need to add it here, regardless of whether the call tracer actually had a value for it; if it's not enabled, then we need to _not_ add it, even if the call tracer _does_ have a value for it.

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -132,8 +151,10 @@ class OpenTelemetryPlugin::KeyValueIterable
       additional_labels_;
   const OpenTelemetryPlugin::ActivePluginOptionsView*
       active_plugin_options_view_;
-  absl::Span<const std::shared_ptr<std::map<std::string, std::string>>>
+  std::map<grpc_core::ClientCallTracer::CallAttemptTracer::OptionalLabelKey,
```
markdroth, line 154: We definitely do not want to make a copy of this map -- that's needless allocation.  I think this should go back to using `absl::Span<>`.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -757,6 +757,142 @@ TEST_F(OpenTelemetryPluginEnd2EndTest,
...
+      data["grpc.server.call.duration"][0].attributes.GetAttributes();
+  EXPECT_THAT(
+      server_duration_attributes,
+      ::testing::Not(::testing::Contains(::testing::Key("grpc.lb.locality"))));
+}
+
+// Tests that when locality label is enabled on the plugin but not provided by
+// gRPC, an empty value is recorded.
+TEST_F(OpenTelemetryPluginEnd2EndTest,
+       OptionalPerCallLocalityLabelWhenNotAvailable) {
```
markdroth, line 821: Please also add a test for the case where the label is added to the call tracer but not enabled via the OT plugin.

markdroth (commented):
This looks like the right approach!  There are a few noteworthy comments, though,

PLMK if you have any questions.  Thanks!

src/cpp/ext/otel/key_value_iterable.h
yashykt: In the previous logic `per_call_optional_label_keys_` had the list of enabled optional labels, and we were iterating over those to make sure we were adding a value for all of them.

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: added

src/cpp/ext/otel/key_value_iterable.h
yashykt: uh yeah, definitely don't want to make a copy

src/cpp/ext/otel/key_value_iterable.h
yashykt: In the previous logic, we only iterate over the set of enabled labels, so it works as expected. (Tested behavior.)

Logic is now updated to std::bitset.

src/cpp/ext/otel/otel_plugin.h
yashykt: I agree. I was trying to avoid conversions between the enum and size_t, but that led to use of a map over an std::array. The benefits of an std::array where we avoid per-call allocations and get contiguous memory access vastly outweigh the conversion downside.

src/cpp/ext/otel/otel_client_call_tracer.h
```diff
@@ -92,20 +92,19 @@ class OpenTelemetryPlugin::ClientCallTracer
...
     // Start time (for measuring latency).
     absl::Time start_time_;
     std::unique_ptr<LabelsIterable> injected_labels_;
-    // The indices of the array correspond to the OptionalLabelComponent enum.
-    std::array<std::shared_ptr<std::map<std::string, std::string>>,
-               static_cast<size_t>(OptionalLabelComponent::kSize)>
-        optional_labels_array_;
+    // Preferring std::map<> over std::vector<std::pair<>> to account for
+    // possible duplicates additions.
+    std::map<OptionalLabelKey, grpc_core::RefCountedStringValue>
```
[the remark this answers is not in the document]
yashykt: Agreed. I was trying to avoid conversions between the enum and size_t, but that led to use of a map over an std::array. The benefits of an std::array where we avoid per-call allocations and get contiguous memory access vastly outweigh the conversion downside.

src/core/load_balancing/xds/xds_cluster_impl.cc
```diff
@@ -197,7 +195,7 @@ class XdsClusterImplLb final : public LoadBalancingPolicy {
     // need to store the locality labels directly only in the case where
```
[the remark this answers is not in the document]
yashykt: done

src/core/lib/channel/call_tracer.h
```diff
@@ -146,9 +163,8 @@ class ClientCallTracer : public CallTracerAnnotationInterface {
     virtual void RecordEnd(const gpr_timespec& latency) = 0;
 
     // Adds optional labels to be reported by the underlying tracer in a call.
```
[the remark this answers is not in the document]
yashykt: Done

src/core/lib/channel/call_tracer.h
```diff
@@ -146,9 +163,8 @@ class ClientCallTracer : public CallTracerAnnotationInterface {
...
-    virtual void AddOptionalLabels(
-        OptionalLabelComponent component,
-        std::shared_ptr<std::map<std::string, std::string>> labels) = 0;
+    virtual void AddOptionalLabel(OptionalLabelKey key,
```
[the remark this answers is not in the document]
yashykt: Done

src/core/ext/xds/xds_client_stats.h
```diff
@@ -89,9 +87,7 @@ class XdsLocalityName final : public RefCounted<XdsLocalityName> {
   const std::string& region() const { return region_; }
   const std::string& zone() const { return zone_; }
   const std::string& sub_zone() const { return sub_zone_; }
-  std::shared_ptr<std::map<std::string, std::string>> locality_labels() const {
-    return locality_labels_;
-  }
+  RefCountedStringValue locality() const { return locality_; }
```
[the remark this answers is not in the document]
yashykt: That means we'll create a new RefCountedStringValue each time we invoke `AsHumanReadableString()` which is unnecessary but it probably also doesn't matter much since this is on the control plane

src/core/ext/xds/xds_cluster.h
```diff
@@ -101,7 +101,9 @@ struct XdsClusterResource : public XdsResourceType::ResourceData {
 
   XdsHealthStatusSet override_host_statuses;
 
-  std::shared_ptr<std::map<std::string, std::string>> telemetry_labels;
+  std::vector<std::pair<ClientCallTracer::CallAttemptTracer::OptionalLabelKey,
```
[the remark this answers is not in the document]
yijiem: Can this be a `std::array<RefCountedStringValue, ClientCallTracer::CallAttemptTracer::OptionalLabelKey::kSize>` since we know the upper-bound of the size in compile-time?

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -336,6 +335,17 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
   };
 
   // StatsPlugin:
+
+  // Returns the string form of \a key
+  static absl::string_view OptionalLabelKeyToString(
```
markdroth, line 340: Do we need this method?  I don't see that we're actually using it anywhere.

test/core/xds/xds_cluster_resource_type_test.cc
```diff
@@ -1653,7 +1657,7 @@ TEST_F(TelemetryLabelTest, MissingMetadataField) {
   ASSERT_TRUE(decode_result.resource.ok()) << decode_result.resource.status();
   auto& resource =
       static_cast<const XdsClusterResource&>(**decode_result.resource);
-  EXPECT_EQ(resource.telemetry_labels, nullptr);
+  EXPECT_TRUE(resource.telemetry_labels.empty());
```
markdroth, line 1660: Prefer to write this as `EXPECT_THAT(resource.telemetry_labels, ::testing::ElementsAre())`, since that will yield more useful output if the test fails.

test/core/xds/xds_cluster_resource_type_test.cc
```diff
@@ -1671,10 +1675,10 @@ TEST_F(TelemetryLabelTest, MissingCsmFilterMetadataField) {
...
```
markdroth, line 1678: Same here.

markdroth (commented):
Only one really significant comment left, which is the one in xds_cluster_stats.h.

PLMK if you have any questions.  Thanks!

src/core/ext/xds/xds_cluster.h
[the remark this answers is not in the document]
yashykt: The main benefit for std::vector is the contiguous storage of elements so that when we iterate over elements for a call, it'll be faster compared to an std::map.

But, as you said, for a small number of elements, the performance is not gonna matter much, so if you feel strongly about this, I'll revert to an `std::map`.

src/cpp/ext/otel/otel_plugin.h
yashykt: I use it in `key_value_iterable.h`

test/core/xds/xds_cluster_resource_type_test.cc
yashykt: done

test/core/xds/xds_cluster_resource_type_test.cc
yashykt: done

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -398,14 +408,15 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
   // Instruments for per-call metrics.
   ClientMetrics client_;
   ServerMetrics server_;
+  static constexpr int kOptionalLabelsSizeLimit = 64;
+  std::bitset<kOptionalLabelsSizeLimit> per_call_optional_label_bits_;
```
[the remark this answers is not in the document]
yashykt: done

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -100,6 +101,26 @@ class OpenTelemetryPlugin::KeyValueIterable
         return false;
       }
     }
+    // Add per-call optional labels
+    if (record_optional_labels_) {
+      for (size_t i = 0;
+           i <
+           static_cast<size_t>(grpc_core::ClientCallTracer::CallAttemptTracer::
```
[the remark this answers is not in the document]
yashykt: done

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -100,6 +101,26 @@ class OpenTelemetryPlugin::KeyValueIterable
...
+           ++i) {
+        if (!otel_plugin_->per_call_optional_label_bits_.test(i)) {
+          continue;
+        }
+        if (!callback(
+                AbslStrViewToOpenTelemetryStrView(OptionalLabelKeyToString(
+                    static_cast<grpc_core::ClientCallTracer::CallAttemptTracer::
+                                    OptionalLabelKey>(i))),
+                AbslStrViewToOpenTelemetryStrView(
+                    optional_labels_[i].as_string_view()))) {
```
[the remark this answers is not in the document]
yashykt: done

src/cpp/ext/otel/key_value_iterable.h
```diff
@@ -100,6 +101,26 @@ class OpenTelemetryPlugin::KeyValueIterable
         return false;
       }
     }
+    // Add per-call optional labels
+    if (record_optional_labels_) {
```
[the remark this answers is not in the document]
yashykt: done

src/core/resolver/xds/xds_dependency_manager.cc
```diff
@@ -639,7 +639,8 @@ void XdsDependencyManager::OnEndpointUpdate(
     for (const auto& priority : endpoint->priorities) {
       for (const auto& p : priority.localities) {
         if (p.second.endpoints.empty()) {
-          empty_localities.insert(p.first->AsHumanReadableString());
+          empty_localities.insert(
+              std::string(p.first->AsHumanReadableString().as_string_view()));
```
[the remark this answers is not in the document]
yashykt: done

src/core/lib/channel/call_tracer.h
```diff
@@ -162,8 +149,9 @@ class ClientCallTracer : public CallTracerAnnotationInterface {
     // library is free to destroy the object.
     virtual void RecordEnd(const gpr_timespec& latency) = 0;
 
-    // Adds optional labels to be reported by the underlying tracer in a call.
-    virtual void AddOptionalLabel(OptionalLabelKey key,
+    // Sets an optional labels to be reported by the underlying tracer in a
```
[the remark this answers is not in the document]
yashykt: done

src/core/ext/xds/xds_client_stats.h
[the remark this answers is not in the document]
yashykt: Yeah, that makes a lot more sense.

src/core/ext/xds/xds_client_stats.h
```diff
@@ -64,7 +64,10 @@ class XdsLocalityName final : public RefCounted<XdsLocalityName> {
   XdsLocalityName(std::string region, std::string zone, std::string sub_zone)
       : region_(std::move(region)),
         zone_(std::move(zone)),
-        sub_zone_(std::move(sub_zone)) {}
+        sub_zone_(std::move(sub_zone)),
+        human_readable_string_(RefCountedStringValue(
```
markdroth, line 68: I don't think we need to say `RefCountedStringValue()` here.

src/core/load_balancing/xds/cds.cc
```diff
@@ -254,7 +254,7 @@ class PriorityEndpointIterator final : public EndpointAddressesIterator {
         const auto& locality = p.second;
         std::vector<RefCountedStringValue> hierarchical_path = {
             RefCountedStringValue(priority_child_name),
-            RefCountedStringValue(locality_name->AsHumanReadableString())};
+            RefCountedStringValue(locality_name->human_readable_string())};
```
markdroth, line 257: No need to construct a new `RefCountedStringValue` here anymore, since the return value is already that type.

src/core/ext/xds/xds_cluster.h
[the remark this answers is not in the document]
markdroth: I would prefer not having an API that allows misuse like duplicate labels.  An alternative would be to just have two fields here, one for each label:

```
RefCountedStringValue service_label;
RefCountedStringValue namespace_label;
```

This avoids all container overhead, with the trade-off that we spend memory for two pointers even if the labels are not present.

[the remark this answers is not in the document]
yashykt: The size of `std::map<>` is 48 vs `RefCountedStringValue` which is 8

src/core/load_balancing/xds/cds.cc
yashykt: done

src/core/ext/xds/xds_client_stats.h
yashykt: done

src/core/load_balancing/xds/xds_wrr_locality.cc
```diff
@@ -167,15 +167,15 @@ absl::Status XdsWrrLocalityLb::UpdateLocked(UpdateArgs args) {
   }
   auto config = args.config.TakeAsSubclass<XdsWrrLocalityLbConfig>();
   // Scan the addresses to find the weight for each locality.
-  std::map<std::string, uint32_t> locality_weights;
+  std::map<RefCountedStringValue, uint32_t> locality_weights;
```
yijiem, line 170: Does this needs to be `std::map<RefCountedStringValue, uint32_t, RefCountedStringValueLessThan>`?

src/cpp/ext/otel/otel_plugin.h
```diff
@@ -366,6 +365,17 @@ class OpenTelemetryPlugin : public grpc_core::StatsPlugin {
   };
 
   // StatsPlugin:
```
yijiem, line 367: Maybe move this comment to line 379 which marks the start of the StatsPlugin's methods.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -527,14 +535,38 @@ OpenTelemetryPlugin::OpenTelemetryPlugin(
...
+
+absl::string_view OpenTelemetryPlugin::OptionalLabelKeyToString(
+    grpc_core::ClientCallTracer::CallAttemptTracer::OptionalLabelKey key) {
+  switch (key) {
+    case grpc_core::ClientCallTracer::CallAttemptTracer::OptionalLabelKey::
+        kLocality:
+      return kLocality;
+    default:
+      grpc_core::Crash("Illegal OptionalLabelKey index");
+  }
```
yijiem, line 558: Since we only have one valid OptionalLabelKey that the user can set for now, we could also write it like:

```
  GPR_ASSERT(key ==
      grpc_core::ClientCallTracer::CallAttemptTracer::OptionalLabelKey::kLocality);
  return kLocality;
```

yijiem (approved):
Thanks Yash!

src/core/load_balancing/xds/xds_cluster_impl.cc
```diff
@@ -241,9 +241,8 @@ class XdsClusterImplLb final : public LoadBalancingPolicy {
 
     RefCountedPtr<CircuitBreakerCallCounterMap::CallCounter> call_counter_;
     uint32_t max_concurrent_requests_;
-    std::vector<std::pair<ClientCallTracer::CallAttemptTracer::OptionalLabelKey,
-                          RefCountedStringValue>>
-        service_labels_;
+    RefCountedStringValue service_telemetry_label;
```
markdroth, line 244: These two fields need an underscore suffix.

src/cpp/ext/otel/otel_plugin.h
yashykt: I see, I didn't realize what you meant by StatsPlugin's methods

src/cpp/ext/otel/otel_plugin.cc
yashykt: The current structure makes it easier to add new labels, and I think the compiler would be smart enough to optimize it.

src/core/load_balancing/xds/xds_wrr_locality.cc
yashykt: Haven't gotten a complaint from the compiler, maybe because I'm just comparing it with other `RefCountedStringValue`. We can add it if needed.

src/core/load_balancing/xds/xds_cluster_impl.cc
yashykt: Oh wow, thanks!

src/core/ext/xds/xds_cluster.h
```diff
@@ -101,8 +101,8 @@ struct XdsClusterResource : public XdsResourceType::ResourceData {
 
   XdsHealthStatusSet override_host_statuses;
 
-  RefCountedStringValue service_telemetry_label;
-  RefCountedStringValue namespace_telemetry_label;
+  RefCountedStringValue service_telemetry_label_;
```
markdroth, line 104: These fields should *not* have an underscore suffix, since this is a struct, not a class.

src/core/load_balancing/xds/xds_wrr_locality.cc
markdroth: We define [`operator<(RefCountedStringValue, RefCountredStringValue)`](https://github.com/grpc/grpc/blob/ddb785674ca6c21f9aeee4018b9ef4594ed41cc4/src/core/lib/gprpp/ref_counted_string.h#L126), so it shouldn't be necessary to use `RefCountedStringValueLessThan` unless we need heterogeneous lookups.

markdroth (approved):

PR #36259, opened 2024-04-04
[Test] PDB files on diet

veblush:
PDB files are consuming a significant amount of storage space on our Windows test machines. With each PDB file being around 200MB and every target having its own corresponding file, the disk usage adds up quickly.

Fortunately, we can address this issue without sacrificing debugging functionality. [Shrink my Program Database (PDB) file](https://devblogs.microsoft.com/cppblog/shrink-my-program-database-pdb-file/) outlines several techniques to reduce PDB file size. We can implement these methods to achieve our goal. (From my experiment, the compression option only can reduce a 200MB PDB file by 60%)

This change only picks options which don't reduce debuggability.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,6 +309,8 @@ if(MSVC)
   set(_gRPC_C_CXX_FLAGS "${_gRPC_C_CXX_FLAGS} /utf-8")
   # Inconsistent object sizes can cause stack corruption and should be treated as an error
   set(_gRPC_C_CXX_FLAGS "${_gRPC_C_CXX_FLAGS} /we4789")
+  # To decrease the size of PDB files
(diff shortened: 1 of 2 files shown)


drfloob (approved):
Neat! I would be curious if this affects our ability to download EXE + PDB from remote builds to help examine stack traces, but it sounds like it should be fine.

veblush:
> Neat! I would be curious if this affects our ability to download EXE + PDB from remote builds to help examine stack traces, but it sounds like it should be fine.

Yes it shouldn't change.

veblush:
Based on [after-log](https://btx.cloud.google.com/invocations/d7c506f3-0e23-465f-afdb-ed2a3900e8cc/targets) and [before-log](
https://btx.cloud.google.com/invocations/8cd2cc38-42ab-4cd7-bb2c-e3c11230c4d2/targets), this was able to decrease disk usage by ~50GB for a single test suite. (out of 160GB)

PR #36262, opened 2024-04-05
inline bounded external:protocol_compiler

mmorel-35:
Related to #36157

This is only focused on one binding so it is validated progressively


gnossen (approved):
Thanks for the contribution!

veblush (approved):

PR #36270, opened 2024-04-08

tanvi-jagtap:

Will not be replacing CHECK with CHECK_EQ , CHECK_NE etc because there are too many callsites. Only a few - which fit into single - line regex will be changed. This would be small in number just to reduce the load later.

Replacing CHECK with CHECK_EQ , CHECK_NE etc could be done using Cider-V once these changes are submitted if we want to clean up later. Given that we have 5000+ instances of GPR_ASSERT to edit, Doing it manually is too much work for both the author and reviewer.


test/core/transport/chttp2/graceful_shutdown_test.cc
```diff
@@ -431,7 +432,7 @@ TEST_F(GracefulShutdownTest, RequestStartedAfterFinalGoawayIsIgnored) {
   op++;
   error = grpc_call_start_batch(s, ops, static_cast<size_t>(op - ops), Tag(101),
                                 nullptr);
-  GPR_ASSERT(GRPC_CALL_OK == error);
+  CHECK_EQ(error, GRPC_CALL_OK);
```
JuneYan, line 435: Just for my learning, why do we prefer CHECK_EQ(error, GRPC_CALL_OK) instead of CHECK(error == GRPC_CALL_OK) here? Thanks.

JuneYan (approved):

test/core/transport/chttp2/graceful_shutdown_test.cc
JuneYan: Got it. Thanks for explaining.

test/core/transport/chttp2/flow_control_fuzzer.cc
```diff
@@ -193,9 +194,9 @@ void FlowControlFuzzer::Perform(const flow_control_fuzzer::Action& action) {
         send_from_remote.ack_initial_window_size =
             sent_to_remote.initial_window_size;
         for (const auto& id_stream : streams_) {
-          GPR_ASSERT(id_stream.second.window_delta +
-                         *sent_to_remote.initial_window_size <=
-                     (1u << 31) - 1);
+          CHECK(id_stream.second.window_delta +
+                    *sent_to_remote.initial_window_size <=
+                (1u << 31) - 1);
```
JuneYan, line 199: Is it then better to use CHECK_LE here (and similar for other places) for the same reason?

PR #36200, opened 2024-03-30
[systemd] Implement socket activation in event_engine

nipil:
Implemented and tested :
- unix absolute and relative socket path
- unix abstract socket
- ipv4 unicast and wildcard
- ipv6 unicast and wildcard, dual-stack or ipv6-only
- wildcard address expansion
- with epoll1 and poll pollers

Not implemented :
- VSock activation
- Interface scope for ipv6 link-local addresses

Incompatible features :
- wildcard ports

Includes dynamic test suite (_edit : all 23 test fixtures in `test.sh` are passing_)
- of the same kind of fixes as were done in previous PR for unix sockets for iomgr (see PR list below)
- of the full IPv4/IPv6 server handling code, which i did not touch in the PR above, but are fully working here

iomgr PR for unix sockets and test suite : #36035, #36036, #36039, #36041 and #36042

PR #36272, opened 2024-04-08
systemd socket activation: add support for abstract unix domain sockets

andresbeltranmsft:
Currently, gRPC supports systemd socket activation for regular AF_UNIX and AF_INET sockets, but not AF_UNIX sockets in the abstract namespace. Add support for those as well.


andresbeltranmsft:
Taken care of by https://github.com/grpc/grpc/pull/36200

PR #36291, opened 2024-04-08
[XdsClient] add grpc.xds_client.server_failure metric

markdroth:


src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -115,11 +115,20 @@ const auto kMetricResourceUpdatesInvalid =
          kMetricLabelXdsResourceType},
         {}, false);
 
+const auto kMetricServerFailure =
+    GlobalInstrumentsRegistry::RegisterUInt64Counter(
+        "grpc.xds_client.server_failure",
+        "EXPERIMENTAL.  A counter of xDS servers going from healthy to "
```
yashykt, line 121: These description messages show up on the UI. I think we should try to be more concise.

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -115,11 +115,20 @@ const auto kMetricResourceUpdatesInvalid =
...
+        "failure or when the ADS stream fails without seeing a response "
+        "message, as per gRFC A57.",
+        "{failure}", {kMetricLabelTarget, kMetricLabelXdsServer}, {}, false);
+
 const auto kMetricConnected =
     GlobalInstrumentsRegistry::RegisterCallbackInt64Gauge(
         "grpc.xds_client.connected",
         "EXPERIMENTAL.  Whether or not the xDS client currently has a "
-        "working ADS stream to the xDS server. For a given server, this "
+        "working ADS stream to the xDS server.  For a given server, this "
```
yashykt, line 131: probably unintended change

src/core/ext/xds/xds_client_grpc.cc
markdroth: I would rather err on the side of verbosity.  There's nothing worse than trying to bebug a problem and not being entirely clear on what a metric actually indicates when you look at it.

src/core/ext/xds/xds_client_grpc.cc
markdroth: It was intentional.  I'm firmly in the "two spaces after a period" camp.  Must have missed this in my earlier PR.

yijiem (approved):

test/cpp/end2end/xds/xds_core_end2end_test.cc
```diff
@@ -1236,6 +1263,26 @@ TEST_P(XdsMetricsTest, MetricValues) {
...
+  // Shut down balancer and wait for metrics to show the failure.
+  balancer_->Shutdown();
+  for (const char* target : {kTarget.c_str(), "#server"}) {
+    const absl::Time deadline =
+        absl::Now() + absl::Seconds(5 * grpc_test_slowdown_factor());
+    while (true) {
+      auto value = stats_plugin_->GetCounterValue(kMetricServerFailure,
+                                                  {target, kXdsServer}, {});
+      if (value.has_value()) {
+        EXPECT_EQ(1, *value);
```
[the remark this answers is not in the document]
markdroth: The former seems equivalent to what's here.  The latter is unnecessary, since we've already checked that the optional has a value.

yousukseung (approved):
Thanks for adding this counter.

src/core/ext/xds/xds_client_grpc.cc
yashykt: wait, what? I didn't know this was a thing..

yashykt (approved):

PR #36311, opened 2024-04-09
[channel-init] Do not register max-age filter when chaotic-good is enabled.

JuneYan:
follow-up on #36225


ctiller (approved):

ctiller (commented):
actually looking at test results maybe this isn't the right fix... let me thing

JuneYan:
@ctiller Just updated the PR to not register max-age filter with chaotic-good.

ctiller (approved):

PR #36340, opened 2024-04-11
Float CocoaPods abseil dep for patches

paulb777:
Float the CocoaPods abseil dependencies so abseil patches get picked up.

This makes CocoaPods behavior consistent with Swift Package Manager.

Note that 1.20240116.2 is required to support macOS and Catalyst with Xcode 15. 

https://github.com/firebase/firebase-ios-sdk/actions/runs/8650156045/job/23718073321 demonstrates the build failure

### DIFF
--- a/gRPC-C++.podspec
+++ b/gRPC-C++.podspec
@@ -22,7 +22,7 @@
 Pod::Spec.new do |s|
   s.name     = 'gRPC-C++'
   # TODO (mxyan): use version that match gRPC version when pod is stabilized
-  version = '1.62.1'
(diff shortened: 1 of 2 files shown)


sampajano (approved):
@HannahShiSFB FYI

Thanks Paul!

HannahShiSFB:
Thanks @paulb777 for the PR.
Those podspec files are generated, so they cannot be modified directly.

I already have #36356 similar to this, though the CIs seems flaky in the past few days.

yashykt:
I'm guessing this needs to be closed

PR #36342, opened 2024-04-11
[event_engine][fork] Don't delete object

eugeneo:
Fixes use-after-free

### DIFF
--- a/src/core/client_channel/subchannel.cc
+++ b/src/core/client_channel/subchannel.cc
@@ -20,6 +20,7 @@
 
 #include <inttypes.h>
 #include <limits.h>
+#include <unistd.h>
 
 #include <algorithm>
 #include <memory>
@@ -697,8 +698,9 @@ void Subchannel::OnRetryTimer() {
 
 void Subchannel::OnRetryTimerLocked() {
   if (shutdown_) return;
-  gpr_log(GPR_INFO, "subchannel %p %s: backoff delay elapsed, reporting IDLE",
-          this, key_.ToString().c_str());
+  gpr_log(GPR_INFO,
+          "pid %d tid %d subchannel %p: backoff delay elapsed, reporting IDLE",
+          getpid(), gettid(), this);
   SetConnectivityStateLocked(GRPC_CHANNEL_IDLE, absl::OkStatus());
 }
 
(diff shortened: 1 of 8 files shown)


drfloob:
For posterity: need to establish that these won't be leaked under all scenarios.

src/core/client_channel/subchannel_pool_interface.cc
```diff
@@ -55,8 +55,7 @@ std::string SubchannelKey::ToString() const {
   auto addr_uri = grpc_sockaddr_to_uri(&address_);
   return absl::StrCat(
       "{address=",
-      addr_uri.ok() ? addr_uri.value() : addr_uri.status().ToString(),
-      ", args=", args_.ToString(), "}");
+      addr_uri.ok() ? addr_uri.value() : addr_uri.status().ToString(), "}");
```
markdroth, line 58: The subchannel args are part of the key, so they're important to log here.  Please revert this change.

src/core/client_channel/subchannel.cc
```diff
@@ -697,8 +698,9 @@ void Subchannel::OnRetryTimer() {
 
 void Subchannel::OnRetryTimerLocked() {
   if (shutdown_) return;
-  gpr_log(GPR_INFO, "subchannel %p %s: backoff delay elapsed, reporting IDLE",
-          this, key_.ToString().c_str());
+  gpr_log(GPR_INFO,
```
markdroth, line 701: This does not seem like a good change.  The subchannel key is useful to log here, and the pid and tid are not.  Please revert this change.

src/core/client_channel/subchannel.cc
```diff
@@ -743,10 +745,10 @@ void Subchannel::OnConnectingFinishedLocked(grpc_error_handle error) {
     const Duration time_until_next_attempt =
         next_attempt_time_ - Timestamp::Now();
     gpr_log(GPR_INFO,
-            "subchannel %p %s: connect failed (%s), backing off for %" PRId64
-            " ms",
-            this, key_.ToString().c_str(), StatusToString(error).c_str(),
-            time_until_next_attempt.millis());
+            "pid %d subchannel %p %s: connect failed (%s), backing off "
+            "for %" PRId64 " ms",
+            getpid(), this, key_.ToString().c_str(),
```
markdroth, line 750: There is no need to add the pid here.  Please revert.

src/core/client_channel/subchannel.cc
```diff
@@ -765,6 +767,9 @@ void Subchannel::OnConnectingFinishedLocked(grpc_error_handle error) {
             self.reset();
           }
         });
+  } else {
+    gpr_log(GPR_INFO, "Connected subchannel %p %s, pid %d", this,
```
markdroth, line 771: This is going to be far too noisy.  Please remove.

src/core/handshaker/handshaker.cc
```diff
@@ -136,6 +138,8 @@ bool HandshakeManager::CallNextHandshakerLocked(grpc_error_handle error) {
     }
     // Cancel deadline timer, since we're invoking the on_handshake_done
     // callback now.
+    gpr_log(GPR_INFO, "Canceling deadline id: %p, pid: %d",
```
markdroth, line 141: Please revert all of the changes in this file.

markdroth (commented):
I don't think we want any of the changes in the subchannel or handshaker code.  Please revert them.

PR #36357, opened 2024-04-15
Create helper APIs for parsing PEM-encoded credentials.

matthewstevenson88:
In a follow-up PR, we will use these helper APIs to add validation checks when creating TLS credentials objects.


gtcooke94 (commented):
This looks good - it does bother me a little bit from an efficiency perspective that we'll be reading and converting everything, then just dropping it. Do you think there's a good way to get this into the main path so we don't throw away that work?

matthewstevenson88:
> This looks good - it does bother me a little bit from an efficiency perspective that we'll be reading and converting everything, then just dropping it. Do you think there's a good way to get this into the main path so we don't throw away that work?

Great point, thanks for bringing this up. Let me elaborate on how I thought we could use these:

* Step 1: Run these checks when the `TlsCredentials` or `TlsServerCredentials` object is created. This gives us a safety-check (we do not start a client or server with unuseable credentials), at the expense of a small performance hit on start-up. This is, I think, acceptable because (a) the performance hit occurs only during start-up, and (b) the performance hit will be quite small because cert chains are generally quite short (and so there are relatively few PEM certs to parse). Furthermore, I was considering making this "safety check" opt-in rather than by default, at least initially in case there are bugs or edge cases that we're missing.
* Step 2: Change the internal plumbing from the credentials object to SSL transport security to use DER instead of PEM, and populate the DER using these functions. This saves us from doing the PEM-to-DER conversion twice, which should negate (most of) the performance hit from Step 1. I don't, however, want to block Step 1 on this because we have a request for this feature and re-doing all of this plumbing will take awhile.

WDYT?

gtcooke94 (approved):

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -375,4 +380,79 @@ absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl) {
...
+  }
+  uint8_t* cert_data = nullptr;
+  long cert_length = 0;
+  size_t number_of_valid_certs = 0;
+  while (PEM_bytes_read_bio(&cert_data, &cert_length, /*pnm=*/nullptr,
+                            PEM_STRING_X509, in, /*cb=*/nullptr,
+                            /*u=*/nullptr)) {
+    // Try to form the X.509 object from the DER-encoding.
+    const uint8_t* der = cert_data;
+    X509* x509 = d2i_X509(/*out=*/nullptr, &der, cert_length);
```
bob-beck, line 399: So, you're doing all the work to parse this, which then, presumably you in the server setup would hand this object to an SSL_CTX as use certificate to set up your context to start the server. 

It seems like this whole module is a bunch of extra work to do this twice instead of restructuring the code that starts the server up to do this, Fail if it does not work, and if it does, hand the X509 object to the SSL_ctx and start the server.  Why would you want to do this twice?

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -375,4 +380,79 @@ absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl) {
...
+    number_of_valid_certs++;
+  }
+  // We always have errors at this point because in the above loop we read until
+  // we reach the end of |cert_chain_pem|, which generates a "no start line"
+  // error. Therefore, this error is OK if we have successfully parsed some
+  // certificate data previously.
+  const int last_error = ERR_peek_last_error();
+  if (ERR_GET_LIB(last_error) != ERR_LIB_PEM ||
+      ERR_GET_REASON(last_error) != PEM_R_NO_START_LINE) {
+    BIO_free(in);
```
davidben, line 417: This works, but given that you're C++, do you all have smart pointers for OpenSSL types? If you're depending on BoringSSL, you could use `bssl::UniquePtr<BIO>`, but I'm guessing that won't work with OpenSSL.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -375,4 +380,79 @@ absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl) {
...
+  }
+  BIO* in = BIO_new_mem_buf(cert_chain_pem.data(), cert_chain_pem.size());
+  if (in == nullptr) {
+    return false;
+  }
+  uint8_t* cert_data = nullptr;
+  long cert_length = 0;
+  size_t number_of_valid_certs = 0;
+  while (PEM_bytes_read_bio(&cert_data, &cert_length, /*pnm=*/nullptr,
+                            PEM_STRING_X509, in, /*cb=*/nullptr,
```
davidben, line 395: Any reason this isn't just using `PEM_read_bio_X509`?

src/core/tsi/ssl_transport_security_utils.h
```diff
@@ -168,6 +168,13 @@ absl::StatusOr<std::string> AkidFromCertificate(X509* cert);
 // crl.
 // return: a std::string of the DER encoding of the AKID or a status on failure.
 absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl);
+
+// Returns true if cert_chain_pem contains a non-empty sequence of PEM
+// certificate blocks.
+bool IsPemCertificateChainNonEmptyAndValid(absl::string_view cert_chain_pem);
+
+// Returns true if private_key_pem contains a non-empty PEM private key block.
```
davidben, line 176: This comment does not match the implementation. It talks about whether it contains a private key, but you're specifically checking for RSA and EC keys. (Which still may not be usable for TLS, e.g. if they're not the expected key type, or if they don't match the certificate.)

What exactly is this API meant to be used for?

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -375,4 +380,79 @@ absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl) {
...
+  return is_rsa_private_key;
+}
+
+bool IsEcPrivateKey(absl::string_view private_key_pem) {
+  BIO* in = BIO_new_mem_buf(private_key_pem.data(), private_key_pem.size());
+  if (in == nullptr) {
+    return false;
+  }
+  EC_KEY* ec = PEM_read_bio_ECPrivateKey(in, /*x=*/nullptr,
+                                         /*cb=*/nullptr, /*u=*/nullptr);
```
davidben, line 444: This is a pretty silly because these APIs just call `PEM_read_bio_PrivateKey` and then check if the resulting key is of the expected type. So by calling the two together, you're just repeating work over and over.

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -375,4 +380,79 @@ absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl) {
   return ret;
 }
 
+bool IsPemCertificateChainNonEmptyAndValid(absl::string_view cert_chain_pem) {
```
davidben, line 383: This is a bit subtle because of what PEM is. These APIs are named if you're checking whether the input "is" a certificate chain, but non-PEM data and PEM blocks of the wrong type are typically skipped over when parsing. This means that this function returning true just tells you that it *contains* a certificate chain, not that it *is* a certificate chain.

davidben, line 383: It's pretty unclear what "valid" means here. "Valid" means a lot of things in X.509 from:

* The surface-level parse (without reaching into the extensions) succeeds
* Extensions that we know about can be parsed
* The certificate chains up to one of some set of trust anchors
* Variations on the above due to implementation issues (it is _not_ guaranteed that, if `d2i_X509` accepts an input in one implementation, that it will also be accepted in another. There's too much tech debt in the X.509 ecosystem for that)

src/core/tsi/ssl_transport_security_utils.cc
```diff
@@ -375,4 +380,79 @@ absl::StatusOr<std::string> AkidFromCrl(X509_CRL* crl) {
...
+  uint8_t* cert_data = nullptr;
+  long cert_length = 0;
+  size_t number_of_valid_certs = 0;
+  while (PEM_bytes_read_bio(&cert_data, &cert_length, /*pnm=*/nullptr,
+                            PEM_STRING_X509, in, /*cb=*/nullptr,
+                            /*u=*/nullptr)) {
+    // Try to form the X.509 object from the DER-encoding.
+    const uint8_t* der = cert_data;
+    X509* x509 = d2i_X509(/*out=*/nullptr, &der, cert_length);
+    if (x509 == nullptr) {
```
davidben, line 400: Note that, by not checking `der` against `cert_data + cert_length`, you're also accepting PEM blocks with extra trailing data. Although the functions inside OpenSSL and BoringSSL also accept that due to a bug... which we hope to fix in the future.

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Unfortunately, I don't think we can use `bssl::UniquePtr` as we need to be compatible with OpenSSL. (As an aside, adding our own smart pointer types is a good idea so we can avoid all the manual cleanup, but I'll leave that out-of-scope for this PR if that's OK.)

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Thanks for the suggestion - I didn't know about this API! Switched to use it.

src/core/tsi/ssl_transport_security_utils.h
matthewstevenson88: I think this comment is obsolete after switching to use `PEM_read_bio_PrivateKey`.

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Thanks for the suggestion! Switched to use `PEM_read_bio_PrivateKey`.

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Based on @bob-beck's suggestion, I changed the API signature so now it parses and returns the vector of `X509`s. Also changed the name to simply `ParsePemCertificateChain` to hopefully make it more clear.

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Ah great call! Is this still relevant now that we switched to using `PEM_read_bio_X509`?

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Thanks @bob-beck! This is a great point. What I'm planning now is the following: in this PR, have the APIs for parsing the PEM; in a follow-up PR, call this API at server creation time and then plumb the parsed PEM down to where it needs to go. (This will be a lot of plumbing hence why I'd like to defer it to a follow-up.)

src/core/tsi/ssl_transport_security_utils.cc
matthewstevenson88: Thanks for calling this out!  I think we are fine skipping over any extra unrelated portions of the input string. The updated comment on the API also know no longer says that `cert_chain_pem` must consist of exactly the PEM cert chain.

matthewstevenson88:
Thanks @davidben and @bob-beck for your reviews! Are there any other major concerns? If not, I'll go ahead and merge. :)

PR #36371, opened 2024-04-16
[Python Otel] Use registered method at server side

XuanWang-Amos:
Add registered method change to sync stack server.

## Note
* We're keeping both generic and registered handler in generated stubs (`*_pb2_grpc.py`) because registered method flow is not implemented in aysncio yet, since the stub is used for both sync and asyncio, we can't remove generic method flow until we fully implement the registered method flow in asyncio.


src/python/grpcio/grpc/_observability.py
```diff
@@ -174,6 +178,9 @@ def set_stats(self, enable: bool) -> None:
         """
         self._stats_enabled = enable
 
+    def save_registered_method(self, method_name: bytes):
+        self._registered_methods.add(method_name)
```
gnossen, line 182: Ideally, plugins shouldn't be accessing underscore-prefixed attributes in the base class. The interface should be method-only. We're already doing this for `tracing_enabled` (via the `@property` decorator). This is somewhat less important for backward-compatibility purposes since we're pinning the version of `grpcio-observability` to be the same as `grpcio`, but it's good practice in general.

src/python/grpcio_observability/grpc_observability/_gcp_observability.py
```diff
@@ -121,8 +121,11 @@ def create_client_call_tracer(
         self, method_name: bytes, target: bytes
     ) -> ClientCallTracerCapsule:
         trace_id = b"TRACE_ID"
+        registered_method = False
+        if method_name in self._registered_methods:
+            registered_method = True
         capsule = _cyobservability.create_client_call_tracer(
-            method_name, target, trace_id
+            method_name, target, trace_id, registered_method
```
gnossen, line 128: Nit: Could be simplified to:

```python
capsule = _cyobservability.create_client_call_tracer(
            method_name, target, trace_id, method_name in self._registered_methods)
```

No need for the preceding 3 lines.

src/python/grpcio_observability/grpc_observability/_gcp_observability.py
```diff
@@ -150,6 +153,14 @@ def record_rpc_latency(
...
+        if method in self._registered_methods:
+            registered_method = True
         _cyobservability._record_rpc_latency(
-            self.exporter, method, target, rpc_latency, status_code
+            self.exporter,
+            method,
+            target,
+            rpc_latency,
+            status_code,
+            registered_method,
```
gnossen, line 165: Likewise here.

src/python/grpcio_observability/grpc_observability/python_census_context.h
```diff
@@ -105,6 +105,7 @@ struct Measurement {
   MetricsName name;
   MeasurementType type;
   MeasurementValue value;
+  bool registered_method;
```
gnossen, line 108: Should this file be renamed?

src/python/grpcio/grpc/_server.py
```diff
@@ -1172,15 +1205,23 @@ def _process_event_and_continue(
...
         with state.lock:
-            state.due.remove(_REQUEST_CALL_TAG)
+            state.due.remove(event.tag)
             concurrency_exceeded = (
                 state.maximum_concurrent_rpcs is not None
                 and state.active_rpc_count >= state.maximum_concurrent_rpcs
             )
             rpc_state, rpc_future = _handle_call(
                 event,
+                registered_method_name,
```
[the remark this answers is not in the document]
XuanWang-Amos: I think we should handle both case with the same `_handle_call` and `_find_method_handler` function because `_handle_call` also takes care of exception handling and `_find_method_handler` takes care of interceptors, both of them works the same for registered and unregistered calls.

Agree that we can refactor them to reduce the number of conditions, let me see what I can do.

src/python/grpcio_observability/grpc_observability/python_census_context.h
XuanWang-Amos: We might still need this for tracing, I think census sounds general enough, I do need to change `GRPC_PYTHON_OPENCENSUS_H` tho.

src/python/grpcio/grpc/_cython/_cygrpc/server.pyx.pxi
```diff
@@ -13,6 +13,22 @@
...
+    cpython.Py_INCREF(self.method)
+    cdef grpc_server *c_server = <grpc_server *>server
+    # TODO(xuanwn): Consider use GRPC_SRM_PAYLOAD_READ_INITIAL_BYTE_BUFFER for unary request
+    # as optimization.
+    # Note that in stubs method is not bound to any host, thus we set host as NULL.
+    self.c_registered_method = grpc_server_register_method(c_server,
+      <const char *>self.method, NULL, GRPC_SRM_PAYLOAD_NONE, 0)
+
+  def __dealloc__(self):
+    cpython.Py_DECREF(self.method)
```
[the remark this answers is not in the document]
gnossen: Can you add a comment in the `__dealloc__` method explaining this?

src/python/grpcio_observability/grpc_observability/python_census_context.h
gnossen: Discussed offline. We'll go with `python_o11y_context` or `python_observability_context`.

src/python/grpcio_tests/tests_py3_only/interop/xds_interop_client.py
```diff
@@ -529,7 +529,6 @@ def _run(
         _XdsUpdateClientConfigureServicer(channel_configs, args.qps),
         _global_server,
     )
-    channelz.add_channelz_servicer(_global_server)
```
XuanWang-Amos, line 532: Remove this line because we're also doing it in `grpc_admin.add_admin_servicers` below.

src/python/grpcio/grpc/_observability.py
gnossen: Ping.

Issue #36265, opened 2024-04-07
bug(python): Regression in `grpcio>=1.59.0` related to grpc thread pool hanging in `grpc_google_default_credentials_create`


parthea:
### What version of gRPC and what language are you using?
Latest gRPC at HEAD
python

### What operating system (Linux, Windows,...) and version?
Linux (ubuntu, rodete)

### What runtime / compiler are you using (e.g. python version or version of gcc)
Python 3.9.16 
gcc version 13.2.0 (Debian 13.2.0-10) 

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.

See unit test in https://github.com/grpc/grpc/commit/da7349b36e289eae80727c07fdd436a50fa8ed98 where the issue from #34672 was created using `grpc.compute_engine_channel_credentials`

### What did you expect to see?
I expected the code to run successfully (not hang)

### What did you see instead?
Code hangs similar to https://github.com/grpc/grpc/issues/34672

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

### Anything else we should know about your project / environment?
See related issue https://github.com/grpc/grpc/issues/34672 where a problem was fixed for `grpc_ssl_credentials_create`. There is a similar issue here where the call to `grpc_google_default_credentials_create` also hangs.

See https://github.com/googleapis/python-bigtable/issues/949 where this issue originally was reported.

PR #36266, opened 2024-04-07
[Fix Python Deadlock] Guard grpc_google_default_credentials_create with nogil

parthea:
This fix is similar to https://github.com/grpc/grpc/pull/34712 but for `grpc_google_default_credentials_create` rather than `grpc_ssl_credentials_create`

Fixes https://github.com/grpc/grpc/issues/36265
Fixes https://github.com/googleapis/python-bigtable/issues/949

PR #36376, opened 2024-04-16
[Fix Python Deadlock] Guard grpc_google_default_credentials_create with nogil (v1.62.x backport)

XuanWang-Amos:
Backport of #36266 to v1.62.x.
---
This fix is similar to https://github.com/grpc/grpc/pull/34712 but for `grpc_google_default_credentials_create` rather than `grpc_ssl_credentials_create`

Fixes https://github.com/grpc/grpc/issues/36265
Fixes https://github.com/googleapis/python-bigtable/issues/949


eugeneo (approved):
Just to confirm - is src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi an actual file that needs to be edited? Path starting with underscores makes it look like a generated file.

XuanWang-Amos:
> Just to confirm - is src/python/grpcio/grpc/_cython/_cygrpc/credentials.pyx.pxi an actual file that needs to be edited? Path starting with underscores makes it look like a generated file.

Yes, that's a manually edited file. Underscore in Python means it's a private file.

PR #36396, opened 2024-04-18
[test] Wait for `RecordEnd` in `FakeClientCallAttemptTracer` destruction

yijiem:

### DIFF
--- a/test/core/util/fake_stats_plugin.h
+++ b/test/core/util/fake_stats_plugin.h
@@ -115,8 +115,7 @@ class FakeClientCallTracer : public ClientCallTracer {
   ~FakeClientCallTracer() override {}
   CallAttemptTracer* StartNewAttempt(bool /*is_transparent_retry*/) override {
     auto call_attempt_tracer =
-        grpc_core::MakeRefCounted<FakeClientCallAttemptTracer>(
-            annotation_logger_);
+        MakeRefCounted<FakeClientCallAttemptTracer>(annotation_logger_);
     call_attempt_tracers_.emplace_back(call_attempt_tracer);
     return call_attempt_tracer.release();  // Released in RecordEnd().
   }
@@ -135,8 +134,7 @@ class FakeClientCallTracer : public ClientCallTracer {
 
  private:
   std::vector<std::string>* annotation_logger_;
-  std::vector<grpc_core::RefCountedPtr<FakeClientCallAttemptTracer>>
-      call_attempt_tracers_;
+  std::vector<RefCountedPtr<FakeClientCallAttemptTracer>> call_attempt_tracers_;
 };
 
 #define GRPC_ARG_INJECT_FAKE_CLIENT_CALL_TRACER_FACTORY \
--- a/test/core/util/fake_stats_plugin.h
+++ b/test/core/util/fake_stats_plugin.h
@@ -23,7 +23,6 @@
 #include "absl/functional/any_invocable.h"
 #include "absl/status/status.h"
 #include "absl/strings/string_view.h"
-#include "absl/synchronization/notification.h"
 #include "absl/types/optional.h"
 #include "absl/types/span.h"
 #include "gmock/gmock.h"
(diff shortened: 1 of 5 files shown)


test/core/util/fake_stats_plugin.h
```diff
@@ -64,7 +65,9 @@ class FakeClientCallTracer : public ClientCallTracer {
     explicit FakeClientCallAttemptTracer(
         std::vector<std::string>* annotation_logger)
         : annotation_logger_(annotation_logger) {}
-    ~FakeClientCallAttemptTracer() override {}
+    ~FakeClientCallAttemptTracer() override {
```
yashykt, line 68: Using a notification wait in a destructor seems scary.. Instead, why not just allocate it on the heap and delete it on `RecordEnd`?

test/core/util/fake_stats_plugin.h
```diff
@@ -105,6 +110,7 @@ class FakeClientCallTracer : public ClientCallTracer {
     }
 
    private:
+    absl::Notification notification;
```
markdroth, line 113: This needs a `_` suffix.

test/core/util/fake_stats_plugin.h
markdroth: I don't think there's necessarily anything wrong with waiting on the notification here, but I agree that it seems cleaner to just ensure the object isn't destroyed until `RecordEnd()` is called.  Note though that we can't just unconditionally destroy the object in `RecordEnd()`, because when we're not yet at the end of the test, we need the object to still exist so that we can access it in the test code.  So I think we'll want to make `FakeClientCallAttemptTracer` be ref-counted, take a new ref to store in `FakeClientCallTracer::call_attempt_tracers_`, and then have `RecordEnd()` drop the original ref.

markdroth (commented):
Thanks for fixing this!

test/core/util/fake_stats_plugin.h
yijiem: Thanks! Done. PTAL.

test/core/util/fake_stats_plugin.h
yijiem: Done.

test/core/util/fake_stats_plugin.h
```diff
@@ -60,14 +60,15 @@ void RegisterFakeStatsPlugin();
...
+        public RefCounted<FakeClientCallAttemptTracer> {
    public:
     explicit FakeClientCallAttemptTracer(
         std::vector<std::string>* annotation_logger)
-        : annotation_logger_(annotation_logger) {}
-    ~FakeClientCallAttemptTracer() override {
-      notification.WaitForNotification();
-    }
+        : RefCounted<FakeClientCallAttemptTracer>(nullptr,
+                                                  2),  // RecordEnd holds a ref.
```
markdroth, line 69: Instead of initialing the ref-count to 2 (an API that I eventually plan to remove), let's change the code in `StartNewAttempt()` to:

```
auto call_attempt_tracer =
    grpc_core::MakeRefCounted<FakeClientCallAttemptTracer>(annotation_logger_);
call_attempt_tracers_.emplace_back(call_attempt_tracer);
return call_attempt_tracer.release();  // Released in RecordEnd().
```

test/core/util/fake_stats_plugin.h
```diff
@@ -60,14 +60,15 @@ void RegisterFakeStatsPlugin();
...
+          annotation_logger_(annotation_logger) {}
+    ~FakeClientCallAttemptTracer() override {}
```
markdroth, line 71: Not needed if it's empty.

test/core/util/fake_stats_plugin.h
yijiem: Nice! Done.

test/core/util/fake_stats_plugin.h
yijiem: Done.

markdroth (approved):

test/core/util/fake_stats_plugin.h
yashykt: That makes sense.

yashykt (approved):

PR #36400, opened 2024-04-18
[Python Distribution] Pin setuptools and wheel to a single version

XuanWang-Amos:


tools/release/verify_python_release.py
```diff
@@ -44,6 +44,7 @@
     "grpcio-testing",
     "grpcio-admin",
     "grpcio-csds",
+    "grpcio-observability",
```
gnossen, line 47: Nice catch!

gnossen (approved):

PR #36413, opened 2024-04-20
[CSM] Fix CSM Observability for trailers-only response

yashykt:
Before this change, on a trailers-only response, Metadata Exchange needed by CSM would just be dropped, and hence CSM labels would not be seen.

Changes - 
* OTel call attempt tracer populates labels from trailers if it's a trailers-only response.
* HTTP2 layer propagates the Metadata Exchange field from headers to trailers for trailers-only responses.
* Add a test to make sure that retries continue to work when Metadata Exchange is enabled and a trailers-only response is sent. (This verifies that a trailers-only response remains a trailers-only response.)


markdroth (approved):
This looks great to me!  Please make sure @ctiller signs off before merging.

ctiller (approved):

PR #36426, opened 2024-04-22
[handshaker] Change handshake manager to clean up read buffer even if endpoint is null

anicr7:
This leak only impacts handshakers that don't have any endpoint associated with them and if the shutdown occurs at the same time as the handshaker finishing successfully. Currently the memory allocated for the read buffer is only cleaned up if the endpoint is not a nullptr triggering the leak. This change unilaterally cleans up the memory for this condition.

cc: @markdroth

### DIFF
--- a/src/core/lib/transport/handshaker.cc
+++ b/src/core/lib/transport/handshaker.cc
@@ -101,11 +101,6 @@ bool HandshakeManager::CallNextHandshakerLocked(grpc_error_handle error) {
             HandshakerArgsString(&args_).c_str());
   }
   GPR_ASSERT(index_ <= handshakers_.size());
-  if (args_.read_buffer != nullptr) {
-    grpc_slice_buffer_destroy(args_.read_buffer);
-    gpr_free(args_.read_buffer);
-    args_.read_buffer = nullptr;
-  }
   // If we got an error or we've been shut down or we're exiting early or
   // we've finished the last handshaker, invoke the on_handshake_done
   // callback.  Otherwise, call the next handshaker.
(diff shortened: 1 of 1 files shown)


markdroth:
I don't think we want this change.  The code in the handshake manager is intended to catch only the case where the `on_handshake_done` callback was already scheduled with OK status before the handshaker got shut down.  In that case, the handshaker will not have cleaned up anything, so cleaning up both the endpoint and the read buffer seems fine.

I do think the cleanup semantics here are currently difficult, and I have some other changes in the works to improve them, but I don't think this particular PR is a step forward, and I don't want to change this code twice.

markdroth:
Upon further discussion, it looks like this is actually needed.  The problem occurs when there is a handshaker registered at priority `kPreTCPConnectHandshakers`.  Those handshakers run before the TCP connect handshaker, so there is no endpoint yet.  If one of those handshakers schedules the completion callback with OK status before it sees shutdown, the handshake manager still needs to clean up the read buffer, even if there is no endpoint.

I'm working on a separate change that will solve this in a better way, but since I need more work to get that landed, let's go ahead with this change for now, and I'll replace it with something better later.

src/core/lib/transport/handshaker.cc
```diff
@@ -101,6 +101,11 @@ bool HandshakeManager::CallNextHandshakerLocked(grpc_error_handle error) {
             HandshakerArgsString(&args_).c_str());
   }
   GPR_ASSERT(index_ <= handshakers_.size());
+  if (args_.read_buffer != nullptr) {
```
markdroth, line 104: We definitely do not want to do this here.  The read buffer has to be passed between handshakers and should not be destroyed unless handshaking fails.

Instead, I think we need to put this code back where it was, but allow it to be cleaned up if `read_buffer` is non-null, independently of whether the endpoint is null:

```
      if (args_.endpoint != nullptr) {
        // TODO(roth): It is currently necessary to shutdown endpoints
        // before destroying then, even when we know that there are no
        // pending read/write callbacks.  This should be fixed, at which
        // point this can be removed.
        grpc_endpoint_shutdown(args_.endpoint, error);
        grpc_endpoint_destroy(args_.endpoint);
      }
      if (args_.read_buffer != nullptr) {
        grpc_slice_buffer_destroy(args_.read_buffer);
        gpr_free(args_.read_buffer);
        args_.read_buffer = nullptr;
      }
      args_.endpoint = nullptr;
```

markdroth: (Made a small edit after sending)

anicr7: Done.

markdroth (approved):

PR #36429, opened 2024-04-23
Add config setting and default_engine_factory for visionOS.

tfg4:
Add config setting and default_engine_factory for visionOS.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -128,6 +128,11 @@ config_setting(
 )
 
 config_setting(
+    name = "visionos",
+    constraint_values = ["//third_party/bazel_platforms/os:visionos"],
+)
+
+config_setting(
     name = "watchos",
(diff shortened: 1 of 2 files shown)


veblush:
@sampajano Would you look at this? If we want to add new `visionos` build configuration, do we need to do for other builds such as podspec?

sampajano:
@veblush Thank you for forwarding it to me!!

---

@HannahShiSFB Hi Hannah, could you help take a look at this request, and check if there's other places that we need to match it like mentioned above? Thanks!!

sampajano:
@tfg4 Hi!

We have worked on another PR for adding visionOS support to grpc here:
https://github.com/grpc/grpc/pull/36476

That PR has all accompanying podspec change too, so i think we will use that one instead of this PR.

Thanks for your contribution!

PR #36435, opened 2024-04-24

yashykt:


sampajano:
@HannahShiSFB Hi Hannah could you help take a look at what are some potential causes for those iOS failures?

I know "Basic Tests ObjC iOS" is flaky ~5% of the time so it could just be that.

But the "Basic Tests ObjC Examples" failures aren't known to be flaky.

templates/src/objective-c/BoringSSL-GRPC.podspec.template
```diff
@@ -70,7 +70,7 @@
 
   Pod::Spec.new do |s|
     s.name     = 'BoringSSL-GRPC'
-    version = '0.0.33'
+    version = '0.0.34'
     s.version  = version
     s.summary  = 'BoringSSL is a fork of OpenSSL that is designed to meet Google\'s needs.'
     # Adapted from the homepage:
```
HannahShiSFB, line 76: Can you try add the include files pattern in line 152 to include sub directories?
Add something like `'src/include/openssl/**/*.h'`

yashykt: That fixed that, but looks like there are other issues now :(

HannahShiSFB: Looks like path to [err_data.c](https://github.com/google/boringssl/blob/master/gen/crypto/err_data.c) changed, can you update it accordingly?

yashykt: Thanks! That fixed it

veblush (approved):

PR #36437, opened 2024-04-24

tanvi-jagtap:


test/core/util/histogram.cc
```diff
@@ -87,8 +89,8 @@ grpc_histogram* grpc_histogram_create(double resolution,
   h->min_seen = max_bucket_start;
   h->max_seen = 0.0;
   h->num_buckets = bucket_for_unchecked(h, max_bucket_start) + 1;
-  GPR_ASSERT(h->num_buckets > 1);
-  GPR_ASSERT(h->num_buckets < 100000000);
+  CHECK_GT(h->num_buckets, 1);
+  CHECK(h->num_buckets < 100000000);
```
yousukseung, line 93: nit: I keep seeing this inconsistency i.e. `CHECK(<, >, !=)` and `CHECK_LE|GE|NE` are mixed here and there. Is this something you want to fix?

yousukseung (approved):
LGTM, had one nit comment which is non-blocking.

test/core/util/histogram.cc
tanvi-jagtap: Will not be replacing all instances of CHECK with CHECK_EQ , CHECK_NE etc because there are too many callsites. Only ones which are doable using very simple regex with least chance of failure will be replaced.

yousukseung: My question was if the regex needs to be checked. In this example, do you know why '>' was replaced with `CHECK_GT` but '<' was not?

PR #36448, opened 2024-04-24
[Python Otel] Remove opentelemetry version pin

XuanWang-Amos:

### DIFF
--- a/requirements.bazel.txt
+++ b/requirements.bazel.txt
@@ -26,8 +26,8 @@ pyasn1==0.5.0
 rsa==4.9
 greenlet==1.1.3.post0
 zope.interface==6.1
-opentelemetry-sdk==1.21.0
-opentelemetry-api==1.21.0
+opentelemetry-sdk>=1.21.0
+opentelemetry-api>=1.21.0
 importlib-metadata==6.11.0
 Deprecated==1.2.14
 opentelemetry-semantic-conventions==0.42b0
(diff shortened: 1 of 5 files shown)


requirements.bazel.txt
```diff
@@ -26,8 +26,8 @@ pyasn1==0.5.0
 rsa==4.9
 greenlet==1.1.3.post0
 zope.interface==6.1
-opentelemetry-sdk==1.21.0
-opentelemetry-api==1.21.0
```
gnossen, line 30: Let's keep the pin for Bazel unless we have a strong reason not to. We want the testing to be deterministic.

src/python/grpcio_tests/tests/observability/BUILD.bazel
```diff
@@ -34,6 +37,7 @@ py_test(
         "//src/python/grpcio/grpc:grpcio",
         "//src/python/grpcio_observability/grpc_observability:pyobservability",
         "//src/python/grpcio_tests/tests/testing",
+        requirement("opentelemetry-sdk"),
```
gnossen, line 40: Interesting. Why is this dependency necessary now?

XuanWang-Amos: I removed sdk requirement from py_library `_opentelemetry_observability`, but sdk is still required in our testing.

gnossen: Ah. This makes a lot of sense. Good change! :+1:

gnossen (approved):

PR #36449, opened 2024-04-24
[metrics] Templatized Metrics API

yijiem:
This adds compile-time checking that the type of the value and the size of the labels and optional labels passed when recording a metric must match with the type and the size specified when the metric is registered.

The RegistrationBuilder API idea is credited to @ctiller.


src/core/lib/channel/metrics.h
```diff
@@ -151,6 +362,26 @@ class CallbackMetricReporter {
 
 class RegisteredMetricCallback;
 
+class CallbackMetricReporterWrapper {
```
markdroth, line 365: I think if my suggestion above pans out, then we won't need this wrapper anymore.

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -313,10 +317,10 @@ GrpcXdsClient::GrpcXdsClient(
               .certificate_providers())),
       stats_plugin_group_(GetStatsPluginGroupForKey(key_)),
       registered_metric_callback_(stats_plugin_group_.RegisterCallback(
-          [this](CallbackMetricReporter& reporter) {
+          [this](CallbackMetricReporterWrapper& reporter) {
             ReportCallbackMetrics(reporter);
           },
-          {kMetricConnected, kMetricResources})) {}
+          {kMetricConnected.convert(), kMetricResources.convert()})) {}
```
markdroth, line 323: The need to call `convert()` here is really ugly.  I think this is a symptom of the fact that we have two different representations for each handle.  I have a suggestion elsewhere that should hopefully avoid the need to have two different types in the first place, but if we do wind up having a second type for some reason, I think it needs to be completely hidden within the API, not exposed to callers like this.

markdroth (commented):
Thanks for doing this!  Overall, this looks very good, but I think we can improve it further.

src/core/lib/channel/metrics.h
```diff
@@ -87,37 +91,238 @@ class GlobalInstrumentsRegistry {
   using GlobalCallbackHandle = absl::variant<GlobalCallbackInt64GaugeHandle,
                                              GlobalCallbackDoubleGaugeHandle>;
 
+  template <typename ValueType, typename InstrumentType, std::size_t M,
+            std::size_t N>
+  struct MagicHandle : public GlobalInstrumentHandle {
```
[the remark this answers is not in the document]
yijiem: Not sure how this will work, the size information needs to be attached on the handle to allow compile-time check if the label value sizes are different from the label key sizes. There is another limitation that we cannot templatize the StatsPlugin interface since C++ does not support virtual template member function.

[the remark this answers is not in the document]
markdroth: Good point about needing the size information in the handle.  I think what I sketched out above can be changed to do that:

```
class GlobalInstrumentsRegistry {
 public:
  // ...

  template <ValueType V, InstrumentType I, size_t N, size_t M>
  struct TypedGlobalInstrumentHandle : public GlobalInstrumentHandle {
    static constexpr ValueType value_type = V;
    static constexpr InstrumentType instrument_type = I;
    static constexpr size_t num_labels = N;
    static constexpr size_t num_optional_labels = M;
  };

  template <ValueType V, InstrumentType I, std::size_t M, std::size_t N>
  class RegistrationBuilder {
   public:
    // ...

    TypedGlobalInstrumentHandle<V, I, N, M> Build() {
      TypedGlobalInstrumentHandle<V, I, N, M> handle;
      handle.index = GlobalInstrumentsRegistry::RegisterInstrument(
          V, I, description_, unit_, enable_by_default_, label_keys_,
          optional_label_keys_);
      return handle;
    }

    // ...
  };

  // Creates instrument in the GlobalInstrumentsRegistry.
  static RegistrationBuilder<ValueType::kUInt64, InstrumentType::kCounter, 0, 0>
  RegisterUInt64Counter(
      absl::string_view name, absl::string_view description,
      absl::string_view unit, bool enable_by_default) {
    return RegistrationBuilder<ValueType::kUInt64, InstrumentType::kCounter, 0, 0>(
        name, description, unit, enable_by_default);
  }

  // ...
};

class StatsPlugin {
 public:
  // ...

  virtual void AddCounter(
      GlobalInstrumentsRegistry::GlobalInstrumentHandle handle,
      uint64_t value, absl::Span<const absl::string_view> label_values,
      absl::Span<const absl::string_view> optional_label_values) = 0;

  // ...
};

class StatsPluginRegistry {
 public:
   // ...

  class StatsPluginGroup {
   public:
    // ...

    template <typename HandleType, typename NativeValueType,
              typename = typename std::enable_if<
                  std::is_same<HandleType::instrument_type, InstrumentType::kCounter>::value>::type>
    void AddCounter(HandleType handle, NativeValueType value,
                    std::array<absl::string_view, HandleType::num_labels> label_values,
                    std::array<absl::string_view, HandleType::num_optional_labels> optional_values) {
      for (auto& state : plugins_state_) {
        state.plugin->AddCounter(handle, value, label_values,
                                 optional_values);
      }
    }

    template <typename HandleType, typename NativeValueType,
              typename = typename std::enable_if<
                  std::is_same<HandleType::instrument_type, InstrumentType::kHistogram>::value>::type>
    void RecordHistogram(HandleType handle, NativeValueType value,
                    std::array<absl::string_view, HandleType::num_labels> label_values,
                    std::array<absl::string_view, HandleType::num_optional_labels> optional_values) {
      for (auto& state : plugins_state_) {
        state.plugin->RecordHistogram(handle, value, label_values,
                                 optional_values);
      }
    }

    // ...
  };

  // ...
};
```

Note that this means we're removing the convenience aliases (`GlobalUInt64CounterHandle` and friends) and changing the `StatsPlugin` methods to use the base `GlobalInstrumentHandle` type.  But there should be no need for a virtual templated function there.  (Note that gRPC components that define metrics don't directly use the `StatsPlugin` API; they go through `StatsPluginGroup`.  So as long as we have the right compile-time checking in `StatsPluginGroup`, I think we're fine.)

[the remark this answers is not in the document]
yijiem: Great suggestion! This indeed simplifies things a bit and good suggestion on using the ValueType and InstrumentType enum class as non-type template parameters. PTAL.

src/core/lib/channel/metrics.h
yijiem: Indeed we don't need this wrapper. Woohoo!

src/core/ext/xds/xds_client_grpc.cc
yijiem: No need of this either, it just gets implicitly converted to its base type.

src/core/lib/channel/metrics.h
```diff
@@ -288,7 +478,7 @@ class GlobalStatsPluginRegistry {
     GRPC_MUST_USE_RESULT std::unique_ptr<RegisteredMetricCallback>
     RegisterCallback(
         absl::AnyInvocable<void(CallbackMetricReporter&)> callback,
-        std::vector<GlobalInstrumentsRegistry::GlobalCallbackHandle> metrics,
+        std::vector<GlobalInstrumentsRegistry::GlobalInstrumentHandle> metrics,
```
markdroth, line 481: This change actually loses us a bit of type safety: previously, it was not possible to register a callback for a metric with InstrumentType != kCallbackGauge, but this change seems to allow it for other types as well.

Is there some reasonable way we can get type safety back for this?

src/core/ext/xds/xds_client_grpc.cc
```diff
@@ -167,7 +171,7 @@ class GrpcXdsClient::MetricsReporter final : public XdsMetricsReporter {
 
   void ReportServerFailure(absl::string_view xds_server) override {
     xds_client_.stats_plugin_group_.AddCounter(
-        kMetricServerFailure, 1, {xds_client_.key_, xds_server}, {});
+        kMetricServerFailure, uint64_t(1), {xds_client_.key_, xds_server}, {});
```
markdroth, line 174: Is this explicit type construction still needed?

Same question throughout.

src/core/load_balancing/rls/rls.cc
```diff
@@ -754,7 +757,10 @@ class RlsLb final : public LoadBalancingPolicy {
   void UpdatePickerLocked() ABSL_LOCKS_EXCLUDED(&mu_);
 
   void MaybeExportPickCount(
-      GlobalInstrumentsRegistry::GlobalUInt64CounterHandle handle,
+      GlobalInstrumentsRegistry::TypedGlobalInstrumentHandle<
```
markdroth, line 760: Is there a way we can avoid writing this type out explicitly?

markdroth (commented):
This looks really good!  Just a few minor comments.

src/core/lib/channel/metrics.h
```diff
@@ -78,46 +79,208 @@ class GlobalInstrumentsRegistry {
...
-  struct GlobalUInt64HistogramHandle : public GlobalInstrumentHandle {};
-  struct GlobalDoubleHistogramHandle : public GlobalInstrumentHandle {};
-  struct GlobalCallbackInt64GaugeHandle : public GlobalInstrumentHandle {};
-  struct GlobalCallbackDoubleGaugeHandle : public GlobalInstrumentHandle {};
-  using GlobalCallbackHandle = absl::variant<GlobalCallbackInt64GaugeHandle,
-                                             GlobalCallbackDoubleGaugeHandle>;
+
+  template <ValueType V, InstrumentType I, size_t M, size_t N>
+  struct TypedGlobalInstrumentHandle : public GlobalInstrumentHandle {
+    static constexpr ValueType value_type = V;
```
markdroth, line 85: I'm not sure we actually need these static data members.  I thought we would wind up using them, but it doesn't look like we're actually doing so anywhere.

src/core/lib/channel/metrics.h
```diff
@@ -139,14 +308,27 @@ class CallbackMetricReporter {
...
-      absl::Span<const absl::string_view> optional_values) = 0;
+  template <GlobalInstrumentsRegistry::ValueType V,
+            GlobalInstrumentsRegistry::InstrumentType I, std::size_t M,
+            std::size_t N, typename ValueType>
+  void Report(
+      GlobalInstrumentsRegistry::TypedGlobalInstrumentHandle<V, I, M, N> handle,
+      ValueType value, std::array<absl::string_view, M> label_values,
+      std::array<absl::string_view, N> optional_values) {
+    Report(
+        static_cast<GlobalInstrumentsRegistry::GlobalInstrumentHandle>(handle),
```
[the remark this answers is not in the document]
yijiem: Without this cast it would call itself and results into stackoverflow.

src/core/ext/xds/xds_client_grpc.cc
yijiem: We need the static_cast somewhere otherwise the call to `AddCounter` is ambiguous, I can move the static_cast into the API though.

src/core/load_balancing/rls/rls.cc
yijiem: I'm not sure and I don't think this is a big concern though. This type is part of the templated API and the caller should be able to referring to it directly.

src/core/lib/channel/metrics.h
```diff
@@ -78,46 +79,208 @@ class GlobalInstrumentsRegistry {
...
+          std::array<absl::string_view, sizeof...(Args)>({args...}));
+    }
+
+    template <ValueType VV = V, InstrumentType II = I,
+              typename =
+                  typename std::enable_if<VV == ValueType::kUInt64 &&
+                                          II == InstrumentType::kCounter>::type>
+    TypedGlobalInstrumentHandle<ValueType::kUInt64, InstrumentType::kCounter, M,
+                                N>
+    Build() {
```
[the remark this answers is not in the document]
yijiem: Great catch! Done.

src/core/lib/channel/metrics.h
yijiem: Done. Removed.

src/core/lib/channel/metrics.h
yijiem: Great point! I changed this API a bit to be able to statically check the handles' type. PTAL.

src/core/lib/channel/metrics.h
```diff
@@ -78,46 +79,128 @@ class GlobalInstrumentsRegistry {
...
-  struct GlobalCallbackDoubleGaugeHandle : public GlobalInstrumentHandle {};
-  using GlobalCallbackHandle = absl::variant<GlobalCallbackInt64GaugeHandle,
-                                             GlobalCallbackDoubleGaugeHandle>;
+
+  template <ValueType V, InstrumentType I, size_t M, size_t N>
+  struct TypedGlobalInstrumentHandle : public GlobalInstrumentHandle {};
+
+  // C++17 has fold expression that may simplify this.
+  template <ValueType V, InstrumentType I, size_t M, size_t N>
+  static constexpr void IsCallbackGaugeHandle(
```
markdroth, line 88: Let's not make this part of the public API surface.  Suggest moving it to `GlobalStatsPluginRegistry` (which is where it's actually used) and making it private.

Also, suggest calling this `AssertIsCallbackGaugeHandle()`.

markdroth (commented):
This is really close!  Just a few minor comments remaining.

src/core/lib/channel/metrics.h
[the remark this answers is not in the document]
yijiem: Done.

src/core/lib/channel/metrics.h
yijiem: Done.

src/core/load_balancing/rls/rls.cc
yijiem: Done. Chose the template approach.

yijiem (commented):
Thanks Mark!

markdroth (approved):
This looks fantastic!  Thanks for doing this!

src/core/lib/channel/metrics.h
```diff
@@ -87,37 +91,238 @@ class GlobalInstrumentsRegistry {
   using GlobalCallbackHandle = absl::variant<GlobalCallbackInt64GaugeHandle,
                                              GlobalCallbackDoubleGaugeHandle>;
 
+  template <typename ValueType, typename InstrumentType, std::size_t M,
```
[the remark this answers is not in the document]
yijiem: Done.

PR #36472, opened 2024-04-29

tanvi-jagtap:
Replacing GPR_ASSERT with absl CHECK.


src/core/lib/promise/activity.cc
```diff
@@ -54,7 +55,7 @@ class FreestandingActivity::Handle final : public Wakeable {
   // Activity is going away... drop its reference and sever the connection back.
   void DropActivity() ABSL_LOCKS_EXCLUDED(mu_) {
     mu_.Lock();
-    GPR_ASSERT(activity_ != nullptr);
+    CHECK_NE(activity_, nullptr);
```
yashykt, line 58: didn't you run into problems using this?

yashykt (approved):

src/core/lib/promise/activity.cc
tanvi-jagtap: The specific checks such as CHECK_NE and CHECK_EQ fail only in specific cases . 
One case is when the == or != operator is working because of an operator overload.
When we convert such a thing to CHECK_NE it fails ( as expected ).
There may be be more scanarios when builds fail with the use of CHECK_NE and CHECK_EQ but given the volume of changes - I have not investigated them in depth yet.

PR #36476, opened 2024-04-30
[ObjC] Support visionos

HannahShiSFB:
This should be able to unblock using grpc on vision os, will add tests later


sampajano (approved):
Thanks Hannah!

PR #36503, opened 2024-05-02

tanvi-jagtap:
Replacing GPR_ASSERT with absl CHECK.


src/core/ext/transport/chaotic_good/server/chaotic_good_server.cc
```diff
@@ -135,7 +136,7 @@ absl::StatusOr<int> ChaoticGoodServerListener::Bind(
 }
 
 absl::Status ChaoticGoodServerListener::StartListening() {
-  GPR_ASSERT(ee_listener_ != nullptr);
+  CHECK(ee_listener_ != nullptr);
```
Vignesh2208, line 139: should it be CHECK_NE ?

Vignesh2208 (approved):

PR #36510, opened 2024-05-02
[core] checking the status of thread creation

chansen3:
It is possible for allocation of a system resource to fail. While the Thread abstraction in the core allows for a 'success' parameter, it is optional and not generally checked where used. This can lead to degenerate behavior which is hard to diagnose. In particular, the Executors assign Closures to their queues regardless of whether the thread handling to queue is actually running, leading those async requests to hang.

This change makes the 'success' parameter required and tries to make the correct decision on how to proceed in the code where those Threads are started.  For the Executor, we will attempt to carry on with the number of threads we were able to create.  In other areas, we assert, assuming that whatever function is being performed by the allocated thread is critical to the overall function of the library.

For the tests, we ignore the value of 'success' as we assume the tests will fail if thread creation failed. This is no different than the existing behavior.

### DIFF
--- a/src/core/lib/gprpp/thd.h
+++ b/src/core/lib/gprpp/thd.h
@@ -92,14 +92,14 @@ class Thread {
   /// Normal constructor to create a thread with name \a thd_name,
   /// which will execute a thread based on function \a thd_body
   /// with argument \a arg once it is started.
-  /// The optional \a success argument indicates whether the thread
+  /// The \a success argument indicates whether the thread
   /// is successfully created.
   /// The optional \a options can be used to set the thread detachable.
   Thread(const char* thd_name, void (*thd_body)(void* arg), void* arg,
-         bool* success = nullptr, const Options& options = Options());
+         bool* success, const Options& options = Options());
 
-  Thread(const char* thd_name, absl::AnyInvocable<void()> fn,
-         bool* success = nullptr, const Options& options = Options())
+  Thread(const char* thd_name, absl::AnyInvocable<void()> fn, bool* success,
+         const Options& options = Options())
       : Thread(
             thd_name,
             [](void* p) {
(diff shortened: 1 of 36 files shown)


chansen3:
Resubmitting original PR: https://github.com/grpc/grpc/pull/33286

veblush (approved):

src/core/lib/gprpp/thd.h
```diff
@@ -92,14 +92,14 @@ class Thread {
   /// Normal constructor to create a thread with name \a thd_name,
   /// which will execute a thread based on function \a thd_body
   /// with argument \a arg once it is started.
-  /// The optional \a success argument indicates whether the thread
+  /// The \a success argument indicates whether the thread
   /// is successfully created.
   /// The optional \a options can be used to set the thread detachable.
   Thread(const char* thd_name, void (*thd_body)(void* arg), void* arg,
-         bool* success = nullptr, const Options& options = Options());
+         bool* success, const Options& options = Options());
```
veblush, line 99: We still need old way of creating Thread because internal code (that you cannot access) still need this.

src/core/lib/gprpp/thd.h
```diff
@@ -92,14 +92,14 @@ class Thread {
...
 
-  Thread(const char* thd_name, absl::AnyInvocable<void()> fn,
-         bool* success = nullptr, const Options& options = Options())
+  Thread(const char* thd_name, absl::AnyInvocable<void()> fn, bool* success,
```
veblush, line 101: This too.

src/core/lib/gprpp/thd.h
chansen3: Adding a new constructor that does not require a `success` argument would undermine the point of this change. It would allow future code to use this constructor and thereby re-introduce the problem we have that pthread_create might fail and the result is a silent, slow degradation of functionality which is very difficult to diagnose.

veblush: It'd be true but we cannot accept it without having a way to not break internal codes.

PR #36511, opened 2024-05-02
Optionally crash benchmark QPS client worker if it fails to connect to the server target.

AshZhang:
By default, the QPS worker will crash if it cannot connect to the server target while operating as a client.

When explicitly set to false, the `die_on_connection_failure` flag (true by default) will keep the QPS worker process running if it fails to connect to the server target. Instead of dying, the QPS worker will finish the client threads, cancel the context to end the benchmark run and terminate the QPS JSON driver, and revert to the idle state of waiting for a `RunClient` or `RunServer` invocation.

This is useful in benchmark setups where runs are invoked continuously on long-lived QPS worker processes, reducing spurious worker deaths and the need for manual restarts due to transient connectivity failures.

CC: @apolcyn

### DIFF
--- a/test/cpp/qps/client.h
+++ b/test/cpp/qps/client.h
@@ -262,6 +262,8 @@ class Client {
     return static_cast<bool>(gpr_atm_acq_load(&thread_pool_done_));
   }
 
+  bool all_channels_connected() { return all_channels_connected_; }
+
   class Thread {
    public:
     Thread(Client* client, size_t idx)
@@ -344,6 +346,7 @@ class Client {
   bool closed_loop_;
   gpr_atm thread_pool_done_;
   double median_latency_collection_interval_seconds_;  // In seconds
+  bool all_channels_connected_;
 
   void StartThreads(size_t num_threads) {
     gpr_atm_rel_store(&thread_pool_done_, static_cast<gpr_atm>(false));
@@ -490,8 +493,10 @@ class ClientImpl : public Client {
       cq.Next(&tag, &ok);
       Channel* channel = static_cast<Channel*>(tag);
       if (!ok) {
-        grpc_core::Crash(absl::StrFormat(
-            "Channel %p failed to connect within the deadline", channel));
+        gpr_log(GPR_ERROR, "Channel %p failed to connect within the deadline",
+                channel);
+                all_channels_connected_ = false;
+        return;
       } else {
         grpc_connectivity_state last_observed = channel->GetState(true);
         if (last_observed == GRPC_CHANNEL_READY) {
@@ -503,6 +508,7 @@ class ClientImpl : public Client {
         }
       }
     }
+    all_channels_connected_ = true;
   }
 
  protected:
--- a/test/cpp/qps/qps_worker.cc
+++ b/test/cpp/qps/qps_worker.cc
@@ -185,7 +185,7 @@ class WorkerServiceImpl final : public WorkerService::Service {
     acquired_ = false;
   }
 
-  Status RunClientBody(ServerContext* /*ctx*/,
+  Status RunClientBody(ServerContext* ctx,
                        ServerReaderWriter<ClientStatus, ClientArgs>* stream) {
     ClientArgs args;
     if (!stream->Read(&args)) {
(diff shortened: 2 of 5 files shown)


test/cpp/qps/qps_worker.cc
```diff
@@ -199,6 +199,16 @@ class WorkerServiceImpl final : public WorkerService::Service {
     if (!client) {
       return Status(StatusCode::INVALID_ARGUMENT, "Couldn't create client");
     }
+    if (!client->AllChannelsConnected()) {
+      if (worker_->ShouldDieOnConnectionFailure()) {
+        grpc_core::Crash("Client failed to connect to all channels");
```
apolcyn, line 204: nit: "Client failed to connect all channels"

test/cpp/qps/qps_worker.cc
```diff
@@ -199,6 +199,16 @@ class WorkerServiceImpl final : public WorkerService::Service {
...
+      } else {
+        client->AwaitThreadsCompletion();
+        ctx->TryCancel();
```
apolcyn, line 207: I don't think `TryCancel` is necessary?

Simply terminating the RPC with the status code and message below seems ideal

test/cpp/qps/qps_worker.h
```diff
@@ -39,11 +39,13 @@ extern std::vector<grpc::testing::Server*>* g_inproc_servers;
...
  public:
   explicit QpsWorker(int driver_port, int server_port,
-                     const std::string& credential_type);
+                     const std::string& credential_type,
+                     bool die_on_connection_failure);
   ~QpsWorker();
 
   bool Done() const;
   void MarkDone();
+  bool ShouldDieOnConnectionFailure();
```
apolcyn, line 48: nit: `bool die_on_connection_failure() const` since this is a trivial getter

test/cpp/qps/qps_worker.h
```diff
@@ -52,6 +54,7 @@ class QpsWorker {
  private:
   std::unique_ptr<WorkerServiceImpl> impl_;
   std::unique_ptr<grpc::Server> server_;
+  bool die_on_connection_failure_;
```
apolcyn, line 57: nit: const

test/cpp/qps/client.h
```diff
@@ -344,6 +346,7 @@ class Client {
   bool closed_loop_;
   gpr_atm thread_pool_done_;
   double median_latency_collection_interval_seconds_;  // In seconds
+  bool all_channels_connected_ = true;
```
apolcyn, line 349: nit: seems a little odd to initialize `true`. Can we initialize `false` and set true and the end of `WaitForChannelsToConnect` ?

test/cpp/qps/client.h
```diff
@@ -262,6 +262,8 @@ class Client {
     return static_cast<bool>(gpr_atm_acq_load(&thread_pool_done_));
   }
 
+  bool AllChannelsConnected() { return all_channels_connected_; }
```
apolcyn, line 265: nit: `all_channels_connected()` since trivial getter

test/cpp/qps/qps_worker.h
```diff
@@ -39,11 +39,13 @@ extern std::vector<grpc::testing::Server*>* g_inproc_servers;
 class QpsWorker {
  public:
   explicit QpsWorker(int driver_port, int server_port,
-                     const std::string& credential_type);
+                     const std::string& credential_type,
+                     bool die_on_connection_failure);
```
apolcyn, line 43: nit: maybe set this as a default parameter to `true` so that we don't need to update other call sites

test/cpp/qps/qps_worker.cc
AshZhang: `TryCancel` ensures that the `stream` passed in from the QPS JSON driver is closed, so that the driver will terminate if the client fails to connect. If the `TryCancel` isn't there, the driver will hang forever waiting for the client to write the initial status.

test/cpp/qps/qps_worker.h
AshZhang: Done.

test/cpp/qps/qps_worker.h
AshZhang: Done.

test/cpp/qps/client.h
AshZhang: Done.

test/cpp/qps/client.h
AshZhang: Done.

test/cpp/qps/qps_worker.h
AshZhang: Done.

PR #36555, opened 2024-05-07
[EventEngine] Rely on RAII for lifeguard

eugeneo:
This makes lifeguard easier to manage and removes one case of hang in fork test.

### DIFF
--- a/src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
+++ b/src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
@@ -38,7 +38,6 @@
 #include "src/core/lib/backoff/backoff.h"
 #include "src/core/lib/event_engine/common_closures.h"
 #include "src/core/lib/event_engine/thread_local.h"
-#include "src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.h"
 #include "src/core/lib/event_engine/trace.h"
 #include "src/core/lib/event_engine/work_queue/basic_work_queue.h"
 #include "src/core/lib/event_engine/work_queue/work_queue.h"
--- a/src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
+++ b/src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
@@ -229,12 +229,6 @@ WorkStealingThreadPool::WorkStealingThreadPoolImpl::WorkStealingThreadPoolImpl(
     size_t reserve_threads)
     : reserve_threads_(reserve_threads), queue_(this) {}
 
-WorkStealingThreadPool::WorkStealingThreadPoolImpl::
-    ~WorkStealingThreadPoolImpl() {
-  grpc_core::MutexLock lock(&lifeguard_ptr_mu_);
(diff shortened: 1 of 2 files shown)


src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -227,18 +227,25 @@ void WorkStealingThreadPool::PostforkChild() { pool_->Postfork(); }
 
 WorkStealingThreadPool::WorkStealingThreadPoolImpl::WorkStealingThreadPoolImpl(
     size_t reserve_threads)
-    : reserve_threads_(reserve_threads), queue_(this), lifeguard_(this) {}
+    : reserve_threads_(reserve_threads), queue_(this) {}
+
+WorkStealingThreadPool::WorkStealingThreadPoolImpl::
+    ~WorkStealingThreadPoolImpl() {
+  grpc_core::MutexLock lock(&lifeguard_ptr_mu_);
```
drfloob, line 234: This should be unnecessary, since the program will crash if Quiesce is not called before destruction, and Quiesce resets the lifeguard_.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
```diff
@@ -227,18 +227,25 @@ void WorkStealingThreadPool::PostforkChild() { pool_->Postfork(); }
...
   }
-  lifeguard_.Start();
+  grpc_core::MutexLock lock(&lifeguard_ptr_mu_);
+  lifeguard_ = std::make_unique<Lifeguard>(this);
 }
 
 void WorkStealingThreadPool::WorkStealingThreadPoolImpl::Run(
     EventEngine::Closure* closure) {
-  DCHECK(quiesced_.load(std::memory_order_relaxed) == false);
+  DCHECK(!IsQuiesced());
```
drfloob, line 248: Let's make this `CHECK`. It's invalid usage otherwise.

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
eugeneo: Would it make sense to add assert? Just in case.

drfloob: See https://github.com/grpc/grpc/blob/641dbfd4c24be9262b1b49a798e7f8116819d820/src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc#L186-L188

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
eugeneo: Done

src/core/lib/event_engine/thread_pool/work_stealing_thread_pool.cc
eugeneo: Done

eugeneo (commented):
Thank you for the review. I addressed the comments, ptal.

drfloob (approved):

PR #36576, opened 2024-05-10
[logging] Centralize configuration for trace flags

drfloob:
All TraceFlags are now configured in `src/core/lib/debug/trace_flags.yaml`. The format is:

```
my_flag:
  default: false                   # the default value; default=false
  description: Some Description
  debug_only: false                # debug_only flags only work in debug builds; default=false
  internal: false                  # internal flags will not show up in documentation; default=false
```

To regenerate the trace flag source code, run `tools/codegen/core/gen_trace_flags.py` (requires mako). This script is also run when sanity checking.

This PR also adds two new features:

### Glob-based flag configuration

Trace flag configuration now supports `?` (single wildcard character) and `*` (one or more wildcard characters).  For example, using `GRPC_TRACE='event_engine*'` will enable all flags that match that glob. It expands to:

* event_engine
* event_engine_client_channel_resolver
* event_engine_dns
* event_engine_endpoint
* event_engine_endpoint_data
* event_engine_poller

### A cleaner trace-logging macro in abseil logging format

If your goal is only to add log statements when the `fault_injection_filter` trace flag is enabled, you can use the macro:

```
GRPC_TRACE_LOG(fault_injection, INFO) << "Filtered:" << 42;
```

When the trace flag is enabled, the the log will show something like this:
```
I0000 00:00:1715733657.430042      16 file.cc:174] Filtered:42
```

---- 

Note: just like with the gpr_log to abseil logging conversion, the pre-existing trace logging usages can be replaced with the new tracing macro across multiple PRs.


src/core/lib/debug/trace.cc
```diff
@@ -104,15 +106,76 @@ void SavedTraceFlags::Restore() {
 }
 
 namespace {
+// A basic glob matcher based on https://research.swtch.com/glob.
+bool GlobMatch(absl::string_view name, absl::string_view trace_glob) {
```
ctiller, line 110: Let's move this to gprpp by itself (there are likely other things we can use this for)

src/core/lib/debug/trace.cc
```diff
@@ -104,15 +106,76 @@ void SavedTraceFlags::Restore() {
...
+  }
+  return true;
+}
+
+bool IsGlob(absl::string_view trace_glob) {
+  return std::any_of(trace_glob.begin(), trace_glob.end(),
+                     [](const char c) { return c == '?' || c == '*'; });
+}
+
 void ParseTracers(absl::string_view tracers) {
```
ctiller, line 158: Would be *really* good to combine this loop with the same (but duplicated) parser for experiments... I could use the glob syntax there too!

src/core/lib/debug/trace_flags.yaml
```diff
@@ -0,0 +1,343 @@
...
+# limitations under the License.
+#
+# File format:
+# Each config var must have a name, a default, and a description.
+# Optionally, a `debug-only: true` flag can be added to ensure that these flags
+# are only enabled in debug builds.
+
+api:
+  default: false
+  description: null
```
ctiller, line 22: Can we populate these?

src/core/lib/debug/trace.h
```diff
@@ -111,6 +112,472 @@ class SavedTraceFlags {
   std::map<std::string, bool> values_;
 };
 
+extern TraceFlag api_trace;
+#define GRPC_API_TRACE_LOG(level) \
```
markdroth, line 116: Instead of defining a separate macro for each tracer, how about just defining it once, such that the tracer can be passed in as a parameter:

```
#define GRPC_TRACE_LOG(tracer, level) \
    LOG_IF(level, GRPC_TRACE_FLAG_ENABLED(tracer))
```

Note that we do want to use `GRPC_TRACE_FLAG_ENABLED()` here instead of directly calling `tracer.enabled()`, as per the comment on line 68.

src/core/lib/debug/trace.h
```diff
@@ -1,6 +1,4 @@
-//
-//
-// Copyright 2015 gRPC authors.
+// Copyright 2024 gRPC authors.
```
markdroth, line 1: We shouldn't change the copyright date here, since this file was pre-existing.

tools/codegen/core/templates/trace.h.mako
```diff
@@ -0,0 +1,135 @@
...
+#include "absl/strings/string_view.h"
+
+#include <grpc/support/port_platform.h>
+
+void grpc_tracer_init();
+void grpc_tracer_shutdown(void);
+
+namespace grpc_core {
+
+class TraceFlag;
```
markdroth, line 36: This file contains both hand-maintained code and generated code.  How about splitting those into two separate files?

src/core/lib/debug/trace.h
```diff
@@ -111,6 +112,472 @@ class SavedTraceFlags {
   std::map<std::string, bool> values_;
 };
 
+extern TraceFlag api_trace;
+#define GRPC_API_TRACE_LOG(level) \
+  LOG_IF(level, grpc_core::api_trace.enabled()) << "(api) "
```
markdroth, line 117: Do we want to automatically log the name of the tracer in every log message?  Usually there's enough info already to tell what the log message is for.

src/core/lib/debug/trace_flags.yaml
drfloob: Autogenerating the documentation is a primary goal here, that'll come next. I left the descriptions null until a first review pass, and it looks like we'll be moving forward! I'll fill them in.

src/core/lib/debug/trace.h
drfloob: I'm flexible. A good handful of uses add that information explicitly every time already, and I think it adds some helpful context, but I can support a YAGNI argument.

markdroth: The case I was thinking of is for things like the tracer argument passed to classes like `RefCounted<>` and `InternallyRefCounted<>`, where we explicitly use a different label for the log messages than the name of the tracer, because a single tracer enables logging for multiple objects.  It's not necessary to log the name of the tracer in those cases.

src/core/lib/debug/trace_flags.yaml
drfloob: Docs are now autogenerated. Details:
* some flags have no meaningful description.
* I added an "internal" metadata bit for things like `xds_cluster_resource_type_test`, which shouldn't show up in the docs.

drfloob (commented):
Thanks for the reviews!

src/core/client_channel/client_channel_filter.cc
```diff
@@ -3406,13 +3308,12 @@ class ClientChannelFilter::FilterBasedLoadBalancedCall::LbQueuedCallCanceller
...
-        gpr_log(GPR_INFO,
-                "chand=%p lb_call=%p: cancelling queued pick: "
-                "error=%s self=%p calld->pick_canceller=%p",
-                chand, lb_call, StatusToString(error).c_str(), self,
-                lb_call->lb_call_canceller_);
-      }
+      GRPC_TRACE_LOG(client_channel_lb_call, INFO)
+          << "chand=" << chand << " lb_call=" << lb_call
+          << ": cancelling queued pick: "
+             "error="
```
markdroth, line 3314: I think this can be merged with the string on the previous line.

markdroth (commented):
This looks really good!  Most of my comments are minor; the only substantive one is the one about needing to remove the existing trace flag object definitions throughout the codebase and make sure that callers are using the new ones.

src/core/lib/debug/trace_flags.yaml
```diff
@@ -0,0 +1,351 @@
...
+#  *  an `internal: true` flag can be added to ensure that this flag does not
+#     show up in the environment_variables.md documentations.
+
+api:
+  default: false
+  description: API calls to the C core.
+auth_context_refcount:
+  debug_only: true
+  default: false
+  description: null
```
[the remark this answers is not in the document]
drfloob: That's excessive. For example, adding "Refcount for the auth context" to the `auth_context_refcount` tracer is not at all useful. If more detail is needed, then probably someone who knows the auth context internals should write a description in more detail here.

If you have specific wants here, you can call them out with your requested verbiage, or follow up with another PR.

src/core/lib/debug/trace.h
```diff
@@ -108,7 +105,217 @@ class SavedTraceFlags {
...
+extern TraceFlag xds_listener_resource_type_test_trace;
+extern TraceFlag xds_override_host_lb_trace;
+extern TraceFlag xds_resolver_trace;
+extern TraceFlag xds_route_config_resource_type_test_trace;
+extern TraceFlag xds_server_config_fetcher_trace;
+extern TraceFlag xds_wrr_locality_lb_trace;
+
+const absl::flat_hash_map<std::string, TraceFlag*>* GetAllTraceFlags();
+
+constexpr const char* g_all_trace_var_names[] = {
```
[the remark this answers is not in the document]
drfloob: We don't anymore. Removed.

src/core/lib/debug/trace.h
```diff
@@ -109,382 +105,120 @@ class SavedTraceFlags {
...
 extern TraceFlag xds_server_config_fetcher_trace;
-#define GRPC_XDS_SERVER_CONFIG_FETCHER_TRACE_LOG(level)               \
-  LOG_IF(level, grpc_core::xds_server_config_fetcher_trace.enabled()) \
-      << "(xds_server_config_fetcher) "
 extern TraceFlag xds_wrr_locality_lb_trace;
-#define GRPC_XDS_WRR_LOCALITY_LB_TRACE_LOG(level)               \
-  LOG_IF(level, grpc_core::xds_wrr_locality_lb_trace.enabled()) \
-      << "(xds_wrr_locality_lb) "
+
+const absl::flat_hash_map<std::string, TraceFlag*>* GetAllTraceFlags();
```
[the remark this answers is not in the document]
drfloob: Good point, done.

tools/codegen/core/templates/trace.h.mako
drfloob: The pragma trick is nice. I've added that, separated the generated content, and we can suffer boilerplate if that ends up not working for some reason.

src/core/lib/debug/trace.h
drfloob: I had the same thought! Done.

drfloob (commented):
Thank you for the review, Mark.

src/core/resolver/xds/xds_resolver_trace.cc
markdroth, line 1: This file can be deleted.

src/core/resolver/xds/xds_resolver_trace.h
markdroth, line 1: This file can be deleted.

src/core/resolver/resolver.cc
```diff
@@ -20,8 +20,8 @@
 
 #include <grpc/support/port_platform.h>
 
-grpc_core::DebugOnlyTraceFlag grpc_trace_resolver_refcount(false,
-                                                           "resolver_refcount");
+grpc_core::DebugOnlyTraceFlag resolver_refcount_trace(false,
```
markdroth, line 23: Delete.

test/core/test_util/BUILD
```diff
@@ -534,8 +534,23 @@ grpc_cc_library(
...
+    deps = [
+        "grpc_test_util",
+        "grpc_test_util_base",
+        "//:gpr",
+        "//src/core:useful",
+    ],
+)
+
+grpc_cc_test(
+    name = "glob_test",
```
markdroth, line 552: Similarly, this test belongs in test/core/gprpp.

test/core/test_util/glob_test.cc
```diff
@@ -0,0 +1,85 @@
...
+#include "src/core/lib/gprpp/glob.h"
+
+#include "absl/strings/match.h"
+#include "gtest/gtest.h"
+
+#include <grpc/grpc.h>
+
+#include "test/core/test_util/test_config.h"
+
+class GlobTest : public ::testing::Test {};
```
markdroth, line 24: If we don't need to put anything in the test class, then we don't need the test class at all.  Instead, just change all of the `TEST_F` calls below to just `TEST`.

test/core/test_util/trace_flags_test.cc
```diff
@@ -20,71 +20,71 @@
 #include "src/core/lib/debug/trace.h"
 #include "test/core/test_util/test_config.h"
 
-class TraceFlagTest : public ::testing::Test {
+class GlobTest : public ::testing::Test {
```
markdroth, line 23: This is testing trace flags, not glob.

markdroth (commented):
Thanks for doing this!

src/core/lib/debug/trace_flags.cc
```diff
@@ -130,9 +131,10 @@ TraceFlag xds_route_config_resource_type_test_trace(
 TraceFlag xds_server_config_fetcher_trace(false, "xds_server_config_fetcher");
 TraceFlag xds_wrr_locality_lb_trace(false, "xds_wrr_locality_lb");
 
-const absl::flat_hash_map<std::string, TraceFlag*>* GetAllTraceFlags() {
+const absl::flat_hash_map<std::string, TraceFlag*>& GetAllTraceFlags() {
   static const absl::flat_hash_map<std::string, TraceFlag*> all({
```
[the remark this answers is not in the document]
drfloob: Good catch. Fixed.

drfloob (commented):
I believe I've addressed all the things. Thanks!

src/core/lib/debug/trace_flags.yaml
```diff
@@ -353,3 +353,11 @@ xds_server_config_fetcher:
 xds_wrr_locality_lb:
   default: false
   description: XDS WRR locality LB policy.
+fuzzing_ee_writes:
```
markdroth, line 356: Should these two be in their proper place in alphabetic order?

src/core/lib/resource_quota/trace.cc
markdroth, line 1: I think this file can be removed now.

src/core/lib/slice/slice_refcount.cc
markdroth, line 1: I think this file can be deleted now.

src/core/lib/debug/trace_flags.yaml
```diff
@@ -192,27 +192,27 @@ inproc:
...
-  description: null
+  description: LB policy refcounting.
 metadata_query:
   default: false
-  description: null
+  description: GCP metadata queries.
 op_failure:
   default: false
-  description: Error information when failure is pushed onto a completion queue.
+  description: Error information when failure is pushed onto a completion queue. API tracing must be enabled for this flag to have any effect.
```
markdroth, line 201: Instead of saying "API tracing", maybe say "The 'api' tracer", to make it clearer what this means?

Same thing on lines 215 and 239.

test/core/test_util/BUILD
markdroth: Looks like you moved the glob test to test/core/debug instead of test/core/gprpp.

markdroth (commented):
This is looking really good!  Just a few minor comments remaining.

@tanvi-jagtap, once this gets merged, you may want to update your absl logging migration PRs to use the new `GRPC_TRACE_LOG()` macro being introduced here.

drfloob (commented):
Pretty close now!

test/core/handshake/client_ssl.cc
```diff
@@ -175,11 +173,10 @@ static int alpn_select_cb(SSL* /*ssl*/, const uint8_t** out, uint8_t* out_len,
 
 static void ssl_log_where_info(const SSL* ssl, int where, int flag,
                                const char* msg) {
-  if ((where & flag) &&
-      GRPC_TRACE_FLAG_ENABLED(client_ssl_tsi_tracing_enabled)) {
-    LOG(INFO) << absl::StrFormat("%20.20s - %30.30s  - %5.10s", msg,
-                                 SSL_state_string_long(ssl),
-                                 SSL_state_string(ssl));
+  if ((where & flag)) {
```
markdroth, line 176: Nit: Don't need the inner parens here.

markdroth (approved):
This looks fantastic!

markdroth (approved):

tanvi-jagtap:
@drfloob : Please can you let me know approximately when this PR will be submitted. I will avoid editing the files in the PR for some time to avoid more merge conflicts for both of us. 
In the meantime I can finish all other files that need gpr to absl migration.

test/core/handshake/client_ssl.cc
```diff
@@ -175,11 +173,10 @@ static int alpn_select_cb(SSL* /*ssl*/, const uint8_t** out, uint8_t* out_len,
 
 static void ssl_log_where_info(const SSL* ssl, int where, int flag,
                                const char* msg) {
-  if ((where & flag) &&
-      GRPC_TRACE_FLAG_ENABLED(client_ssl_tsi_tracing_enabled)) {
-    LOG(INFO) << absl::StrFormat("%20.20s - %30.30s  - %5.10s", msg,
-                                 SSL_state_string_long(ssl),
-                                 SSL_state_string(ssl));
+  if (where & flag) {
```
ctiller, line 176: I'd prefer `(where & flag) != 0` here, for the same reasons we usually write `!= nullptr` for pointers -- this is an int not a bool.

markdroth: +1 -- but don't need the parens.

drfloob:
> @drfloob : Please can you let me know approximately when this PR will be submitted. I will avoid editing the files in the PR for some time to avoid more merge conflicts for both of us. In the meantime I can finish all other files that need gpr to absl migration.

@tanvi-jagtap Thanks. I'll try to wrap it up this week, I have a bit of codegen work to do.

test/core/handshake/client_ssl.cc
drfloob: Fixed. It does require parens though, `!=` has higher precedence than `&`. http://go/cppref/cpp/language/operator_precedence

PR #36582, opened 2024-05-10
[channelz] modernize channel trace test

markdroth:
This cleans up the channel trace test to provide better output when tests fail, which will help diagnose the problem with b/339066599.

### DIFF
--- a/test/core/channelz/channel_trace_test.cc
+++ b/test/core/channelz/channel_trace_test.cc
@@ -163,7 +163,7 @@ MATCHER_P(IsEmptyChannelTrace, num_events_logged_expected,
 
 void ValidateJsonProtoTranslation(const Json& json) {
   std::string json_str = JsonDump(json);
-  grpc::testing::ValidateChannelTraceProtoJsonTranslation(json_str.c_str());
+  grpc::testing::ValidateChannelTraceProtoJsonTranslation(json_str);
 }
 
 }  // anonymous namespace
--- a/test/core/channelz/channel_trace_test.cc
+++ b/test/core/channelz/channel_trace_test.cc
@@ -20,10 +20,12 @@
 
 #include <stdlib.h>
 
+#include <list>
 #include <string>
 #include <thread>
 
 #include "absl/synchronization/notification.h"
+#include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
 #include <grpc/credentials.h>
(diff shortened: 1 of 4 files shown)


test/cpp/util/channel_trace_proto_helper.h
```diff
@@ -19,17 +19,23 @@
...
 
-void ValidateChannelTraceProtoJsonTranslation(const char* json_c_str);
-void ValidateChannelProtoJsonTranslation(const char* json_c_str);
-void ValidateGetTopChannelsResponseProtoJsonTranslation(const char* json_c_str);
-void ValidateGetChannelResponseProtoJsonTranslation(const char* json_c_str);
-void ValidateGetServerResponseProtoJsonTranslation(const char* json_c_str);
-void ValidateSubchannelProtoJsonTranslation(const char* json_c_str);
-void ValidateServerProtoJsonTranslation(const char* json_c_str);
-void ValidateGetServersResponseProtoJsonTranslation(const char* json_c_str);
+void ValidateChannelTraceProtoJsonTranslation(const std::string& json_string);
```
drfloob, line 27: Why not use absl::string_view? Protobuf's `JsonStringToMessage` takes a string_view, so I think it should be a safe change.

test/core/channelz/channel_trace_test.cc
```diff
@@ -43,82 +45,126 @@ namespace grpc_core {
...
+      IsJsonObject(::testing::ElementsAre(
+          ::testing::Pair("creationTimestamp", IsJsonString(::testing::_)),
+          ::testing::Pair("numEventsLogged",
+                          IsJsonStringNumber(num_events_logged_expected)))),
+      arg, result_listener);
+}
+
+void ValidateJsonProtoTranslation(const Json& json) {
+  std::string json_str = JsonDump(json);
+  grpc::testing::ValidateChannelTraceProtoJsonTranslation(json_str.c_str());
```
drfloob, line 166: `.c_str()` is not needed.

drfloob (commented):
Mostly looks good! It's much more readable now. I left two suggestions.

drfloob (approved):

test/cpp/util/channel_trace_proto_helper.h
markdroth: Thanks for pointing this out!  I somehow thought that none of the protobuf APIs used absl yet, so I assumed that this needed to be `std::string`.

Done.

test/core/channelz/channel_trace_test.cc
markdroth: Done.

PR #36583, opened 2024-05-11
[TEST][DO NOT SUBMIT] Upgrade cares to 1.20.1

yijiem:


yijiem:
This version of cares has a known issue which is breaking gRPC: https://github.com/c-ares/c-ares/issues/571. If we ever decide to upgrade, we should skip this version (the issue is fixed in 1.21.0).

PR #36608, opened 2024-05-14

tanvi-jagtap:

We have the following mapping

Reviewers need to check :


test/core/network_benchmarks/low_level_ping_pong.cc
```diff
@@ -477,7 +473,7 @@ static int create_sockets_tcp(fd_pair* client_fds, fd_pair* server_fds) {
 static int create_sockets_socketpair(fd_pair* client_fds, fd_pair* server_fds) {
   int fds[2];
   if (socketpair(AF_UNIX, SOCK_STREAM, 0, fds) < 0) {
-    gpr_log(GPR_ERROR, "socketpair: %s", grpc_core::StrError(errno).c_str());
+    LOG(ERROR) << "socketpair: " << grpc_core::StrError(errno).c_str();
```
JuneYan, line 476: I think the conversion is no longer needed here (same as other places in this file)?

tanvi-jagtap: fixed. Thanks for the catch

JuneYan (approved):

PR #36611, opened 2024-05-14
[lb] Fix wrong memory allocation size from strlen for string copying

### LINKED ISSUE #36610: Wrong memory allocation size for string copy at load_balancer_api.cc
Hi,
I found a wrong memory allocation during a string copy with Svace static analyzer.

During a string copy in `GrpcLbLoadReportRequestCreate()` the result of `strlen` is used to allocate memory, but it is not count a null terminating symbol:
https://github.com/grpc/grpc/blob/15850972ddba9c1262a9d51341da03bc607bd934/src/core/load_balancing/grpclb/load_balancer_api.cc#L107-L109

apach301:
Fixes #36610


markdroth (approved):
Thanks for the fix!

eugeneo (approved):

markdroth:
Looks like the existing code was actually correct.  We don't want to include the null byte in the string in this case, because we're encoding the size into the protobuf.

PR #36628, opened 2024-05-15
[Test] Install g++ before gen_upb_api

veblush:
This is to address the following error error thrown when running protobuf-at-head. This is because Bazel needs a C++ compiler to run `tools/codegen/core/gen_upb_api.sh`. 

```
Auto-Configuration Error: Cannot find gcc or CC (clang); either correct your path or set the CC environment variable
```

This is not ideal as installing `build-essential` package takes time but let's fix it first and get this installed in the CI image later.


veblush:
Verified in the separate test-run.

yashykt (approved):

PR #36656, opened 2024-05-17
[grpc][Gpr_To_Absl_Logging] Changing definition of GPR_ASSERT macro

tanvi-jagtap:
[grpc][Gpr_To_Absl_Logging] Remove gpr_assertion_failed

1. Remove gpr_assertion_failed
2. Using LOG(FATAL) instead.
3. Defining GPR_ASSERT as CHECK


include/grpc/support/log.h
```diff
@@ -22,6 +22,8 @@
 #include <stdarg.h>
 #include <stdlib.h> /* for abort() */
 
+#include "absl/log/log.h"
+
```
tanvi-jagtap, line 26: Should be check.h

Issue #36663, opened 2024-05-17
Use-after-free bug on StatsPluginChannelScope


yashykt:
There's an ASAN issue on the usage of `StatsPluginChannelScope` as shown in https://github.com/googleapis/google-cloud-cpp/pull/14259/checks?check_run_id=25119432817

yashykt:
Needs to be backported to 1.63 and 1.64 as well

PR #36664, opened 2024-05-17
[StatsPlugin] Fix use-after-free issue

yashykt:
Fix https://github.com/grpc/grpc/issues/36663

### DIFF
--- a/include/grpc/support/metrics.h
+++ b/include/grpc/support/metrics.h
@@ -28,8 +28,7 @@ class StatsPluginChannelScope {
  public:
   StatsPluginChannelScope(absl::string_view target,
                           absl::string_view default_authority)
-      : target_(target),
-        default_authority_(default_authority) {}
+      : target_(target), default_authority_(default_authority) {}
 
   /// Returns the target used for creating the channel in the canonical form.
   /// (Canonicalized target definition -
--- a/include/grpc/support/metrics.h
+++ b/include/grpc/support/metrics.h
@@ -41,8 +41,8 @@ class StatsPluginChannelScope {
   // Disable copy constructor and copy-assignment operator.
   StatsPluginChannelScope(const StatsPluginChannelScope&) = delete;
   StatsPluginChannelScope& operator=(const StatsPluginChannelScope&) = delete;
-  std::string target_;
-  std::string default_authority_;
+  absl::string_view target_;
+  absl::string_view default_authority_;
 };
 
 }  // namespace experimental
--- a/include/grpc/support/metrics.h
+++ b/include/grpc/support/metrics.h
@@ -28,7 +28,8 @@ class StatsPluginChannelScope {
  public:
   StatsPluginChannelScope(absl::string_view target,
                           absl::string_view default_authority)
-      : target_(target), default_authority_(default_authority) {}
+      : target_(std::string(target)),
+        default_authority_(std::string(default_authority)) {}
 
   /// Returns the target used for creating the channel in the canonical form.
   /// (Canonicalized target definition -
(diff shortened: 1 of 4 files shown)


include/grpc/support/metrics.h
```diff
@@ -28,7 +28,8 @@ class StatsPluginChannelScope {
  public:
   StatsPluginChannelScope(absl::string_view target,
                           absl::string_view default_authority)
-      : target_(target), default_authority_(default_authority) {}
+      : target_(std::string(target)),
```
markdroth, line 31: Nit: No need to explictly say `std::string()` on these two lines, since we're already passing the parameters to `std::string` ctor by virtue of the data members now being that type.

markdroth (commented):
Can we craft a test that fails without this fix?

include/grpc/support/metrics.h
yashykt: Thanks!

yijiem (approved):

test/core/channel/metrics_test.cc
```diff
@@ -646,6 +646,19 @@ TEST_F(MetricsDeathTest, RegisterTheSameMetricNameWouldCrash) {
       "Metric name double_histogram has already been registered.");
 }
 
+// Test that we can use the target and authority after the original strings are
+// freed.
+TEST(StatsPluginChannelScopeTest, UseAfterFree) {
```
markdroth, line 651: This is kind of too low-level of a test.  What I was really asking for was a test that demonstrates the crash that prompted this fix.  In what circumstance does that happen?

Actually, this makes me realize that I really don't understand why this fix is necessary in the first place.  The ChannelScope shouldn't be used asynchronously, so the underlying strings should always exist until after they're accessed here.  In what situation is that not happening?

yashykt: https://github.com/grpc/grpc/blob/fb6a57b153c1290c08028cd909d759fcd393ff22/src/core/lib/surface/legacy_channel.cc#L96 for example. The authority is a temporary string and also where the issue exists. I could also change the invocation to make sure that `GetStatsPluginForChannel` is invoked in the same scope as target and authority, but that seems too brittle.

markdroth: I think we should absolutely fix that caller to keep the string in scope instead of changing `ChannelScope` to store copies of the strings.  We should not be using `ChannelScope` in a situation where it needs its own copy.

yashykt: I still think that this is too brittle, but the test would atleast make sure that target and authority are accessible.

src/core/lib/surface/legacy_channel.cc
```diff
@@ -93,13 +93,13 @@ absl::StatusOr<OrphanablePtr<Channel>> LegacyChannel::Create(
     *(*r)->stats_plugin_group =
         GlobalStatsPluginRegistry::GetStatsPluginsForServer(args);
   } else {
-    experimental::StatsPluginChannelScope scope(
-        target, args.GetOwnedString(GRPC_ARG_DEFAULT_AUTHORITY)
-                    .value_or(CoreConfiguration::Get()
-                                  .resolver_registry()
-                                  .GetDefaultAuthority(target)));
     *(*r)->stats_plugin_group =
```
markdroth, line 96: Suggest writing this as:

```
std::string authority =
    args.GetOwnedString(GRPC_ARG_DEFAULT_AUTHORITY)
                    .value_or(CoreConfiguration::Get()
                                  .resolver_registry()
                                  .GetDefaultAuthority(target);
*(*r)->stats_plugin_group =
    GlobalStatsPluginRegistry::GetStatsPluginsForChannel(
        experimental::StatsPluginChannelScope(target, authority));
```

markdroth (approved):

markdroth (approved):

PR #36672, opened 2024-05-20
[PSM Interop] Skip Circuit Breaking test in the legacy test driver

sergiitk:
Circuit breaking ported to the new framework: https://github.com/grpc/psm-interop/blob/main/tests/circuit_breaking_test.py.

To avoid backports, skipping it in the legacy test driver.

ref b/227678751


murgatroid99 (approved):

sergiitk:
- [ ] [grpc/core/master/linux/grpc_xds_v3](https://source.cloud.google.com/results/invocations/97af7aa8-9d98-49f8-abbb-5b1a5b545f1d)

sergiitk:
Ah, the test is useless because the old driver can't test changes until they're already merged.

PR #36703, opened 2024-05-23

tanvi-jagtap:

We have the following mapping

Reviewers need to check :


test/core/security/credentials_test.cc
```diff
@@ -1028,7 +1025,7 @@ void validate_sts_token_http_request(const grpc_http_request* request,
       absl::StrFormat("%s?%s", test_sts_endpoint_url, body);
   absl::StatusOr<URI> url = URI::Parse(get_url_equivalent);
   if (!url.ok()) {
-    gpr_log(GPR_ERROR, "%s", url.status().ToString().c_str());
+    LOG(ERROR) << url.status().ToString();
```
yousukseung, line 1028: I think we should look into using Abseil Stringfy with gRPC types if you have not.

test/core/test_util/cmdline_test.cc
```diff
@@ -20,14 +20,13 @@
...
+#include "absl/log/log.h"
 #include "gtest/gtest.h"
 
-#include <grpc/support/log.h>
-
 #include "src/core/util/useful.h"
 #include "test/core/test_util/test_config.h"
 
-#define LOG_TEST() gpr_log(GPR_INFO, "test at %s:%d", __FILE__, __LINE__)
+#define LOG_TEST() LOG(INFO) << "test at " << __FILE__ << ":" << __LINE__
```
yousukseung, line 29: Maybe unnecessary if abseil logging already adds file/line?

yousukseung (approved):

test/core/test_util/cmdline_test.cc
tanvi-jagtap: Even gpr_log added file and line . So behaviourally gpr_log and LOG are equivalent. I don't know why they did this.

test/core/security/credentials_test.cc
tanvi-jagtap: Missed this. Fixed it.

yousukseung (approved):

PR #36707, opened 2024-05-23
[Clean-up] Fix MSAN and UBSAN issues found by clang-19

veblush:
Fixed various MSAN and UBSAN issues found in an attempt to bump the clang version used for RBE. (https://github.com/grpc/grpc/pull/36685) As our xSAN tests are using RBE, it revealed a few new issues. This PR is to fix all of those.

### DIFF
--- a/src/core/lib/iomgr/ev_poll_posix.cc
+++ b/src/core/lib/iomgr/ev_poll_posix.cc
@@ -267,19 +267,29 @@ struct grpc_pollset_set {
 //
 
 static void fork_fd_list_remove_node(grpc_fork_fd_list* node) {
+  gpr_mu_lock(&fork_fd_list_mu);
+  if (fork_fd_list_head == node) {
+    fork_fd_list_head = node->next;
+  }
+  if (node->prev != nullptr) {
+    node->prev->next = node->next;
+  }
+  if (node->next != nullptr) {
+    node->next->prev = node->prev;
+  }
+  gpr_free(node);
+  gpr_mu_unlock(&fork_fd_list_mu);
+}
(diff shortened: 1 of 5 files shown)


src/core/lib/iomgr/ev_poll_posix.cc
```diff
@@ -267,19 +267,29 @@ struct grpc_pollset_set {
 //
 
 static void fork_fd_list_remove_node(grpc_fork_fd_list* node) {
+  gpr_mu_lock(&fork_fd_list_mu);
```
yashykt, line 270: this seems like a change in logic

veblush: Not really. `fork_fd_list_remove_grpc_fd` is a new replacement for the old  `fork_fd_list_remove_node` and all call sites for `fork_fd_list_remove_node` are updated to use `fork_fd_list_remove_grpc_fd`

yashykt: @drfloob / @eugeneo should review this

eugeneo: I believe
```
if (track_fds_for_fork) {
```
might be needed so there's no overhead if fork support is not necessary

veblush: ^ `fork_fd_list_remove_grpc_fd` does. No one calls `fork_fd_list_remove_node` other than `fork_fd_list_remove_grpc_fd`

yashykt (approved):

PR #36709, opened 2024-05-24
[EventEngine] Fix race between connection and its deadline timer on Windows

drfloob:


src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -43,20 +45,113 @@
...
+
+void WindowsEventEngine::ConnectionState::AbortOnConnect() {
+  on_connected_cb_.reset();
+}
+
+void WindowsEventEngine::ConnectionState::AbortDeadlineTimer() {
+  deadline_timer_cb_.reset();
+}
+
+void WindowsEventEngine::ConnectionState::OnConnectedCallback::Run() {
```
Vignesh2208, line 117: Nit: Probably code between WindowsEventEngine::ConnectionState::OnConnectedCallback::Run and WindowsEventEngine::ConnectionState::DeadlineTimerCallback::Run can be shared.

src/core/lib/event_engine/windows/windows_engine.cc
```diff
@@ -253,38 +349,60 @@ bool WindowsEventEngine::IsWorkerThread() { grpc_core::Crash("unimplemented"); }
...
+    grpc_core::MutexLock lock(&connection_state->mu());
+    cancelled = CancelConnectFromDeadlineTimer(connection_state.get());
+  }
+  if (cancelled) {
+    auto cb = connection_state->TakeCallback();
+    connection_state.reset();
+    cb(absl::DeadlineExceededError("Connection timed out"));
+    return;
+  }
+  GRPC_EVENT_ENGINE_TRACE(
```
Vignesh2208, line 400: Nit: I dont think this needs to be traced.

Vignesh2208 (approved):
LGTM. Only minor nits

src/core/lib/event_engine/windows/windows_engine.cc
drfloob: Trace removed.

src/core/lib/event_engine/windows/windows_engine.cc
drfloob: To eliminate 10 SLOC duplication, options include: introducing a shared base class with protected shared state, adding friend classes/functions to access private members of the ConnectionState, or putting the engine callback invocation in a lambda. I'm not a fan of any, I think this finite (2-count) duplication is best.

drfloob (commented):
Thanks for the review!

PR #36713, opened 2024-05-24

tanvi-jagtap:

We have the following mapping

Reviewers need to check :


src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
```diff
@@ -277,14 +274,13 @@ absl::StatusOr<int> ListenerContainerAddAllLocalAddresses(
     addr = EventEngine::ResolvedAddress(ifa_it->ifa_addr, len);
     ResolvedAddressSetPort(addr, requested_port);
     std::string addr_str = *ResolvedAddressToString(addr);
-    gpr_log(GPR_DEBUG,
-            "Adding local addr from interface %s flags 0x%x to server: %s",
-            ifa_name, ifa_it->ifa_flags, addr_str.c_str());
+    VLOG(2) << "Adding local addr from interface " << ifa_name << " flags 0x"
+            << std::hex << ifa_it->ifa_flags << " to server: " << addr_str;
     // We could have multiple interfaces with the same address (e.g.,
```
tanvi-jagtap, line 279: TODO : validate std::hex usage

sastryvp (approved):

PR #36714, opened 2024-05-24
doc: add A53 (ignore resource deletion) to xds features

dfawley:
Noticed this feature was missing from the list, so I added it.

@murgatroid99 - is this supported in Node?  Let me know what version and I'll include it.


murgatroid99:
Yes, Node added this in version 1.7.0

murgatroid99 (approved):

markdroth (approved):
Historically, we've really intended this file to be mainly for those writing their own control planes that needed to know what features they could configure clients for, so it didn't seem necessary to include A53 here, since it doesn't impact interactions with the control plane.  But I don't really object to doing so.

Note that we never added A46 or A57, for the same reason.  Feel free to add them as well if that's what we're preferring.

dfawley:
There was a question that came up about this feature and when/if it was implemented, and I noticed it was absent from this list.  I like having all these things in one place to consolidate, but if we say we should limit this as above, then we should instead add info about when Go and Node implemented this in the gRFC (and also add when _all_ languages implemented A46/A57 in those gRFCs as well).

markdroth (approved):

PR #36715, opened 2024-05-24
[build] rename android platforms to match CPU arch

jdcormie:
Work around for https://github.com/bazelbuild/bazel/issues/15837#issuecomment-2127758141

Fixes a problem where the example client and server APKs would build but could not be installed (`INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries`)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -66,30 +66,6 @@ bool_flag(
     build_setting_default = False,
 )
 
-platform(
-    name = "android_x86_64",
-    constraint_values = [
-        "@platforms//os:android",
-        "@platforms//cpu:x86_64",
-    ],
-)
-
-platform(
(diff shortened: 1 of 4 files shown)


jdcormie:
> Thanks!
> 
> 1. I think we should move these platform declaration to subfolder if we want to remove `android_` prefix, maybe `bazel/platforms/android/BUILD`. (In google's internal repo we declare them at //.../platforms/android/BUILD)
> 2. Change PR and commit message title to `[build] Rename android platforms to match CPU arch`

done

sifmelcara (approved):
Some tests are still failing after rerun, but the failure doesn't look like relating to this change

PR #36726, opened 2024-05-24
[xDS] fix "tls" channel cred in bootstrap to actually work

markdroth:
This fixes a fairly embarrassing bug and lack of testing from #33234.  Prior to this fix, attempting to use the "tls" creds type would always cause a crash.

@gtcooke94 @matthewstevenson88 Note that the root cause of this bug was that when I wrote this code, I assumed that `grpc_tls_credentials_options` had a reasonable default for the cert verifier.  But it turns out that it doesn't do that directly; instead, we are only imposing that default in [`CredentialOptionSanityCheck()`](https://github.com/grpc/grpc/blob/621aa4e5ce5a1ac93a31439adcec00e301206072/src/core/lib/security/credentials/tls/tls_credentials.cc#L85), which is called only when we call [`grpc_tls_credentials_create()`](https://github.com/grpc/grpc/blob/621aa4e5ce5a1ac93a31439adcec00e301206072/src/core/lib/security/credentials/tls/tls_credentials.cc#L160), not when we directly instantiate `TlsCredentials` as my code was doing.  As part of the TlsCreds API cleanup you're working on, we should fix this so that callers get the right behavior even if they are internal callers that instantiate the TlsCreds object directly rather than calling the C-core API.


yashykt (approved):

yashykt:
release notes: yes?

matthewstevenson88:
Thanks for flagging this @markdroth. We will make sure this "reasonable default" issue gets fixed as part of the de-experimentalization effort. cc: @gtcooke94

PR #36729, opened 2024-05-25
[OTPlugin] Per-channel OpenTelemetry plugin

yijiem:


include/grpcpp/ext/otel_plugin.h
```diff
@@ -142,6 +152,10 @@ class OpenTelemetryPluginBuilder {
   /// Registers a global plugin that acts on all channels and servers running on
   /// the process.
   absl::Status BuildAndRegisterGlobal();
+  /// Builds an open telemetry plugin, returns the plugin object when succeeded
+  /// or an error status when failed.
+  GRPC_MUST_USE_RESULT absl::StatusOr<std::shared_ptr<OpenTelemetryPlugin>>
```
yashykt, line 157: any special reason why this returns a shared_ptr?

include/grpcpp/ext/otel_plugin.h
```diff
@@ -26,24 +26,31 @@
...
 
 class OpenTelemetryPluginOption {
  public:
   virtual ~OpenTelemetryPluginOption() = default;
 };
 
+class OpenTelemetryPlugin {
+ public:
+  // Adds this OpenTelemetryPlugin to the channel args \a args.
+  virtual void AddToChannelArguments(grpc::ChannelArguments* args) = 0;
```
yashykt, line 51: probably should not be virtual?

include/grpcpp/ext/otel_plugin.h
```diff
@@ -113,8 +120,8 @@ class OpenTelemetryPluginBuilder {
   /// If set, \a generic_method_attribute_filter is called per call with a
   /// generic method type to decide whether to record the method name or to
   /// replace it with "other". Non-generic or pre-registered methods remain
-  /// unaffected. If not set, by default, generic method names are replaced with
-  /// "other" when recording metrics.
+  /// unaffectePd. If not set, by default, generic method names are replaced
```
yashykt, line 123: please fix

yashykt:
Let's make sure we're passing this channel arg to the rls channel and the xds client channel as well

include/grpcpp/ext/otel_plugin.h
```diff
@@ -26,24 +26,31 @@
...
 class OpenTelemetryPluginBuilderImpl;
-class OpenTelemetryPlugin;
 }  // namespace internal
 
 class OpenTelemetryPluginOption {
  public:
   virtual ~OpenTelemetryPluginOption() = default;
 };
 
+class OpenTelemetryPlugin {
```
yashykt, line 48: I'm thinking - mark this as an experimental API first. After we confirm that our users are happy with this, we can stabilize it.

include/grpcpp/ext/otel_plugin.h
```diff
@@ -142,6 +149,10 @@ class OpenTelemetryPluginBuilder {
   /// Registers a global plugin that acts on all channels and servers running on
   /// the process.
   absl::Status BuildAndRegisterGlobal();
+  /// Builds an open telemetry plugin, returns the plugin object when succeeded
+  /// or an error status when failed.
+  GRPC_MUST_USE_RESULT absl::StatusOr<std::shared_ptr<OpenTelemetryPlugin>>
+  Build();
```
yashykt, line 155: I'm thinking - mark this as an experimental API first. After we confirm that our users are happy with this, we can stabilize it.

yashykt: We should also add a comment saying that you can only Build once (irrespective  of whether `Build()` is used or `BuildAndRegisterGlobal()` is used

include/grpcpp/ext/otel_plugin.h
yashykt: I was thinking - 
`grpc::OpenTelemetryPlugin` plugin be a non-abstract class with a private member `std::shared_ptr<grpc::internal::OpenTelemetry> plugin_`.

include/grpcpp/ext/otel_plugin.h
```diff
@@ -26,24 +26,31 @@
...
-class OpenTelemetryPlugin;
 }  // namespace internal
 
 class OpenTelemetryPluginOption {
  public:
   virtual ~OpenTelemetryPluginOption() = default;
 };
 
+class OpenTelemetryPlugin {
+ public:
```
yashykt, line 49: disable copy constructor

include/grpcpp/ext/otel_plugin.h
yijiem: done.

src/core/lib/surface/legacy_channel.cc
```diff
@@ -97,9 +97,23 @@ absl::StatusOr<OrphanablePtr<Channel>> LegacyChannel::Create(
...
+        GlobalStatsPluginRegistry::GetStatsPluginsForChannel(scope);
+    // Add per-channel stats plugins.
+    auto* stats_plugin_list = args.GetPointer<
+        std::shared_ptr<std::vector<std::shared_ptr<StatsPlugin>>>>(
+        GRPC_ARG_STATS_PLUGINS);
+    if (stats_plugin_list != nullptr) {
+      for (const auto& plugin : **stats_plugin_list) {
+        bool is_enabled = false;
+        std::shared_ptr<StatsPlugin::ScopeConfig> config;
+        std::tie(is_enabled, config) = plugin->IsEnabledForChannel(scope);
```
markdroth, line 111: This doesn't seem like quite the right interface here.  In this case, we are being explicitly told to use this stats plugin for this channel, so we shouldn't ask the plugin if it's okay and give it the opportunity to say no.  Instead, I think we should have a method on the plugin to get the config for the channel unconditionally.

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -1007,4 +1037,9 @@ absl::Status OpenTelemetryPluginBuilder::BuildAndRegisterGlobal() {
   return impl_->BuildAndRegisterGlobal();
 }
 
+absl::StatusOr<std::shared_ptr<grpc::OpenTelemetryPlugin>>
+OpenTelemetryPluginBuilder::Build() {
+  return impl_->Build();
```
markdroth, line 1042: Not directly related to this PR, but is there anything left stopping us from moving the functionality of `OpenTelemetryPluginBuilderImpl` directly into `OpenTelemetryPluginBuilder`, so that we can stop having this level of indirection?

test/cpp/ext/otel/otel_test_library.h
```diff
@@ -59,7 +59,7 @@ class MockMetricReader : public opentelemetry::sdk::metrics::MetricReader {
 };
 
 class OpenTelemetryPluginEnd2EndTest : public ::testing::Test {
- protected:
+ public:
```
markdroth, line 62: Why does this need to be public instead of protected?  I don't think this should be used by anything but subclasses.

Is this because of the standalone `ConfigureOTBuilder()` function you have in the .cc file?  If so, I suggest making that a static method on the class instead of a standalone function.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1969,6 +1970,92 @@ TEST(OpenTelemetryPluginMetricsEnablingDisablingTest, TestEnableDisableAPIs) {
               ::testing::UnorderedElementsAre("grpc.test.metric_3"));
 }
 
+TEST_F(OpenTelemetryPluginEnd2EndTest, RegisterMultiplePerChannelStatsPlugins) {
```
markdroth, line 1973: Please add tests for the following cases:
- The same stats plugin being added directly to two different channels.
- Stats plugins being added directly to servers.

src/core/lib/surface/legacy_channel.cc
```diff
@@ -97,9 +97,23 @@ absl::StatusOr<OrphanablePtr<Channel>> LegacyChannel::Create(
...
+        if (is_enabled) {
+          (*r)->stats_plugin_group->AddStatsPlugin(plugin, std::move(config));
```
markdroth, line 113: @yashykt, didn't we say that we would need to store a bit in the `StatsPluginGroup` indicating whether each plugin has come from the global registry or was directly added to the channel, since we'll need that information to properly propagate stats plugins to child channels?

markdroth (commented):

src/core/lib/surface/legacy_channel.cc
yashykt: We shouldn't need to, since we already have the channel arg that only has the per channel stats plugins.

include/grpcpp/ext/otel_plugin.h
yashykt: In the approach I was imagining, the returned object would have the `shared_ptr` member.

include/grpcpp/ext/otel_plugin.h
yashykt: > Yash, why is that better than the virtual method?

Yeah, we talked about this offline later, and I'm fine with the current approach as well. 
+1 to renaming `internal::OpenTelemetryPlugin` if sticking with the current approach.

test/cpp/ext/otel/otel_plugin_test.cc
yijiem: great ideas! done, ptal

test/cpp/ext/otel/otel_test_library.h
yijiem: exactly, done

include/grpcpp/ext/otel_plugin.h
yijiem: done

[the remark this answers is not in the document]
yijiem: done

yijiem (commented):
Thanks for the review!

BUILD
```diff
@@ -2988,6 +2988,7 @@ grpc_cc_library(
     ],
     language = "c++",
     deps = [
+        ":grpc++",
```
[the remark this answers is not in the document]
markdroth: Since we do need the impl to build, I think it's better to have the dep on `grpc++` than to use forward declarations.

markdroth (approved):
Looks great!  Only a few minor comments remaining.  Feel free to merge after addressing.

include/grpcpp/ext/otel_plugin.h
```diff
@@ -146,11 +152,13 @@ class OpenTelemetryPluginBuilder {
   OpenTelemetryPluginBuilder& SetChannelScopeFilter(
       absl::AnyInvocable<bool(const ChannelScope& /*scope*/) const>
           channel_scope_filter);
-  /// Registers a global plugin that acts on all channels and servers running on
-  /// the process.
+  /// Builds and registers a global plugin that acts on all channels and servers
+  /// running on the process. Note that calling BuildAndRegisterGlobal more than
+  /// once when the first call succeeded is undefined behavior.
```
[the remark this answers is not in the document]
yijiem: done

include/grpcpp/ext/otel_plugin.h
```diff
@@ -146,11 +152,13 @@ class OpenTelemetryPluginBuilder {
...
   absl::Status BuildAndRegisterGlobal();
   /// Builds an open telemetry plugin, returns the plugin object when succeeded
-  /// or an error status when failed.
+  /// or an error status when failed. Note that calling Build more than once
+  /// when the first call succeeded is undefined behavior.
```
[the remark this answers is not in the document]
yijiem: done

BUILD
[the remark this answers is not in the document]
yijiem: done

yijiem (commented):
Thanks Mark!

markdroth (approved):

include/grpcpp/ext/otel_plugin.h
yashykt: Can we mark this as experimental as well please?

src/core/lib/channel/context.h
```diff
@@ -28,9 +28,7 @@
 /// Call context is represented as an array of \a grpc_call_context_elements.
 /// This enum represents the indexes into the array, where each index
 /// contains a different type of value.
-typedef enum {
-  GRPC_CONTEXT_COUNT
-} grpc_context_index;
+typedef enum { GRPC_CONTEXT_COUNT } grpc_context_index;
```
yashykt, line 31: revert unnecessary changes

yashykt (commented):
minor comments

include/grpcpp/ext/otel_plugin.h
yijiem: done

include/grpcpp/ext/otel_plugin.h
yijiem: done

src/core/lib/channel/context.h
yijiem: done

include/grpcpp/ext/otel_plugin.h
markdroth: If this is experimental, we should probably put it in an `experimental` namespace.

yijiem: done

yashykt (approved):

Issue #36738, opened 2024-05-27
Build fails with the latest gazelle (0.37.0)


aira-xite:
### What version of gRPC and what language are you using?
1.54.1 (but on the latest 1.64.0 is still present)

### What operating system (Linux, Windows,...) and version?
Linux and MacOs

### What runtime / compiler are you using (e.g. python version or version of gcc)
bazel 6.4.0

### What did you do?
I updated to gazelle 0.37.0

### What did you expect to see?
Build with no errors

### What did you see instead?
Builds started to fail with 
```
failed to build tools: # github.com/bazelbuild/bazel-gazelle/cmd/fetch_repo
src/github.com/bazelbuild/bazel-gazelle/cmd/fetch_repo/go_mod_download.go:64:17: undefined: errors.Join
```
This happens because `errors.Join` was introduced in go 1.20 ([link](https://pkg.go.dev/errors@go1.20#Join)), but grpc/grpc has hard-coded go 1.18 ([link](https://github.com/grpc/grpc/blob/master/bazel/grpc_extra_deps.bzl#L55)). 

I tried to specify the version in my WORKSPACE file
```
load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains")
go_register_toolchains(version = "1.20")
```

but it conflicts with the version specified in grpc
```
    grpc_extra_deps()
File ".../external/com_github_grpc_grpc/bazel/grpc_extra_deps.bzl", line 56, column 27, in grpc_extra_deps
    go_register_toolchains(version = "1.18")
File ".../external/io_bazel_rules_go/go/private/sdk.bzl", line 713, column 13, in go_register_toolchains
    fail("go_register_toolchains: version set after go sdk rule declared ({})".format(", ".join([r["name"] for r in sdk_rules])))
Error in fail: go_register_toolchains: version set after go sdk rule declared (go_sdk)
```

Could you please update the go version and/or make non hard-coded, so grpc can be used with the latest gazelle?

drfloob:
I've updated our dependencies and backported the change to 1.64, it should be out in a 1.64.1 patch release ~soon.

PR #36745, opened 2024-05-28
[build] Update Gazelle bazel dependency

drfloob:


drfloob:
Fixes #36738

veblush (approved):
LGTM. Just make sure that it works with both Bazel 6.x and 7.x as we need to support Bazel 6.x for the next 6 months.

bazel/grpc_extra_deps.bzl
```diff
@@ -52,7 +52,7 @@ def grpc_extra_deps(ignore_version_differences = False):
     api_dependencies()
 
     go_rules_dependencies()
-    go_register_toolchains(version = "1.18")
+    go_register_toolchains(version = "1.20")
```
yashykt, line 55: let's try using "host" here

drfloob:
> LGTM. Just make sure that it works with both Bazel 6.x and 7.x as we need to support Bazel 6.x for the next 6 months.

Our tests would catch this, yes? If not, we need a bazel 6 test.

veblush:
We do have Bazel 6 test so you'll know ;)

bazel/grpc_extra_deps.bzl
drfloob: Done, let's see.

PR #36757, opened 2024-05-29
[StatsPlugin] Add API to check if an instrument is enabled

yashykt:


test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1281,7 +1282,17 @@ TEST_F(OpenTelemetryPluginOptionEnd2EndTest,
...
-using OpenTelemetryPluginNPCMetricsTest = OpenTelemetryPluginEnd2EndTest;
+class OpenTelemetryPluginNPCMetricsTest
+    : public OpenTelemetryPluginEnd2EndTest {
+ protected:
+  void TearDown() override {
+    // We are tearing down OpenTelemetryPluginEnd2EndTest first to ensure that
+    // gRPC has shutdown before we reset the instruments registry.
+    OpenTelemetryPluginEnd2EndTest::TearDown();
+    grpc_core::GlobalInstrumentsRegistryTestPeer::
+        ResetGlobalInstrumentsRegistry();
```
yijiem, line 1293: Thanks for fixing this! Though this means we need to keep the `OpenTelemetryPluginNPCMetricsTest`s to be the last to run.

yijiem (commented):
Thanks Yash!

src/core/telemetry/metrics.h
```diff
@@ -329,6 +329,9 @@ class StatsPlugin {
   // Removes a callback previously added via AddCallback().  The stats
   // plugin may not use the callback after this method returns.
   virtual void RemoveCallback(RegisteredMetricCallback* callback) = 0;
+  // Returns true if instrument \a handle is enabled.
+  virtual bool IsInstrumentEnabled(
```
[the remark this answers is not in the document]
markdroth: +1

markdroth (approved):

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: Only if the OTel plugin tests the globally registered metrics.. which we probably shouldn't, or if we need to, we can register them again.

markdroth (approved):

PR #36759, opened 2024-05-29
[benchmark] Disable benchmarks for dotnet in CI.

paulosjca:
Build is failing, so temporarily removing dotnet from benchmarks CI.


gnossen (approved):

ctiller:
why are we commenting out code instead of just deleting it?

PR #36785, opened 2024-05-31
[ObjC] default GRPC_IOS_EVENT_ENGINE_CLIENT to 1

HannahShiSFB:
It's still possible to define GRPC_IOS_EVENT_ENGINE_CLIENT=0 to disable event engine for iOS.


src/core/lib/channel/context.h
```diff
@@ -28,9 +28,7 @@
 /// Call context is represented as an array of \a grpc_call_context_elements.
 /// This enum represents the indexes into the array, where each index
 /// contains a different type of value.
-typedef enum {
-  GRPC_CONTEXT_COUNT
-} grpc_context_index;
+typedef enum { GRPC_CONTEXT_COUNT } grpc_context_index;
```
sampajano, line 31: nit: is this change necessary to include in this PR? If not could you dis-include it?

sampajano (commented):
Oh wow!!! Wonderful!! Thank you so much for this change, Hannah!! 😊

Just TRIPLE check — You've verified both the effectiveness of this change in both 1) testing logs, and 2) local apps. Right? 😃

HannahShiSFB:
"Basic Tests C/C++ MacOS" is not running, probably because it was renamed in #36631

Windows build broken is no related

src/core/lib/channel/context.h
HannahShiSFB: It's for sanity test to pass, will rebase.

HannahShiSFB:
I've verified the test logs. 
Not sure what do you mean by "2) local apps"?

sampajano:
> "Basic Tests C/C++ MacOS" is not running, probably because it was renamed in #36631

Right!! I've tried to fix that one! Maybe it'll work this time!!

> I've verified the test logs.
> Not sure what do you mean by "2) local apps"?

Ok i guess that's good enough.

I was hoping you could verify using a test app as well! So we're even more sure! 😃

sampajano (approved):
Thanks so much Hannah!!! 😃

If you could also verify using a test app that'll be even better! (but i mean you know the code so well so this is probably already obvious to you.. in that case, it's not necessary. 😄)

Thanks again!! 😊

PR #36797, opened 2024-06-02
[ObjC] enable cronet test

HannahShiSFB:


HannahShiSFB:
"Basic Tests C/C++ MacOS" is not running, probably because it was renamed in #36631

PR #36798, opened 2024-06-03
[grpc][Gpr_To_Absl_Logging] Supporting verbosity flag

tanvi-jagtap:
[grpc][Gpr_To_Absl_Logging] 
1. Supporting the legacy GRPC_VERBOSITY environment variable in the new absl logging implementation. 
2. Adding a new way to disable VLOG logging.
3. Documenting the recommendations clearly.
4. Editing the init code. 

Additional Context : 
Check function gpr_default_log() for more context 
https://github.com/search?q=repo%3Agrpc%2Fgrpc%20gpr_default_log&type=code


src/core/util/log.cc
```diff
@@ -123,6 +124,33 @@ static gpr_atm parse_log_severity(absl::string_view str, gpr_atm error_value) {
...
+  if (absl::EqualsIgnoreCase(verbosity, "INFO")) {
+    LOG(WARNING) << "Not suitable for production. Prefer WARNING or ERROR. "
+                    "However if you see this message in a debug environmenmt "
+                    "or test environmenmt it is safe to ignore this message.";
+    absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfo);
+  } else if (absl::EqualsIgnoreCase(verbosity, "DEBUG")) {
+    LOG(ERROR) << "Not suitable for production. Prefer WARNING or ERROR. "
+                  "However if you see this message in a debug environmenmt or "
+                  "test environmenmt it is safe to ignore this message.";
+    absl::SetVLogLevel("*grpc*/*", 2);
```
sastryvp, line 141: Why is the magic number 2? Lets say user uses VLOG(3), it doesn't get logged right? If there is any guideline to use debug statements less than or equal to 2, it makes sense. Is it there somewhere?

src/core/util/log.cc
```diff
@@ -123,6 +124,33 @@ static gpr_atm parse_log_severity(absl::string_view str, gpr_atm error_value) {
...
+    absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfo);
+  } else if (absl::EqualsIgnoreCase(verbosity, "ERROR")) {
+    absl::SetMinLogLevel(absl::LogSeverityAtLeast::kError);
+  } else if (absl::EqualsIgnoreCase(verbosity, "NONE")) {
+    absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfinity);
+  } else if (absl::EqualsIgnoreCase(verbosity, "UNKNOWN")) {
```
sastryvp, line 147: How does gpr_log handle the verbosity level "UNKNOWN"? Whats the behavior if we don't set to any level? Logging gets disabled?

src/core/util/log.cc
```diff
@@ -123,6 +124,33 @@ static gpr_atm parse_log_severity(absl::string_view str, gpr_atm error_value) {
   return error_value;
 }
 
+void disable_vlog_for_grpc() { absl::SetVLogLevel("*grpc*/*", -1); }
```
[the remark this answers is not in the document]
tanvi-jagtap: https://yaqs.corp.google.com/eng/q/2204561495621107712#a1 
I validated it because I had the same doubt yesterday . They said this should be ok.

src/core/util/log.cc
tanvi-jagtap: This log level will never be passed to gpr_log.
This will happen only and only when no value is present for GRPC_VERBOSITY. This would be the default value.

src/core/util/log.cc
tanvi-jagtap: gpr(DEBUG , ... ) corresponds to VLOG(2) . If anytone wants to alter it from 2 to something else, rather than using DEBUG, it would be better for them to set it via absl --v etc.

src/core/util/log.cc
```diff
@@ -123,6 +124,33 @@ static gpr_atm parse_log_severity(absl::string_view str, gpr_atm error_value) {
...
+void gpr_to_absl_verbosity_setting_init(void) {
+  absl::string_view verbosity = grpc_core::ConfigVars::Get().Verbosity();
+  DVLOG(2) << "Log verbosity: " << verbosity;
+  if (absl::EqualsIgnoreCase(verbosity, "INFO")) {
+    LOG(WARNING) << "Not suitable for production. Prefer WARNING or ERROR. "
+                    "However if you see this message in a debug environmenmt "
+                    "or test environmenmt it is safe to ignore this message.";
+    absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfo);
+  } else if (absl::EqualsIgnoreCase(verbosity, "DEBUG")) {
+    LOG(ERROR) << "Not suitable for production. Prefer WARNING or ERROR. "
```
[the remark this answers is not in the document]
tanvi-jagtap: Fixed.

drfloob:
See my previous attempt here https://github.com/grpc/grpc/pull/36501. It had to be rolled back due to custom abseil sink issues around gRPC shutdown. Please run a TGP for this.

ctiller (approved):

ctiller (approved):

PR #36800, opened 2024-06-03

tanvi-jagtap:

We have the following mapping

Reviewers need to check :


src/core/lib/gprpp/posix/thd.cc
```diff
@@ -33,9 +33,8 @@
 #include <unistd.h>
 
 #include "absl/log/check.h"
-#include "absl/log/log.h"  // IWYU pragma: keep
+#include "absl/log/log.h"
```
ananda1066, line 36: Please check if IWYU can be removed here

tanvi-jagtap: I don't understand. 
I deleted "// IWYU pragma: keep" .

ananda1066: I'm asking if we need to keep the IWYU pragma - why have you removed it?

tanvi-jagtap: Oh. I had added IWYU pragma 4 weeks ago because I was working with these files in Critique and the linter was complaining. We dont need it anymore.

ananda1066 (approved):

PR #36805, opened 2024-06-04
[Fix] new clang -Wmissing-template-arg-list-after-template-kw warning

amykhuang:
Clang now requires a template argument list after the use of the template keyword. Edit this
instance to remove the template keyword since there are no template arguments.

See https://github.com/llvm/llvm-project/pull/80801.

### DIFF
--- a/src/core/lib/promise/detail/basic_seq.h
+++ b/src/core/lib/promise/detail/basic_seq.h
@@ -99,7 +99,7 @@ class BasicSeqIter {
           }
           cur_ = next;
           state_.~State();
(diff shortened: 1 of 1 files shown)


ctiller (approved):

eugeneo (approved):

amykhuang:
The patch appears to need a release-notes label, could that be added?

eugeneo:
@amykhuang added.

amykhuang:
edited change to remove the template keyword, as I don't think the previous thing was working

PR #36807, opened 2024-06-04
Remove extraneous template keyword

aeubanks:
The template keyword should only be used if we have angle brackets after the method.

Fixes code broken by upcoming clang change: https://github.com/llvm/llvm-project/pull/80801


aeubanks:
would like to add the
lang/c++
release notes: no
labels but don't have permissions

aeubanks:
obsoleted by https://github.com/grpc/grpc/pull/36805/files

PR #36817, opened 2024-06-05

tanvi-jagtap:
Replacing GPR_ASSERT with c assert.


src/ruby/ext/grpc/rb_call.c
```diff
@@ -437,7 +439,7 @@ static int grpc_rb_md_ary_fill_hash_cb(VALUE key, VALUE val, VALUE md_ary_obj) {
                  tmp_str);
         return ST_STOP;
       }
```
tanvi-jagtap, line 441: Manage the optimization part.

src/ruby/ext/grpc/rb_grpc.h
```diff
@@ -78,4 +80,10 @@ void grpc_rb_fork_unsafe_end();
 
 void grpc_ruby_init();
 
+#define GRPC_RUBY_ASSERT(x)                 \
+  if (!(x)) {                               \
+    fprintf(stderr, "%s", "Assert failed"); \
```
apolcyn, line 85: It would be useful to keep file and line number in particular, can we do something like this:

`fprintf(stderr, "%s:%d assert failed\n", __FILE__, __LINE__);`

?

tanvi-jagtap: Done. @apolcyn Thanks for the really good catch !! 🙂

apolcyn (approved):

PR #36831, opened 2024-06-06
[build] Support clang-cl for Windows builds (best effort)

drfloob:
Usage: `bazel build --config=clang-cl --build_tag_filters=-no_windows :all`

The highlight, for me anyway: thread safety annotation checking can now be done on Windows-only code.

clang-cl's interpretation of `-Wall` differs from that of clang on linux. This PR uses the set of warnings enabled on linux, and whittles down the list until all builds pass. I left comments in `copts.bzl` describing the warnings, many of which can be removed with targeted code cleanups.


third_party/protobuf.10007.patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/build_defs/BUILD.bazel b/build_defs/BUILD.bazel
```
veblush, line 1: We're trying to remove all existing patch files making it harder to upgrade the dependency. Is it feasible to remove it with help of Protobuf team?

bazel/copts.bzl
```diff
@@ -57,7 +57,104 @@ GRPC_LLVM_WARNING_FLAGS = [
     "-Wno-unused-function",
 ]
 
+GRPC_LLVM_WINDOWS_WARNING_FLAGS = GRPC_LLVM_WARNING_FLAGS + [
```
veblush, line 60: I guess those are necessarily added but wonder why windows happened to have to have all those exceptions despite the fact that both Linux and Windows use the same clang?

third_party/protobuf.10007.patch
drfloob: This patch is pulled from https://github.com/protocolbuffers/protobuf/issues/10007. The code is in protobuf master branch now, we can delete this patch after the next protobuf release.

bazel/copts.bzl
drfloob: clang-cl aims to be MSVC-compatible, and I think `clang-cl -Wall` is translated roughly to `clang -Weverything`, because of the way MSVC's `cl.exe /Wall` works.

https://clang.llvm.org/docs/UsersManual.html#clang-cl:~:text=/Wall%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20Enable%20%2DWeverything

We could structure these lists differently, but I'm not sure it'd be much better since the diagnostic flags are not identical between clang-cl and clang.

third_party/protobuf.10007.patch
veblush: Great then we can keep this temporarily as we're about to upgrade protobuf to v27 soon.

bazel/copts.bzl
veblush: I see. Thank you for the context. I don't worry too much about this lengthy list as long as we don't have to massage this often over time when changing the version of msvc or clang on Windows.

veblush (approved):

PR #36835, opened 2024-06-06
Fix race due to use of TryLock in grpc_event_engine::experimental::TimerList::TimerCheck

scwhittle:
TryLock returns true only with high-probility when the lock is held [link](https://github.com/abseil/abseil-cpp/blob/master/absl/synchronization/mutex.h#L191), because it uses relaxed load for initial check if unlocked. If it spuriously returned
false when the lock was not held, it was possible that there was a queued timer which was not
detected and the next wakeup time was left to InfiniteFuture.
This could lead to delays in firing the timer if no other timers were set to kick the TimerListHost.

Instead of using TryLock just use a regular Lock.  If there are multiple threads calling TimerCheck this could lead to some contention, but AFAICT such threads woudl eventually just block anyway waiting to be kicked so I don't think it would matter.  If there are concerns it could perhaps be switched to an atomic-based method for single execution similar to the spinlock used in [the generic run_some_expired_timers function](https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/timer_generic.cc#L590)

### DIFF
--- a/src/core/lib/event_engine/posix_engine/timer.cc
+++ b/src/core/lib/event_engine/posix_engine/timer.cc
@@ -298,12 +298,10 @@ TimerList::TimerCheck(grpc_core::Timestamp* next) {
     return std::vector<experimental::EventEngine::Closure*>();
   }
 
-  if (!checker_mu_.TryLock()) return absl::nullopt;
-  std::vector<experimental::EventEngine::Closure*> run =
-      FindExpiredTimers(now, next);
-  checker_mu_.Unlock();
-
-  return std::move(run);
+  {
+    absl::MutexLock lock(&checker_mu_);
+    return FindExpiredTimers(now, next);
+  }
 }
 
 }  // namespace experimental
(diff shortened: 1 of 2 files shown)


scwhittle:
Fixed accidental inclusion of unrelated directories in commit.

drfloob:
Thanks for pointing this out. This code can use a refactoring of a different sort. As it sits today, only one thread will ever call `TimerCheck`, and the caller will crash the process if `TryLock` returns false. https://github.com/grpc/grpc/blob/73d016e39973da726b43ba31a5e86a50e81f6815/src/core/lib/event_engine/posix_engine/timer_manager.cc#L68-L71

TryLock _could_ still spuriously returns false and cause problems here, as you point out. But I wonder if a better solution is to state that the TimerList is not thread-safe, and treat it as such.

I'm curious, did you discover this issue due to some kind of failure you witnessed?

CC @Vignesh2208 @ctiller

PR #36837, opened 2024-06-06
Explicitly check if candidate is not None when closing aio channel

mboutet:
When using gRPC client alongside kopf (Python framework for writing K8s operators), the existing code fails because of https://github.com/nolar/kopf/blob/9243c8fbb2a36c1b7cad1f65a647d3b9876a5b17/kopf/_cogs/aiokits/aiotoggles.py#L40. This is because the existing `if candidate:` condition invokes the `__bool__` magic method of the `Toggle` class from kopf.

### DIFF
--- a/src/python/grpcio/grpc/aio/_channel.py
+++ b/src/python/grpcio/grpc/aio/_channel.py
@@ -419,7 +419,10 @@ class Channel(_base_channel.Channel):
             # Locate ones created by `aio.Call`.
             frame = stack[0]
             candidate = frame.f_locals.get("self")
-            if candidate is not None:  # https://github.com/grpc/grpc/pull/36837
+            # Explicitly check for a non-null candidate instead of the more pythonic 'if candidate:'
+            # because doing 'if candidate:' assumes that the coroutine implements '__bool__' which
+            # might not always be the case.
+            if candidate is not None:
                 if isinstance(candidate, _base_call.Call):
                     if hasattr(candidate, "_channel"):
                         # For intercepted Call object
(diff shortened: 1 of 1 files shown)


src/python/grpcio/grpc/aio/_channel.py
```diff
@@ -419,7 +419,7 @@ async def _close(self, grace):  # pylint: disable=too-many-branches
             # Locate ones created by `aio.Call`.
             frame = stack[0]
             candidate = frame.f_locals.get("self")
-            if candidate:
+            if candidate is not None:  # https://github.com/grpc/grpc/pull/36837
```
XuanWang-Amos, line 422: I don't think this comment is useful, please remove it or replace it with a helpful comment.

mboutet: @XuanWang-Amos, the goal is to avoid a regression if someone in the future reverts to `if candidate:` because it is "more pythonic". By linking to this PR, I would expect any contributor to not blindly do so and go see the PR to have more context.

I'll look into adding a unit test for it instead.

XuanWang-Amos: I agree we should have some comments here, but rather than a link, let's use a sentence or two to explain why we need `if candidate is not None:` instead of just `if candidate:`.

mboutet: @XuanWang-Amos, let me know what you think of the new comment.

XuanWang-Amos (approved):

PR #36841, opened 2024-06-06
[xds] Do not crash when XdsLocalityName is null

eugeneo:


yashykt (approved):

markdroth:
In what circumstance is the locality name null here?

eugeneo:
Not clear... Happens rarely in PSM Interop testing.

markdroth:
I don't think it should ever happen.  If it is, then there's probably some deeper bug here, and this PR is just working around the symptom without solving the underlying problem.

eugeneo:
I assumed it was fine since the rest of the code could survive the null. I will reopen the PSM bug but lower its priority.

PR #36842, opened 2024-06-06
[Example] C++ Auth/SSL example

veblush:
Adding a new SSL example.


examples/cpp/auth/ssl_client.cc
```diff
@@ -0,0 +1,131 @@
...
+
+#include <grpcpp/grpcpp.h>
+
+#ifdef BAZEL_BUILD
+#include "examples/protos/helloworld.grpc.pb.h"
+#else
+#include "helloworld.grpc.pb.h"
+#endif
+
+ABSL_FLAG(std::string, target, "localhost:50051", "Server address");
```
yijiem, line 38: I think the hostname of the target has to be `localhost` since that is the name in the server's certificate. Maybe we only allow the client to specify the port number?

yijiem (approved):
Thanks Esun!

examples/cpp/auth/ssl_client.cc
veblush: That's a good point. Only localhost works.

examples/cpp/auth/ssl_server.cc
```diff
@@ -0,0 +1,104 @@
...
+    ServerUnaryReactor* reactor = context->DefaultReactor();
+    reactor->Finish(Status::OK);
+    return reactor;
+  }
+};
+
+constexpr char kServerCertPath[] = "credentials/localhost.crt";
+constexpr char kServerKeyPath[] = "credentials/localhost.key";
+
+std::string LoadStringFromFile(std::string path) {
```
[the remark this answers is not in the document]
veblush: Done.

PR #36857, opened 2024-06-07
[Security] add knob for deny_undetermined

gtcooke94:
This has been a TODO in the CRL Provider stack for some time.
If a CRL for a given issuer is _not_ found in a provider, there resulting revocation status for _any_ certificate from that issuer will be `Undetermined`. This flag allows a user to treat `Undetermined` as `Revoked`. The default behavior is to treat `Undetermined` as `Unrevoked`.

### DIFF
--- a/include/grpc/grpc_crl_provider.h
+++ b/include/grpc/grpc_crl_provider.h
@@ -68,9 +68,7 @@ class CrlProvider {
   // Not finding a CRL is a specific behavior. Per RFC5280, not having a CRL to
   // check for a given certificate means that we cannot know for certain if the
   // status is Revoked or Unrevoked and instead is Undetermined. How a user
-  // handles an Undetermined CRL is up to them. We use absl::IsNotFound as an
-  // analogue for not finding the Crl from the provider, thus the certificate in
-  // question is Undetermined.
+  // handles an Undetermined CRL is up to them.
   // Defaults to treating it as `Unrevoked`.
   bool DenyUndetermined() { return deny_undetermined_; };
 
--- a/include/grpc/grpc_crl_provider.h
+++ b/include/grpc/grpc_crl_provider.h
@@ -63,10 +63,14 @@ class CrlProvider {
   // Get the CRL associated with a certificate. Read-only.
   virtual std::shared_ptr<Crl> GetCrl(
       const CertificateInfo& certificate_info) = 0;
+  bool DenyUndetermined() { return deny_undetermined_; };
+
+ protected:
+  bool deny_undetermined_ = false;
 };
 
 absl::StatusOr<std::shared_ptr<CrlProvider>> CreateStaticCrlProvider(
-    absl::Span<const std::string> crls);
+    absl::Span<const std::string> crls, bool deny_undetermined = false);
 
 // Creates a CRL Provider that periodically and asynchronously reloads a
 // directory. The refresh_duration minimum is 60 seconds. The
@@ -77,7 +81,8 @@ absl::StatusOr<std::shared_ptr<CrlProvider>> CreateStaticCrlProvider(
 // your monitoring and alerting setup.
 absl::StatusOr<std::shared_ptr<CrlProvider>> CreateDirectoryReloaderCrlProvider(
     absl::string_view directory, std::chrono::seconds refresh_duration,
-    std::function<void(absl::Status)> reload_error_callback);
+    std::function<void(absl::Status)> reload_error_callback,
+    bool deny_undetermined = false);
 
 }  // namespace experimental
 }  // namespace grpc_core
--- a/include/grpc/grpc_crl_provider.h
+++ b/include/grpc/grpc_crl_provider.h
@@ -65,6 +65,12 @@ class CrlProvider {
       const CertificateInfo& certificate_info) = 0;
   // Returns whether a revocation status of `Undetermined` should be treated as
   // `Revoked` (deny connections) or `Unrevoked` (allowing connections).
+  // Not finding a CRL is a specific behavior. Per RFC5280, not having a CRL to
+  // check for a given certificate means that we cannot know for certain if the
+  // status is Revoked or Unrevoked and instead is Undetermined. How a user
+  // handles an Undetermined CRL is up to them. We use absl::IsNotFound as an
+  // analogue for not finding the Crl from the provider, thus the certificate in
+  // question is Undetermined.
   // Defaults to treating it as `Unrevoked`.
   bool DenyUndetermined() { return deny_undetermined_; };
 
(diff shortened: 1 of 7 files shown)


include/grpc/grpc_crl_provider.h
```diff
@@ -63,10 +63,21 @@ class CrlProvider {
   // Get the CRL associated with a certificate. Read-only.
   virtual std::shared_ptr<Crl> GetCrl(
       const CertificateInfo& certificate_info) = 0;
+  // Returns whether a revocation status of `Undetermined` should be treated as
+  // `Revoked` (deny connections) or `Unrevoked` (allowing connections).
+  // Not finding a CRL is a specific behavior. Per RFC5280, not having a CRL to
```
matthewstevenson88, line 68: When reading this paragraph, I found "Not finding a CRL is a specific behavior." confusing until I read the rest. Should we consider removing this sentence, since the same idea is repeated in more detail in the next sentence?

include/grpc/grpc_crl_provider.h
```diff
@@ -77,7 +88,8 @@ absl::StatusOr<std::shared_ptr<CrlProvider>> CreateStaticCrlProvider(
 // your monitoring and alerting setup.
 absl::StatusOr<std::shared_ptr<CrlProvider>> CreateDirectoryReloaderCrlProvider(
     absl::string_view directory, std::chrono::seconds refresh_duration,
-    std::function<void(absl::Status)> reload_error_callback);
+    std::function<void(absl::Status)> reload_error_callback,
```
matthewstevenson88, line 91: This is for a follow-up PR, just noting it now because I happened to notice it: should this be `absl::AnyInvocable` instead of `std::function`?

src/core/tsi/ssl_transport_security.cc
```diff
@@ -1076,8 +1076,9 @@ static int CheckCertRevocation(grpc_core::experimental::CrlProvider* provider,
   // analogue for not finding the Crl from the provider, thus the certificate in
   // question is Undetermined.
   if (absl::IsNotFound(crl.status())) {
-    // TODO(gtcooke94) knob for undetermined being revoked or unrevoked. By
-    // default, unrevoked.
+    if (provider->DenyUndetermined()) {
+      return 0;
```
matthewstevenson88, line 1080: In this case (and in the `else if` below): do we need to free `*crl`?

test/core/tsi/crl_ssl_transport_security_test.cc
```diff
@@ -469,6 +469,40 @@ TEST_P(CrlSslTransportSecurityTest, CrlFromBadCa) {
                                         *provider, false, false, false);
   fixture->Run();
 }
+TEST_P(CrlSslTransportSecurityTest,
+       CrlProviderGoodCertMissingCrlDenyUndetermined) {
+  // Create a provider with no CRLs
+  absl::StatusOr<std::shared_ptr<grpc_core::experimental::CrlProvider>>
+      provider = grpc_core::experimental::CreateStaticCrlProvider({}, true);
```
matthewstevenson88, line 476: nit: Please add /*...=*/ inline comments for the variables where we are passing in a literal. Here and throughout the 2 newly-added tests.

test/cpp/end2end/crl_provider_test.cc
```diff
@@ -78,6 +78,14 @@ TEST(DirectoryReloaderCrlProviderTestNoFixture, Construction) {
   auto provider = grpc_core::experimental::CreateDirectoryReloaderCrlProvider(
       kCrlDirectoryPath, std::chrono::seconds(60), nullptr);
   ASSERT_TRUE(provider.ok()) << provider.status();
+  EXPECT_FALSE((*provider)->DenyUndetermined());
+}
+
+TEST(DirectoryReloaderCrlProviderTestNoFixture, DenyUndeterminedTrue) {
+  auto provider = grpc_core::experimental::CreateDirectoryReloaderCrlProvider(
+      kCrlDirectoryPath, std::chrono::seconds(60), nullptr, true);
```
matthewstevenson88, line 86: nit: Same here RE inline comment for literals.

matthewstevenson88 (commented):
Thanks for the PR Greg! LGTM, just a couple minor comments.

include/grpc/grpc_crl_provider.h
gtcooke94: I agree, done

include/grpc/grpc_crl_provider.h
gtcooke94: I think we can do this in a separate PR with more of the de-experimentalization work - added a TODO

src/core/tsi/ssl_transport_security.cc
gtcooke94: I don't believe so - this is an absl::StatusOr, so in the case that it is an absl::Status, there won't be a CRL to free

test/cpp/end2end/crl_provider_test.cc
gtcooke94: Done

test/core/tsi/crl_ssl_transport_security_test.cc
gtcooke94: Done

include/grpc/grpc_crl_provider.h
matthewstevenson88: SG, thanks!

test/core/security/grpc_tls_crl_provider_test.cc
```diff
@@ -203,17 +203,15 @@ TEST_F(CrlProviderTest, StaticCrlProviderLookupIssuerNotFound) {
 }
 
 TEST_F(CrlProviderTest, StaticCrlProviderDenyUndeterminedDefault) {
-  std::vector<std::string> crl_strings = {GetFileContents(kCrlPath.data())};
   absl::StatusOr<std::shared_ptr<CrlProvider>> provider =
-      experimental::CreateStaticCrlProvider(crl_strings);
+      experimental::CreateStaticCrlProvider({});
```
matthewstevenson88, line 207: micro-nit: Please add `/*crl=*/`

test/core/security/grpc_tls_crl_provider_test.cc
```diff
@@ -203,17 +203,15 @@ TEST_F(CrlProviderTest, StaticCrlProviderLookupIssuerNotFound) {
...
   ASSERT_TRUE(provider.ok()) << provider.status();
   EXPECT_FALSE((*provider)->DenyUndetermined());
 }
 
 TEST_F(CrlProviderTest, StaticCrlProviderDenyUndeterminedTrue) {
-  std::vector<std::string> crl_strings = {GetFileContents(kCrlPath.data())};
   absl::StatusOr<std::shared_ptr<CrlProvider>> provider =
-      experimental::CreateStaticCrlProvider(crl_strings, true);
+      experimental::CreateStaticCrlProvider({}, true);
```
matthewstevenson88, line 214: micro-nit: Please add `/*crl=*/` and `/*deny_undetermined=*/`

src/core/tsi/ssl_transport_security.cc
matthewstevenson88: Yup good call, agreed.

matthewstevenson88 (approved):

PR #36859, opened 2024-06-08
[grpc][Gpr_To_Absl_Logging] Deleting all instances of gpr_set_log_verbosity

tanvi-jagtap:
Deleting all instances of gpr_set_log_verbosity

### DIFF
--- a/test/cpp/qps/driver.cc
+++ b/test/cpp/qps/driver.cc
@@ -364,7 +364,7 @@ std::unique_ptr<ScenarioResult> RunScenario(
     g_inproc_servers = new std::vector<grpc::testing::Server*>;
   }
   // Log everything from the driver
-  gpr_set_absl_verbosity_debug();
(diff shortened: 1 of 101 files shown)


JuneYan (approved):

test/cpp/qps/driver.cc
```diff
@@ -363,8 +363,6 @@ std::unique_ptr<ScenarioResult> RunScenario(
   if (run_inproc) {
     g_inproc_servers = new std::vector<grpc::testing::Server*>;
   }
-  // Log everything from the driver
-  gpr_set_log_verbosity(GPR_LOG_SEVERITY_DEBUG);
```
tanvi-jagtap, line 366: This is really not a good place to add a debug log. This kind of work should not be done in the middle of some test code. 
This should ideally be done in the main() function of the test, or setup or some other place.

JuneYan (commented):
LGTM.

Issue #36060, opened 2024-03-06
Use go1.22 for go interop tests and for future interop_matrix tests


arvindbr8:
gRPC-Go [added support](https://github.com/grpc/grpc-go/pull/7005) for go v1.22. This issue is to track adding go1.22 for go interop tests and for future interop_matrix tests similar to https://github.com/grpc/grpc/pull/31732

arjan-bal:
I have a PR to fix this: https://github.com/grpc/grpc/pull/36755

Will mark it as ready for review soon.

arvindbr8:
Thanks for looking into this @arjan-bal. Assigning you to this issue.

PR #36864, opened 2024-06-10
Use go1.22 for Go master branch interop tests

arjan-bal:
Similar to https://github.com/grpc/grpc/pull/36755 which was done for a release branch.

Fixes: https://github.com/grpc/grpc/issues/36060


arvindbr8:
> Also prevent installing the same python dependencies twice due to duplication in template from https://github.com/grpc/grpc/issues/36060.

@arjan-bal, Since we have multiple changes going in, might be worth it to split this^ to a separate PR for isolation.

arjan-bal:
> > Also prevent installing the same python dependencies twice due to duplication in template from #36060.
> 
> @arjan-bal, Since we have multiple changes going in, might be worth it to split this^ to a separate PR for isolation.

I've kept this PR to remove the duplicate installation. Will raise a followup PR once this is merged. Since both PRs depend on each other, once will need to be submitted before the other can be opened.

Fixing the duplication in the second PR would require regenerating the images that were just created, so I've not done that.

arvindbr8 (approved):

PR #36879, opened 2024-06-11

tanvi-jagtap:

We have the following mapping

Reviewers need to check :


test/core/slice/slice_test.cc
```diff
@@ -395,14 +394,19 @@ size_t SumSlice(const Slice& slice) {
...
   external_string.reset();
   // In ASAN (where we can be sure that it'll crash), go ahead and read the
   // bytes we just deleted.
   if (BuiltUnderAsan()) {
-    ASSERT_DEATH({ gpr_log(GPR_DEBUG, "%" PRIdPTR, SumSlice(slice)); }, "");
+    ASSERT_DEATH(
+        {
+          size_t sum = SumSlice(slice);
+          VLOG(2) << sum;
+        },
```
tanvi-jagtap, line 408: Using VLOG(2) << SumSlice(slice) does not work because the absl macro expands around this call, and then the ASSERT_DEATH does not work the way we want it to work.

test/core/iomgr/fd_posix_test.cc
```diff
@@ -164,8 +164,7 @@ static void session_read_cb(void* arg,  // session
       // before notify_on_read is called.
       grpc_fd_notify_on_read(se->em_fd, &se->session_read_closure);
     } else {
-      grpc_core::Crash(absl::StrFormat("Unhandled read error %s",
-                                       grpc_core::StrError(errno).c_str()));
+      LOG(FATAL) << "Unhandled read error " << grpc_core::StrError(errno);
```
JuneYan, line 167: For my learning, will LOG(FATAL) maintains locations information as grpc_core::Crash (http://shortn/_BojiaqEO6l)? Thanks

test/core/slice/slice_test.cc
```diff
@@ -395,14 +394,19 @@ size_t SumSlice(const Slice& slice) {
...
   Slice slice = Slice::FromExternalString(*external_string);
-  const auto initial_sum = SumSlice(slice);
+  const size_t initial_sum = SumSlice(slice);
   Slice owned = slice.AsOwned();
   EXPECT_EQ(initial_sum, SumSlice(owned));
   external_string.reset();
   // In ASAN (where we can be sure that it'll crash), go ahead and read the
   // bytes we just deleted.
   if (BuiltUnderAsan()) {
-    ASSERT_DEATH({ gpr_log(GPR_DEBUG, "%" PRIdPTR, SumSlice(slice)); }, "");
```
JuneYan, line 405: How about `PRIdPTR`?

JuneYan (approved):

PR #36883, opened 2024-06-11
[Gpr_To_Absl_Logging] Migrating from gpr to absl logging - gpr_log

tanvi-jagtap:

We have the following mapping

Reviewers need to check :


src/core/load_balancing/outlier_detection/outlier_detection.cc
```diff
@@ -924,25 +910,24 @@ void OutlierDetectionLb::EjectionTimer::OnTimerLocked() {
...
-            failure_percentage_ejection_candidates.size(), ejected_host_count,
-            success_rate_sum);
+    LOG(INFO) << "[outlier_detection_lb " << parent_.get() << "] found "
+              << success_rate_ejection_candidates.size()
+              << " success rate candidates and "
+              << failure_percentage_ejection_candidates.size()
+              << " failure percentage candidates; ejected_host_count="
+              << ejected_host_count
+              << "; success_rate_sum=" << success_rate_sum;
   }
```
tanvi-jagtap, line 920: Loss of the floating point formatting seems acceptable here.

src/core/load_balancing/outlier_detection/outlier_detection.cc
```diff
@@ -1021,11 +1001,10 @@ void OutlierDetectionLb::EjectionTimer::OnTimerLocked() {
...
-          gpr_log(GPR_INFO,
-                  "[outlier_detection_lb %p] random_key=%d "
-                  "ejected_host_count=%" PRIuPTR " current_percent=%.3f",
-                  parent_.get(), random_key, ejected_host_count,
-                  current_percent);
+          LOG(INFO) << "[outlier_detection_lb " << parent_.get()
+                    << "] random_key=" << random_key
+                    << " ejected_host_count=" << ejected_host_count
+                    << " current_percent=" << current_percent;
         }
```
tanvi-jagtap, line 1008: Loss of the floating point formatting seems acceptable here. I decided against using absl::StrFormat(...)

src/core/load_balancing/outlier_detection/outlier_detection.cc
tanvi-jagtap: Fixed

tools/run_tests/sanity/banned_functions.py
```diff
@@ -164,7 +164,6 @@
         "./src/core/lib/transport/connectivity_state.cc",
         "./src/core/lib/transport/transport.h",
         "./src/core/load_balancing/grpclb/grpclb.cc",
-        "./src/core/load_balancing/outlier_detection/outlier_detection.cc",
         "./src/core/load_balancing/pick_first/pick_first.cc",
         "./src/core/load_balancing/priority/priority.cc",
```
tanvi-jagtap, line 168: This file has a list of deprecated functions.

nanahpang (approved):

Issue #36613, opened 2024-05-14
Python asyncio servicer context invocation_metadata returns a tuple instead of Metadata


artificial-aidan:
### What version of gRPC and what language are you using?

Python 3.11 grpcio 1.63.0

### What operating system (Linux, Windows,...) and version?

OSX 10.14

### What runtime / compiler are you using (e.g. python version or version of gcc)

3.11

### What did you do?

Created a python asyncio server. Called `invocation_metadata` and had a `tuple` returned

```
    async def UnaryUnary(self, request: Request, context: grpc.aio.ServicerContext) -> Response:  # noqa: N802
        md = context.invocation_metadata()
        if isinstance(md, grpc.aio.Metadata):
            print('hello')
        type(md) # returns tuple
```

### What did you expect to see?

`invocation_metadata` to return a `grpc.aio.Metadata` object like the type hint suggests

gnossen:
Thanks for pointing this out! This does indeed seem to be an unintentional discrepancy. In order to minimize the impact of a fix here, I think the right thing to do is to change the _type_ to match the implementation rather than the other way around.

artificial-aidan:
That's definitely a bit of a bummer. But makes sense. It would be quite the breaking change to change the return type. Is there any precedence for adding a differently named method that returns metadata, or as a fuction parameter.

gnossen:
No precedent, but it's definitely something that could be discussed. We definitely verbally discussed something like `invocation_metadata2` when reviewing this issue. Though adding a whole new method would be a bit of a bigger effort (gRFC, implementation, testing, and documentation). I think we'd need someone to volunteer to pitch in to make something like that happen.

artificial-aidan:
Its a lot of effort for something that can be fixed with a one-liner turning it into metadata again.

I think fixing the typing would solve 90% of the problem

iamWing:
Was about to report this issue as well. Yes I agree with @gnossen that I wouldn't want to see a fix that changes the return type, as it'd break the current code base I have (which I do `dict(context.invocation_metadata())` to convert it to a dictionary to allow me using the `.get` function for getting a specific metadata). Something like `invocation_metadata2` would be welcome as it means that I can gradually migrate to the new function instead of having to deal with the breaking change all at once.

sreenithi:
the above PR should have updated the invocation_metadata() return type. Hence closing the issue

artificial-aidan:
[...]

PR #36894, opened 2024-06-12
[fix] updated invocation_metadata return type hint

sreenithi:
Updated the type hint for return type of grpc.aio.ServicerContext.invocation_metadata() to Optional[MetadataType] instead of Optional[Metadata].

MetadataType is already defined to be a Sequence[MetadatumType] and hence the type hint will reflect the sequence being returned, hence solving the [user's reported issue](https://github.com/grpc/grpc/issues/36613)

### DIFF
--- a/src/python/grpcio/grpc/aio/_base_server.py
+++ b/src/python/grpcio/grpc/aio/_base_server.py
@@ -18,7 +18,6 @@ from typing import Generic, Iterable, Mapping, NoReturn, Optional, Sequence
 
 import grpc
 
-from ._metadata import Metadata
 from ._typing import DoneCallbackType
 from ._typing import MetadataType
 from ._typing import RequestType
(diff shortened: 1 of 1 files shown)


XuanWang-Amos:
Looks good as your first PR!

Couple of small things:
* Use markdown format in description so it's easier to review, example: `grpc.aio.ServicerContext.invocation_metadata()`.
* This changed the return type of a public API, even it's a bug fix, we should still label it with "release note: yes" so it appears in our release notes (`release note` is also a required label for all PRs).

sreenithi:
hmm, but even the sync implementation `grpc._server.Context.invocation_metadata()` has the return type as `Optional[MetadataType]`. I thought it would make sense to keep them both same

XuanWang-Amos:
> hmm, but even the sync implementation `grpc._server.Context.invocation_metadata()` has the return type as `Optional[MetadataType]`. I thought it would make sense to keep them both same

Checked again, actually since `Tuple` is a type of `Sequence`, I think it should be fine to use `MetadataType`

XuanWang-Amos (approved):

PR #36903, opened 2024-06-12
[ruby] improve the way completion queue pluck operations handle signals and process shutdown

apolcyn:
Fixes the CBF of `src/ruby/end2end/killed_client_thread_test.rb` (failure mode is a hang of the child process that receives the SIGTERM) that has been happening since https://github.com/grpc/grpc/pull/36724

So far grpc-ruby CQ pluck operations have so far used a 20ms-interval busy poll to check interrupts in case we've received a signal, handle process shutdown, etc. This means ongoing RPCs will not terminate their CQ operations if we need to terminate the process (the loop simply exits without waiting for the CQ op to finish), causing a leak. Those RPCs can leave refs over their corresponding channels preventing [this](https://github.com/grpc/grpc/blob/8564f72e8e0334c25c480e0aec1df75bdc1fce14/src/ruby/ext/grpc/rb_channel.c#L653) from terminating (the channels don't reach state SHUTDOWN after being destroyed).

Fix is to unblock CQ pluck operations by cancelling calls, and thus allowing the CQ pluck to actually complete its operation. For server listening CQ operations, we unblock them by shutting down the server.

A side win here is to remove the [20ms-interval busy poll](https://github.com/grpc/grpc/blob/8564f72e8e0334c25c480e0aec1df75bdc1fce14/src/ruby/ext/grpc/rb_completion_queue.c#L44) on CQ operations, which was only needed to handle shutdown.

### DIFF
--- a/src/ruby/end2end/killed_client_thread_client.rb
+++ b/src/ruby/end2end/killed_client_thread_client.rb
@@ -19,6 +19,8 @@
 
 require_relative './end2end_common'
 
+ENV["GRPC_VERBOSITY"] = "DEBUG"
+
 def main
   parent_controller_port = ''
   server_port = ''
--- a/src/ruby/end2end/killed_client_thread_client.rb
+++ b/src/ruby/end2end/killed_client_thread_client.rb
@@ -19,8 +19,6 @@
 
 require_relative './end2end_common'
 
-ENV["GRPC_VERBOSITY"] = "DEBUG"
-
 def main
   parent_controller_port = ''
   server_port = ''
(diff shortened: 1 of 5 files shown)


stanley-cheung (approved):
LGTM. Any tests affected?

apolcyn:
> LGTM. Any tests affected?

This doesn't require any test *changes*

It fixes a CBT in `src/ruby/end2end/killed_client_thread_test.rb` though - that has been failing since a ~week ago as noted:

`Timed out waiting for client process. It likely freezes when killed while in the middle of an rpc (RuntimeError)`

PR #36915, opened 2024-06-13
[promises]: fix UB on uninitialized bool

deeedob:
Add the missing initializations in the constructors of 'NextResult' as the std doesn't provide it for primitive types.

### DIFF
--- a/src/core/lib/promise/pipe.h
+++ b/src/core/lib/promise/pipe.h
@@ -61,9 +61,9 @@ struct Pipe;
 template <typename T>
 class NextResult final {
  public:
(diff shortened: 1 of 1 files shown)


ctiller:
This is uninitialized by intent: note the comment on line 81 -- this is exactly that case.
We want to leave the value uninitialized so that ubsan can flag invalid usage.

Is there something you're seeing that's triggering this change?

ctiller:
Of course, and thanks for the PR regardless!

I'd be interested in seeing a callstack if you ever do find this error again.

PR #36921, opened 2024-06-13
[chttp2] don't access endpoint in transport ops if it's already been destroyed

markdroth:
This fixes a bug introduced in #36195.

b/345103652


ctiller (approved):

sampajano (commented):
Thanks Mark! 😊

@HannahShiSFB FYI

PR #36925, opened 2024-06-14
[Security] Distributor refactor

gtcooke94:
Rename `grpc_tls_certificate_distributor` to `TlsCertificateDistributor` and swap it from `grpc_core::RefCountedPtr` to `std::shared_ptr`

### DIFF
--- a/src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.h
+++ b/src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.h
@@ -212,6 +212,4 @@ class TlsCertificateDistributor {
       ABSL_GUARDED_BY(mu_);
 };
 
-typedef TlsCertificateDistributor TlsCertificateDistributor;
-
 #endif  // GRPC_SRC_CORE_LIB_SECURITY_CREDENTIALS_TLS_GRPC_TLS_CERTIFICATE_DISTRIBUTOR_H


src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.h
```diff
@@ -213,4 +212,6 @@ struct grpc_tls_certificate_distributor
       ABSL_GUARDED_BY(mu_);
 };
 
+typedef TlsCertificateDistributor TlsCertificateDistributor;
```
matthewstevenson88, line 215: Is this necessary?

src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.h
```diff
@@ -40,8 +40,7 @@ struct grpc_tls_identity_pairs {
 };
 
 // TLS certificate distributor.
-struct grpc_tls_certificate_distributor
-    : public grpc_core::RefCounted<grpc_tls_certificate_distributor> {
+class TlsCertificateDistributor {
```
matthewstevenson88, line 43: Should we consider changing the filename too?

matthewstevenson88 (approved):

src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.h
gtcooke94: Ah good catch, that was a temporary typedef I put in for the old `grpc_tls_certificate_distributor` name

Then at some point it got changed in a full find and replace - removed

src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.h
gtcooke94: I don't think we need to change the filenames - they have their own convention, for example the `CrlProviders` are called `grpc_crl_provider.h`, then the internal tls specific impl is `grpc_tls_crl_provider.{h/cc}`.

gtcooke94:
@markdroth can you take a look at this when you get the chance? I think the failing tests are unrelated flakes?

PR #36928, opened 2024-06-14
[test] Deflake //test/core/xds:xds_client_test

eugeneo:
Make sure all updates are received before checking the expected values.


markdroth:
Looks like the failure was caused by #36875, so let's revert that instead.

PR #36931, opened 2024-06-15
[log] Do not set absl min log level if user does not set GRPC_VERBOSITY

yashykt:
Do not set `absl::MinLogLevel` if user does not set `GRPC_VERBOSITY`. Only set `absl::MinLogLevel` if the user has specified `GRPC_VERBOSITY`.

Background:
gRPC Core overrides the setting of `absl::MinLogLevel` based on the setting of `GRPC_VERBOSITY`. 
`GRPC_VERBOSITY` defaults to the setting of `GPR_DEFAULT_LOG_VERBOSITY_STRING` if that env var is not set, resulting in the consequence that gRPC Core would override the setting of `absl::MinLogLevel` even if the user had previously modified it.

Note that even with this change, the minimum log severity that's printed remains at `ERROR` by default.


ctiller (approved):

ctiller (approved):

ctiller:
Adding a note that this and https://github.com/grpc/grpc/pull/36934 are essentially the same thing and I'm counting on communication not via me to ensure that only one is submitted :)

PR #36957, opened 2024-06-18
[log] Do not set absl min log level if user does not set GRPC_VERBOSITY (#36931)

yashykt:
Backport #36931 to 1.65.x


src/core/util/log.cc
```diff
@@ -150,6 +146,9 @@ void gpr_to_absl_verbosity_setting_init(void) {
   } else if (absl::EqualsIgnoreCase(verbosity, "NONE")) {
     absl::SetVLogLevel("*grpc*/*", -1);
     absl::SetMinLogLevel(absl::LogSeverityAtLeast::kInfinity);
+  } else if (verbosity.empty()) {
+    // Do not set absl::MinLogLevel if verbosity has not been set. Note that the
+    // default gRPC min log severity that is printed will still be ERROR.
```
tanvi-jagtap, line 151: I think we can avoid commenting on what the default will be. Because that is now controlled by absl. Not by us

tanvi-jagtap: "Do not override absl settings if GRPC_VERBOSITY has not been set."

tanvi-jagtap (approved):

XuanWang-Amos (approved):

src/core/util/log.cc
yashykt: leaving it to you in https://github.com/grpc/grpc/pull/36961/files

PR #36961, opened 2024-06-18
[Gpr_To_Absl_Logging] Fixing bugs 

tanvi-jagtap:
### Problem 1

Context : 
gpr_log() internally calls gpr_log_message() . 
gpr_log_message() may either call gpr_default_log() or a user provided logging function.
gpr_default_log() uses absl LOG. This is the default. Most users will log this way.
For the small percentage of users who have customized the logging function, gpr_log will log to custom this function. 

Problem :
We have converted half the instances of gpr_log to absl LOG().
For users who use the defaults - there will be no issue.
For the users who use a customized logging function
1. All the absl LOGs will log to the absl log sink.
2. All the gpr_log statements will log via this user provided function.
This is in-consistent behaviour and will cause confusion and difficulty in debugging.

Solution:
All logs should go to the same sink.
So we decided to make gpr_set_log_function a no op in this release.
The function will be deleted in the next release. 
https://github.com/grpc/proposal/pull/425

### Problem 2

Context :
gpr_should_log is used to avoid computing expensive stuff for logging if the log is not going to be visible.

Problem :
gpr_should_log was referencing the GRPC_VERBOSITY flag and values set by gpr_set_log_verbosity .
However, actual logging happens based on the absl settings.
This is incorrect. Because if the old settings are not honoured, they should not be checked and no decision in code should be made based on settings which are not going to get used.

Solution :
Given the above changes in Problem 1, since all custom logging is disabled, all logging from gRPC with honour the absl LOG settings. Hence we modified the gpr_should_log function to refer to absl settings.

### Problem 3

We still have the issue of php using a custom log sink. We will address this in a separate PR.

### Problem 4

Tests related to test/core/end2end/tests/no_logging.cc are broken . These will be fixed in another PR.


test/core/end2end/tests/no_logging.cc
```diff
@@ -54,9 +55,11 @@ class Verifier {
     }
     grpc_tracer_set_enabled("all", 0);
     grpc_set_absl_verbosity_debug();
+    // This is broken. Replace with an absl log sink.
```
yashykt, line 58: i'm assuming you are going to fix this as part of this PR?

tanvi-jagtap: No . This is a separate PR. About 2 days of work I think

ctiller: I can live with this being broken for a little bit. Let's get the release sorted and then put this as top priority.

yashykt (approved):

PR #36962, opened 2024-06-18
[Gpr_To_Absl_Logging] Remove custom logger from php.

tanvi-jagtap:
[Gpr_To_Absl_Logging] Remove custom logger from php. Not supported anymore.
While I am writing a custom absl LogSink to fix the issue in the next Pull Request,
I wanted to know if not having this LogSink is a release blocker or not.

### DIFF
--- a/src/php/ext/grpc/php_grpc.c
+++ b/src/php/ext/grpc/php_grpc.c
@@ -90,8 +90,6 @@ ZEND_GET_MODULE(grpc)
                      grpc_verbosity, zend_grpc_globals, grpc_globals)
    STD_PHP_INI_ENTRY("grpc.grpc_trace", NULL, PHP_INI_SYSTEM, OnUpdateString,
                      grpc_trace, zend_grpc_globals, grpc_globals)
-   STD_PHP_INI_ENTRY("grpc.log_filename", NULL, PHP_INI_SYSTEM, OnUpdateString,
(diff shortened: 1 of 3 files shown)


src/php/ext/grpc/php_grpc.c
```diff
@@ -249,38 +249,38 @@ void apply_ini_settings(TSRMLS_D) {
...
-               now.tv_nsec, display_file, args->line);
-
-  gpr_asprintf(&final, "%-60s %s\n", prefix, args->message);
-
-  fprintf(fp, "%s", final);
-  fclose(fp);
-  gpr_free(prefix);
-  gpr_free(final);
-}
+// static void custom_logger(gpr_log_func_args* args) {
```
stanley-cheung, line 252: I would prefer to just remove the code rather than commenting it out.

Also we need to remove this line from the src/php/README: https://github.com/grpc/grpc/blob/master/src/php/README.md?plain=1#L339

tanvi-jagtap: Fixed.

tools/run_tests/sanity/banned_functions.py
```diff
@@ -34,7 +34,6 @@
         "./src/core/util/log.cc",
```
tanvi-jagtap, line 34: This file has a list of deprecated functions.

PR #36968, opened 2024-06-18
[LB policy API] change metadata mutation API to handle discarded picks

markdroth:
Previously, metadata mutations were made by the picker directly, which meant that they would be applied even if the channel winds up discarding the pick due to the returned subchannel having been disconnected by the time the pick result is returned.  This changes the API such that pickers return metadata mutations along with the pick result, so that the mutations won't get applied unless the pick result is actually used.


src/core/load_balancing/lb_policy.h
```diff
@@ -116,27 +118,33 @@ class LoadBalancingPolicy : public InternallyRefCounted<LoadBalancingPolicy> {
...
+    void Add(absl::string_view key,
+             grpc_event_engine::experimental::Slice value) {
+      additions_.push_back({key, std::move(value)});
+    }
+
+   private:
+    friend class MetadataMutationHandler;
+
+    absl::InlinedVector<
+        std::pair<absl::string_view, grpc_event_engine::experimental::Slice>, 3>
```
ctiller, line 146: Let's add a comment explaining why 3 (in case we ever think to change it)

src/core/load_balancing/lb_policy.h
```diff
@@ -116,27 +118,33 @@ class LoadBalancingPolicy : public InternallyRefCounted<LoadBalancingPolicy> {
...
-    /// Adds a key/value pair.
-    /// Does NOT take ownership of \a key or \a value.
-    /// Implementations must ensure that the key and value remain alive
-    /// until the call ends.  If desired, they may be allocated via
-    /// CallState::Alloc().
-    virtual void Add(absl::string_view key, absl::string_view value) = 0;
+    virtual absl::optional<absl::string_view> Lookup(
+        absl::string_view key, std::string* buffer) const = 0;
 
     /// Produce a vector of metadata key/value strings for tests.
```
ctiller, line 124: Noting it would be nice to figure out how to eliminate this

ctiller (approved):

src/core/load_balancing/lb_policy.h
markdroth: Good point.  And this PR actually made it easy to do that, so I've gone ahead and done so.

src/core/load_balancing/lb_policy.h
markdroth: Done.

src/core/load_balancing/lb_policy.h
ctiller: TestOnlyCopyToVector still seems to be here

markdroth: It's gone from the public API, only present in the internal implementation, which the tests reach into.

PR #36977, opened 2024-06-19
[Gpr_To_Absl_Logging] Cleaning up the allow list for banned functions

tanvi-jagtap:
[Gpr_To_Absl_Logging] Cleaning up the allow list for deprecated functions.

This file has a list of deprecated functions.


tools/run_tests/sanity/banned_functions.py
```diff
@@ -14,19 +14,31 @@
...
+# It would be wise to do the file clean up and the altering of the allow list
+# in the same PR. This will make sure that any roll back of a clean up PR will
+# also alter the allow list and avoid build failures.
 
 import os
 import sys
 
 os.chdir(os.path.join(os.path.dirname(sys.argv[0]), "../../.."))
 
 # map of banned function signature to allowlist
```
sastryvp, line 38: NIT: Update this comment

tools/run_tests/sanity/banned_functions.py
```diff
@@ -291,7 +288,9 @@
             continue
         with open(path) as f:
             text = f.read()
-        for banned, exceptions in list(BANNED_EXCEPT.items()):
+        for banned, exceptions in list(
```
sastryvp, line 291: NIT: You may want to rename the variable exceptions

sastryvp (approved):

tools/run_tests/sanity/banned_functions.py
tanvi-jagtap: Done

tools/run_tests/sanity/banned_functions.py
tanvi-jagtap: Done.

sastryvp (approved):

PR #36979, opened 2024-06-19

tanvi-jagtap:

We have the following mapping

Reviewers need to check :


src/core/util/posix/tmpfile.cc
```diff
@@ -48,14 +48,15 @@ FILE* gpr_tmpfile(const char* prefix, char** tmp_filename) {
...
+    LOG(ERROR) << "mkstemp failed for filename_template " << filename_template
+               << " with error " << grpc_core::StrError(errno).c_str();
     goto end;
   }
   result = fdopen(fd, "w+");
   if (result == nullptr) {
-    gpr_log(GPR_ERROR, "Could not open file %s from fd %d (error = %s).",
-            filename_template, fd, grpc_core::StrError(errno).c_str());
+    LOG(ERROR) << "Could not open file " << filename_template << " from fd "
+               << fd << " (error = " << grpc_core::StrError(errno).c_str()
```
JuneYan, line 58: Is c_str() still needed here? Same as above.

tanvi-jagtap: grpc_core::StrError returns std::string which is supported by ABSL << overloaded operator.

tanvi-jagtap: Fixed

JuneYan (approved):

PR #24072, opened 2020-09-08
[Ruby] Implement setter for logger of GRPC module

nvh0412:
This commit implements `logger=` for GRPC module. It will allow us to
easy inject a custom logger on the fly. Using `self.extend` inside this
setter will help us on dealing with backward compatible for some custom
Logger Module.

Fixed #24084 

@markdroth

PR #36989, opened 2024-06-20
 [Ruby] Implement setter for logger of GRPC module

### LINKED ISSUE #24084: [Ruby] Add setter for GRPC.logger
Inside our system, we're using logger object which is passed from some upstream services, sometime it's instance of Logger ruby module, sometime it's Rails logger and sometime it's logging gem. It leads us to quite hard to inject these logger instance into `GRPC.logger` to get `info logs`

Implement a setter for logger of GRPC module which allows upstream service (or used place) easy to inject their custom logger

Upstream service can inject their logger instance by themselve by using GRPC.extend(Module.new.....), but I believe it isn't a suitable way to use a custom logger inside a library

nvh0412:
This commit implements `logger=` for GRPC module. It will allow us to inject a custom logger on the fly easily. Using `self.extend` inside this setter will help us deal with backward compatible for some custom Logger Module.

Fixed #24084 

cc @sampajano , @apolcyn and @alto-ruby TIA

re: https://github.com/grpc/grpc/pull/24072

### DIFF
--- a/src/ruby/lib/grpc/logconfig.rb
+++ b/src/ruby/lib/grpc/logconfig.rb
@@ -14,6 +14,19 @@
 
 # GRPC contains the General RPC module.
 module GRPC
+  def self.logger=(logger_obj)
+    # Need a free variable here to keep value of logger_obj for logger closure
+    @logger = logger_obj
+
+    extend(
+      Module.new do
+        def logger
+          @logger
+        end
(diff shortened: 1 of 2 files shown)


sampajano (approved):
@nvh0412 Thank you again for your contribution! 😃

PR #36994, opened 2024-06-20
[Gpr_To_Absl_Logging] Fixing bugs 

tanvi-jagtap:
[Gpr_To_Absl_Logging] Fixing bugs 

### Problem 1

Context : 
gpr_log() internally calls gpr_log_message() . 
gpr_log_message() may either call gpr_default_log() or a user provided logging function.
gpr_default_log() uses absl LOG. This is the default. Most users will log this way.
For the small percentage of users who have customized the logging function, gpr_log will log to custom this function. 

Problem :
We have converted half the instances of gpr_log to absl LOG().
For users who use the defaults - there will be no issue.
For the users who use a customized logging function
1. All the absl LOGs will log to the absl log sink.
2. All the gpr_log statements will log via this user provided function.
This is in-consistent behaviour and will cause confusion and difficulty in debugging.

Solution:
All logs should go to the same sink.
So we decided to make gpr_set_log_function a no op in this release.
The function will be deleted in the next release. 
https://github.com/grpc/proposal/pull/425

### Problem 2

Context :
gpr_should_log is used to avoid computing expensive stuff for logging if the log is not going to be visible.

Problem :
gpr_should_log was referencing the GRPC_VERBOSITY flag and values set by gpr_set_log_verbosity .
However, actual logging happens based on the absl settings.
This is incorrect. Because if the old settings are not honoured, they should not be checked and no decision in code should be made based on settings which are not going to get used.

Solution :
Given the above changes in Problem 1, since all custom logging is disabled, all logging from gRPC with honour the absl LOG settings. Hence we modified the gpr_should_log function to refer to absl settings.

### DIFF
--- a/include/grpc/support/log.h
+++ b/include/grpc/support/log.h
@@ -28,15 +28,10 @@
 extern "C" {
 #endif
 
-/** GPR log API.
-
-   Usage (within grpc):
-
-   int argument1 = 3;
-   char* argument2 = "hello";
-   gpr_log(GPR_DEBUG, "format string %d", argument1);
-   gpr_log(GPR_INFO, "hello world");
-   gpr_log(GPR_ERROR, "%d %s!!", argument1, argument2); */
+/**
+ * Logging functions in this file are deprecated.
+ * Please use absl ABSL_LOG instead.
+ */
 
 /** The severity of a log message - use the #defines below when calling into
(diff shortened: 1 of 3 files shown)


yashykt:
no, let's not do this. Let's wait for the original PR to get merged. Doing it this way will make future backports harder

PR #37034, opened 2024-06-24
[example] Add C++ retry example

yijiem:

### DIFF
--- a/examples/cpp/retry/CMakeLists.txt
+++ b/examples/cpp/retry/CMakeLists.txt
@@ -1,0 +1,73 @@
+# Copyright 202 gRPC authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


veblush (commented):
Can we have a cmake build?

examples/cpp/retry/server.cc
```diff
@@ -0,0 +1,84 @@
...
+using helloworld::Greeter;
+using helloworld::HelloReply;
+using helloworld::HelloRequest;
+
+// Logic and data behind the server's behavior.
+class GreeterServiceImpl final : public Greeter::Service {
+ public:
+  Status SayHello(ServerContext* context, const HelloRequest* request,
+                  HelloReply* reply) override {
+    if (++request_counter_ % request_modulo_ != 0) {
```
veblush, line 45: Having some comments on this logic would be helpful.

yijiem:
> Can we have a cmake build?

done

examples/cpp/retry/server.cc
yijiem: done. thanks!

examples/cpp/retry/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
+# Copyright 202 gRPC authors.
```
veblush, line 1: 2024?

veblush (approved):
LGTM! Thanks!

examples/cpp/retry/CMakeLists.txt
yijiem: good catch! I blame the keyboard lol.

PR #37051, opened 2024-06-25
[LB policy API] add TODOs about fixing ABA problems

markdroth:


src/core/load_balancing/subchannel_interface.h
```diff
@@ -76,6 +76,8 @@ class SubchannelInterface : public DualRefCounted<SubchannelInterface> {
 
   // Cancels a connectivity state watch.
   // If the watcher has already been destroyed, this is a no-op.
+  // TODO(roth): This interface has an ABA issue.  Fix this before we
+  // make this API public.
```
ctiller, line 80: I think we need this for AddDataWatcher too

markdroth: Done.

ctiller (approved):

PR #37070, opened 2024-06-27
unit testing setup and unit-tests script

sourabhsinghs:
1. Added script to install all dependencies then run all unit tests for python grpcio.


tools/internal_ci/linux/grpc_unittests_python_in_docker.sh
```diff
@@ -0,0 +1,70 @@
...
+      maybe_run_command build_package_protos
+      python3 -m pip install .;
+    popd;
+  done
+popd;
+
+# Run tests for grpcio_tests
+pushd src/python/grpcio_tests;
+  python3 setup.py test_lite
+# python setup.py test_aio
```
XuanWang-Amos, line 66: Why skipping aio here?

sreenithi (commented):
are these tests meant for Linux setup only or do we want to add the same for macos and windows too? 
In that case we would need additional configs and scripts under `tools/internal_ci/macos/pull_request` and `tools/internal_ci/windows/pull_request`

tools/internal_ci/linux/grpc_unittests_python_in_docker.sh
```diff
@@ -0,0 +1,70 @@
...
+cd "$BASEDIR";
+
+# unit-tests setup starts from here
+function maybe_run_command () {
+  if python3 setup.py --help-commands | grep "$1" &>/dev/null; then
+    python3 setup.py "$1";
+  fi
+}
+
+python3 -m pip install --upgrade "cython<3.0.0rc1";
```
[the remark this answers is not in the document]
sourabhsinghs: updated

tools/internal_ci/linux/grpc_unittests_python_in_docker.sh
sourabhsinghs: these testes were flacky during development. now they seems stable passing in local also un commented them

tools/internal_ci/linux/grpc_unittests_python_in_docker.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
gnossen, line 1: This seems to be a (near) copy of [this file](https://github.com/grpc/grpc/blob/master/tools/distrib/install_all_python_modules.sh). The `install_all_python_modules.sh` file name and path clearly indicate that it is meant for consumption by developers.

This file, however, lives in `internal_ci` and it has `in_docker` in the name. Shoudl we instead just update `install_all_python_modules.sh` to pull in any needed fixes and then call _it_ from `grpc_unittests.sh`?

tools/internal_ci/linux/pull_request/grpc_unittests_python.cfg
```diff
@@ -0,0 +1,26 @@
+# Copyright 2021 The gRPC Authors
```
gnossen, line 1: Reminder that when we change the g3 kokoro job name, the name of this cfg file will have to change as well.

tools/internal_ci/linux/grpc_unittests_python_in_docker.sh
sourabhsinghs: Updated as suggested.

tools/internal_ci/linux/pull_request/grpc_unittests_python.cfg
sourabhsinghs: updated name to tooling

tools/distrib/install_all_python_modules.sh
```diff
@@ -13,45 +13,58 @@
...
 function maybe_run_command () {
-  if python setup.py --help-commands | grep "$1" &>/dev/null; then
-    python setup.py "$1";
+  if python3 setup.py --help-commands | grep "$1" &>/dev/null; then
+    python3 setup.py "$1";
   fi
 }
 
-set -e
+python3 -m pip install --upgrade "cython==3.0.0";
```
XuanWang-Amos, line 31: We should probably use `cython<4.0.0rc1`

tools/distrib/install_all_python_modules.sh
```diff
@@ -13,45 +13,58 @@
...
+      maybe_run_command build_package_protos
+      python3 -m pip install .;
+    popd;
+  done
+popd;
+
+# Run tests for grpcio_tests
+pushd src/python/grpcio_tests;
+  python3 setup.py test_lite
+  python setup.py test_aio
```
XuanWang-Amos, line 66: `python3`?

tools/distrib/install_all_python_modules.sh
sourabhsinghs: updated

tools/distrib/install_all_python_modules.sh
```diff
@@ -13,45 +13,58 @@
...
+      python3 setup.py clean;
+      maybe_run_command preprocess
+      maybe_run_command build_package_protos
+      python3 -m pip install .;
+    popd;
+  done
+popd;
+
+# Run tests for grpcio_tests
+pushd src/python/grpcio_tests;
```
gnossen, line 64: Why are we running tests in the install script? This script should just build/install all packages, correct? If the goal is to use these to ensure that our dev script ran correctly, I would suggest:

- splitting out a top-level script that _invokes_ `install_all_python_modules.sh` and then runs a test after that -- this gives us a separation between our developer tooling and the _test_ for that developer tooling
- let's choose a much smaller subset of tests -- the goal here isn't to test every facet of the gRPC library. I think a test that just imports all relevant modules successfully would be reasonable.

tools/distrib/install_all_python_modules.sh
```diff
@@ -13,45 +13,58 @@
...
-      pushd "${PACKAGE}";
-        python setup.py clean;
-        maybe_run_command preprocess
-        maybe_run_command build_package_protos
-        python -m pip install .;
-      popd;
-    done
-  popd;
-)
+chmod -R 755 src/
```
gnossen, line 70: Why is this necessary? Seems very odd to be modifying permissions on our source code.

sourabhsinghs: This was required for internal_ci jobs to pass. It tries to move some files at the end of build and fails due to permissions. This was one of the ways other have resolved this issues as suggested in internal stackoverflow

gnossen: Which files? Are you able to share the failure? Are we generating junk files that shouldn't be under the src repo? This seems like treating the symptom rather than the cause.

sourabhsinghs: `
warning: could not open directory 'src/python/grpcio_tests/tests/protoc_plugin/tmpkolcwjrfwkt_test/': Permission denied
warning: could not open directory 'src/python/grpcio_tests/tests/protoc_plugin/tmprgzxk_wwwkt_test/': Permission denied
warning: could not open directory 'src/python/grpcio_tests/tests/protoc_plugin/tmpjtqox7o7wkt_test/': Permission denied
warning: could not open directory 'src/python/grpcio_tests/tests/protoc_plugin/tmpvm0psk_kwkt_test/': Permission denied
`

gnossen: Looks like these are artifacts generated by the test run above. This command would be reasonable to have in a _testing_ script, but not in the developer-oriented script itself. I think if you split out the test like in [this comment](https://github.com/grpc/grpc/pull/37070#discussion_r1773848498) and move it to a separate script, then organizationally, this should be fine.

tools/distrib/install_all_python_modules.sh
sourabhsinghs: 1.  Created an uber file containing install and run tests scripts
2. install script is this file contains installation of all modules
3. run tests only runs the tooling tests
4. I think it's ok, to keep all tests in tests script . user can re-run all test by running this script

gnossen (approved):

tools/distrib/python_tooling_tests.sh
```diff
@@ -0,0 +1,28 @@
...
+
+set -ex
+
+BASEDIR=$(dirname "$0")/../..
+cd "$BASEDIR";
+
+# Run tests for grpcio_tests
+pushd src/python/grpcio_tests;
+  python3 setup.py test_lite
+  python3 setup.py test_aio
```
gnossen, line 24: This is going to become a source of flakiness and maintenance burden for us (it will inherit any existing flakiness, doubling the existing rate and potentially cause future repo lockdowns). Please file a bug against yourself to minimize the set of tests run by this script.

tools/distrib/install_all_python_modules.sh
gnossen: > user can re-run all test by running this script

Users will not be running these tests. Our CI will. The concern is that we are doubling any existing flakiness rate in these tests without a clear testing goal. (we just need to test that installation has succeeded)

XuanWang-Amos (approved):

PR #37075, opened 2024-06-27
[LB policy API] change metadata mutations to overwrite instead of append

markdroth:


src/core/load_balancing/lb_policy.h
```diff
@@ -124,8 +124,8 @@ class LoadBalancingPolicy : public InternallyRefCounted<LoadBalancingPolicy> {
   /// A list of metadata mutations to be returned along with a PickResult.
   class MetadataMutations {
    public:
-    /// Adds a key/value pair.  If the key is already present, the new
-    /// value will be appended with a comma delimiter.
+    /// Adds a key/value pair.  If the key is already present, it will
+    /// be replaced with the new value.
     void Add(absl::string_view key, absl::string_view value) {
```
ctiller, line 129: Probably `Set` is a better name -- later on if we want to support appending entries we might want this name.

ctiller (approved):
One nit that should be fixed up, otherwise lgtm.
Approving to get out of the way.

src/core/load_balancing/lb_policy.h
markdroth: Good point, done.

PR #37086, opened 2024-06-27
[call v3] add missing channelz updates in subchannel

markdroth:


ctiller (approved):

src/core/client_channel/subchannel.cc
```diff
@@ -856,6 +855,24 @@ bool Subchannel::PublishTransportLocked() {
             ->client_transport());
     InterceptionChainBuilder builder(
         connecting_result_.channel_args.SetObject(transport.get()));
+    if (channelz_node_ != nullptr) {
+      // TODO(ctiller): If/when we have a good way to access the subchannel
```
ctiller, line 859: @Vignesh2208 this is likely a good candidate for the connection context idea we talked about

markdroth: Just noting that we definitely don't want to do this with the current call context, since that would make it impossible to add hedging support.  But if we had some sort of per-connection context, I could see that working quite well.

PR #37096, opened 2024-06-28
[CI] Fixed build_cxx to accept additional cmake arguments

veblush:
Along with the fix, some code massage had to be made to resolve C++23 unique_ptr incomplete type errors.

### DIFF
--- a/tools/run_tests/helper_scripts/build_cxx.sh
+++ b/tools/run_tests/helper_scripts/build_cxx.sh
@@ -24,7 +24,7 @@ INSTALL_PATH="$(pwd)/cmake/install"
 cd third_party/abseil-cpp
 mkdir build
 cd build
-cmake -DCMAKE_CXX_STANDARD=14 -DABSL_BUILD_TESTING=OFF -DCMAKE_BUILD_TYPE="${MSBUILD_CONFIG}" -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" "$@" ..
+cmake -DABSL_BUILD_TESTING=OFF -DCMAKE_BUILD_TYPE="${MSBUILD_CONFIG}" -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" "$@" ..
 make -j"${GRPC_RUN_TESTS_JOBS}" install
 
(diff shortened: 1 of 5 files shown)


ctiller:
Does this mean we can use C++17? (asking for a friend)

veblush:
> Does this mean we can use C++17? (asking for a friend)

Haha no. This is to fix the bug in the test script. Currently build_cxx.sh ignores the cpp version given by other scripts, giving no valid result for C++20, C++23 tests.

tools/run_tests/helper_scripts/build_cxx.sh
```diff
@@ -24,15 +24,15 @@ INSTALL_PATH="$(pwd)/cmake/install"
...
+cmake -DABSL_BUILD_TESTING=OFF -DCMAKE_BUILD_TYPE="${MSBUILD_CONFIG}" -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" "$@" ..
 make -j"${GRPC_RUN_TESTS_JOBS}" install
 
 # Install opentelemetry-cpp since we only support "package" mode for opentelemetry at present.
 cd ../../..
 cd third_party/opentelemetry-cpp
 mkdir build
 cd build
-cmake -DCMAKE_CXX_STANDARD=14 -DWITH_ABSEIL=ON -DBUILD_TESTING=OFF -DWITH_BENCHMARK=OFF -DCMAKE_BUILD_TYPE="${MSBUILD_CONFIG}" -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" "$@" ..
+cmake -DWITH_ABSEIL=ON -DBUILD_TESTING=OFF -DWITH_BENCHMARK=OFF -DCMAKE_BUILD_TYPE="${MSBUILD_CONFIG}" -DCMAKE_INSTALL_PREFIX="${INSTALL_PATH}" "$@" ..
```
yashykt, line 35: as discussed we might need to make `-DCMAKE_CXX_STANDARD=<version>` a required argument set by the caller

veblush: It turned out that we need this only for Mac so I'll add this only to Mac.

yashykt (approved):

PR #37099, opened 2024-06-28
[call v3] implement ClientChannel::WatchConnectivityState()

markdroth:
We can't actually test this until we change `WorkSerializer` to use `absl::AnyInvocable<>` and then implement `ClientChannel::AddConnectivityWatcher()`, but I wanted to get this in place so that it's ready when the time comes.

### DIFF
--- a/src/core/client_channel/client_channel.cc
+++ b/src/core/client_channel/client_channel.cc
@@ -688,10 +688,93 @@ grpc_connectivity_state ClientChannel::CheckConnectivityState(
   return state;
 }
 
-void ClientChannel::WatchConnectivityState(grpc_connectivity_state, Timestamp,
-                                           grpc_completion_queue*, void*) {
-  // TODO(ctiller): implement
-  Crash("not implemented");
+namespace {
+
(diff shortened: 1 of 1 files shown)


src/core/client_channel/client_channel.cc
```diff
@@ -688,10 +688,93 @@ grpc_connectivity_state ClientChannel::CheckConnectivityState(
...
+    MutexLock lock(&mu_);
+    if (watcher_ == nullptr) return;  // Ignore subsequent notifications.
+    // Cancel watch.
+    channel_->RemoveConnectivityWatcher(watcher_);
+    watcher_ = nullptr;
+    // Cancel timer.
+    channel_->event_engine()->Cancel(timer_handle_);
+    // Send CQ completion.
+    Ref().release();  // Released in FinishedCompletion().
+    grpc_cq_end_op(cq_, tag_, status, FinishedCompletion, this,
```
ctiller, line 749: @yousukseung noting there's CQ ops that are not call ops (and you should be aware of such)

ctiller (approved):

PR #37154, opened 2024-07-03
[CSM] Increase CPP PSM CSM timeout to 180min

XuanWang-Amos:
Pyhton timeout was already set to 240min:
https://github.com/grpc/grpc/blob/3e8cdb9f396f290cc8725fe9e6180c37542d9e40/tools/internal_ci/linux/psm-csm-python.cfg#L19


sergiitk (changes requested):
Let's make the cpp change in the same PR?

sergiitk (approved):
Misread the PR, sorry

PR #37156, opened 2024-07-03
[xDS e2e tests] set channel/server creds in tests instead of as a test parameter

markdroth:
As part of the xDS authority rewriting implementation in #37087, I need to write some tests showing the authority being rewritten.  However, the fake security connector currently crashes on unexpected authorities.  (As a side note, I think the fake security connector is very cumbersome and should be redesigned, but that's a separate project.)  As a result, I need a way to use InsecureCreds on a per-test basis.

I thought about just adding an option to `XdsTestType` to trigger use of InsecureCreds, but the logic we use for determining which creds type to use for what is already very cumbersome, and adding another option there would have just made that worse.  Instead, I have switched to a simpler approach where the individual tests can decide what creds type to use directly.  This both unblocks my other PR and makes the existing code more maintainable.

### DIFF
--- a/test/cpp/end2end/xds/xds_core_end2end_test.cc
+++ b/test/cpp/end2end/xds/xds_core_end2end_test.cc
@@ -166,6 +166,9 @@ TEST_P(XdsClientTest, XdsStreamErrorPropagation) {
 
 class XdsServerTlsTest : public XdsEnd2endTest {
  protected:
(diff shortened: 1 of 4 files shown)


yashykt:
what's the rationale?

yashykt (approved):

markdroth:
I've updated the PR description with the details.

PR #37169, opened 2024-07-05
[Ruby] add ruby distrib test with protobuf 3.25

alto-ruby:
Add distrib test with protobuf 3.25

https://btx.cloud.google.com/invocations/3bbd9817-4f69-4ae6-b70b-77fac674c6be/targets/github%2Fgrpc%2Fdistribtests;config=default/tests

### DIFF
--- a/test/distrib/ruby/run_distrib_test.sh
+++ b/test/distrib/ruby/run_distrib_test.sh
@@ -25,6 +25,8 @@ function die {
 ARCH="$1"
 PLATFORM="$2"
 PACKAGE_TYPE="$3"
+PROTOBUF_VERSION="$4"
+
 echo "$EXTERNAL_GIT_ROOT"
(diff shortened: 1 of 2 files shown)


sampajano (commented):
@alto-ruby Thanks for the change!!

There are some Ruby test failures i could tell tho?

alto-ruby:
It's separate issue that ruby tests timeouts sometimes that I'm looking into but no clues. CC: @stanley-cheung

sampajano (approved):
> It's separate issue that ruby tests timeouts sometimes (50% accroding to @stanley-cheung ) that I'm looking into but yet no clues. PS, not only ruby, other tests sometimes timeout as well.

Oh i see!! Thanks for explaining! 😃 (It would be good to know how to slightly increase the timeout if it's timing out 50% of the time?)

Thanks again for the change!

PR #37177, opened 2024-07-08
[Gpr_To_Absl_Logging] Fixing unit test that flags log noise

tanvi-jagtap:
1.  Fixing unit test that flags log noise. 
2. This test was broken for many months. As a result , a lot of log noise was added. Removing the noise as a part of the PR.
3. If we want to retain any log line as `INFO` instead of `VLOG(2)`, please let me know, I will add it to allow list.
4. In this PR , we replace the old `gpr_set_log_function` mechanism with an `absl LogSink` . So here , `Send` function will do everything that `NoLog` used to do before.


test/core/end2end/tests/no_logging.cc
```diff
@@ -90,30 +105,23 @@ class Verifier {
     }
     auto it = allowed_logs_by_module->find(filename);
     if (it != allowed_logs_by_module->end() &&
```
yashykt, line 107: If we have this, seems like we don't need the regex?

test/core/end2end/tests/no_logging.cc
```diff
@@ -152,13 +160,9 @@ CORE_END2END_TEST(NoLoggingTest, NoLoggingTest) {
     GTEST_SKIP() << "not for windows + event engine listener";
   }
 #endif
-  Verifier verifier;
-  verifier.FailOnNonErrorLog();
+  VerifyLogNoiseLogSink nolog_verifier(absl::LogSeverityAtLeast::kInfo, 2);
   for (int i = 0; i < 10; i++) {
     SimpleRequest(*this);
   }
-  verifier.FailOnAnyLog();
```
yashykt, line 160: I don't understand this. It seems that previously, we were failing on any log, but the check that you have allows info logs and verbosity logs > 0

test/core/end2end/tests/no_logging.cc
tanvi-jagtap: Regex allow lists only specific text. So if an allow listed files has a new log line, which is noisy we want that to be flagged. I am not sure if I understand your question fully though.

test/core/end2end/tests/no_logging.cc
tanvi-jagtap: The current test only checks for LOG(INFO), LOG(ERROR) and LOG(WARNING) being noisy.
The current test ignores all VLOG(1) , VLOG(2) and VLOG(N) for N >= 1.

Why did I do this?

There are a tonne of VLOG(2) statements in our code. If I start failing this test for VLOG(2), our allow list would be flooded with entries. And this will become hard to maintain.

Production systems dont have VLOG turned on. I am assuming that the intent here is to protect systems which have absl log severity set to INFO, WARNING and ERROR in production systems. This test will protect server using log severity INFO, WARNING and ERROR.

If anyone is using log verbosity >= 1 in a production system, they will see massive log noise.

Removing these VLOG(2) statements will cause important debugging data to go away.
Moving all VLOG(2) statements inside the tracer will be a large effort. Also, we will have to make new tracers because many files dont have their own tracer.
Adding so many VLOG(2) log lines to the allow list would make for an insanely lengthy allow list. 
So ignoring them in this test was intentional.

tanvi-jagtap: If we want to clean out the usage of VLOG(2) and put it behind a trace, that ideally then should be a separate CL.
Because 
1. It is going to be large. 
2. It will take about 1-3 days to prepare. And another 1-3 days to review. In the mean time, we should atleast get the current test submitted and running. Before we accidentally submit PRs with more log noise.

test/core/end2end/tests/no_logging.cc
yashykt: I see, ok

test/core/end2end/tests/no_logging.cc
yashykt: My understanding is that you just converted existing gpr_log statements into VLOG statements. If the statements were not being logged previously, I expect them to not be logged now either.

> There are a tonne of VLOG(2) statements in our code. If I start failing this test for VLOG(2), our allow list would be flooded with entries. And this will become hard to maintain.

Did you verify this? Can you show me what VLOG statements show up if you send another RPC?

tanvi-jagtap: While I was converting gpr_log to absl log, this test was broken for about 5-10 weeks. In the meantime, large PRs were submitted which added VLOG(2) statements. And the test missed them. 
A few examples of which VLOGs are noisy are available in this PR itself. Because I converted LOG(INFO) to VLOG(2). This conversion was necessary coz many prod systems still choose to keep INFO as their default log level.
The full list of noisy VLOGs I will provide in a few hours.

tanvi-jagtap: Fixed. Now VLOG(2) will be checked too. A lot of entries from dual_ref_counted.h . But other than that the allow list is small.
The latest code should work.

tanvi-jagtap: Fixed. Now VLOG(2) will be checked too. A lot of entries from dual_ref_counted.h . But other than that the allow list is small.
The latest code should work.

tanvi-jagtap: I was wrong about the list being small . My bad . Locally when I run the test, the allow list is small. But on Kokoro , I see a lot of flakes , with logs showing up here and there . 
This VLOG(2) allow list does not look practical to maintain IMO.
Additionally, it does not serve much purpose to regulate the usage of VLOG(2) because this is almost guaranteed to be off in production. 
The goal IIUC was to prevent log spam. I did not initially intend to maintain parity with the older test. Please can you share the rationale for wanting to regulate VLOG(2) , I am curious. 
@yashykt

yashykt: ok, I can accept VLOG not being monitored since they are not logged by default.

yashykt (commented):
ok, I am fine with `VLOG` statements being ignored as long as `LOG()` statements are monitored.

yashykt (approved):

PR #37185, opened 2024-07-08
[xDS] implement system root cert support

markdroth:
As per gRFC A82 (https://github.com/grpc/proposal/pull/436).


src/core/load_balancing/xds/xds_cluster_impl.cc
```diff
@@ -674,23 +674,32 @@ XdsClusterImplLb::MaybeCreateCertificateProviderLocked(
...
-      return absl::InternalError(
-          absl::StrCat("Certificate provider instance name: \"",
-                       root_provider_instance_name, "\" not recognized."));
-    }
-  }
+  bool use_system_root_certs = false;
+  absl::Status status = Match(
+      cluster_resource.common_tls_context.certificate_validation_context
+          .ca_certs,
+      [](const absl::monostate&) { return absl::OkStatus(); },
```
yashykt, line 683: can you add a comment saying that this means that no root certs are used.

yashykt (approved):

src/core/load_balancing/xds/xds_cluster_impl.cc
markdroth: Done.

yashykt (approved):

PR #37212, opened 2024-07-13
resolve NaN compile-time constant issue in message.c

kagancapar:
Resolved the issue with `NAN` not being a compile-time constant in Windows 11 SDK 10.0.26100.0. Updated the definition of `kUpb_NaN` to use `std::numeric_limits<double>::quiet_NaN()` for better portability and compatibility.

Changes:
- Replaced the `NAN` macro with `std::numeric_limits<double>::quiet_NaN()` for the `kUpb_NaN` constant.

Testing:
- Verified that the constants correctly represent positive infinity and NaN for both `float` and `double`.

Important Note:
- NaN is no longer a compile-time constant in Windows 11 SDK 10.0.26100.0. Rolling back to Windows 11 SDK 10.0.22000.0 resolves this issue.

### DIFF
--- a/third_party/upb/upb/message/internal/message.c
+++ b/third_party/upb/upb/message/internal/message.c
@@ -9,6 +9,7 @@
 
 #include <math.h>
 #include <string.h>
+#include <limits> // Added for std::numeric_limits
 
 #include "upb/base/internal/log2.h"
 #include "upb/mem/arena.h"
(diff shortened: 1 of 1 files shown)


yashykt:
This needs to be done in the protobuf repo - https://github.com/protocolbuffers/protobuf

PR #37135, opened 2024-07-01
[test-flake] Increase Ruby run_tests.py timeout to 20 minutes

stanley-cheung:
7zfxqBLsQo8ciuq

Curiously, some of the Ruby sub-tests job were given a 20-minute timeout, while the main `run_ruby.sh` job was given a 10-minute timeout. All the current flakes have the `run_ruby.sh` job timeout at 10 minutes. So this PR tries to increase that limit to 20-minute to see if that helps.

```
GRPC::GenericService::Dsl
  can be included in new classes

GRPC::GenericService
  #underscore
    should convert CamelCase to underscore separated
  including it
    adds a class method, rpc
    adds rpc descs using the added class method, #rpc
    give subclasses access to #rpc_descs
    adds a default service name
    adds a default service name to subclasses
    adds the specified service name
    adds the specified service name to subclasses
  #include
    raises if #rpc is missing an arg
    is ok for services that expect the default {un,}marshal methods
    is ok for services that override the default {un,}marshal methods
    when #rpc args are incorrect
      raises if an arg does not have the marshal or unmarshal methods
      raises if a type arg only has the marshal method
      raises if a type arg only has the unmarshal method
  #rpc_stub_class
    generates a client class that defines any of the rpc methods
    the generated instances
I0000 00:00:1719872817.473215   21463 rb_grpc.c:331] GRPC_RUBY: grpc_ruby_init_threads g_bg_thread_init_done=1
      can be instanciated with just a hostname and credentials
I0000 00:00:1719872817.481099   21463 rb_grpc.c:331] GRPC_RUBY: grpc_ruby_init_threads g_bg_thread_init_done=1
      has the methods defined in the service

Finished in 0.83804 seconds (files took 0.51661 seconds to load)
150 examples, 0 failures

Coverage report generated for idiomatic, wrapper to /var/local/git/grpc/coverage. 1001 / 1110 LOC (90.18%) covered.

2024-07-01 22:36:20,441 TIMEOUT: tools/run_tests/helper_scripts/run_ruby.sh [pid=18527, time=604.5sec]
2024-07-01 22:36:20,442 FAILED: tools/run_tests/helper_scripts/run_ruby.sh
```

PR #37230, opened 2024-07-16
[Ruby] Increased ruby build timeout to 20min

veblush:
Ruby tests have been flaky for a while on Linux. All I looked at failed with timeout during the build step which has 10 minutes timeout which sounds tight. So I'm increasing this to mitigate this problem.


stanley-cheung:
Just to make a note here (talked offline) - I attempted this once just a while ago #37135, and that didn't seem to help. The Ruby tests now just time out after 20m instead. So this is not a root cause.

PR #37236, opened 2024-07-17
[benchmark] Add a scalable_unrestricted benchmark category.

paulosjca:
The following benchmark hangs on 8-core machines, but passes on 30-core
machines:

```
cpp-protobuf-async-client-unary-1channel-64wide-128breq-8mbresp-insecure
```

This change creates a `scalable_unrestricted` category in addition to
the `scalable` category typically run in CI, and applies it to all
`scalable` benchmarks except the one that fails on 8-core machines.

This change also changes the set of benchmarks run on 8-core machines to
use the `scalable_unrestricted` category. Benchmarks run on 30-core
machines remain unchanged.

### DIFF
--- a/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
+++ b/tools/internal_ci/linux/grpc_e2e_performance_gke.sh
@@ -113,7 +113,7 @@ buildConfigs() {
     -a ci_gitCommit_java="${GRPC_JAVA_COMMIT}" \
     -a ci_gitActualCommit="${KOKORO_GIT_COMMIT}" \
     --prefix="${LOAD_TEST_PREFIX}" -u "${UNIQUE_IDENTIFIER}" -u "${pool}" \
-    -a pool="${pool}" --category=scalable \
+    -a pool="${pool}" \
     --allow_client_language=c++ --allow_server_language=c++ \
     --allow_server_language=node \
     -o "loadtest_with_prebuilt_workers_${pool}.yaml"
@@ -139,8 +139,8 @@ for language in "${disabledLanguages[@]}"; do
 done
 
 # Add arguments for languages.
-declare -a configLangArgs8core=()
-declare -a configLangArgs32core=()
+declare -a configLangArgs8core=("--category=scalable_unrestricted")
+declare -a configLangArgs32core=("--category=scalable")
 declare -a runnerLangArgs=()
 
 # c++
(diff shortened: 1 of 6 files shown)


tools/internal_ci/linux/grpc_e2e_performance_gke.sh
```diff
@@ -139,8 +139,8 @@ for language in "${disabledLanguages[@]}"; do
 done
 
 # Add arguments for languages.
-declare -a configLangArgs8core=()
-declare -a configLangArgs32core=()
+declare -a configLangArgs8core=("--category=scalable_unrestricted")
```
drfloob, line 142: This is backwards, right? `scalable_unrestricted` should contain the test that fails on 8-core machines.

tools/internal_ci/linux/grpc_e2e_performance_gke_experiment.sh
```diff
@@ -134,8 +134,8 @@ for language in "${disabledLanguages[@]}"; do
 done
 
 # Add arguments for languages.
-declare -a configLangArgs8core=()
-declare -a configLangArgs32core=()
+declare -a configLangArgs8core=("--category=scalable_unrestricted")
```
drfloob, line 137: Also backwards, yes?

tools/run_tests/performance/scenario_config.py
```diff
@@ -525,7 +534,8 @@ def scenarios(self):
                 resp_size=8 * 1024 * 1024,
                 secure=secure,
                 minimal_stack=not secure,
-                categories=inproc_categories + [SCALABLE],
+                categories=inproc_categories
+                + ([SCALABLE, SCALABLE_UNRESTRICTED] if secure else [SCALABLE]),
```
drfloob, line 538: Can't this just be `inproc_categories + [SCALABLE_UNRESTRICTED]`? The `secure` logic seems a bit like obfuscation, since the goal is to disable this on 8-core machines.

drfloob (commented):
Wouldn't it be simpler to add negative matchers, like `--exclude_categories=CATEGORY[,CATEGORY,...]`, or `--exclude_regex=<name matching for tests to skip>`? I feel like that would A) make the category sets easier to manage and reason about, and B) make it more explicit that some configurations have problems with some tests.

Using something like `--category=scalable --exclude_categories=8-core-problems` would require you to add just one category to the bad test, and add maybe 10 lines of filter code. I prefer this one.

Using something like `--category=scalable --exclude_regex='cpp-protobuf-async-client-unary-1channel-64wide-128breq-8mbresp-insecure'` would require no category changes, and would add maybe 5 lines of filter code.

tools/run_tests/performance/scenario_config.py
drfloob: Oh I see this is only an issue on insecure runs. I think a comment is warranted here, unless you go with one of my simpler suggestions.

Issue #37239, opened 2024-07-17
grpc.aio.Server.stop() is not graceful in case of aborted methods


ikalnytskyi:
### What version of gRPC and what language are you using?

The issue can be reproduced using both the latest python package (`v1.64.1`) and the `master` branch.

### What operating system (Linux, Windows,...) and version?

I successfully reproduced the issue on Linux (Arch Linux, Ubuntu 22.04) and macOS 14.5.

### What runtime / compiler are you using (e.g. python version or version of gcc)

CPython 3.11, CPython 3.12.
GCC 14.1.1

### What did you do?

I'm doing the graceful shutdown via `await server.stop(<timeout>)`, but it's not graceful for the aborted servicer methods. 

See example server and client implementations below:

<details>
<summary><code>dreamer.proto</code></summary>

```proto
syntax = "proto3";

message SleepRequest { int32 duration = 1; }
message Empty {}

service Dreamer {
  rpc Sleep(SleepRequest) returns (Empty) {}
}
```
</details>

<details>
<summary><code>dreamer_server.py</code></summary>

```python
import asyncio
import signal

import grpc

pb2, pb2_grpc = grpc.protos_and_services("dreamer.proto")


class DreamerServicer(pb2_grpc.DreamerServicer):
    async def Sleep(
        self,
        request: pb2.SleepRequest,
        context: grpc.aio.ServicerContext,
    ) -> pb2.Empty:
        print(f"Sleep(duration={request.duration}) - BEGIN")
        try:
            await context.abort(grpc.StatusCode.FAILED_PRECONDITION, "no way")
        finally:
            await asyncio.sleep(request.duration)
            print(f"Sleep(duration={request.duration}) - END")
        return pb2.Empty()


async def main() -> None:
    address = "127.0.0.1:50051"
    server = grpc.aio.server()
    shutdown = False
    pb2_grpc.add_DreamerServicer_to_server(DreamerServicer(), server)
    server.add_insecure_port(address)

    def _shutdown(signum: int, _) -> None:
        print(f"Received signal {signum}, shutting down the server...")
        nonlocal shutdown
        shutdown = True

    signal.signal(signal.SIGINT, _shutdown)
    signal.signal(signal.SIGTERM, _shutdown)

    print(f"Starting server at {address}")
    await server.start()
    while await server.wait_for_termination(timeout=1):
        if shutdown:
            await server.stop(10)


if __name__ == "__main__":
    asyncio.run(main())
```
</details>

<details>
[...]

PR #37240, opened 2024-07-17
python: await tasks in grpc.aio.Server.stop()

ikalnytskyi:
When `grpc.aio.Server.stop(<timeout>)` is awaited, it's expected for the server to be graceful shut down. This means that all in-flight requests are finished and clients received their responses, and all servicer methods and/or interceptors are fully finished.

Unfortunately, this is not what happening when a servicer method is aborted via `await context.abort()`. When `abort()` happens and the response is transferred to a client, gRPC core things that there are not in-flight requests and that the server can be shut down. While this is correct from the Core point of view, it's not necessarily correct from Python point of view, where some unfinished coroutines (rpc tasks) may not be finished yet.

It's important to make sure they are finished before returning control to a user from `stop()` because they may contain important logic in their `finally` blocks, such as releasing some distributed locks or other resources.

This patch updates the AioServer implementation to keep track of unfinished rpc coroutines, and make sure they are awaited by the shutdown procedure before returning control to a user.

Fixes: #37239


sreenithi:
Hi @ikalnytskyi, the approach seems to be good enough. Thank you for your contribution. Please also update the tests for this change

src/python/grpcio_tests/tests_aio/unit/server_test.py
```diff
@@ -230,6 +235,18 @@ async def _invalid_trailing_metadata(self, request, context):
...
+    async def _abort_and_slow_finalization(self, request, context):
+        try:
+            try:
+                await context.abort(grpc.StatusCode.INVALID_ARGUMENT)
+            finally:
+                # server.stop() should either wait for this to complete (graceful),
+                # or cancel this and finish execution
+                await asyncio.sleep(test_constants.SHORT_TIMEOUT)
+        finally:
+            self.abort_and_slow_finalization_finished = True
```
ikalnytskyi, line 247: **n.b.** Prior to this patch, this line was never executed because `server.stop()` completed much earlier, never waiting for the in-progress async tasks to complete.

ikalnytskyi:
Hey @sreenithi,

Thanks for the feedback. I think I have green tests now (at least locally) and added 2 more tests. I'm not particularly happy about those tests, so I'm open for feedback.

ikalnytskyi:
Hey @sreenithi,

Did you have any chance to take a look?

sreenithi:
Hi @ikalnytskyi , apologies for the delay. It looks like I completely missed your PR. I will take a look at it in the coming weeks. Meanwhile, would like to ask if you have tried your patch with the repro case mentioned in your issue [#37239](https://github.com/grpc/grpc/issues/37239) .

ikalnytskyi:
Closing this PR due to inactivity and lack of maintainer feedback. It doesn't seem like there's any interest in merging or moving this forward.

PR #37243, opened 2024-07-17
[Test] Increase registered callback's duration to avoid flakyness

yijiem:
Previously the registered callback's duration is set too low (10ms) that 2 different OTel callbacks get called from OTel consecutively would trigger the registered callback again, making the test flaky.

### DIFF
--- a/test/cpp/ext/otel/otel_plugin_test.cc
+++ b/test/cpp/ext/otel/otel_plugin_test.cc
@@ -1919,7 +1919,7 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
         reporter.Report(double_gauge_handle, double_value_1++, kLabelValuesSet2,
                         kOptionalLabelValuesSet2);
       },
-      grpc_core::Duration::Milliseconds(10) * grpc_test_slowdown_factor(),
+      grpc_core::Duration::Milliseconds(50) * grpc_test_slowdown_factor(),
       integer_gauge_handle, double_gauge_handle);
   int report_count_2 = 0;
   int64_t int_value_2 = 1;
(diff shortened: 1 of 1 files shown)


test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1956,8 +1956,8 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
       data = collector.Stop();
   // Verify that data is incremental without duplications (cached
   // values).
-  EXPECT_EQ(report_count_1, kIterations);
-  EXPECT_EQ(report_count_2, kIterations);
+  EXPECT_GE(report_count_1, kIterations);
```
yashykt, line 1959: can you dig into why this might be happening?

yijiem: reverted back and change the duration instead

yashykt (approved):
Thanks!

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1936,9 +1936,9 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
         reporter.Report(double_gauge_handle, double_value_2++, kLabelValuesSet2,
                         kOptionalLabelValuesSet2);
       },
-      grpc_core::Duration::Milliseconds(10) * grpc_test_slowdown_factor(),
+      grpc_core::Duration::Milliseconds(50) * grpc_test_slowdown_factor(),
       integer_gauge_handle, double_gauge_handle);
-  constexpr int kIterations = 100;
+  constexpr int kIterations = 50;
```
yashykt, line 1941: maybe reduce it to 20?

PR #37253, opened 2024-07-18
[log] Reducing log spam

yashykt:
This PR changes many of the `LOG(INFO)` statements that are not guarded by TraceFlags to either be -
* guarded by a TraceFlag
* logged under `VLOG(2)`

### DIFF
--- a/src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
+++ b/src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
@@ -120,8 +120,8 @@ int InitMaxAcceptQueueSize() {
   max_accept_queue_size = n;
 
   if (max_accept_queue_size < MIN_SAFE_ACCEPT_QUEUE_SIZE) {
-    LOG(INFO) << "Suspiciously small accept queue (" << max_accept_queue_size
-              << ") will probably lead to connection drops";
+    LOG(WARNING) << "Suspiciously small accept queue (" << max_accept_queue_size
+                 << ") will probably lead to connection drops";
   }
   return max_accept_queue_size;
 }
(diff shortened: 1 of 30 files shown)


markdroth:
This seems a little dangerous as a mechanical change.  Some of these cases are things that we've always logged by default and probably want to continue doing so.  In particular, I'm concerned about some of the security code, which for historical reasons is not written to return errors to the caller but only logs them, so there's really no other way to determine what the actual underlying problem was.

src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
```diff
@@ -120,8 +120,8 @@ int InitMaxAcceptQueueSize() {
   max_accept_queue_size = n;
 
   if (max_accept_queue_size < MIN_SAFE_ACCEPT_QUEUE_SIZE) {
-    LOG(INFO) << "Suspiciously small accept queue (" << max_accept_queue_size
-              << ") will probably lead to connection drops";
+    LOG(WARNING) << "Suspiciously small accept queue (" << max_accept_queue_size
+                 << ") will probably lead to connection drops";
   }
```
tanvi-jagtap, line 125: This may not be so bad. Unless you are sure about this, we should avoid making it a warning,

src/core/lib/iomgr/tcp_server_utils_posix_common.cc
```diff
@@ -72,8 +72,9 @@ static void init_max_accept_queue_size(void) {
   s_max_accept_queue_size = n;
 
   if (s_max_accept_queue_size < MIN_SAFE_ACCEPT_QUEUE_SIZE) {
-    LOG(INFO) << "Suspiciously small accept queue (" << s_max_accept_queue_size
-              << ") will probably lead to connection drops";
+    LOG(WARNING) << "Suspiciously small accept queue ("
+                 << s_max_accept_queue_size
+                 << ") will probably lead to connection drops";
   }
```
tanvi-jagtap, line 78: Same as above. Changing INFO to WARNING feels out of place as a part of a log noise reduction PR.

src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
tanvi-jagtap: Changing INFO to WARNING feels out of place as a part of a log noise reduction PR.

yashykt:
> This seems a little dangerous as a mechanical change. Some of these cases are things that we've always logged by default and probably want to continue doing so. In particular, I'm concerned about some of the security code, which for historical reasons is not written to return errors to the caller but only logs them, so there's really no other way to determine what the actual underlying problem was.

@markdroth Note that I'm only changing `LOG(INFO)` logs. Previously, we only logged ERROR logs by default if `GRPC_VERBOSITY` was not mentioned. This means that by default these were not logged. I want to preserve that default behavior.

src/core/lib/event_engine/posix_engine/posix_engine_listener_utils.cc
yashykt: reverting

src/core/lib/iomgr/tcp_server_utils_posix_common.cc
yashykt: reverting

yashykt:
I'll revert security changes and move them to a separate PR.

PR #37266, opened 2024-07-20
[security] Change security level classification for local connections

erm-g:
Set security level to NONE for LOCAL_TCP type, Privacy&Integrity for UDS.


src/core/lib/security/security_connector/local/local_security_connector.cc
```diff
@@ -143,9 +143,13 @@ void local_check_peer(tsi_peer peer, grpc_endpoint* ep,
   }
   if (peer.properties != nullptr) gpr_free(peer.properties);
   peer.properties = new_properties;
-  // TODO(yihuazhang): Set security level of local TCP to TSI_SECURITY_NONE.
-  const char* security_level =
-      tsi_security_level_to_string(TSI_PRIVACY_AND_INTEGRITY);
+  // Set security level to NONE for TCP type, privacy&integrity otherwise.
+  const char* security_level;
+  if (type == LOCAL_TCP) {
```
matthewstevenson88, line 148: Right now this isn't a big problem because there are only 2 types, but to be more future-proof: can the default be "none" rather than "privacy & integrity"? Said differently, can we re-arrange this to:

```
if (type == UDS) {
  security_level = tsi_security_level_to_string(TSI_PRIVACY_AND_INTEGRITY);
} else {
  security_level = tsi_security_level_to_string(TSI_SECURITY_NONE);
}
```

src/core/lib/security/security_connector/local/local_security_connector.cc
```diff
@@ -262,6 +266,8 @@ class grpc_local_server_security_connector final
 };
 }  // namespace
 
+// just pass local and uds in the format of Chttp2FullstackLocalIpv4, Chttp2FullstackLocalIpv6, Chttp2FullstackLocalUds, check for results
```
matthewstevenson88, line 269: Please remove commented out code.

src/core/lib/security/security_connector/local/local_security_connector.cc
matthewstevenson88, line 148: We need to guard the new functionality with a gRPC experiment, so we can safely roll it out internally. 

See https://github.com/grpc/grpc/pull/37035 as a recent example of how to create an experiment and use it in your code.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+  const grpc_auth_property* prop = grpc_auth_property_iterator_next(&it);
+  ASSERT_NE(prop, nullptr);
+  EXPECT_STREQ(prop->value, tsi_security_level_to_string(level));
+  connector.reset();
+  auth_context.reset();
+  grpc_server_credentials_release(server_creds);
+}
+
+//
+// Tests for Certificate Providers in ChannelSecurityConnector.
```
matthewstevenson88, line 135: Please remove this comment, as it is not relevant here.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+//
+
+TEST_F(LocalSecurityConnectorTest, CheckUDSType) {
+//  auto server_creds = grpc_local_server_credentials_create(LOCAL_TCP);
```
matthewstevenson88, line 139: nit: Please remove commented-out code.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+#include "src/core/lib/security/credentials/tls/tls_credentials.h"
+#include "src/core/tsi/transport_security.h"
+#include "test/core/test_util/test_config.h"
+#include "test/core/test_util/tls_utils.h"
+#include "googletest/include/gtest/gtest.h"
+
+namespace grpc_core {
+namespace testing {
+
+class LocalSecurityConnectorTest : public ::testing::Test {
```
matthewstevenson88, line 50: This is not needed. You can remove this and use `TEST` instead of `TEST_F` below.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+
+static absl::string_view me_get_local_address_unix(grpc_endpoint* /*ep*/) {
+  return "unix:";
+}
+
+static int me_get_fd(grpc_endpoint* /*ep*/) { return -1; }
+
+static bool me_can_track_err(grpc_endpoint* /*ep*/) { return false; }
+
+static const grpc_endpoint_vtable vtable_unix = {me_read,
```
matthewstevenson88, line 84: To simplify the code, I think we can just use `nullptr` for all of the functions that we do not use. Similarly for the other vtable.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+                                            me_get_peer,
+                                            me_get_local_address_local,
+                                            me_get_fd,
+                                            me_can_track_err};
+
+static void check_tsi_security_level(grpc_local_connect_type connect_type,
+                                 tsi_security_level level, grpc_endpoint ep) {
+  grpc_server_credentials* server_creds = grpc_local_server_credentials_create(connect_type);
+  ChannelArgs args;
+  EXPECT_NE(server_creds, nullptr);
```
matthewstevenson88, line 114: Please either remove `EXPECT_NE` (because `grpc_local_server_credentials_create` never returns null) or change it to `ASSERT_NE`.

If we decide to leave it, please move it to the next line immediately after `grpc_local_server_credentials_create` is called.

Similarly for `EXPECT_NE(connector, nullptr);`

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+                                            me_add_to_pollset,
+                                            me_add_to_pollset_set,
+                                            me_delete_from_pollset_set,
+                                            me_destroy,
+                                            me_get_peer,
+                                            me_get_local_address_local,
+                                            me_get_fd,
+                                            me_can_track_err};
+
+static void check_tsi_security_level(grpc_local_connect_type connect_type,
```
matthewstevenson88, line 110: nit: Prefer an anonymous namespace rather than using `static` throughout.

matthewstevenson88, line 110: nit: Prefer C++-style naming here and also a more explicit name, e.g. `ExpectSecurityLevelOfLocalEndpoint`.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+
+  RefCountedPtr<grpc_auth_context> auth_context;
+  connector->check_peer(peer, &ep, args, &auth_context, nullptr);
+  tsi_peer_destruct(&peer);
+  auto it = grpc_auth_context_find_properties_by_name(
+      auth_context.get(), GRPC_TRANSPORT_SECURITY_LEVEL_PROPERTY_NAME);
+  const grpc_auth_property* prop = grpc_auth_property_iterator_next(&it);
+  ASSERT_NE(prop, nullptr);
+  EXPECT_STREQ(prop->value, tsi_security_level_to_string(level));
+  connector.reset();
```
matthewstevenson88, line 129: I think `connector.reset()` and `auth_context.reset()` may be unnecessary, as both are shared pointers.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+  auth_context.reset();
+  grpc_server_credentials_release(server_creds);
+}
+
+//
+// Tests for Certificate Providers in ChannelSecurityConnector.
+//
+
+TEST_F(LocalSecurityConnectorTest, CheckUDSType) {
```
matthewstevenson88, line 138: nit: Prefer a more descriptive test name than `CheckUDSType`, e.g. `SecurityLevelOfUdsConnection`. Similarly below.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+#include <grpc/support/alloc.h>
+#include <grpc/support/log.h>
+#include <grpc/support/string_util.h>
+
+#include "src/core/lib/channel/channel_args.h"
+#include "src/core/lib/config/config_vars.h"
+#include "src/core/lib/gprpp/crash.h"
+#include "src/core/lib/gprpp/unique_type_name.h"
+#include "src/core/lib/security/context/security_context.h"
+#include "src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h"
```
matthewstevenson88, line 39: There seem to be several unnecessary includes, e.g. the TLS ones or "googletest/include/gtest/gtest.h". Please remove these.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,165 @@
...
+    };
+  check_tsi_security_level(LOCAL_TCP, TSI_SECURITY_NONE, ep);
+}
+
+}  // namespace testing
+}  // namespace grpc_core
+
+int main(int argc, char** argv) {
+  grpc::testing::TestEnvironment env(&argc, argv);
+  grpc_core::ConfigVars::Overrides overrides;
```
matthewstevenson88, line 158: I think `grpc_core::ConfigVars::Overrides overrides;` and `grpc_core::ConfigVars::SetOverrides(overrides);` are unnecessary, unless I'm missing something.

matthewstevenson88 (commented):
Thanks for the PR @erm-g! Please also address the failing tests.

src/core/lib/security/security_connector/local/local_security_connector.cc
erm-g: Ok, would it be better to use `switch with default` construction for it?

test/core/security/local_security_connector_test.cc
erm-g: I saw 2 styles in surrounding tests - with test related class and without. Which one is preferred? I liked the idea of keeping PrintAuthContext debugging function under this class, but dropped it because wasn't sure about the performance.

test/core/security/local_security_connector_test.cc
erm-g: Agreed for removing `EXPECT_NE(server_creds)`
For the connector one - I think technically you're right (server_creds is not null for this test) -
https://github.com/grpc/grpc/blob/b1896a2136e9f60e026daeaaf393cb8cb143be72/src/core/lib/security/security_connector/local/local_security_connector.cc#L298
but in such cases is it better to drop or to use `ASSERT_NE` ?

test/core/security/local_security_connector_test.cc
erm-g: Surrounding tests use both styles - is camel case a newer pattern?

test/core/security/local_security_connector_test.cc
erm-g: GitHub test suite was indicating a memory leak without it - is there a better way to fix it?

test/core/security/local_security_connector_test.cc
erm-g: SInce it's effectively the name of this test, would `CheckSecurityLevelOfUdsConnection` / `SecurityLevelOfTcpConnection` be better?

test/core/security/local_security_connector_test.cc
erm-g: Removed

test/core/security/local_security_connector_test.cc
erm-g: Cleaned up

test/core/security/local_security_connector_test.cc
erm-g: Yeah that was an ugly part of the test, thanks for showing a better solution!

test/core/security/local_security_connector_test.cc
erm-g: Cleaned up

test/core/security/local_security_connector_test.cc
erm-g: Changed the comment to `Tests for grpc_local_channel_security_connector.` - in surrounding tests it's used as a separator between test util code and tests themselves. Is it better to remove it completely?

src/core/lib/security/security_connector/local/local_security_connector.cc
erm-g: Cleaned up

src/core/lib/security/security_connector/local/local_security_connector.cc
matthewstevenson88: Either is fine by me. Switch is probably more future-proof, good idea.

test/core/security/local_security_connector_test.cc
matthewstevenson88: Thanks for fixing the wording. I generally omit this comment, and only include it if there are a huge number of very different tests in the same test file. I'll leave it to you whether you want to include it.

test/core/security/local_security_connector_test.cc
matthewstevenson88: Having this test class (rather than using the autogenerated one) is unnecessary unless you have setup that you need to do (which we don't in this case). If you see this in other files, I'm guessing what happened is: there used to be setup, the setup got deleted, and they didn't bother to also delete the test class.

test/core/security/local_security_connector_test.cc
matthewstevenson88: It's a bit confusing unfortunately. :)

The history here: originally all of these files were written in C and then 3-4 years ago we were allowed to use C++. Over time, whenever changes were made to the existing files, we tried to C++-ify the parts of the code that we're being modified. This led to the situation where a lot of the files have a mix of C-style and C++-style, which is what you alluded to above.

For new files such as this one, we should try to use C++-style only.

test/core/security/local_security_connector_test.cc
matthewstevenson88: No it's fine to leave it, if the tests flag it as a memory leak. The reason I was confused is because `RefCountedPtr` calls `reset()` in the destructor, so I thought it was unnecessary to do so here. But I wonder if the order in which `connector` and `auth_context` are freed matters? Doesn't matter, just curious. :)

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,114 @@
...
+                                            nullptr,
+                                            nullptr,
+                                            nullptr,
+                                            nullptr,
+                                            nullptr,
+                                            me_get_local_address_local,
+                                            nullptr,
+                                            nullptr};
+
+static void check_tsi_security_level(grpc_local_connect_type connect_type,
```
matthewstevenson88, line 63: This test only checks the security level for the server credentials - can we have an analogous test for the channel credentials?

test/core/security/local_security_connector_test.cc
matthewstevenson88: SGTM.

test/core/security/local_security_connector_test.cc
matthewstevenson88: Np!

test/core/security/local_security_connector_test.cc
matthewstevenson88: Took a look at the code and it seems that `connector` could be nullptr, so let's keep the check and use `ASSERT_NE`.

src/core/lib/experiments/rollouts.yaml
```diff
@@ -75,6 +75,8 @@
     windows: true
 - name: free_large_allocator
   default: false
+- name: local_connector_secure
+  default: true
```
matthewstevenson88, line 79: Please make the default `false`. We will need to roll this out carefully. :)

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,152 @@
...
+                                                  nullptr,
+                                                  nullptr,
+                                                  me_get_local_address_local,
+                                                  nullptr,
+                                                  nullptr};
+
+void CheckSecurityLevelForServer(grpc_local_connect_type connect_type,
+                                            tsi_security_level level,
+                                            grpc_endpoint ep) {
+  grpc_server_credentials
```
matthewstevenson88, line 64: micro-nit: Is there an empty space between `grpc_server_credentials` and `*`? If so, can we remove the space, so that it looks like `grpc_server_credentials*`?

Similarly for the channel test.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,152 @@
...
+#include "src/core/tsi/transport_security.h"
+#include "test/core/test_util/test_config.h"
+#include "src/core/client_channel/client_channel_filter.h"
+#include "include/grpc/impl/grpc_types.h"
+
+namespace grpc_core {
+namespace testing {
+namespace {
+
+absl::string_view me_get_local_address_unix(grpc_endpoint* /*ep*/) {
```
matthewstevenson88, line 31: nit: Can we use C++ naming for this function and its local equivalent? E.g. "GetLocalUnixAddress"?

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,152 @@
...
+  return "unix:";
+}
+
+const grpc_endpoint_vtable vtable_unix = {nullptr,
```
matthewstevenson88, line 35: nit: Since this is a constant, the variable name can be written as `kUnixEndpointVtable`. Similarly for the local version.

matthewstevenson88 (commented):
Please add release notes and address the failing tests.

sreenithi:
@erm-g Does this count as a breaking change? if so, it will need to be added to the release notes.

erm-g:
@sreenithi It's protected by an experiment but I'd say yes. I'm not sure which release we're targeting for it, so I think it's to early to put it in the notes.

test/core/end2end/end2end_tests.h
```diff
@@ -683,6 +684,11 @@ class CoreEnd2endTestRegistry {
     GTEST_SKIP() << "Disabled for initial v3 testing";      \
   }
 
+#define SKIP_IF_LOCAL_CREDS()                                   \
+  if (GetParam()->feature_mask & FEATURE_MASK_IS_LOCAL_CREDS) { \
```
matthewstevenson88, line 688: Since we're skipping only for local TCP and not local UDS, should we rename the feature mask to `FEATURE_MASK_IS_LOCAL_TCP_CREDS` and the variable to `SKIP_IF_LOCAL_TCP_CREDS`?

matthewstevenson88 (approved):

test/core/end2end/tests/call_creds.cc
```diff
@@ -264,11 +264,17 @@ CORE_END2END_TEST(PerCallCredsOnInsecureTest,
 }
 
 CORE_END2END_TEST(PerCallCredsTest, RequestResponseWithPayloadAndCallCreds) {
+  if (IsLocalConnectorSecureEnabled()) {
+    SKIP_IF_LOCAL_TCP_CREDS();
```
markdroth, line 268: Why do we need a special case here for local TCP credentials when we don't seem to have one for insecure creds?  Presumably these tests also don't work for insecure creds, but we don't have a special case here to exclude insecure creds.  That implies to me that there's some other mechanism to exclude these tests for insecure creds, and I'm wondering why we can't use the same mechanism for both.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,139 @@
...
+  return "ipv4:127.0.0.1:12667";
+}
+
+const grpc_endpoint_vtable kTcpEndpointVtable = {
+    nullptr, nullptr, nullptr, nullptr,
+    nullptr, nullptr, nullptr, GetLocalTcpAddress,
+    nullptr, nullptr};
+
+void CheckSecurityLevelForServer(grpc_local_connect_type connect_type,
+                                 tsi_security_level level, grpc_endpoint ep) {
```
markdroth, line 49: Please pass the endpoint in as either a pointer or a reference, not by value.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,139 @@
...
+  const grpc_auth_property* prop = grpc_auth_property_iterator_next(&it);
+  ASSERT_NE(prop, nullptr);
+  EXPECT_STREQ(prop->value, tsi_security_level_to_string(level));
+  connector.reset();
+  auth_context.reset();
+  grpc_server_credentials_release(server_creds);
+}
+
+void CheckSecurityLevelForChannel(grpc_local_connect_type connect_type,
+                                  tsi_security_level level, grpc_endpoint ep) {
```
markdroth, line 73: Please pass the endpoint in as either a pointer or a reference, not by value.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,139 @@
...
+absl::string_view GetLocalTcpAddress(grpc_endpoint* /*ep*/) {
+  return "ipv4:127.0.0.1:12667";
+}
+
+const grpc_endpoint_vtable kTcpEndpointVtable = {
+    nullptr, nullptr, nullptr, nullptr,
+    nullptr, nullptr, nullptr, GetLocalTcpAddress,
+    nullptr, nullptr};
+
+void CheckSecurityLevelForServer(grpc_local_connect_type connect_type,
```
markdroth, line 48: The pattern of putting expectations inside helper functions makes tests much harder to debug, because when the expectation fails, the line number of the failure makes it hard to see exactly which test is failing.  Instead, I suggest changing these functions to return the actual security level, and then having the test case set the expectation on the returned value.

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,139 @@
...
+
+TEST(LocalSecurityConnectorTest, CheckSecurityLevelOfUdsConnectionServer) {
+  grpc_endpoint ep;
+  ep.vtable = &kUnixEndpointVtable;
+  CheckSecurityLevelForServer(UDS, TSI_PRIVACY_AND_INTEGRITY, ep);
+}
+
+TEST(LocalSecurityConnectorTest, SecurityLevelOfTcpConnectionServer) {
+  if (!IsLocalConnectorSecureEnabled()) {
+    return;
```
markdroth, line 106: Instead of skipping the entire test when the experiment is not enabled, I suggest checking for the appropriate expected value either way.

```
CheckSecurityLevelForServer(
    LOCAL_TCP,
    IsLocalConnectorSecureEnabled() ? TSI_SECURITY_NONE : TSI_PRIVACY_AND_INTEGRITY,
    ep);
```

test/core/security/local_security_connector_test.cc
```diff
@@ -0,0 +1,139 @@
...
+
+TEST(LocalSecurityConnectorTest, CheckSecurityLevelOfUdsConnectionChannel) {
+  grpc_endpoint ep;
+  ep.vtable = &kUnixEndpointVtable;
+  CheckSecurityLevelForChannel(UDS, TSI_PRIVACY_AND_INTEGRITY, ep);
+}
+
+TEST(LocalSecurityConnectorTest, SecurityLevelOfTcpConnectionChannel) {
+  if (!IsLocalConnectorSecureEnabled()) {
+    return;
```
markdroth, line 121: Same here: Let's check the value either way.

src/core/lib/experiments/experiments.yaml
```diff
@@ -91,6 +91,11 @@
   expiry: 2024/12/01
   owner: alishananda@google.com
   test_tags: [resource_quota_test]
+- name: local_connector_secure
+  description: Local Security Connector uses TSI_PRIVACY_AND_INTEGRITY only for UDS.
```
markdroth, line 95: Might be a good idea for this to be re-worded to indicate what's actually changing:

Local security connector uses TSI_SECURITY_NONE for LOCAL_TCP connections.

markdroth (commented):
Sorry for the delay on this!

test/core/end2end/tests/call_creds.cc
erm-g: We discussed offline with @ctiller that it's ok to leave it as is for now (Craig, pls correct me if I'm wrong). Once experiment is done, we can reorganize the suites.

src/core/BUILD
```diff
@@ -4118,9 +4118,9 @@ grpc_cc_library(
         "//:promise",
         "//:ref_counted_ptr",
         "//:sockaddr_utils",
-        "//src/core:experiments",
         "//:tsi_base",
         "//:uri",
+        "//src/core:experiments",
```
markdroth, line 4123: This should just be "experiments", since we're already in this directory.

markdroth (approved):

PR #37274, opened 2024-07-22
Added excessive logging threshold test

sourabhsinghs:
* Added excessive logging threshold tests
* Excessive logs sample:

```
Warning: Excessive error output detected (7 lines):
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR

I0000 00:00:1724852923.140229      45 chttp2_transport.cc:1182] ipv6:%5B::1%5D:45545: Got goaway [2] err=UNAVAILABLE:GOAWAY received; Error code: 2; Debug Text: Cancelling all calls {file:"src/core/ext/transport/chttp2/transport/chttp2_transport.cc", file_line:1171, created_time:"2024-08-28T13:48:43.140192497+00:00", http2_error:2, grpc_status:14}
```


bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,45 @@
...
+LOGGING_ERROR_THRESHOLD = 50
+LOGGING_OUT_THRESHOLD = 20
+
+if __name__ == "__main__":
+ 
+    if len(sys.argv) != 2:
+        print(f"USAGE: {sys.argv[0]} TARGET_MODULE", file=sys.stderr)
+
+    target_module = sys.argv[1]
+    command = [sys.executable, "./src/python/grpcio_tests/tests/unit/_single_module_tester.py", target_module]
```
XuanWang-Amos, line 29: We should get the path using `os.path`: https://github.com/grpc/grpc/blob/bb2e7e44fca2fd466247b6856245f7ee32f79825/bazel/_gevent_test_main.py#L75

bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,45 @@
...
+            client_stderr.seek(0)
+
+            stdout_count = len(client_stdout.readlines()) 
+            stderr_count = len(client_stderr.readlines())
+
+            if result.returncode != 0:
+                sys.exit('Test failure')
+            
+            if stderr_count > LOGGING_ERROR_THRESHOLD or stdout_count > LOGGING_OUT_THRESHOLD:
+                sys.exit('Test failure')
```
XuanWang-Amos, line 45: Can we add more text to explain why test failed?

bazel/logging_threshold_test.bzl
```diff
@@ -0,0 +1,75 @@
+# Copyright 2023 The gRPC Authors
```
XuanWang-Amos, line 1: [Minor] Should be 2024.

src/python/grpcio_tests/tests/unit/_single_module_tester.py
```diff
@@ -0,0 +1,56 @@
...
+
+if __name__ == "__main__":
+
+    if len(sys.argv) != 2:
+        print(f"USAGE: {sys.argv[0]} TARGET_MODULE", file=sys.stderr)
+
+    target_module = sys.argv[1]
+
+    test_kwargs = {}
+    test_kwargs["verbosity"] = 3
```
XuanWang-Amos, line 49: Shouldn't we use 0 here?

bazel/_logging_threshold_test_main.py
gnossen: Let's use an assert statement here instead as well. `sys.exit` doesn't leave a very readable trace and stops us from executing the other tests.

src/python/grpcio_tests/tests/unit/_single_module_tester.py
```diff
@@ -0,0 +1,56 @@
...
+            raise AssertionError("Expected only 1 test module. Found {}".format(tests))
+        self.suite.addTest(tests[0])
+
+    def loadTestsFromNames(self, names: Sequence[str], module: str = None) -> unittest.TestSuite:
+        return self.suite
+
+if __name__ == "__main__":
+
+    if len(sys.argv) != 2:
+        print(f"USAGE: {sys.argv[0]} TARGET_MODULE", file=sys.stderr)
```
gnossen, line 44: Should we add a `sys.exit(1)` in this conditional to terminate on invalid flags?

bazel/_logging_threshold_test_main.py
gnossen: It's not conventional to encode paths like this with a test harness. Instead, let's supply this as a CLI flag to this file and provide it via Bazel. This gives us a single source of truth for the location of the entrypoint for the subprocesses. This will mean supplying the `_single_module_tester.py` file as a `data` arg to the `native.py_test` target in `bazel/logging_threshold_test.bzl` as well as to its `args` arg. ([reference](https://bazel.build/reference/be/python#py_test))



I think it's also probably cleaner to move `_single_module_test.py` up to the `bazel/` directory because this file will never be used in the `setup.py` based tests and it may mess with module discovery for that test harness. This will mean backing out the changes to `src/python/grpcio_tests/tests/unit/BUILD.bazel`

bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import tempfile
+import sys
+import subprocess
+
+LOGGING_ERROR_THRESHOLD = 50
+LOGGING_OUT_THRESHOLD = 20
```
gnossen, line 21: Switching to a zero logs or "skipped" model is a blocker for merging this PR.

bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,45 @@
...
+
+if __name__ == "__main__":
+ 
+    if len(sys.argv) != 2:
+        print(f"USAGE: {sys.argv[0]} TARGET_MODULE", file=sys.stderr)
```
gnossen, line 26: Same comment. Should this conditional include `sys.exit(1)`?

gnossen (commented):
Overall. this structurally looks pretty good. The main blocker here is going to be the exact conditions for the assertion.

bazel/_logging_threshold_test_main.py
sourabhsinghs: made updates as suggested. 
- moved `_single_module_test.py` to `bazel/`
- reverted changes in `src/python/grpcio_tests/tests/unit/BUILD.bazel`
- `_single_module_tester.py` passed as data and arg.

bazel/_logging_threshold_test_main.py
sourabhsinghs: - Added more explanation failure.
- Also printing the exact undesired output lines
- Removed sys.exit(1) . And output above is a warning.

src/python/grpcio_tests/tests/unit/_single_module_tester.py
sourabhsinghs: reduced verbosity to 0

bazel/_logging_threshold_test_main.py
sourabhsinghs: Added sys.exit(1)

bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,89 @@
...
+
+import os
+import re
+import subprocess
+import sys
+import tempfile
+
+_OK_TEST_REGEX = r"^-+.*Ran ([\d]+) tests* in ([\d.]+)s.*OK(?: \(skipped=(\d+)\))?\n$"
+
+_SKIP_TESTS = [
```
XuanWang-Amos, line 23: We should add some comments explain why we're skipping them.

bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,89 @@
...
+if __name__ == "__main__":
+    if len(sys.argv) != 3:
+        print(f"USAGE: {sys.argv[0]} TARGET_MODULE", file=sys.stderr)
+        sys.exit(1)
+
+    test_script = sys.argv[1]
+    target_module = sys.argv[2]
+
+    if target_module in _SKIP_TESTS:
+        sys.exit(0)
```
XuanWang-Amos, line 46: [Minor] Can we add a log to print the name of skipped tests?

src/python/grpcio_tests/tests/unit/_signal_handling_test.py
```diff
@@ -208,7 +208,6 @@ def testStreamingHandlerWithException(self):
                 self._handler.await_connected_client()
                 client.send_signal(signal.SIGINT)
                 client.wait()
-                print(_read_stream(client_stderr))
```
XuanWang-Amos, line 211: Is this intentional?

XuanWang-Amos:
In general looks good, thanks for working on this! Can you take a look of the failed tests and re-trigger them if necessary?

src/python/grpcio_tests/tests/unit/_signal_handling_test.py
sourabhsinghs: Yes this was the only line being output in stdout. all other tests have 0 stdout

gnossen: Instead of deleting, can we capture this to a variable and conditionally log in case the assertion on the next line fires?

bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+import re
+import subprocess
+import sys
+import tempfile
+
+_OK_TEST_REGEX = r"^-+.*Ran ([\d]+) tests* in ([\d.]+)s.*OK(?: \(skipped=(\d+)\))?\n$"
+
+_SKIP_TESTS = [
+    "_rpc_part_1_test",
```
gnossen, line 24: Skipping the `_rpc_part_1_test` _and_ the `_rpc_part_2_test` seems a bit excessive. These are the most basic unit tests we have. If we can't run ensure that logging regressions don't happen in those, I'm not sure how much value these tests have.

bazel/_logging_threshold_test_main.py
sourabhsinghs: we are skipping the tests with warning for now

bazel/_logging_threshold_test_main.py
sourabhsinghs: Added

bazel/_logging_threshold_test_main.py
sourabhsinghs: I'll Actively work to get logging Issue with _rpc_part_1 and _rpc_part_2 resolved .
We can move rest, so that logging tests start running for rest of tests.

bazel/_logging_threshold_test_main.py
```diff
@@ -0,0 +1,92 @@
...
+                for line in stderr_file:
+                    print(line)
+
+            if stdout_count > 0:
+                print(
+                    f"Warning: Unexpected output detected ({stdout_count} lines):"
+                )
+                stdout_file.seek(0)
+                for line in stdout_file:
+                    print(line)
```
XuanWang-Amos, line 92: [Minor] Missing new line.

bazel/_single_module_tester.py
```diff
@@ -0,0 +1,54 @@
...
+
+    target_module = sys.argv[1]
+    unittest_path = sys.argv[2]
+
+    loader = SingleLoader(target_module, unittest_path)
+    runner = unittest.TextTestRunner(verbosity=0)
+    result = runner.run(loader.suite)
+    
+    if not result.wasSuccessful():
+        sys.exit("Test failure.")
```
XuanWang-Amos, line 54: [Minor] Missing new line.

bazel/_logging_threshold_test_main.py
XuanWang-Amos: What's the current issue with `_rpc_part_1` and `_rpc_part_2`?

XuanWang-Amos (commented):
I don't have other comments now, can you also put some example logs to description so that we know what it looks like if those tests failed?

bazel/_logging_threshold_test_main.py
sourabhsinghs: @XuanWang-Amos  We have some logs still in output . Need to work with core team on those.

bazel/_single_module_tester.py
sourabhsinghs: Added new line

XuanWang-Amos (approved):
Thanks for working on this!

src/python/grpcio_tests/tests/unit/_signal_handling_test.py
```diff
@@ -208,8 +208,12 @@ def testStreamingHandlerWithException(self):
...
-                self.assertEqual(0, client.returncode)
+                client_stderr_output = _read_stream(client_stderr)
+                try:
+                    self.assertEqual(0, client.returncode)
```
gnossen, line 213: Nit: Instead of doing a try-except, you can just use the [`msg` kwarg](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual).

gnossen (approved):

src/python/grpcio_tests/tests/unit/_signal_handling_test.py
gnossen: Nevermind. I see the intention here.

PR #37302, opened 2024-07-24
[WIP]Add typings support python

sourabhsinghs:
* Adding typings to grpc python


src/python/grpcio/grpc/aio/_interceptor.py
```diff
@@ -111,9 +115,9 @@ class ClientCallDetails(
         wait_for_ready: An optional flag to enable :term:`wait_for_ready` mechanism.
     """
 
-    method: str
+    method: bytes
     timeout: Optional[float]
-    metadata: Optional[Metadata]
+    metadata: Optional[Sequence[Tuple[Union[str, bytes], Union[str, bytes]]]]
```
XuanWang-Amos, line 120: Can we use `MetadataType` here?

src/python/grpcio/grpc/aio/_interceptor.py
```diff
@@ -543,7 +550,7 @@ async def _proxy_writes_as_request_iterator(self):
             yield value
 
     async def _write_to_iterator_queue_interruptible(
-        self, request: RequestType, call: InterceptedCall
+        self, request: object, call: InterceptedCall
```
XuanWang-Amos, line 553: Maybe just use `Any` instead of `object`?

PR #37304, opened 2024-07-24
[Gpr_To_Absl_Logging] Convert some tracing macros to use << operator

tanvi-jagtap:
[Gpr_To_Absl_Logging] Convert some tracing macros to use << operator

### DIFF
--- a/src/core/lib/iomgr/ev_apple.cc
+++ b/src/core/lib/iomgr/ev_apple.cc
@@ -48,6 +48,8 @@
 #define GRPC_POLLING_TRACE(...)
 #endif  // NDEBUG
 
+#define GRPC_POLLSET_KICK_BROADCAST ((grpc_pollset_worker*)1)
+
 struct GlobalRunLoopContext {
   grpc_core::CondVar init_cv;
   grpc_core::CondVar input_source_cv;
--- a/src/core/lib/iomgr/ev_apple.cc
+++ b/src/core/lib/iomgr/ev_apple.cc
@@ -39,10 +39,14 @@
 #include "src/core/lib/gprpp/time_util.h"
 #include "src/core/lib/iomgr/ev_apple.h"
 
-#define GRPC_POLLING_TRACE() \
-  DVLOG_IF(2, GRPC_TRACE_FLAG_ENABLED(apple_polling)) << "(polling) "
-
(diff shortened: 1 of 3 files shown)


src/core/lib/iomgr/ev_posix.h
```diff
@@ -33,10 +35,8 @@
 #include "src/core/lib/iomgr/pollset_set.h"
 #include "src/core/lib/iomgr/wakeup_fd_posix.h"
 
-#define GRPC_FD_TRACE(format, ...)                                      \
-  if (GRPC_TRACE_FLAG_ENABLED(fd_trace)) {                              \
-    LOG(INFO) << "(fd-trace) " << absl::StrFormat(format, __VA_ARGS__); \
-  }
+#define GRPC_FD_TRACE() \
+  LOG_IF(INFO, GRPC_TRACE_FLAG_ENABLED(fd_trace)) << "(fd-trace) "
```
drfloob, line 39: nit: This macro should be deleted and replaced with `GRPC_TRACE_LOG(fd, INFO) << "(fd-trace) " <<`. We can do that in another pass.

src/core/lib/iomgr/ev_apple.cc
```diff
@@ -39,14 +39,8 @@
...
 
-#ifndef NDEBUG
-#define GRPC_POLLING_TRACE(format, ...)                              \
-  if (GRPC_TRACE_FLAG_ENABLED(apple_polling)) {                      \
-    VLOG(2) << "(polling) " << absl::StrFormat(format, __VA_ARGS__); \
-  }
-#else
-#define GRPC_POLLING_TRACE(...)
-#endif  // NDEBUG
+#define GRPC_POLLING_TRACE() \
```
drfloob, line 42: nit: This macro should be deleted and replaced with `GRPC_TRACE_VLOG(apple_polling, INFO) << "(polling) " <<`. We can do that in another pass.

drfloob (approved):

src/core/lib/iomgr/ev_apple.cc
tanvi-jagtap: Same as above . Changed the usage , but cant delete the macro - it messes up some old internal code which I will need to fix before I clean this up. Will do that in the next iteration.... This is going to be a long chain of PRs...

drfloob (approved):

PR #37320, opened 2024-07-26
[Gpr_To_Absl_Logging] Editing the Troubleshooting doc

tanvi-jagtap:
[Gpr_To_Absl_Logging] Editing the Troubleshooting doc

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -22,9 +22,11 @@ Example
 ./helloworld_application_using_grpc --v=-1 --minloglevel=3
 ```
 
-## GRPC_VERBOSITY (DEPRECATED)
+## GRPC_VERBOSITY
 
-[Environment Variables Overview](doc/environment_variables.md)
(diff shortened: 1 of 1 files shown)


TROUBLESHOOTING.md
```diff
@@ -22,9 +22,11 @@ Example
 ./helloworld_application_using_grpc --v=-1 --minloglevel=3
 ```
 
-## GRPC_VERBOSITY (DEPRECATED)
+## GRPC_VERBOSITY
 
-[Environment Variables Overview](doc/environment_variables.md)
+This is not recommended for gRPC C++. For gRPC C++ it is strongly recommended that you use absl to control gRPC verbosity. For other languages, GRPC_VERBOSITY will be supported. Users need to note that this will change settings for all libraries/binaries that use absl in the application. GRPC_VERBOSITY will alter the global absl settings, not just settings for gRPC.
```
yashykt, line 27: Thanks! Let's be specific about which languages we are talking about here. Python, ObjC, PHP, Ruby.

tanvi-jagtap: Done

yashykt (approved):

PR #37338, opened 2024-07-30
Remove win32 handling on grpc_ares_init

adyaksaw:
Purpose: Fix TSAN caused by early call to ares_library_cleanup while gRPC is still using ares

`ares_library_init` and `ares_library_cleanup` is not required to be called on non-windows. But if other library is calling these functions, then it will potentially introduce data race. Some example that I encounter: [Curl](https://github.com/curl/curl/blob/abdf612221db39f31abde0ef211bad28e272dde0/lib/asyn-ares.c#L152) calling `ares_library_cleanup` when gRPC is still processing request.

### DIFF
--- a/src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
+++ b/src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
@@ -62,6 +62,7 @@
 #include "src/core/lib/channel/channel_args.h"
 #include "src/core/lib/gprpp/debug_location.h"
 #include "src/core/lib/gprpp/host_port.h"
+#include "src/core/lib/gprpp/sync.h"
 #include "src/core/lib/gprpp/time.h"
 #include "src/core/lib/iomgr/error.h"
 #include "src/core/lib/iomgr/exec_ctx.h"
@@ -1179,8 +1180,10 @@ static void grpc_cancel_ares_request_impl(grpc_ares_request* r) {
 
 void (*grpc_cancel_ares_request)(grpc_ares_request* r) =
     grpc_cancel_ares_request_impl;
+grpc_core::Mutex ares_setup_mu_;
 
 grpc_error_handle grpc_ares_init(void) {
+  grpc_core::MutexLock lock(&ares_setup_mu_);
   int status = ares_library_init(ARES_LIB_INIT_ALL);
   if (status != ARES_SUCCESS) {
     return GRPC_ERROR_CREATE(
@@ -1189,6 +1192,9 @@ grpc_error_handle grpc_ares_init(void) {
   return absl::OkStatus();
 }
 
-void grpc_ares_cleanup(void) { ares_library_cleanup(); }
+void grpc_ares_cleanup(void) {
+  ares_library_cleanup();
+  grpc_core::MutexLock lock(&ares_setup_mu_);
+}
 
 #endif  // GRPC_ARES == 1
--- a/src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
+++ b/src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
@@ -1180,10 +1180,6 @@ static void grpc_cancel_ares_request_impl(grpc_ares_request* r) {
 void (*grpc_cancel_ares_request)(grpc_ares_request* r) =
     grpc_cancel_ares_request_impl;
 
-// ares_library_init and ares_library_cleanup are currently no-op except under
-// Windows. Calling them may cause race conditions when other parts of the
-// binary calls these functions concurrently.
-#ifdef GPR_WINDOWS
 grpc_error_handle grpc_ares_init(void) {
   int status = ares_library_init(ARES_LIB_INIT_ALL);
   if (status != ARES_SUCCESS) {
@@ -1194,9 +1190,5 @@ grpc_error_handle grpc_ares_init(void) {
 }
 
 void grpc_ares_cleanup(void) { ares_library_cleanup(); }
-#else
-grpc_error_handle grpc_ares_init(void) { return absl::OkStatus(); }
-void grpc_ares_cleanup(void) {}
-#endif  // GPR_WINDOWS
 
 #endif  // GRPC_ARES == 1
--- a/src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
+++ b/src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
@@ -55,7 +55,6 @@
 
 #include <grpc/support/alloc.h>
 #include <grpc/support/string_util.h>
-#include <grpc/support/sync.h>
 
 #include "src/core/lib/address_utils/parse_address.h"
 #include "src/core/lib/address_utils/sockaddr_utils.h"
@@ -1180,10 +1179,8 @@ static void grpc_cancel_ares_request_impl(grpc_ares_request* r) {
 
 void (*grpc_cancel_ares_request)(grpc_ares_request* r) =
     grpc_cancel_ares_request_impl;
-grpc_core::Mutex ares_setup_mu_;
 
 grpc_error_handle grpc_ares_init(void) {
-  grpc_core::MutexLock lock(&ares_setup_mu_);
   int status = ares_library_init(ARES_LIB_INIT_ALL);
   if (status != ARES_SUCCESS) {
     return GRPC_ERROR_CREATE(
@@ -1192,9 +1189,6 @@ grpc_error_handle grpc_ares_init(void) {
   return absl::OkStatus();
 }
 
-void grpc_ares_cleanup(void) {
-  ares_library_cleanup();
-  grpc_core::MutexLock lock(&ares_setup_mu_);
-}
+void grpc_ares_cleanup(void) { ares_library_cleanup(); }
 
 #endif  // GRPC_ARES == 1


yijiem:
Are you seeing TSAN race on Windows, if so how is this change fixes the race?

src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
```diff
@@ -1179,12 +1180,10 @@ static void grpc_cancel_ares_request_impl(grpc_ares_request* r) {
 
 void (*grpc_cancel_ares_request)(grpc_ares_request* r) =
     grpc_cancel_ares_request_impl;
+grpc_core::Mutex ares_setup_mu_;
 
-// ares_library_init and ares_library_cleanup are currently no-op except under
-// Windows. Calling them may cause race conditions when other parts of the
-// binary calls these functions concurrently.
-#ifdef GPR_WINDOWS
```
yijiem, line 1186: Why removing this ifdef? I don't think we want to call these functions on non-Windows platforms.

yijiem (commented):
So you are seeing TSAN error on Linux and you think this will cause the same issue on Windows, is that the rationale?

src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
```diff
@@ -1179,12 +1180,10 @@ static void grpc_cancel_ares_request_impl(grpc_ares_request* r) {
 
 void (*grpc_cancel_ares_request)(grpc_ares_request* r) =
     grpc_cancel_ares_request_impl;
+grpc_core::Mutex ares_setup_mu_;
```
yijiem, line 1183: I'm not sure adding a mutex in gRPC is the right thing to do. If an application doesn't use curl and gRPC together, it would not see this race and it doesn't need this mutex. I think application needs to make sure that `ares_library_init()` and `ares_library_cleanup()` are not called concurrently from different threads.

adyaksaw: I added the mutex to address the previous comment (`Calling them may cause race conditions when other parts of the binary calls these functions concurrently.`). Ares is not intended to be thread-safe (https://github.com/c-ares/c-ares/issues/449#issuecomment-996681949), so I think the previous comment make sense and we need to handle it here.

src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
adyaksaw: There are 2 parts answer to this question:

1. Consider this sequence of process in linux:
a. gRPC call `grpc_ares_init` (no-op)
b. curl call `ares_library_init` ([ares_initialized](https://github.com/c-ares/c-ares/blob/0e4b735a1060afa640b427c5c37df50f6014d991/src/lib/ares_library_init.c#L100C7-L100C23) incremented by one).
c. curl stop using ares, call `ares_library_cleanup`
d. gRPC process request, try to dns lookup using `grpc_dns_lookup_ares_continued` and eventually call [ares_init_options](https://github.com/grpc/grpc/blob/5638afba6eca657fe38c7566ac5a18004cec421f/src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc#L532)

On our case, when (d) is happening, it introduces TSAN because ares is in the middle of cleanup while gRPC is still processing request. We can prevent it by calling `ares_library_init` on gRPC, so ares will know that there's another library that is still using ares (i.e. `ares_initialized` is still non-zero) and prevent it from cleanup.

2. After digging the issue more, I found out that we actually want to call this functions not on Windows, but on Android ([Code](https://github.com/c-ares/c-ares/blob/0e4b735a1060afa640b427c5c37df50f6014d991/src/lib/ares_library_init.c#L142) and [c-ares maintainer comment](https://github.com/c-ares/c-ares/issues/475#issuecomment-1166627950))

adyaksaw:
> So you are seeing TSAN error on Linux and you think this will cause the same issue on Windows, is that the rationale?

Actually the reverse, we saw this issue on Linux and expect the issue is already fixed on Windows because of the ifdef (Hence the ifdef removal).

src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
yijiem: I'm not sure we need to call this on Android, gRPC never called `ares_library_init_jvm()` nor `ares_library_init_android()`, so `ares_library_cleanup_android()` just looks like noop in this case. We haven't been doing that and haven't heard any issue yet.

`ares_init_options()` will call `ares_library_initialized()`: https://github.com/c-ares/c-ares/blob/main/src/lib/ares_init.c#L273 which seems to be noop on non-Windows platforms, so why TSAN complains? Can you post the TSAN error?

src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
yijiem: Again I think it's the application's job to make sure it's not causing this race condition. Adding the mutex here is not necessary for users who do not use curl and gRPC in the same binary. In general, we don't want to add things in gRPC that would only address a specific use case or problem. This solution would actually add mutex ops to all gRPC users which is not necessary.

adyaksaw: I see. Actually adding mutex here also doesn't help when user use curl and gRPC in the same binary, it only help synchronize `ares_library_init()` in the same process (i.e. when gRPC call the `ares_library_init` in different thread). I saw that the previous comment noted their concern about it and tried to address it, but it seems it's not a problem in gRPC right now so it shouldn't be a problem to remove this mutex.

Removed.

src/core/resolver/dns/c_ares/grpc_ares_wrapper.cc
adyaksaw: Ack on no issue in the Android as of now.

`ares_library_initialized()` -> I don't think it's no-op. IIUC this line always return `ARES_SUCCESS` on non-Windows, that mean https://github.com/c-ares/c-ares/blob/main/src/lib/ares_init.c#L273 won't return early -> `ares_init_options` will continue as usual.

TSAN error can be seen on sponge2/306229f2-ef87-49bb-acb9-422486498e56

yijiem: Got it, so it's actually a race on accessing `ares_malloc`. With this change, wouldn't you still have the race on `ares_initialized`?

adyaksaw: Agree on `ares_initialized` is still not thread-safe, but I'm not sure the correct way to solve it without fixing c-ares, and the last attempt to fix it on c-ares is getting rejected on https://github.com/c-ares/c-ares/issues/475.

This PR only aims to prevent cleanup caused by `ares_library_cleanup` when gRPC is still using it (Which causes `ares_malloc` to be cleaned too early), and only fixing the thread-safety of `ares_initialized` won't fix this issue (At least that's what I tried).

yijiem: I think this change makes sense. You may need to add synchronization logic in your application around curl's init/cleanup and gRPC's init/shutdown or call them on the same thread to prevent this race.

@apolcyn WDYT?

adyaksaw: Friendly ping to this @apolcyn

yashykt: @apolcyn ping

yashykt: @apolcyn ping

apolcyn: Apologies for the long review delay.

I'm not crazy about this change because, while fixing one thing, it looks like it potentially introduces a *new* bug: a data race on `ares_initialized` (if something outside grpc is also initializing/de-initializing c-ares concurrently with grpc).

Given that there is a potential data race i.e. undefined behavior at play, I'd opt to keep the current code as is since at least the existing bug is staying in the same place. I.e. this way we're less likely to cause a *new* hard-to-debug issue.

Here are a few other ideas that I think would be safer:

1) Have the application itself invoke `ares_library_init` once at the very top of main, before using `curl` or `grpc`.

2) Somehow prevent `curl` from invoking `ares_library_cleanup`. E.g. do not invoke invoke `curl_global_cleanup` or compile `curl` without `CARES_HAVE_ARES_LIBRARY_CLEANUP` (I am looking at [here](https://github.com/curl/curl/blob/0cececef0f70f9f349b185952bba6d387602bdd4/lib/asyn-ares.c#L146)

3) Maybe a bit hacky, but if we could patch c-ares to simply not [write](https://github.com/c-ares/c-ares/blob/a70637c67ba2657adb54b077f4f8434f00ecbf83/src/lib/ares_library_init.c#L155) to shared globals in `ares_library_cleanup` (i.e. leave `ares_malloc` and friends at the current state and only override them if subsequent initialization calls for it.

4) c-ares patch c-ares/c-ares#475 (I realize this has been rejected already, just listing for completeness)

5) Do the reverse patch of this - instead of patching c-ares to invoke `ares_library_init`/`ares_library_cleanup`, patch curl to *not* invoke this (if not running on windows).

PR #37399, opened 2024-08-05
Update experiment expirations

ananda1066:


Vignesh2208 (approved):

src/core/lib/experiments/experiments.yaml
```diff
@@ -55,14 +55,14 @@
 - name: canary_client_privacy
```
ctiller, line 55: Just to check... when can we remove this one (and it's friends?)

ananda1066: I saw a change from the DDT team recently that still relied on the experiments, I think probably end of year - I'll check with them.

PR #37409, opened 2024-08-07
[FakeStatsPlugin] Add locks around adding/removing callbacks

yashykt:


test/core/test_util/fake_stats_plugin.h
```diff
@@ -426,6 +428,7 @@ class FakeStatsPlugin : public StatsPlugin {
   void TriggerCallbacks() {
     VLOG(2) << "FakeStatsPlugin[" << this << "]::TriggerCallbacks(): START";
     Reporter reporter(*this);
+    MutexLock lock(&callback_mu_);
```
ctiller, line 431: Holding a mutex whilst triggering callbacks is a bit of an anti-pattern... it'd be much better if we gathered the list of callbacks to call under a mutex, and then called them without the mutex being held.

yashykt: If we want to do that, we'll need to do some refcounting to make sure that the callbacks and the objects that they refer are still around after we let go of the mutex.. but it's possible

yashykt: actually, that seems like a good idea. Seems like an easier way to avoid deadlock issues that i'm facing

yijiem: yeah, if that is the issue, then the callback object and the actual callback that it holds need to be refcounted. The actual callback can take a refcount of the object (e.g. RlsLb) at the point of its creation. So that once we atomically extend the lifetime of the callback object the object's lifetime is also extended.

yijiem (approved):
Thanks Yash!

test/core/test_util/fake_stats_plugin.h
yashykt: given that opentelemetry still going to hold its mutex while invoking our callbacks, for safety, i'm going to have the fake stats plugin do the same till we can have opentelemetry not hold a global mutex before removing it.

yijiem (approved):

Issue #37352, opened 2024-08-01
MSVC Warning: unary minus operator on unsigned type


bebuch:
When I compile with MSVC (VS 2022) I get this Warning in nearly every compilation unit:

```
core/lib/gprpp/time.h(331): warning C4146: unary minus operator applied to unsigned type, result still unsigned
```

```cpp
return Duration::Milliseconds(
    time_detail::MillisAdd(lhs.milliseconds_after_process_epoch(),
                           -rhs.milliseconds_after_process_epoch())); // Here is the warning
```

https://github.com/grpc/grpc/blob/a83a7ff3545a9687fa319b739759e6e5b22d0628/src/core/lib/gprpp/time.h#L331

I'm not sure if this is just a warning that can be ignored or if this minus on unsigned is a real bug.

PR #37416, opened 2024-08-07
[Build] Decrease the number of warnings on Windows.

eugeneo:
Fixes: #37352


src/core/util/useful.h
```diff
@@ -133,19 +133,19 @@ inline uint32_t RoundUpToPowerOf2(uint32_t v) {
 
 // Return a value with only the lowest bit left on.
 GPR_ATTRIBUTE_ALWAYS_INLINE_FUNCTION inline uint8_t LowestOneBit(uint8_t x) {
-  return x & -x;
+  return x & ~x + 1;
```
ctiller, line 136: please add brackets so I don't have to remember weird precedence rules

eugeneo: Done

PR #37420, opened 2024-08-08
[Gpr_To_Absl_Logging] Remove gpr_should_log from the header file

tanvi-jagtap:
[Gpr_To_Absl_Logging] Remove gpr_should_log from the header file

### DIFF
--- a/src/core/util/log.cc
+++ b/src/core/util/log.cc
@@ -40,6 +40,8 @@ void gpr_unreachable_code(const char* reason, const char* file, int line) {
                    grpc_core::SourceLocation(file, line));
 }
 
+int absl_vlog2_enabled() { return ABSL_VLOG_IS_ON(2); }
+
 int gpr_should_log(gpr_log_severity severity) {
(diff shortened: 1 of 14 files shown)


src/core/util/log.cc
```diff
@@ -40,6 +40,8 @@ void gpr_unreachable_code(const char* reason, const char* file, int line) {
                    grpc_core::SourceLocation(file, line));
 }
 
+int absl_vlog2_enabled() { return ABSL_VLOG_IS_ON(2); }
```
apolcyn, line 43: LGTM for this and the ruby changes. I'm just wondering if this should be named something like `gpr_absl_vlog2_enabled` since it's in the global scope.

apolcyn (approved):

src/core/util/log.cc
tanvi-jagtap: Addressing that in this PR - with other related functions.
https://github.com/grpc/grpc/pull/37431

PR #37427, opened 2024-08-09
Use /usr/bin/base64 to decode because some Darwin users use GNU base64

antonysouthworth-halter:
When you have `coreutils` brew package installed, `base64` is usually going to be the coreutils version rather than macos system version at /usr/bin/base64. This causes package installation to fail with a pretty cryptic error message.

This seems to trip people up fairly often, see:

- https://stackoverflow.com/questions/72499392/pod-install-in-flutter-ios-project-gives-base64-invalid-input/72500565#72500565
- https://stackoverflow.com/questions/61603856/flutter-boringssl-grpc-unable-to-install
- https://github.com/grpc/grpc/issues/32489
- me, the last hour or so pulling hair and scratching head

It seems like this script is fundamentally assuming we are using the system-provided `base64` command anyway which is fair enough, so I think it makes sense to just call it directly.

### DIFF
--- a/src/objective-c/BoringSSL-GRPC.podspec
+++ b/src/objective-c/BoringSSL-GRPC.podspec
@@ -173,10 +173,11 @@ Pod::Spec.new do |s|
     # /src/boringssl/boringssl_prefix_symbols.h. Here we decode the content and inject the header to
     # the correct location in BoringSSL.
     case "$(uname)" in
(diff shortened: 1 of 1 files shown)


HannahShiSFB:
Thanks for the PR, could you sign the CLA and rebase to resolve the conflicts?

PR #37431, opened 2024-08-09
[Gpr_To_Absl_Logging] Remove gpr_log. Adding absl LOG wrappers.

tanvi-jagtap:
[Gpr_To_Absl_Logging] Remove gpr_log. Adding absl LOG wrappers

List of changes in this PR

1. Replacing all instances of gpr_log in PHP and RUBY with the new absl wrapper APIs. The replacement mapping is given below

gpr_log(GPR_ERROR, ...) 
	=> grpc_absl_log_error

gpr_log(GPR_INFO, ...)
	=> grpc_absl_log_info - Printing a simple message
	=> grpc_absl_log_info_int - Printing a message and a number
	=> grpc_absl_log_info_str - Printing 2 strings.

gpr_log(GPR_DEBUG, ...)
	=> grpc_absl_vlog - Printing a simple message
	=> grpc_absl_vlog_int - Printing a message and a number
	=> grpc_absl_vlog_str - Printing 2 strings.

Adding grpc_absl_vlog2_enabled() check around gpr_log(GPR_DEBUG, ...)

2. src/python/grpcio_observability/grpc_observability/observability_util.cc One instance of gpr_log to absl LOG replacement was missed earlier. Fixing that.

3. Deleting deprecated gpr stuff : gpr_log_severity , GPR_DEBUG , GPR_INFO , GPR_ERROR , gpr_log .

4. Adding new APIs for Ruby and PHP. These APIs are very simple wrappers around absl. 

5. Removing the legacy functions in platform specific log.cc files. These files are safe to delete now.

6. Fixing the allow list in banned_functions.py . This makes sure that these new wrappers don't get used all over the place by everyone. We carefully only allow list the PHP and RUBY files and allow the use of these wrappers. Everywhere else - using these wrappers should fail Sanity Tests.


apolcyn (approved):
ruby changes LGTM

stanley-cheung:
PHP changes LGTM

stanley-cheung (approved):

include/grpc/support/log.h
```diff
@@ -28,31 +28,39 @@
 extern "C" {
 #endif
 
-/**
- * Logging functions in this file are deprecated.
- * Please use absl ABSL_LOG instead.
- */
+#define GPR_LOCATION __FILE__, __LINE__
+
+/** Should only be used from gRPC PHP and RUBY. **/
```
yashykt, line 33: this requires a gRFC

tanvi-jagtap: I will try to find some other folder/file to put the new functions/macros - other than includes. Any pointers would help.
If that does not work, I will add some changes to the RFC.

yashykt: i'm unsure but check with alex/stanley if a src folder works

tanvi-jagtap: Alex said that it won't work any other way unless we put in non trivial amount of effort. I will prepare the RFC and upload it within 90 minutes.
I guess we have to wait a week or two before submitting the RFC and then this PR. 
If you have other concerns in the PR content please let me know - so I can fix them.
@yashykt

tanvi-jagtap: RFC https://github.com/grpc/proposal/pull/450

PR #37442, opened 2024-08-09
[EventEngine] Migrate httpcli to use EventEngine DNSResolver

yijiem:


src/core/util/http_client/httpcli.cc
```diff
@@ -173,7 +175,11 @@ HttpRequest::HttpRequest(
       pollent_(pollent),
       pollset_set_(grpc_pollset_set_create()),
       test_only_generate_response_(std::move(test_only_generate_response)),
-      resolver_(GetDNSResolver()) {
+      resolver_(
+          ChannelArgs::FromC(channel_args_)
+              .GetObjectRef<EventEngine>()
+              ->GetDNSResolver(EventEngine::DNSResolver::ResolverOptions())
+              .value()) {
```
drfloob, line 182: I don't think it's safe to crash here. That's what'll happen if the resolver returns a `Status` and you call `StatusOr<...>::value()`. 

It's likely safer to pull a resolver from the Engine in the constructor body, and fail all requests immediately if `GetDNSResolver` returns a status.

drfloob (commented):
Looks pretty good! Just a few fixes needed.

src/core/util/http_client/httpcli.cc
yijiem: Great catch! Done.

yijiem (commented):
Thanks AJ!

drfloob (approved):

PR #37456, opened 2024-08-12
[python] fix complaint raises by `ruff check` with xxx_pb2_grpc.py that generated by grpc_tools.

### LINKED ISSUE #37455: [python] `ruff chec` raises `F541 [*] f-string without any placeholders` with xxx_pb2_grpc.py that was generated by grpc_tools
$ python -m pip list|grep grpc
```
grpcio            1.65.4
grpcio-tools      1.65.4
```

Centos7/MacOS

N/A

### What did you do?
Please provide either 1) A unit test for reproducing the bug or 2) Specific steps for us to follow to reproduce the bug. If there’s not enough information to debug the problem, gRPC team may close the issue at their discretion. You’re welcome to re-open the issue once you have a reproduction.
To reproduce:
```bash
python3.11 -m venv venv
source venv/bin/activate
pip install --upgrade pip grpcio-tools ruff
echo 'syntax = "proto3";

service MsgService {
 rpc handler (MsgRequest) returns (MsgResponse){}
}

message MsgRequest {
  string name = 1;
}

message MsgResponse {
  uint64 id = 1;
}
' > msg.proto
python -m grpc_tools.protoc -I . --python_out=. --grpc_python_out=. msg.proto
ruff check msg_pb2_grpc.py
```

```
All checks passed!
```

### What did you see instead?
```
msg_pb2_grpc.py:23:11: F541 [*] f-string without any placeholders
Found 1 error.
[*] 1 fixable with the `--fix` option.
```

Make sure you include information that can help us debug (full error message, exception listing, stack trace, logs).

See [TROUBLESHOOTING.md](https://github.com/grpc/grpc/blob/master/TROUBLESHOOTING.md) for how to diagnose problems better.

NO

waketzheng:
Closes #37455

### DIFF
--- a/src/compiler/python_generator.cc
+++ b/src/compiler/python_generator.cc
@@ -692,7 +692,7 @@ bool PrivateGenerator::PrintPreamble(grpc_generator::Printer* out) {
   var["Package"] = config.grpc_package_root;
   out->Print(var, "import $Package$\n");
   if (config.grpc_tools_version.size() > 0) {
-    out->Print(var, "import warnings\n");
+    out->Print(var, "import warnings\n;from pathlib import Path\n");
   }
   if (generate_in_pb2_grpc) {
     out->Print("\n");
@@ -773,7 +773,7 @@ bool PrivateGenerator::PrintPreamble(grpc_generator::Printer* out) {
           out->Print(
               var,
               "f'The grpc package installed is at version {GRPC_VERSION},'\n"
-              "+ f' but the generated code in $Pb2GrpcFileName$_pb2_grpc.py "
+              "+ f' but the generated code in {Path(__file__).name} "
               "depends on'\n"
               "+ f' grpcio>={GRPC_GENERATED_VERSION}.'\n"
               "+ f' Please upgrade your grpc module to "


waketzheng:
Can be labelled as `release notes: no`

sourabhsinghs (approved):

XuanWang-Amos (approved):

src/compiler/python_generator.cc
```diff
@@ -773,7 +773,7 @@ bool PrivateGenerator::PrintPreamble(grpc_generator::Printer* out) {
           out->Print(
               var,
               "f'The grpc package installed is at version {GRPC_VERSION},'\n"
-              "+ f' but the generated code in $Pb2GrpcFileName$_pb2_grpc.py "
+              "+ f' but the generated code in {Path(__file__).name} "
```
sergiitk: No need to force f-string if it's not needed. The actual solution is much simpler.
```suggestion
              "+ ' but the generated code in $Pb2GrpcFileName$_pb2_grpc.py "
```

src/compiler/python_generator.cc
```diff
@@ -692,7 +692,7 @@ bool PrivateGenerator::PrintPreamble(grpc_generator::Printer* out) {
   var["Package"] = config.grpc_package_root;
   out->Print(var, "import $Package$\n");
   if (config.grpc_tools_version.size() > 0) {
-    out->Print(var, "import warnings\n");
+    out->Print(var, "import warnings\nfrom pathlib import Path\n");
```
sergiitk: This is not needed either
```suggestion
    out->Print(var, "import warnings\n");
```

sergiitk (changes requested):

src/compiler/python_generator.cc
sergiitk: updated the suggestion (copy-pasted a wrong thing initially)

PR #37460, opened 2024-08-12
[Python Otel] Manage call tracer life cycle use call arena.

XuanWang-Amos:
We're seeing segfault in Python CSM tests:
```
2024-08-03T09:49:45.720555997Z *** SIGSEGV received at time=1722678585 on cpu 0 ***
2024-08-03T09:49:45.721761998Z PC: @     0x7847ffd5c1c9  (unknown)  (unknown)
2024-08-03T09:49:45.722070502Z     @     0x7847fa309d8c         64  absl::lts_20240116::WriteFailureInfo()
2024-08-03T09:49:45.722175904Z     @     0x7847fa309a15        272  absl::lts_20240116::AbslFailureSignalHandler()
2024-08-03T09:49:45.722187675Z     @     0x7847ffc3d050       1592  (unknown)
2024-08-03T09:49:45.723432238Z     @     0x7847e97f9390  (unknown)  (unknown)
2024-08-03T09:49:45.723487349Z     @ ... and at least 1 more frames
2024-08-03T09:49:45.829702781Z [INFO  tini (1)] Spawned child process '/xds_interop_client' with pid '7'
2024-08-03T09:49:45.829766869Z [DEBUG tini (1)] Received SIGCHLD
2024-08-03T09:49:45.829778749Z [DEBUG tini (1)] Reaped child with pid: '7'
2024-08-03T09:49:45.829787070Z [INFO  tini (1)] Main child exited with signal (with signal 'Segmentation fault')
```

### The issue

After investigation, we found that the call tracer was deleted before `RecordEnd` was called.

### Why this fix

* To fix this, we decide to use arena to manage the life cycle of CallTracer.
* Since CallTracer was created in another shard object library (`grpcio_observability`) which don't have a dependency on grpc core, we can't use `grpc_core::Arena` directly when creating the call tracer.
* As a workaround, we created a wrapper class `ClientCallTracerWrapper` to wrap the CallTracer and created another core API `grpc_call_tracer_set_and_manage` so that we can manage the life cycle of CallTracer use the wrapper class.

### DIFF
--- a/src/core/lib/surface/call.h
+++ b/src/core/lib/surface/call.h
@@ -269,9 +269,9 @@ void grpc_call_tracer_set(grpc_call* call, grpc_core::ClientCallTracer* tracer);
 // When using this API, the tracer will be destroyed by grpc_call arena when
 // grpc_call is about to be destroyed. The caller of this API SHOULD NOT
 // manually destroy the tracer. This API is used by Python as a way of using
-// Arena to manage the lifetime of the call tracer. Python needs this API because
-// the tracer was created within a separate shared object library which doesn't
-// have access to core functions like arena->ManagedNew<>.
+// Arena to manage the lifetime of the call tracer. Python needs this API
+// because the tracer was created within a separate shared object library which
+// doesn't have access to core functions like arena->ManagedNew<>.
 void grpc_call_tracer_set_and_manage(grpc_call* call,
                                      grpc_core::ClientCallTracer* tracer);
 
--- a/src/core/lib/surface/call.h
+++ b/src/core/lib/surface/call.h
@@ -282,7 +282,7 @@ uint8_t grpc_call_is_client(grpc_call* call);
 
 class ClientCallTracerWrapper {
  public:
-  ClientCallTracerWrapper(grpc_core::ClientCallTracer* tracer)
+  explicit ClientCallTracerWrapper(grpc_core::ClientCallTracer* tracer)
       : tracer_(tracer) {}
 
  private:
--- a/src/core/lib/surface/call.h
+++ b/src/core/lib/surface/call.h
@@ -265,7 +265,13 @@ void grpc_call_log_batch(const char* file, int line, const grpc_op* ops,
 
 void grpc_call_tracer_set(grpc_call* call, grpc_core::ClientCallTracer* tracer);
 
-// Both sets and manages call tracer life cycle using ClientCallTracerWrapper.
+// Both sets and manages call tracer life cycle.
+// When using this API, the tracer will be destroyed by grpc_call arena when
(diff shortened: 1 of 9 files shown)


gnossen:
@XuanWang-Amos For reference, can you please link the PR where the new Core API was introduced into the description?

XuanWang-Amos:
> @XuanWang-Amos For reference, can you please link the PR where the new Core API was introduced into the description?

It's added in this PR.

src/core/lib/surface/call.h
```diff
@@ -265,6 +265,10 @@ void grpc_call_log_batch(const char* file, int line, const grpc_op* ops,
 
 void grpc_call_tracer_set(grpc_call* call, grpc_core::ClientCallTracer* tracer);
 
+// Both sets and manages call tracer life cycle using ClientCallTracerWrapper.
```
gnossen, line 268: Please specifically document the lifecycle for the ClientCallTracer, not referencing ClientCallTracerWrapper (which is an implementation detail).

gnossen (approved):

src/core/lib/surface/call.h
```diff
@@ -265,6 +265,10 @@ void grpc_call_log_batch(const char* file, int line, const grpc_op* ops,
...
+void grpc_call_tracer_set_and_manage(grpc_call* call,
```
yashykt, line 269: Let's add a comment saying something to the effect of - Used by Python as a convenient way of using the arena to manage the lifetime of the call tracer.

Also add a comment explaining why we couldn't use `arena->ManagedNew<>` directly in Python

src/core/lib/surface/call.h
```diff
@@ -265,6 +265,16 @@ void grpc_call_log_batch(const char* file, int line, const grpc_op* ops,
 
 void grpc_call_tracer_set(grpc_call* call, grpc_core::ClientCallTracer* tracer);
 
+// Both sets and manages call tracer life cycle.
```
yashykt, line 268: Replace with - Sets call tracer on the call and manages its life by using the call's arena.

src/core/lib/surface/call.h
```diff
@@ -265,6 +265,16 @@ void grpc_call_log_batch(const char* file, int line, const grpc_op* ops,
...
+// When using this API, the tracer will be destroyed by grpc_call arena when
+// grpc_call is about to be destroyed. The caller of this API SHOULD NOT
+// manually destroy the tracer. This API is used by Python as a way of using
+// Arena to manage the lifetime of the call tracer. Python need this API because
```
yashykt, line 272: s/need/needs
s/don\'t/doesn\'t

yashykt (approved):

PR #37467, opened 2024-08-13
[xDS] ORCA to LRS propagation changes

markdroth:
Implements gRFC A85 (https://github.com/grpc/proposal/pull/454).


src/core/xds/xds_client/lrs_client.cc
```diff
@@ -1065,19 +1143,17 @@ void LocalityStatsPopulate(
   envoy_config_endpoint_v3_UpstreamLocalityStats_set_total_issued_requests(
       output, snapshot.total_issued_requests);
   // Add backend metrics.
+  MaybeAddLoadMetric(context, output, "cpu_utilization",
```
yousukseung, line 1146: Are we not using top level fields?

yousukseung (approved):

src/core/xds/xds_client/lrs_client.cc
markdroth: I'm not sure what you're asking here.  Prior to this PR, we don't support propagating the top-level fields from ORCA to LRS, since that wasn't part of A64.  With the changes in A85, though, we are gaining the ability to propagate those fields if we want to.

yousukseung: I meant top-level fields in `load_report.proto`.  I was under the impression that these three metrics will be populated in top-level fields for efficiency?
```
UnnamedEndpointLoadMetricStats cpu_utilization = 12;
UnnamedEndpointLoadMetricStats mem_utilization = 13;
UnnamedEndpointLoadMetricStats application_utilization = 14;
```

markdroth: Oh!  You're absolute right, we should be doing that.  Fixed.

Thanks for catching this!

PR #37468, opened 2024-08-14
[xDS] generalize CDS metadata handling

markdroth:
This is a step toward gRFC A83 (https://github.com/grpc/proposal/pull/438).

### DIFF
--- a/src/core/xds/grpc/xds_cluster_parser.cc
+++ b/src/core/xds/grpc/xds_cluster_parser.cc
@@ -641,34 +641,49 @@ absl::StatusOr<std::shared_ptr<const XdsClusterResource>> CdsResourceParse(
     cds_update->override_host_statuses.Add(
         XdsHealthStatus(XdsHealthStatus::kHealthy));
   }
-  // Record telemetry labels (if any).
+  // Parse metadata.
   const envoy_config_core_v3_Metadata* metadata =
       envoy_config_cluster_v3_Cluster_metadata(cluster);
   if (metadata != nullptr) {
-    google_protobuf_Struct* telemetry_labels_struct;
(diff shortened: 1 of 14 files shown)


src/core/xds/grpc/xds_cluster_parser.cc
```diff
@@ -641,34 +641,49 @@ absl::StatusOr<std::shared_ptr<const XdsClusterResource>> CdsResourceParse(
...
+      absl::string_view key = UpbStringToAbsl(
+          envoy_config_core_v3_Metadata_FilterMetadataEntry_key(entry));
+      auto json = ParseProtobufStructToJson(
+          context,
+          envoy_config_core_v3_Metadata_FilterMetadataEntry_value(entry));
+      if (!json.ok()) {
+        ValidationErrors::ScopedField field(
+            &errors, absl::StrCat(".metadata.filter_metadata[", key, "]"));
+        errors.AddError(json.status().message());
+      } else if (!cds_update->metadata.contains(key)) {
```
yashykt, line 685: is ignoring duplicates the expected behavior? I don't see it mentioned in the gRFC

yashykt (approved):
approved with comment

src/core/xds/grpc/xds_cluster_parser.cc
markdroth: The gRFC says:

"""
The value for a given metadata key will come from only one of the two maps; the value from `filter_metadata` will be used only if the key is not present or is of an unknown protobuf message type in `typed_filter_metadata`.
"""

We check `typed_filter_metadata` first, and we use the entry in `filter_metadata` only if we didn't already get one from `typed_filter_metadata`.

PR #37485, opened 2024-08-15
[OTel C++] Fix race when adding and removing callbacks

yashykt:
Split off from https://github.com/grpc/grpc/pull/37425

We are adding and removing callbacks on the OpenTelemetry Async Instruments without synchronization. This opens us to races where we have an AddCallback and RemoveCallback operation happening at the same time. The correct result after these operations is to still have a callback registered with OpenTelemetry at the end, but the two operations could race and we could just decide to remove the OpenTelemetry callback.

The fix delays removing OpenTelemetry callbacks to plugin destruction time.

### DIFF
--- a/src/cpp/ext/otel/otel_plugin.cc
+++ b/src/cpp/ext/otel/otel_plugin.cc
@@ -575,6 +575,7 @@ OpenTelemetryPluginImpl::~OpenTelemetryPluginImpl() {
         [](const std::unique_ptr<opentelemetry::metrics::Histogram<double>>&) {
         },
         [](const std::unique_ptr<CallbackGaugeState<int64_t>>& state) {
+          CHECK(state->caches.empty());
           if (state->ot_callback_registered) {
             state->instrument->RemoveCallback(
                 &CallbackGaugeState<int64_t>::CallbackGaugeCallback,
@@ -583,6 +584,7 @@ OpenTelemetryPluginImpl::~OpenTelemetryPluginImpl() {
           }
         },
         [](const std::unique_ptr<CallbackGaugeState<double>>& state) {
+          CHECK(state->caches.empty());
           if (state->ot_callback_registered) {
             state->instrument->RemoveCallback(
                 &CallbackGaugeState<double>::CallbackGaugeCallback,
--- a/src/cpp/ext/otel/otel_plugin.cc
+++ b/src/cpp/ext/otel/otel_plugin.cc
@@ -563,6 +563,36 @@ OpenTelemetryPluginImpl::OpenTelemetryPluginImpl(
       });
 }
 
+OpenTelemetryPluginImpl::~OpenTelemetryPluginImpl() {
+  for (const auto& instrument_data : instruments_data_) {
+    grpc_core::Match(
+        instrument_data.instrument, [](const Disabled&) {},
+        [](const std::unique_ptr<opentelemetry::metrics::Counter<double>>&) {},
+        [](const std::unique_ptr<opentelemetry::metrics::Counter<uint64_t>>&) {
+        },
+        [](const std::unique_ptr<
+            opentelemetry::metrics::Histogram<uint64_t>>&) {},
+        [](const std::unique_ptr<opentelemetry::metrics::Histogram<double>>&) {
+        },
+        [](const std::unique_ptr<CallbackGaugeState<int64_t>>& state) {
+          if (state->ot_callback_registered) {
+            state->instrument->RemoveCallback(
+                &CallbackGaugeState<int64_t>::CallbackGaugeCallback,
+                state.get());
+            state->ot_callback_registered = false;
+          }
+        },
+        [](const std::unique_ptr<CallbackGaugeState<double>>& state) {
+          if (state->ot_callback_registered) {
+            state->instrument->RemoveCallback(
+                &CallbackGaugeState<double>::CallbackGaugeCallback,
+                state.get());
+            state->ot_callback_registered = false;
+          }
+        });
+  }
+}
+
 namespace {
 constexpr absl::string_view kLocality = "grpc.lb.locality";
 }
@@ -823,9 +853,6 @@ void OpenTelemetryPluginImpl::AddCallback(
 
 void OpenTelemetryPluginImpl::RemoveCallback(
     grpc_core::RegisteredMetricCallback* callback) {
-  std::vector<
-      absl::variant<CallbackGaugeState<int64_t>*, CallbackGaugeState<double>*>>
-      gauges_that_need_to_remove_callback;
   {
     grpc_core::MutexLock lock(&mu_);
     callback_timestamps_.erase(callback);
@@ -848,11 +875,6 @@ void OpenTelemetryPluginImpl::RemoveCallback(
           CHECK_NE(callback_gauge_state, nullptr);
           CHECK((*callback_gauge_state)->ot_callback_registered);
           CHECK_EQ((*callback_gauge_state)->caches.erase(callback), 1u);
-          if ((*callback_gauge_state)->caches.empty()) {
-            gauges_that_need_to_remove_callback.push_back(
-                callback_gauge_state->get());
-            (*callback_gauge_state)->ot_callback_registered = false;
-          }
           break;
         }
         case grpc_core::GlobalInstrumentsRegistry::ValueType::kDouble: {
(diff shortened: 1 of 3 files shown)


test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1996,6 +1990,91 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
...
+  MetricsCollectorThread new_collector{
+      this,
+      grpc_core::Duration::Milliseconds(100) * grpc_test_slowdown_factor(),
+      kIterations,
+      [&](const absl::flat_hash_map<
+          std::string,
+          std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
+              data) { return false; }};
+  // We shouldn't get any new callbacks
+  EXPECT_EQ(new_collector.Stop().size(), 0);
```
yijiem, line 2073: `EXPECT_TRUE(new_collector.Stop().empty());`?

yijiem (commented):
Thanks Yash!

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -563,6 +563,36 @@ OpenTelemetryPluginImpl::OpenTelemetryPluginImpl(
       });
 }
 
+OpenTelemetryPluginImpl::~OpenTelemetryPluginImpl() {
+  for (const auto& instrument_data : instruments_data_) {
```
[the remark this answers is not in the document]
yashykt: Destructors usually do not require locks to be held. We usually refcount our implementation which means that the destructor will only be invoked when all users of this plugin are done with it.

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1996,6 +1990,91 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
...
+      }};
+  absl::flat_hash_map<
+      std::string,
+      std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>
+      data = collector.Stop();
+  // Verify that callbacks are invoked
+  EXPECT_EQ(report_count_1, kIterations);
+  EXPECT_EQ(report_count_2, kIterations);
+  // Reset the global stats plugins
+  registered_metric_callback_1.reset();
```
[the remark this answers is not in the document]
yashykt: I was mostly interested in what happens after all the callbacks are removed and stats plugins are cleared as well, so that we are making sure that we are clearing callbacks, but you make a good point.
Changed the test to first remove one callback, then the other and finally remove stats plugin as well.

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: done

src/cpp/ext/otel/otel_plugin.cc
```diff
@@ -563,6 +563,36 @@ OpenTelemetryPluginImpl::OpenTelemetryPluginImpl(
...
+        [](const std::unique_ptr<opentelemetry::metrics::Counter<double>>&) {},
+        [](const std::unique_ptr<opentelemetry::metrics::Counter<uint64_t>>&) {
+        },
+        [](const std::unique_ptr<
+            opentelemetry::metrics::Histogram<uint64_t>>&) {},
+        [](const std::unique_ptr<opentelemetry::metrics::Histogram<double>>&) {
+        },
+        [](const std::unique_ptr<CallbackGaugeState<int64_t>>& state) {
+          if (state->ot_callback_registered) {
+            state->instrument->RemoveCallback(
```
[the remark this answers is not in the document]
yashykt: done

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1715,14 +1715,8 @@ TEST_F(OpenTelemetryPluginNPCMetricsTest, InstrumentsEnabledTest) {
...
 
-class OpenTelemetryPluginCallbackMetricsTest
-    : public OpenTelemetryPluginEnd2EndTest {
- protected:
-  OpenTelemetryPluginCallbackMetricsTest()
-      : endpoint_config_(grpc_core::ChannelArgs()) {}
-
-  grpc_event_engine::experimental::ChannelArgsEndpointConfig endpoint_config_;
-};
+typedef OpenTelemetryPluginNPCMetricsTest
```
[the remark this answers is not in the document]
yashykt: done

src/cpp/ext/otel/otel_plugin.cc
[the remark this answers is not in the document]
yijiem: Thinking about this, if each component eventually owns a shared ownership to a StatsPlugin and when the StatsPlugin is destroyed that guarantees that all components has been destroyed, so does their `RegisteredMetricCallback`. So we might be fine here.

yijiem (approved):
Thanks Yash!

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1996,6 +1990,117 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
...
+          std::string,
+          std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
+              data) {
+        return !data.contains(kInt64CallbackGaugeMetric) ||
+               !data.contains(kDoubleCallbackGaugeMetric);
+      }};
+  absl::flat_hash_map<
+      std::string,
+      std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>
+      data = collector.Stop();
```
yijiem, line 2055: You may simplify this part with just this if you are not using the data that `Stop` returns:
```
{
  MetricsCollectorThread collector{
        this,
        grpc_core::Duration::Milliseconds(100) * grpc_test_slowdown_factor(),
        kIterations,
        [&](const absl::flat_hash_map<
            std::string,
            std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
                data) {
          return !data.contains(kInt64CallbackGaugeMetric) ||
                 !data.contains(kDoubleCallbackGaugeMetric);
        }};
}
```

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1996,6 +1990,117 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
...
+  MetricsCollectorThread new_new_collector{
+      this,
+      grpc_core::Duration::Milliseconds(100) * grpc_test_slowdown_factor(),
+      kIterations,
+      [&](const absl::flat_hash_map<
+          std::string,
+          std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
+              data) { return false; }};
+  // We shouldn't get any new callbacks
+  EXPECT_THAT(new_new_collector.Stop(), ::testing::IsEmpty());
```
yijiem, line 2083: Here you may do:
```
{
  MetricsCollectorThread collector{
        this,
        grpc_core::Duration::Milliseconds(100) * grpc_test_slowdown_factor(),
        kIterations,
        [&](const absl::flat_hash_map<
            std::string,
            std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
                data) {
          return false;
        }};
  EXPECT_THAT(collector.Stop(), ::testing::IsEmpty());
}
```

same below

test/cpp/ext/otel/otel_plugin_test.cc
```diff
@@ -1996,6 +1990,117 @@ TEST_F(OpenTelemetryPluginCallbackMetricsTest,
...
+  registered_metric_callback_1.reset();
+  MetricsCollectorThread new_collector{
+      this,
+      grpc_core::Duration::Milliseconds(100) * grpc_test_slowdown_factor(),
+      kIterations,
+      [&](const absl::flat_hash_map<
+          std::string,
+          std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
+              data) { return false; }};
+  new_collector.Stop();
```
yijiem, line 2069: Maybe just:
```
{
  MetricsCollectorThread collector{
        this,
        grpc_core::Duration::Milliseconds(100) * grpc_test_slowdown_factor(),
        kIterations,
        [&](const absl::flat_hash_map<
            std::string,
            std::vector<opentelemetry::sdk::metrics::PointDataAttributes>>&
                data) {
          return false;
        }};
}
```

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: done

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: done

test/cpp/ext/otel/otel_plugin_test.cc
yashykt: that's not helping though. All it is doing is allowing us to use the same variable again. For tests, we should be aiming for readability rather than efficiency.

PR #37486, opened 2024-08-15
[OTel C++] Simplify tests

yashykt:
It is hard to reason about tests if multiple callbacks record values for the same metrics with the same label sets.


yijiem (approved):
Thanks Yash!

PR #37513, opened 2024-08-16
[Gpr_To_Absl_Logging] Remove gpr logging header include from other headers

tanvi-jagtap:
[Gpr_To_Absl_Logging] Remove gpr logging header include from other headers

Some of the cc files are using gpr_log_verbosity_init() functions.

So I removed the #include <grpc/support/log.h> from all headers and put it selectively only in the cc files that used gpr_log_verbosity_init()

### DIFF
--- a/src/core/lib/surface/init.cc
+++ b/src/core/lib/surface/init.cc
@@ -24,6 +24,7 @@
 #include <grpc/fork.h>
 #include <grpc/grpc.h>
 #include <grpc/impl/channel_arg_names.h>
+#include <grpc/support/log.h>
 #include <grpc/support/port_platform.h>
 #include <grpc/support/sync.h>
 #include <grpc/support/time.h>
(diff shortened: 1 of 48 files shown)


src/core/lib/surface/init.cc
```diff
@@ -24,6 +24,7 @@
 #include <grpc/fork.h>
 #include <grpc/grpc.h>
 #include <grpc/impl/channel_arg_names.h>
+#include <grpc/support/log.h>
 #include <grpc/support/port_platform.h>
 #include <grpc/support/sync.h>
```
tanvi-jagtap, line 29: Some of the cc files are using gpr_log_verbosity_init() functions.

So I removed the #include <grpc/support/log.h> from all headers and put it selectively only in the cc files that used gpr_log_verbosity_init()

nanahpang (approved):

PR #37544, opened 2024-08-21
[call creds] implement GcpServiceAccountIdentityCredentials

markdroth:
As per gRFC A83 (https://github.com/grpc/proposal/pull/438).

For now, I am not exposing this new call creds type via the C-core API or in any C++ or wrapped language public APIs, so there's no way to use it externally.  We can easily add that in the future if someone asks, but for now the intent is to use it only internally via the xDS GCP authentication filter, which I'll implement in a subsequent PR.

As part of this, I changed the test framework in credentials_test to check the status code in addition to the message on failure.  This exposed several places where existing credential types are returnign the wrong status code (unsurprisingly, because of all of the tech debt surrounding grpc_error).  I have not fixed this behavior, but I have added TODOs in the test showing which ones I think need to be fixed.


src/core/lib/security/credentials/gcp_service_account_identity/gcp_service_account_identity_credentials.h
```diff
@@ -0,0 +1,87 @@
...
+
+ private:
+  class HttpFetchRequest;
+
+  virtual OrphanablePtr<HttpRequest> StartHttpRequest(
+      grpc_polling_entity* pollent, Timestamp deadline,
+      grpc_http_response* response, grpc_closure* on_complete) = 0;
+};
+
+// GCP service account identity call credentials.
```
matthewstevenson88, line 59: optional: Consider adding a link or some kind of reference to the gRFC.

matthewstevenson88 (approved):

src/core/lib/security/credentials/gcp_service_account_identity/gcp_service_account_identity_credentials.h
markdroth: Will do in a subsequent PR.

PR #37547, opened 2024-08-21
[CI] Upgraded all `Debian 10` based images to `Debian 11`

veblush:
Debian 10 is long gone and it has too old cmake to get new Abseil.


drfloob (approved):

drfloob:
We could arguably drop Debian 11 too. Based on the [Foundational C++ support policy](https://opensource.google/documentation/policies/cplusplus-support), Debian 11 is in an "extended support" lifetime ([staffed only by volunteers](https://endoflife.date/debian) IIUC), and quoting the doc:

> We only support our libraries with distributions still under "Active" or "Full" support.

veblush:
I used this [table](https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md#user-content-fn-next-change-3bd690e98ed33f4cf1e899ec7897d04a) to determine its oldest version. This table says that Debian 11 will be supported by 2026-06-30. Debian typically provides 5 years of standard support (including security updates) and an additional 5 years of extended support. We aim to align with the 5-year standard support window for Debian but won't cover the extended support period.

drfloob:
> Debian typically provides 5 years of standard support (including security updates) and an additional 5 years of extended support.

I believe our table is wrong. https://www.debian.org/security/faq#lifespan. Maybe this changed recently, but the Debian security team only supports a release until 3 years after its initial release. Which ended 1 week ago for Debian 11.

I don't think Debian 11 is the long pole in any of our future improvement hopes (C++17, etc), so I have no _strong_ argument here.

veblush:
This [page](https://www.debian.org/releases/bullseye/) says

> The Debian 11 life cycle encompasses five years: the initial three years of full Debian support, until August 14th, 2024, and two years of Long Term Support (LTS), until August 31st, 2026. The set of supported architectures is limited during the Bullseye LTS term to i386, amd64, armhf and arm64. For more information, please refer to the [Security Information](https://www.debian.org/security/) webpage and the [LTS section of the Debian Wiki](https://wiki.debian.org/LTS).

5 years = 3 years of active support + 2 years of long term support.

PR #37552, opened 2024-08-22
[SourceTreeLayout] Move promise folder

tanvi-jagtap:
[SourceTreeLayout] Move promise folder from
`src/core/lib/promise` to 
`src/core/util/promise`


src/core/util/promise/activity.cc
```diff
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "src/core/lib/promise/activity.h"
+#include <grpc/support/port_platform.h>
```
markdroth, line 15: We probably want to add a .clang-format file in this directory to avoid moving port_platform.h to the top of the file, since we've been trying to make that rule go away, and this would be a regression.

Or even better, just try removing this include from these files.  My guess is that it's not actually needed here.

markdroth (approved):

PR #37534, opened 2024-08-20
Fix some label references for Bzlmod support

Wyverald:
These changes are compatible with WORKSPACE.

PR #37561, opened 2024-08-23
[Flakiness] Fix flakiness in retry_transparent_max_concurrent_streams_test

yashykt:
Refer b/333896115 for more details

### DIFF
--- a/test/core/end2end/tests/retry_transparent_max_concurrent_streams.cc
+++ b/test/core/end2end/tests/retry_transparent_max_concurrent_streams.cc
@@ -40,14 +40,10 @@ namespace {
 // restarted.  The second call will fail in that transport instance and
 // will be transparently retried after the server starts up again.
 CORE_END2END_TEST(RetryHttp2Test, RetryTransparentMaxConcurrentStreams) {
-  // The minimum time between subsequent connection attempts helps with cases
-  // where the connection handshake does not realize that the socket is not
-  // working in time. (b/333896115)
   const auto server_args =
       ChannelArgs()
           .Set(GRPC_ARG_MAX_CONCURRENT_STREAMS, 1)
-          .Set(GRPC_ARG_MAX_CONCURRENT_STREAMS_OVERLOAD_PROTECTION, false)
-          .Set(GRPC_ARG_MIN_RECONNECT_BACKOFF_MS, 5000);
+          .Set(GRPC_ARG_MAX_CONCURRENT_STREAMS_OVERLOAD_PROTECTION, false);
   InitServer(server_args);
   InitClient(ChannelArgs());
   auto c =
--- a/test/core/end2end/tests/retry_transparent_max_concurrent_streams.cc
+++ b/test/core/end2end/tests/retry_transparent_max_concurrent_streams.cc
@@ -116,7 +116,7 @@ CORE_END2END_TEST(RetryHttp2Test, RetryTransparentMaxConcurrentStreams) {
   Expect(201, true);
   // Give enough time for the handshake to timeout, and a new handshake to start
   // if needed. (b/333896115)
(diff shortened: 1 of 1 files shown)


markdroth:
If I'm understanding the problem correctly based on the description in the internal bug, the issue is that when the client attempts to reconnect, it thinks that the TCP connection attempt succeeded, even though it actually didn't, so it sits there waiting for the H2 SETTINGS frame timeout, which takes longer than the test wants.  Is that right?

If so, then this PR seems like at best a hacky work-around.  I would prefer to instead try to root-cause the problem: Why does the client think that the TCP connection attempt succeeded in the first place?  If we fix that, then we won't need to add sleeps in this test.

yashykt:
Shutdown on the server side happens asynchronously.
If you look at the logs at https://fusion2.corp.google.com/ci/kokoro/prod:grpc%2Fcore%2Fmaster%2Flinux%2Farm64%2Fgrpc_bazel_test_c_cpp/activity/b25bf0bf-7387-4c45-9f9a-eea747ca2f21/log, we see that `tcp_connect` happens before we shutdown the listening socket -
```
I0823 17:56:59.495409   23145 handshaker.cc:191] handshake_manager 0x3fe9c00: calling handshaker tcp_connect [0x3fe9da0] at index 0
```
```
E0823 17:56:59.495694   23493 posix_engine_listener.h:113] ~AsyncConnectionAcceptor 0x3fd1560
E0823 17:56:59.495705   23493 ev_poll_posix.cc:354] Orphan Handle 0x3fd1700 
E0823 17:56:59.495710   23493 ev_poll_posix.cc:379] calling shutdown 0x3fd1700
E0823 17:56:59.502651   23505 ev_poll_posix.cc:121] closing fd 0x3fd1700
```

I wouldn't call this a bug. The ideal behavior on the client side would have been to get a poll notification that the server socket has shutdown without accepting the connection, but we don't see that. To dig into why we don't see that event would require digging into the arm64 debian11 kernel which I'm not sure is worth it.

yashykt:
b/333896115 unfortunately doesn't reproduce when polling logs are enabled, but we see this with both epoll1 and poll poling engines.

test/core/end2end/tests/retry_transparent_max_concurrent_streams.cc
```diff
@@ -40,10 +40,14 @@ namespace {
...
 CORE_END2END_TEST(RetryHttp2Test, RetryTransparentMaxConcurrentStreams) {
+  // The minimum time between subsequent connection attempts helps with cases
+  // where the connection handshake does not realize that the socket is not
+  // working in time. (b/333896115)
   const auto server_args =
       ChannelArgs()
           .Set(GRPC_ARG_MAX_CONCURRENT_STREAMS, 1)
-          .Set(GRPC_ARG_MAX_CONCURRENT_STREAMS_OVERLOAD_PROTECTION, false);
+          .Set(GRPC_ARG_MAX_CONCURRENT_STREAMS_OVERLOAD_PROTECTION, false)
+          .Set(GRPC_ARG_MIN_RECONNECT_BACKOFF_MS, 5000);
```
markdroth, line 50: I think this needs to be multiplied by `grpc_test_slowdown_factor()`.

markdroth, line 50: Will we still need this after b/362326480 is fixed?  If not, please add a TODO.

yashykt: that would mess up the timing instead, but we don't need this anyway, removing it.

markdroth (approved):

malkia:
Why was this (seemingly) unrelated fix was undoed? - https://github.com/grpc/grpc/pull/37534 ? cc: @Wyverald

PR #37563, opened 2024-08-23
[php] Fix PHPDoc so that UnaryCall defines the proper return type

### LINKED ISSUE #33431: Typing support for grpc-php
I've recently started using the php-codegen for gRPC and it is really annoying that although the generator knows the return value of an rpc call, non-streaming rpc calls (with ->wait()) will just return an instance of `UnaryCall`. This might be fine for some, but PHP has been introducing more type safety and tools like PHPStan have become a must-have for many. It is quite annoying that there is no type-safety on the response of a grpc call at all. 

I'd like the `UnaryCall` class to extend PHPStans generic template system like the following:
```php
/**
 * @template T of object
 * Represents an active call that sends a single message and then gets a
 * single response.
 */
class UnaryCall extends AbstractCall
```

with 

```php
    /**
     * Wait for the server to respond with data and a status.
     *
     * @return array{0: T, 1: any} [response data, status]
     */
    public function wait()
    {
        $batch = [
            OP_RECV_MESSAGE => true,
            OP_RECV_STATUS_ON_CLIENT => true,
        ];
        if ($this->metadata === null) {
            $batch[OP_RECV_INITIAL_METADATA] = true;
        }
        $event = $this->call->startBatch($batch);
        if ($this->metadata === null) {
            $this->metadata = $event->metadata;
        }
        $status = $event->status;
        $this->trailing_metadata = $status->metadata;

        return [$this->_deserializeResponse($event->message), $status];
    }
```

Instead of `any` for the status, one could also use a typed version of the Status data.
Now, what the generator could do is instead of just inserting 
```php
/**
 * @returns UnaryCall
 **/
 ```
 
 it could instead return 
 ```php
/**
 * @returns UnaryCall<my/simple/return/type>
 **/
 ```
This should allow for IDEs like PHPStorm to detect the type of the response data, which would make development so much more intuitive.

This is just an example of where it could be implemented. This could be implemented in streaming-rpcs, server code and other places as well.

I have not come up with any good alternative to this suggestion yet.

### PR DESCRIPTION
**Pull request by sayuprc**
Type information is added to the result of `UnaryCall::wait()` execution.

resolve #33431

### CODE REVIEW

File: src/compiler/php_generator.cc [hunk 1]
```diff
@@ -106,17 +106,22 @@ void PrintMethod(const MethodDescriptor* method, Printer* out) {
...
+                "public function $name$(\\$input_type_id$ $$argument,\n"
+                "  $$metadata = [], $$options = []) {\n");
     }
-    out->Print(vars,
-               " * @param \\$input_type_id$ $$argument input argument\n"
-               " * @param array $$metadata metadata\n"
-               " * @param array $$options call options\n"
-               " * @return $return_type_id$\n */\n"
-               "public function $name$(\\$input_type_id$ $$argument,\n"
-               "  $$metadata = [], $$options = []) {\n");
```
bshaffer, line 119: I don't like all the code duplication this presents - why not do something like this instead? Then you'd only have to change a single line of code

```c
    if (method->server_streaming()) {
      vars["return_type_id"] = "\\Grpc\\ServerStreamingCall";
    } else {
      vars["return_type_id"] = sprintf("\\Grpc\\UnaryCall<%s>", $vars["input_type_id"]);
    }
    out->Print(vars,
               " * @param \\$input_type_id$ $$argument input argument\n"
               " * @param array $$metadata metadata\n"
               " * @param array $$options call options\n"
               " * @return $return_type_id$\n */\n"
               "public function $name$(\\$input_type_id$ $$argument,\n"
               "  $$metadata = [], $$options = []) {\n");

```

bshaffer (commented):
This is great, but we also need some tests for the generated output

sayuprc:
@bshaffer 

Thank you for your review.
I will make the corrections you pointed out and add the necessary tests.

I have one question, though—I wasn't sure where to place the test code.
For the changes I made this time, where would be the correct location to place the test code?

bshaffer:
@sayuprc you just need to follow the instructions in `src/php/README.md` to run the tests, and then run `src/php/bin/generate_proto_php.sh` to update the Grpc Client classes. You should get something like this:

https://github.com/bshaffer/grpc/commit/d3cf6db72ed35552491538a56fde1927e52b18bb

sayuprc:
@bshaffer 
Thank you for the thorough explanation.
I will give it a try.

bshaffer:
This is looking great! I'm not sure what's up with that "NO CHECK IN GEN CODE", though. We will need to look into that. 

@stanley-cheung what do you think?

sayuprc:
I apologize for the delayed response.

Regarding `# NO CHECKED-IN PROTOBUF GENCODE` appearing in the diff, I thought it might be related to this commit: https://github.com/protocolbuffers/protobuf/commit/c668e2e4da3eccf268f4ff74be087899fe87df8b
However, I wasn’t able to understand much beyond that.

What should I do?

Nyholm:
This is simply amazing. Is there anything I can do to help getting this merged?

bshaffer:
Messaging here for visibility is very helpful @Nyholm ! It's just waiting on me or @stanley-cheung to approve and merge.

stanley-cheung:
Sorry, there's some clang format diff here, do you mind fixing that? Thanks

```
--- /var/local/git/grpc/src/compiler/php_generator.cc	2025-06-24 18:31:09.427418725 +0000
+++ /tmp/tmp.UMLUXWZXww	2025-06-24 18:38:18.370327357 +0000
@@ -108,7 +108,8 @@
     if (method->server_streaming()) {
       vars["return_type_id"] = "\\Grpc\\ServerStreamingCall";
     } else {
-      vars["return_type_id"] = "\\Grpc\\UnaryCall<\\" + vars["output_type_id"] + ">";
+      vars["return_type_id"] =
+          "\\Grpc\\UnaryCall<\\" + vars["output_type_id"] + ">";
     }
     out->Print(vars,
                " * @param \\$input_type_id$ $$argument input argument\n"
```
You may be able to just run `./tools/distrib/clang_format_code.sh`

sayuprc:
@stanley-cheung 
Thanks for the review!
I fixed the format.

stanley-cheung (approved):

apolcyn (approved):

stanley-cheung:
This has been merged. Thanks for the contributions. This will be part of the next 1.74.0 release which will happen in the next couple of weeks.

src/php/lib/Grpc/UnaryCall.php
```diff
@@ -50,7 +52,7 @@ public function start($data, array $metadata = [], array $options = [])
     /**
      * Wait for the server to respond with data and a status.
      *
-     * @return array [response data, status]
+     * @return array{0: T|null, 1: stdClass} [response data, status]
```
shu-yusa, line 55: @sayuprc shouldn't `stdClass` be imported or used as `\stdClass` under a namespace?

sayuprc: Thank you for your feedback. You are correct—I'll make the necessary changes and open a PR.

PR #37565, opened 2024-08-23
[tls] Add ValidateCredentials API to the TLS certificate provider.

matthewstevenson88:
Add a ValidateCredentials API to the TLS certificate provider interface. A user can call this API to check that the credentials currently held by the certificate provider instance are valid. The definition of "valid" depends on provider that is being used. For the static data and file watcher providers, "valid" means that the credentials consist of valid PEM.

~Currently there is no check to ensure that credentials consist of valid PEM blocks before a TLS handshake commences. This PR creates a static factory for FileWatcherCertificateProvider (and marks the constructor as deprecated) which performs this validation check. The analogous work for StaticDataCertificateProvider will be done in a follow-up PR.~


src/cpp/common/tls_certificate_provider.cc
```diff
@@ -23,6 +23,8 @@
 #include <grpc/grpc_security.h>
 #include <grpcpp/security/tls_certificate_provider.h>
 
+#include "src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h"
```
gtcooke94, line 26: I think here we'd prefer to alias the new APIs in `grpcpp/security/tls_certificate_provider.h` if possible, rather that bring in the `src/core` import. And I believe we want this API exposed in the`grpcpp` include header so it can be used.
See the CRL Provider APIs as an exampe - https://github.com/grpc/grpc/blob/a615c505d1adea224a4b303f92f48fb60311a6c3/include/grpcpp/security/tls_crl_provider.h#L29-L34

src/core/tsi/test_creds/BUILD
```diff
@@ -32,4 +32,6 @@ exports_files([
     "multi-domain.pem",
     "leaf_signed_by_intermediate.key",
     "leaf_and_intermediate_chain.pem",
+    "malformed-cert.pem",
```
gtcooke94, line 35: How did you generate these files? I try to have a README describing how test data is created and ideally a script that can be run to generate it, can you add one or both of those?

e.g. these two :
1. https://github.com/grpc/grpc/blob/fc8ab2b7b454cad2bad42ef18a79d2893524cfc0/test/core/tsi/test_creds/crl_data/README#L1

2. https://github.com/grpc/grpc/blob/fc8ab2b7b454cad2bad42ef18a79d2893524cfc0/test/core/tsi/test_creds/crl_data/intermediate_gen.sh#L1

src/cpp/common/tls_certificate_provider.cc
matthewstevenson88: This works in the CRL provider case because there is no C++ layer implementation - it's shared with the C-core implementation. In this case, there is already a `FileWatcherCertificateProvider` defined in `tls_crl_provider.{h,cc}`, so I'm not sure we can do this without converging the two implementations.

src/core/tsi/test_creds/BUILD
matthewstevenson88: Sure! Added to the README. But note that there is no script to generate this testdata: I just copied one of the existing PEM blocks and deleted a random character. :)

src/cpp/common/tls_certificate_provider.cc
gtcooke94: I think you can add a similar looking `::Create` to the definition in `tls_certificate_provider` that ends up wrapping the C-Core implementation or however you want to do it?

gtcooke94: Discussed offline

matthewstevenson88:
@gtcooke94 Any further concerns?

gtcooke94 (approved):
Thanks for the PR, LGTM!

include/grpcpp/security/tls_certificate_provider.h
```diff
@@ -36,6 +40,11 @@ class GRPCXX_DLL CertificateProviderInterface {
  public:
   virtual ~CertificateProviderInterface() = default;
   virtual grpc_tls_certificate_provider* c_provider() = 0;
+
+  // Returns an OK status if the credentials held by the provider are valid.
+  // What it means for a credential to be valid is determined by the provider
+  // implementation.
+  virtual absl::Status ValidateCredentials() const { return absl::OkStatus(); }
```
markdroth, line 47: I don't think it makes sense for this to be a method on the `CertificateProviderInterface`, because the interface is actually designed for more general-purpose cases where there can be multiple certs provided by the same provider, and this API is ambiguous in cases where some of the certs are valid and others are not.

I think this method makes sense only for provider implementations that supply only a single cert, such as the file-watcher and static providers.  Given that, I think we should support this method only on the file-watcher and static providers specifically.  Applications will know which specific subclass they instantiated, so it should be fine for them to call this method.

BUILD
```diff
@@ -2493,6 +2495,7 @@ grpc_cc_library(
         "//src/core:grpc_backend_metric_provider",
         "//src/core:grpc_crl_provider",
         "//src/core:grpc_service_config",
+        "//src/core:grpc_tls_credentials",
```
markdroth, line 2498: I don't think we want this dependency, because this will pull in the SSL libraries as a dependency, which we don't want in insecure builds.

Why is this needed?

include/grpcpp/security/tls_certificate_provider.h
```diff
@@ -36,6 +40,11 @@ class GRPCXX_DLL CertificateProviderInterface {
  public:
   virtual ~CertificateProviderInterface() = default;
   virtual grpc_tls_certificate_provider* c_provider() = 0;
+
+  // Returns an OK status if the credentials held by the provider are valid.
+  // What it means for a credential to be valid is determined by the provider
```
markdroth, line 45: This statement seems so weak that I'm not sure the API actually has value.  I think we probably need to define this a bit more rigorously in order for it to be useful.

src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h
```diff
@@ -77,6 +77,11 @@ struct grpc_tls_certificate_provider
   // instances of that provider implementation.
   virtual grpc_core::UniqueTypeName type() const = 0;
 
+  // Returns an OK status if the credentials held by the provider are valid.
+  // What it means for a credential to be valid is determined by the provider
+  // implementation.
+  virtual absl::Status ValidateCredentials() const { return absl::OkStatus(); }
```
markdroth, line 83: As mentioned elsewhere, I don't think this method belongs on the base class, only on the file-watcher and static implementations.

test/cpp/server/credentials_test.cc
```diff
@@ -116,6 +120,102 @@ TEST(
   CHECK_NE(server_credentials.get(), nullptr);
 }
 
+TEST(CredentialsTest,
```
markdroth, line 123: I don't think we need to test every success and error case both here and at the C-core layer.  The C-core tests already cover all of the cases; I think the code here should simply check a single success case and a single error case for each of the two provider impls, just to prove that we're propagating the result from the C-core layer correctly.

markdroth (commented):
This looks like the right approach!  Just a few comments to address.

Please let me know if you have any quesitons.  Thanks!

src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h
matthewstevenson88: Removed from the base class.

test/cpp/server/credentials_test.cc
matthewstevenson88: Removed redundant tests.

include/grpcpp/security/tls_certificate_provider.h
matthewstevenson88: Removed from the base class.

include/grpcpp/security/tls_certificate_provider.h
matthewstevenson88: That sounds reasonable, thanks. Removed from the base class.

BUILD
matthewstevenson88: Good callout. To confirm, you're mostly worried including about the `libcrypto` and `libssl` external deps in the insecure build - is that correct? Or are you worried about pulling in SSL TSI as well?

If yes, I think both `libcrypto` and `libssl` are already pulled in to `grpc++_base` because they are external deps of `//src/core:grpc_crl_provider`, which is included here. However, I agree that this new dependency is pulling in SSL TSI also

The reason this is needed is because of including `src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h` in `src/cpp/common/tls_certificate_provider.cc`. For a similar reason, we have the `//src/core:grpc_crl_provider` dependency.

One option to avoid this dependency is to add C wrapper APIs to `include/grpc/credentials.h` and use those from `src/cpp/common/tls_certificate_provider.cc`, rather than directly including `src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h`. My understanding is that this is the older model that we were moving away from, but I may have misunderstood.

WDYT?

markdroth: Sorry, nevermind this comment -- I was confused about the BUILD target hierarchy here.  I thought I remembered that `grpc++_base` was a common target used for both secure and insecure builds, but it looks like it's actually the secure-specific target; users that want insecure builds need to use `grpc++_base_unsecure` instead.

So I think this is fine.

markdroth (approved):

src/cpp/common/tls_certificate_provider.cc
```diff
@@ -47,7 +47,9 @@ StaticDataCertificateProvider::~StaticDataCertificateProvider() {
 };
 
 absl::Status StaticDataCertificateProvider::ValidateCredentials() const {
-  return c_provider_->ValidateCredentials();
+  auto* provider =
+      reinterpret_cast<grpc_core::StaticDataCertificateProvider*>(c_provider_);
```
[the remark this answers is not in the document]
matthewstevenson88: Done.

markdroth (approved):

PR #37566, opened 2024-08-23
[xDS] validate Audience cluster metadata

markdroth:
Add validation of the `Audience` cluster metadata type, as per gRFC A83 (https://github.com/grpc/proposal/pull/438).

I had previously changed the metadata to be represented as JSON in #37468.  However, while working on the GCP Authentication filter implementation, I realized that that's not an ideal representation, because it would have required us to validate the JSON on a per-RPC basis, which would be bad for performance.  So I've changed the representation of metadata to be an abstract type, and we now store the `Audience` metadata as a simple string.  I've also moved metadata into its own type with its own validation code, so that in the future we can use it in places other than CDS (many xDS resource types have metadata fields).

While I was at it, I also add some helper functions for validating the `UInt32Value` and `UInt64Value` wrapper protos.


test/core/xds/xds_metadata_test.cc
```diff
@@ -0,0 +1,276 @@
...
+namespace testing {
+namespace {
+
+class XdsMetadataTest : public ::testing::Test {
+ protected:
+  XdsMetadataTest()
+      : xds_client_(MakeXdsClient()),
+        decode_context_{xds_client_.get(),
+                        *xds_client_->bootstrap().servers().front(),
+                        // FIXME: why does each test need its own tracer?
```
yashykt, line 59: was this meant to be fixed in this PR?

test/core/xds/xds_metadata_test.cc
```diff
@@ -0,0 +1,276 @@
+//
+// Copyright 2022 gRPC authors.
```
yashykt, line 2: 2024? looks like a new file with new contents?

yashykt (approved):

test/core/xds/xds_metadata_test.cc
markdroth: Whoops!  Yes, I meant to fix this as part of this PR, but I somehow managed to do it on the wrong branch.  I've added the fix.

test/core/xds/xds_metadata_test.cc
markdroth: I forked this from another file and kept the original date.

PR #37579, opened 2024-08-26
[EventEngine] Replace ApplicationCallbackExecCtx with EventEngine::Run

drfloob:
WIP / experiment.


yousukseung (approved):

src/core/lib/surface/completion_queue.cc
```diff
@@ -847,13 +852,22 @@ static void cq_end_op_for_callback(
     cq_finish_shutdown_callback(cq);
   }
 
+  auto* functor = static_cast<grpc_completion_queue_functor*>(tag);
+  if (grpc_core::IsEventEngineApplicationCallbacksEnabled()) {
+    // Run the callback on EventEngine threads.
+    cqd->event_engine->Run(
```
markdroth, line 858: Is this going to cause a thread hop that would affect performance?

I suspect it's probably fine, because if this runs in an EE thread, it will wind up getting run in the same thread (unless there's congestion, in which case the thread hop helps anyway), and if this doesn't run in an EE thread, the iomgr code would have caused a thread hop in that case anyway.

That having been said, it might be a good idea to confirm that that line of thinking is correct by running some benchmarks to verify that there's no noticeable performance impact.

markdroth (approved):

src/core/lib/surface/completion_queue.cc
drfloob: I ran a few benchmarks, and did not find any decent evidence of a difference in performance. The sensitivity of these benchmarks leaves a bit to be desired at the moment, due to recent large fluctuations, so statistical significance is not going to happen. We will roll this change out slowly, and be on alert for news of performance regressions.

PR #37580, opened 2024-08-27
[SourceTreeLayout] Move promise folder

tanvi-jagtap:
[SourceTreeLayout] Move promise folder from
`src/core/lib/promise` 
to
`src/core/util/promise`


src/core/util/promise/.clang-format
```diff
@@ -0,0 +1,50 @@
+---
```
tanvi-jagtap, line 1: Adding this to keep formatting same for the source and destination folder.

ctiller (approved):

PR #37581, opened 2024-08-27
[SourceTreeLayout] Moving call_destination.h, call_filters*, call_spi…

tanvi-jagtap:
Moving 
`call_destination.h`, `call_filters*`, `call_spine*` 
from
`scr/core/lib/transport/`
to 
`src/core/call/`

Question : Do we also want to move `lib/transport/call_final_info.h` ?


build_autogenerated.yaml
```diff
@@ -1095,10 +1098,7 @@ libs:
   - src/core/lib/surface/validate_metadata.h
   - src/core/lib/transport/bdp_estimator.h
   - src/core/lib/transport/call_arena_allocator.h
-  - src/core/lib/transport/call_destination.h
-  - src/core/lib/transport/call_filters.h
   - src/core/lib/transport/call_final_info.h
-  - src/core/lib/transport/call_spine.h
```
tanvi-jagtap, line 1101: Do we want to move lib/transport/call_final_info.h too?

sastryvp (approved):

build_autogenerated.yaml
markdroth: No, that file is part of the legacy filter API and will go away as we finish the migration to promises.

